PR #2781, opened 2020-10-30
QA: Make sure we stop salt-minion service on clean-up

srbarrios:
## What does this PR change?

Recently we updated our step definition to clean up minion's content, once we deleted the system from SUSE Manager.
We simplified it by using the `systemctl` command to stop the salt-minion service in all kinds of minions, while before we were using a different command for SLE minions. This change, cause a problem in our tests, as sometimes on SLE clients we stopped the service at the same time that SUMA was acting on it. In order to fix it, I changed the `run` command to another command that retries several times until it gives a successful result. Also, this step only removes the information from the cache, so it needs a rename to don't cause confusion. By the way, I use the verb "clean up" instead of the noun "cleanup" as it was before.

In addition, I would like to keep (commented) a TODO which explains better the reason why we are only cleaning the cache of the minion (so it doesn't try to connect to master again) instead of perform a full clean-up where we can uninstall the salt and salt-minion packages. I'm offering in the TODO description the step definition that we can use in case we unblock the reason why we can't use it ;)

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.1
- Manager-4.0

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/salt_steps.rb
+++ b/testsuite/features/step_definitions/salt_steps.rb
@@ -719,11 +719,28 @@ When(/^I enter "([^"]*)" password$/) do |host|
   step %(I enter "#{ENV['VIRTHOST_XEN_PASSWORD']}" as "password") if host == "xen_server"
 end
 
-And(/^I cleanup minion "([^"]*)"$/) do |minion|
+And(/^I clean up the minion's cache on "([^"]*)"$/) do |minion|
   raise "#{minion} is not a salt minion" unless minion.include? 'minion'
   node = get_target(minion)
-  node.run('systemctl stop salt-minion')
+  node.run_until_ok('systemctl stop salt-minion')
   node.run('rm -Rf /var/cache/salt/minion')
+  # TODO: Ideally we should remove /etc/salt/* /var/run/salt/* /var/log/salt/* and salt and salt-minion packages
+  #       But we can't do that as we don't have products synced, so it will fail installing salt and sal-minion
+  #       Instead we inject those packages when deploying through sumaform and we can't remove them.
+  #       If someday we have synced products, we can proceed to run a full cleanup:
+  #
+  # And(/^I perform a full salt minion cleanup on "([^"]*)"$/) do |minion|
+  #   raise "#{minion} is not a salt minion" unless minion.include? 'minion'
+  #   node = get_target(minion)
+  #   if minion.include? 'ceos'
+  #     node.run('yum -y remove salt salt-minion')
+  #   elsif minion.include? 'ubuntu'
+  #     node.run('apt-get --assume-yes remove salt salt-minion')
+  #   else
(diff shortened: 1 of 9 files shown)


ktsamis (approved):
Looks great, a renaming for now, later on with products synced we can uncomment the code and do the proper cleanup. :rocket:

testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -719,11 +719,28 @@ def pillar_get(key, minion)
...
 
-And(/^I cleanup minion "([^"]*)"$/) do |minion|
+And(/^I clean up the minion's cache on "([^"]*)"$/) do |minion|
   raise "#{minion} is not a salt minion" unless minion.include? 'minion'
   node = get_target(minion)
-  node.run('systemctl stop salt-minion')
+  node.run_until_ok('systemctl stop salt-minion')
   node.run('rm -Rf /var/cache/salt/minion')
+  # TODO: Ideally we should remove /etc/salt/* /var/run/salt/* /var/log/salt/* and salt and salt-minion packages
+  #       But we can't do that as we don't have products synced, so it will fail installing salt and sal-minion
```
Bischoff, line 728: "salt-minion"

In theory, we could re-enable original system repositories while we reinstall the packages... but... complicated, error-prone.

testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -719,11 +719,28 @@ def pillar_get(key, minion)
...
+  #   node = get_target(minion)
+  #   if minion.include? 'ceos'
+  #     node.run('yum -y remove salt salt-minion')
+  #   elsif minion.include? 'ubuntu'
+  #     node.run('apt-get --assume-yes remove salt salt-minion')
+  #   else
+  #     node.run('zypper --non-interactive remove -y salt salt-minion')
+  #   end
+  #   node.run('rm -Rf /var/cache/salt/minion /var/run/salt /var/log/salt /etc/salt')
+  # end
```
Bischoff, line 743: Maybe we could keep this snippet outside of this file, instead of a big comment.

But I don't mind too much either.

Bischoff (approved):
LGTM. See small nitpick.

testsuite/features/step_definitions/salt_steps.rb
srbarrios: If we keep outside.. it will be lost sooner or later, I will keep it.

testsuite/features/step_definitions/salt_steps.rb
srbarrios: I made a try about that in the past, and found some problems.
Anyhow, yeah, something to research. Better only clean the cache for now.

PR #2650, opened 2020-09-24
Feat: add support for i18n of user-facing strings

mbologna:
## What does this PR change?

Feat: add gettext translation hook to strings
Only user-facing strings have been touched and a gettext hook has been
added `_('...')`
Debug strings are left untouched

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: translation related

This page should be modified to remove spacecmd when this is merged: https://github.com/uyuni-project/uyuni/wiki/Translating-Uyuni-to-your-language#limitations

- [x] **DONE**

## Test coverage

- No tests: internal

- [X] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/12327

- [X] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_lint_checkstyle"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #2798, opened 2020-11-05
Internationalization of dnf-plugin-spacewalk

hustodemon:
## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: i18n

- [x] **DONE**

## Test coverage
- No tests: i18n

- [x] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/12328

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/client/rhel/dnf-plugin-spacewalk/po/dnf-plugin-spacewalk.pot
+++ b/client/rhel/dnf-plugin-spacewalk/po/dnf-plugin-spacewalk.pot
@@ -1,0 +1,71 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-11-05 13:12+0100\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=CHARSET\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: ../spacewalk.py:41
+msgid "Spacewalk based repositories will be disabled."
+msgstr ""
+
+#: ../spacewalk.py:42
+msgid "Spacewalk channel support will be disabled."
+msgstr ""
+
+#: ../spacewalk.py:43
+msgid "There was an error communicating with Spacewalk server."
+msgstr ""
+
+#: ../spacewalk.py:44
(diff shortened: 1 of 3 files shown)


hustodemon:
@mcalmer IIUC, it's enough to include the `pot` file. The `po` for each supported language will be brough by weblate. Is that correct?

mcalmer:
> @mcalmer IIUC, it's enough to include the `pot` file. The `po` for each supported language will be brough by weblate. Is that correct?

yes

mcalmer:
But you should "install" all translations. So check the spec file if it already collect language files.

hustodemon:
Thanks @mcalmer , will do.

client/rhel/dnf-plugin-spacewalk/po/dnf-plugin-spacewalk.pot
```diff
@@ -0,0 +1,71 @@
...
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-11-05 13:12+0100\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=CHARSET\n"
```
parlt91, line 17: I am not sure if weblate generates the charset for the `po` files from the `pot` file. But just to be sure could you `charset=UTF-8`?

parlt91 (approved):
Thanks. LGTM
Just one comment below

mbologna:
I am a bit confused: for other PRs we also have a Makevars file to declare the ugettext function:
https://github.com/uyuni-project/uyuni/pull/2650/files#diff-aa16b3e3d95f1d23b6c0fab8ab77a6cfb91a9b22901273ccc0b5d7f4cf79cf82
but in dnf-plugin-spacewalk I cannot find any gettext hook imported.

hustodemon:
@mbologna True, but where is this file used? Weblate? I couldn't find any usage in uyuni.

mbologna:
IIRC the Makevars instructs gettext about the method name (`_`) that is used in gettext calls

hustodemon:
Thanks @mbologna it's [true indeed](https://www.gnu.org/software/gettext/manual/html_node/po_002fMakevars.html). Although, according to gettext manual, the `_` seems to be explicitly used for python, it doesn't hurt to explicitly add it. We also do that for some other python projects in Uyuni (`spacecmd`).

mcalmer:
If `Makevars` is needed or not also depends on how the Makefile was written. Not sure where this was copied from.
We have also examples in this git where no Makevars file is needed.

client/rhel/dnf-plugin-spacewalk/po/dnf-plugin-spacewalk.pot
hustodemon: It seems `charset=CHARSET` is common for our `pot` files. I'd stick to that.

hustodemon:
@mcalmer The makefile was copied from other project from `client/` dir (I think it was `yum-rhn-plugin`). But the `Makefile` doesn't need to specify `Makevars` file - the file is taken into account even without it (I've tried running running `make update-po` while changing the `Makevars` content and the results differed).

PR #2845, opened 2020-11-12
QA: Refactor extraction of logs to support QAM clients

srbarrios:
## What does this PR change?

Small refactor to extract logs from all the active nodes in our testsuite. In that way, we fix the current failure on the QAM test suite at the same time that we don't need to worry if we add more clients to the list of nodes.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -324,16 +324,12 @@ When(/^I execute spacewalk-debug on the server$/) do
   raise "Download debug file failed" unless code.zero?
 end
 
-Then(/^I get logfiles from "([^"]*)"$/) do |target|
-  node = get_target(target)
-  os_version, os_family = get_os_version(node)
-  if os_family =~ /^opensuse/
-    node.run('zypper mr --enable os_pool_repo os_update_repo && zypper --non-interactive install tar')
+When(/^I extract the log files from all our active nodes$/) do
+  $nodes.each do |node|
+    next if node.nil?
+
+    extract_logs_from_node(node)
   end
-  node.run("journalctl > /var/log/messages && (tar cfvJP /tmp/#{target}-logs.tar.xz /var/log/ || [[ $? -eq 1 ]])")
-  `mkdir logs` unless Dir.exist?('logs')
-  code = file_extract(node, "/tmp/#{target}-logs.tar.xz", "logs/#{target}-logs.tar.xz")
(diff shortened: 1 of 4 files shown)


testsuite/features/support/commonlib.rb
```diff
@@ -228,3 +228,16 @@ def generate_repository_name(repo_url)
...
 end
+
+def extract_logs_from_node(node)
+  _os_version, os_family = get_os_version(node)
+  if os_family =~ /^opensuse/
+    node.run('zypper mr --enable os_pool_repo os_update_repo') unless $qam_test
+    node.run('zypper --non-interactive install tar')
+    node.run('zypper mr --disable os_pool_repo os_update_repo') unless $qam_test
+  end
+  node.run("journalctl > /var/log/messages && (tar cfvJP /tmp/#{node.full_hostname}-logs.tar.xz /var/log/ || [[ $? -eq 1 ]])")
```
srbarrios, line 239: This might fail in SLES11 nodes, right? :thinking:

Bischoff: No idea.

Bischoff (approved):
LGTM

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -324,16 +324,12 @@
...
-  os_version, os_family = get_os_version(node)
-  if os_family =~ /^opensuse/
-    node.run('zypper mr --enable os_pool_repo os_update_repo && zypper --non-interactive install tar')
-  end
-  node.run("journalctl > /var/log/messages && (tar cfvJP /tmp/#{target}-logs.tar.xz /var/log/ || [[ $? -eq 1 ]])")
-  `mkdir logs` unless Dir.exist?('logs')
-  code = file_extract(node, "/tmp/#{target}-logs.tar.xz", "logs/#{target}-logs.tar.xz")
-  raise "Download log archive failed" unless code.zero?
+When(/^I extract the log files from all our active nodes$/) do
+  $nodes.each do |node|
```
calancha, line 328: I don't think we have used until now a global variable `$nodes`.
This PR gives a good point to add one.

srbarrios: Thank you, indeed, missed.

PR #2848, opened 2020-11-13
Testsuite: Use an xml available in the centos minion

calancha:
## What does this PR change?
Use an XML that is available in the CentOS minion for the openscap scan.

### Ports
- Manager-4.0:  https://github.com/SUSE/spacewalk/pull/13121
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13120
## Changelogs

- [x] No changelog needed


Bischoff:
Thanks, not sure how I missed that.

PR #2858, opened 2020-11-17
Added RHEL8 compatibility

sbluhm:
## What does this PR change?

- Added compatibility for RHEL8.
- Removed RHEL6 logic.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: *No functional change.

- [X] **DONE**

## Test coverage
- No tests: Will be tested by auto-build.

- [X] **DONE**

## Links

Fixes #

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/uyuni/base/uyuni-base.spec
+++ b/uyuni/base/uyuni-base.spec
@@ -26,14 +26,20 @@
 %define apache_group www
 %else
 %define www_path %{_var}
+%if 0%{?rhel}
+%define apache_user root
+%define apache_group root
+%else
 %define apache_user apache
 %define apache_group apache
+%endif
 %endif
 
 Name:           uyuni-base
 Version:        4.2.1
 Release:        1
 Url:            https://github.com/uyuni-project/uyuni
+Source0:        %{name}-%{version}.tar.gz
 Summary:        Uyuni Base Package
 License:        GPL-2.0-only
 Group:          System/Fhs
@@ -98,10 +104,11 @@ mkdir -p %{buildroot}/%{_prefix}/share/rhn/config-defaults
 getent group susemanager >/dev/null || %{_sbindir}/groupadd -r susemanager
 getent passwd salt >/dev/null && %{_sbindir}/usermod -a -G susemanager salt
 getent passwd tomcat >/dev/null && %{_sbindir}/usermod -a -G susemanager tomcat
-getent passwd wwwrun >/dev/null && %{_sbindir}/usermod -a -G susemanager wwwrun
+getent passwd %{apache_user} >/dev/null && %{_sbindir}/usermod -a -G susemanager %{apache_user}
 %endif
 
 %files common
+%license LICENSE
 %defattr(-,root,root)
 %dir %attr(750,root,%{apache_group}) /etc/rhn
 %dir %{_prefix}/share/rhn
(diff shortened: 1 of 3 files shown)


sbluhm:
The difference in folder on line 119 is required for RHEL. Otherwise dnf will complain about folder conflicts (/var/spacewalk is used by multiple packages).

uyuni/base/uyuni-base.spec
```diff
@@ -26,14 +26,20 @@
...
 %define apache_user apache
 %define apache_group apache
 %endif
+%endif
 
 Name:           uyuni-base
 Version:        4.2.1
 Release:        1
 Url:            https://github.com/uyuni-project/uyuni
+Source:		noop.tar.gz
```
mcalmer, line 42: Is this empty tar really needed to build on RHEL?
(just try to learn something new)

sbluhm: Actually no. It is necessary for the Copr build process (not sure how OBS takes it on CentOS packages). Personally, I would like to keep it in to have that flexibility. Otherwise I can take it out.

mcalmer: @juliogonzalez any objections?

juliogonzalez: Not from me, but will autobuild accept it?

mcalmer: I think yes, but what about tito? 
What about adding a README and a LICENSE file and make it a standard package with tar like all the other?

juliogonzalez: Ah, right, tito would rewrite this.

I'd say it's better if we convert this to a standard package, yes.

sbluhm: Can you please guide me what to do? What is a standard package?

mcalmer: - copy the LICENSE file from uyuni/common-libs/ to uyuni/base/
- modify uyuni-base.spec
  * add `Source0:        %{name}-%{version}.tar.gz`
  * add `%license LICENSE` (to filelist `%files common`)

I think thats all.

sbluhm: Done.

uyuni/base/uyuni-base.spec
```diff
@@ -88,29 +94,34 @@ Basic filesystem hierarchy for Uyuni proxy.
 %install
 mkdir -p %{buildroot}/etc/rhn
 mkdir -p %{buildroot}/usr/share/rhn/proxy
-%if ! (0%{?rhel} == 6 || 0%{?suse_version} == 1110)
```
mcalmer, line 91: uyuni-base-common is used in osad.spec which also build for these old versions.
We have this bad situations that some packages contain client and server components.

So it is better to keep the rhel 6 test here until we really do not build for RHEL 6 clients anymore.
Here and the other 3 places in this file.

mcalmer (changes requested):

uyuni/base/uyuni-base.spec
sbluhm: Done.

mcalmer (approved):

PR #2867, opened 2020-11-18
Drop obsoleted ssl_available/--no-ssl/--use-ssl options

moio:
## What does this PR change?

Drops the configuration option to disable SSL, given AFAIU there is not really that possibility any more.

## GUI diff

Before:
before

After:
after

(sorry for the SUSE Manager branding)

- [x] **DONE**

## Documentation

-  [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pull/682)

- [ ] **DONE**

## Test coverage
- No tests: **no need to cover something dropped**
- [x] **DONE**

## Links

- [x] **DONE**

Fixes https://github.com/SUSE/spacewalk/issues/10858

## Re-run a test


mcalmer:
There are more places which say that SSL is possible. If we remove this option, we should do it everywhere.

Main other place is spacewalk-certs-tools wich generate the bootstrap scripts.
When you call mgr-bootstrap you have the option --no-ssl . This use a `noSSLServerURL` variable from `client-config-overrides.txt` which is generated on the server/proxy and downloaded to the clients.

I think we also need to think about updates. What happens to customers with old bootstrap scripts?

I think this change should only be merged in uyuni master (4.2) and not backported to 4.1.
We may need to document this properly. I think it needs a chapter in the release-notes for 4.2/uyuni .

moio:
> Main other place is spacewalk-certs-tools wich generate the bootstrap scripts.
> When you call mgr-bootstrap you have the option --no-ssl . This use a `noSSLServerURL` variable from `client-config-overrides.txt` which is generated on the server/proxy and downloaded to the clients.

This should now be clean with commit 8d84f6d.

> I think we also need to think about updates. What happens to customers with old bootstrap scripts?

If the bootstrap script was created with `--no-ssl`, I guess it will have to be regenerated without that option. And I guess the only possibility for us is to document that, right?

> I think this change should only be merged in uyuni master (4.2) and not backported to 4.1.

Definitely, this is for master only.

> We may need to document this properly. I think it needs a chapter in the release-notes for 4.2/uyuni .

Anything else other than what described above?

mcalmer (approved):
LGTM: check the test results - not sure if the error is related to your change.

moio:
@juliogonzalez do we have a place for release note collection?

juliogonzalez:
Since this is neither for 2020.11 or 4.1.4, not yet. Ping me again on Monday so I create the cards and you can mention this on the cards.

juliogonzalez:
@moio 

https://github.com/SUSE/spacewalk/issues/13227
https://github.com/SUSE/spacewalk/issues/13228
https://github.com/SUSE/spacewalk/issues/13230

moio:
@mcalmer I added clean ups both for https://github.com/SUSE/spacewalk/issues/10858 and another option I found during the cleanups (`--use-ssl`).

As the removals were fairly straightforward I am proceeding with merging as soon as tests are green - unless you object of course.

moio:
All tests green, rebasing one last time to fix changelogs and merging

PR #2869, opened 2020-11-18
Fix mandatory channels JS API to finish loading in case of error (bsc#1178839)

cbbayburt:
CLM "add sources" panel is stuck with "Loading" when the mandatory channels API returns an error. This should not prevent users from working with CLM. This change fixes the API to go out of loading state when the request finishes with any result.

https://bugzilla.suse.com/1178839

Port of https://github.com/SUSE/spacewalk/pull/13150

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/spacewalk-java.changes
+++ b/java/spacewalk-java.changes
@@ -1,3 +1,5 @@
+- Update exception message in findSyncedMandatoryChannels
+
 -------------------------------------------------------------------
 Wed Nov 25 12:22:07 CET 2020 - jgonzalez@suse.com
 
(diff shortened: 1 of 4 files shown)


java/spacewalk-java.changes
```diff
@@ -1,3 +1,4 @@
+- Update exception message in findSyncedMandatoryChannels
```
juliogonzalez, line 1: Why is the bsc missing in this entry?

cbbayburt: It's intentional, the actual fix is only in `spacewalk-web`

juliogonzalez: Thanks for the clarification :-)

PR #2873, opened 2020-11-19
Testsuite: Fix srv_content_lifecycle tests

calancha:
## What does this PR change?
Fix errors at srv_content_lifecycle tests about unclickable buttons.

Search for the text messages in the specific environment where
they appear.
Wait until a target button is enabled before to click it.

- testsuite/features/secondary/srv_content_lifecycle.feature:
Ensure we search for the expected text in the right environment.

Add step to wait a custom number of seconds until a target string
shows up in the target envitonment.

### Ports:
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13168
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/13167

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -1187,20 +1187,26 @@ end
 
 # content lifecycle steps
 When(/^I click the environment build button$/) do
-  raise "Click on environment build failed" unless find(:xpath, '//*[@id="cm-build-modal-save-button"]').click
+  raise 'Click on environment build failed' unless find_button('cm-build-modal-save-button', disabled: false, wait: DEFAULT_TIMEOUT).click
 end
 
 When(/^I click promote from Development to QA$/) do
-  raise "Click on promote from Development failed" unless find(:xpath, '//*[@id="dev_name-promote-modal-link"]').click
+  raise 'Click on promote from Development failed' unless find_button('dev_name-promote-modal-link', disabled: false, wait: DEFAULT_TIMEOUT).click
 end
 
 When(/^I click promote from QA to Production$/) do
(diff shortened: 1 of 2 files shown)


testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1187,15 +1187,17 @@ def server_secret
 
 # content lifecycle steps
 When(/^I click the environment build button$/) do
-  raise "Click on environment build failed" unless find(:xpath, '//*[@id="cm-build-modal-save-button"]').click
+  raise 'Click on environment build failed' unless find(:xpath, '//*[@id="cm-build-modal-save-button"]').click
```
srbarrios, line 1190: Just to easy maintain those steps, I would use the same method to click in all of them, so if find_button works, let's use it in the three methods.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1187,15 +1187,17 @@ def server_secret
...
 end
 
 When(/^I click promote from Development to QA$/) do
-  raise "Click on promote from Development failed" unless find(:xpath, '//*[@id="dev_name-promote-modal-link"]').click
+  step %(I wait until button "dev_name-promote-modal-link" becomes enabled)
```
srbarrios, line 1194: I wonder if just to set a "wait: XXX" parameter in the find method, with a big value, could make the same result avoiding one extra step.

srbarrios (approved):
Some comments made, but if it's tested and working, I'm fine to merge it like that.

testsuite/features/step_definitions/common_steps.rb
calancha: It sounds good; I am going to investigate it.

PR #2880, opened 2020-11-20
Fix openSCAP results

Bischoff:
## What does this PR change?

The result of the openSCAP audit may list many not applicable rules. They make the test for passing rules difficult as they may start on page 2. This PR uses the filter button to focus on the passing rules.

Now we will install all needed packages from the test suite, instead of a mixture of sumaform and the test suite (see uyuni-project/sumaform#801). This PR adapts.

The list of packages to install for openSCAP tests was wrong, For example, the CentOS traditional client is both a CentOS client and a traditional client. Constant package lists were not fit for this, this PR computes the list dynamically.

This PR also cleans up the generated reports, the installed packages and the enabled repositories at the end to make these tests idempotent.

## Links

Ports:
* 4.1: SUSE/spacewalk#13187 SUSE/spacewalk#13189 SUSE/spacewalk#13195 SUSE/spacewalk#13197
* 4.0: SUSE/spacewalk#13188 SUSE/spacewalk#13190 SUSE/spacewalk#13194 SUSE/spacewalk#13196

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/secondary/trad_openscap_audit.feature
+++ b/testsuite/features/secondary/trad_openscap_audit.feature
@@ -23,8 +23,9 @@ Feature: openSCAP audit of traditional client
     Then I should see a "Details of XCCDF Scan" text
     And I should see a "Default" text
     And I should see a "XCCDF Rule Results" text
-    And I should see a "pass" text or "notapplicable" text
-    And I should see a "rule-" link
+    When I enter "pass" as the filtered XCCDF result type
+    And I click on the filter button
+    Then I should see a "rule-pwd-warnage" link
 
   Scenario: Cleanup: remove audit scans retention period from traditional client
     Given I am on the Organizations page
--- a/testsuite/features/secondary/trad_openscap_audit.feature
+++ b/testsuite/features/secondary/trad_openscap_audit.feature
@@ -1,12 +1,17 @@
 # Copyright (c) 2015-2020 SUSE LLC
 # Licensed under the terms of the MIT license.
 
-Feature: openSCAP audit of traditional client
+Feature: OpenSCAP audit of traditional client
   In order to audit a traditional client
   As an authorized user
-  I want to run an openSCAP scan on it
+  I want to run an OpenSCAP scan on it
 
-  Scenario: Schedule an audit job using the SUSE profile
+  Scenario: Install the OpenSCAP packages on the traditional client
+    When I enable repository "os_pool_repo os_update_repo" on this "sle_client"
+    And I enable SUSE Manager tools repositories on "sle_client"
+    And I install OpenSCAP dependencies on "sle_client"
+
+  Scenario: Schedule an OpenSCAP audit job on the traditional client using SUSE profile
     Given I am on the Systems overview page of this "sle_client"
     When I follow "Audit" in the content area
     And I follow "Schedule" in the content area
@@ -51,3 +56,8 @@ Feature: openSCAP audit of traditional client
     And I enter "90" as "scap_retention_period"
     And I click on "Update Organization"
     Then I should see a "Organization SUSE Test was successfully updated." text
+
+  Scenario: Cleanup: remove the OpenSCAP packages from the traditional client
+    When I remove OpenSCAP dependencies from "sle_client"
+    And I disable SUSE Manager tools repositories on "sle_client"
+    And I disable repository "os_pool_repo os_update_repo" on this "sle_client"
--- a/testsuite/features/secondary/trad_openscap_audit.feature
+++ b/testsuite/features/secondary/trad_openscap_audit.feature
@@ -35,16 +35,16 @@ Feature: openSCAP audit of traditional client
     And I click on "Update Organization"
     Then I should see a "Organization SUSE Test was successfully updated." text
 
-  Scenario: Delete audit results from traditional client
+  Scenario: Cleanup: delete audit results from traditional client
     Given I am on the Systems overview page of this "sle_client"
     When I follow "Audit" in the content area
     And I follow "List Scans" in the content area
     And I click on "Select All"
     And I click on "Remove Selected Scans"
     And I click on "Confirm"
-    Then I should see a "deleted. 0 SCAP Scan(s) retained" text
+    Then I should see a " SCAP Scan(s) deleted. 0 SCAP Scan(s) retained" text
 
-  Scenario: Restore audit scans retention period on traditional client
+  Scenario: Cleanup: restore audit scans retention period on traditional client
     Given I am on the Organizations page
     When I follow "SUSE Test" in the content area
     And I follow "Configuration" in the content area
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -659,7 +659,7 @@ When(/^I enable IPv6 forwarding on all interfaces of the SLE minion$/) do
   $minion.run('sysctl net.ipv6.conf.all.forwarding=1')
 end
 
-When(/^I wait for the openSCAP audit to finish$/) do
+When(/^I wait for the OpenSCAP audit to finish$/) do
   host = $server.full_hostname
   @sle_id = retrieve_server_id($minion.full_hostname)
   @cli = XMLRPC::Client.new2('http://' + host + '/rpc/api')
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -771,17 +771,30 @@ When(/^I remove pattern "([^"]*)" from this "([^"]*)"$/) do |pattern, host|
 end
 
 When(/^I (install|remove) the traditional stack utils (on|from) "([^"]*)"$/) do |action, where, host|
-  step %(I #{action} packages "#{TRADITIONAL_STACK_RPMS}" #{where} this "#{host}")
+  pkgs = 'spacewalk-client-tools spacewalk-check spacewalk-client-setup mgr-daemon mgr-osad mgr-cfg-actions'
+  step %(I #{action} packages "#{pkgs}" #{where} this "#{host}")
 end
 
-When(/^I (install|remove) OpenSCAP (traditional|salt|centos) dependencies (on|from) "([^"]*)"$/) do |action, client_type, where, host|
-  if client_type == 'traditional'
-    step %(I #{action} packages "#{OPEN_SCAP_TRAD_DEPS}" #{where} this "#{host}")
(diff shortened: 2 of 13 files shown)


testsuite/features/support/constants.rb
```diff
@@ -197,12 +197,3 @@
...
                       'ubuntu2004_minion' => 'x86_64' }.freeze
-
-TRADITIONAL_STACK_RPMS = 'spacewalk-client-tools spacewalk-check spacewalk-client-setup '\
-                         'mgr-daemon mgr-osad mgr-cfg-actions'.freeze
-
-OPEN_SCAP_CENTOS_DEPS = 'spacewalk-oscap scap-security-guide'.freeze
-
-OPEN_SCAP_TRAD_DEPS = 'spacewalk-oscap'.freeze
-
-OPEN_SCAP_SALT_DEPS = 'openscap-utils openscap-content scap-security-guide'.freeze
```
srbarrios, line 208: I don't like the idea to remove these constants.
I prefer to have outside from a method this information, as it's easier to tweak and handle it.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -729,6 +729,32 @@
...
+  node.run("sed -i '/^#\\s*deb http:\\/\\/archive.ubuntu.com\\/ubuntu .* universe/ s/^#\\s*deb /deb /' /etc/apt/sources.list")
+  node.run("apt-get update")
+end
+
+When(/^I disable universe repositories on "([^"]*)"$/) do |host|
+  node = get_target(host)
+  node.run("sed -i '/^deb http:\\/\\/archive.ubuntu.com\\/ubuntu .* universe/ s/^deb /# deb /' /etc/apt/sources.list")
+  node.run("apt-get update")
+end
+
```
srbarrios, line 757: I'd prefer if we join these two steps definitions that are doing the same, but with the opposite result

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -729,6 +729,32 @@
...
+  if os_family =~ /^opensuse/ || os_family =~ /^sles/
+    repos, _code = node.run('zypper lr | grep "tools" | cut -d"|" -f2')
+    node.run("zypper mr --disable #{repos.gsub(/\s/, ' ')}")
+  elsif os_family =~ /^centos/
+    repos, _code = node.run('yum repolist enabled 2>/dev/null | grep "tools" | cut -d" " -f1')
+    repos.gsub(/\s/, ' ').split.each do |repo|
+      node.run("sed -i 's/enabled=.*/enabled=0/g' /etc/yum.repos.d/#{repo}.repo")
+    end
+  end
+end
```
srbarrios, line 744: I hope we never pass Sonarcube in our code, as we are adding a lot of duplicated code :S 
I would also refactor this code in the manner you wish, to don't have two "step definitions" doing exactly the same behavior but with the opposite result.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -771,17 +771,30 @@
...
+  step %(I #{action} packages "#{pkgs}" #{where} this "#{host}")
+end
+
+# On CentOS 7, OpenSCAP files are for RedHat and need a small adaptation for CentOS
+When(/^I fix CentOS 7 OpenSCAP files on "([^"]*)"$/) do |host|
+  node = get_target(host)
+  script = '/<\/rear-matter>/a  <platform idref="cpe:/o:centos:centos:7"/>'
+  file = "/usr/share/xml/scap/ssg/content/ssg-rhel7-xccdf.xml"
+  node.run("sed -i '#{script}' #{file}")
 end
```
srbarrios, line 798: Does it only happen in CentOS7? It doesn't happen on CentOS 8?

testsuite/features/support/constants.rb
Bischoff: I knew you would say that. I still don't understand how having things at different places helps. Having them where they are needed avoid me losing overview.

Said overwise, a constant (or a method) that is used only once does not really make sense to me.

testsuite/features/secondary/trad_openscap_audit.feature
```diff
@@ -34,19 +40,24 @@ Feature: openSCAP audit of traditional client
...
-  Scenario: Delete audit results from traditional client
+  Scenario: Cleanup: delete audit results from traditional client
     Given I am on the Systems overview page of this "sle_client"
     When I follow "Audit" in the content area
     And I follow "List Scans" in the content area
     And I click on "Select All"
     And I click on "Remove Selected Scans"
     And I click on "Confirm"
-    Then I should see a "deleted. 0 SCAP Scan(s) retained" text
+    Then I should see a "1 SCAP Scan(s) deleted. 0 SCAP Scan(s) retained" text
```
srbarrios, line 50: These kinds of assumptions are very problematic for idempotency and it's a source of issues if we decide to run the tests in parallel. Please keep that in mind.

testsuite/features/step_definitions/common_steps.rb
Bischoff: I don't think that helps readability and I prefer it like this. But if you wish to "fix" this someday I won't oppose either.

testsuite/features/secondary/trad_openscap_audit.feature
```diff
@@ -23,8 +28,9 @@ Feature: openSCAP audit of traditional client
     Then I should see a "Details of XCCDF Scan" text
     And I should see a "Default" text
     And I should see a "XCCDF Rule Results" text
-    And I should see a "pass" text or "notapplicable" text
-    And I should see a "rule-" link
+    When I enter "pass" as the filtered XCCDF result type
+    And I click on the filter button
+    Then I should see a "rule-pwd-warnage" link
```
srbarrios, line 33: Unless "rule-pwd-warnage" text is visible on the UI and for the user, I'd prefer to have a step definition hiding this technical detail.

testsuite/features/step_definitions/common_steps.rb
Bischoff: Same, feel free to do it. I don't think it helps readability, and strictly speaking it slows down execution a bit, but ok.

testsuite/features/step_definitions/command_steps.rb
Bischoff: This is an interesting and open question. In the doubt, I took a conservative approach and chose  to use it on CentOS 7 only. We can add it to CentOS 8 too if we see `notapplicable` only.

I am sure it happens in CentOS 6 because of https://forums.centos.org/viewtopic.php?t=50462 .

It also happens in Ubuntu !!! But I did not find the correct fix and gave up after one hour searching. I added a `TODO` in the code as a way to remember the issue.

testsuite/features/secondary/trad_openscap_audit.feature
```diff
@@ -34,19 +40,24 @@ Feature: openSCAP audit of traditional client
...
     When I follow "SUSE Test" in the content area
     And I follow "Configuration" in the content area
     And I enter "90" as "scap_retention_period"
     And I click on "Update Organization"
     Then I should see a "Organization SUSE Test was successfully updated." text
+
+  Scenario: Cleanup: remove the openSCAP packages from the traditional client
+    When I remove OpenSCAP dependencies from "sle_client"
+    And I disable SUSE Manager tools repositories on "sle_client"
+    And I disable repository "os_pool_repo os_update_repo" on this "sle_client"
```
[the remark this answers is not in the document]
Bischoff: No, we do hide this, but for other stuff : Docker, Prometeus, etc.

To be noticed: the existing code was not hiding the details either (`And I enable repository "CentOS Base"`).

I tried to hide it the same way:

`And I install packages before testing OpenSCAP`

but it resulted complicated, especially due to the fact we reuse steps like `I disable SUSE Manager tools repositories` that are already doing part of the hiding, but are also used standalone for bootstrapping.

I agree this refactor would be beneficial, but I also already lost too much time on openSCAP, which is somehow marginal.

Feel free to write these "hinding" steps. Would be great.

testsuite/features/secondary/trad_openscap_audit.feature
Bischoff: Agree. Will remove it everywhere. Was done to align, but I aligned the wrong way round.

testsuite/features/step_definitions/common_steps.rb
srbarrios: Using `add-apt-repository` on Ubuntu20, will be safer and will simplify it in a step.
If we use Ubuntu16 or Ubuntu18, we just need to first install `software-properties-common`

testsuite/features/secondary/trad_openscap_audit.feature
Bischoff: It is visible on UI and therefore easy to check when we get screenshots.

testsuite/features/step_definitions/common_steps.rb
srbarrios: :laughing: Really do you feel the time of execution will change more than a microsecond?
I would prefer we refactor those steps to use both enable/disable options, but I will not go behind other's code refactoring it in another PR...

testsuite/features/secondary/trad_openscap_audit.feature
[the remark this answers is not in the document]
srbarrios: As I said, I will not go behind others' code refactoring it. If you really feel it's ok like that, I will not oppose, I'm just giving my personal opinion :)

testsuite/features/step_definitions/command_steps.rb
srbarrios: Ok, keep in mind it, and maybe add a comment in the card to upgrade our CI to CentOS8, as that could emerge at some moment.

testsuite/features/support/constants.rb
srbarrios: That's a lost battle, you will never agree with me in relation to encapsulation and isolation, so... go ahead.

srbarrios (approved):
Giving approval, but I left a few remarks because I don't fully agree with the way we handle some things on that PR. Nothing that I can't live with.

testsuite/features/secondary/trad_openscap_audit.feature
Bischoff: Done.

Bischoff: As a side note, it could not have created problems with parallelism, as it happens on different clients, and each has its own audit screen. But OK.

PR #2882, opened 2020-11-21
mgr-setup: Added RHEL/CentOS compatibility.

sbluhm:
## What does this PR change?

- Added RHEL/CentOS Java alternatives options.
- Create PostgreSQL Datadir should it not exist.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: No functional changes.

- [X] **DONE**

## Test coverage
- No tests: Don't know how...

- [X] **DONE**

## Links

Fixes #

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


susemanager/bin/mgr-setup
```diff
@@ -181,7 +182,17 @@ cleanup_hostname() {
 }
 
 setup_db_postgres() {
+    DATADIR="/var/lib/pgsql/data" 
     systemctl --quiet enable postgresql 2>&1
+    # Create the PostgreSQL data folder, should it not exist.
+    if [ ! -f $DATADIR/PG_VERSION ]; then
```
mcalmer, line 188: Please add here also a check for not SUSE.
On SUSE systems this happens when starting the DB in the systemd start script (`systemctl start postgresql`).

You can check via /etc/os-release that ID_LIKE contains "suse":

```
if [ -e /etc/os-release ]; then
   . /etc/os-release
   if [[ "$ID_LIKE" = *"suse"* ]]; then
      echo "Do something for any SUSE/openSUSE distro"
   fi
fi
```

mcalmer (changes requested):
A small check for SUSE distro is required

susemanager/bin/mgr-setup
sbluhm: I had two thoughts for not making this distro dependent:
1. The folder would already exist for SUSE. Or is postgresql service started only AFTER mgr-setup?
2. It is the same code as in postgresql.service. This is where I got it from. So it wouldn't make a difference what creates it.

Overall adding the distro check shouldn't be more than 2 lines (if/fi) as it is already done at the top of mgr-setup for Java.

sbluhm: I have made the change in my branch and tested it. If you still insist, let me know and I will push it.

mcalmer: > 1. The folder would already exist for SUSE. Or is postgresql service started only AFTER mgr-setup?

I think mgr-setup start postgresql the first time. 

> 2. It is the same code as in postgresql.service. This is where I got it from. So it wouldn't make a difference what creates it.

Yes, but I fear times when something must be changed in the init process and we do not get the info about it.
And the init code here on my system looks already a bit different then the one in this PR.
Especially exporting the correct environment is important. We had in the past a lot of fun with it.

If you don't mind I would like to use the initialization provided by the OS vendor.

sbluhm: Pushed.

mcalmer (approved):
thanks

PR #2899, opened 2020-11-24
Testsuite: prevent special chars in the output xml files

calancha:
## What does this PR change?
Prevent special chars in the output xml files.

Otherwise, they break the `minidom` parser and the CI won't send
the mail with the results.
Do not pass `--force-color` option to the `salt-call`.

### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13225

## Changelogs

- [x] No changelog needed


srbarrios (approved):
:clap:

PR #2921, opened 2020-11-27
Fix file download links

ricardoasmarques:
## What does this PR change?

This PR fixes file download links that are broken since `senna.js` was introduced.

This PR is a follow up of https://github.com/uyuni-project/uyuni/pull/2649 that fixes a similar issue for `CSV` downloads.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: This is just a fix

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1179324
Tracks: https://github.com/SUSE/spacewalk/pull/13428
Tracks: https://github.com/SUSE/spacewalk/pull/13431

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/webapp/WEB-INF/pages/common/fragments/configuration/files/contents.jspf
```diff
@@ -27,8 +27,9 @@
                         <bean:message key="filedetails.contents.jspf.toolarge-edit" arg0="${requestScope.maxEditBytes}"/>
 
                         <strong><bean:message key="filedetails.contents.jspf.download-header"/></strong><br />
+                        <%-- TODO Shouldn't we add the 'data-senna-off="true"' on 'StringResource_*.xml' files directly, instead of injecting it here? --%>
```
ricardoasmarques, line 30: @ncounter @moio We should discuss this next week.

java/code/webapp/WEB-INF/pages/common/fragments/configuration/files/contents.jspf
```diff
@@ -28,7 +28,7 @@
 
                         <strong><bean:message key="filedetails.contents.jspf.download-header"/></strong><br />
                         <bean:message key="filedetails.contents.jspf.download"
-                                arg0="/rhn/configuration/file/FileDownload.do?crid=${revision.id}&amp;cfid=${revision.configFile.id}"
+                                arg0="/rhn/configuration/file/FileDownload.do?crid=${revision.id}&amp;cfid=${revision.configFile.id}\" data-senna-off=\"true"
```
ricardoasmarques, line 31: Shouldn't we add the `data-senna-off="true"` on `StringResource_*.xml` files directly, instead of injecting it here?

ricardoasmarques: `data-senna-off="true"` was now moved to `StringResource_en_US.xml`.

moio:
@ricardoasmarques I am fine with the patch, please complete the PR complete (description).

@ncounter can you have a look from your side?

It would also be good to add a Cucumber test, more as an occasion to get familiar with the end-to-end testsuite than because of an absolute need for this PR.

The Cucumber testsuite requires a separate set of VMs that can be deployed via sumaform by following these instructions: https://github.com/uyuni-project/sumaform/blob/master/README_TESTING.md#cucumber-testsuite

Once deployed, you will need to run the so-called "core features" and, assuming all is green, you can start developing a new feature. I would recommend taking VM snapshots before and after to quickly roll back when needed. Feel free to ping for more guidance.

ncounter (approved):
Thanks for the complete fix @ricardoasmarques , I went through it once, IIRC another colleague went through it as well, but it seems we left something out. I agree with @moio suggestions, and it LGTM :+1:

ricardoasmarques:
@moio I've now updated this PR to include a Cucumber test that will check if the configuration file download link is working, please take another look.

moio (approved):
LGTM. @lkotek can you take a very quick look at test code?

lkotek (approved):
@moio Yep, sure. I did moment ago. IIRC we never passed variables in testsuite this way, but I have no problem with that if it is tested and works :-)

testsuite/features/secondary/min_action_chain.feature
```diff
@@ -92,6 +92,16 @@ Feature: Action chains on Salt minions
...
     And I should see a "Update Configuration File" button
 
+  Scenario: Download the configuration file from configuration channel
+    Given I am authorized as "admin" with password "admin"
+    When I follow the left menu "Configuration > Channels"
+    And I follow "Action Chain Channel"
+    And I follow "List/Remove Files"
+    And I follow "/etc/action-chain.cnf"
+    And I follow "Download File"
+    And I wait until file "${PWD}/action-chain.cnf" exists on "localhost"
```
ricardoasmarques, line 102: I think `${HOME}` is more correct here.. I'll test it and update PR accordingly if works.

ricardoasmarques: I've not changed location from `${PWD}` to `${HOME}`.

PR #2926, opened 2020-11-29
spacewalk-utils: RHEL8 changes. Python changes for pylint.

sbluhm:
## What does this PR change?

- Updated Python files to pass the pylint check.
- Updated SPEC file for RHEL8.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: Spec file is tested during auto build.
Not sure if python tests are necessary. Please advice if tests should be created (and how...)

Build tested sucessful on CentOS 8 and LEAP 15.2

- [X] **DONE**

## Links

Fixes #

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


utils/cloneByDate.py
```diff
@@ -57,10 +55,7 @@
...
 def confirm(txt, options):
     if not options.assumeyes:
-        if six.PY2:
-            inputfn = raw_input
-        else:
-            inputfn = input
+        inputfn = input()
         response = inputfn(txt)
         while ['y', 'n'].count(response.lower()) == 0:
             response = inputfn(txt)
```
meaksh, line 61: I think this is not correct. Apparently what you want would be:
```suggestion
        inputfn = input
        response = inputfn(txt)
        while ['y', 'n'].count(response.lower()) == 0:
            response = inputfn(txt)
```


But, since you're dropping the Python 2 compatibility, then it does not make sense anymore to have an `inputfn` alias. Therefore I think we should go like:

```suggestion
        response = input(txt)
        while ['y', 'n'].count(response.lower()) == 0:
            response = input(txt)
```

meaksh (changes requested):
Thanks for the PR @sbluhm ! I've added a comment so you can take a look.

Apart from that, rest of the changes look good.

utils/cloneByDate.py
sbluhm: I did not look left or right/up or down....

meaksh (approved):
LGTM :+1:
Thanks!

PR #2927, opened 2020-11-29
Fix incorrect password autocompletion on older Firefox

Etheryte:
## What does this PR change?

Passwords are no longer autocompleted where inappropriate, for example `/rhn/admin/setup/MirrorCredentials.do`. Chrome already didn't do this, but some older versions of Firefox did.  
This is the best spec-compliant option we have, if the browser doesn't respect this, there isn't much else we can do.  
There is a history of a lot of controversy around this and vendors play around with this a fair bit.  
See:
 - https://stackoverflow.com/q/12374442/1470607  
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1119063  

## GUI diff

GUI itself unchanged, but passwords are no longer autocompleted where inappropriate.

- [x] **DONE**

## Documentation
- No documentation needed: No documentation update, fixes a bug that wasn't reproducible on the latest build of Firefox, but was present on an older version. Couldn't repro on other browsers.

- [x] **DONE**

## Test coverage
- No tests: Practical functionality unchanged.

- [x] **DONE**

## Links

Fixes [bsc#1148357](https://bugzilla.suse.com/show_bug.cgi?id=1148357), [SUSE/spacewalk#9285](https://github.com/SUSE/spacewalk/issues/9285)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


Etheryte:
As a related question, is it intentional that no `autocomplete` is set on the log in password field in `/web/html/src/manager/login/login.js`?

hustodemon:
> As a related question, is it intentional that no autocomplete is set on the log in password field in /web/html/src/manager/login/login.js?

Honestly, I don't know. I just tried in 2 browsers (FF 78, Chromium 86) and got 2 different results: whereas Firefox auto-completed my credentials, Chromium didn't do it (I even tried to turn change the settings that turn chromium more friendly to the self-signed certificates, but still couldn't persuade the browser to auto-complete the data). I think the behavior should be consistent - either we allow autocompletion or we don't...

hustodemon (commented):
Thanks for the PR @Etheryte !
Left an opinion about the auto-completion behavior.
Moreover: the changelog file is incorrect: the files changed belong to packages with these changelogs: `spacewalk-java.changes` and `spacewalk-web.changes`. This is quite confusing, I agree :)

Otherwise the PR LGTM!

ncounter (approved):
:+1: LGTM.


About the login question:
> As a related question, is it intentional that no autocomplete is set on the log in password field in /web/html/src/manager/login/login.js?

No reason I am aware of actually. I don't know if there are security limitations, but I don't really think this is the case. Maybe @mbologna knows more as he worked on the SSO prototype a while ago?

PR #2932, opened 2020-11-30
[Port] QAM: Fix extraction of logs from old clients not having /etc/os-relea…

srbarrios:
## What does this PR change?

On this PR we treat differently the way to obtain family and version from a CentOS6 node. After analyzing all the current supported nodes in our test suite, we noticed that only CentOS6 has not `/etc/os-release` file. Instead, it has a different file with a different format.

We decided that while we don't integrate other distros not supporting `/etc/os-release`, as it is a unique case, as soon as we notice it exist a file `/etc/centos-release` we can set the values for family and version to `centos` and `6`.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
 - Manager-4.1 https://github.com/SUSE/spacewalk/pull/13134

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/.rubocop_todo.yml
+++ b/testsuite/.rubocop_todo.yml
@@ -362,10 +362,7 @@ Style/DateTime:
 # Offense count: 4
 # Cop supports --auto-correct.
 Style/DisableCopsWithinSourceCodeDirective:
-  Exclude:
-    - 'features/step_definitions/common_steps.rb'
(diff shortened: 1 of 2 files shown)


Bischoff:
Thanks.

PR #2944, opened 2020-11-30
spacewalk-cert-tools: Added RHEL8 build and simplified SPEC file.

sbluhm:
## What does this PR change?

Added RHEL8 build support.
Minor simplification of the SPEC file.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: Tested during automated build.

Tested the build successfully on CentOS 8, Fedora 31+, LEAP 15.2

- [X] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacewalk/certs-tools/spacewalk-certs-tools.spec
+++ b/spacewalk/certs-tools/spacewalk-certs-tools.spec
@@ -31,10 +31,13 @@
 %endif
 %global rhnroot %{_datadir}/rhn
 
-%if 0%{?fedora} || 0%{?suse_version} > 1320
+%global __python /usr/bin/python2
+
+%if 0%{?fedora} || 0%{?suse_version} > 1320 || 0%{?rhel} >= 8
 %global build_py3   1
 %global default_py3 1
 %endif
+
 
 %define pythonX %{?default_py3: python3}%{!?default_py3: python2}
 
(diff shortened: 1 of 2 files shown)


spacewalk/certs-tools/spacewalk-certs-tools.spec
```diff
@@ -84,13 +86,16 @@ Spacewalk.
...
 Group:          Applications/Internet
+BuildRequires:  (python or python2)
 Requires:       %{name} = %{version}-%{release}
 Requires:       python2-rhn-client-tools
 Requires:       python2-uyuni-common-libs
-%if 0%{?rhel} && 0%{?rhel} <= 5
-Requires:       python-hashlib
-%endif
+%if 0%{?suse_version} || 0%{?rhel}
+BuildRequires:  python2
```
mcalmer, line 94: Seems we have no this BuildRequires 2 times.

In Line 89 you added `BuildRequires:  (python or python2)` and here we have it again with using macros.
Or should be be just a `Requires` ?

mcalmer (changes requested):
Either a duplicate dependency or used BuildRequires vs Requires. Please check.

spacewalk/certs-tools/spacewalk-certs-tools.spec
sbluhm: True, I forgot to remove line 89. This took me a long time to sort out and identify the issue.
It seems that if you just require "python" on Leap, it replaces it with a preferred version and therefore never installs python 2 in my test case. So line 89 unfortunately will not work for python2 as a simplification.
I will remove line 89.

mcalmer (approved):

PR #2952, opened 2020-12-01
Use sanitized repo label to build reposync repo cache path (bsc#1179410)

meaksh:
## What does this PR change?

This PR makes fixes some issues we have in `spacewalk-repo-sync` when dealing with custom DEB repositories that contains non-valid characters on its label (like `/`).

The summary of changes is:
- Add validation to custom repository label (same as channel name): allow uppercase, spaces, slashes, etc.
- Fix `deb_src` reposync plugin to use a sanitized "repository label" when building the repository cache root path.
- Fix `yum_src` reposync plugin to use a sanitized "repository label" (instead of channel label) when building the repository cache root path.

These are the current regex we use for validation:
```java
CHANNEL_NAME_REGEX = "^[a-zA-Z\\d][\\w\\d\\s\\-\\.\\'\\(\\)\\/\\_]*$"
CHANNEL_LABEL_REGEX = "^[a-z\\d][a-z\\d\\-\\.\\_]*"
REPO_LABEL_REGEX = "^[a-zA-Z\\d][\\w\\d\\s\\-\\.\\'\\(\\)\\/\\_]*$"
```

So, with this PR, reposync will use a sanitized "repository label" (which only allows alphanumerical characters) to build the reposync repo cache path for both RPM and DEB reposync plugins.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **bugfix**

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/13294

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/channel/test/BaseRepoCommandTest.java
+++ b/java/code/src/com/redhat/rhn/manager/channel/test/BaseRepoCommandTest.java
@@ -194,6 +194,7 @@ public class BaseRepoCommandTest extends RhnBaseTestCase {
 
         try {
             ccc.store();
+            fail("invalid repository label should have thrown error: " + label);
         }
         catch (InvalidRepoUrlException e) {
             fail("non duplicate url caused error");
--- a/java/code/src/com/redhat/rhn/manager/channel/test/BaseRepoCommandTest.java
+++ b/java/code/src/com/redhat/rhn/manager/channel/test/BaseRepoCommandTest.java
@@ -130,4 +130,82 @@ public class BaseRepoCommandTest extends RhnBaseTestCase {
             fail("valid repo type caused error: " + url);
         }
     }
+
+    public void testRepoLabelInput() {
+
+        // Repository label must contain only letters, hyphens, periods, underscores and numeral
+
+        // I N V A L I D
+        invalidRepoLabelInput("");
+        invalidRepoLabelInput("example/repo");
+        invalidRepoLabelInput("example;repo");
+        invalidRepoLabelInput("example_repo$");
+        invalidRepoLabelInput("my*examplerepo");
+        invalidRepoLabelInput("example^repo");
+
+        // V A L I D
+        validRepoLabelInput("my example repo");
+        validRepoLabelInput("my-example-repo");
+        validRepoLabelInput("my-example_repo");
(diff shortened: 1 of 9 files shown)


java/code/src/com/redhat/rhn/manager/channel/test/BaseRepoCommandTest.java
```diff
@@ -130,4 +130,82 @@ private void validUrlInput(String url, String type) {
...
+        invalidRepoLabelInput("my*examplerepo");
+        invalidRepoLabelInput("example^repo");
+
+        // V A L I D
+        validRepoLabelInput("my example repo");
+        validRepoLabelInput("my-example-repo");
+        validRepoLabelInput("my-example_repo");
+        validRepoLabelInput("my_repo_22");
+        validRepoLabelInput("My-Example-Repo");
+        validRepoLabelInput("My/Example/Repo/15");
```
mcalmer, line 152: Why is this valid?

meaksh: I used here the same regex we have for validating the channel names. The slashes will be replaced as non-valid characters in the reposync plugins (as for already created custom repository labels in the database).

On the other hand, we could also not accept slashes from the UI to be safer and prevent posible related issues with path maybe even in some other places from the code base where we use the "repository label" to build paths in the system.

mcalmer: What I mean is: in Line 140 we have this test:
```
    invalidRepoLabelInput("example/repo");
```

Why is "example/repo" INVALID but "My/Example/Repo/15" is VALID ?
Is it not because of the slash?

meaksh: Oh! That's correct.. there should be a problem with the logic of the `invalidRepoLabelInput` and `validRepoLabelInput` helpers. I'll fix that.

With the current implementation, the "slashes" are allowed.

mcalmer (approved):

PR #2967, opened 2020-12-03
Better safety net mechanism

Bischoff:
## What does this PR change?

This PR:
 - explains better how "targetted assassinations" of reposyncs work,
    both in terms of workflow of the test suite and of implementation
 - adds safety net for SLES 15 SP2 and SLES 12 SP5 which we already use, or will use, in our test suite
 - does not add safety net for auto-installation if we don't perform it
 - removes possible duplicates from the list for faster execution

In 4.0 and 4.1 versions, it also adds a bugfix that was not backported (`15SP2` versus `15-SP2`).

## Links

Fixes SUSE/spacewalk#12895

Ports:
* 4.0: SUSE/spacewalk#13348
* 4.1: SUSE/spacewalk#13349

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -215,6 +215,11 @@ When(/^I execute mgr\-sync refresh$/) do
   $command_output = sshcmd('mgr-sync refresh', ignore_err: true)[:stderr]
 end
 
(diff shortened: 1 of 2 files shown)


srbarrios (approved):

lkotek (approved):
Thanks! LGTM.

cbosdo (approved):
:+1:

PR #2975, opened 2020-12-05
QA: Cucumber profiles to filter features by focus areas

srbarrios:
## What does this PR change?

This is a draft/proposal. The idea is to add the support to run the test suite using [Cucumber profiles](https://cucumber.io/docs/cucumber/configuration/) , each profile will filter the features by a [focus area](https://confluence.suse.com/pages/viewpage.action?pageId=580223405), so if we made changes on a concrete area, we can run the cucumber .feature files related only to this focus area.

In this draft, we only add the new tags at the feature level of all our secondary features, and those must be reviewed so we agree if each of them fits in the right feature file. As the idea is to always run the core stage + init_clients stage and then use these Cucumber profiles on the secondary features.

The Rake task will get from the environment the variable PROFILE, so the integration on the susemanager-ci pipelines is trivial.

Related to: 
Epic: https://github.com/SUSE/spacewalk/issues/6426
Card: https://github.com/SUSE/spacewalk/issues/7296
Sub-Card: https://github.com/SUSE/spacewalk/issues/12831

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Minimal documentation provided for now

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


jordimassaguerpla:
I think this is an excellent idea. I can't review the code details given I don't know enough on cubumer and suma.

Bischoff:
Too long to review the details... but definitely something needed.

Just a small request: please change `@long_tests` introduced by Cédric into several of these settings.

srbarrios:
> Too long to review the details... but definitely something needed.
> 
> Just a small request: please change `@long_tests` introduced by Cédric into several of these settings.

This is a draft, so no need to review all the details, but I would like to know your opinion @Bischoff about how we can handle the execution of a sub-set of tests that could be related to a Pull Request.
As you remember, our first agreement was to try to split the different feature files into different yaml files, so we organize the features of a same kind in  yaml files. Here, I propose a different idea, introducing Cucumber profiles to filter the features we use by a combination of tags per profile, into each feature file.

Bischoff:
> > Too long to review the details... but definitely something needed.
> > Just a small request: please change `@long_tests` introduced by Cédric into several of these settings.
> 
> This is a draft, so no need to review all the details, but I would like to know your opinion @Bischoff about how we can handle the execution of a sub-set of tests that could be related to a Pull Request.
> As you remember, our first agreement was to try to split the different feature files into different yaml files, so we organize the features of a same kind in yaml files. Here, I propose a different idea, introducing Cucumber profiles to filter the features we use by a combination of tags per profile, into each feature file.
> If you like this new approach, I will test it first and come back with some outcome, as this is just code written without test it.

I am not religious for one approach or the other.

The yaml files and the tags could have represented orthogonal directions. But two different kind of tags might also work just fine.

Maybe we should just make very clear in which category is every tag:
* the "variable geometry", depending on which resources we have, example, `@sshminion`
* the "functionalities", like the tags introduceds in this PR.

Maybe prefixes like `@res_XX` and `@focus_XX` could help keep the distinction under control?


> Note: `@long_tests` is a different kind of tag, with its concrete implementation, we can review it in a different Pull Request.
> The difference is that Cucumber profile will filter all the features and only run those with the tags defined at the feature level, but the long_tests tag is a tag that defines if some features will run if an environment variable is set up on the controller.

That's my point, I'd like to keep the whole thing simple, and have one type of tags less. Said otherwise, `@long_tests` = `@virtualization` + `build_image`. We could split in into two "focus" tags with benefit.

srbarrios:
> Maybe prefixes like `@res_XX` and `@focus_XX` could help keep the distinction under control?

Let's make a try without anything prepended if we find it brings confusion between tags, we re-visit it.
I would like to avoid longer tags.

> > Note: `@long_tests` is a different kind of tag, with its concrete implementation, we can review it in a different Pull Request.
> > The difference is that Cucumber profile will filter all the features and only run those with the tags defined at the feature level, but the long_tests tag is a tag that defines if some features will run if an environment variable is set up on the controller.
> 
> That's my point, I'd like to keep the whole thing simple, and have one type of tags less. Said otherwise, `@long_tests` = `@virtualization` + `build_image`. We could split in into two "focus" tags with benefit.

The behavior of how we use the Cucumber profiles and how we use the long_tests tag is different. A Cucumber profile can have one or multiple tags, but what it does is run ONLY the features with this tag at feature level. Instead, the long_tests tag is used for is to run the whole CI tests PLUS the features tagged with long_tests tag.

PR #2978, opened 2020-12-07
PoC: Migrate CommonJS components to ES6

Etheryte:
## What does this PR change?

While working on https://github.com/SUSE/spacewalk/issues/13208, I ran into the problem that we currently use a mix of CommonJS (`require`/`module.exports`) and ES6 (`import`/`export`) across different React components.  
Typescript has trouble jumping between the two correctly, even with interop and other shenanigans enabled.  

As a solution, I propose migrating the CommonJS based components to ES6 so we have one uniform way of doing things, excluding async require usage which has no equivalent in ES6.  
This is one possible way to resolve the situation, but I'm also very open to alternative suggestions.  
This is mergeable as-is, but if the PR is too big I'm open to other ideas.

Currently I've migrated imports using [jscodeshift](https://github.com/facebook/jscodeshift), and exports with a mix of that and manual work depending on the situation.  
I've also gone through some of the common star imports manually to ensure they're consistently used.  

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Not sure if this needs additional documentation, no user-facing changes?

- [x] **DONE**

## Test coverage
- No tests: Needs to pass against all existing test suites.

- [x] **DONE**

## Links

Related: https://github.com/SUSE/spacewalk/issues/13208

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


cbbayburt (approved):
LGTM :+1: Nice cleanup. Thanks!

hustodemon (approved):
Thanks for this effort @Etheryte ! I was always confused by the mix of 2 different ways to write import/exports and this helps a lot.

web/html/src/manager/admin/config/use-monitoring-api.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import {useState} from 'react';
-import * as Network from 'utils/network';
+import Network from 'utils/network';
 
 import type {JsonResult} from "../../../utils/network";
```
ncounter, line 5: Now that we are on it, I wonder if with some good `regex` it is possible to:
- find and replace all the `"....."` to `'.....'`
- keep a space between brackets from `{abc}` to `{ abc }`

while using imports. I see we are standardizing it so it would be nice, but not really required.

ncounter (approved):
I left one small comment, feel free to give it a try, or ignore it for now.

Thank you @Etheryte for this cleanup. We need consistency and standardization in our codebase so, very glad to see this. :+1:

parlt91 (approved):
LGTM. Thanks for cleaning this up ;)

web/html/src/manager/admin/config/use-monitoring-api.js
cbbayburt: This disturbs me as well :rofl:

On the other hand, I prefer single quotes and `{abc}` without spaces, but I guess we can leave it to @Etheryte to decide

Etheryte: This caught my eye as well, but I didn't want to throw everything along with the kitchen sink into one PR. I'll create a follow-up card and do it there if that works for you?

ncounter: If I could vote, I would:

- 2 space only, I think we already have it in place as a *standardization*, probably we don't have it enforced nor a checkstyle rule to blame who breaks it.
- single quotes
- no real preference for space or no-space on `{abc}`

Time for a strong check style test to fail if someone breaks them, whatever the rules will be? We do in Java, why not here?

Etheryte: I created https://github.com/SUSE/spacewalk/issues/13437 to track this with a short summary of the above.

PR #2980, opened 2020-12-07
spacewalk-java: SPEC file updates

sbluhm:
## What does this PR change?

- SPEC file libxml2-devel addition
- Source0 update
- removed RHEL commons-jexl prettification.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: no functional change.

- [X] **DONE**

## Test coverage
- No tests: tested during auto-build.
Manually tested build successfully on CentOS8, LEAP15.2

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/spacewalk-java.spec
```diff
@@ -748,7 +752,7 @@ chown tomcat:%{apache_group} /var/log/rhn/gatherer.log
 
 %{jardir}/snakeyaml.jar
 # SUSE extra runtime dependencies: spark, jade4j, salt API client + dependencies
-%{jardir}/apache-commons-jexl_commons-jexl.jar
+%{jardir}/commons-jexl.jar
```
sbluhm, line 755: Can you please check on this change? It is confusing me a bit.
1. This built successfully on my LEAP15.2 test.
2. [https://build.opensuse.org/package/view_file/systemsmanagement:Uyuni:Master:Other/apache-commons-jexl/apache-commons-jexl.spec](https://build.opensuse.org/package/view_file/systemsmanagement:Uyuni:Master:Other/apache-commons-jexl/apache-commons-jexl.spec) also puts the file into `%{_javadir}/*`
So I am wondering why the original location `%{jardir}/apache-commons-jexl_commons-jexl.jar `worked in the first place...

mcalmer: I found out that we have also an official build of this package in Leap. So we do not need to build it on our own.
I have disabled apache-commons-jexl building for SUSE and removed the last build from the repo.

The Leap package has this path similar to the official CentOS package. AFAIK you added the condition which now also match the Leap Package.

I forgot to disable the CentOS package. I am right, that there is package names apache-commons-jexl in CentOS 8 and it also have the path apache-commons/jexl_commons-jexl.jar ? If yes, I would just use `%{jardir}/apache-commons-jexl_commons-jexl.jar` everywhere and we can remove apache-commons-jexl from systemsmanagement:Uyuni:Master:Other

sbluhm: There isn't. Let me have another look at this. My current strategy:
1. Adapt/verify the SPEC to the official SUSE bundle version.
2. remove the bundle from systemsmanagement:Uyuni:Master:Other (SUSE and CentOS)
3. Add the bundle to my "EL8 only" wip branch (home:sbluhm:systemsmanagement:Uyuni:Other:EL8) and make it match RHEL conventions.
4. Adapt the SPEC file to RHEL if required.

sbluhm: OK, tested and verified now on LEAP and CentOS. Your parallel changes 3 days ago to the spec file is what really confused me as I couldn't figure out the non-existing delta :)

Note that we should eventually take out [https://github.com/uyuni-project/uyuni/blob/b9e6a913b778e770f2243625026da90cc2ee9b8a/java/buildconf/build-props.xml#L79](https://github.com/uyuni-project/uyuni/blob/b9e6a913b778e770f2243625026da90cc2ee9b8a/java/buildconf/build-props.xml#L79) to make everything consistent. 

This would conflict with everyone having installed apache-commons-jexl from the Uyuni repository though. I noticed that the Uyuni package always installs before the OS package on LEAP.
An alternative option would be to release a new version/revision with the OS path and force this change on the next package update.

mcalmer (approved):

java/spacewalk-java.spec
mcalmer: It was the idea to remove the Uyuni build version. But when it is required for CentOS we should think about it.
I think I will revisit this on the weekend and check your suggestion with our internal repos for SLES.

sbluhm: It is required for CentOS. But the naming conventions should match up with the SUSE ones. So the real issue is **current** package in Uyuni:Others. I am rebuilding it to RHEL/SUSE standards in my repo atm.

sbluhm: @mcalmer , I just thought about this. We should just require %{jardir}/commons-jexl.jar. That would/should force the installation of the OS package and ignore the Uyuni package version. I won't have this on my todo/test list though as I am currently playing in different areas (unless you ask me to).

PR #2981, opened 2020-12-07
Faster and more robust checkbox test

Bischoff:
## What does this PR change?

Test for a selected product can be made faster (do not search in the page twice) and more robuts (current test fails in certain conditions).

## Links

Ports:
* 4.1: SUSE/spacewalk#13381
* 4.0: SUSE/spacewalk#13383

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -511,8 +511,9 @@ end
 
 Then(/^I should see the "(.*?)" selected$/) do |product|
   xpath = "//span[contains(text(), '#{product}')]/ancestor::div[contains(@class, 'product-details-wrapper')]"
-  product_identifier = find(:xpath, xpath)['data-identifier']
-  raise "#{product_identifier} is not checked" unless has_checked_field?('checkbox-for-' + product_identifier)
+  within(:xpath, xpath) do
+    raise "#{find(:xpath, '.')['data-identifier']} is not checked" unless find(:xpath, "./div/input[@type='checkbox']").checked?
+  end
 end
 
 And(/^I wait until I see "(.*?)" product has been added$/) do |product|


ktsamis (approved):
LGTM

lkotek (approved):
Great, thanks :-)

calancha (approved):

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -511,8 +511,9 @@
 
 Then(/^I should see the "(.*?)" selected$/) do |product|
   xpath = "//span[contains(text(), '#{product}')]/ancestor::div[contains(@class, 'product-details-wrapper')]"
-  product_identifier = find(:xpath, xpath)['data-identifier']
-  raise "#{product_identifier} is not checked" unless has_checked_field?('checkbox-for-' + product_identifier)
+  product = find(:xpath, xpath)
+  selected = product.find(:xpath, "./div/input[@type='checkbox']").checked?
+  raise "#{product['data-identifier']} is not checked" unless selected
```
srbarrios, line 516:
```suggestion
within(xpath) do
  raise "#{product['data-identifier']} is not checked" unless product.find(:xpath, "./div/input[@type='checkbox']").checked?
end
```

Bischoff: why not, it's more elegant as you suggest

BUT `product` is a node, what you want in the error message is `product['data-identifier']`, a string

Bischoff: Done.

srbarrios: @Bischoff it is impossible that product is undefined, we are passing a value as a parameter in the step definition.
With your extra change, we are adding an extra find, anyhow, as it is only called on a raise, fine with this approach.

PR #2983, opened 2020-12-07
WIP: Github action for Javascript unit tests

Etheryte:
## What does this PR change?

As I learned in the review meeting, frontend unit tests are currently not run against PRs.  
This is an initial implementation for the action.  

I could use some help with integrating `actions/cache@v2` for dependencies similar to `java-checkstyle`, I'm not that familiar with the tool.   
Perhaps @renner could give a helping hand?  

Also, I'm not sure how to correctly integrate with the test rerunner bot.

Other than that, I tested it on my own fork and it seems to work as intended.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Workflow setup, no user facing changes.

- [x] **DONE**

## Test coverage
- Unit tests were added to the pipeline.

- [x] **DONE**

## Links

Fixes: https://github.com/SUSE/spacewalk/issues/13382

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/javascript-unit-tests.yml
+++ b/.github/workflows/javascript-unit-tests.yml
@@ -14,16 +14,36 @@ jobs:
   javascript_unit_tests:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        node-version: [10.x]
+
     steps:
     - uses: actions/checkout@v2
 
-    - name: Use Node.js
+    - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v1
       with:
-        node-version: '10.x'
+        node-version: ${{ matrix.node-version }}
 
+    # See: https://github.com/actions/cache/blob/main/examples.md#node---yarn
+    - name: Get Yarn cache directory
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn cache dir)"
+
+    # FYI, caching node_modules directly instead of Yarn's cache offers no considerable speed benefit and increases the cache size
+    - name: Use Yarn cache
+      uses: actions/cache@v2
+      id: yarn-cache
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ matrix.node-version }}-${{ hashFiles('**/yarn.lock') }}
+
+    # This step is required even on a cache hit or the cache won't hydrate node_modules.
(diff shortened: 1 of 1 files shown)


renner (approved):
This looks good, it will run and re-run automatically on the main PR events (especially push of commits or push of a rebase) in case the patch is changing `web/html/src/**`, as well as when a patch is merged to `master` and this path is affected. No integration is needed with the test re-runner bot apart from removing a corresponding job from there if it exists, which is basically the advantage of using this kind of integrated CI. There is a button on the workflow results page to trigger another run if you manually want to do that.

The disadvantage is: it won't (yet) work on our downstream repo and branches until we have a self-hosted runner up, which is already on our (@jordimassaguerpla's) agenda.

The caching is optional of course, but should improve the workflow runtime a lot. Idea is to download dependencies only if the `yarn.lock` file has actually changed, otherwise jobs can re-use the cached dependencies. There is yarn examples available from [here](https://github.com/actions/cache/blob/main/examples.md#node---yarn) or see my [Java example](https://github.com/uyuni-project/uyuni/blob/master/.github/workflows/java-checkstyle.yml). Let me know if you have specific questions, I found it most difficult to understand the  `restore-keys`, and eventually omitted them for the Java case. You can test the workflow including the caching by adding test commits to this PR that can be removed before the final merge.

BTW: `javascript_lint` is another candidate of a job that we might want to migrate from Jenkins to a workflow. As long as we run it in Jenkins a community contributor will not be able to actually see the results in case of failure!

Etheryte:
I added caching using the recommended Yarn cache approach with slight modifications. I also tested out a raw `node_modules` based approach, but the performance was identical while having a larger cache size. I added a validation condition for Node version as well, which should avoid any problems if we move from 10.x to 12.x some years down the road.

I couldn't get `restore-keys` to work reliably either, even after reading a number of related bug reports, and eventually opted to not use it.

cbbayburt (commented):
> BTW: javascript_lint is another candidate of a job that we might want to migrate from Jenkins to a workflow. As long as we run it in Jenkins a community contributor will not be able to actually see the results in case of failure!

IIUC it's a very similar work so maybe we can get it done as well?

Etheryte:
As discussed on chat, linting will be handled separately since there's some additional friction there with Flow related package problems. Tracked in https://github.com/SUSE/spacewalk/issues/13403.

PR #2985, opened 2020-12-08
Simplify login page inputs

Etheryte:
## What does this PR change?

Small things that caught my eye when working on https://github.com/uyuni-project/uyuni/pull/2927.
The effect can be replaced with [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus). Using `tabindex` with a greater value than 0 can [break accessibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) so I would suggest we remove it.  
Thoughts?  

## GUI diff

No difference. Tab indexing has changed, but this would be more widely user friendly.

- [x] **DONE**

## Documentation
- No documentation needed: No visual difference.

- [x] **DONE**

## Test coverage
- No tests: No visual difference.

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/login/login.js
+++ b/web/html/src/manager/login/login.js
@@ -79,11 +79,6 @@ const Login = (props: Props) => {
   const loginInput = useInputValue('');
   const passwordInput = useInputValue('');
   const { onLogin, success, messages } = useLoginApi();
-  const loginInputRef = useRef();
-
-  useEffect(() => {
-    loginInputRef && loginInputRef.current && loginInputRef.current.focus();
-  }, []);
 
   const product = props.isUyuni ? products.uyuni : products.suma;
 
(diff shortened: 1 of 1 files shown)


mbologna:
> Small things that caught my eye when working on #2927.
> The effect can be replaced with [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus). Using `tabindex` with a greater value than 0 can [break accessibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) so I would suggest we remove it.
> Thoughts?

Looks good. Please keep an eye on the testsuite (and you are also the helper) after you merge this.

mbologna (approved):

PR #2987, opened 2020-12-08
Add API call to get all formula data

witekest:
## What does this PR change?

Add API method

`listEndpoints(User, List<Long>)`

to list all available application and Prometheus exporters endpoints for the systems whose IDs
match with the passed systems IDs.

Requires SUSE/salt-formulas#102

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: change only to XMLRPC API with automatically generated reference documentation.

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes #
Tracks https://github.com/SUSE/spacewalk/issues/12777

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/dto/EndpointInfo.java
+++ b/java/code/src/com/redhat/rhn/domain/dto/EndpointInfo.java
@@ -1,0 +1,99 @@
+/**
+ * Copyright (c) 2021 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package com.redhat.rhn.domain.dto;
+
+import java.util.Optional;
+
+/**
+ * Class for representing the endpoint information of applications or Prometheus exporters of a minion system
+ */
+public class EndpointInfo {
+
+    private final Long systemID;
+    private final String endpointName;
+    private final Integer port;
+    private final String path;
+    private final String module;
+    private final String exporterName;
+    private final Boolean tlsEnabled;
+
+    /**
+     * Instantiates a new endpoint information
+     * @param systemIDIn server ID
+     * @param endpointNameIn endpoint name
+     * @param exporterNameIn
+     * @param portIn
+     * @param moduleIn
+     * @param pathIn
+     * @param tlsEnabledIn
+     */
+    public EndpointInfo(Long systemIDIn,
+                        String endpointNameIn,
+                        String exporterNameIn,
+                        Integer portIn,
+                        String moduleIn,
+                        String pathIn,
+                        Boolean tlsEnabledIn) {
+        this.systemID = systemIDIn;
+        this.endpointName = endpointNameIn;
+        this.exporterName = exporterNameIn;
+        this.port = portIn;
+        this.module = moduleIn;
+        this.path = pathIn;
+        this.tlsEnabled = tlsEnabledIn;
+    }
+
+    public Long getSystemID() {
+        return systemID;
+    }
+
+    public String getEndpointName() {
+        return endpointName;
+    }
+
+    public Integer getPort() {
+        return port;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    public String getModule() {
+        return module;
+    }
+
+    public Optional<String> getExporterName() {
+        return Optional.ofNullable(exporterName);
+    }
+
+    public Boolean isTlsEnabled() {
+        return tlsEnabled;
+    }
+
+    @Override
+    public String toString() {
+        return "EndpointInfo{" +
+                "systemID=" + systemID +
+                ", endpointName='" + endpointName + '\'' +
+                ", port=" + port +
+                ", path='" + path + '\'' +
+                ", module='" + module + '\'' +
+                ", exporterName='" + exporterName + '\'' +
+                ", tlsEnabled='" + tlsEnabled + '\'' +
+                '}';
+    }
+}
--- a/java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
+++ b/java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
@@ -17,6 +17,8 @@ package com.redhat.rhn.domain.formula;
 import com.redhat.rhn.GlobalInstanceHolder;
 import com.redhat.rhn.common.validator.ValidatorError;
 import com.redhat.rhn.common.validator.ValidatorException;
+import com.redhat.rhn.domain.dto.EndpointInfo;
+import com.redhat.rhn.domain.dto.FormulaData;
 import com.redhat.rhn.domain.org.Org;
 import com.redhat.rhn.domain.server.MinionServer;
 import com.redhat.rhn.domain.server.MinionServerFactory;
@@ -69,7 +71,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
-
 
 /**
  * Factory class for working with formulas.
@@ -943,9 +944,15 @@ public class FormulaFactory {
     @SuppressWarnings("unchecked")
     private static boolean hasMonitoringDataEnabled(Map<String, Object> formData) {
         Map<String, Object> exporters = (Map<String, Object>) formData.get("exporters");
-        return exporters.values().stream()
-                .map(exporter -> (Map<String, Object>) exporter)
-                .anyMatch(exporter -> (boolean) exporter.get("enabled"));
+        if (exporters != null && !exporters.isEmpty()) {
+            return exporters.values().stream()
+                    .filter(Map.class::isInstance)
+                    .map(exporter -> (Map<String, Object>) exporter)
+                    .anyMatch(exporter -> (boolean) exporter.get("enabled"));
+        }
+        else {
+            return false;
+        }
     }
 
     /**
@@ -1108,4 +1115,58 @@ public class FormulaFactory {
         return dataDir + PILLAR_DIR;
     }
 
+    /**
+     * Find endpoint information from given formula data
+     * @param formulaName name of the formula to examine
+     * @param formulaData formula data to extract information from
+     * @return list of endpoint information objects
+     */
+    public static List<EndpointInfo> getEndpointsFromFormulaData(String formulaName, FormulaData formulaData) {
+        return getExportersEndpoints(formulaData);
+    }
+
+    private static List<EndpointInfo> getExportersEndpoints(FormulaData formulaData) {
+        Map<String, Object> formulaValues = formulaData.getFormulaValues();
+        if (formulaValues.containsKey("exporters")) {
+            Boolean proxyEnabled = getValueByPath(formulaValues, "proxy_enabled")
+                    .filter(Boolean.class::isInstance)
+                    .map(Boolean.class::cast)
+                    .orElse(false);
+            Optional<Integer> proxyPort = proxyEnabled ? getValueByPath(formulaValues, "proxy_port")
+                    .filter(Number.class::isInstance)
+                    .map(Number.class::cast)
+                    .map(Number::intValue) : Optional.empty();
+            String proxyPath = proxyEnabled ? "/proxy" : null;
+            Boolean tlsEnabled = getValueByPath(formulaValues, "tls:enabled")
+                    .filter(Boolean.class::isInstance)
+                    .map(Boolean.class::cast)
+                    .orElse(false);
+
+            Map<String, Object> exportersMap = getValueByPath(formulaValues, "exporters")
+                    .filter(Map.class::isInstance)
+                    .map(Map.class::cast)
+                    .orElseGet(Collections::emptyMap);
+
+            return exportersMap.entrySet().stream()
+                    .map(exporterEntry -> new ExporterConfig(exporterEntry.getKey(),
+                            Optional.ofNullable(exporterEntry.getValue())
+                                    .filter(Map.class::isInstance)
+                                    .map(Map.class::cast)
+                                    .orElseGet(Collections::emptyMap)))
+                    .filter(ExporterConfig::isEnabled)
+                    .filter(e -> (proxyEnabled && proxyPort.isPresent()) || (!proxyEnabled && e.getPort().isPresent()))
+                    .map(exporterConfig -> new EndpointInfo(
+                            formulaData.getSystemID(),
+                            exporterConfig.getEndpointNameOrFallback(),
+                            exporterConfig.getEndpointNameOrFallback(),
+                            proxyEnabled ? proxyPort.get() : exporterConfig.getPort().get(),
+                            proxyEnabled ? exporterConfig.getProxyModuleOrFallback() : null,
+                            proxyPath,
+                            tlsEnabled))
+                    .collect(Collectors.toList());
+        }
+        else {
+            return new ArrayList<>();
+        }
+    }
 }
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/monitoring/SystemMonitoringHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/monitoring/SystemMonitoringHandler.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2021 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package com.redhat.rhn.frontend.xmlrpc.system.monitoring;
+
+import com.redhat.rhn.domain.dto.EndpointInfo;
+import com.redhat.rhn.domain.user.User;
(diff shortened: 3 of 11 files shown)


java/code/src/com/redhat/rhn/manager/formula/FormulaManager.java
```diff
@@ -348,6 +348,15 @@ public void enableFormula(String minionId, String formulaName) throws IOExceptio
                 .collect(Collectors.toList());
     }
 
+    public List<Map<String, FormulaData>> getCombinedFormulaDataForSystems(User user, List<Integer> systemIDs) {
```
cavalheiro, line 351: Would it make sense to pass a flag to this method to optionally get only the monitoring formulas?

java/code/src/com/redhat/rhn/manager/formula/FormulaManager.java
```diff
@@ -348,6 +348,15 @@ public void enableFormula(String minionId, String formulaName) throws IOExceptio
...
+        return systemIDs.stream()
+                .map(Integer::longValue)
+                .map(sID -> FormulaFactory.getCombinedFormulasByServerId(sID).stream()
+                .collect(Collectors.toMap(formula -> formula,
+                        formula -> getCombinedFormulaDataForSystems(user, List.of(sID), formula).get(0))))
```
cavalheiro, line 356: Is this call expected to perform well on large setups?

java/code/src/com/redhat/rhn/manager/formula/FormulaManager.java
witekest: I also don't like the fact that we dump a lot of not needed and possibly sensitive data. In the second commit I've proposed an alternative new API method to get only relevant information about endpoints.

java/code/src/com/redhat/rhn/manager/formula/FormulaManager.java
witekest: I haven't measured performance but taking into account that all the information is available locally on the disk it should not be a problem.

moio:
@witekest I see this has `WIP` in the title - do you want me to take a look at this time?

witekest:
That would be nice. The code is not production ready yet but functional.
It requires SUSE/salt-formulas#102
It examines all applied formulas for configured exporters and exposes relevant endpoints information. The code can be easily extended to provide information not only about exporters but all managed applications as requested in upstream review.

renner (changes requested):
This patch will need changes before it can be merged, I left a number of really minor comments that should become obsolete when checkstyle is fixed on the patch. See the checkstyle results for more details.

My main feedback is the following, and I am curious about @moio's opinion here: Is there really a usecase for an API call to return all application endpoints of a managed system that were configured via Formulas? From my perspective for the given usecase of service discovery, aren't we rather looking for an API call to list all (Prometheus specific) monitoring endpoints of a system, while we actually don't care if those were configured via Formulas or not?

So my suggestion would be: let's rename the new API call to `listMonitoringEndpoints()` and move it to a namespace that is not specific to Formulas, e.g. the `SystemHandler`. We could also create a `SystemMonitoringHandler` and keep the current name of the call, so it can be used as `system.monitoring.listEndpoints()`, but only returning Prometheus exporter endpoints! Most of the code in this patch could be used as it is, but the usecase would be much more specific and clear to me. We would also not return additional data about application endpoints only to filter it out afterwards, while I'm currently not even sure which applications would be included there, and which ones maybe not? Also how would you suggest to maintain this in the future when Formulas with application endpoint configuration are added, removed or changed?

I understand it like this: Uyuni is provisioning and configuring Prometheus monitoring exporters on managed systems. How this is done is an implementation detail (currently via Formulas, but that could change in theory to e.g. just a Salt state, or an Ansible playbook, whatever), but the Uyuni server centrally knows about the monitoring endpoints incl. reverse proxies as it is rolling out all of this configuration. With this patch we want to centrally provide this information to feed the Prometheus SD. If this is not the way how SD is supposed to be used, then we might alternatively want to investigate pushing the configuration to the Prometheus server, that is also a manged system that the Uyuni server is in control of. IIRC this has been the original idea anyways, and might make even more sense from a design perspective as we could automatically push the configuration on demand, e.g. right after it was changed.

witekest:
> My main feedback is the following, and I am curious about @moio's opinion here: Is there really a usecase for an API call to return all application endpoints of a managed system that were configured via Formulas? From my perspective for the given usecase of service discovery, aren't we rather looking for an API call to list all (Prometheus specific) monitoring endpoints of a system, while we actually don't care if those were configured via Formulas or not?

Feedback from Prometheus cores was, we're looking for all installed systems/applications/endpoints/formulas. That's what service discovery does. They not necessarily have to be Prometheus or monitoring specific. We indeed don't care if those were configured by formulas or not.

> So my suggestion would be: let's rename the new API call to `listMonitoringEndpoints()` and move it to a namespace that is not specific to Formulas, e.g. the `SystemHandler`. We could also create a `SystemMonitoringHandler` and keep the current name of the call, so it can be used as `system.monitoring.listEndpoints()`, but only returning Prometheus exporter endpoints! Most of the code in this patch could be used as it is, but the usecase would be much more specific and clear to me.

I'm fine with moving this call to `system` namespace. As I explained above it should not be monitoring specific.
Prometheus cores were not happy about consuming only exporters metrics. They evangelize for instrumenting all applications directly so that no additional exporters are needed.

> We would also not return additional data about application endpoints only to filter it out afterwards, while I'm currently not even sure which applications would be included there, and which ones maybe not? Also how would you suggest to maintain this in the future when Formulas with application endpoint configuration are added, removed or changed?

That's exactly opposite to what Prometheus cores have suggested: Drop not needed targets during relabeling in Prometheus configuration.
The list of applications exposing metrics directly is available [here](https://prometheus.io/docs/instrumenting/exporters/#software-exposing-prometheus-metrics). Currently that would be _Grafana_ only. But Uyuni allows adding custom formulas.
I think for the first implementation it would be enough to just list the installed formulas and return their names. Alternatively we could add some metadata to the formulas or filter on some field in FormulaData.

I could limit the new API call to returning just exporter endpoints, but then would have to use `formulas.getCombinedFormulasByServerId` for every server or, to make it more efficient, introduce another API call `formulas.getCombinedFormulas`.

> I understand it like this: Uyuni is provisioning and configuring Prometheus monitoring exporters on managed systems. How this is done is an implementation detail (currently via Formulas, but that could change in theory to e.g. just a Salt state, or an Ansible playbook, whatever), but the Uyuni server centrally knows about the monitoring endpoints incl. reverse proxies as it is rolling out all of this configuration. With this patch we want to centrally provide this information to feed the Prometheus SD. If this is not the way how SD is supposed to be used, then we might alternatively want to investigate pushing the configuration to the Prometheus server, that is also a manged system that the Uyuni server is in control of. IIRC this has been the original idea anyways, and might make even more sense from a design perspective as we could automatically push the configuration on demand, e.g. right after it was changed.

Yes, using [file-based service discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config) is another way to go. The problem which we have here is that configuration file has to be updated on Prometheus server every time some minion's configuration changes. From what I understand we cannot do it with the current Salt integration which we have and would need some extra logic.

moio:
> My main feedback is the following, and I am curious about @moio's opinion here: Is there really a usecase for an API call to return all application endpoints of a managed system that were configured via Formulas? From my perspective for the given usecase of service discovery, aren't we rather looking for an API call to list all (Prometheus specific) monitoring endpoints of a system, while we actually don't care if those were configured via Formulas or not?

I agree with that, as I mentioned earlier via email:

"From a design perspective, I think the Uyuni SD shouldn't really be aware of the existence of formulas [... I] would think about an endpoint (or a set of endpoints) that return any relevant monitoring configuration data - no matter where it comes from, and how Uyuni came to know about it."

moio:
> Feedback from Prometheus cores was, we're looking for all installed systems/applications/endpoints/formulas. That's what service discovery does. They not necessarily have to be Prometheus or monitoring specific. 

I re-read all comments in https://github.com/prometheus/prometheus/pull/8190 and I think there's a bit of a mix of three different aspects here. Let me try to clarify.

1. design of an SD is that the SD returns all possible scraping targets "in one shot" - any filtering or relabeling should happen later in Prometheus itself. Reason: they do not want filtering/relabeling logic duplicated in each SD. That is good engineering and definitely something I agree with.
2. design of Prometheus is that application instrumenting is the gold standard, and exporters are a second-class citizens when that is not possible. So Prometheus has exporters as a "fallback mechanism" when application instrumentation is not possible only! Or at least that was the original intent. I understand application instrumentation is what SoundCloud created Prometheus for, and sympathize with the design ideal as well - but as a pragmatic engineer I also observe that today exporters have also grown a lot and saw massive adoption, and they are very well used within the community. De facto, Prometheus is today a very valid system monitoring solution as well, and that's primarily the way we use it in the Uyuni context too. So I do not see our focus to system monitoring as a bad thing - the Prometheus project has simply outgrown its original scope, a very common (and positive!) phenomenon in open source and in general. I think we should respect the ideal with a bit of pragmatism, and just embrace the fact that Uyuni (and its community) might not be ready for pervasive application instrumentation yet, and we help where it helps the most
3. the APIs consumed by the SD should not be Prometheus (or monitoring) specific. This is something I could not find in the PR comments above, and also something I am convinced about.
 
The reason is that, from an Uyuni perspective, the APIs we are adding are de facto Prometheus and monitoring specific - they contain all the data that is required for the Prometheus SD and only that. And that is fine by me - monitoring and Prometheus integration specifically is what those XMLRPC methods were made for.

So from a naming perspective, I think that is not a bad thing to signal why those APIs were created. I think that, if we are explicit and specific in naming and documentation about our intents here, future unaware programmers will be happy to find hints about what we are trying to achieve and why. So I would go for a `monitoring` namespace or even `prometheus`. With an extreme example: the day we will migrate to a completely new monitoring platform it will be very clear what code needs to be dropped!

Do you see other pros/cons that I am currently not considering?

BTW, and sorry if this only comes as the last line: thanks for all the continued effort. Do not get discouraged!!!

witekest:
Alright, I'll go then with `system.monitoring.listEndpoints()` and keep the current implementation for now. Thanks for your reviews.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/monitoring/SystemMonitoringHandler.java
```diff
@@ -0,0 +1,62 @@
...
+     * @return a list containing endpoint details for all Prometheus exporters on the passed system IDs.
+     *
+     * @xmlrpc.param #session_key()
+     * @xmlrpc.param #array_single("int", "systemID")
+     * @xmlrpc.returntype
+     *   #array_begin()
+     *     $EndpointInfo
+     *   #array_end()
+     */
+    public List<EndpointInfo> listEndpoints(User loggedInUser, List<Long> systemIDs) {
```
moio, line 59: Did you try this endpoint (eg. from a Python script)?

I was under the impression that in order to return a custom type one really had to add a custom `Serializer` class (see for example `com.redhat.rhn.frontend.xmlrpc.serializer.SystemOverviewSerializer`).

If that is wrong then I'll be happy to learn that plain beans are also completely OK to return!

java/code/src/com/redhat/rhn/domain/dto/EndpointInfo.java
```diff
@@ -0,0 +1,92 @@
...
+    public String getModule() {
+        return module;
+    }
+
+    public void setModule(String moduleIn) {
+        this.module = moduleIn;
+    }
+
+    public String getExporterName() {
+        return exporterName;
```
moio, line 74: What if an endpoint is an application endpoint, ie. it does not originate from an exporter? Would the name be `null` in that case?

java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
```diff
@@ -1107,4 +1109,75 @@ public static String getClusterPillarDir() {
...
+        List<EndpointInfo> endpointInfos = new ArrayList<>();
+        endpointInfos.addAll(getExportersEndpoints(formulaData));
+        endpointInfos.addAll(getApplicationEndpoints(formulaName, formulaData));
+        return endpointInfos;
+    }
+
+    private static List<EndpointInfo> getApplicationEndpoints(String formulaName, FormulaData formulaData) {
+        // TODO
+        return Collections.emptyList();
+    }
```
moio, line 1128: I would remove this method as of today, and add it only when we will actually support application endpoints.

java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
```diff
@@ -1107,4 +1109,75 @@ public static String getClusterPillarDir() {
...
+                    endpointInfo.setExporterName(endpointInfo.getEndpointName());
+                    endpointInfo.setPort(proxyPort.orElse(exporterConfig.getPort()));
+                    if (proxyEnabled.isPresent() && proxyEnabled.get()) {
+                        endpointInfo.setModule(Optional.ofNullable(
+                                exporterConfigMap.getOrDefault("proxy_module", endpointInfo.getEndpointName()))
+                                .filter(String.class::isInstance).map(String.class::cast).get());
+                        endpointInfo.setPath("/proxy");
+                    }
+
+                    endpointInfos.add(endpointInfo);
```
moio, line 1176: It would be cool if this algorithm could be refactored with streams, so that `EndpointInfo` could be turned into a read-only-after-construction class.

In general, we prefer creating collections via streams rather than modifying methods such as `put`, `add` and so on. Would that be possible here?

moio (commented):
What is the plan with the previous endpoint (`listSystemGroupsForSystemsWithEntitlement`)? Should this PR remove it as well?

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/monitoring/SystemMonitoringHandler.java
witekest: Good point. I've added custom serializer.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/monitoring/SystemMonitoringHandler.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.redhat.rhn.domain.dto.EndpointInfo;
+import com.redhat.rhn.domain.user.User;
+import com.redhat.rhn.frontend.xmlrpc.BaseHandler;
+import com.redhat.rhn.manager.formula.FormulaManager;
+
+import java.util.List;
+
+/**
+ * ServerMonitoringHandler
```
renner, line 25:
```suggestion
 * SystemMonitoringHandler
```
This comment does not match the actual class name.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -7787,5 +7787,4 @@ public TaskomaticApi getTaskomaticApi() {
     private static List<String> maybeActivationKeys(String activationKey) {
         return StringUtils.isEmpty(activationKey) ? Collections.emptyList() : List.of(activationKey);
     }
-
```
renner, line 7790: This change seems unrelated to the patch and could probably be omitted (= 1 file less that is changed).

java/code/src/com/redhat/rhn/manager/formula/FormulaManager.java
```diff
@@ -381,6 +408,32 @@ private FormulaData getCombinedFormulaDataForSystem(MinionIds minionID,
...
+                .flatMap(name -> FormulaFactory.getGroupFormulaValuesByNameAndGroupId(name,
+                        cluster.getGroup().getId()));
+    }
+
+    /**
+     * Gets all endpoints details for given systems
+     * @param systemIDs list of system IDs
+     * @return list of endpoint information objects
+     */
+    public List<EndpointInfo> listEndpoints(List<Long> systemIDs) {
```
renner, line 420: At least for this method it would be great to implement some unit tests.

java/code/src/com/redhat/rhn/domain/dto/EndpointInfo.java
```diff
@@ -0,0 +1,96 @@
+/**
+ *
```
renner, line 2:
```suggestion
 * Copyright (c) 2021 SUSE LLC
 *
```
The copyright header is missing for all new classes in this patch, wouldn't checkstyle usually remind us about that?

java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
```diff
@@ -1108,4 +1109,54 @@ public static String getClusterPillarDir() {
...
+                    .orElse(false);
+
+            Map<String, Object> exportersMap = getValueByPath(formulaValues, "exporters")
+                    .filter(Map.class::isInstance)
+                    .map(Map.class::cast).get();
+
+            return exportersMap.entrySet().stream()
+                    .map(exporterEntry -> new ExporterConfig(exporterEntry.getKey(),
+                            Optional.ofNullable(exporterEntry.getValue())
+                                    .filter(Map.class::isInstance).map(Map.class::cast).get()))
```
renner, line 1146: Not sure how big is the risk here, but could calling `get()` result in a NPE here (or on the previous assignment)?

java/code/src/com/redhat/rhn/domain/dto/EndpointInfo.java
renner: I understand that we would only return monitoring endpoints in the meantime.

renner (approved):
Good work! I like the general direction, but left some comments to be considered before merging. Also a changelog entry seems to be missing so far and obviously some unit test would be great, at least for those parts of the code that contain the actual logic.

moio (commented):
Hey @witekest, thanks for this contribution!

I added some nitpicks here and there, and other than that the only biggish request from my end would be to add a jUnit test (as @renner has also suggested).

Good job!

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
witekest: Reset to the original version.

java/code/src/com/redhat/rhn/domain/dto/EndpointInfo.java
mcalmer: Please add the copyright line everywhere. without it we cannot let this pass. When the checkstyle failed, we need to check why. But it is not an option to remove it.

witekest: Done

mbologna (approved):

rjmateus (approved):

juliogonzalez:
As discussed at the chat, this arrives too late for Uyuni 2021.08. It will be merged after the release and will be part of the next Uyuni release.

PR #2986, opened 2020-12-08
Testsuite: Fix get_gpg_keys function

calancha:
## What does this PR change?
It fixes get_gpg_keys function

- testsuite/features/support/client_stack.rb:
Get the host and pass it as argument to get_os_version.

### Ports
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/13385
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13386

- [ ] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

PR #2989, opened 2020-12-08
Add second arg head

calancha:
## What does this PR change?

Pass the right type for the first argument
Add a 2nd optional arg with the target from where to obtain the GPG keys ($server, default).

## Links

Fixes https://github.com/uyuni-project/uyuni/pull/2986
### Ports
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/13393
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13392

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -667,7 +667,7 @@ When(/^I bootstrap (traditional|minion) client "([^"]*)" using bootstrap script 
   # Prepare bootstrap script for different types of clients
   client = client_type == 'traditional' ? '--traditional' : ''
   node = get_target(host)
-  gpg_keys = get_gpg_keys(node)
+  gpg_keys = get_gpg_keys(node, target)
   cmd = "mgr-bootstrap #{client} &&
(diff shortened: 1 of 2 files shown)


lkotek (approved):
Thanks for fix :-)

srbarrios:
Thanks @calancha, it is better passing directly the node as we do in the other functions indeed 👍

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -667,7 +667,7 @@
   # Prepare bootstrap script for different types of clients
   client = client_type == 'traditional' ? '--traditional' : ''
   node = get_target(host)
-  gpg_keys = get_gpg_keys(node)
+  gpg_keys = get_gpg_keys(node, target)
```
srbarrios, line 670: Naming it is just a bit confusing now, as to obtain the node we use the function get_target, and then as second parameter of the next function we use the variable target extracted from few lines above 🤔
I just hope it is clear enough the next time we revisit this code

calancha: I agree; there is room for improvement in the naming; it is worth improving it.

PR #2990, opened 2020-12-09
Testsuite: Pacify Rubocop

calancha:
## What does this PR change?
Rewrite a list of constant strings as suggested by Rubocop

### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13400
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/13401

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -69,12 +69,17 @@ end
 #
 # This is a safety net only, the best thing to do is to not start the reposync at all.
 # rubocop:disable Metrics/MethodLength
+# rubocop:disable Metrics/BlockLength
 def compute_list_to_leave_running
   do_not_kill = []
   if $long_tests_enabled
     # keep the repos needed for the auto-installation tests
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

srbarrios:
My personal opinion is that we should move these lists to a map of key(distro)-value(channels_list) in constants.rb
But... well I know this is also a controversial topic :p

lkotek (changes requested):
Thanks for changes! I see we currently hit different rubocop problem (maximum method length, at least in case of ports). 

Can you align the list of repos for 15SP1 and 15SP2, so we have additional ones like in the case of 4.1, please?

Apart of this I don't have problem with formatting if it resolves the rubocop problem, but I am also interested what @Bischoff opinion about it is.

calancha:
One channel per line as suggested by Rubocop, greatly improves readability.

Bischoff:
yeah formatting is just fine @lkotek, fanks for the fix @calancha

PR #2994, opened 2020-12-09
Fix permission issue in Salt user group

Etheryte:
## What does this PR change?

Fix `salt_user_group` being set to `root` which caused minion bootstrapping to fail due to permission issues.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Bugfix.

- [x] **DONE**

## Test coverage
- No tests: Needs to pass existing tests.

- [x] **DONE**

## Links

Related test runs: https://ci.suse.de/view/Manager/view/Manager-Head/job/manager-Head-cucumber-NUE/1046/testReport/

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/spacewalk-java.spec
+++ b/java/spacewalk-java.spec
@@ -32,14 +32,13 @@
 %define appdir          /srv/tomcat/webapps
 %define jardir          /srv/tomcat/webapps/rhn/WEB-INF/lib
 %define apache_group    www
-%define salt_user-group    salt
+%define salt_user_group salt
 %define apache2         apache2
 %define java_version    11
 %else
 %define appdir          %{_localstatedir}/lib/tomcat/webapps
 %define jardir          %{_localstatedir}/lib/tomcat/webapps/rhn/WEB-INF/lib
 %define apache_group    apache
-%define salt_user_group root 
 %define apache2         httpd
 %define java_version    1:11
 %endif
--- a/java/spacewalk-java.spec
+++ b/java/spacewalk-java.spec
@@ -39,6 +39,7 @@
 %define appdir          %{_localstatedir}/lib/tomcat/webapps
 %define jardir          %{_localstatedir}/lib/tomcat/webapps/rhn/WEB-INF/lib
 %define apache_group    apache
+%define salt_user_group salt
 %define apache2         httpd
 %define java_version    1:11
 %endif


Etheryte:
@sbluhm Could you please review this as well? I couldn't add you as a reviewer from the UI.

java/spacewalk-java.spec
```diff
@@ -32,14 +32,13 @@
...
 %define apache_group    www
-%define salt_user-group    salt
+%define salt_user_group salt
 %define apache2         apache2
 %define java_version    11
 %else
 %define appdir          %{_localstatedir}/lib/tomcat/webapps
 %define jardir          %{_localstatedir}/lib/tomcat/webapps/rhn/WEB-INF/lib
 %define apache_group    apache
-%define salt_user_group root
```
mcalmer, line 42: @sbluhm we have a special user because tomcat need to write in that directory. Not sure as which user tomcat is running on CentOS.

@Etheryte I think removing is wrong as it it must be defined when building this package for CentOS.

mcalmer (changes requested):
Either keep the root definition for CentOS or sbluhm tell you what user should be used.

java/spacewalk-java.spec
sbluhm: tomcat is actually by the user tomcat on CentOS7. Why are you running into the else statement? For SUSE it should stay as `salt`.

1. We could put back the user as `salt`  in line 702 and 703 and just keep the variable salt_user_group as salt_user. We would cover the groups salt (SUSE), tomcat (other).
2. Add tomcat to the group salt. I assume this is being done for SUSE systems?

mcalmer: I think I was wrong. tomcat uses salt-api to write in this directory. So it is important under which user salt-api/salt-master is running. With the salt package shipped with Uyuni this is the user "salt".

I would not suggest to use a different salt package, as upstream is slow with merging our new features. Even the latest salt upstream package might miss features which are required for uyuni.

So I think this is now good to be merged. We can make detail changes at any time later if needed.

sbluhm: To remove complexity, does it make sense to revert my changes and hardcode the user/group again? Or the least remove the definition out of the if-loop?

Etheryte: Personally I think it's fine as-is right now, as good overview of the similarities and differences in the setups similar to apache etc. I'm going to merge this now to see if it resolves the test suite issues fully, and we can come back to this later if it rubs someone the wrong way.

PR #2995, opened 2020-12-09
Allow to specify 'Revision' when updating 'init.sls' via CLI or API

ricardoasmarques:
## What does this PR change?

With this PR, users will be able to set the "Revision Number" when updating `init.sls` via CLI:

```
spacecmd> configchannel_updateinitsls ... -r <revision>
```
Or via API:

```
configchannel.updateInitSls(..., <revision>)
```

Note that this is already supported by the WebUI, so we are only adapting the CLI and API accordingly.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: The API documentation, which is acessible via "Help > API" on the left menu, is automatically generated.

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1179566 (Partially)

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/configchannel/ConfigChannelHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/configchannel/ConfigChannelHandler.java
@@ -379,7 +379,7 @@ public class ConfigChannelHandler extends BaseHandler {
      * Update the init.sls file for the given state channel with the given contents.
      * @param user The current user
      * @param channelLabel the label of the config channel.
-     * @param data a map containing 'content' and 'contents_enc64'
+     * @param data a map containing 'content', 'contents_enc64' and 'revision'
      * @return returns the updated config revision..
      *
      * @xmlrpc.doc Update the init.sls file for the given state channel. User can only update contents, nothing else.
@@ -389,6 +389,7 @@ public class ConfigChannelHandler extends BaseHandler {
      *  #struct_begin("path info")
      *      #prop_desc("string","contents", "Contents of the init.sls file")
      *      #prop_desc("boolean","contents_enc64", "Identifies base64 encoded content(default: disabled)")
+     *      #prop_desc("int", "revision", "next revision number, auto increment for null")
      *  #struct_end()
      * @xmlrpc.returntype
      * $ConfigRevisionSerializer
@@ -399,6 +400,7 @@ public class ConfigChannelHandler extends BaseHandler {
         Set<String> validKeys = new HashSet<String>();
         validKeys.add(ConfigRevisionSerializer.CONTENTS);
         validKeys.add(ConfigRevisionSerializer.CONTENTS_ENC64);
+        validKeys.add(ConfigRevisionSerializer.REVISION);
         validateMap(validKeys, data);
         XmlRpcConfigChannelHelper helper = XmlRpcConfigChannelHelper.getInstance();
         ConfigChannel channel = helper.lookupGlobal(user, channelLabel);
@@ -415,6 +417,9 @@ public class ConfigChannelHandler extends BaseHandler {
             form.setGroup(configInfo.getGroupname());
             form.setOwner(configInfo.getUsername());
             form.setPermissions(configInfo.getFilemode().toString());
+            if (data.containsKey(ConfigRevisionSerializer.REVISION)) {
+                form.setRevNumber(String.valueOf(data.get(ConfigRevisionSerializer.REVISION)));
+            }
             result = ConfigFileBuilder.getInstance().update(form, user, configFile);
 
          }
--- a/spacecmd/src/spacecmd/configchannel.py
+++ b/spacecmd/src/spacecmd/configchannel.py
@@ -848,6 +848,7 @@ def help_configchannel_updateinitsls(self):
 
 options:
   -c CHANNEL
+  -r REVISION
   -f local path to file contents
   -y automatically proceed with file contents
 '''))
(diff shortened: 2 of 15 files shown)


ricardoasmarques:
@moio Do I need to update `spacecmd/po/*.po` files in this PR?

spacecmd/src/spacecmd/configchannel.py
```diff
@@ -847,6 +847,7 @@ def help_configchannel_updateinitsls(self):
 
 options:
   -c CHANNEL
+  -r REVISION
```
moio, line 850: It would be cool to have a (Cucumber?) test to cover this.

java/code/src/com/redhat/rhn/frontend/xmlrpc/configchannel/ConfigChannelHandler.java
```diff
@@ -389,6 +389,7 @@ public ConfigChannel update(User loggedInUser, String label,
      *  #struct_begin("path info")
      *      #prop_desc("string","contents", "Contents of the init.sls file")
      *      #prop_desc("boolean","contents_enc64", "Identifies base64 encoded content(default: disabled)")
+     *      #prop_desc("int", "revision", "next revision number, auto increment for null")
```
moio, line 392: What are the pros and cons to add this as a field in this map rather than a full new parameter?

moio (commented):
@ricardoasmarques, thanks for this PR.

My understanding is that XMLRPC Javadoc is not translated, rationale being that users need to know English to make sense of method names, parameter names and so on. Basically the assumptions is that programmers know English, while sysadmins may not.

`spacecmd`, OTOH, is a translated tool, so that means this PR is expected to update the English version of the files. Please ping @parlt91 if you need any further guidance on that topic.

java/code/src/com/redhat/rhn/frontend/xmlrpc/configchannel/ConfigChannelHandler.java
ricardoasmarques: **PROS**
- Consistency with the syntax already used on `configchannel.createOrUpdatePath`, which is the equivalent method for updating "files" (instead of "states")
- No need to overload this method to guarantee that we are not breaking the current method signature
- Do not pollute method signature with optional parameters
- Since `contents` is a required field, and lives inside the map, It doesn't seem right to keep the **required** `contents` field in the map and the **optional** `revision` field as a new parameter (since `revision` is less important and less used)

**CONS**
- Field is not explicitly declared, so we have to rely on documentation to check which "fields" are accepted, but unfortunately Java does not have a great way to deal with optional parameters (requires method overload for that)

Let me know if you prefer a new parameter, so I can update the PR accordingly.

ricardoasmarques:
> ...
> `spacecmd`, OTOH, is a translated tool, so that means this PR is expected to update the English version of the files. Please ping @parlt91 if you need any further guidance on that topic.

As discussed with @parlt91, there's no need to update `*.po` files in this PR as they will be updated next time we merge translations from https://l10n.opensuse.org/projects/uyuni/

spacecmd/src/spacecmd/configchannel.py
ricardoasmarques: Sure! I'll add a Cucumber test for this.

java/code/src/com/redhat/rhn/frontend/xmlrpc/configchannel/ConfigChannelHandler.java
moio: > **PROS**
> 
>     * Consistency with the syntax already used on `configchannel.createOrUpdatePath`, which is the equivalent method for updating "files" (instead of "states")

Good point.

>     * No need to overload this method to guarantee that we are not breaking the current method signature

Overloading is not a problem with Java and XMLRPC, so that is not a concern I would say.

>     * Do not pollute method signature with optional parameters
> 
>     * Since `contents` is a required field, and lives inside the map, It doesn't seem right to keep the **required** `contents` field in the map and the **optional** `revision` field as a new parameter (since `revision` is less important and less used)
> 
> 
> **CONS**
> 
>     * Field is not explicitly declared, so we have to rely on documentation to check which "fields" are accepted, but unfortunately Java does not have a great way to deal with optional parameters (requires method overload for that)
> 
> 
> Let me know if you prefer a new parameter, so I can update the PR accordingly.

I am convinced about the pros, OK with the current direction :+1:

moio (approved):

ricardoasmarques:
> > ...
> > `spacecmd`, OTOH, is a translated tool, so that means this PR is expected to update the English version of the files. Please ping @parlt91 if you need any further guidance on that topic.
> 
> As discussed with @parlt91, there's no need to update `*.po` files in this PR as they will be updated next time we merge translations from https://l10n.opensuse.org/projects/uyuni/

After further discussion, we concluded that `*.po` files should be updated in this PR by running `make update-po`, and then `make clean` to remove `*.gmo` files.

ricardoasmarques:
> > > ...
> > > `spacecmd`, OTOH, is a translated tool, so that means this PR is expected to update the English version of the files. Please ping @parlt91 if you need any further guidance on that topic.
> > 
> > 
> > As discussed with @parlt91, there's no need to update `*.po` files in this PR as they will be updated next time we merge translations from https://l10n.opensuse.org/projects/uyuni/
> 
> After further discussion, we concluded that `*.po` files should be updated in this PR by running `make update-po`, and then `make clean` to remove `*.gmo` files.

Done.

spacecmd/src/spacecmd/configchannel.py
ricardoasmarques: @moio I've now updated this PR with a Cucumber test that will guarantee `-r REVISION` option is working properly.

moio: You have a green light from my side! Thanks!

PR #3008, opened 2020-12-11
Fix "working directory" on cucumber tests that rely on downloaded files

ricardoasmarques:
## What does this PR change?

PR https://github.com/uyuni-project/uyuni/pull/2921 assumed that E2E tests are executed from the `/root` location on the `controller` node.

If `CI` is running E2E tests from a different location, we will need to adapt accordingly.

This PR assumes that E2E tests are executed from `/root/spacewalk/testsuite/` location.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: just fixing tests

- [x] **DONE**

## Test coverage
- Cucumber tests were fixed

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_action_chain.feature
+++ b/testsuite/features/secondary/min_action_chain.feature
@@ -99,8 +99,8 @@ Feature: Action chains on Salt minions
     And I follow "List/Remove Files"
     And I follow "/etc/action-chain.cnf"
     And I follow "Download File"
-    And I wait until file "/root/action-chain.cnf" exists on "localhost"
-    Then file "/root/action-chain.cnf" should contain "Testchain=YES_PLEASE" on "localhost"
+    And I wait until file "/tmp/downloads/action-chain.cnf" exists on "localhost"
+    Then file "/tmp/downloads/action-chain.cnf" should contain "Testchain=YES_PLEASE" on "localhost"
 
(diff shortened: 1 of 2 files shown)


ricardoasmarques:
@lkotek is `/root/spacewalk/testsuite/` the working directory where E2E tests are executed?

srbarrios:
> @lkotek is `/root/spacewalk/testsuite/` the working directory where E2E tests are executed?

Can we make it differently, so we specify for instance /tmp/ folder as the path to download and handle this file?
We are assuming that we always run the tests from the controller, but it could not be the case. For instance... when I make runs from my IDE (IntelliJ)

ricardoasmarques:
> > @lkotek is `/root/spacewalk/testsuite/` the working directory where E2E tests are executed?
> 
> Can we make it differently, so we specify for instance /tmp/ folder as the path to download and handle this file?
> We are assuming that we always run the tests from the controller, but it could not be the case. For instance... when I make runs from my IDE (IntelliJ)

@srbarrios this is definitely the best approach. I've now updated this PR accordingly, please review again. Thanks.

srbarrios (approved):

PR #3015, opened 2020-12-14
mgr-cfg: Updated Python definitions for RHEL8 and quoted text comparisons.

sbluhm:
## What does this PR change?

- Updated Python definitions for RHEL8.
- Quoted text comparisons as words are not supported anymore.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: Tested during automatic build.

Manually tested successfully on CentOS6+, LEAP 15.2, Fedora32+

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/client/tools/mgr-cfg/mgr-cfg.spec
+++ b/client/tools/mgr-cfg/mgr-cfg.spec
@@ -33,6 +33,7 @@
 %if 0%{?fedora} || 0%{?suse_version} > 1320 || 0%{?rhel} >= 8
 %global build_py3   1
 %global default_py3 1
+%global __python /usr/bin/python2 
 %endif
 
 %if ( 0%{?fedora} && 0%{?fedora} < 28 ) || ( 0%{?rhel} && 0%{?rhel} < 8 ) || 0%{?suse_version} || 0%{?ubuntu} || 0%{?debian}
@@ -62,7 +63,7 @@
 
 %define pythonX %{?default_py3: python3}%{!?default_py3: python2}
 
-%if %{_vendor} == "debbuild"
+%if "%{_vendor}" == "debbuild"
 # Bash constructs in scriptlets don't play nice with Debian's default shell, dash
 %global _buildshell /bin/bash
 %endif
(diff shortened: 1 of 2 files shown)


client/tools/mgr-cfg/mgr-cfg.spec
```diff
@@ -125,16 +126,24 @@ Obsoletes:      python-%{name} < %{oldversion}
 Provides:       python2-%{oldname} = %{oldversion}
 Obsoletes:      python2-%{oldname} < %{oldversion}
 Requires:       %{name} = %{version}-%{release}
+%if 0%{?suse_version} || 0%{?rhel} >= 8
+Requires:       python2
```
mcalmer, line 130: On SLE12 this result in an "unresolvable: nothing provides python2".
Just do this %if for `rhel` only. At SUSE we have the definition python == python2 .

client/tools/mgr-cfg/mgr-cfg.spec
```diff
@@ -125,16 +126,24 @@ Obsoletes:      python-%{name} < %{oldversion}
...
 Requires:       python
+%endif
 Requires:       python2-rhn-client-tools >= 2.8.4
 Requires:       rhnlib >= 2.8.3
 Requires:       python2-uyuni-common-libs
 %if 0%{?rhel} && 0%{?rhel} <= 5
 Requires:       python-hashlib
 %endif
+%if 0%{?suse_version} || 0%{?rhel} >= 8
+BuildRequires:  python2
```
mcalmer, line 141: same as above: just remove the `suse_version` here.

mcalmer (changes requested):

client/tools/mgr-cfg/mgr-cfg.spec
sbluhm: ok, I think I need more understanding on this as I hit this before.
My understanding (and tests on LEAP 15.2), SUSE detects the required Python version. so if you put
`Requires: python`
it will either resolve to python2 or python3. On my test builds, the package failed because of missing Python 2 (3 was installed instead). This is why I am forcing Python 2 by using Requires: python2.
Or should I also include the actual SUSE version number? (I noticed you wrote SLE12).

mcalmer: This would be new to me. In SLE and AFAIK also in Leap python is always python2.

https://build.opensuse.org/package/binary/openSUSE:Leap:15.2/python/standard/x86_64/python-2.7.17-lp152.2.17.x86_64.rpm

The package name in Leap is really only "python". python2 is a "Provides" value. 

Ahh, but our "python3" package "Provides: python = 3.x".

So if you found problems, we can say 

Requires: python < 3

sbluhm: THANK YOU! I love your approach. This makes it a lot simpler. I believe I have to revisit one or two of my previous submissions for this.

sbluhm: @mcalmer , will the "or" work for SLE12?  If yes, I would do:
`Requires:       (python < 3 or python2)`

mcalmer: No, I think it will not work on SLE12.

mcalmer: We build this package also for RHEL6 and SLES11. So keep the python2 for rhel >= 8 only, please.

sbluhm:
Tested build successfully on Leap15.2, CentOS6+, Fedora 32+

mcalmer (approved):

PR #3017, opened 2020-12-14
Testsuite - do not use beta client tools for SLE15

lkotek:
## What does this PR change?

PR unselects `SUSE Manager Tools 15 x86_64 (BETA)` so only released client tools are used.

## Links

https://github.com/SUSE/spacewalk/issues/13284

## Changelogs

- [x] No changelog needed

## Re-run a test


juliogonzalez (approved):

lkotek:
I put a comment with explanation. It is very easy to miss this line. Please share if you disagree :-)

juliogonzalez:
I guess that ideally we should have a flag somewhere. But not for this PR. Just think about it as it will be needed for testing milestones, I guess :-)

srbarrios (approved):

lkotek:
> I guess that ideally we should have a flag somewhere. But not for this PR. Just think about it as it will be needed for testing milestones, I guess :-)

Yep, that sounds very reasonable :-)

PR #3025, opened 2020-12-15
Testsuite - remove duplicity scenario and use tag for suma specific one

lkotek:
## What does this PR change?

This PR adds tag for SUSE Manager specific scenario and removes `List all products for Uyuni` scenario as this is fully covered by the steps from previous one (`List all products for SUSE Manager`) and also is specific for SUSE Manager again.

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/reposync/srv_sync_channels.feature
+++ b/testsuite/features/reposync/srv_sync_channels.feature
@@ -23,6 +23,7 @@ Feature: Be able to list available channels and enable them
     And I shouldn't get "sles11-extras"
 
 @scc_credentials
+@susemanager
   Scenario: List products
     When I execute mgr-sync "list products"
     Then I should get "[ ] SUSE Linux Enterprise Server 12 SP4 x86_64"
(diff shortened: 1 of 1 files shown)


calancha (approved):

Bischoff:
Nice catch @lukas

srbarrios:
I'm sorry @lkotek  but I don't agree :p 

I think we should let:

```
@scc_credentials
@uyuni
  Scenario: List all products for Uyuni
```

As this will not be covered in any place when running our Uyuni CI. The fix will be to check for different products, so instead of these ones:

```
    Then I should get "[ ] SUSE Linux Enterprise Server 12 SP4 x86_64"
    And I should get "[ ] SUSE Manager Proxy 4.0 x86_64"
```

We need to check other products visible in the list :

```
uyuni-master-srv:~ # mgr-sync list products --expand
Available Products:

(R) - recommended extension

Status:
  - [I] - product is installed
  - [ ] - product is not installed, but is available
  - [U] - product is unavailableI'm sorry but I don't agree :p 

I think we should let:

```
@scc_credentials
@uyuni
  Scenario: List all products for Uyuni
```

As this will not be covered in any place when running our Uyuni CI. The fix will be to check for different products, so instead of these ones:

```

```


[ ] Ubuntu 18.04
[ ] Ubuntu 16.04
[ ] RHEL5 Base x86_64
[ ] RHEL5 Base i386
[ ] RHEL6 Base x86_64
[ ] RHEL6 Base i386
[ ] RHEL7 Base x86_64
....
```

lkotek:
@srbarrios Nothing to sorry about :-) Different opinions are welcome ;-) Btw. we still have this scenario: 
```
@scc_credentials
  Scenario: List products with filter
    When I execute mgr-sync "list products --expand --filter x86_64"
    Then I should get "[ ] SUSE Linux Enterprise Server 12 SP4 x86_64"
    And I shouldn't get "ppc64"
    And I shouldn't get "s390x"
```
The only difference is `--filter x86_64` used.

srbarrios:
Ah right! So then it's fine! I just went through the PR and the description :)
We are good then, we can just test both together.

PR #3027, opened 2020-12-15
Allow to specify file contents when creating state channel via API

ricardoasmarques:
## What does this PR change?

With this PR, users will be able to provide "init.sls file contents" when creating a new state config channel via XML-RPC.

**Motivation:** Allow "inter server sync" to create an `init.sls` with contents and "revision 1".

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: The API documentation, which is acessible via "Help > API" on the left menu, is automatically generated.

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Fixes: https://github.com/SUSE/spacewalk/issues/13345

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/configchannel/ConfigChannelHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/configchannel/ConfigChannelHandler.java
@@ -54,6 +54,7 @@ import com.redhat.rhn.manager.system.SystemManager;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


moio (approved):
LGTM, thanks!

In general, we customarily at least CC a testsuite expert (eg. @lkotek) before merging any PR that touches the testsuite code.

Also, I regret not telling you earlier, but we frequently test XMLRPC endpoints in JUnit tests rather than integration tests - mostly because it's quicker to do so, and normally it's good enough. In this case you covered the feature in Cucumber and that's even better, but feel free to evaluate the JUnit path as well next time :wink:

PR #3044, opened 2020-12-17
Typescript integration & migration helper

Etheryte:
## What does this PR change?

Add Typescript to the build pipeline, tooling, configurations etc.  
Additionally, adds a migration helper script that takes care of the bigger more common issues.  

There's a bunch of commits in this PR because I had to hack around a fair bit to find a good way to approach the bulk of this, but this can be squashed down before merging.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Includes some developer documentation, but could surely use more. I will firstly try migrating some core parts of the codebase with this approach to get a better feeling for the workflow and then write more relevant guidelines.

- [x] **DONE**

## Test coverage
- No tests: Needs to pass all existing tests.

- [x] **DONE**

## Links

Part of https://github.com/SUSE/spacewalk/issues/13208

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
ncounter, line 1: I think this file was not intended to be added in the git history, is it?

Etheryte: This is actually intentional since the IDE cannot find the correct Typescript instance otherwise. We can omit it, but then multiple other people will run into the same problem and need to look for a solution. Unless you have other suggestions?

ncounter (approved):
Very good job, thanks a lot @Etheryte ! :medal_military:

Issue #3040, opened 2020-12-17
Unexpected error: <class 'spacewalk.server.importlib.importLib.InvalidPackageError'>


StefanSa:
# Additional Information
i get this error message while importing / linking a new repo.

```
2020/12/17 09:14:54 +02:00   Linking packages to the channel.
2020/12/17 09:14:55 +02:00 Unexpected error: <class 'spacewalk.server.importlib.importLib.InvalidPackageError'>
2020/12/17 09:14:55 +02:00 Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/reposync.py", line 592, in sync
    ret = self.import_packages(plugin, data['id'], url, is_non_local_repo)
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/reposync.py", line 1107, in import_packages
    importer.run()
  File "/usr/lib/python3.6/site-packages/spacewalk/server/importlib/importLib.py", line 777, in run
    self.submit()
  File "/usr/lib/python3.6/site-packages/spacewalk/server/importlib/packageImport.py", line 126, in submit
    self.backend.lookupPackages(self.batch, self.checksums)
  File "/usr/lib/python3.6/site-packages/spacewalk/server/importlib/backend.py", line 822, in lookupPackages
    raise_with_tb(not_found[0], not_found[1])
  File "/usr/lib/python3.6/site-packages/uyuni/common/usix.py", line 77, in raise_with_tb
    raise value
  File "/usr/lib/python3.6/site-packages/spacewalk/server/importlib/backend.py", line 814, in lookupPackages
    self.__lookupObjectCollection([package], 'rhnPackage')
  File "/usr/lib/python3.6/site-packages/spacewalk/server/importlib/backend.py", line 2493, in __lookupObjectCollection
    raise InvalidPackageError(object, "Could not find object %s in table %s" % (object, tableName))
[...]

PR #2157, opened 2020-04-24
Reposync speedup part 2

moio:
## What does this PR change?

It replaces the existing mechanism in `spacewalk-repo-sync` to insert package capabilities, changelogs and checksums with a more efficient one.

**Total speedup is ~25%**, individual contributions from relevant commits as measured in my setup (syncing `sles12-sp5-updates-x86_64` from a local mirror, robustly transferable to other channels) follow:

| Commit                                                | Speedup |
|-------------------------------------------------------|---------|
| reposync speedup: use execute_values for capabilities | 5.18%   |
| Add INDEX to rhnPackageCapability.name                | 8.16%   |
| reposync speedup: use execute_values for checksums    | 4.57%   |
| reposync speedup: use execute_values for changelogs   | 5.28%   |

## The old mechanism being replaced

Capabilities/checksums/changelogs were inserted one by one via a `lookup_` stored procedure. The stored procedure was coded in a way to avoid that inserting an already-existing object would cause a unique constraint error - it did so by opening a secondary Postgres connection that was dedicated to just those insertions.

## The new mechanism proposed in this PR

Main differences are:

 * normal `INSERT INTO` statements are used instead of stored procedures
 * insertions are batched (one statement for many `VALUE`s)
 * batching happens via the via the `psycopg2` [execute_many helper](https://www.psycopg.org/docs/extras.html#fast-execution-helpers#psycopg2.extras.execute_values). According to the library documentation and my tests [using such helper is much faster than simply looping over a statement or the previously-used `executemany` method](https://www.psycopg.org/docs/extras.html#fast-execution-helpers)
 * potential conflicts with concurrent insertions are resolved via the the `INSERT INTO ... ON CONFLICT DO NOTHING` [clause of PostgreSQL](https://www.postgresql.org/docs/12/sql-insert.html#SQL-ON-CONFLICT) instead of the secondary connection
 * code implementing the secondary connection is dropped

This requires the latest psycopg2 library in order to work. [This is being added to SLE as part of the addition of PostgreSQL 12](https://smelt.suse.de/request/215193/) and [we will add the package to Uyuni temporarily](https://build.opensuse.org/request/show/797156).

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **internal change only**
- [x] **DONE**

## Test coverage
- No tests: **Cucumber tests to be implemented after reposync is fast enough**

- [x] **DONE**

## Links

Fixes partially https://github.com/SUSE/spacewalk/issues/9275
[...]

PR #3047, opened 2020-12-18
Improve extra key handling

mcalmer:
## What does this PR change?

Adding new extratag keys during reposync can fail with 

```
duplicate key value violates unique constraint "rhn_pkg_extra_tag_key_idx"
DETAIL:  Key (name)=(Cnf-Extra-Commands) already exists.
```
This result in failed imports of a bunch of packages. To improve the creation of new keys, this PR declare lookup and insert functions and make use of it in the import code.

Second case found while inserting into rhnPackageKey.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **manual**

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/3040
Tracks https://github.com/SUSE/spacewalk/pull/13545

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/backend/server/importlib/backend.py
+++ b/backend/server/importlib/backend.py
@@ -307,12 +307,12 @@ class Backend:
             missing AS (
               SELECT nextval('rhn_package_extra_tags_keys_id_seq') AS id, wanted.name
                 FROM wanted
-           LEFT JOIN rhnPackageExtraTagKey ON rhnPackageExtraTagKey.name = wanted.name
+                    LEFT JOIN rhnPackageExtraTagKey ON rhnPackageExtraTagKey.name = wanted.name
                WHERE rhnPackageExtraTagKey.id IS NULL
             )
             INSERT INTO rhnPackageExtraTagKey (id, name)
-            SELECT * from missing
-            ON CONFLICT DO NOTHING
+                SELECT * from missing
+                ON CONFLICT DO NOTHING
         """
         values = [key for key in extraTags.keys() if key != '']
         if not values:
--- a/backend/server/importlib/backend.py
+++ b/backend/server/importlib/backend.py
@@ -60,8 +60,7 @@ sequences = {
     'suseEula': 'suse_eula_id_seq',
     'suseProducts': 'suse_products_id_seq',
     'suseSCCRepository': 'suse_sccrepository_id_seq',
-    'suseProductSCCRepository': 'suse_prdrepo_id_seq',
-    'rhnPackageExtraTagKey': 'rhn_package_extra_tags_keys_id_seq'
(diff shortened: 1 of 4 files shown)


moio (changes requested):
@mcalmer while I see that the patch here works for addressing the issue, I think a better approach could be use `INSERT INTO... ON CONFLICT DO NOTHING`. That has the advantage of allowing batches of insertions, which is faster than individual ("per object") calls to the lookup stored procedures. That would also allow not to add another "nested insert SP" hack.

This is the pattern I have been using during the reposync speedup, and would recommend keeping it in new code as well.

You can see an example of such patch here:

https://github.com/uyuni-project/uyuni/pull/2157/commits/82ee0130a08ce2f8dce9270b21aab7fd2d84b585

You can see several other instances of this pattern in [backlog.py](https://github.com/uyuni-project/uyuni/pull/2157/commits/82ee0130a08ce2f8dce9270b21aab7fd2d84b585) by grepping for `ON CONFLICT DO NOTHING`.

I hope I am not missing any important aspect that would prevent the above approach from working here - in that case just let me know.

backend/server/importlib/backend.py
```diff
@@ -299,40 +298,40 @@ def processCVEs(self, cveHash):
...
+        if not extraTags:
+            return
+        sql = """
+            WITH wanted (name) AS (
+              VALUES %s
+            )
+            missing AS (
+              SELECT nextval('rhn_package_extra_tags_keys_id_seq') AS id, wanted.name
+                FROM wanted
+           LEFT JOIN rhnPackageExtraTagKey ON rhnPackageExtraTagKey.name = wanted.name
```
moio, line 310:
```suggestion
                    LEFT JOIN rhnPackageExtraTagKey ON rhnPackageExtraTagKey.name = wanted.name
```

backend/server/importlib/backend.py
```diff
@@ -299,40 +298,40 @@ def processCVEs(self, cveHash):
...
+               WHERE rhnPackageExtraTagKey.id IS NULL
+            )
+            INSERT INTO rhnPackageExtraTagKey (id, name)
+            SELECT * from missing
```
moio, line 314:
```suggestion
                SELECT * from missing
```

backend/server/importlib/backend.py
```diff
@@ -299,40 +298,40 @@ def processCVEs(self, cveHash):
...
+            ON CONFLICT DO NOTHING
```
moio, line 315:
```suggestion
                ON CONFLICT DO NOTHING
```

moio (approved):
LGTM

PR #3055, opened 2020-12-22
Checked server-app-module in uyuni head testsuite

atighineanu:
## What does this PR change?
without this commit, the uyuni head testuite is
failing, because it expects to have all recommended channels.

Testsuite should not stumble into the `./features/step_definitions/common_steps.rb:551:in` error anymore, since all recommended channels are checked accordingly.

## Links

uyuni branch only

## Changelogs

- [x] No changelog needed


Bischoff (approved):
LGTM

PR #3063, opened 2020-12-30
Cleanup tomcat versions

mcalmer:
## What does this PR change?

In our build configs we still have configurations and targets for tomcat5.
Currently we only support tomcat9.
Cleanup all the configs and keep only tomcat9 and tomcat which is also configured to version 9 just in case
on CentOS a different config is needed.
It also give us an idea how to configure a tomcat10 in future.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **build config changes only**

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


mcalmer:
@sbluhm this might be interesting for your as well. Not sure what tomcat config CentOS need here.

sbluhm (approved):

sbluhm (approved):
Looking good on CentOS8. I have just built, installed and ran it.

moio (approved):
I have no problem with the cleanups proposed here. As always, there's no shortage of other things to clean up! But those can of course go into a new PR.

Thanks for this contribution!

PR #3075, opened 2021-01-05
Make HMR and key usage consistent

Etheryte:
## What does this PR change?

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1180580. The problem was a mix of old and new HMR use, but it was triggered by the key issue in the related page.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: No gui diff, fixes existing page.

- [x] **DONE**

## Test coverage
- No tests: Needs to pass existing tests.

- [x] **DONE**

## Links

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1180580

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/content-management/create-project/create-project.js
+++ b/web/html/src/manager/content-management/create-project/create-project.js
@@ -6,7 +6,7 @@ import TopPanelButtons from "./top-panel-buttons";
 import PropertiesCreate from "../shared/components/panels/properties/properties-create";
 import {showErrorToastr} from "components/toastr/toastr";
 import withPageWrapper from 'components/general/with-page-wrapper';
-import {hot} from 'react-hot-loader';
+import { hot } from 'react-hot-loader/root';
 import useRoles from "core/auth/use-roles";
 import {isOrgAdmin} from "core/auth/auth.utils";
 import useLifecycleActionsApi from "../shared/api/use-lifecycle-actions-api";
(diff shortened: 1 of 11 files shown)


web/spacewalk-web.changes
```diff
@@ -1,3 +1,4 @@
+- Fix inconsistent HMR and key usage bug (bsc#1180580)
```
mcalmer, line 1: This changelog is often read by normal user of Uyuni/SUSE Manager. Try to write a message which they can understand. 

Like "fixing display error of Package State page" or something similar.

web/html/src/manager/content-management/create-project/create-project.js
```diff
@@ -6,7 +6,7 @@ import TopPanelButtons from "./top-panel-buttons";
 import PropertiesCreate from "../shared/components/panels/properties/properties-create";
 import {showErrorToastr} from "components/toastr/toastr";
 import withPageWrapper from 'components/general/with-page-wrapper';
-import {hot} from 'react-hot-loader';
+import { hot } from 'react-hot-loader/root';
```
cbbayburt, line 9: `react-hot-loader` [docs](https://github.com/gaearon/react-hot-loader#getting-started) say: "Make sure react-hot-loader is required before react and react-dom". Since you're already touching all the imports, maybe you can apply this as well.

cbbayburt (approved):
LGTM :+1: Thanks, @Etheryte for fixing it super quick :)

Just one thing in the comments..

web/html/src/manager/content-management/create-project/create-project.js
Etheryte: @cbbayburt Good catch, I've updated it manually for now and I'll see if we can add linting to automatically cover this for the future.

PR #3080, opened 2021-01-06
spacewalk-admin: Fixed salt certificate creation.

sbluhm:
## What does this PR change?

- Added salt-secrets-config to the default service list.
- Updated salt certificate script to variable certificate path and message on failing update-ca-certificates.
- Updated source URL in spec file.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: no functional changes

- [X] **DONE**

## Test coverage
- No tests: tested during automatic build.
Manual tests of code segments done on Leap 15.2 and CentOS 8.
Build tested successfully for CentOS 7/8 and Leap 15.2. Installation tested successfully on CentOS 8

- [X] **DONE**

## Links

Fixes #

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacewalk/admin/salt-secrets-config.py
+++ b/spacewalk/admin/salt-secrets-config.py
@@ -69,5 +69,5 @@ if (not all([os.path.isfile(f) for f in ["/etc/salt/pki/api/salt-api.crt", "/etc
     # Detect CA management tool.
     if os.system("update-ca-certificates"):
         print('Using "update-ca-trust" instead of "update-ca-certificates".')
-        os.system("update-ca-trust")
+        os.system("update-ca-trust extract")
 
--- a/spacewalk/admin/salt-secrets-config.py
+++ b/spacewalk/admin/salt-secrets-config.py
@@ -14,6 +14,11 @@ initCFG('java')
 
 thread_pool_size = CFG.salt_event_thread_pool_size
 
+# To be moved into a config file in future
+cert_location = "/etc/pki/trust/anchors"
+if not os.path.isdir(cert_location):
+    cert_location = "/etc/pki/ca-trust/source/anchors"
+
 initCFG()
 
 mgr_events_config = {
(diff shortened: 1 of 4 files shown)


spacewalk/admin/salt-secrets-config.py
```diff
@@ -60,5 +65,7 @@
     os.chmod("/etc/salt/pki/api/salt-api.crt", 0o600)
     os.chown("/etc/salt/pki/api/salt-api.key", pwd.getpwnam("salt").pw_uid, grp.getgrnam("salt").gr_gid)
     os.chmod("/etc/salt/pki/api/salt-api.key", 0o600)
-    shutil.copyfile("/etc/salt/pki/api/salt-api.crt", "/etc/pki/trust/anchors/salt-api.crt")
-    os.system("update-ca-certificates")
+    shutil.copyfile("/etc/salt/pki/api/salt-api.crt", cert_location + "/salt-api.crt")
+    if os.system("update-ca-certificates"):
+        print("update-ca-certificates failed. Perform the required activities manually.")
```
mcalmer, line 70: I think on CentOS you need to use this tool `/usr/bin/update-ca-trust`, right?
Why not putting it into another variable and enhance the check above to set also the ca tool to call?

spacewalk/admin/spacewalk.target
```diff
@@ -14,6 +14,7 @@ Requires=rhn-search.service
 Requires=cobblerd.service
 Requires=taskomatic.service
 Requires=spacewalk-wait-for-taskomatic.service
+Requires=salt-secrets-config.service
```
mcalmer, line 17: Oh yes. We did not really maintain spacewalk.target as we use the file spacewalk.target.SUSE .
Please check if other changes we made there should be ported as well.

Btw: we made jabberd, osa-dispatcher "optional" and start them manual in `spacewalk-service` script.
I you have a salt only system, you do not what them.

mcalmer (commented):
Some suggestions added

spacewalk/admin/salt-secrets-config.py
sbluhm: Thanks, I didn't know that. Will check it out and update.

sbluhm: Both tools seem to do a similar thing, so I added update-ca-trust. Before a final release, we really have to go through those hard coded /etc/pki folders and match them up...
I did follow a different strategy to your suggestion. Instead of trying to complicatedly detect the right tool at the beginning, I run one tool. If it fails, the other one is run. This keeps the code clean from one time variables and complex logic.

spacewalk/admin/spacewalk.target
sbluhm: Ah, I was indeed confused about the optional. I will add the line-up between the two files to my to-do list. Maybe we can even drop the .SUSE version for a generic one.

mcalmer (approved):
I made a minor change as the RHEL tool requires a parameter

PR #3084, opened 2021-01-06
Let time in "Pending" tab be configurable

Bischoff:
## What does this PR change?

When we bootstrap a minion, the time spent in "Pending" tab can be long for the first events.

This PR makes this time configurable.

## Links

Ports:
* 4.0: SUSE/spacewalk#13590
* 4.1: SUSE/spacewalk#13589

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -70,16 +70,23 @@ When(/^I wait until event "([^"]*)" is completed$/) do |event|
   step %(I wait at most #{DEFAULT_TIMEOUT} seconds until event "#{event}" is completed)
 end
 
-When(/^I wait at most (\d+) seconds until event "([^"]*)" is completed$/) do |final_timeout, event|
+When(/^I wait at most (\d+) seconds until event "([^"]*)" is completed$/) do |timeout, event|
+  starting = Time.now
+  initial_timeout = timeout.to_i
   # The code below is not perfect because there might be other events with the
   # same name in the events history - however, that's the best we have so far.
   steps %(
     When I follow "Events"
     And I follow "Pending"
-    And I wait until I do not see "#{event}" text, refreshing the page
+    And I wait at most #{initial_timeout} seconds until I do not see "#{event}" text, refreshing the page
     And I follow "History"
     And I wait until I see "System History" text
     And I wait until I see "#{event}" text, refreshing the page
+  )
+  ending = Time.now
+  final_timeout = timeout.to_i - (ending - starting).to_i
+  final_timeout = 3 if final_timeout < 3
+  steps %(
     And I follow first "#{event}"
     And I wait at most #{final_timeout} seconds until the event is completed, refreshing the page
   )
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -1,4 +1,4 @@
-# Copyright (c) 2010-2020 SUSE LLC.
+# Copyright (c) 2010-2021 SUSE LLC.
 # Licensed under the terms of the MIT license.
 
 #
@@ -91,9 +91,9 @@ When(/^I wait until I see the name of "([^"]*)", refreshing the page$/) do |host
   step %(I wait until I see "#{system_name}" text, refreshing the page)
 end
 
-When(/^I wait until I do not see "([^"]*)" text, refreshing the page$/) do |text|
+When(/^I wait at most (\d+) seconds until I do not see "([^"]*)" text, refreshing the page$/) do |timeout, text|
   next unless has_content?(text)
-  repeat_until_timeout(message: "Text '#{text}' is still visible") do
+  repeat_until_timeout(timeout: timeout.to_i, message: "Text '#{text}' is still visible") do
     break unless has_content?(text)
     begin
       accept_prompt do
@@ -103,6 +103,10 @@ When(/^I wait until I do not see "([^"]*)" text, refreshing the page$/) do |text
       # ignored
     end
   end
+end
+
+When(/^I wait until I do not see "([^"]*)" text, refreshing the page$/) do |text|
+  step %(I wait at most #{DEFAULT_TIMEOUT} seconds until I do not see "#{text}" text, refreshing the page)
 end
 
 When(/^I wait until I do not see the name of "([^"]*)", refreshing the page$/) do |host|


lkotek (approved):
Yes, I understand your point. Approved :+1:

Bischoff:
Testing locally first

calancha (approved):

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -70,18 +70,24 @@
...
   steps %(
     When I follow "Events"
     And I follow "Pending"
-    And I wait until I do not see "#{event}" text, refreshing the page
+    And I wait at most #{timeout} seconds until I do not see "#{event}" text, refreshing the page
     And I follow "History"
     And I wait until I see "System History" text
     And I wait until I see "#{event}" text, refreshing the page
+  )
+  ending = Time.now
```
srbarrios, line 85: I'd avoid this instantiation, and use Time.now directly on the next line.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -70,18 +70,24 @@
...
+  timeout -= ending - starting
+  timeout = 3 if timeout < 3
```
srbarrios, line 87: Do we really need to use a different timeout value for the next step?
I'd use the timeout value coming from the step definition parameter in both steps without taking into account this difference of seconds "spent" in the first block. The code would be much easier and the advantage of having a bit smaller timeout is in my opinion none.

Bischoff: It's a matter of semantics of this step.

We want to wait "at most" XX seconds in this step, but if we do like you say we will in fact spend at most XX seconds in first timeout and agains at most XX seconds in second step. For a total wait that can be up to 2 * XX seconds.

Now, if we accept that the step does not do exactly like the English sentence pretends it does, we can do like you want. It will indeed produce simpler code.

Alternatively, we could pass 2 parameters to the event, one for the time spent in "pending", and the other one for the time spent in running the event.

testsuite/features/step_definitions/common_steps.rb
Bischoff: code more readable...

Bischoff:
@oscar-barrios apart from the implementation details, I am having doubts about this anyway. Is more than 4 minutes acceptable for getting the event to be picked up at bootstrap? Am I trying to solve a product issue with this change to the testsuite?

@moio, as performance specialist, could you propvide your advice here?

srbarrios:
> @oscar-barrios apart from the implementation details, I am having doubts about this anyway. Is more than 4 minutes acceptable for getting the event to be picked up at bootstrap? Am I trying to solve a product issue with this change to the testsuite?
> 
> @moio, as performance specialist, could you propvide your advice here?

It depends on the conditions, as it picks the events FIFO, if we are running in parallel a dozen different events on different systems, it might indeed take more time. As we experienced in QAM testsuite during the smoke tests.

testsuite/features/step_definitions/common_steps.rb
srbarrios: Ok, I understand now your concern :)
Might be that we want to achieve too much in the same wrapper step definition? Might be that we need to write a bit more in our scenarios to clarify it? I'm asking my self, not sure, but when we use that much numbers of steps inside a step, having even some extra logic between them, maybe is because we are trying to wrap to much.

Bischoff:
> It depends on the conditions, as it picks the events FIFO, if we are running in parallel a dozen different events on different systems, it might indeed take more time. As we experienced in QAM testsuite during the smoke tests, in concrete when running several hardware refresh events at same time.

This happens on production test suite, where we have no parallélism, and we try to bootstrap only one minion on a quite powerful hypervisor.

The fact that the issue is a) reproducible b) new lets me think the bootstrap mechanism has degraded its performance recently.

It's indeed even worse in QAM context where we have parallelism.


> Now I wonder if taskomatic support some kind of priority levels for each kind of event.

Maybe. But I suspect the issue is rather on Salt side. Salt is a catastrophe performance-wise.

srbarrios:
> This happens on production test suite, where we have no parallélism, and we try to bootstrap only one minion on a quite powerful hypervisor.

What is the name of the event that gets stuck that much time? Might it be applying states when bootstrapping? 
If it's this case, we can also consider that the variable being the bottleneck is not the hypervisor or the VM hardware, but most a network condition, in case we have some network connections implied on those states.

srbarrios (approved):
In any case, I think we can approve this PR, and we can take into consideration my comment about splitting the step in a further moment or just forget it ;)

testsuite/features/step_definitions/common_steps.rb
srbarrios: ok

Bischoff:
> What is the name of the event that gets stuck that much time?

`Hardware List Refresh`.

> Might it be applying states when bootstrapping?

Yes, that's exactly that. This very first event stays too long in "Pending" tab.

> If it's this case, we can also consider that the variable being the bottleneck is not the hypervisor or the VM hardware, but most
> a network condition between the server and the node.

Maybe. But I think it's salt the guilty part here.

I explain that the error appeared recently by the fact they have been complexifying the bootstrap process at salt level.

> In my opinion, there are two mysteries to solve:
> 
> * Understand if the event's state on the WebUI (pending, picked up, completed..) reflects the real state of this event

Good point. I did not investigate that... :-(

>   * Might it be shown as pending and being in progress, or picked up and be completed.

It would be shown as pending and being in progress then.

> * If it gets stuck during a concrete state

Highest probability on this one...

>   * Find the reason why.

Yup :-) .

Bischoff:
> In any case, I think we can approve this PR, and we can take into consideration my comment about splitting the step in a further moment or just forget it ;)

Yes, we'll re-discuss the implementation details again if we decide to merge after all.

(I like the sovietic approach with simpler code. I just have a problem with the semantics of the step)

moio:
@Bischoff 4 minutes sounds like really long to me (on healthy minions and a not-loaded Server).

I'd suggest to try to break down where that time is spent. One way is via minion logs (not sure if logging level needs to be raised to see it) or looking at Salt events via `salt-run state.event pretty=True` on the Server. That should provide an indication of the "pure Salt" time. If that is the culprit, a Salt expert should be able to guide you further - and it can very well be a genuine bug. Hardware refresh should take seconds - maybe even 30 seconds in some corner cases - but definitely not minutes.

Otherwise, if the Salt execution is fast, it could be a problem on the Java side, for example Taskomatic taking too long to run the Action.

While we are at it we could tune [idleWaitTime](https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/salt/large-scale-tuning.html#org-quartz-scheduler-idleWaitTime) for testsuite - and actually any sumaform use: it will have Taskomatic check for new Actions more often than every 5 seconds. I guess those add up since we waste up to 5 seconds per single Action (manually or automatically scheduled!). That will certainly improve things a bit here - but does not add up to 4 minutes for onboarding alone, for sure.

srbarrios:
Even running on CI, where at that moment we don't have parallel tests enabled, it might be that a previous scenario let some other pending event.

The most important step it will be to reproduce this behaviour isolated.
As a reminder, we have a card to research ssh-minionw failing on hardware refresh when running simultanously in QAM tests.

Bischoff:
> (...)
> The most important step it will be to reproduce this behaviour isolated.
> (...)

Very good point. Especially that we always hit it in same scenario.

Bischoff:
> @Bischoff 4 minutes sounds like really long to me (on healthy minions and a not-loaded Server).
> (...)

Thanks for all the tips @moio.

> (...) Hardware refresh should take seconds - maybe even 30 seconds in some corner cases - but definitely not minutes.

No, it does not happen during the Hardware refresh - it takes time *before* the event is even picked up - it stays in "Pending".

moio:
Then we have a Java-land problem. To me personally it qualifies for a Bugzilla report

Bischoff:
> Then we have a Java-land problem. To me personally it qualifies for a Bugzilla report

OK. But let me try to reproduce locally as O%scar suggested (and as I should have tried in the first place anyway - my bad).

Bischoff:
Closing this for the time being. Needs more investigation.

moio:
> No, it does not happen during the Hardware refresh - it takes time _before_ the event is even picked up - it stays in "Pending".

Whoops, sorry, I got that completely wrong. In any case, it's a Javaland problem.

Bischoff:
I'm working on it under this card: #13627 .

PR #3088, opened 2021-01-07
Change behavior of "mgrcompat" wrapper after deprecation changes on Salt 3002

meaksh:
## What does this PR change?

This PR changes the behavior of `mgrcompat.module_run` wrapper to adapt it to the new deprecation policies released in Salt 3002.

In 3002, the `module.run` state syntax is again set as deprecated in "Phosphorus", so warning messages are printed. SaltStack decided to keep the old syntax (the initial syntax for module.run) still available under `module._run`.

Therefore, we don't need to tailor the old arguments passed to `mgrcompat.module_run`, but instead just pointing to the right function to avoid deprecation messages. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **internal changes**

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/13568

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mcalmer (commented):
But is the new/old syntax available in 3000, 3001, 3002? Do we need to handle it if a customer has it turned on?

meaksh:
@mcalmer the old syntax is still available in those versions, SaltStack decided to keep it in 3002 under `module._run`.

In our context, we want SUMA to use that `module._run` (which does not produce any deprecation warnings) regardless of customer configuration, since we don't want those messages to be printed, so this time, we call the old syntax function directly (`module._run`), instead of allowing the decorator from `module.run` to resolve the corresponding function.

mcalmer:
Hmm. The compat module expected the old syntax. We converted it to the new syntax in case the client was configured to expect the new syntax. Why should this not the case anymore when the client has salt 3000?

mcalmer (approved):

PR #3035, opened 2020-12-16
Testsuite: Add test for service pack migration

calancha:
## What does this PR change?
it adds a test for service pack migration on a SSH minion.

- testsuite/features/finishing/minssh_sp_migration.feature:
New test.

- testsuite/features/reposync/srv_sync_products.feature
- testsuite/features/step_definitions/common_steps.rb:
Add the product we want to migrate.
Add an activation key for it.

- testsuite/features/support/twopence_init.rb
- testsuite/features/support/constants.rb:
- testsuite/documentation/cucumber-steps.md:
Add label 'ssh_spack_migrated_minion'.

- testsuite/run_sets/finishing.yml:
List the new test at the end of the YAML file.

## Links
https://github.com/SUSE/spacewalk/issues/11471

### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13486

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

PR #3090, opened 2021-01-07
Testsuite: Add test for service pack migration

calancha:
Run the service pack migration for both, the minion and the
SSH managed minion at the init_clients stage.
This means we run the whole secondary features on
the migrated minions, i.e., on a newer service pack.

- testsuite/features/init_clients/minssh_sp_migration.feature
- testsuite/features/init_clients/min_sp_migration.feature:
New tests.
- testsuite/run_sets/init_clients.yml: Run the sp migration
  right after bootstrap the corresponding minion.

- testsuite/features/support/env.rb:
Add tag @service_pack_migration to enable/disable the migration.

- testsuite/features/reposync/srv_sync_products.feature
- testsuite/features/support/commonlib.rb
Ensure we add both products, the one before/after the migration.
Add activation keys to support the service pack migration: one
for the normal minion and one for the SSH minion.

- testsuite/features/support/twopence_init.rb
- testsuite/features/support/constants.rb:
- testsuite/documentation/cucumber-steps.md:
Add labels 'ssh_spack_migrated_minion', 'sle_spack_migrated_minion'.

## Links
https://github.com/SUSE/spacewalk/issues/11471
This PR is a reimplementation of https://github.com/uyuni-project/uyuni/pull/3035
### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13594
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/13601
 (at 4.0 only port the refactor of `compute_list_to_leave_running`)

## Changelogs

- [x] No changelog needed


testsuite/features/support/env.rb
```diff
@@ -23,8 +23,10 @@
 
 server = ENV['SERVER']
 $debug_mode = true if ENV['DEBUG']
-$long_tests_enabled = true if ENV['LONG_TESTS']
+$long_tests_enabled = !ENV['LONG_TESTS'].nil?
```
srbarrios, line 26: In my humble opinion is less readable in that format :P

testsuite/features/support/env.rb
```diff
@@ -319,6 +321,11 @@ def enable_assertions
   skip_this_scenario unless $long_tests_enabled
 end
 
+# do test only if we want to run service pack migration
+Before('@service_pack_migration') do
+  skip_this_scenario unless $service_pack_migration_enabled
+end
```
srbarrios, line 327: As far as I see the test strategy for the SP migration, we want to execute it always.
As we will deploy from the terraform file a SLES15SP1 version, and we want to test SLES15SP2 version for all our test suites features. So, I'd say is not optional to migrate to SP2.

testsuite/features/support/constants.rb
```diff
@@ -225,3 +228,75 @@
                       'ubuntu1804_minion' => 'x86_64',
                       'ubuntu2004_ssh_minion' => 'x86_64',
                       'ubuntu2004_minion' => 'x86_64' }.freeze
+
+CHANNEL_TO_SYNCH_BY_OS_VERSION = {
```
srbarrios, line 232: Very nice, I love it! Thanks :)

srbarrios (approved):
Some thoughts about the idea of making it optional. But anyhow LGTM

testsuite/features/support/commonlib.rb
```diff
@@ -69,41 +69,23 @@ def compute_image_name
...
   do_not_kill = []
-  if $long_tests_enabled
-    # keep the repos needed for the auto-installation tests
-    do_not_kill += CHANNEL_TO_SYNCH_BY_OS_VERSION['default']
+  [$minion, $build_host, $sshminion].each do |node|
+    next unless node
+    os_version, os_family = get_os_version(node)
+    next unless os_family == 'sles'
+    raise "Can't build list of reposyncs to leave running" unless ['12-SP4', '12-SP5', '15-SP1', '15-SP2', '15-SP3'].include? os_version
+    do_not_kill += CHANNEL_TO_SYNCH_BY_OS_VERSION[os_version]
```
srbarrios, line 80: Awesome

testsuite/features/support/constants.rb
calancha: Haha You forgot it! It was **you** suggestion in a previous PR, so I brought here ;-)

testsuite/features/support/env.rb
calancha: You are absolutely right! I am going to use the previous one.

testsuite/features/support/env.rb
calancha: Agree. Once merged, let's keep this tag for some CI runs (passing the env variable to the pipe in order to run the SP migration) for easy comparison with/without SP migration; if we don't observe breakages, then it has sense to drop this optional tag.

Bischoff:
For @srbarrios ' information, we discussed (Tino and I), and decided to do like this:

In `init-clients`, either run `sle_minion.feature` or `min_sp_migration.feature`, but not both. same for SSH minion. Remove the initial "unregister" too.

Goal is to accelerate `init_clients` stage.

We already do something similar for the proxy: we have 3 ways to register it, but only run one of them.

srbarrios:
> For @srbarrios ' information, we discussed (Tino and I), and decided to do like this:
> 
> In `init-clients`, either run `sle_minion.feature` or `min_sp_migration.feature`, but not both. Same for SSH minion. Remove the initial "unregister" too.
> 
> Goal is to accelerate `init_clients` stage.
> 
> We already do something similar for the proxy: we have 3 ways to register it, but only run one of them.

@Bischoff @calancha I think better not make it optional, we should always run the sp migration feature, otherwise, we will be testing the whole test suite with sles15sp1 instead of sles15sp2.

calancha:
> In `init-clients`, either run `sle_minion.feature` or `min_sp_migration.feature`, but not both. Same for SSH minion. Remove the initial "unregister" too.
> 
> Goal is to accelerate `init_clients` stage.

See the comment addressing this suggestion; I have tested it at head and 4.1.

testsuite/features/init_clients/min_sp_migration.feature
```diff
@@ -0,0 +1,68 @@
...
+  As an authorized user
+  I want to migrate from one service pack to the other on Salt minions
+
+  Scenario: Create the bootstrap repository for a service pack migrated Salt client
+     Given I am authorized
+     And I create the "x86_64" bootstrap repository for "sle_spack_migrated_minion" on the server
+
+  Scenario: Register this minion for service pack migration
+    Given I am authorized
+    And I go to the bootstrapping page
```
Bischoff, line 22: When

testsuite/features/init_clients/min_sp_migration.feature
```diff
@@ -0,0 +1,68 @@
...
+    Then I should see a "Bootstrap Minions" text
+    And I enter the hostname of "sle_spack_migrated_minion" as "hostname"
```
Bischoff, line 24: When

testsuite/features/init_clients/minssh_sp_migration.feature
```diff
@@ -0,0 +1,67 @@
...
+@scope_sp_migration
+@service_pack_migration
+Feature: Service pack migration for SSH minion
+  In order to update my systems
+  As an authorized user
+  I want to migrate from one service pack to the other on SSH managed minions
+
+  Scenario: Register this SSH minion for service pack migration
+    Given I am authorized
+    And I go to the bootstrapping page
```
Bischoff, line 20: When

testsuite/features/init_clients/minssh_sp_migration.feature
```diff
@@ -0,0 +1,67 @@
...
+    Then I should see a "Bootstrap Minions" text
+    And I check "manageWithSSH"
```
Bischoff, line 22: When

testsuite/features/support/commonlib.rb
```diff
@@ -68,80 +68,21 @@ def compute_image_name
 # This function computes a list of reposyncs to avoid killing, because they might be involved in bootstrapping.
 #
 # This is a safety net only, the best thing to do is to not start the reposync at all.
-# rubocop:disable Metrics/MethodLength
-# rubocop:disable Metrics/BlockLength
 def compute_list_to_leave_running
   do_not_kill = []
-  if $long_tests_enabled
```
Bischoff, line 75: No. The products for auto-installation tests should not disappear.

We need to guarantee that these products are added to the whitelist when we are testing auto-installation and that they are not when we don't.
￼
It is exactly same situation as for the service pack migration

The argument "we are going to run the service pack migration by default" is invalid, the test `  if $service_pack_migration_enabled` tells exactly the contrary.

Don't worry about putting duplicate packages in the list, the list ist reduced at the end anyway thanks to `.uniq`.

Bischoff (changes requested):
A few nitpicks, but also a serious alert.

Bischoff:
> @Bischoff @calancha I think better not make it optional, we should always run the sp migration feature, otherwise, we will be testing the whole test suite with sles15sp1 instead of sles15sp2.

@srbarrios It's possible we want to see how the testsuite reacts when using a "normal" minion  instead of a migrated minion. So there is a real use case for making it optional.

Also, I think it perfectly matches your "focus areas" strategy. What if someone just wants a quick test and is not interested into this service pack migration?

A valid roadmap would be to have these tags:
* `@scope_servicepack_migration`     (instead of current `@servicepack_migration`)
* `@scope_build_osimage` (instead of current `@long_tests`)
* `@scope_autoinstall` (instead of current `@long_tests` too)

What do you think?

testsuite/features/support/commonlib.rb
srbarrios: I don't understand your message :)
I think nothing has been disappeared just well refactored :)

srbarrios:
> > @Bischoff @calancha I think better not make it optional, we should always run the sp migration feature, otherwise, we will be testing the whole test suite with sles15sp1 instead of sles15sp2.
> 
> @srbarrios It's possible we want to see how the testsuite reacts when using a "normal" minion instead of a migrated minion. So there is a real use case for making it optional.
> 
> Also, I think it perfectly matches your "focus areas" strategy. What if someone just wants a quick test and is not interested into this service pack migration?
> 
> A valid roadmap would be to have these tags:
> 
> * `@scope_servicepack_migration`     (instead of current `@servicepack_migration`)
> * `@scope_build_osimage` (instead of current `@long_tests`)
> * `@scope_autoinstall` (instead of current `@long_tests` too)
> 
> What do you think?

I still think that if we don't pass the sp_migration feature, we will test the full testsuite with sles15sp1 instead of sles15sp2.
And it's hard to design a mechanism where we deploy sles15sp2 instead of sles15sp1 when a different .feature test is enabled.

Let's have an agreement, let's merge as it is, so we have the same approach in the .yaml file that we have on proxy.
BUT, let's add a warning, advising just after the sp_migration feature on the .yaml file, that if we don't run it, the rest of the test suite will run using sles15sp1.

Bischoff:
> I still think that if we don't pass the sp_migration feature, we will test the full testsuite with sles15sp1 instead of sles15sp2.
> And it's hard to design a mechanism where we deploy sles15sp2 instead of sles15sp1 when a different .feature test is enabled.

Very true, but I think it's an acceptable problem.

> Let's have an agreement, let's merge as it is, so we have the same approach in the .yaml file that we have on proxy.
> BUT, let's add a warning, advising just after the sp_migration feature on the .yaml file, that if we don't run it, the rest of the test suite will run using sles15sp1.

Agree. Just like we do with the proxy.

For the moment, the old minion bootstrap features are commented out in the yaml file (to keep the complexity of this PR low). But on the long run we could uncomment them and proceed according to the tag to provide a real choice between "direct bootstrap" and "sp migration bootstrap".

testsuite/features/support/commonlib.rb
calancha: OK, I see Eric's point; they are needed only at HEAD because we run such a test only at that branch.
I have just pushed a commit to reenable them into `do_not_kill`

testsuite/features/init_clients/min_sp_migration.feature
calancha: Thank you. Done.

Bischoff (approved):

Bischoff:
@calancha I approved, but if i may: the name of the array of channels to whitelist if we are testing Cedric's autoinstallation should bear a reference to this autoinstallation focus area, so we know it's in that case only.

calancha:
> @calancha I approved, but if i may: the name of the array of channels to whitelist if we are testing Cedric's autoinstallation should bear a reference to this autoinstallation focus area, so we know it's in that case only.

Addressed with a comment at the entry of the array; I mention the purpose of such an entry.

PR #3093, opened 2021-01-07
Provide Custom Info as Pillar data

nadvornik:
## What does this PR change?

Provide Custom Info as Pillar data

## GUI diff

No difference.

## Documentation

- TBD
- [ ] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/12016

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/systems/sdc/UpdateCustomDataAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/systems/sdc/UpdateCustomDataAction.java
@@ -134,6 +134,11 @@ public class UpdateCustomDataAction extends RhnAction {
             cdv.setModified(new Date());
             cdv.setLastModifier(user);
             server.addCustomDataValue(cdv);
+            server.asMinionServer().ifPresent(minion -> {
+                MinionPillarManager.INSTANCE.generatePillar(minion, false,
+                    MinionPillarManager.PillarSubset.CUSTOM_INFO);
+            });
+
             request.setAttribute(VAL_PARAM, form.get(VAL_PARAM));
             return getStrutsDelegate().forwardParams(mapping.findForward("updated"),
                     params);
--- a/java/code/src/com/redhat/rhn/frontend/action/systems/sdc/UpdateCustomDataAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/systems/sdc/UpdateCustomDataAction.java
@@ -27,6 +27,7 @@ import com.redhat.rhn.frontend.struts.RhnHelper;
 import com.redhat.rhn.frontend.struts.RhnValidationHelper;
 import com.redhat.rhn.frontend.struts.StrutsDelegate;
 import com.redhat.rhn.manager.system.SystemManager;
+import com.suse.manager.webui.services.pillar.MinionPillarManager;
 
 import org.apache.struts.action.ActionForm;
 import org.apache.struts.action.ActionForward;
@@ -134,6 +135,9 @@ public class UpdateCustomDataAction extends RhnAction {
             cdv.setCreator(user);
             cdv.setLastModifier(user);
             request.setAttribute(VAL_PARAM, form.get(VAL_PARAM));
+            server.asMinionServer().ifPresent(minion -> {
+                MinionPillarManager.INSTANCE.generatePillar(minion);
+            });
             return getStrutsDelegate().forwardParams(mapping.findForward("updated"),
                     params);
         }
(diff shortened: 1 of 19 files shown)


java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -2075,6 +2076,10 @@ public int setCustomValues(User loggedInUser, Integer sid, Map<String, String> v
             }
         }
 
+        server.asMinionServer().ifPresent(minion -> {
+            MinionPillarManager.INSTANCE.generatePillar(minion);
```
moio, line 2080: I know it's a bit more work, but I'd encourage creating an ad-hoc `MinionPillarGenerator` for this use case, or any change to custom info will entail the full regeneration of all pillar data, which can take a bit of time.

java/code/src/com/redhat/rhn/frontend/action/systems/sdc/UpdateCustomDataAction.java
```diff
@@ -134,6 +135,9 @@ public ActionForward execute(ActionMapping mapping,
             cdv.setCreator(user);
             cdv.setLastModifier(user);
             request.setAttribute(VAL_PARAM, form.get(VAL_PARAM));
+            server.asMinionServer().ifPresent(minion -> {
+                MinionPillarManager.INSTANCE.generatePillar(minion);
+            });
```
moio, line 140: I am a bit confused by the logic from lines above this. It seems to me that `addCustomDataValue` already creates a new `CustomDataValue` instance and sets it if null, and now we have the logic repeated here.

I would not mind in the context of this PR normally, but here we are adding code which is clearly back-end (pillar generation) to a front-end class (an Action), and was wondering where it would be placed best.

Would it be possible to fold the code above and the pillar generation inside of `addCustomDataValue` - either to be left in `Server` or moved into `ServerFactory`?

moio (commented):
Thanks for the efforts here, I am being a bit nitpicky but I think we are already on a good path!

Just not to forget, once the code is done please also open an issue or PR to our docs, this is a small but nice piece of news to share for the scripters out there.

Thanks, it's been a long time since I hoped this will one day be done!

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
nadvornik: Do you mean adding new pillar file or calling only MinionGeneralPillarGenerator?
Anyway, if I understand correctly, the write should go via MinionPillarManager.INSTANCE, right?
Is it OK to change MinionPillarManager to use named variables instead of list to hold MinionPillarFileManager instances and add MinionPillarManager.generatePillar with arguments to use only selected MinionPillarFileManager ?

moio: > Do you mean adding new pillar file or calling only MinionGeneralPillarGenerator?

I meant adding a new pillar file and related Generator class. Actually I made a copypaste mistake in that comment - meaning `MinionPillarGenerator` but actually writing `MinionGeneralPillarGenerator`, sorry for having confused you.

> Anyway, if I understand correctly, the write should go via MinionPillarManager.INSTANCE, right?

Only if you want to regenerate all pillar data, which IIUC should not be the case here.

> Is it OK to change MinionPillarManager to use named variables instead of list to hold MinionPillarFileManager instances and add MinionPillarManager.generatePillar with arguments to use only selected MinionPillarFileManager ?

It's not a bad idea FMPOV, the only thing I see is that, so far, the pattern has been to just instantiate a `new MinionPillarFileManager(new MySpecificPillarGenerator())` and call `updatePillarFile()`.

So I would welcome such change, at the condition it comes in its own separate commit (or even PR if you so prefer) and that all similar cases are changed in one shot (I see three of them apart from tests, see callers of `MinionPillarGenerator.generatePillarData()` as displayed by IntelliJ's `Navigate -> Call Hierarchy`).

image

nadvornik: OK, I am working on it.
For completeness, ad-hoc MinionPillarFileManager instances do not work with unit tests which change pillar root.

moio: One more reason to go ahead with such a refactoring!

nadvornik:
Updated as requested.

PR #3094, opened 2021-01-07
Truncate author name in the changelog

jvanz:
## What does this PR change?

The table in the database to store the changelog has a varchar column of
128 to store the author name. Thus, the sync fails if, for some reason,
the author name has more chars. This commits changes the code to truncate
the author name in 128 chars.

Before:

Changelog author was not truncate. Allowing name with more than 128 chars

After:

The changelog author is truncated in 128 chars

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13541

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

- [ ] Re-run test "javascript_lint" (Test skipped, there are no changes to test)

### DIFF
--- a/backend/spacewalk-backend.changes
+++ b/backend/spacewalk-backend.changes
@@ -1,6 +1,7 @@
 - Drop Transfer-Encoding header from proxy respone to fix error response messages (bsc#1176906)
 - Prevent tracebacks on missing mail configuration (bsc#1179990)
 - Fix pycurl.error handling in suseLib.py (bsc#1179990)
+- Truncate author name in the changelog (bsc#1180285)
 - harden extratag key import by execute_values to ignore conflicts
 - internal code cleanup (dropping unused table rhnErrataTmp)
 - Fix Debian package version comparison
--- a/backend/server/importlib/packageImport.py
+++ b/backend/server/importlib/packageImport.py
@@ -288,9 +288,10 @@ class PackageImport(ChannelPackageSubscription):
         unique_package_changelog_hash = {}
         unique_package_changelog = []
         for changelog in package['changelog']:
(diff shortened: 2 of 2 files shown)


parlt91 (approved):
LGTM
I tested if your changes actually fix the issue and can confirm that's the case.
Thank you very much for providing the fix. :smile: 
There is one thing to mention:
This is our upstream repository.
If you want to create a PTF for the customer you should base it on the product version that's affected in this case https://github.com/SUSE/spacewalk/tree/Manager-4.1
(there might be some changes between the two versions that can cause the ptf to fail to apply for the customer)
Please also open a PR with the very same changes for https://github.com/SUSE/spacewalk/tree/Manager-4.1 as well.
You can add me as a reviewer there and I will have another look as well.

parlt91:
There is one thing I forgot. Since this is a bug fix please also reference the id of the bug in the changelog entry.
Like this: `Truncate author name in the changelog (bsc#1180285)`
Please do the same for your Manager-4.1 port

backend/spacewalk-backend.changes
```diff
@@ -1,3 +1,4 @@
+- Truncate author name in the changelog
```
parlt91, line 1: See my comment about the bug id for the changelog

parlt91 (changes requested):

jvanz:
> bsc#1180285

Done!

jvanz:
> There is one thing I forgot. Since this is a bug fix please also reference the id of the bug in the changelog entry.
> Like this: `Truncate author name in the changelog (bsc#1180285)`
> Please do the same for your Manager-4.1 port

Done!

parlt91 (approved):
Thanks looks good now. Thanks for helping us out on this. :+1:

PR #3098, opened 2021-01-08
Open raw output in new tab for ScriptRunAction

parlt91:
## What does this PR change?

When clicking `view/download raw script output` in a remote command
execution event summary it links to the SSM Overview Remote command tab.
instead it should open a new tab with the raw output to view/download.

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: bugfix

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13567

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/script/ScriptRunAction.java
+++ b/java/code/src/com/redhat/rhn/domain/action/script/ScriptRunAction.java
@@ -64,7 +64,7 @@ public class ScriptRunAction extends ScriptAction {
                         .getReturnCode().toString()));
                 retval.append("</br>");
                 retval.append(ls.getMessage("system.event.scriptRawOutput"));
-                retval.append("<a href=\"" +
+                retval.append("<a data-senna-off=\"true\" target=\"_blank\" href=\"" +
                         DownloadManager.getScriptRawOutputDownloadPath(
                                 this.getId(), sr.getActionScriptId(), currentUser) +
                         "\">");
--- a/java/spacewalk-java.changes
+++ b/java/spacewalk-java.changes
@@ -1,3 +1,5 @@
+- Open raw output in new tab for ScriptRunAction (bsc#1180547)
+
 -------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


java/code/src/com/redhat/rhn/domain/action/script/ScriptRunAction.java
```diff
@@ -64,7 +64,7 @@ public String getHistoryDetails(Server server, User currentUser) {
                         .getReturnCode().toString()));
                 retval.append("</br>");
                 retval.append(ls.getMessage("system.event.scriptRawOutput"));
-                retval.append("<a href=\"" +
+                retval.append("<a target=\"_blank\" href=\"" +
```
ncounter, line 67: While I see this could work, and it could even be better in the end to open it into a different tab, I believe the bug is due to the single page application behavior. To properly fix the bug and make it work as it was before, we should disable the `spa` behavior with the `data-senna-off` attribute. But I personally even like better to open it into a different tab: in the end we are no longer into the product webUI, but we are displaying a **raw output** into the browser.
So, in the end, we should add them both the attributes: one to make sure the application does not try to behave as a `spa`, the other one to open it into a new tab.

ncounter (commented):
Take a look at my comment. In the end your way to fix it works, but I would take the chance to fix it from all the perspectives.
Thank you :+1:

java/code/src/com/redhat/rhn/domain/action/script/ScriptRunAction.java
parlt91: Makes sense!
I updated the PR with you suggestion please have another look and approve if you're satisfied :smile:

juliogonzalez:
Ok, this something nice to have, the fix is simple, and (last famous words) should not break anything, not even the testsuite.

Let's merge it.

Issue #2241, opened 2020-05-21
Refreshing repositories on .deb servers through Uyuni proxy very slow


phibid:
Uyuni release 2020.03

The "apt update" command  is really slow, around 2 minutes on a server registered with salt-ssh and going through a proxy server:

```
root@server:/etc/apt# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
root@server:/etc/apt# apt clean all
root@server:/etc/apt# time apt update
Ign:1 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-universe-uyuni InRelease
Ign:2 https://uyuni_proxy:443/rhn/manager/download ubuntu-18.04-pool-amd64-uyuni InRelease
Ign:3 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-main-security-uyuni InRelease
Ign:4 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-main-uyuni InRelease
Hit:5 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-universe-uyuni Release
Hit:6 https://uyuni_proxy:443/rhn/manager/download ubuntu-18.04-pool-amd64-uyuni Release
Hit:7 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-main-security-uyuni Release
Hit:8 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-main-uyuni Release
Ign:9 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-universe-uyuni Release.gpg
Ign:10 https://uyuni_proxy:443/rhn/manager/download ubuntu-18.04-pool-amd64-uyuni Release.gpg
Ign:11 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-main-security-uyuni Release.gpg
Ign:12 https://uyuni_proxy:443/rhn/manager/download ubuntu-1804-amd64-main-uyuni Release.gpg
Reading package lists... Done
Building dependency tree       
Reading state information... Done
1 package can be upgraded. Run 'apt list --upgradable' to see it.
[...]

PR #3104, opened 2021-01-10
Wsgi drop chuncked transfer encoding

mcalmer:
## What does this PR change?

When copying headers from parent requests in the proxy we can get also the `Transfer-Encoding` header field which may be set to `chunked`. Using this as input for WSGI to return a result to the client cause timeout issues as it does not handle correctly (at least not when setting it from external).
The proxy code remove this header already for successful responses, but in case of an error (404 Not Found) that header was still present.

This PR remove the Transfer Encoding header just before sending it back to the client.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/2241
Tracks https://github.com/SUSE/spacewalk/pull/13618

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/backend/spacewalk-backend.changes
+++ b/backend/spacewalk-backend.changes
@@ -1,3 +1,4 @@
+- drop Transfer-Encoding header from proxy respone to fix error response messages (bsc#1176906)
 - harden extratag key import by execute_values to ignore conflicts
 - internal code cleanup (dropping unused table rhnErrataTmp)
 - Fix Debian package version comparison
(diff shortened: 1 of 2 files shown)


mbologna:
@mcalmer this PR looks good. I wanted to provide a comment to simplify the logic a little bit, but in the end it was easier to open another PR based on your work: https://github.com/uyuni-project/uyuni/pull/3109

mbologna:
Closed in favor of https://github.com/uyuni-project/uyuni/pull/3109

PR #3107, opened 2021-01-11
Add Typescript compilation & type checking to Github Actions

Etheryte:
## What does this PR change?

Adds Typescript type checking and compilation to Github Actions so we have better confidence in the type safety of every PR.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Github Action was added

- [x] **DONE**

## Links

Fixes: https://github.com/SUSE/spacewalk/issues/13502

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/typescript-compilation.yml
+++ b/.github/workflows/typescript-compilation.yml
@@ -22,7 +22,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node-version }}
 
--- a/.github/workflows/typescript-compilation.yml
+++ b/.github/workflows/typescript-compilation.yml
@@ -1,0 +1,44 @@
+name: Typescript compilation
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/typescript-compilation.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    strategy:
+      matrix:
+        node-version: [10.x]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
```
renner, line 25: Not sure if it brings many advantages, but It seems there is a v2 available that could bring some performance benefits by caching node versions especially with self-hosted runners (IIUC): https://github.com/actions/setup-node#v2

renner (approved):
Looks really good to me! :+1: Only one comment added inline, otherwise feel free to merge.

.github/workflows/typescript-compilation.yml
Etheryte: Upgraded to use v2, thanks a lot for the suggestion!

PR #3115, opened 2021-01-12
Disable linting for Flow files

Etheryte:
## What does this PR change?

Fixing linting for Typescript files will break linting for Flow typed files since we have to use either one parser or another.  
Since work on TS is moving forward at a good pace, I propose this is a small price to pay and outlines a good path for what needs to be migrated first.  

In short, the idea is to ignore these files for now to get lint working consistently, then migrate those files, and then reenable lint on them again.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: no user facing changes

- [x] **DONE**

## Links

Related:
 - https://github.com/SUSE/spacewalk/issues/13208
 - https://github.com/SUSE/spacewalk/issues/13403
 - https://github.com/SUSE/spacewalk/issues/13515

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


susemanager-frontend/susemanager-nodejs-sdk-devel/package.json
```diff
@@ -37,7 +36,6 @@
     "eslint-config-react-app": "^5.2.1",
     "eslint-import-resolver-alias": "^1.1.2",
     "eslint-loader": "^3.0.3",
-    "eslint-plugin-flowtype-errors": "^4.4.0",
     "eslint-plugin-flowtype": "^5.2.0",
```
cbbayburt, line 39: Do we still need these other deps then?

Etheryte: The preset we use, `extends: ["react-app"]`, relies on `eslint-plugin-flowtype` being available, we just manually don't use it. Hopefully I can clean this out when I get around to https://github.com/SUSE/spacewalk/issues/13515.  

For now, removing it will make ESLint throw, but eventually the plan is to remove this too, yes.

web/html/src/.eslintignore
```diff
@@ -4,4 +4,44 @@ build/**/*
 build.js
 flow-typed/**/*
 global.d.ts
-imports.d.ts
\ No newline at end of file
+imports.d.ts
+
+# These files need Flow to TS migration before they're lintable again
```
ncounter, line 9: Do we need to declare them one by one, instead of grouping and migrating folder by folder?

Etheryte: We could ignore whole folders, but that would be too broad. For example, we don't have issues with everything in `components/input`, only `InputBase.js`. Same for other folders. I'd prefer to disable linting for as few files as possible, plus this cuts the list of must-migrate-asap files down considerably.

ncounter: Fine by me then :wink:

cbbayburt (approved):

PR #3120, opened 2021-01-13
Add Javascript lint to Github Actions

Etheryte:
## What does this PR change?

Add Javascript lint to Github Actions, tested on my own fork and works as expected.

Once this gets merged, it would make sense to remove the corresponding job from Jenkins for Uyuni, I'm not familiar with the process for that.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: linter integration.

- [x] **DONE**

## Links

Fixes: https://github.com/SUSE/spacewalk/issues/13403

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/javascript-lint.yml
+++ b/.github/workflows/javascript-lint.yml
@@ -5,10 +5,12 @@ on:
     branches:
       - master
     paths:
-      - 'web/html/src/**'
+      - 'web/html/src/**/*jsx?'
+      - 'web/html/src/**/*tsx?'
   pull_request:
     paths:
-      - 'web/html/src/**'
+      - 'web/html/src/**/*jsx?'
+      - 'web/html/src/**/*tsx?'
 
 jobs:
   javascript_lint:
--- a/.github/workflows/javascript-lint.yml
+++ b/.github/workflows/javascript-lint.yml
@@ -1,0 +1,44 @@
+name: Javascript lint
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'web/html/src/**'
+  pull_request:
+    paths:
+      - 'web/html/src/**'
+
+jobs:
+  javascript_lint:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x]
+
+    steps:
+    - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


.github/workflows/javascript-lint.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'web/html/src/**'
+  pull_request:
+    paths:
+      - 'web/html/src/**'
```
renner, line 11: Do we want to run the linting also when non-javascript files are changed within this folder (it seems there is at least some *.css, *.md, *.json files in there)? Otherwise I would recommend to restrict it here like this: `'web/html/src/**.js'`

.github/workflows/javascript-lint.yml
```diff
@@ -0,0 +1,44 @@
+name: Javascript lint
+
+on:
+  push:
+    branches:
+      - master
```
renner, line 6: This will need to be changed for the downstream version by the way and adapted to our release branches, also for other workflows. There is no master branch on the downstream repo.

renner (approved):
Looks good, only one suggestion about restricting linting to changes on specific files was added for your consideration before merge.

.github/workflows/javascript-lint.yml
Etheryte: Thanks for pointing this out, that's a good idea. I've added matchers for all of `.js`, `.jsx`, `.ts` and `.tsx`, I hope I got the syntax correct.

.github/workflows/javascript-lint.yml
Etheryte: That's a good point, I didn't consider this. What branches do we want to target for this? We could use [the syntax outlined here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags) to match specific ones.

Etheryte: As discussed on chat, I'll target 4.1, and also 4.0 if possible.

PR #3123, opened 2021-01-14
Fix CVE audit results for affected and patched entries (bsc#1180893)

cbbayburt:
The specific CVE audit query list_systems_by_patch_status merges server package data from both affected and non-affected systems. If the system is not affected, all the package related data (including EVR) is filled with nulls. But since epoch is a non-mandatory field, it can be null in both cases.

With the recent update to PackageEvr, the code assumes that an epoch is always present in an EVR, and ignores the rest of it if it is not. This causes an NPE down the line.

This fix checks the version to determine if we have an EVR instead.

Fixes: https://bugzilla.suse.com/1180893
Port of: https://github.com/SUSE/spacewalk/pull/13649

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/audit/CVEAuditManager.java
+++ b/java/code/src/com/redhat/rhn/manager/audit/CVEAuditManager.java
@@ -775,13 +775,13 @@ public class CVEAuditManager {
 
         return results.stream()
                 .map(row -> {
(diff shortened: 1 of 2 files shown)


mbologna:
Closing in favor of https://github.com/uyuni-project/uyuni/pull/3126 which is based on this PR but adds tests.

PR #3112, opened 2021-01-11
Migrate core to Typescript

Etheryte:
## What does this PR change?

~~Followup to https://github.com/uyuni-project/uyuni/pull/3106, no need to review this until that gets merged.~~

Migrate all code in `web/html/src/core` to Typescript.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes

- [x] **DONE**

## Test coverage
- No tests: Needs to pass all existing tests.

- [x] **DONE**

## Links

Related: https://github.com/SUSE/spacewalk/issues/13208

- [x] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #3124, opened 2021-01-14
Add Typescript support to Jest tests

Etheryte:
## What does this PR change?

~~Blocked by https://github.com/uyuni-project/uyuni/pull/3106 and https://github.com/uyuni-project/uyuni/pull/3112, otherwise should be fine.~~  

Adds Typescript support to our unit testing setup so we tests can benefit from type checking as well.  
Updated build dependencies and configurations where necessary.  
Migrated tests for code that are already in Typescript, the other tests can be migrated along with their respective main code.  

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: Upgrades testing environment.

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13610

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


ncounter (approved):

web/html/src/utils/data-providers/page-control.ts
```diff
@@ -6,14 +6,14 @@ type SortType = {
 export default class PageControl {
   page: number;
   pageSize: number;
-  query: string | null | undefined;
-  sort: SortType | null | undefined;
+  query?: string | null;
+  sort?: SortType | null;
```
cbbayburt, line 10: so is this equivalent in functionality?

cbbayburt (approved):

web/html/src/utils/data-providers/page-control.ts
Etheryte: @cbbayburt It's not quite equivalent, but in the above case that's how it was used. As for the difference, `arg: T | undefined` means the arg has to be specified, even if `undefined`:

```ts
function foo(arg: string | undefined) {...}
foo("bar"); // valid
foo(undefined); // valid
foo(); // NOT valid
```

Making the arg optional means that both `undefined` and omitting the arg is fine:  

```ts
function foo(arg?: string) {...}
foo("bar"); // valid
foo(undefined); // valid
foo(); // valid
```

cbbayburt: ok so TS will know the difference. Thanks :+1:

PR #3126, opened 2021-01-14
Fix: CVE audit results for affected and patched entries (bsc#1180893)

mbologna:
## What does this PR change?

The specific CVE audit query `list_systems_by_patch_status` and `list_images_by_patch_status` merges server package data from both affected and non-affected systems. If the system is not affected, all the package related data (including EVR) is filled with nulls. But since epoch is a non-mandatory field, it can be null in both cases.

With the recent update to `PackageEvr`, the code assumes that an epoch is always present in an EVR, and ignores the rest of it if it is not. This causes an NPE down the line.

This fix checks the version to determine if we have an EVR instead.

Fixes: https://bugzilla.suse.com/1180893

## Documentation

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Fixes: bugzilla.suse.com/1180893

To be ported to Manager-4.1, Manager-4.0

- [ ] **DONE**

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/audit/CVEAuditManager.java
+++ b/java/code/src/com/redhat/rhn/manager/audit/CVEAuditManager.java
@@ -775,13 +775,15 @@ public class CVEAuditManager {
 
         return results.stream()
                 .map(row -> {
-                    Optional<PackageEvr> packageEvr =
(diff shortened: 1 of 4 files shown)


cbbayburt (approved):
LGTM :+1: Thanks, @mbologna 
Please keep in mind that the tests need to be backported to `Manager-4.1`

PR #3142, opened 2021-01-18
Fix: login gets an ISE when SSO is enabled (bsc#1181048)

mbologna:
## What does this PR change?

Some variables needed for rendering the Jade login template were added after the introduction of SSO (e.g. `webTheme`).
Back in the day, the needed variables for SSO were only in `SparkApplicationHelper`. Now, login with SSO gets an ISE because the
Jade compiler cannot find model variables that were added.

This PR unifies the code path for non-SSO and SSO fixes the problem above while also unifies the code paths for non-SSO and SSO to
avoid the problem in the future.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1181048

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13680
Tracks https://github.com/SUSE/spacewalk/pull/13688

- [X] **DONE**

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/SSOController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/SSOController.java
@@ -121,7 +121,7 @@ public final class SSOController {
                         }
                     }
                     else {
-                        LOG.error("SAML attribute named 'uid' not found in SAML attributes. Cannot log the user in." +
+                        LOG.error("SAML attribute named 'uid' not found in SAML attributes. Cannot log the user in. " +
                                 "Please check with your Identity Provider (IdP) the presence of this attribute.");
                     }
                 }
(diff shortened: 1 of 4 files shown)


ncounter (approved):
LGTM, thanks for the quick fix. In the end, IIUC the problem was in the `return new ModelAndView(new HashMap<>,.....)` that you removed, it was returning an empty model we need instead to fulfill it with `Config` variables, do I get it right?
:+1:

ncounter:
@mbologna don't forget to backport to 4.1, thank you.

mbologna:
> LGTM, thanks for the quick fix. In the end, IIUC the problem was in the `return new ModelAndView(new HashMap<>,.....)` that you removed, it was returning an empty model we need instead to fulfill it with `Config` variables, do I get it right?
> +1

Yes, I unified the model population - the model is still rendered even if SSO is enabled and we get a redirect. That was the root cause of the ISE.

mbologna:
> @mbologna don't forget to backport to 4.1, thank you.

You mean 4.0? You are right.

ncounter:
> > @mbologna don't forget to backport to 4.1, thank you.
> 
> Do you mean 4.0? In 4.0 we do not have SSO so this problem does not happen.

Yes in 4.0 it is not needed. I meant 4.1, but I didn't seen there was already a [PR in place](https://github.com/SUSE/spacewalk/pull/13688). Thanks.

PR #3143, opened 2021-01-18
Updated 'node' requirements.

sbluhm:
## What does this PR change?

Added build requirements for node.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: No changes

- [X] **DONE**

## Test coverage
- No tests: Tested during package build process.
Build of CentOS 8 verified. Leap 15.2 not due to test system issues.

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/spacewalk-web.changes
+++ b/web/spacewalk-web.changes
@@ -1,3 +1,5 @@
+- Improved node requirements in spec file.
+- Updated Source to Uyuni repo.
 - Fix source selection form in CLM project page
 - Default to preferred items per page in content lifecycle lists (bsc#1180558)
 - Fix sorting in content lifecycle projects and cluster tables (bsc#1180558)
--- a/web/spacewalk-web.spec
+++ b/web/spacewalk-web.spec
@@ -40,18 +40,20 @@ Group:          Applications/Internet
 Version:        4.2.4
 Release:        1%{?dist}
 Url:            https://github.com/uyuni-project/uyuni
-Source0:        https://github.com/spacewalkproject/spacewalk/archive/%{name}-%{version}.tar.gz
+Source0:        https://github.com/uyuni-project/uyuni/archive/%{name}-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildArch:      noarch
 Requires(pre):  uyuni-base-common
 BuildRequires:  uyuni-base-common
 BuildRequires:  perl(ExtUtils::MakeMaker)
 BuildRequires:  susemanager-nodejs-sdk-devel
+BuildRequires:  nodejs-packaging
 
 %if 0%{?suse_version}
 BuildRequires:  apache2
-BuildRequires:  nodejs-packaging
-
+BuildRequires:  nodejs10
+%else
+BuildRequires:  nodejs
 %endif
 
 %description


web/spacewalk-web.spec
```diff
@@ -40,18 +40,20 @@ Group:          Applications/Internet
...
 BuildRequires:  uyuni-base-common
 BuildRequires:  perl(ExtUtils::MakeMaker)
 BuildRequires:  susemanager-nodejs-sdk-devel
+BuildRequires:  nodejs-packaging
 
 %if 0%{?suse_version}
 BuildRequires:  apache2
-BuildRequires:  nodejs-packaging
-
+BuildRequires:  nodejs-common
```
AdamMajer, line 54: For TW, the current version of node is `nodejs-default` provided by nodejs-common sources. I will submit this update to SLE so this can be used in the future. For now, the only sure way is to BR: on nodejs10 (if you need 32bit) or nodejs14 (it's available in all supported non-LTSS codestreams)

Etheryte: @sbluhm I think this looks good with the above suggestion applied. I think that Node 10 is a good target for now since that's what our pipeline tests against as well, I'll do the update everywhere at once when we're ready to move forward. Could you please also fix the typo of `nose` → `node` in the changefile?

sbluhm: @Etheryte and @AdamMajer, I am unsure what to put. As far as I understand we should keep nodejs-common as it includes nose (purpose of this PR). It also can provide nodejs-default for TW or SLE in future. So I don't see a code change requirement. Or do you want me to put nodejs10/nodejs14 instead of nodejs-common?

I intentionally put "nose" in the change log as this is the command where CenOS is failing/is missing.

AdamMajer: It would be better to list specific nodejs version, like nodejs14 or nodejs10 or another. This can be changed later.

sbluhm:
Tested successful build of changes on CentOS 8 and Leap 15.2

Etheryte:
Looks good, thanks a lot for your effort and patience on helping clear this up.

Cc @juliogonzalez so you're aware of this as well.

web/spacewalk-web.spec
```diff
@@ -40,18 +40,20 @@ Group:          Applications/Internet
...
 BuildRequires:  susemanager-nodejs-sdk-devel
+BuildRequires:  nodejs-packaging
 
 %if 0%{?suse_version}
 BuildRequires:  apache2
-BuildRequires:  nodejs-packaging
-
+BuildRequires:  nodejs10
+%else
+BuildRequires:  nodejs
```
juliogonzalez, line 56: Just a comment from my side, shouldn't this be `BuildRequires:  nodejs = 10` or similar, so we also stick to a given nodejs version for CentOS?

Etheryte: If I recall correctly, we already previously tried this, either on master or on 4.1 and there was a problem with it, but I can't recall what the specific problem was. I can't find a paper trail right now.

web/spacewalk-web.changes
```diff
@@ -1,3 +1,5 @@
+- Improved nose requirements in spec file.
```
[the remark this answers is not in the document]
Etheryte: I'll take care of cleaning this up.

web/spacewalk-web.spec
sbluhm: I can certainly try it as it is now in an %else clause.

web/spacewalk-web.changes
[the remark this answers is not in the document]
Etheryte: Covered in https://github.com/uyuni-project/uyuni/pull/3159

web/spacewalk-web.spec
sbluhm: This would work out to 
BuildRequires:    nodejs >= 1:10, nodejs <= 1:11
This looks very RHEL8 specific. Not sure if we would really want it like that.
Either way, putting line 56 to the above builds on Leap and CentOS 8.

Etheryte: I found the old discussion about version numbers as well, from [our internal chat](https://chat.suse.de/channel/galaxy-naica?msg=6qcZnem8xkxjwG9jZ):  

> Our nodejs packages for not provide the simple version number.  
> ```
> nothing provides nodejs = 10 needed by nodejs-packaging
> (got version 8.17.0 provided by nodejs8)
> (got version 12.18.4 provided by nodejs12)
> (got version 14.15.0 provided by nodejs14)
> (got version 10.22.1 provided by nodejs10)```

With a link to https://build.opensuse.org/package/view_file/openSUSE:Leap:15.2:Update/nodejs10/nodejs10.spec?expand=1 line 284

sbluhm: This might be because you put `nodejs = 10`. That won't work (as far as I understand. You have to put `>=` and `<`. So I would expect` Requires nodejs >= 10, nodejs < 11` to work (unless you call the version `1:10` and `1:11` like in the RHEL8 case...).

PR #3145, opened 2021-01-19
Add origin server hostname to Salt pillar data (bsc#1180439)

witekest:
## What does this PR change?

Origin server hostname is useful e.g. for setting up monitoring in environments
where Prometheus server is installed on the minion registered on the
proxy. The name _origin server_ always refers to the server where Uyuni server is installed.

Partially fixes https://bugzilla.suse.com/1180439

## GUI diff

No difference.

## Documentation
- [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pull/772)
- [Follow-up Documentation PR](https://github.com/uyuni-project/uyuni-docs/pull/797)

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes # SUSE/spacewalk#13543

- [x] **DONE**

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/utils/test/SSHMinionBootstrapperTest.java
+++ b/java/code/src/com/suse/manager/webui/controllers/utils/test/SSHMinionBootstrapperTest.java
@@ -82,6 +82,7 @@ public class SSHMinionBootstrapperTest extends AbstractMinionBootstrapperTestBas
     protected Map<String, Object> createPillarData(Optional<ActivationKey> key) {
         Map<String, Object> pillarData = new HashMap<>();
         pillarData.put("mgr_server", ConfigDefaults.get().getCobblerHost());
+        pillarData.put("mgr_origin_server", ConfigDefaults.get().getCobblerHost());
         pillarData.put("minion_id", "myhost");
         pillarData.put("contact_method", key
                 .map(k -> k.getContactMethod().getLabel())
--- a/java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java
+++ b/java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java
@@ -63,6 +63,7 @@ public class MinionGeneralPillarGenerator implements MinionPillarGenerator {
 
         pillar.add("contact_method", minion.getContactMethod().getLabel());
         pillar.add("mgr_server", minion.getChannelHost());
+        pillar.add("mgr_origin_server", ConfigDefaults.get().getCobblerHost());
         pillar.add("machine_password", MachinePasswordUtils.machinePassword(minion));
 
         Map<String, Object> chanPillar = new HashMap<>();
--- a/java/code/src/com/suse/manager/webui/controllers/utils/AbstractMinionBootstrapper.java
+++ b/java/code/src/com/suse/manager/webui/controllers/utils/AbstractMinionBootstrapper.java
@@ -220,6 +220,7 @@ public abstract class AbstractMinionBootstrapper {
                 .orElse(ConfigDefaults.get().getCobblerHost());
 
         pillarData.put("mgr_server", mgrServer);
+        pillarData.put("mgr_origin_server", ConfigDefaults.get().getCobblerHost());
         pillarData.put("minion_id", input.getHost());
         pillarData.put("contact_method", contactMethod);
         pillarData.put("mgr_sudo_user", SaltSSHService.getSSHUser());
--- a/java/code/src/com/suse/manager/webui/controllers/utils/test/RegularMinionBootstrapperTest.java
+++ b/java/code/src/com/suse/manager/webui/controllers/utils/test/RegularMinionBootstrapperTest.java
@@ -121,6 +121,7 @@ public class RegularMinionBootstrapperTest extends AbstractMinionBootstrapperTes
     protected Map<String, Object> createPillarData(Optional<ActivationKey> key) {
         Map<String, Object> pillarData = new HashMap<>();
         pillarData.put("mgr_server", ConfigDefaults.get().getCobblerHost());
+        pillarData.put("mgr_origin_server", ConfigDefaults.get().getCobblerHost());
         pillarData.put("contact_method", key
                 .map(k -> k.getContactMethod().getLabel())
                 .orElse(getDefaultContactMethod()));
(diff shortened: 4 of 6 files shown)


java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -2194,6 +2201,11 @@ public String getChannelHost() {
                 .orElseGet(() -> ConfigDefaults.get().getCobblerHost());
     }
 
+    public String getHubServerHost() {
+        return this.getLastServerPath().map(p -> p.getHostname())
+                .orElseGet(() -> ConfigDefaults.get().getCobblerHost());
+    }
```
mcalmer, line 2207: I think this is not correct. What happens when we do not have a Hub scenario?
I think we call getHubServerHost() and we get an answer. But this is just a server and not a Hub. Is this wanted?

And the javadoc is missing

java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java
```diff
@@ -63,6 +63,7 @@
 
         pillar.add("contact_method", minion.getContactMethod().getLabel());
         pillar.add("mgr_server", minion.getChannelHost());
+        pillar.add("hub_server", minion.getHubServerHost());
```
mcalmer, line 66: I think in most cases hub_server is set with a not hub server name.
I think this is not what we should do.

mcalmer (changes requested):

mcalmer:
I added @rjmateus as reviewer as well. I think somebody who is familar with "Hub" should review this as well.

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -490,6 +490,13 @@ public void setServerPaths(Set<ServerPath> serverPathsIn) {
                 .findFirst();
     }
 
+    public Optional<ServerPath> getLastServerPath() {
+        int lastPosition = serverPaths.size() - 1;
+        return serverPaths.stream()
+                .filter(path -> path.getPosition() == lastPosition)
+                .findFirst();
+    }
```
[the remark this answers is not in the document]
witekest: That's correct.
I've added java docs.

java/code/src/com/redhat/rhn/domain/server/Server.java
witekest: Javadoc added.

java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java
witekest: Currently we have the field `mgr_server` which points to the server or to the nearest proxy. I'd actually expect it to point always to the actual Uyuni server, as the name suggests, but that's not the case.
For configuring Prometheus targets I need information where the server runs. I've been thinking about the appropriate name for the new field and `hub_server` is the best which came to my mind.

java/code/src/com/redhat/rhn/domain/server/Server.java
witekest: The method returns the Hub server (last in the chain) or the server itself if no proxy is present.

moio (changes requested):
@witekest in the SUSE Manager context, Hub has a special meaning and that could mislead readers who are familiar with it.

In short, a Hub is a Server of Servers - and that is a separate concept than having a Server with Proxies (possibly chained, what we call a "server path", which is your intent here).

The information you really need is the canonical Server hostname, and it is set in the configuration variable `cobbler.host`, available via the `com.redhat.rhn.common.conf.ConfigDefaults#getCobblerHost` method. Indirectly the "last server path" is set to this value anyway, so I would encourage you not to consult the "server path" in the database at all and just return the Cobbler host.

A better name for the pillar could be for example `server_fqdn`.

PS. I hear you, that `getCobblerHost` method name is terrible and the idea of conflating the Cobbler hostname and the Server hostname is not good! I would be happy to see this occasion used to clean that up via a proper refactoring (but only in a separate PR after the user case is addressed).

PPS. I also hear you, `mgr_server` is also a bad pillar name - I would again be happy to see that cleaned up via a second refactoring PR.

Those are not requirements from my part, but well received nice-to-haves! And feel free to ping me if you need a hand.

mcalmer:
Other suggestions for a better pillar name:

- origin_server


PS: renaming "cobbler.host" and functions using it is a thing I see as possible and not so critical in terms of possible regressions.
But renaming the pillar `mgr_server` could lead into customers have to re-write there scripts. Pillar data are like public APIs. Changing them requires thinking about consequences, maybe with a deprecation phase and proper docs. Better not touch it :-)

witekest:
`server_fqdn` would be a good name only if `mgr_server` gets renamed. Otherwise they seem to be too similar and confusing.

How about `root_server`?

witekest:
> The information you really need is the canonical Server hostname, and it is set in the configuration variable `cobbler.host`, available via the `com.redhat.rhn.common.conf.ConfigDefaults#getCobblerHost` method. Indirectly the "last server path" is set to this value anyway, so I would encourage you not to consult the "server path" in the database at all and just return the Cobbler host.

Thanks for clarification. I'll implement that.

renner:
> `server_fqdn` would be a good name only if `mgr_server` gets renamed. Otherwise they seem to be too similar and confusing.
> 
> How about `root_server`?

I would have a slight preference for `server_fqdn` over `root_server` for the new pillar because it implies that the value is a FQDN. I also see though that `root_server` and `mgr_server` is a better pair naming wise, if we assume that we won't or cannot change the name of the `mgr_server` pillar.

Do I understand correctly that `mgr_server` contains either the proxy (IP or hostname?) in case of a proxy setup, or otherwise what is actually the difference between them? Trying to come up with a possibly better name for it.

java/code/src/com/suse/manager/webui/controllers/utils/test/SSHMinionBootstrapperTest.java
```diff
@@ -82,6 +82,7 @@ protected String getDefaultContactMethod() {
     protected Map<String, Object> createPillarData(Optional<ActivationKey> key) {
         Map<String, Object> pillarData = new HashMap<>();
         pillarData.put("mgr_server", ConfigDefaults.get().getCobblerHost());
+        pillarData.put("root_server", ConfigDefaults.get().getCobblerHost());
```
renner, line 85: Is the additional pillar data value really needed in those two cases where we set it to the exactly same value as `mgr_server`?

mcalmer: This seems to be a test. I do not think we should omit one if both are the same.

mcalmer (approved):

moio (approved):
This is OK FMPOV. If it's easy, it would be cool to check this in a unit test (probably I expect very similar ones exist already).

witekest:
_Merging is blocked_ for me. Is it intended?

mcalmer:
> _Merging is blocked_ for me. Is it intended?

yes, master is locked as we want to release uyuni and 4.2 Alpha 2. Julio wrote a mail

PR #3152, opened 2021-01-20
Add Javascript build to Github Actions

Etheryte:
## What does this PR change?

Add the frontend application build step to Github Actions, with this change we should be 100% on par with everything that used to run in Jenkins on PRs.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes: https://github.com/SUSE/spacewalk/issues/13703

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/javascript-build.yml
+++ b/.github/workflows/javascript-build.yml
@@ -1,0 +1,44 @@
+name: Javascript build
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'web/html/src/**'
+  pull_request:
+    paths:
+      - 'web/html/src/**'
+
+jobs:
+  javascript_build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x]
+
+    steps:
+    - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


renner (approved):
Nice, it looks good! :+1: As soon as the previous jenkins job is obsolete (sounds like it is the case after we merged this workflow) we could disable it to save some resources. Otherwise going forward we could start thinking about combining those workflows, but I'm currently not sure yet if it is desirable or if we would prefer to keep checks running separately (e.g. lint, build and test in one workflow)? In any case I would suggest to merge this PR first.

Etheryte:
I'm not really sure whether it's better to keep the actions separately or merge them, from my side there isn't a big difference since you can inspect the output easily. Perhaps @juliogonzalez has a preference, do we incur extra costs or something similar for extra jobs?  

I'm going to merge this job now, it will be easy to merge them as a followup if we want to.

juliogonzalez:
@Etheryte no idea, TBH I don't know much about GH actions in general, and much less about the billing. But IIRC @renner was having a look at that.

PR #3170, opened 2021-01-25
Allow setting system primary FQDN via WebUI or XMLRPC-API

ricardoasmarques:
## What does this PR change?

With this PR, it will now be possible to set/get the primary FQDN of a given system (via WebUI or XMLRPC-API).

**XMLRPC-API**
- The existing `system.getNetworkForSystems` method will now return a new `fqdn` field with the primary FQDN
- A new `system.setPrimaryFqdn` method has been added to set the primary FQDN of a given system

**WEBUI**
- The primary FQDN of a given systems can ve visualized/set via `System > Details > Hardware` page.

**Schema**
- The primary FQDN is stored in a new `is_primary` column on the `rhnServerFqdn` table

## GUI diff

Before:

Screenshot from 2021-01-25 15-12-12

After:

107793387-24f75e00-6d4e-11eb-9214-378c32f62231

- [x] **DONE**

## Documentation
- Documentation issue was created: https://github.com/SUSE/spacewalk/issues/13759

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13480

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/server/Server.java
+++ b/java/code/src/com/redhat/rhn/domain/server/Server.java
@@ -54,6 +54,7 @@ import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -943,6 +944,36 @@ public class Server extends BaseDomainHelper implements Identifiable {
             return ni.getIPv4Addresses().stream().findFirst().map(ServerNetAddress4::getAddress).orElse(null);
         }
         return null;
+    }
+
+    /**
+     * Return the ServerFQDN which is considered the primary.
+     *
+     * If primary FQDN is not explicitly set, the first one
+     * is returned, based on the alphabetic order.
+     *
+     * @return The primary FQDN for this server
+     */
+    public ServerFQDN findPrimaryFqdn() {
+        return this.fqdns.stream()
+                .filter(ServerFQDN::isPrimary)
+                .findFirst()
+                .orElseGet(() -> {
+                    if (!fqdns.isEmpty()) {
+                        return Collections.min(fqdns, Comparator.comparing(ServerFQDN::getName));
+                    }
+                    return null;
+                });
+    }
+
+    /**
+     * @param fqdnName The FQDN to be obtained
+     * @return The fqdn with the specified name
+     */
+    public Optional<ServerFQDN> lookupFqdn(String fqdnName) {
+        return this.fqdns.stream()
+                .filter((fqdn) -> fqdn.getName().equals(fqdnName))
+                .findFirst();
     }
 
     /**
@@ -1936,7 +1967,9 @@ public class Server extends BaseDomainHelper implements Identifiable {
             n.setPrimary(null);
         }
         SystemManager.storeServer(this);
-        primaryInterface.setPrimary("Y");
+        if (primaryInterface != null) {
+            primaryInterface.setPrimary("Y");
+        }
     }
 
     /**
@@ -1953,6 +1986,23 @@ public class Server extends BaseDomainHelper implements Identifiable {
             }
         }
         return null;
+    }
+
+    /**
+     * @param fqdnName name of the primary FQDN
+     */
+    public void setPrimaryFQDNWithName(String fqdnName) {
+        ServerFQDN newPrimaryFQDN = null;
+        for (ServerFQDN fqdn: fqdns) {
+            fqdn.setPrimary(false);
+            if (fqdn.getName().equals(fqdnName)) {
+                newPrimaryFQDN = fqdn;
+            }
+        }
+        SystemManager.storeServer(this);
+        if (newPrimaryFQDN != null) {
+            newPrimaryFQDN.setPrimary(true);
+        }
     }
 
     /**
--- a/java/code/src/com/redhat/rhn/domain/server/ServerFQDN.java
+++ b/java/code/src/com/redhat/rhn/domain/server/ServerFQDN.java
@@ -28,6 +28,7 @@ public class ServerFQDN extends BaseDomainHelper implements Identifiable {
     private Long id;
     private Server server;
     private String name;
+    private boolean primary;
 
     /**
      * Constructor for the class ServerFQDN
@@ -71,6 +72,20 @@ public class ServerFQDN extends BaseDomainHelper implements Identifiable {
      */
     public void setName(String nameIn) {
         this.name = nameIn;
+    }
+
+    /**
+     * @return Boolean which indicates primary FQDN
+     */
+    public boolean isPrimary() {
+        return primary;
+    }
+
+    /**
+     * @param primaryIn Boolean which sets primary FQDN
+     */
+    public void setPrimary(boolean primaryIn) {
+        primary = primaryIn;
     }
 
     /**
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/NoSuchFQDNException.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/NoSuchFQDNException.java
@@ -1,0 +1,34 @@
+/**
+ * Copyright (c) 2021 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package com.redhat.rhn.frontend.xmlrpc;
+
+import com.redhat.rhn.FaultException;
+import com.redhat.rhn.common.localization.LocalizationService;
+
+/**
+ * No Such FQDN Interface Exception
+ */
+public class NoSuchFQDNException extends FaultException  {
+
+    /**
+     * NoSuchFQDNException
+     * @param fqdn FQDN which has not been found
+     */
+    public NoSuchFQDNException(String fqdn) {
+        super(1074, "No Such FQDN" , LocalizationService.getInstance().
+                getMessage("api.system.nosuchfqdn", fqdn));
+    }
+}
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
@@ -77,6 +77,7 @@ import com.redhat.rhn.domain.server.NetworkInterface;
 import com.redhat.rhn.domain.server.Note;
 import com.redhat.rhn.domain.server.PushClient;
 import com.redhat.rhn.domain.server.Server;
+import com.redhat.rhn.domain.server.ServerFQDN;
 import com.redhat.rhn.domain.server.ServerFactory;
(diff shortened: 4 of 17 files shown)


moio:
@ricardoasmarques how would we define primary from a user perspective? What makes "primary" special compared to others?

moio:
@ncounter can you please take a look at the before/after screenshots above and provide a suggestion about the UX aspect? I personally find it a bit odd that two similar concepts (primary network interface and primary FQDN) are displayed differently (via a dropdown vs. radio buttons). What would be best and why, or does it make sense to keep them different?

ncounter:
> @ncounter can you please take a look at the before/after screenshots above and provide a suggestion about the UX aspect? I personally find it a bit odd that two similar concepts (primary network interface and primary FQDN) are displayed differently (via a dropdown vs. radio buttons). What would be best and why, or does it make sense to keep them different?

From my perspective we should align the two for having the same UX, if possible. IIUC the UX is exactly the same: the user picks one among the multiple options, and the picked one is the `primary`, no more than that. A couple of picky question to choose the right graphical object to use:
1- if the user is **not** looking for changing the `primary` value, are other values useful to be displayed, or they are just taking space because the user is not looking at them unless he really wants to change it? [taking space | have value]
2- how many options do we usually expect to have? [less than 5 | more than 5]

Answer to 1 is more important than 2 actually, anyway this is the guideline:
- use a `select` (AKA `dropdown`) if the main value we need to take care is the "chosen" one, and other options can be hided. Also, use a `select` if the list of multiple options is veeeery long, because when you don't need them they are collapsed.
- use a `radio` button multiple option if the list is short (otherwise it takes too much space and it confuses the user) and also if you still want to see all of the other possibilities once you are done with the selection. I.e. when I need to know all the properties, but I need to name one as primary but still, I want to see them all.

Wrapping up: from my POV a `select` is a good element in both of the cases here. In addition to that, we should also consider to rename the label of the `FQDNs` to `Primary FQDN`, because when the dropdown is collapsed, we will see only the primary selection, and that's the one the user wants to see in the end.

@moio : do all of my assumptions make any sense or are they wrong, from what the user wants perspective?

If for some reason we instead need to present all of the values, I'd rather go with a `radio` button approach, but with something different: a list of options where you can click on the value and the one you click gets a badge next to it saying `primary`, and the others have a "make this primary" clickable text.

@ricardoasmarques I'd still go for the `select` option, but if for some reason we need to fallback on this option, ping me again I will prepare a mock-up or something (a picture explains it better than 1000 words sometime :sweat_smile: )

ricardoasmarques:
@ncounter @moio Thanks for the feedback, I've now changed it to a dropdown and updated the label accordingly:

Screenshot from 2021-02-12 16-10-44

schema/spacewalk/upgrade/susemanager-schema-4.2.7-to-susemanager-schema-4.2.8/701-rhn_server_fqdn.sql
```diff
@@ -0,0 +1,6 @@
+
+ALTER TABLE rhnServerFQDN ADD COLUMN IF NOT EXISTS is_primary VARCHAR(1);
```
moio, line 2: Would it make sense to set one per server as the primary as part of the migration?

java/spacewalk-java.changes
```diff
@@ -1,3 +1,5 @@
+- Allow setting system primary FQDN via WebUI or XMLRPC-API
```
moio, line 1:
```suggestion
- Allow setting a primary FQDN per client, either via WebUI or XMLRPC-API
```

java/code/src/com/redhat/rhn/domain/server/ServerFQDN.java
```diff
@@ -73,6 +74,20 @@ public void setName(String nameIn) {
         this.name = nameIn;
     }
 
+    /**
+     * @return primary String which indicates primary interface
+     */
+    public String getPrimary() {
```
moio, line 80: Could we turn those accessors into `boolean` ones? I find the "YN" string confusing.

See how this is done, for example, in the PAYG attribute of Server:

https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/server/Server_legacyUser.hbm.xml#L35

https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/server/Server.java#L130

https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/server/Server.java#L2079-L2093

My understanding was that the "magic" `yes_no` Hibernate column type hides away all string handling - or at least that would be my hope :angel:

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -945,6 +947,26 @@ public String getIpAddress() {
...
+            return fqdn.getName();
+        }
+        return null;
+    }
+
+    /**
+     * @param fqdnName The FQDN to be checked
+     * @return Returns true if the FQDN exists, otherwise false
+     */
+    public boolean existsFqdn(String fqdnName) {
```
moio, line 966: Does this method have to be `public`?

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -1866,6 +1868,7 @@ public int deleteSystem(User loggedInUser, Integer serverId, String cleanupType)
             network.put("ip", StringUtils.defaultString(server.getIpAddress()));
             network.put("ip6", StringUtils.defaultString(server.getIp6Address()));
             network.put("hostname", StringUtils.defaultString(server.getHostname()));
+            network.put("fqdn", StringUtils.defaultString(server.getFqdn()));
```
moio, line 1871: Since we are adding a new attribute here, should we make clear this is the primary one?

java/code/src/com/redhat/rhn/frontend/xmlrpc/NoSuchFQDNException.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.redhat.rhn.frontend.xmlrpc;
+
+import com.redhat.rhn.FaultException;
+import com.redhat.rhn.common.localization.LocalizationService;
+
+/**
+ * No Such FQDN Interface Exception
+ *
+ * @version $Rev$
```
moio, line 24: These tags should be removed everywhere and definitely not appear in new code :sweat_smile:

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -6605,6 +6608,30 @@ public int setPrimaryInterface(User loggedInUser, Integer serverId,
...
+     * @xmlrpc.doc Sets new primary FQDN
+     * @xmlrpc.param #param("string", "sessionKey")
+     * @xmlrpc.param #param("int", "serverId")
+     * @xmlrpc.param #param("string", "fqdn")
+     * @xmlrpc.returntype #return_int_success()
+     */
+    public int setPrimaryFqdn(User loggedInUser, Integer serverId, String fqdn) {
+        Server server = lookupServer(loggedInUser, serverId);
+
+        if (!server.existsFqdn(fqdn)) {
```
moio, line 6628: Would it be more straightforward to have a finder method that returns an `Optional<ServerFQDN>` instance and set the flag there if it is found?

ricardoasmarques:
@moio I believe I've now addressed all your comments, can you please do another round of review? Thanks!

java/code/src/com/redhat/rhn/frontend/xmlrpc/NoSuchFQDNException.java
ricardoasmarques: See https://github.com/uyuni-project/uyuni/pull/3294

witekest (commented):
Tested with Uyuni SD. Works as expected. Two nits inline.
An implementation detail about storing server is not clear to me.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
ricardoasmarques: I've renamed it from `fqdn` to `primary_fqdn`.

java/code/src/com/redhat/rhn/frontend/xmlrpc/NoSuchFQDNException.java
```diff
@@ -0,0 +1,35 @@
...
+public class NoSuchFQDNException extends FaultException  {
+
+    /**
+     * NoSuchFQDNException
+     * @param fqdn FQDN which has not been found
+     */
+    public NoSuchFQDNException(String fqdn) {
+        super(1074, "No Such FQDN" , LocalizationService.getInstance().
+                getMessage("api.system.nosuchfqdn",
+                new Object [] {fqdn}));
```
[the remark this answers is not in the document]
ricardoasmarques: Done

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -945,6 +946,36 @@ public String getIpAddress() {
...
+     * Return the ServerFQDN which is considered the primary.
+     *
+     * If primary FQDN is not explicitly set, the first one
+     * is returned, based on the alphabetic order.
+     *
+     * @return The primary FQDN for this server
+     */
+    public ServerFQDN findPrimaryFqdn() {
+        return this.fqdns.stream()
+                .filter((fqdn) -> fqdn.isPrimary())
```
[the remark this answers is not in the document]
ricardoasmarques: Done

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -1955,6 +1988,23 @@ private NetworkInterface lookupForPrimaryInterface() {
...
+     */
+    public void setPrimaryFQDNWithName(String fqdnName) {
+        ServerFQDN newPrimaryFQDN = null;
+        for (ServerFQDN fqdn: fqdns) {
+            fqdn.setPrimary(false);
+            if (fqdn.getName().equals(fqdnName)) {
+                newPrimaryFQDN = fqdn;
+            }
+        }
+        SystemManager.storeServer(this);
```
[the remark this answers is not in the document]
ricardoasmarques: This `storeServer` call is needed to guarantee that we don't hit the DB unique constraint that prevents more than one primary FQDN for each server, but we don't have to call it again because DB will then be updated via SessionFilter:

https://github.com/uyuni-project/uyuni/blob/93a51e3edc32dbc7cbaccc53d8da5743aeba9176/java/code/src/com/redhat/rhn/frontend/servlets/SessionFilter.java

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -1853,6 +1855,7 @@ public int deleteSystem(User loggedInUser, Integer serverId, String cleanupType)
      *       #prop_desc("string", "ip", "IPv4 address of system")
      *       #prop_desc("string", "ip6", "IPv6 address of system")
      *       #prop_desc("string", "hostname", "Hostname of system")
+     *       #prop_desc("string", "fqdn", "FQDN of system")
```
witekest, line 1858:
```suggestion
     *       #prop_desc("string", "primary_fqdn", "Primary FQDN of system")
```

ricardoasmarques: Nice catch! Fixed, thanks.

witekest (approved):

moio (approved):
I think the changes are good to go, the only other requirement from my part is some minimal coverage in our Cucumber testsuite, as this touched the UI.

Thanks!

ricardoasmarques:
A minimal Cucumber test has been added.

PR #3171, opened 2021-01-25
Eliminate dead VHM API code

Bischoff:
## What does this PR change?

This PR eliminates dead code.

(a bit of cleanup piggybacked here too)

## Links

See also https://github.com/uyuni-project/sumaform/pull/840.

Ports:
* 4.0: SUSE/spacewalk#13764
* 4.1: SUSE/spacewalk#13765

## Changelogs

- [x] No changelog needed


lkotek (approved):
LGTM. Let's remove it.

calancha (approved):
If it is dead, let's bury it and rest in piece.

testsuite/features/step_definitions/security_steps.rb
```diff
@@ -11,14 +11,14 @@
...
                   "Header '#{arg1}' not present in '#{@url}'")
   assert_equal(arg2, @headers[arg1.downcase],
                "Header '#{arg1}' in '#{@url}' is not '#{arg2}'")
 end
 
-And(/^the response header "(.*?)" should not be "(.*?)"$/) do |arg1, arg2|
+Then(/^the response header "(.*?)" should not be "(.*?)"$/) do |arg1, arg2|
   refute_equal(arg2, @headers[arg1.downcase],
                "Header '#{arg1}' in '#{@url}' is '#{arg2}'")
 end
```
srbarrios, line 24: That's the kind of step that I would like to avoid.
I would prefer to hide this check, wrapping it inside the other step, instead of using an instance variable.

Also, I understand that is nicer to have When/Then/Given properly, but I feel more useful to have argument names that describe their purpose :)

srbarrios (approved):

testsuite/features/step_definitions/security_steps.rb
Bischoff: Hiding it: yes, definitely.

Argument names: absolutely.

Reason for not doing it: I was already piggybacking, I did not want to add to much to it :-) .

Also: this file is missing copyright notice and dates.

And while we are at it: we might have much more dead code around XML-RPC. I did not search in depth.

PR #3176, opened 2021-01-27
QAM: Implementing remote command for traditional client

srbarrios:
## What does this PR change?

This PR adds a remote command scenario for the traditional clients of a Build Validation test suite.
Also, it removes a deprecated rake task, as we renamed everything from qam to build_validation, as it's a more generic name that involves Maintenance Updates testing and new major/minor versions testing (which also use Maintenance Incidence mechanism in order to provide the un-released packages)

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/13778
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13777

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/Rakefile
+++ b/testsuite/Rakefile
@@ -84,16 +84,8 @@ namespace :utils do
     end
   end
 
-  desc 'Generate all the features need for Build Validation from templates'
+  desc 'Generate all the features we need for a Build Validation from templates'
   task :generate_build_validation_features do
-    Rake::Task['utils:generate_smoke_tests'].invoke
-    Rake::Task['utils:generate_smoke_tests'].reenable
-    Rake::Task['utils:generate_add_activation_keys_features'].invoke
-    Rake::Task['utils:generate_add_activation_keys_features'].reenable
-  end
-
-  desc 'Generate all the features need for QAM from templates'
-  task :generate_qam_features do
     Rake::Task['utils:generate_smoke_tests'].invoke
     Rake::Task['utils:generate_smoke_tests'].reenable
     Rake::Task['utils:generate_add_maintenance_update_repository_features'].invoke
--- a/testsuite/features/build_validation/smoke_tests/smoke_tests.template
+++ b/testsuite/features/build_validation/smoke_tests/smoke_tests.template
@@ -29,7 +29,7 @@ Feature: Smoke tests for <client>
     And I click on "Confirm"
     Then I should see a "1 patch update has been scheduled for" text
     When I force picking pending events on "<client>" if necessary
-    Then I wait until event "Patch Update:" is completed
+    And I wait until event "Patch Update:" is completed
 
   Scenario: Install a package on the <client>
     Given I am on the Systems overview page of this "<client>"
@@ -40,7 +40,7 @@ Feature: Smoke tests for <client>
     And I click on "Confirm"
     Then I should see a "1 package install has been scheduled for" text
     When I force picking pending events on "<client>" if necessary
-    Then I wait until event "Package Install/Upgrade scheduled by admin" is completed
+    And I wait until event "Package Install/Upgrade scheduled by admin" is completed
 
   Scenario: Remove package from <client>
     Given I am on the Systems overview page of this "<client>"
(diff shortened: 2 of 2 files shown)


testsuite/Rakefile
```diff
@@ -84,16 +84,8 @@ namespace :utils do
     end
   end
 
-  desc 'Generate all the features need for Build Validation from templates'
+  desc 'Generate all the features need for a Build Validation from templates'
```
ktsamis, line 87: nitpick: `features we need for a Build` or `features needed for a Build`

testsuite/features/build_validation/smoke_tests/smoke_tests.template
```diff
@@ -62,13 +62,25 @@ Feature: Smoke tests for <client>
...
-    When I wait until I do not see "pending" text
-    And I click on run
+    And I wait until I do not see "pending" text
+    When I click on run
     And I wait until I see "show response" text
     And I expand the results for "<client>"
     Then I should see "My remote command output" in the command output for "<client>"
 
-  #TODO: Run a remote command on traditional clients
+@skip_for_minion
```
ktsamis, line 71: Here we need to install the package that allows this (not installed by default mgr-cfg-actions) and run the command via ssh on the client first `mgr-actions-control --enable-all` which allows the UI to have the tab there. Then we can run this scenario.

ktsamis (changes requested):
Some modification needed. But otherwise this is needed, great work :+1:

testsuite/features/build_validation/smoke_tests/smoke_tests.template
srbarrios: This is already done in the initialization of each traditional client.

ktsamis: You are absolutely correct, sorry

ktsamis (approved):
Great let's merge

testsuite/features/build_validation/smoke_tests/smoke_tests.template
```diff
@@ -62,13 +62,25 @@ Feature: Smoke tests for <client>
     And I enter the hostname of "<client>" as "target"
     And I click on preview
     Then I should see a "Target systems (1)" text
-    When I wait until I do not see "pending" text
-    And I click on run
+    And I wait until I do not see "pending" text
```
Bischoff, line 65: When, this is an action

testsuite/features/build_validation/smoke_tests/smoke_tests.template
```diff
@@ -62,13 +62,25 @@ Feature: Smoke tests for <client>
...
+    When I click on run
```
Bischoff, line 66: And

Bischoff (approved):
LGTM, just a small nitpick

testsuite/features/build_validation/smoke_tests/smoke_tests.template
srbarrios: I changed because in natural language those "When I wait until something happen Then I expect this other thing" it's very weird to me, it sounds better to me "Then I wait until something happen and I expect this other thing" :man_shrugging: 
BUt ok, changing, not strong opinions with those Given/When/Then/And

Bischoff: @srbarrios "Then" in English has 2 meanings:
* chronology: I do this, then I do that
* cause-consequence: If I do this, then that happens

Here we use second meaning.

srbarrios: > * cause-consequence: If I do this, then that happens
> 

What we write in theses steps "I wait until xxx" is:

When "I do this until that happens"

And using "Then", it sounds better to me:

Then "I do this until that happens"

PR #3188, opened 2021-01-28
Channel subscription via SSM ignores the child channel

ricardoasmarques:
## What does this PR change?

Channel subscription via SSM was not sending the `childChannelActions` field in the server request due to the fact that `JSON.stringify` does not convert `Map`fields.

With this PR, channel subscription via SSM will no longer ignore child channels:

A complete list of steps to reproduce the issue can be found in https://bugzilla.suse.com/show_bug.cgi?id=1181520

## GUI diff

Before:
Screenshot from 2021-01-28 18-21-37

After:
Screenshot from 2021-01-28 18-21-27

- [x] **DONE**

## Documentation
- No documentation needed: This is just a bugfix, feature is already documented.

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13809

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/systems/ssm/ssm-subscribe-channels.js
+++ b/web/html/src/manager/systems/ssm/ssm-subscribe-channels.js
@@ -7,6 +7,7 @@ import ReactDOM from 'react-dom';
 import { AsyncButton, Button } from 'components/buttons';
 import { ActionSchedule } from 'components/action-schedule';
 import Network from 'utils/network';
+import { replacer } from 'utils/json';
 import { Utils, Formats } from 'utils/functions';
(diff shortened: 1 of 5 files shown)


juliogonzalez (approved):
I am not a Javascript expert at all, but as far as I can understand from the changes, this should work.

So approving and merging. If we are lucky, we'll see nice results today, and worst worst case, we can always rollback.

Thanks a lot for the fix!

PR #3189, opened 2021-01-29
Testsuite: Combine SP migration scenarios inside client registration

calancha:
## What does this PR change?
Service pack migration test is stable; let's simplify it. 

Instead of having 2 separated feature files, one w/ and another w/o
SP migration, combine them in just one file;  make use of
the tags @service_pack_migration and @skip_service_pack_migration
to select the proper scenarios.

Also, we don't need dedicated activation keys to bootstrap the SP
migrated minions; delete them.

- testsuite/features/reposync/srv_sync_products.feature
  (Add the initial product for the service pack migration):
  Drop @ssh_minion tag from this scenario.
  Delete the two scenarios creating sp migration keys.

- testsuite/features/init_clients/sle_minion.feature
- testsuite/features/init_clients/sle_ssh_minion.feature:
Absorb here the scenarios about SP migration.

- testsuite/features/init_clients/min_sp_migration.feature
- testsuite/features/init_clients/minssh_sp_migration.feature:
Delete these files.

testsuite/features/support/env.rb (@skip_service_pack_migration):
Rename from @skip_sp_migration and move its definition closer
to its partner, i.e.  @service_pack_migration.

- testsuite/run_sets/init_clients.yml:
Adjust with the features files we run; no more commented out features.

## Links
refactor of https://github.com/uyuni-project/uyuni/pull/3090
### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/13827
(This test doesn't exist at Manager-4.0 because channel synch is much slower).

## Changelogs

- [x] No changelog needed


testsuite/features/support/env.rb
```diff
@@ -332,6 +328,11 @@ def enable_assertions
   skip_this_scenario unless $service_pack_migration_enabled
 end
 
+# do test only if we don't want to run service pack migration
+Before('@skip_sp_migration') do
+  skip_this_scenario if $service_pack_migration_enabled
```
srbarrios, line 333: hmmm 🤔 the name is not coherent right?
If this variable is enabled then we DON'T want to skip it. If it's false, then we want to skip the scenario.

testsuite/features/init_clients/sle_ssh_minion.feature
```diff
@@ -22,6 +22,7 @@ Feature: Bootstrap a Salt host managed via salt-ssh
 # Package 'sle-manager-tools-release' is automatically installed during bootstrap and
 # stays installed after removal of channel containing it. So it is not possible to update it.
 # Package needs to be removed from highstate to avoid failure when updating it.
+@skip_sp_migration
```
srbarrios, line 25: What about naming it the same that the other tag, but adding "skip_".
So : "skip_service_pack_migration"
Or calling both: (skip_)sp_migration

srbarrios (approved):
Only a pair of small comments, approving in any case, as the changes seems ok.

testsuite/features/init_clients/sle_ssh_minion.feature
calancha: I have changed it to: `(skip_)service_pack_migration`

testsuite/features/support/env.rb
calancha: To me skip is very clear :-)

Bischoff (commented):
OK, but please note that for the 3 ways to create the proxy, we have 3 different files @calancha . I liked to have a common pattern here.

But we could also go in the reverse direction and systematize the one-file approach.

PR #3195, opened 2021-02-01
Increase default request timeout (and more...)

ricardoasmarques:
## What does this PR change?

**Problem**
When a request times out after 30 seconds, the SPA engine automatically reloads the page which may confuse the user because all "manual" changes done in the page will be reverted (e.g. if the user changes a dropdown value, that dropdown value will be "reset") - This problem is reported in https://bugzilla.suse.com/show_bug.cgi?id=1178767.

**Solution**
- This PR changes the default timeout from 30 seconds to 120 seconds.
- User can change this timeout value via `web.spa.timeout` config property.
- Additionally, an alert message will now be displayed to the user before automatically reloading the page.

## GUI diff

After:

Screenshot from 2021-02-01 13-56-16

- [x] **DONE**

## Documentation
- Documentation issue was created: https://github.com/SUSE/spacewalk/issues/13847

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13117

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


moio (approved):
Hey @ricardoasmarques, thanks for this contribution!

I think this is one of those edge cases that make life needlessly difficult in a lot of different use cases, and I like the way you nailed it down and proposed a very sensible solution.

:+1: 

I will leave it to @ncounter and/or @Etheryte to also give a final green light on the Javascript side of the solution.

Etheryte:
I agree that this definitely improves the situation, I think we can improve it a bit more though.  

The problem with using `alert()` is that the user can't do anything about it — their only option is to press OK and then watch the page reload. What I think we should do instead is show an error banner ([something like this](https://suse.eosdesignsystem.com/alerts/global)) that lets the user choose when to perform this action.  

Something a-la 

> Request timed out, please [button: reload the page]  

The upside of this solution is that it empowers the user: if the user wants to write down what changes they've made on the page or whatnot, they have the option to do so. With `alert()`, the user can't do anything besides clicking OK.  

I'm not that familiar with the SPA code here, maybe Dario can chime in with more, maybe we can even offer the option to retry without a full reload?

rjmateus (approved):
LGTM. But I think you should have a look at Karl suggestion, it looks to make a lot of sense

ncounter:
@ricardoasmarques thanks a lot for the contribution and the very good improvement of the behavior. While we are at this, I would really go with what @Etheryte already described, if possible. Maybe you can start having a look and we can discuss if it is doable or not, etc.

There is one point I'm still not sure about the `timeout` value: if the time to respond depends on the amount of data in the database (as usual, but in this case some query is slow because the data can be reeeeeally huge), on top of what do we set a good `timeout` value for everything? There will always be a condition limit the customer will reach and he will never be able to load the page because of a constant timeout (like the bug you are fixing). We are now moving the limit up, but why not to set it up to 5 minutes instead of 2 then? I am not complaining at all, I am asking for more opinions because I am not sure what should we expect in this case. Should we drop the timeout completely and just keep it loading forever (I mean from this manageable layer)?


@Etheryte partial reloading and keeping the values selected would be very nice but I don't see it easy to achieve: we have very different pages, behaviors and stack behind each request, we would need to handle it in very different cases and implementing again different solutions on top of different stacks for the same need it is something I'd like to avoid. If the improvement gained was huge maybe it would be worth it, but I don't see it in this case. I mean it is nice, but it is not worth the effort.

For the SPA context, `senna.js` is the engine that fires the URL request in async and instead of reloading the entire page it merges the diff in predefined id boxes reloading only what is changed. IIRC this reload in case of errors was added to cover corner cases where the SPA cannot handle the request properly for some reason, then reloading the page completely is the way to make sure we solve it, as described in the comment at [line 76](https://github.com/uyuni-project/uyuni/pull/3195/files#diff-97ea6b25d1c03bface1bd6648a788d8819e73a49d4f3fceb5df5d0be700fa566R76) ` If an error happens we make a full refresh to make sure the original request is shown instead of a SPA replacement`.

Last but not least, we also need to consider the possibility of the user who will disable the SPA behavior, and in that case we should keep all the rest of the behaviors consistent as much as possible. Therefore we would have an additional condition to keep in mind for a partial reload keeping the selected values in place.

**Wrapping up**: @ricardoasmarques I really like your change, :+1: . Just think about the way we want to force the user through the issue: in this case @Etheryte 's suggestion fits really well. We tell the user there was an error, we only provide the option to reload the page from the error message because that's what we want him to do, but still the user is able to decide to do it or not, to load a different page instead or just stay on the issue as long as he wants.

moio:
> There is one point I'm still not sure about the `timeout` value: if the time to respond depends on the amount of data in the database (as usual, but in this case some query is slow because the data can be reeeeeally huge), on top of what do we set a good `timeout` value for everything? There will always be a condition limit the customer will reach and he will never be able to load the page because of a constant timeout (like the bug you are fixing). We are now moving the limit up, but why not to set it up to 5 minutes instead of 2 then? I am not complaining at all, I am asking for more opinions because I am not sure what should we expect in this case. Should we drop the timeout completely and just keep it loading forever (I mean from this manageable layer)?

As general design principles:
 - every UI element that takes more than a few seconds to load should have a spinner/progress bar or similar indicator. A very busy page (let's say displaying thousands of systems) might take 10s - but that's already a pretty high time
 - we should not start adding spinner/progress bars on things that are slow without a good reason for them to be slow. In most cases, pages load slowly because we have very evident code defects, and we should treat performance bugs like any other bug
 - hence, anything that takes let's say 30s is very very likely signalling a problem: either a performance defect (eg. slow query), or a UX one (eg. missing spinner) or a completely out-of-tune configuration (eg. improper hardware). In any case, we should alert the user that the case is an anomaly and that should be what is conveyed by the dialog here

I am not super worried about choosing "the perfect amount of time" here. Let's pick a value which is more or less sensible and concentrate in clarifying with users that the situation is unexpected and requires a deeper look.

Etheryte:
(Sorry for hijacking this PR review with an architecture discussion.)  

In my opinion, 10 seconds should not be a loading time we ever serve to our users. For comparison, you wouldn't expect your Facebook feed to load for 10 seconds before you see anything. I understand that our use cases are different, but I think the general principle is the same — if something loads for more than a few seconds, as a user I assume their server is down or something else and I go to a different page. The only exception to this is when I absolutely have to be on that page, e.g. for my work, which our product is for most of the users.  

I'm think not being concerned with specific values works well. As a broad strokes example though, I think we should already log an error on 10 seconds so that we (or the instance manager) can see there's a problem. We can grandfather old pages if they're a lot of effort, but I think it would be a good guide for new pages.  

<hr>

Regarding spinners, from the React side, this is a good fit for [decorators](https://www.typescriptlang.org/docs/handbook/decorators.html). This is a solution I've used in a big project previously and it works great. Instead of every component manually tracking loading state, with decorators you can do something along these lines:  

```ts
class Component extends React.PureComponent implements Loading {
  async componentDidMount() {
    const data = await this.fetchData();
    this.setState({ data });
  }

  @HasLoading
  async fetchData() {...}
}
```

The main benefit of using decorators is that you solve error handling and the like at a single point, instead of every component reimplementing it.

ncounter:
> I think not being concerned with specific values works well. 

..and this is what I was talking about I wanted to raise up a discussion.

@moio I was not focusing on design principles or UI and so. I am more concerned about why increasing to a certain number is a good point, and why that number works better than the previous one. Say we fix to 30 seconds the timeout, there will be a page that takes always 31 seconds, this page will **never** be rendered, and this is the most important problem in this case. It doesn't matter if we are talking about 5 seconds or 120. Increasing the limit will just let the user think it is only a matter of performance, and only some of the users will come back to us reporting it. And even less will reach the timeout, but who will he will be back to us with a bug.
From my perspective we should report a problem after the time **we** define as a limit that we want the user to get back to us with a feedback, and as you said 30 seconds are already a symptom of something odd. But in the meanwhile, I would avoid pages to never be rendered because they always exceed a timeout.
Ideally (but from a general perspective):
- presenting nothing while loading if it is `< 1s`
- presenting a loading spinner if it is `1s < x < 10s`
- presenting a progress bar if it is `> 10s`
Other than that, a timeout parameter to highlights the user that something is going unexpectedly slow or wrong is a good message to deliver with @ricardoasmarques implementation. But if there was no error during the loading, then keep loading it.

@Etheryte IIRC in our codebase at the moment we have a `Loading` component to be reused by other components. A proper wrapper/decorator sounds better to me, but we need to standardize components shape before.

web/html/src/core/spa/spa-engine.tsx
```diff
@@ -73,14 +76,22 @@ window.pageRenderers.spaengine.init = function init() {
...
+             navigation.error.invalidStatus ||
+             navigation.error.requestError) {
+          window.location = navigation.path;
+        } else if (navigation.error.timeout) {
+          // Stop loading bar
+          jQuery(document.documentElement).removeClass('senna-loading')
+          // Inform user that page must be reloaded
+          const message = (
+            <>
+                Request has timed out, please <a className="btn-link" onClick={() => window.location = navigation.path}>reload the page</a>.
```
Etheryte, line 90: If this is a link anyway, it should be possible to use `href={navigation.path}` instead of `onClick={...}`.  

This is something that we have in older code here and there, but would be nice to avoid in new code. In short, things that navigate should be links and use `href`, and things that trigger actions should be buttons and use `onClick` and similar.

Etheryte:
Outside of one small comment, this looks good to me unless Darios has any further commentary.  

I think it would be good to gather the rest of the discussion into a separate ticket or an RFC or something similar to not lose the discussion that was had.

Etheryte (approved):

web/html/src/core/spa/spa-engine.tsx
ricardoasmarques: Note that `href` only works when navigating to an URL that is different from the current page. So I've changed it to a button.

ricardoasmarques:
@moio @rjmateus @Etheryte @ncounter thanks for all the feedback!

I've now updated this PR based on the discussion, so please review it again.

Here is the summary of what this PR is now doing:

- We keep de default timeout of `30 seconds`
- Request timeout is configurable (in case we really need to change it in production to fix an urgent issue)
- Request timeout notification is displayed to the user via a `Toastr` message
- The `Toastr` message has an action that allows the user to refresh the page with one click

Screenshot from 2021-02-11 14-23-55

ncounter (approved):
Thank you @ricardoasmarques for your work, and sorry for hijacking the focus a bit :+1:

moio:
> From my perspective we should report a problem after the time **we** define as a limit that we want the user to get back to us with a feedback

Absolutely, that was the intention since the beginning, sorry not having made that very clear!

>, and as you said 30 seconds are already a symptom of something odd. But in the meanwhile, I would avoid pages to never be rendered because they always exceed a timeout.
> Ideally (but from a general perspective):
>     * presenting nothing while loading if it is `< 1s`
>     * presenting a loading spinner if it is `1s < x < 10s`
>     * presenting a progress bar if it is `> 10s`
>       Other than that, a timeout parameter to highlights the user that something is going unexpectedly slow or wrong is a good message to deliver with @ricardoasmarques implementation. But if there was no error during the loading, then keep loading it.

I agree on the ideal model. This PR should only concern the very last case and FMPOV we could also make the time non-configurable (eg. hardcoded 30s), as users increasing the value will only hide problems.

It is my understanding that:
 - after 30s the page will present a bubble but continue loading. User gets an explanation that something may be wrong, can decide to navigate away, reload the page or wait some more
 - if user just wants to wait he/she of course can. Content will be presented as soon as available (but bubble does not get automatically dismissed)

ricardoasmarques:
> ...
> 
> I agree on the ideal model. This PR should only concern the very last case and FMPOV we could also make the time non-configurable (eg. hardcoded 30s), as users increasing the value will only hide problems.

@moio Should I remove the `web.spa.timeout` config?

> 
> It is my understanding that:
> 
> * after 30s the page will present a bubble but continue loading. User gets an explanation that something may be wrong, can decide to navigate away, reload the page or wait some more
> * if user just wants to wait he/she of course can. Content will be presented as soon as available (but bubble does not get automatically dismissed)

Just to clarify, this is not how SPA works... if a timeout happens page will no longer load. User will have to reload or navigate to a  different page.

moio:
> Just to clarify, this is not how senna/SPA works... if a timeout happens page will no longer load. User will have to reload or navigate to a different page.

I see, thanks for clarifying, this is something I did not understand well.

If this is not something we can decently workaround then no, the timeout must stay as a parameter to provide quick workarounds in case of problems.

PR #3197, opened 2021-02-01
Detect debian products

mcalmer:
## What does this PR change?

Debian 9/10 were not found as Installed Products. The detection of these products were missing.

## GUI diff

No difference.

Before:

image

After:

image

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13875
Tracks https://github.com/SUSE/spacewalk/pull/13874

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/utils/SaltUtils.java
+++ b/java/code/src/com/suse/manager/utils/SaltUtils.java
@@ -1369,6 +1369,9 @@ public class SaltUtils {
                 .map(StateApplyResult::getChanges)
                 .filter(ret -> ret.getStdout() != null)
                 .map(CmdResult::getStdout);
+
+        ValueMap grains = new ValueMap(result.getGrains());
+
         if (rhelReleaseFile.isPresent() || centosReleaseFile.isPresent() ||
                 oracleReleaseFile.isPresent() || resReleasePkg.isPresent()) {
             Set<InstalledProduct> products = getInstalledProductsForRhel(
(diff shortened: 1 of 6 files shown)


java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1369,6 +1369,9 @@ private void handlePackageProfileUpdate(MinionServer server,
                 .map(StateApplyResult::getChanges)
                 .filter(ret -> ret.getStdout() != null)
                 .map(CmdResult::getStdout);
+
+        ValueMap grains = new ValueMap(result.getGrains());
```
mbologna, line 1373: IIUC moving the declaration here has the side effect that grains will always get updated for Ubuntu (see https://github.com/uyuni-project/uyuni/commit/939378ecb4570d89ebf698b389210281a3b89818)
Would that be a problem?

mcalmer: I think this will not happen. The "if" statement for the update code block has not changed. We still test for `result.getGrains().isEmpty()` (see https://github.com/uyuni-project/uyuni/pull/3197/files#diff-eb3aa818a08d5c5af464fbc549102ad7ab37289784f95ee48750babe41091b59L1398)

mcalmer: But I wonder if it would crash (even without my change). What will happen with this:

```
"ubuntu".equalsIgnoreCase((String) null))
```
If `result.getGrains()` is empty and we call `get("something")` on it, it will return `NULL`.

mbologna: Yes, I believe it would crash even without this PR.

mcalmer: I will try to come up with a fix for this as well

mcalmer: I use now the ValueMap to get the `os` grains. When it does not exist, it return an empty string.

mbologna (approved):

PR #3202, opened 2021-02-02
Fallback to sysfs when reading info from python-dmidecode fails (bsc#1182603, bsc#1173893)

hustodemon:
(review commit-by-commit - there is some small refactoring as well)

We use `python-dmidecode` to get various hardware info on traditional clients. The problem is, that it doesn't behave well on systems that have kernel lockdown feature enabled (i.e. when `Secure boot enabled and kernel locked down` appears in the `dmesg` output).

Whereas current (plain) `dmidecode` works on such systems, `python-dmidecode` does not as it relies on reading `/dev/mem`, which seems to be restricted.

The best solution would be to align `python-dmidecode` to the latest `dmidecode` codebase, but `python-dmidecode` seems to be a inactive project since some years and aligning wourd require too much effort.
Likewise, backporting just the `/dev/mem` from `dmidecode` to `python-dmidecode` doesn't seem to be an option either, since the codebases diverged massively in the latest years.

This PR solves the problem in our `up2date_client`. The code reads the needed data directly from sysfs (`/sys/devices/virtual/dmi/id`), see `dmidecode.8`:
```
Note: on Linux, most of these strings can alternatively be read  directly
from sysfs, typically from files under /sys/devices/virtual/dmi/id.  Most
of these files are even readable by regular users.
```

The code reading info from sysfs kick in only in cases when reading via `python-dmidecode` fails (so on systems without kernel lockdown, nothing should change).

## GUI diff

- [x] **DONE**

## Documentation
bugfix

## Test coverage
- No tests in the legacy client code AFAIK

- [x] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/13654

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/client/rhel/spacewalk-client-tools/spacewalk-client-tools.changes
+++ b/client/rhel/spacewalk-client-tools/spacewalk-client-tools.changes
@@ -1,3 +1,4 @@
+- Fallback to sysfs when reading info from python-dmidecode fails (bsc#1182603)
 - log an error when product detection failed (bsc#1182339)
 
 -------------------------------------------------------------------
--- a/client/rhel/spacewalk-client-tools/src/up2date_client/hardware.py
+++ b/client/rhel/spacewalk-client-tools/src/up2date_client/hardware.py
@@ -900,9 +900,6 @@ def get_hal_system_and_smbios():
 def get_smbios():
     """ Returns dictionary with values we are interested for.
         For historical reason it is in format, which use HAL.
-        Currently in dictionary are keys:
-        smbios.system.uuid, smbios.bios.vendor, smbios.system.serial,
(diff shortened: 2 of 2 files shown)


meaksh (approved):

mcalmer (commented):
I have a question: how does this behave on `ppc64le` and `s390x` ?

mcalmer (approved):
Answering myself: exception are catched and will return empty string. So no crash. Same result as before when no dmi is available.

hustodemon:
Thanks for the reviews @meaksh and @mcalmer  ! Will merge this as soon as get positive feedback from the user affected by this.

PR #3122, opened 2021-01-13
Remove ABRT and the "Software Crashes" feature

ricardoasmarques:
## What does this PR change?

This PR removes the **ABRT** and the "**Software Crashes**" feature which includes removing all related WebUI and command-line parts as well as the corresponding database tables.

## GUI diff

Before:

Screenshot from 2021-01-13 16-33-49

After:

All "Sofware Crashes" occurrences were removed, including the one from the left main menu.

- [x] **DONE**

## Documentation
- Documentation PR: https://github.com/uyuni-project/uyuni-docs/pull/770

- [x] **DONE**

## Test coverage
- No tests: This PR only removes features, so it does not add any tests.

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13075

- [x] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #3214, opened 2021-02-04
Fix schema migration when 'rhnservercrash' was already dropped

ricardoasmarques:
### PROBLEM

Since PR https://github.com/uyuni-project/uyuni/pull/3122, the CI [1] is failing while running `susemanager-schema-4.2.4-to-susemanager-schema-4.2.5/702-debvercmp.sql` on a database that was already migrated to `4.27` (because the `rhnservercrash` was already dropped on `susemanager-schema-4.2.6-to-susemanager-schema-4.2.7/801-drop-servercrash.sql`).

From Jenkins logs:
```
susemanager-schema-4.2.7-to-susemanager-schema-4.2.8/0033-702-debvercmp.sql.postgresql

...

psql:<stdin>:363: ERROR:  relation "rhnservercrash" does not exist

LINE 3:          from rhnservercrash p

                      ^

ERROR: Upgrade failed! Most probably the new SQL scripts are not idempotent!
```

### SOLUTION

Use a pl/pgsql block to check if `rhnservercrash` exists before updating it.

[1] https://ci.suse.de/blue/organizations/jenkins/manager-Head-schema-migration-test-pgsql/detail/manager-Head-schema-migration-test-pgsql/11372/pipeline/

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Not visible to user

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/schema/spacewalk/upgrade/susemanager-schema-4.2.4-to-susemanager-schema-4.2.5/702-debvercmp.sql
+++ b/schema/spacewalk/upgrade/susemanager-schema-4.2.4-to-susemanager-schema-4.2.5/702-debvercmp.sql
@@ -365,18 +365,31 @@ where rhnserverprofilepackage.server_profile_id = sub.server_profile_id and
 
 
 --    TABLE "rhnservercrash" CONSTRAINT "rhn_server_crash_evr_id_fk" FOREIGN KEY (package_evr_id) REFERENCES rhnpackageevr(id)
-with sub as (
-    select p.server_id, p.crash, t.label, pe.evr
-         from rhnservercrash p
-         join rhnpackagearch pa on p.package_arch_id = pa.id
-         join rhnarchtype t on pa.arch_type_id = t.id
-         join rhnpackageevr pe on p.package_evr_id = pe.id
-)
-update rhnservercrash
-set package_evr_id = lookup_evr((sub.evr).epoch, (sub.evr).version, (sub.evr).release, sub.label)
-from sub
-where rhnservercrash.server_id = sub.server_id and
-      rhnservercrash.crash = sub.crash;
+DO $$
+    BEGIN
+      -- Schema migration must be idempotent, so we need to guarantee that it will not fail
+      -- if 'rhnservercrash' table was already dropped by the '801-drop-servercrash.sql' script
+      IF EXISTS (
+            SELECT 1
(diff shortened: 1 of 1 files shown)


mcalmer (commented):
What about the "delete statement" later ? does it work if the table does not exist?

juliogonzalez (changes requested):
It seems `schema_migration_test_pgsql` didn't run for #3122, which explains what this problem wasn't detected.

Now you can see it's still failing because the update is not idempotent.

But since you're modifying a schema upgrade SQL script that was *already* released, what's going to happen to Uyuni 2020.01 users without this?

ricardoasmarques:
> What about the "delete statement" later ? does it work if the table does not exist?

Good point, I've now updated this PR to re-create the table instead of checking if it exits.

ricardoasmarques:
> But since you're modifying a schema upgrade SQL script that was _already_ released, what's going to happen to Uyuni 2020.01 users without this?

Not sure if I'm missing anything, but I believe it will work because the problem only happens when we run `4.2.4-to-4.2.5` after `4.2.6-to-4.2.7`, so:

1) if the user is in the **OLD** version, then `4.2.6-to-4.2.7` does **not exist** and `4.2.4-to-4.2.5` can run multiple times

2) if the user is in the **NEW** version, then `4.2.6-to-4.2.7` **exists** but `4.2.4-to-4.2.5` can run multiple times even if `4.2.6-to-4.2.7` was already executed, due to the fix in this PR

mcalmer:
@ricardoasmarques I think the current error exists because it execute only the changed scripts. Can you try to run standard migration test on your branch

  susemanager-utils/testing/automation/schema-migration-test-pgsql.sh

That should run exactly the same as the currently failing `manager-Head-schema-migration-test-pgsql` job.
If this is good, we can merge this.

ricardoasmarques:
@mcalmer I've switched to the PLSQL solution again to avoid an intermediate state that temporarily "pollutes" the database with a table that is not supposed to exist.

Tests are passing now, so it's ready for a new review.

mcalmer (approved):

PR #3217, opened 2021-02-04
Do not hide navigation steps in implementation

Bischoff:
## What does this PR change?

This PR does two things:

- it removes reference to bsc#1124634. Since this is an INVALID bug, I prefer we do not mention it.
  We need this scenario because of the way SUSE products work starting with SLES 15, that's all.

- it moves navigation steps from the ruby implementation to the feature. Rationale:
  * makes test suite execution easier to follow (more details)
  * it makes reading and understanding of scenario easier
  * it makes debugging easier when it fails at one of these steps
  * it is better if all our scenarios more or less follow a common pattern

Tested to work in 4.1 branch.

## Links

Ports:
* 4.0: SUSE/spacewalk#13894
* 4.1: SUSE/spacewalk#13893

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/init_clients/sle11sp4_ssh_minion.feature
+++ b/testsuite/features/build_validation/init_clients/sle11sp4_ssh_minion.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 SUSE LLC
+# Copyright (c) 2020-2021 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 @sle11sp4_ssh_minion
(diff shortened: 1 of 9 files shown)


lkotek (approved):
Yes, it is definitely more readable now. LGTM.

srbarrios:
I'm not very fun of long scenarios, but ok to approve it.
If it was me refactoring it, I would have a step called "When I go to Package States section" or similar.
In any case, approved.

srbarrios (approved):

testsuite/features/build_validation/init_clients/sle11sp4_ssh_minion.feature
```diff
@@ -20,13 +20,19 @@ Feature: Bootstrap a SLES 11 SP4 Salt SSH Minion
     Then I wait until I see "Successfully bootstrapped host!" text
     And I wait until onboarding is completed for "sle11sp4_ssh_minion"
 
-# WORKAROUD for bsc#1124634
```
srbarrios, line 23: Fine to remove the reference, but I'd like to keep the keyword "WORKAROUND"

Bischoff: One WORKs AROUND a product problem, but there's no product problem. Other than broken by design at least ;-) .

HACK maybe ?

srbarrios: Something that we can identify actions that we made that a regular customer will not do.

Bischoff: HACK then...

PR #3227, opened 2021-02-05
Filter by package state name

srbarrios:
## What does this PR change?

In our QAM and BV tests, we have a long list of package states in the table, so we need to first filter by name to be able to remove one of them.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.0 https://github.com/SUSE/spacewalk/pull/14015
- Manager-4.1 https://github.com/SUSE/spacewalk/pull/14018

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/build_validation/init_clients/sle11sp4_ssh_minion.feature
+++ b/testsuite/features/build_validation/init_clients/sle11sp4_ssh_minion.feature
@@ -33,6 +33,8 @@ Feature: Bootstrap a SLES 11 SP4 Salt SSH Minion
     Then I should see a "Package States" text
     When I follow "Search" in the content area
     And I wait until button "Search" becomes enabled
+    And I enter "sle-manager-tools-release" as the filtered package states name
+    And I click on "Search" in element "search-row"
     And I remove package "sle-manager-tools-release" from highstate
 
   # WORKAROUND bsc#1181847
--- a/testsuite/features/build_validation/init_clients/sle12sp4_ssh_minion.feature
+++ b/testsuite/features/build_validation/init_clients/sle12sp4_ssh_minion.feature
@@ -33,6 +33,8 @@ Feature: Bootstrap a SLES 12 SP4 Salt SSH Minion
     Then I should see a "Package States" text
     When I follow "Search" in the content area
     And I wait until button "Search" becomes enabled
+    And I enter "sle-manager-tools-release" as the filtered package states name
+    And I click on "Search" in element "search-row"
     And I remove package "sle-manager-tools-release" from highstate
 
   # WORKAROUND bsc#1181847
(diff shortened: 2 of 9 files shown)


Bischoff (changes requested):
What about taking the occasion to simplify the implementation of￼

   `I remove package "..." from highstate`
   
Because you now spot the right package state, and that step loops over the whole table. But, if i am not wrong, you will get a single entry now, so mucher simpler code might be used.

srbarrios:
> What about taking the occasion to simplify the implementation of
> 
> `I remove package "..." from highstate`
> 
> Because you now spot the right package state, and that step loops over the whole table. But, if i am not wrong, you will get a single entry now, so mucher simpler code might be used.

Make sense.

srbarrios:
> What about taking the occasion to simplify the implementation of
> 
> `I remove package "..." from highstate`
> 
> Because you now spot the right package state, and that step loops over the whole table. But, if i am not wrong, you will get a single entry now, so mucher simpler code might be used.

@Bischoff if you don't mind, I prefer to keep this step as it is, it has some steps... that make me doubt how it can behave with a refactor, and I prefer to keep it, in any case it will only pass once on the loop, because we expect only one row.

This is the current implementation:
```
When(/^I remove package "([^"]*)" from highstate$/) do |package|
  event_table_xpath = "//div[@class='table-responsive']/table/tbody"
  rows = find(:xpath, event_table_xpath)
  rows.all('tr').each do |tr|
    next unless tr.text.include?(package)
    puts tr.text
    tr.find("##{package}-pkg-state").select('Removed')
    next if has_css?('#save[disabled]')
    steps %(
      Then I click on "Save"
      And I click on "Apply"
    )
  end
end
```

Bischoff:
> > What about taking the occasion to simplify the implementation of
> > `I remove package "..." from highstate`
> > Because you now spot the right package state, and that step loops over the whole table. But, if i am not wrong, you will get a single entry now, so mucher simpler code might be used.
> 
> @Bischoff if you don't mind, I prefer to keep this step as it is, it has some steps... that make me doubt how it can behave with a refactor, and I prefer to keep it, in any case it will only pass once on the loop, because we expect only one row.
> 
> This is the current implementation:
> 
> ```
> When(/^I remove package "([^"]*)" from highstate$/) do |package|
>   event_table_xpath = "//div[@class='table-responsive']/table/tbody"
>   rows = find(:xpath, event_table_xpath)
>   rows.all('tr').each do |tr|
>     next unless tr.text.include?(package)
>     puts tr.text
>     tr.find("##{package}-pkg-state").select('Removed')
>     next if has_css?('#save[disabled]')
>     steps %(
>       Then I click on "Save"
>       And I click on "Apply"
>     )
>   end
> end
> ```

We can do it later.

Bischoff (approved):
lgtm, we can do the small refactoring separately

calancha (approved):

srbarrios:
I finally had time to test it and works. Ready to be merged.

juliogonzalez:
Bypassing `java_pgsql_tests`, unrelated to this PR.

PR #3228, opened 2021-02-05
Testsuite: Replace pidof with pgrep

calancha:
## What does this PR change?
The current sles15sp3o image at the CI doesn't ship `pidof`; use `pgrep`
instead which is available.

Replace pidof with pgrep.

### Ports:
- Manager4-.1: https://github.com/SUSE/spacewalk/pull/13923
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/13924

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -493,8 +493,8 @@ When(/^the server starts mocking an IPMI host$/) do
 end
 
 When(/^the server stops mocking an IPMI host$/) do
-  $server.run('kill $(pidof ipmi_sim)')
-  $server.run('kill $(pidof -x fake_ipmi_host.sh)')
+  $server.run('pkill ipmi_sim')
+  $server.run('pkill fake_ipmi_host.sh || :')
 end
 
 When(/^the server starts mocking a Redfish host$/) do


calancha:
Could this be solved more simple at sumaform (`salt/server/testsuite.sls`) by requiring `pidof` been installed?

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -493,8 +493,8 @@
 end
 
 When(/^the server stops mocking an IPMI host$/) do
-  $server.run('kill $(pidof ipmi_sim)')
-  $server.run('kill $(pidof -x fake_ipmi_host.sh)')
+  $server.run('kill $(pgrep ipmi_sim)')
+  $server.run('if pgrep fake_ipmi_host.sh &> /dev/null; then kill $(pgrep fake_ipmi_host.sh); else :; fi')
```
Bischoff, line 497: `pgrep fake_ipmi_host.sh &> /dev/null && kill $(pgrep fake_ipmi_host.sh)` maybe?

Bischoff (changes requested):
suggested change

testsuite/features/step_definitions/command_steps.rb
calancha: In my local tests the second `pgrep` doesn't find the process; in such a case, with your form,
run will raise (the form would exit with code 1).
It might be that when using `pgrep`, we are killing everything in the first line; then we would not need the second.
I tried to be defensive, so that my form always return 0 exit code.

calancha: Indeed, we don't need a subshell: `pkill` is also available, we can just `pkill process-foo`.

calancha:
It's working well in my test machine; I want this fixed at head before leaving, so I am merging it.

PR #3239, opened 2021-02-08
Modernize static Proxy home page

ncounter:
## Reviewers

- @mcalmer please take a look only at the overview change if it makes sense or not
- @juliogonzalez please take a look at the `.spec` only, that's something I think you have the best up-to-date knowledge
- @cbbayburt pretty trivial and straight forward static code, but it deserves a review anyway

Thank you

## What does this PR change?

**Context**: when accessing a proxy system via `http://<proxy_ip_address>` the machine provides a static homepage where there is a link to the filesystem `/pub` directory content. The page was really outdated (showing a SUSE Manager 1.2 logo page).

This PR is about to wildly refactoring of the static proxy homepage. In addition to that, there are changes to the `.spec` file in order to serve a *suse_proxy* or a *uyuni_proxy* branded homepage depending on the product being installed.

**Drawback**: being the page a complete static resource and not willing to invest more in this to make it dynamic anyway for now, the code has to be duplicated and customized for both the product versions. But that's really a small chunk of code. Whenever it will grow, a proper dynamic switching and code-sharing should be implemented instead.

## GUI diff

Before:
Screenshot from 2021-02-08 11-20-14

After:
Screenshot from 2021-02-08 11-20-25

or
Screenshot from 2021-02-08 15-09-41

- [x] **DONE**

## Documentation
- No documentation needed: @mcalmer @Loquacity do we have this documented anywhere?

- [x] **DONE**

## Test coverage
- No tests: static files not tested

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/12686

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


proxy/proxy-html/spacewalk-proxy-html.spec
```diff
@@ -49,18 +56,17 @@ displays, if you navigate to it using your browser.
...
-%{htmldir}/_rhn_proxy/*.ico
-%{htmldir}/_rhn_proxy/*.png
+%dir %{htmldir}
+%config %{htmldir}/index.html
+%{htmldir}/sources/img/*.ico
+%{htmldir}/sources/img/*.png
 %doc LICENSE
 %if 0%{?suse_version}
-%dir %dir %{htmldir}/_rhn_proxy
+%dir %dir %{htmldir}
```
juliogonzalez, line 69: You are already included `%{htmldir} at line 64 for all OS.

Looks to me that this whole if can go away?

juliogonzalez (commented):
One comment for the spec.

Loquacity:
> * No documentation needed: @mcalmer @Loquacity do we have this documented anywhere?

I don't think so. Perhaps @keichwa can think of something I'm missing though.

keichwa:
thanks for pinging.  No, we do not mention the (old) proxy home page in the docs.  I'm not sure whether the user has to know about it.

proxy/proxy-html/spacewalk-proxy-html.spec
ncounter: Good point @juliogonzalez , it was the case when there was the subfolder `_rhn_proxy` before this PR, but now it is removed and the whole `if` condition can be dropped. Thanks

proxy/proxy/httpd-conf/spacewalk-proxy-wsgi.conf
```diff
@@ -21,7 +21,7 @@ WSGIPythonPath "/usr/share/rhn"
 </IfVersion>
 
 <LocationMatch "^/*">
-    DirectoryIndex _rhn_proxy/index.html index.html index.htm index.html.var index.shtml index.php index.php4 index.php3 index.phtml index.cgi
+    DirectoryIndex suse_proxy/index.html index.html index.htm index.html.var index.shtml index.php index.php4 index.php3 index.phtml index.cgi
```
mcalmer, line 24: I think this is wrong. It seems that `suse_proxy` and `uyuni_proxy` only exists on the sources.
After installation, its exists directly in the root dir, right?

I think `suse_proxy/index.html` can be removed

proxy/proxy/httpd-conf/spacewalk-proxy.conf
```diff
@@ -39,7 +39,7 @@
     SetHandler None
 </LocationMatch>
 
-<LocationMatch "^/_rhn_proxy">
+<LocationMatch "^/suse_proxy">
```
mcalmer, line 42: I think this can be removed as well. I think there is no `suse_proxy` in the installed system.

mcalmer (changes requested):

proxy/proxy/httpd-conf/spacewalk-proxy-wsgi.conf
ncounter: Good catch, thanks for pointing it out. Initially I only renamed it, but then I went for the double option removing the subfolder, so yes everything goes in the web root (which is `/srv/www/htdocs`). Fixed now :+1:

proxy/proxy/httpd-conf/spacewalk-proxy.conf
ncounter: Fixed :+1:

cbbayburt (approved):
LGTM :+1: (I reviewed the frontend only)

BTW, that favicon is also outdated as hell :rofl: 
It's not relevant for this PR though, it's the same in the server anyway.

mcalmer (approved):

ncounter:
> BTW, that favicon is also outdated as hell rofl
> It's not relevant for this PR though, it's the same in the server anyway.

I updated it anyway. Time to touch this path once and never go back again :wink:

juliogonzalez:
> I updated it anyway. Time to touch this path once and never go back again wink

By "never" you mean until next year, right? :trollface: 

Remember to add the `merge-candidate` label if you want this merged soon.

PR #3242, opened 2021-02-08
Migrate free-range components

Etheryte:
## What does this PR change?

~~Blocked by waiting on the merge of https://github.com/uyuni-project/uyuni/pull/3180~~

Migrate all components that aren't part of a subdirectory to Typescript.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Part of: https://github.com/SUSE/spacewalk/issues/13739
Fixes: https://github.com/SUSE/spacewalk/issues/13931

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


Etheryte:
The previous PR got merged so this is no ready for review.

web/html/src/components/ace-editor.tsx
```diff
@@ -7,25 +6,15 @@ type Props = {
   minLines: number;
   maxLines: number;
   readOnly: boolean;
-  onChange: (...args: any[]) => any;
+  onChange?: (...args: any[]) => any;
   className: string;
   id: string;
   content: React.ReactNode;
+  /** Obsolete, do not use */
+  name?: string;
```
hustodemon, line 14: Looks like the only usage of the ace editor is in the `config-channels.tsx`: we pass the `name`, but we don't use it in the component. Can we remove it?

hustodemon (approved):
Thanks for the big effort @Etheryte :+1:  Looks good to me, I only left 1 minor comment to check.

Have you run the whole cucumber testsuite on your branch? It'd be worth it as it could reveal any possible unpleasant surprises. Just ping me if you need more info about that.

web/html/src/components/ace-editor.tsx
Etheryte: Good shout, thanks for pointing this out. Might even be worth keeping an eye out in the future, perhaps we can cut this dependency out altogether at some point.

Etheryte:
I haven't run Cucumber on this branch alone, but almost all of the changes with all TS migrations should be one-to-one, i.e. functionally equivalent.  
I have tested this sporadically manually, but mostly I rely on the TS compiler and JS build to find any issues I might've missed.

PR #3248, opened 2021-02-09
Migrate `components/dialog`

Etheryte:
## What does this PR change?

Migrate all components in `components/dialog` to Typescript

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/components/dialog/ActionConfirm.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { DangerDialog } from "./DangerDialog";
+
+type Props = {
+  id: string;
+  type: string;
+  name: string;
+  itemName: string;
+  icon: string;
+  selected: any[];
+  // TODO: This should be renamed `onConfirm` or something similar
```
ncounter, line 11: Feel free to do the rename, `onConfirm` makes sense to me as it is used only in that case IIUC.

web/html/src/components/dialog/ModalLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+  title?: string;
+  text: React.ReactNode;
+  icon?: string;
+  disabled?: boolean;
+  item?: any;
+  onClick: (...args: any[]) => any;
+};
+
+/**
+ * DEPRECATED: Do **NOT** use this component for new code, prefer `ModalButton` instead
```
ncounter, line 17: :+1:

web/html/src/components/dialog/ModalButton.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react";
+import { Button } from "../buttons";
+import { showDialog } from "./util";
+
+// TODO: Extend Button's props here once that branch gets merged
```
ncounter, line 5: Make sure to not forget about this, we try to avoid merging `TODO`s , but I also see a reason here so, just keep it in mind to get rid of it ASAP, or track it into a proper issue at least.

ncounter (approved):

web/html/src/components/dialog/ModalButton.tsx
Etheryte: That's a fair point, a number of places in the migration process rely on essentially "fix this once something else gets done as well" which may be weeks apart. I'll add a ticket to look over all of these, thanks for the suggestion. 👍

Etheryte: Created https://github.com/SUSE/spacewalk/issues/13952

web/html/src/components/dialog/ActionConfirm.tsx
Etheryte: Thanks, I went ahead and changed all the occurrences.

juliogonzalez:
java_pgsql_tests was aborted, but doesn't affect this PR. I will merge it.

PR #3249, opened 2021-02-09
Remove no longer used images

ncounter:
## What does this PR change?

Dropping `/branding/img` images no longer used in the project.

Script used to check and drop files not in use:
```bash
#!/bin/bash

cd ~/spacewalk
path_to_clean=~/spacewalk/branding/img/

for absolute_file in `find $path_to_clean -type f
do
  echo $absolute_file
	file_name=$(basename $absolute_file)
	echo $file_name
	if [ -z "$(rg $file_name)" ]; then
		echo "no usage found.........deleting the file"
		echo `rm $absolute_file`
	else
		echo "There are usages, see the following"
		echo `rg $file_name`
	fi
	echo
done

for dir in `find $path_to_clean -type d -empty`
do
	echo "$dir is now empty.........deleting the dir"
	echo `rm -r $dir`
	echo
done
exit 0
```
Note: In the script the `rg` command stands for [`rip grep`](https://github.com/BurntSushi/ripgrep).

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: 

- [x] **DONE**

## Test coverage
- No tests: 
- 
- [x] **DONE**

## Links

Fixes #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


Etheryte:
To be honest, I don't know how to review this outside checking the paths. I don't know whether the paths might be dynamically generated somewhere as well? This looks good as far as I can see, but it might be good to get a second opinion too. 😛

Etheryte (approved):

ncounter:
@Etheryte thanks for the quick review. Yeah, for a little bit of context, the most of the dropped files are `*.gif` and they belong to the original spacewalk project, everything we added during the years was introduced in the subfolder `/branding/img/susemanager`. Anyway, we started using font-awesome for everything more or less, plus a little bit of custom font icons. AFAIK we don't have any image in use a part from the logo and few more. So my job here was to drop some *spacewalk-related* files, plus some no longer used.
For the dynamically generated ones, I agree we have some generated like that, but I confidentially believe it is only about the react part and only for font icons like `fa-`+*some_dynamic_string*.

In the end, let's see how much I can break this time :smiley: 

Added @hustodemon as an additional reviewer, also for checking if the script I used it makes any sense or it is buggy.

Thank you

hustodemon (approved):
LGTM. I also re-checked with my simple script.

PR #3250, opened 2021-02-10
Traditional stack change ssl implementation

mcalmer:
## What does this PR change?

Change rhnlib SSL implementation to use python ssl instead of pyOpenSSL. This brings us better SAN support and
better hostname matching during SSL handshake.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue: https://github.com/SUSE/spacewalk/issues/13902
Tracks https://github.com/SUSE/spacewalk/pull/13988 https://github.com/SUSE/spacewalk/pull/13989

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/client/rhel/spacewalk-client-tools/src/up2date_client/rhnregGui.py
+++ b/client/rhel/spacewalk-client-tools/src/up2date_client/rhnregGui.py
@@ -65,7 +65,7 @@ from up2date_client import up2dateUtils
 from up2date_client import config
 import OpenSSL
 from up2date_client import up2dateLog
-from rhn import rpclib
+from rhn import rpclib, SSL
 from rhn.connections import idn_puny_to_unicode
 from up2date_client import rhnreg_constants
 from up2date_client.pmPlugin import PM_PLUGIN_NAME, PM_PLUGIN_CONF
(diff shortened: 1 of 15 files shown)


client/rhel/spacewalk-client-tools/src/up2date_client/rhnregGui.py
```diff
@@ -960,7 +960,7 @@ def _provideCertificatePageApply(self):
                     errorWindow(rhnreg_constants.SSL_CERT_ERROR_MSG % (certFile, server_url))
 
                 return ERROR_WAS_HANDLED
-            except OpenSSL.SSL.Error:
+            except (OpenSSL.SSL.Error, SSL.SSL.SSLError):
```
meaksh, line 963: Why do you keep `OpenSSL.SSL.Error` here ?

Shouldn't this simply be:

```suggestion
            except SSL.SSL.SSLError:
```

mcalmer: This code use both, SSL and OpenSSL. I have no idea what can be raised here.
Can we be sure that pyOpenSSL will not raise its own `Error` ?

meaksh: Ah, I see what you meant - we're calling some methods from the `pyOpenSSL.crypto` that might raise that custom pyOpenSSL exceptions.

Ok, looks good then :+1:

mcalmer:
CC @juliogonzalez Before you tag this and submit it for release, we need to add versioned dependencies to new rhnlib version.

agraul (approved):
Looks good to me. The SSL abstraction is leaking a bit but this is only the traditional stack and I think it is ok to leave it as it is.

PR #3255, opened 2021-02-11
Migrate `components/formulas`

Etheryte:
## What does this PR change?

Migrate all components in `components/formulas` to Typescript.  
These are pretty messy files to begin with, with a lot of things thrown into one pile.  
For now, I opted to only make this pass compilation without trying to fix the architectural problems.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


ncounter (approved):
I completely agree with 
> These are pretty messy files to begin with, with a lot of things thrown into one pile.

It is out of this PR scope anyway so, this LGTM :+1:

web/html/src/components/formulas/EditGroup.tsx
```diff
@@ -0,0 +1,403 @@
+/* eslint-disable */
```
cbbayburt, line 1:
```suggestion
```

web/html/src/components/formulas/PasswordInput.tsx
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable */
```
cbbayburt, line 1:
```suggestion
```

cbbayburt (approved):
LGTM :+1: 
However, I'm not familiar with the formulas code at all, so I might have missed something. Maybe one of the original authors should also take a look.

Below I pointed out some leftover `eslint-disable` directives

web/html/src/components/formulas/EditGroup.tsx
Etheryte: Hmm, looks like the migration script isn't cleaning these up correctly. I'll have to have a look later. I cleaned these up manually for now.

PR #3257, opened 2021-02-11
Migrate `components/panels`

Etheryte:
## What does this PR change?

Migrate all components in `components/panels` to Typescript.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/components/panels/BootstrapPanel.tsx
```diff
@@ -31,6 +32,4 @@ BootstrapPanel.defaultProps = {
   footer: undefined,
 };
```
cbbayburt, line 33: Above: I think `defaultProps` is no-op here, right?

cbbayburt (approved):

web/html/src/components/panels/BootstrapPanel.tsx
Etheryte: Good catch, that's indeed the case. Thanks for looking through it. 👍

ncounter (approved):

PR #3258, opened 2021-02-11
Migrate `components/picker`

Etheryte:
## What does this PR change?

Migrate all components in `components/picker` to Typescript.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


hustodemon (approved):
LGTM!

web/html/src/global.d.ts
```diff
@@ -47,6 +47,9 @@ declare global {
 
   // Used by section-toolbar, defined in spacewalk-essentials.js
   var handleSst: Function;
+
+  /** DEPRECATED: Do **NOT** use this global for new code, prefer `useUserLocalization()` instead */
```
ncounter, line 51: :+1:

web/html/src/global.d.ts
```diff
@@ -47,6 +47,9 @@ declare global {
...
+  var localTime: string | undefined;
 }
 
 export {};
```
ncounter, line 55: Not part of this PR but, could you please add a final empty line? Git/GitHub complain about it, thanks.

web/html/src/components/picker/recurring-event-picker.tsx
```diff
@@ -0,0 +1,484 @@
...
+import { DateTimePicker } from "../datetimepicker";
+import { Combobox } from "../combobox";
+import { ComboboxItem } from "../combobox";
+import { Form } from "components/input/Form";
+import { Text } from "components/input/Text";
+import { Utils } from "utils/functions";
+import styles from "./recurring-event-picker.css";
+
+type RecurringType = "hourly" | "daily" | "weekly" | "monthly" | "cron";
+// TODO: This should be `Record<string, string>` or `Record<string, number>`, but currently they're used mixed up
```
ncounter, line 12: I didn't took a deep look into this: how hard/long could it be to address and fix this? Would it be worth of doing it right now or we can live with it?

ncounter (approved):
I left a couple of comments but more as questions. The PR LGTM :+1:

web/html/src/components/picker/recurring-event-picker.tsx
Etheryte: This would require functional changes, something that I'm trying to avoid in the migration since it would require considerably deeper testing of everything related. Essentially the problem is that some places set the values as strings while others set it as numbers and it isn't trivial to refactor this without testing it through the full stack.  

Since this code works right now, I don't see a need to put this time and effort into here right now, but this TODO is a good option for anyone who happens to work more in depth on this piece in the future.

PR #3262, opened 2021-02-11
We have 2 BETA channels in HEAD

Bischoff:
## What does this PR change?

In step `And I should see a "Build (4)" text`, the text `Build(6)` appears, because we have two extra channels:
* SLE-Manager-Tools12-BETA-Pool
* SLE-Manager-Tools12-BETA-Updates

This PR adapts the test. It will need to be reverted when we branch to 4.2.

## Links

No ports, HEAD only.

## Changelogs

- [x] No changelog needed


calancha (approved):
Thank you.

lkotek (approved):

ktsamis (approved):

PR #3282, opened 2021-02-15
eslint: The href attribute requires a valid value to be accessible (anchor-is-valid)

ricardoasmarques:
## What does this PR change?

This PR fixes all `anchor-is-valid` eslint warnings.

**Before**
```
$ yarn lint
...
warning: The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. If you cannot provide a valid href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md (jsx-a11y/anchor-is-valid) at components/table/Header.tsx:44:9:
...
42 warnings found.
```

**After**
```
$ yarn lint
...
37 warnings found.
```

Learn more: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md

## GUI diff

No difference.

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/components/dialog/ModalLink.tsx
+++ b/web/html/src/components/dialog/ModalLink.tsx
@@ -21,13 +21,11 @@ export function ModalLink(props: Props) {
   const margin = props.text ? "" : " no-margin";
   var icon = props.icon && <i className={"fa " + props.icon + margin} />;
 
-  // TODO: Using `href="#"` is bad for accessibility and semantics, this should be a `<button>` that looks like it needs to
   return (
-    <a
+    <button
       id={props.id}
-      href="#"
       title={props.title}
-      className={props.className}
+      className={'btn-link ' + (props.className || '')}
       onClick={() => {
         if (props.onClick) props.onClick(props.item);
         showDialog(props.target);
@@ -35,6 +33,6 @@ export function ModalLink(props: Props) {
     >
       {icon}
       {props.text}
-    </a>
+    </button>
   );
 }
--- a/branding/css/tables.less
+++ b/branding/css/tables.less
@@ -133,14 +133,18 @@ tr.alphaResult td.sortedCol:before {
   font-family: 'FontAwesome';
 }
 
-th.ascSort a:after {
+th.ascSort a:after, th.ascSort button:after {
   content: "  \f160";
   font-family: 'FontAwesome';
 }
 
-th.descSort a:after {
+th.descSort a:after, th.descSort button:after {
   content: "  \f161";
   font-family: 'FontAwesome';
+}
+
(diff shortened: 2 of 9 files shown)


web/html/src/components/dialog/ModalLink.tsx
```diff
@@ -21,11 +21,9 @@ export function ModalLink(props: Props) {
...
   var icon = props.icon && <i className={"fa " + props.icon + margin} />;
 
-  // TODO: Using `href="#"` is bad for accessibility and semantics, this should be a `<button>` that looks like it needs to
   return (
-    <a
+    <button
       id={props.id}
-      href="#"
       title={props.title}
       className={props.className}
```
ncounter, line 28: Is this perhaps missing the `btn-link` className? Maybe it is inherited but, just to make sure, I'd rather add it here. Am I missing something?

ncounter (commented):
I left one single comment. Other than that, it LGTM, thanks a lot for taking the time to fix this issue.

Etheryte:
This is very nice, something I've had on my mind for ages but never got around to actually addressing. Thanks! Outside of the question by Dario, looks good to me.

Etheryte (approved):

Etheryte:
One small addition I forgot before, if this fixes all of the occurrences of the above rule, I think we should set it to error in `.eslintrc.js`. This is a pretty common issue and is prevalent in code samples online as well, making the rule an error will make it hard to reintroduce this type of issue again.

ricardoasmarques:
> One small addition I forgot before, if this fixes all of the occurrences of the above rule, I think we should set it to error in `.eslintrc.js`. This is a pretty common issue and is prevalent in code samples online as well, making the rule an error will make it hard to reintroduce this type of issue again.

Done.

web/html/src/components/dialog/ModalLink.tsx
ricardoasmarques: Nice catch, in some cases the `btn-link` className were not inherited.

@ncounter I've updated PR accordingly, please take another look.

cbbayburt (approved):
LGTM too :+1:

ncounter (approved):
:+1:

juliogonzalez:
Ping @ktsamis (as next RRTG), this is changing a link to a button, in case you see problems at Master/head in the next cucumber runs.

PR #3302, opened 2021-02-19
Remove unused variables in frontend code

ricardoasmarques:
## What does this PR change?

This PR removes unused variables in frontend code, which reduces the number of `eslint` warnings.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/components/datetimepicker.tsx
+++ b/web/html/src/components/datetimepicker.tsx
@@ -131,10 +131,10 @@ class TimePicker extends React.Component<TimePickerProps> {
       roundingFunction: (seconds, options) => seconds,
     });
     this._input?.on("change", () => {
-      const timepickerValue = this._input?.timepicker("getTime");
+      // Do nothing
     });
     this._input?.on("timeFormatError", () => {
-      const timepickerValue = this._input?.timepicker("getTime");
+      // Do nothing
     });
     this._input?.timepicker("setTime", this.props.value);
     this._input?.on("changeTime", () => {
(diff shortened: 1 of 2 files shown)


web/html/src/components/datetimepicker.tsx
```diff
@@ -131,10 +131,10 @@ class TimePicker extends React.Component<TimePickerProps> {
       roundingFunction: (seconds, options) => seconds,
     });
     this._input?.on("change", () => {
-      const timepickerValue = this._input?.timepicker("getTime");
+      this._input?.timepicker("getTime");
```
Etheryte, line 134: It isn't clear whether this is the right action here. If we get the time and then don't use it I don't think we need this call either, as long as it has no side effects. Perhaps @ncounter is more familiar with it?  

Personally I would recommend making the whole block empty with the comment `// Do nothing` so that it's clear that it's intentionally an empty block. The value in keeping the empty block is that if someone needs to change this code next, they won't have to hunt down the docs for the plugin.

ricardoasmarques: Nice catch! It's fixed now, thanks.

Etheryte (approved):

Etheryte:
This looks good to me, Dario can speak now or forever hold his silence. 😃

web/html/src/components/datetimepicker.tsx
ncounter: I agree with you guys. I can see the same pattern few lines after where it is also used, but in these two cases I don't see the need not only for that line, but also for the entire event action if we do not react in anyway.
Maybe for completeness we should ask @lucidd if there is a particular reason why he added back in the day almost 5 years ago in [this commit](https://github.com/uyuni-project/uyuni/commit/e2f937e9f245df82c406b43e7bbc94e21432bb28).

ncounter (approved):
I left a comment in the ongoing discussion. Other than that, it LGTM :+1:

PR #3306, opened 2021-02-19
Testsuite - fix of package refresh by spacecmd

lkotek:
## What does this PR change?

PR fixes the way package list refresh is handled by `spacecmd` in testsuite. 
We need to wait for all package refreshes to be completed.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1362,12 +1362,22 @@ Then(/^I wait until refresh package list on "(.*?)" is finished$/) do |client|
   timeout_time = (Time.now + long_wait_delay + round_minute).strftime('%Y%m%d%H%M')
   node = get_system_name(client)
   $server.run("spacecmd -u admin -p admin clear_caches")
-  cmd = "spacecmd -u admin -p admin schedule_listcompleted #{current_time} #{timeout_time} #{node} | grep 'Package List Refresh scheduled by admin' | head -1"
+  # Gather all the ids of package refreshes existing at SUMA
+  refreshes, = $server.run("spacecmd -u admin -p admin schedule_list | grep 'Package List Refresh scheduled by admin' | cut -f1 -d' '", false)
+  node_refreshes = ""
+  refreshes.split(' ').each do |refresh_id|
+    next unless refresh_id.match('/[0-9]{1,4}/')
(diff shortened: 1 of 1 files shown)


lkotek:
I tested `schedule_list` -> it contains all events including the failed ones.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -1362,12 +1362,13 @@
   timeout_time = (Time.now + long_wait_delay + round_minute).strftime('%Y%m%d%H%M')
   node = get_system_name(client)
   $server.run("spacecmd -u admin -p admin clear_caches")
-  cmd = "spacecmd -u admin -p admin schedule_listcompleted #{current_time} #{timeout_time} #{node} | grep 'Package List Refresh scheduled by admin' | head -1"
+  cmd = "spacecmd -u admin -p admin schedule_list #{current_time} #{timeout_time} #{node} | grep 'Package List Refresh scheduled by admin'"
```
mcalmer, line 1365:
```
schedule_list: List all actions
usage: schedule_list [BEGINDATE] [ENDDATE]
```
There is no "system" as parameter. So it makes no sense to give `#{node}` as parameter

mcalmer (changes requested):
please remove that option. It give the wrong impression that we look for these system only.

testsuite/features/step_definitions/command_steps.rb
lkotek: Thanks. Yes, this is good point. I am re-working this a bit, so it still will be possible to get status of package refresh specific for the selected system.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -1362,12 +1362,13 @@
...
   repeat_until_timeout(timeout: long_wait_delay, message: "'refresh package list' did not finish") do
     result, code = $server.run(cmd, false)
+    sleep 1
+    next if result.include? '0    0    1'
     break if result.include? '1    0    0'
```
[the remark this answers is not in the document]
lkotek: It is correct, but there is `next if result.include? '0    0    1'`, which addresses exactly this.

lkotek:
Current solution was tested multiple times. It worked perfectly :-)

Bischoff (approved):
LGTM. There's quite a number of calls to spacecmd, but i guess there's no way around it...

mcalmer (approved):
Let's give it a try

PR #3307, opened 2021-02-19
Testsuite: Fix check for env. vars set by terracumber

calancha:
## What does this PR change?
It fixes a few environment variable checks.

Terracumber sets them as the strings 'true' or 'false'.
At Ruby, `if 'false'` returns truthy.

- testsuite/features/support/env.rb:
Check explicitely for either 'true' or 'false'.

### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/14044
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/14045
## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -18,12 +18,9 @@ require 'multi_test'
 
 server = ENV['SERVER']
 $debug_mode = true if ENV['DEBUG']
-$long_tests_enabled = true if ENV['LONG_TESTS']
+$long_tests_enabled = true if ENV['LONG_TESTS'] == 'true'
 puts "Executing long running tests" if $long_tests_enabled
-$service_pack_migration_enabled = true if ENV['SERVICE_PACK_MIGRATION']
-# To disable SP migration, comment the previous line, uncoment the next one
-# and adjust testsuite/run_sets/init_clients.yml
-# $service_pack_migration_enabled = false
+$service_pack_migration_enabled = true if ENV['SERVICE_PACK_MIGRATION'] == 'true'
 puts "Executing service pack migrations" if $service_pack_migration_enabled
 
 # maximal wait before giving up


testsuite/features/support/env.rb
```diff
@@ -18,9 +18,9 @@
 
 server = ENV['SERVER']
 $debug_mode = true if ENV['DEBUG']
```
Bischoff, line 20: we should probably use same startegy for `DEBUG` variable too then

(btw do we use that variable at all? if yes, is it documented?)

testsuite/features/support/env.rb
```diff
@@ -18,9 +18,9 @@
...
-$long_tests_enabled = true if ENV['LONG_TESTS']
+$long_tests_enabled = true if ENV['LONG_TESTS'] == 'true'
```
Bischoff, line 21: we should really get rid of that variable and replace it with 2 scopes:
* build images
* virtualization test

testsuite/features/support/env.rb
```diff
@@ -18,9 +18,9 @@
...
 puts "Executing long running tests" if $long_tests_enabled
-$service_pack_migration_enabled = true if ENV['SERVICE_PACK_MIGRATION']
+$service_pack_migration_enabled = true if ENV['SERVICE_PACK_MIGRATION'] == 'true'
 # To disable SP migration, comment the previous line, uncoment the next one
 # and adjust testsuite/run_sets/init_clients.yml
```
Bischoff, line 25: is it still needed to adjust that yaml file?

Bischoff (changes requested):
see small change requested and comments

testsuite/features/support/env.rb
calancha: It seems not.  Good catch.

testsuite/features/support/env.rb
calancha: I agree, but I could not find where we set `DEBUG`: it's not set by terracumber.
I prefer to keep the scope of this PR as it is: fix the vars that we set with terracumber, and we fail to check correctly.

testsuite/features/support/env.rb
calancha: Out of scope of this PR: this is a bug fix; a refactor is welcome, but please, in a next PR.

srbarrios:
Oups! I missed that one.
I'm fine with that approach, initial design was the same as with tags, if defined (not matters which value get) then is true.
But if we want to follow this design where we explicitely define a boolean value and we parse this value, then we should apply the same mechanism to every "switch" variable in our test framework.
For instance, DEBUG.
Last thing, did you checked if that change works when the variable is not defined? I would say it will, but it's worth to confirm.

calancha:
> Oups! I missed that one @calancha
> I'm fine with that approach, initial design was the same as with tags, if defined (not matters which value get) then is true.
> But if we want to follow this design where we explicitely define a boolean value and we parse this value, then we should apply the same mechanism to every "switch" variable in our test framework.
> For instance, DEBUG.
> Last thing, did you checked if that change works when the variable is not defined? I would say it will, but it's worth to confirm.

Out of scope. This PR was just a fix of something broken.
You are very welcome to revisit the design and extend it as needed.

srbarrios:
@calancha I'm sorry but this concern is not out of scope :)

Last thing, did you checked if that change works when the variable is not defined? I would say it will, but it's worth to confirm.

calancha:
I haven't checked it, but it's trivial: if it's not defined, then it will be nil, so my change is OK:
```
ENV['foo']
=> nil
ENV['foo'] == 'true'
=> false
```

PR #3310, opened 2021-02-19
Testsuite: Ensure we get the lastest Salt after SP migration

calancha:
## What does this PR change?
Ensure we use the latest Salt version available after the SP migration.

After the SP migration, the Salt version installed in the
SLE minion is the one from SCC; instead, we want to
test the latest Salt version available.

- testsuite/features/init_clients/sle_minion.feature
- testsuite/features/init_clients/sle_ssh_minion.feature
Install the latest Salt version in the SP migrated minion.
Combine the two steps checking the result of the SP migration in just one.
- testsuite/features/step_definitions/salt_steps.rb: Step definitions.

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14030
### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/14051
and https://github.com/SUSE/spacewalk/pull/14078
## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/init_clients/sle_minion.feature
+++ b/testsuite/features/init_clients/sle_minion.feature
@@ -56,17 +56,19 @@ Feature: Bootstrap a Salt minion via the GUI
     Then I should see a "This system is scheduled to be migrated to" text
 
 @service_pack_migration
-  Scenario: Check the migration status for this minion
+  Scenario: Check the migration is successful for this minion
     Given I am on the Systems overview page of this "sle_spack_migrated_minion"
     When I follow "Events"
     And I follow "History"
     And I wait at most 600 seconds until event "Service Pack Migration scheduled by admin" is completed
+    And I follow "Details" in the content area
+    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
 
 @service_pack_migration
-  Scenario: Check the migration is successful for this minion
-    Given I am on the Systems overview page of this "sle_spack_migrated_minion"
-    When I follow "Details" in the content area
-    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
+  Scenario: Install the latest Salt on this minion
+    When I enable repositories before installing Salt on this "sle_spack_migrated_minion"
+    And I install Salt packages from "sle_spack_migrated_minion"
+    And I disable repositories after installing Salt on this "sle_spack_migrated_minion"
 
   Scenario: Subscribe the SLES minion to a base channel
     Given I am on the Systems overview page of this "sle_minion"
--- a/testsuite/features/init_clients/sle_ssh_minion.feature
+++ b/testsuite/features/init_clients/sle_ssh_minion.feature
@@ -67,17 +67,20 @@ Feature: Bootstrap a Salt host managed via salt-ssh
     Then I should see a "This system is scheduled to be migrated to" text
 
 @service_pack_migration
-  Scenario: Check the migration status for this SSH minion
+  Scenario: Check the migration is successful for this SSH minion
     Given I am on the Systems overview page of this "ssh_spack_migrated_minion"
     When I follow "Events"
     And I follow "History"
     And I wait at most 600 seconds until event "Service Pack Migration scheduled by admin" is completed
+    And I follow "Details" in the content area
+    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
 
 @service_pack_migration
-  Scenario: Check the migration is successful for this SSH minion
-    Given I am on the Systems overview page of this "ssh_spack_migrated_minion"
-    When I follow "Details" in the content area
-    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
+@ssh_minion
+  Scenario: Install the latest Salt on this SSH-managed minion
+    When I enable repositories before installing Salt on this "ssh_spack_migrated_minion"
+    And I install Salt packages from "ssh_spack_migrated_minion"
+    And I disable repositories after installing Salt on this "ssh_spack_migrated_minion"
 
 @ssh_minion
   Scenario: Subscribe the SSH-managed SLES minion to a base channel
--- a/testsuite/features/step_definitions/salt_steps.rb
+++ b/testsuite/features/step_definitions/salt_steps.rb
@@ -659,6 +659,14 @@ When(/^I install Salt packages from "(.*?)"$/) do |host|
   end
 end
 
+When(/^I enable repositories before installing Salt on this "([^"]*)"$/) do |host|
+  step %(I enable repository "tools_additional_repo" on this "#{host}" without error control)
+end
+
+When(/^I disable repositories after installing Salt on this "([^"]*)"$/) do |host|
+  step %(I disable repository "tools_additional_repo" on this "#{host}" without error control)
+end
+
 # minion bootstrap steps
 Then(/^I run spacecmd listevents for "([^"]*)"$/) do |host|
   system_name = get_system_name(host)


testsuite/features/init_clients/sle_minion.feature
```diff
@@ -56,17 +56,19 @@ Feature: Bootstrap a Salt minion via the GUI
...
 
 @service_pack_migration
-  Scenario: Check the migration status for this minion
+  Scenario: Check the migration is successful for this minion
     Given I am on the Systems overview page of this "sle_spack_migrated_minion"
     When I follow "Events"
     And I follow "History"
     And I wait at most 600 seconds until event "Service Pack Migration scheduled by admin" is completed
+    And I follow "Details" in the content area
+    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
```
Bischoff, line 65: Cool verification

testsuite/features/init_clients/sle_minion.feature
```diff
@@ -56,17 +56,19 @@ Feature: Bootstrap a Salt minion via the GUI
...
 
 @service_pack_migration
-  Scenario: Check the migration is successful for this minion
-    Given I am on the Systems overview page of this "sle_spack_migrated_minion"
-    When I follow "Details" in the content area
-    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
+  Scenario: Install the latest Salt on this minion
+    And I enable repository "tools_additional_repo" on this "sle_spack_migrated_minion"
```
Bischoff, line 69: When

testsuite/features/init_clients/sle_ssh_minion.feature
```diff
@@ -67,17 +67,20 @@ Feature: Bootstrap a Salt host managed via salt-ssh
...
+    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
 
 @service_pack_migration
-  Scenario: Check the migration is successful for this SSH minion
-    Given I am on the Systems overview page of this "ssh_spack_migrated_minion"
-    When I follow "Details" in the content area
-    Then I should see a "SUSE Linux Enterprise Server 15 SP2" text
+@ssh_minion
+  Scenario: Install the latest Salt on this minion
+    And I enable repository "tools_additional_repo" on this "ssh_spack_migrated_minion"
```
Bischoff, line 81: When

Bischoff (changes requested):
LGTM, minor nitpicks

Bischoff:
Reformulating Tino's explanations, as I did to explain this thing to Oscar:

The salt we use in the test suite is "experimental" salt that has not been published to SCC yet.

When we do a service pack migration, we end up, like customers, with SCC repositories obtained via the SUSE Manager server.

This means we are not testing anymore with the "new" Salt for next version of SUSE Manager, but with old, already published, Salt, for previous version of SUSE Manager. And that's wrong.

srbarrios (approved):
LGTM.
Long term goal I think it would be to get rid of those sumaform injected repositories, and let SUMA handle the repositories. For that purpose, I suggest setting up the scc qa proxy.

Bischoff:
> LGTM.
> Long term goal I think it would be to get rid of those sumaform injected repositories, and let SUMA handle the repositories. For that purpose, I suggest setting up the scc qa proxy.

Yes. Or through the test suite.

But remember developers are also using sumaform to deploy, and they like to get automatically N+1 Salt that way.

So we should carefully consider such moves.

srbarrios (changes requested):
We update salt-minion. But... what about salt package?
```
suma-41-min-sles15:~ # rpm -qa | grep salt
salt-minion-3000-19.1.x86_64
python3-salt-3000-19.1.x86_64
salt-3000-19.1.x86_64
```

srbarrios (approved):

testsuite/features/init_clients/sle_minion.feature
```diff
@@ -56,17 +56,19 @@ Feature: Bootstrap a Salt minion via the GUI
...
+    And I run "zypper install -y salt-minion" on "sle_minion"
```
Bischoff, line 70: We have specialized steps to update packages. Hide details from feature please.

Also, we could explicitely update "salt" (it should be dragged automatically by dependency, but just to be on safe side").

testsuite/features/init_clients/sle_ssh_minion.feature
```diff
@@ -67,17 +67,20 @@ Feature: Bootstrap a Salt host managed via salt-ssh
...
+    And I run "zypper install -y salt-minion" on "sle_minion"
```
Bischoff, line 82: Use dedicated step.

Bischoff (changes requested):
More nitpicks

testsuite/features/init_clients/sle_minion.feature
calancha: Do you mean this?
```
When I enable repositories before installing Salt on this "sle_spack_migrated_minion"
And I install the latest Salt on this "sle_spack_migrated_minion"
And I disable repositories after installing Salt on this "sle_spack_migrated_minion"
```
 Otherwise, please clarify.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -755,6 +755,18 @@
   node.run("apt-get update")
 end
 
+When(/^I enable repositories before installing Salt on this "([^"]*)"$/) do |host|
+  step %(I enable repository "tools_additional_repo" on this "#{host}")
+end
+
+When(/^I install the latest Salt on this "([^"]*)"$/) do |host|
+  step %(I run "zypper install -y salt-minion" on "#{host}")
```
Bischoff, line 763: 1 - Why not `I install packages "..." on this "..."`

2 - Also  specify "salt" package

Bischoff (changes requested):
see remarks

testsuite/features/step_definitions/common_steps.rb
calancha: Actually, we have the following step in `salt_steps.rb`:
```
When(/^I install Salt packages from "(.*?)"$/) do |host|
  target = get_target(host)
  if %w[sle_minion ssh_minion sle_client sle_migrated_minion].include?(host)
    target.run("test -e /usr/bin/zypper && zypper --non-interactive install -y salt salt-minion", false)
.
.
.
```
We don't need to add a new step: we could use teh existent one.

Bischoff (approved):

testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -659,6 +659,14 @@ def pillar_get(key, minion)
   end
 end
 
+When(/^I enable repositories before installing Salt on this "([^"]*)"$/) do |host|
+  step %(I enable repository "tools_additional_repo" on this "#{host}")
```
meaksh, line 663: One question, shouldn't this be something like:

```suggestion
  step %(I enable repository "tools_additional_repo" on this "#{host}" without error control)
```

IIUC, without this, if `tools_additional_repo` repo is not defined in this node, the scenario will fail.

What do you think?

testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -659,6 +659,14 @@ def pillar_get(key, minion)
...
+end
+
+When(/^I disable repositories after installing Salt on this "([^"]*)"$/) do |host|
+  step %(I disable repository "tools_additional_repo" on this "#{host}")
```
meaksh, line 667: Same consideration as above

testsuite/features/step_definitions/salt_steps.rb
calancha: In such a case, I prefer it fails immediately: it helps diagnostic the reason for the failure.
Otherwise, it might or might not fail later; should not fail, we would use the wrong Salt without notice; should fail instead, 
it won't point straight to the final reason for this failure (i.e., a missing repository).

meaksh: The problem I see with this, is that you're making the definition of "additional_repo" in sumaform, an particularly a repository called `tools_additional_repo` as mandatory for the SLE minion in any environment/deployment where you want to be able to run the testsuite (since this scenario runs as part "init_client" stage).

I'm not sure we really want this, IMHO the testsuite should be able to run (at least reaching secondary steps) without having to define this particular "additional repo".

If we really want to make this repository mandatory for the testsuite, then I'd rather make it mandatory in "sumaform" and not be defined as an optional "additional repo".

What do you think? :)

srbarrios: @calancha please double-check they usages of this step, in case it can break other Cucumber features, for instance features running for QAM/BV.

calancha: > I'm not sure we really want this, IMHO the testsuite should be able to run (at least reaching secondary steps) without having to define this particular "additional repo".

OK.

testsuite/features/step_definitions/common_steps.rb
srbarrios: The word "from" on the current step, doesn't look coherent :thinking:

PR #3311, opened 2021-02-20
more links changed to buttons

mcalmer:
## What does this PR change?

Found more links which changed into buttons in PR https://github.com/uyuni-project/uyuni/pull/3282
I wonder if this was really a good idea to do it.

@ricardoasmarques @ncounter @Etheryte - I wonder if this was intended or we we better rollback this change?

See examples below

## GUI diff

The column header texts are not links anymore, but buttons.:

image

Here the right side actions `Attach/Detach Sources`, `Attach/Detach Filter`, etc are now also buttons:

image

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/build_validation/add_custom_repositories/add_ceos8_repositories.feature
+++ b/testsuite/features/build_validation/add_custom_repositories/add_ceos8_repositories.feature
@@ -75,18 +75,18 @@ Feature: Adding the CentOS 8 distribution custom repositories
     And I enter "no-appstream" as "label"
     And I click on "Create"
     Then I should see a "Content Lifecycle Project - Remove AppStream metadata" text
-    When I follow "Attach/Detach Sources"
+    When I click on "Attach/Detach Sources"
     And I select "RHEL8-Pool for x86_64" from "selectedBaseChannel"
     And I check "Custom Channel for CentOS 8 DVD"
     And I check "RES-AS-8-Updates for x86_64"
     And I click on "Save"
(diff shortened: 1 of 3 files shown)


ncounter (approved):
I still believe it was a good idea to do this change, we were misusing the `link` (AKA `<a href...>` DOM element) before, so it is a good thing to move to a `button` instead. Probably the not-very-good idea was merging without giving a thought about the cucumber testsuite and give it a full run before merging. I personally reviewed and approved the PR but I forgot about the testsuite. Sorry.

On one hand, it should be transparent from a test perspective if the code is using a link or a button actually, but I know we have workarounds/limitations/etc that we rely in the end on the HTML composition.

Just to make it clear: the UX is untouched because the link and the button they work and look the same here.

Anyway, I would not go for reverting, we have the change and it is a good thing. I could expect some more failure, but the fix should be straight forward, same as this PR. Keep me in the loop and feel free to assign the required fixes to me.

Etheryte:
I agree with Dario, these elements should've been buttons to begin with. We're fixing up old mistakes, sorry for causing trouble with the test suite.

mcalmer:
Ok, in this case we adapt the testsuite. I just wanted to be sure that this change was really intended.

PR #3316, opened 2021-02-22
improve fromdir with better mapping of URL to local files

mcalmer:
## What does this PR change?

To be able to use Uyuni with a local mirror we had to improve the conversion of http URLs to local Files.
This fixes two problems:
1. Ubuntu/Debian URLs which requires multiple subdirs to work. supporting apt-mirror standard config
2. map mirrorlist using query parameters into a path (centos case)

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Documentation issue was created: https://github.com/SUSE/spacewalk/issues/14071

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13719
Fixes https://github.com/SUSE/spacewalk/issues/13432
Tracks https://github.com/SUSE/spacewalk/pull/14126 https://github.com/SUSE/spacewalk/pull/14127

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/content/MgrSyncUtils.java
+++ b/java/code/src/com/redhat/rhn/manager/content/MgrSyncUtils.java
@@ -31,12 +31,8 @@ import org.apache.log4j.Logger;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.URL;
-import java.net.URLDecoder;
-import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


java/code/src/com/redhat/rhn/manager/content/MgrSyncUtils.java
```diff
@@ -202,22 +221,48 @@ public static URI urlToFSPath(String urlString, String name) {
             host = url.getHost();
             path = url.getPath();
 
+            // Case 1
             if ("localhost".equals(host)) {
                 return url.toURI();
             }
+            String qPath = Pattern.compile("&")
```
lucidd, line 228: not sure how much more overhead this pattern is but i would just use `Arrays.stream(url.getQuery().split("&"))` for something simple like this.

java/code/src/com/redhat/rhn/manager/content/MgrSyncUtils.java
```diff
@@ -202,22 +221,48 @@ public static URI urlToFSPath(String urlString, String name) {
...
+                    .splitAsStream(Optional.ofNullable(url.getQuery()).orElse(""))
+                    .filter(f -> f.contains("=")) // filter out possible auth tokens
+                    .map(s -> Arrays.copyOf(s.split("=", 2), 2))
```
lucidd, line 231: `copyOf` should not be needed here since the filter makes sure there is always a `=` and split will always return 2 elements with empty string in case there is nothing after/before `=`

lucidd: i think this should be equivalent.
```java
      Arrays.stream(url.getQuery().split("&"))
                .filter(p -> p.contains("="))
                .map(p ->
                    Arrays.stream(p.split("=", 2))
                        .map(s -> URLDecoder.decode(s, StandardCharsets.UTF_8))
                        .collect(joining("/"))
                )
                .sorted()
                .collect(Collectors.joining("/"));
 ```

lucidd:
i'm not sure if this is already done somewhere but we should make sure the resulting path is properly sandboxed to a directory and does not escape.

java/code/src/com/redhat/rhn/manager/content/MgrSyncUtils.java
mcalmer: Trying out

mcalmer: changed

lucidd (approved):

PR #3318, opened 2021-02-23
Move push_package feature from core to secondary

maximenoel8:
Move push_package feature from core to secondary

The changes have been tested

https://github.com/SUSE/spacewalk/issues/13559

- [x] No changelog needed

### DIFF
--- a/testsuite/features/secondary/srv_push_package.feature
+++ b/testsuite/features/secondary/srv_push_package.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2015-2019 SUSE LLC
+# Copyright (c) 2015-2021 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 Feature: Push a package with unset vendor
--- a/testsuite/features/secondary/srv_push_package.feature
+++ b/testsuite/features/secondary/srv_push_package.feature
@@ -9,10 +9,6 @@ Feature: Push a package with unset vendor
   Background:
     Given I am authorized as "admin" with password "admin"
 
-  Scenario: Download the SSL certificate
-    When I download the SSL certificate
-    And I make the SSL certificate available to zypper
-
   Scenario: Push a package with unset vendor
     When I push package "/root/subscription-tools-1.0-0.noarch.rpm" into "test_base_channel" channel
     Then I should see package "subscription-tools-1.0-0.noarch" in channel "Test Base Channel"
--- a/testsuite/features/secondary/srv_push_package.feature
+++ b/testsuite/features/secondary/srv_push_package.feature
@@ -9,6 +9,10 @@ Feature: Push a package with unset vendor
   Background:
     Given I am authorized as "admin" with password "admin"
 
+  Scenario: Download the SSL certificate
+    When I download the SSL certificate
+    And I make the SSL certificate available to zypper
+
   Scenario: Push a package with unset vendor
     When I push package "/root/subscription-tools-1.0-0.noarch.rpm" into "test_base_channel" channel
     Then I should see package "subscription-tools-1.0-0.noarch" in channel "Test Base Channel"
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -20,20 +20,6 @@ When(/^I mount as "([^"]+)" the ISO from "([^"]+)" in the server$/) do |mountpoi
   $server.run("mount -o loop #{iso_path} /srv/www/htdocs/#{mountpoint}", true, 500, 'root')
 end
 
-When(/^I download the SSL certificate$/) do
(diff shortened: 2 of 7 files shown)


ktsamis (approved):
Looks great :100:

maximenoel8:
@juliogonzalez Hi, you can ignore the error. There is no need of release note because it's a small testsuite change.

mcalmer:
@maximenoel8 you said "move", but the changes show only that you add that test to secondary* but not removed it from core.
This does not look correct. Would you please double check?

ktsamis (changes requested):
Indeed @maximenoel8 you need to remove from here https://github.com/uyuni-project/uyuni/blob/master/testsuite/run_sets/core.yml#L16 Sorry I didn't catch it, I saw the rename

ktsamis:
Removing the label until the changes are made

srbarrios:
@maximenoel8 double-check with @calancha if this feature has not any dependency and can go to the parallelizable set of features, or it needs to stick into the secondary.yml file, where we have the features that have dependencies with others, and must run in a concrete order.

maximenoel8:
> @maximenoel8 you said "move", but the changes show only that you add that test to secondary* but not removed it from core.
> This does not look correct. Would you please double check?

Good catch, indeed I forget to remove from core on this second PR.

maximenoel8:
> @maximenoel8 double-check with @calancha if this feature has not any dependency and can go to the parallelizable set of features, or it needs to stick into the secondary.yml file, where we have the features that have dependencies with others, and must run in a concrete order.

I check if the package was call in any other part of the code and seems to be a standalone.

calancha:
@maximenoel8 
- Please, move `srv_push_package.feature` in `secondary_parallelizable.yml`
   above; together with the rest of `srv_` features.
- Drop `srv_push_package.feature` also from:
```
virtualization.yml
sle-updates.yml
refhost.yml
```

- According with my checks, the following Scenario is not needed anymore:
```
  Scenario: Download the SSL certificate
    When  I download the SSL certificate
    And I make the SSL certificate available to zypper
```

I have seen that the SSL certificate is downloaded and the link created
when we run `sle_client.feature`.  Note that, we only refer to these two
steps in this feature.

Please, double-check it; if I am right, then delete such a scenario from
the feature and their step definitions from `common_steps.rb`.

Bischoff (changes requested):
Please apply Tino's changes from last comment.

Apart from that, LGTM. Thanks for that change, the more we can remove from core, the best.

maximenoel8:
Add the change to the pull requests. After some testing, the change are working fine !

ktsamis (approved):
We are doing 2 things now, moving the push_package feature to secondary and removing the "Download the SSL certificate" feature and implementation. For now this is fine if it's tested. Next time please 2 PRs with clear goals.

maximenoel8:
> We are doing 2 things now, moving the push_package feature to secondary and removing the "Download the SSL certificate" feature and implementation. For now this is fine if it's tested. Next time please 2 PRs with clear goals.

OK sorry. Download SSL was only use in push_package. That's why I did it in the same PR.

testsuite/features/secondary/srv_push_package.feature
```diff
@@ -9,10 +9,6 @@ Feature: Push a package with unset vendor
   Background:
     Given I am authorized as "admin" with password "admin"
 
-  Scenario: Download the SSL certificate
```
Bischoff, line 12: What was this scenario needed for?

Bischoff (changes requested):
I am okay to remove Scenario `Download the SSL certificate`, but what was it needed for, and why isn't it needed anymore? We would need some context here.

mcalmer:
The `Download the SSL certificate` needs a bit or research. Originally I added it to make rhnpush work.
rhnpush works only out of the box then it uses SSL. But to use SSL, the CA certificate store must be available.
The default config say, that it should look for it at `/usr/share/rhn/RHN-ORG-TRUSTED-SSL-CERT` which is available on every registered client. **But** we execute rhnpush on the "Server". This step was there to download the SSL certificate and make it available on the server at the expected path. See Commit fa3fc505a963edc104fa112e46b3b126ca8964b7

But it seems later this was missunderstand and Commit 2667a2c51cf71e2e017a992202550c04190bb3e8 changed it to execute it on the "client". But from my point of view this is not required on the client as it should be there already after a successful registration.

1. check, how rhnpush can be executed on the server using SSL? Where does the certificate come from?
2. Try to remove that step from the client. From my point of view this is not needed. At least not on the client.

maximenoel8:
I removed the SSL part from the PR and will create a card for that. Thanks for you feedback !

Bischoff (approved):

srbarrios (approved):

PR #3319, opened 2021-02-23
Migrate `manager/admin`

Etheryte:
## What does this PR change?

Migrate everything in `manager/admin`.  
There were some weird bugs in this one, I'll add comments to them to turn reviewer's attention to those places.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/manager/admin/setup/products/products-scc-dialog.tsx
```diff
@@ -0,0 +1,200 @@
+import * as React from "react";
+import Network from "utils/network";
+import { Messages, MessageType } from "components/messages";
+import { Button } from "components/buttons";
+
+const msgMap = {
```
Etheryte, line 6: This was originally missing in the source file, we had a discussion about this with Dario and our best guess is that it simply got lost over time in refactoring. My guess is that no one noticed because the network error branch was hit rarely enough or never.

web/html/src/components/table/TableDataHandler.tsx
```diff
@@ -341,10 +341,9 @@ export class TableDataHandler extends React.Component<Props, State> {
                   selectable={this.props.selectable}
                 >
                   {this.props.searchField}
-                  {this.props.additionalFilters &&
-                    this.props.additionalFilters.map((filter, i) => (
-                      <span key={"additional-filter-" + i}>{filter}&nbsp;</span>
-                    ))}
+                  {this.props.additionalFilters?.map((filter, i) => (
+                    <span key={"additional-filter-" + i}>{filter}&nbsp;</span>
```
Etheryte, line 345: There is a small bug here, created a separate ticket for it: https://github.com/SUSE/spacewalk/issues/14072

web/html/src/manager/admin/setup/products/products.js
```diff
@@ -1,972 +0,0 @@
...
-      return data.filter(p => this.state.archCriteria.includes(p.arch));
-    }
-    return data;
-  };
-
-  handleSelectedItems = (items) => {
-    this.props.handleSelectedItems(items);
-  };
-
-  handleUnselectedItems = (items) => {
```
Etheryte, line 431: This file had a mess with a mixup of `handleUnselectedItems` and `handleUnSelectedItems` (different capitalization of `s/S`), where both were referenced in some places. I did best I could do test it and fix it up as seemed correct, but would be good if you could also give this a test please.

web/html/src/manager/content-management/shared/api/use-lifecycle-actions-api.ts
```diff
@@ -49,8 +49,13 @@ const useLifecycleActionsApi = (props: Props): returnUseProjectActionsApi => {
       setIsLoading(true);
 
       const apiUrl = getApiUrl(props.resource, props.nestedResource, id);
-      const networkMethod: NetworkMethod = networkAction[action] || networkAction["get"];
-      const networkRequest = networkMethod(apiUrl, JSON.stringify(actionBodyRequest), "application/json");
```
Etheryte, line 53: There was a bug here, where GET requests would also be called with the following signature:  

```ts
get(url, payload, mimeType)
```

However, their actual signature is:

```ts
get(url, mimeType)
```

This only accidentally worked because GET requests have a default value for `contentType: string = "application/json"` (other requests don't have this default value).  
The below refacto aims to fix this problem.

web/html/src/manager/admin/task-engine-status/taskotop.tsx
```diff
@@ -0,0 +1,268 @@
...
+    const headerTabs = (
+      <div className="spacewalk-content-nav">
+        <ul className="nav nav-tabs">
+          <li>
+            <a className="js-spa" href="/rhn/admin/TaskStatus.do">
+              {t("Last Execution Times")}
+            </a>
+          </li>
+          <li className="active js-spa">
+            <a href="/rhn/manager/admin/runtime-status">{t("Runtime Status")}</a>
```
Etheryte, line 167: This link had an empty href (`href=""`) in the original source, based on other uses I _think_ this should be the right URL.

ncounter: Yes, it is the same link of the current page actually, so it makes sense to fill with it. Thanks for the catch

web/html/src/manager/content-management/shared/api/use-lifecycle-actions-api.ts
```diff
@@ -49,8 +49,13 @@ const useLifecycleActionsApi = (props: Props): returnUseProjectActionsApi => {
...
+
+      let networkRequest: ReturnType<NetworkMethod>;
+      if (action === "get" || !networkAction[action]) {
```
ncounter, line 54: Just to make sure we don't miss anything 100% : could it happen that action is capitalized instead (`GET` instead of `get`)? If so, I'd check for `action.toLowerCase() === "get"`, otherwise forget about this comment :)

ncounter (approved):

web/html/src/manager/content-management/shared/api/use-lifecycle-actions-api.ts
Etheryte: In that case we'd fall through to the second condition (`!networkAction["GET"] === true`), but just for sanity I add both that and a type check for the action parameter so that it only takes the correct keys. Good catch! 👍

PR #3334, opened 2021-02-25
Reposync fix (bsc#1182197)

vzhestkov:
## What does this PR change?

In case of exeption inside deb_src repo plugin it doesn't restore config files namespace back, causing reading of some config variables from improper namespace.
Also fixes improper exception handling in ThreadedDownloader class causing another exception.

## GUI diff

No difference.

## Documentation

## Links

Fixes [bsc#1182197](https://github.com/SUSE/spacewalk/issues/13992)  - spacewalk-repo-sync: AttributeError: AUTO_GENERATE_BOOTSTRAP_REPO

## Re-run a test


mcalmer (approved):

backend/satellite_tools/download.py
```diff
@@ -284,29 +284,33 @@ def __init__(self, retries=3, log_obj=None, force=False):
...
+                )
+            try:
+                self.minrate = int(CFG.REPOSYNC_MINRATE)
+            except ValueError:
+                raise ValueError(
+                    "Minimal transfer rate in bytes pre second expected, found: '%s'"
+                    % CFG.REPOSYNC_MINRATE
+                )
+        except Exception as e:
+            raise e from None
```
agraul, line 307: This is only here to silence "During handling of the above exception, another exception occurred", right?

I think it would be cleaner (didn't think about this when I added the other `try:...except:`s) not to handle anything here and do it like this:
```python
try:
    try:
        int("foo")
    except ValueError:
        raise ValueError("nice msg") from None
finally:
    print("initCFG call")
```
Which results in:
```python
ValueError                                Traceback (most recent call last)
<ipython-input-8-89bd5c21ec97> in <module>
      3         int("foo")
      4     except ValueError:
----> 5         raise ValueError("nice msg") from None
      6 finally:
      7     print("initCFG call")

ValueError: nice msg

File: backend/satellite_tools/repo_plugins/deb_src.py [hunk 2]
```diff
@@ -295,7 +296,8 @@ def __init__(self, url, name, insecure=False, interactive=True, yumsrc_conf=None
             self.timeout = int(CFG.REPOSYNC_TIMEOUT)
         except ValueError:
             self.timeout = 300
-        initCFG(comp)
+        finally:
```
agraul, line 299: I think this needs another `finally` for the first `try` clause

PR #3335, opened 2021-02-25
Homogenizes style in filter buttons, facilitating testability

srbarrios:
## What does this PR change?

**To find the filter button, we use the class name `spacewalk-button-filter` as we can see in this example:**

image

------
**But in this other case, we don't have this class name:**

image

This PR fix it, and facilitate testability.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Ports:
- Manager-4.1

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


Etheryte:
Looks good as a one-off, in the long term however, I would recommend targeting on something else rather than class name, a-la the closest submit button to the search bar or something similar. From development side, this class name doesn't hold any special significance so it's entirely possible it may change at some point.

Etheryte (approved):

srbarrios:
> Looks good as a one-off, in the long term however, I would recommend targeting on something else rather than class name, a-la the closest submit button to the search bar or something similar. From development side, this class name doesn't hold any special significance so it's entirely possible it may change at some point.

Yeah... relative locators will help a lot, sadly we still use an older version of Selenium and we don't support that kind of search.

PR #3340, opened 2021-03-01
Fix: increase password length in the database (bsc#1182687)

mbologna:
## What does this PR change?

This commit increases the password length for the susecredentials table.
Without this commit, the error of password length overflown in the
database field is happening when using a JWT Token (760 chars) as the
password.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

Tracks TBD

- [] **DONE**

## Re-run a test

### DIFF
--- a/schema/spacewalk/common/tables/suseCredentials.sql
+++ b/schema/spacewalk/common/tables/suseCredentials.sql
@@ -26,7 +26,7 @@ CREATE TABLE suseCredentials
                  REFERENCES suseCredentialsType (id),
     url      VARCHAR(256),
     username VARCHAR(64) NOT NULL,
-    password VARCHAR(64) NOT NULL,
+    password VARCHAR(4096) NOT NULL,
     created  TIMESTAMPTZ DEFAULT (current_timestamp) NOT NULL,
     modified TIMESTAMPTZ DEFAULT (current_timestamp) NOT NULL
 )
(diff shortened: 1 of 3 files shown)


schema/spacewalk/common/tables/suseCredentials.sql
```diff
@@ -26,7 +26,7 @@ CREATE TABLE suseCredentials
                  REFERENCES suseCredentialsType (id),
     url      VARCHAR(256),
     username VARCHAR(64) NOT NULL,
-    password VARCHAR(64) NOT NULL,
+    password VARCHAR(1031) NOT NULL,
```
mcalmer, line 29: Do not calculate the exact number of bytes. Next time it will be one more and we have the same problem again.
If it started to become normal to use passwords with more than 64 characters, just go for 2048 or 4096 directly.
It is varchar and should not need more space.

mcalmer (changes requested):

schema/spacewalk/common/tables/suseCredentials.sql
mbologna: You are right. Fixed

mcalmer (approved):

moio (approved):

PR #3345, opened 2021-03-01
Migrate `manager/maintenance`

Etheryte:
## What does this PR change?

Migrate everything in `manager/maintenance`.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/manager/maintenance/api/maintenance-windows-api.ts
```diff
@@ -0,0 +1,31 @@
+import Network from "utils/network";
+
+/* Returns a list of maintenance schedules or calendars.
+ * type == "schedule" for schedules
+ * type == "calendar" for calendars
+ */
+function list(type: "schedule" | "calendar") {
+  const endpoint = "/rhn/manager/api/maintenance/" + type + "/list";
+  return Network.get(endpoint, "application/json").promise.then(schedules => schedules);
```
ncounter, line 9: To be picky, probably not the best names for the `.then(schedules => schedules)` as it seems it can be either `schedules` **or** `calendars`, so I could think about a renaming for a more generic name (like following methods do): `list` or `result` could be valid options to me. Anyway, not part of this PR so feel free to ignore it.

Etheryte: Thanks for highlighting this, the point about names is good but even more so that bit is redundant here and below as well. `return Promise(...).then(a => a)` is equivalent to `return Promise(...)`. Good catch!

ncounter (approved):

PR #3360, opened 2021-03-03
Wait for reposync, including in Build Validation

Bischoff:
## What does this PR change?

This PR waits for all reposyncs to finish, even when we are in Build Validation case

## Links

Solves SUSE/spacewalk#14055

Ports:
* 4.0: SUSE/spacewalk#14176
* 4.1: SUSE/spacewalk#14175

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/reposync/srv_wait_for_reposync.feature
+++ b/testsuite/features/reposync/srv_wait_for_reposync.feature
@@ -1,8 +1,9 @@
-# Copyright (c) 2019 SUSE LLC
+# Copyright (c) 2019-2021 SUSE LLC
 # Licensed under the terms of the MIT license.
 
-Feature: Abort all reposync activity
+Feature: Wait for reposync activity to finish
 
+@regular_ci
   Scenario: Delete scheduled reposyncs
     Given I am authorized as "admin" with password "admin"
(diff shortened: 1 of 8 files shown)


testsuite/features/reposync/srv_wait_for_reposync.feature
```diff
@@ -1,8 +1,9 @@
-# Copyright (c) 2019 SUSE LLC
+# Copyright (c) 2019-2021 SUSE LLC
 # Licensed under the terms of the MIT license.
 
-Feature: Abort all reposync activity
+Feature: Wait for reposync activity to finish
 
+@regular_ci
```
srbarrios, line 6: Maybe the word "ci_tests" instead? or just "ci"?

srbarrios (approved):
LGTM, just a suggestion with regular_ci tag, but I'm fine to keep what it is.

testsuite/features/reposync/srv_wait_for_reposync.feature
Bischoff: Maybe `continuous_integration then

We could also avoid using new tags, and just use a different feature file. But it seemed to me more consistent that way.

calancha (approved):

PR #3361, opened 2021-03-04
QA: Hold the browser session and page between scenarios

srbarrios:
## What does this PR change?

Hold the browser session (and page) between Cucumber scenarios, but not between Cucumber Features.

Bonus:
This also fixes a problem in XEN/KVM tests, that were opening a graphical console in full-screen mode in a new tab and not taking into account to close the tab, coming back to the previous page before end the scenario. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.0 https://github.com/SUSE/spacewalk/pull/14188
- Manager-4.1 https://github.com/SUSE/spacewalk/pull/14187

- [x] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #3370, opened 2021-03-07
Revert "QA: Hold the browser session and page between scenarios"

srbarrios:
Reverts uyuni-project/uyuni#3361

I tried it first, and apparently, it was good, but I was wrong.
I will need to think deeper about this change. It's very disruptive.

Revert this before https://github.com/uyuni-project/uyuni/pull/3371, please

## Changelogs

- [x] No changelog needed


srbarrios:
Revert this before https://github.com/uyuni-project/uyuni/pull/3371, please

PR #3376, opened 2021-03-08
Migrate `manager/state`

Etheryte:
## What does this PR change?

Migrate everything in `manager/state`.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/manager/state/recurring-states.tsx
```diff
@@ -0,0 +1,254 @@
...
+import { Messages } from "components/messages";
+import Network from "utils/network";
+import { RecurringStatesDetails } from "./recurring-states-details";
+import { RecurringStatesList } from "./recurring-states-list";
+import { RecurringStatesEdit } from "./recurring-states-edit";
+import { Utils as MessagesUtils } from "components/messages";
+import SpaRenderer from "core/spa/spa-renderer";
+
+/**
+ * See:
```
Etheryte, line 11: If I'm not mistaken, these are all the files that may render this view and where we might inherit global variables from. Please correct me if I'm wrong.

hustodemon (approved):
LGTM :+1:  just left an affirmative comment. Thanks!

web/html/src/manager/state/recurring-states.tsx
Etheryte: Thanks a lot for pointing that out, I've added that to the list. :)

PR #3380, opened 2021-03-09
When using the clone-by-date tool, the modules.yaml of the target channel is not aligned to the source.

hustodemon:
Not aligning the modules.yaml in this situation is wrong, since the source modules.yaml of the source can get updated, whereas the modules.yaml of the target still stays the same and might get incorrect for the new state of the channel.

Aligning the modules.yaml (this PR) is also wrong: if the clone-by-date tool is executed with a date "far in the past", then the modules.yaml of the source might be incorrect for the new state of the clone, since it contains references to packages that were filtered out when cloning.

While both approaches are wrokng, I believe aligning the metadata is "less wrong".

In general, this problem might be insolvable, since we don't control the generating of the modules.yaml.

## TODO
- [x] Check the DNF behavior of cloned channel created by `spacewalk-clone-by-date` (with a date far in the past) and new `modules.yaml`: it seems DNF displays the modules according to new `modules.yaml` (as expected), but installing a new module that is not present in the cloned channel fails.

## Documentation
Maybe instead of this fix, we only need to update the documentation.

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/14139

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/channel/software/ChannelSoftwareHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/channel/software/ChannelSoftwareHandler.java
@@ -32,9 +32,9 @@ import com.redhat.rhn.domain.channel.ChannelFactory;
 import com.redhat.rhn.domain.channel.ContentSource;
 import com.redhat.rhn.domain.channel.ContentSourceFilter;
 import com.redhat.rhn.domain.channel.InvalidChannelRoleException;
+import com.redhat.rhn.domain.errata.ClonedErrata;
 import com.redhat.rhn.domain.errata.Errata;
 import com.redhat.rhn.domain.errata.ErrataFactory;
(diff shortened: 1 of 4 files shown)


hustodemon:
> Is there something this tool can do what cannot be done with CLM and API?

AFAIK, the tool can solve dependencies. Moreover, CLM will have the same problem as spacewalk-clone-by-date - we don't control the `modules.yaml` content, so any tool that somehow modifies the channel packages can hit it.

cbbayburt (approved):
Looks good :+1: 

Messy stuff indeed

hustodemon:
Sorry for the delay here. I'm asking the user affected by this, if this fixed their problems. I'll preferably merge only after getting positive feedback.

PR #3382, opened 2021-03-09
Avoid race condition due multiple reposync import threads (bsc#1183151)

meaksh:
## What does this PR change?

This PR fixes a race condition that might happen at the time of running `spacewalk-repo-sync`, particularly at the time of importing the packages. This race condition make `spacewalk-repo-sync` to crash.

Importing the packages during reposync is done by multiple threads, and it seems there might be the case where a race condition happen at the time of cleaning some directories:

```console
File "/usr/lib64/python3.6/multiprocessing/pool.py", line 119, in worker
result = (True, func(*args, **kwds))
File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/reposync.py", line 1285, in import_package_batch
raise exc
File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/reposync.py", line 1280, in import_package_batch
os.rmdir(os.path.dirname(stage_path))
FileNotFoundError: [Errno 2] No such file or directory: '/var/spacewalk/packages/1/stage/81ca6528cad433169ea96ad3ff82811b02ef16be95740c22df3511777fb28a0f'
```

The code from `import_package_batch` fuction runs in different threads and causes the race condition in the following part:

https://github.com/uyuni-project/uyuni/blob/f88add35dc97aa0f6dc16a94fb476c99dbaf5057/backend/satellite_tools/reposync.py#L1297-L1305

As you can see, we check if the directory exists and after that we try to remove it but the directory is already gone.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **bugfix**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/14212

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/backend/spacewalk-backend.changes
+++ b/backend/spacewalk-backend.changes
@@ -1,3 +1,5 @@
+- Avoid race condition due multiple reposync import threads (bsc#1183151)
+
 -------------------------------------------------------------------
 Fri Mar 05 15:41:07 CET 2021 - jgonzalez@suse.com
 


agraul (approved):

moio (approved):
LGTM! thanks

meaksh:
>  The changelog has not been updated. Please update it @meaksh.

What? I do pushed a changelog entry... @juliogonzalez do you know about it? is this bot producing false possitives?

agraul:
I think @brejoc had a similar experience with that bot lately

PR #3397, opened 2021-03-12
QA: Speed up select step definition

srbarrios:
## What does this PR change?

Speed up select step definition, by using a different method to identify if we have to React component or not, that will not wait default timeout of 10 seconds.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Ports:
- Manager-4.0
- Manager-4.1

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -148,7 +148,7 @@ end
 
 When(/^I select "([^"]*)" from "([^"]*)"$/) do |arg1, arg2|
   xpath = "//input[@id='#{arg2}']/../../../../.."
-  if has_xpath?(xpath)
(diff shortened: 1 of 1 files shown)


srbarrios:
I tested in local ;)

ktsamis (approved):

srbarrios:
Ports pending because it seems we have a previous pending port, that collide with this PR.

srbarrios:
> Ports pending because it seems we have a previous pending port, that collide with this PR.

After discussing it with @cbosdo I should not port this change, as 4.1, 4.0 will never have Select React component.

PR #3398, opened 2021-03-12
BV: Create the bootstrap repositories with custom channels

Bischoff:
@moio @juliogonzalez @mcalmer please review only the first commit (the short one, that removes existing stuff).

## What does this PR change?

This PR removes the creation of bootstrap repositories from normal CI test suite
* in most cases, we are not synchronizing the repositories used to boot,
   so it is not possible to create the corresponding bootstrap repositories
* it had every chance to fail, as it was not specifying the parent channels (failures were ignored)
* it was not called anyway from proxy, traditional client, SSH minion, CentOS, nor Ubuntu

This PR also adds the creation of bootstrap repositories to BV test suite
* in that test suite, we use Maintenance Incidents as custom channels;
   we must therefore make sure that the bootstrap repositories use these custom channels,
   using `--with-custom-channels` option.

## Links

Ports:
* 4.0: SUSE/spacewalk#14279
* 4.1: SUSE/spacewalk#14278

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/init_clients/sle12sp4_client.feature
+++ b/testsuite/features/build_validation/init_clients/sle12sp4_client.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2015 SUSE LLC
+# Copyright (c) 2015-2021 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 @sle12sp4_client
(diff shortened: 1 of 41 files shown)


mcalmer (approved):
I think we never required a bootstrap repo. Simply because it took to long to mirror everything what we need. So in normal devel environment we assigned the channels directly to the clients.

srbarrios (approved):

testsuite/features/build_validation/init_clients/sle12sp4_client.feature
```diff
@@ -7,6 +7,10 @@ Feature: Bootstrap a SLES 12 SP4 traditional client
   Scenario: Clean up sumaform leftovers on a SLES 12 SP4 traditional client
     When I perform a full salt minion cleanup on "sle12sp4_client"
 
+  Scenario: Create the bootstrap repository for a SLES 12 SP4 traditional client
+    Given I am authorized
+    When I create the bootstrap repository for "sles12sp4_client" on the server
```
srbarrios, line 12: Extra 's'.
Needs to be sle12sp4_client

PR #3413, opened 2021-03-13
delete scenario which delete product package from package state

mcalmer:
## What does this PR change?

This scenario should not be needed since round about a year now.
Adding the products to the package state was dropped in https://github.com/uyuni-project/uyuni/pull/2158

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/pull/14287

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/init_clients/sle_ssh_minion.feature
+++ b/testsuite/features/init_clients/sle_ssh_minion.feature
@@ -18,26 +18,6 @@ Feature: Bootstrap a Salt host managed via salt-ssh
     And I wait until I see the name of "ssh_minion", refreshing the page
     And I wait until onboarding is completed for "ssh_minion"
 
-# HACK
-# Package 'sle-manager-tools-release' is automatically installed during bootstrap and
-# stays installed after removal of channel containing it. So it is not possible to update it.
-# Package needs to be removed from highstate to avoid failure when updating it.
-@skip_service_pack_migration
-@ssh_minion
-  Scenario: Remove sle-manager-tools-release from state after bootstrap
-    Given I am on the Systems overview page of this "ssh_minion"
-    When I wait until I see "States" text
-    And I follow "States" in the content area
-    And I wait until I see "Highstate" text
-    And I follow "Packages" in the content area
-    Then I should see a "Package States" text
-    When I follow "Search" in the content area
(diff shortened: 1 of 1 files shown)


srbarrios:
@mcalmer could we please don't merge not urgent PRs during the weekend?

If you would wait a bit, I would tell you that we are also removing this package in our automated MU and milestones validations, I would ask you to explain me with more detail the reason why we don't need to remove this package, and I would ask you a doubt I have with sles12sp4, which has two packages with this name, one of them finishing in -POOL.

mcalmer:
> @mcalmer thanks for taking care of CI, but please, try to wait a bit more for non urgent PRs, it's hard to keep track of PRs created and merged during the weekend.

A not full test run over the weekend is imporant for me. Sorry.
 
> If I would saw it before, I would tell you that we are also removing this package in our automated MU and milestones validations, and I would ask you to keep both testsuites synch on that kind of changes, I would ask you to explain me with more detail the reason why we don't need to remove this package, and I would ask you a doubt I have with sles12sp4, which has two packages with this name, one of them finishing in -POOL.

As we can now see, there is also another test we need to check.

  `Remove sle-manager-tools-release from state after bootstrap via XML-RPC`

This is the reason why I added you all as reviewer. We can get all this removed at all places where it happens.
I will try to come up with another PR.

About the reason why this is not required anymore.

In the past (over a year ago) we detected the products during bootstrapping and added the release packages to the Package State to get them installed. This did not work so well as it required that the channels were added before this and this could not be guaranteed as it was an async task.

We implemented a different mechnism and wrote a salt module which is always part of the channel state and add the required product packages automatically. At that time the Package State was not required anymore. Some weeks later, we removed the code from bootstrapping which generate the package state.

This means you can wait as long as you want, you will never see `suse-manager-tools-release` package in the Package State.
I am not sure why the state was working before. Maybe we have something like: "it is ok when the package is not found" ?

Important: the package is still installed, but not via a Package State. When there are tests which remove the package, they might still be required. But cleaning up the Package State is not required anymore.

PR #3416, opened 2021-03-15
Bootstrap repositories must be generated sequentially

Bischoff:
## What does this PR change?

See SUSE/susemanager-ci#202 (Add new stage for BV create bootstrap repos)

Move the creation of bootstrap repos away from parallel stages, as `mgr-create-bootstrap-repo` cannot run in parallel.

## Links

Ports:
* 4.0: SUSE/spacewalk#14300
* 4.1: SUSE/spacewalk#14301

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/create_bootstrap_repositories/srv_create_bootstrap_repositories.feature
+++ b/testsuite/features/build_validation/create_bootstrap_repositories/srv_create_bootstrap_repositories.feature
@@ -1,0 +1,136 @@
+# Copyright (c) 2021 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Create bootstrap repositories
+
+@proxy
+  Scenario: Create the bootstrap repository for the SUSE Manager proxy
+    When I create the bootstrap repository for "proxy" on the server
+
+@sle11sp4_client
+  Scenario: Create the bootstrap repository for a SLES 11 SP4 traditional client
+    When I create the bootstrap repository for "sle11sp4_client" on the server
+
+@sle11sp4_minion
+  Scenario: Create the bootstrap repository for a SLES 11 SP4 minion
+    When I create the bootstrap repository for "sle11sp4_minion" on the server
+
+@sle11sp4_ssh_minion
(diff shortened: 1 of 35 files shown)


ktsamis (approved):

srbarrios:
@moio how difficult is that mgr-create-bootstrap-repo command can run multiple times and process multiple products in parallel?

moio:
I do not know the tool internals, but I am surprised to learn it can't run in parallel. Maybe @mcalmer knows?

mcalmer:
Why do you want this? Please explain the use case for parallel run. How do you call it? What is the trigger to start a run?

srbarrios:
When we prepare our Build Validation environment, we first synchronize all the products, once we confirm this, we create and sync custom channels that will include the repositories from IBS, once all of that it's done, we need to run the bootstrap repository creation for all our supported clients, so all those that we have synced.
We were doing it just before we bootstrap each kind of client, all our bootstraps are made in parallel. Then we discovered this is not technically possible, as this command can't run in parallel, so I wonder if it could be feasible to refactor this command to allow parallelization, or it's worth it just run in sequentially (keeping in mind that we need to gain as much speed as we can in our setup).

mcalmer:
The main problem is, that a bootstrap repo can be created by multiple product combinations and they are multi arch.
Running 2 processes which generate the same destination path at the same time can cause trouble.

I think if somebody would re-design the algorithms and maybe also the resulting bootstrap repos such a thing would be possible.
But I think we do not have the resources to do it.
I think it is not a few hours task.

srbarrios:
> The main problem is, that a bootstrap repo can be created by multiple product combinations and they are multi arch.
> Running 2 processes which generate the same destination path at the same time can cause trouble.
> 
> I think if somebody would re-design the algorithms and maybe also the resulting bootstrap repos such a thing would be possible.
> But I think we do not have the resources to do it.
> I think it is not a few hours task.

Ok! I was just curious, let's not worry more about it then. We will run sequentially.

PR #3420, opened 2021-03-16
BV: Add new run sets for waiting for reposyncs

Bischoff:
## What does this PR change?

This PR adds new run sets for waiting for all reposyncs.

It also does some cleaning in the run sets comments.

## Links

* 4.0: SUSE/spacewalk#14316
* 4.1: SUSE/spacewalk#14317

## Changelogs

- [x] No changelog needed


ktsamis (approved):

srbarrios (approved):
It should be good.
I'm just worried about the moment that we create custom channels and activation keys, just after we call the sync of the products, as we must have the product related to each AK or custom channel, available in the dropdown list at the moment we create those, if we do that too fast, as we don't wait for it, it could cause troubles.
But... ok, let's give a try, why not.

Bischoff:
> It should be good.
> I'm just worried about the moment that we create custom channels and activation keys, just after we call the sync of the products, as we must have the product related to each AK or custom channel, available in the dropdown list at the moment we create those, if we do that too fast, as we don't wait for it, it could cause troubles.
> But... ok, let's give a try, why not.

Modified the PR according to your remarks. We're back to 2 waits.

PR #3428, opened 2021-03-17
Testsuite: Drop some machine aliases mentions

calancha:
## What does this PR change?

Keep the aliases only in those contexts where they are needed.

### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/14327
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/14328

## Changelogs

- [x] No changelog needed


srbarrios (approved):
Make sense

PR #3435, opened 2021-03-18
Close the <p> tags in several channel descriptions

calancha:
## What does this PR change?

Close the `<p>` tags in several channel descriptions.
Otherwise, the HTML tags are visible:

close-html-tag-Screenshot from 2021-03-18 11-35-54

## Links
Found during manual test cycle for 4.2.0 Beta 1.
https://github.com/SUSE/spacewalk/issues/12722

### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/14348
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/14347

## Changelogs

- [x] No changelog needed

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/content/test/data1/productsUnscoped.json
+++ b/java/code/src/com/redhat/rhn/manager/content/test/data1/productsUnscoped.json
@@ -527,7 +527,7 @@
         "product_class": "SLE-LP",
         "cpe": "cpe:/o:suse:sle-live-patching:12",
         "free": false,
-        "description": "<p> SUSE Linux Enterprise Live Patching provides packages to update critical kernel modules live in SUSE Linux Enterprise. With SUSE Linux Enterprise Live Patching, you can perform critical kernel patching without shutting down your system, reducing the need for planned downtime and increasing service availability. <p>",
+        "description": "<p> SUSE Linux Enterprise Live Patching provides packages to update critical kernel modules live in SUSE Linux Enterprise. With SUSE Linux Enterprise Live Patching, you can perform critical kernel patching without shutting down your system, reducing the need for planned downtime and increasing service availability. </p>",
         "release_stage": "released",
         "eula_url": "https://updates.suse.com/SUSE/Products/SLE-Live-Patching/12/x86_64/product.license/",
         "product_type": "extension",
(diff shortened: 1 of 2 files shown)


lucidd:
this is just some data use in test but has no effect on production systems. The real data comes from scc and the test data is just copied from there. If i search scc data i find a lot of those cases.

grep -rn '<p>"' unscoped.json | wc -l
69

In general i'm not a big fan of embedding markup in the description since the description may be presented in multiple places like web but also terminal where the html tags are noise. In addition its also questionable from a security point of view since we need to make sure this is properly sanitized to not allow arbitrary html i.e scripts to be embedded. So if we tackle those html tags i would suggest we do it scc and push towards removing them completely.

calancha:
> In addition its also questionable from a security point of view since we need to make sure this is properly sanitized to not allow arbitrary html i.e scripts to be embedded. So if we tackle those html tags i would suggest we do it scc and push towards removing them completely.

Your point about security is a good one to drop those tags.
I am not familiar with doing scc changes.  Can you do that instead of me?

lucidd:
i just remembered that i complained about those tags a long time ago and was talking to @mcalmer  about it i think the outcome was that the description is actually defined in obs and that its hard to make everybody change their product definitions of old products. Maybe we could still post process those strings on our side and remove those tags.

mcalmer:
The product description is in the build service and SCC reads it from there and we read it from SCC.
As the product is "frozen" after it is released, it is a bigger task to unlock, change, lock and re-sync.
Nobody like that and we just accept it :-)

Etheryte:
This discussion has come up a number of times, but I can't find a link to the older ones at the moment. The problem isn't as simple as removing the tags, for example what's the expected output for lists with items (`ul li`/`ol li`), what about images, etc. I don't know to what extent any products actually use any fancy markup, but even just for paragraphs simply omitting them could leave to incorrect results. If we want to address this then in my opinion the best option we have is to have a tag whitelist and strip any markup that isn't in the list (including properties etc). All of this is far from a trivial piece of code, and would also need to be duplicated in frontend code too, because we're essentially talking about injecting arbitrary code from an external source.

calancha:
Let's better continue the discussion in the existing open issue and bug report. 
https://github.com/SUSE/spacewalk/issues/9280
https://bugzilla.suse.com/show_bug.cgi?id=1143633

PR #3436, opened 2021-03-18
Do a zypper ref before installing a pattern

Bischoff:
## What does this PR change?

This PR makes sure to do a `zypper ref` before installing (or removing) a pattern.

Technical notes:
 * we need to call `run()` twice because of different handling of error codes
 * we don't need to test for existence of `/usr/bin/zypper` as `zypper` will fail if it's not available

## Links

Ports:
* 4.0: SUSE/spacewalk#14350
* 4.1: SUSE/spacewalk#14349

## Changelogs

- [x] No changelog needed


lkotek (approved):
LGTM :-)

PR #3482, opened 2021-03-29
Highstate summary page

cbbayburt:
Add a highstate summary section with a list of states and config channels to highstate pages in system details, system groups, and SSM.

## GUI diff
state-summary

## Documentation
- Documentation issue was created: https://github.com/SUSE/spacewalk/issues/14490

## Test coverage
- Unit tests were added

## Links
Fixes https://github.com/SUSE/spacewalk/issues/14233

## Changelogs

- [ ] No changelog needed

## Re-run a test


cbbayburt:
@Etheryte TS complains about `server.mockGetJson()` in the test file so I kept it as JS for now.

web/html/src/manager/state/display-highstate.tsx
```diff
@@ -1,99 +1,37 @@
...
-  };
-
-  render() {
-    return (
-      <div className="panel panel-default">
-        <div className="panel-heading" onClick={this.expand} style={{ cursor: "pointer" }}>
-          <span>{this.props.data.name}</span>
+  return (
+    <div className="panel panel-default" style={{ marginBottom: 10 }}>
+      <div className="panel-heading" onClick={() => { setShow(!show) }} style={{ cursor: "pointer" }}>
```
Etheryte, line 20: This was already here previously, but a small nitpick since we're working on it: if this is something that can be clicked on to change state, it should probably be a `button`.

web/html/src/manager/state/highstate-summary.tsx
```diff
@@ -0,0 +1,137 @@
...
+    <div className="row">
+      <pre>{highstate}</pre>
+    </div >
+  );
+}
+
+function State({ minionId, state }: { minionId: number, state: StateSource }) {
+  if (state.type === "STATE") {
+    return (
+      <>
```
Etheryte, line 82: I'm not sure, but there might be a small issue here: there's multiple children but no keys, I would expect there to be a warning in development mode? But perhaps I misunderstand how the pieces fit together here.

Etheryte (approved):
Two small questions, other than that the frontend parts look very nice. Also, good job on the unit tests. 👍

web/html/src/manager/state/display-highstate.tsx
cbbayburt: It's technically an accordion and has some specific styling from bootstrap via `panel-heading` so I'm not sure if it's a good idea

web/html/src/manager/state/highstate-summary.tsx
cbbayburt: In any case, only one of the fragments is returned so it shouldn't be a problem. There are no "key" warnings in the dev console either.

web/html/src/manager/state/display-highstate.tsx
cbbayburt: A better implementation would be to use native accordions from bootstrap all across, but I think it's too much for this PR.

Etheryte: Fair enough, just something that caught my eye.

mcalmer (approved):

cbbayburt:
@Etheryte can I have another review for the last commit (to make the table sortable as per feedback)

Etheryte:
Looks good to me, nice. 👍

Etheryte (approved):

ncounter (approved):
Thank you very much for the implementation of this.
You made it easy and clean. 🤘

PR #3483, opened 2021-03-29
Reinstall salt broker after cleanup on proxy

Bischoff:
## What does this PR change?

Removing file `/etc/salt/broker` may break transmission of Salt protocol between server and minions.

This PR reinstalls `spacewalk-proxy-salt` package after the salt cleanup in order to recreate that file.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/init_clients/proxy.feature
+++ b/testsuite/features/build_validation/init_clients/proxy.feature
@@ -11,7 +11,11 @@ Feature: Setup SUSE Manager proxy
   I want to register the proxy to the server and configure it also as branch server
 
   Scenario: Clean up sumaform leftovers on a SUSE Manager proxy
-    When I perform a full salt minion cleanup on "proxy"
+    When I remove package "spacewalk-proxy-salt" from this "proxy" without error control
+    And I perform a full salt minion cleanup on "proxy"
+
+  Scenario: Install proxy pattern on the proxy
+    When I install pattern "suma_proxy" on this "proxy"
     # WORKAROUND to set proper product when JeOS image for SLE15SP2 is used
     And I set correct product for "proxy"
     # End of WORKAROUND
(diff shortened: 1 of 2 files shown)


ktsamis (approved):

srbarrios:
uhmm what about not removing the broker file?

Bischoff:
> uhmm what about not removing the broker file? It will make the trick?

Yes, but how ? You do `rm -r /etc/salt`, that removes the broker file `/etc/salt/broker`

Of course we could move it to `/tmp` before cleanup and back afterwards, but:
* it would be hacky
* what guarantee do we have there are no others like that?
* there's a problem of logic here... do we want to remove sumaform leftovers or not ?

I think that, to test like customers, it's better to install the proxy patterns ourselves. We already do that for the retail pattern anyway. Yes, that means undoing some work from sumaform.

srbarrios:
> > uhmm what about not removing the broker file? It will make the trick?
> 
> Yes, but how ? You do `rm -r /etc/salt`, that removes the broker file `/etc/salt/broker`
> 
> Of course we could move it to `/tmp` before cleanup and back afterwards, but:
> 
> * it would be hacky
> * what guarantee do we have there are no others like that?
> * there's a problem of logic here... do we want to remove sumaform leftovers or not ?
> 
> I think that, to test like customers, it's better to install the proxy pattern ourselves. We already do that for the retail pattern a few lines below anyway. Yes, that means undoing some work from sumaform.

I agree. With a only 'but', I would like that we are homogeneus in this kind of steps in CI and BV, so if you succeed with it, let's port it to CI too.

Bischoff:
> > I think that, to test like customers, it's better to install the proxy pattern ourselves. We already do that for the retail pattern a few lines below anyway. Yes, that means undoing some work from sumaform.
> 
> I agree. With a only 'but', I would like that we are homogeneus in this kind of steps in CI and BV, so if you succeed with it, let's port it to CI too.

Why not. Will do that tomorrow.

Bischoff:
Closing, I will move it to a private branch, I did not notice I was on Uyuni directly.

PR #3484, opened 2021-03-29
Set up Java logging with level TRACE in testsuite

agraul:
## What does this PR change?
Add Java logging (TRACE) in the testsuite

## GUI diff

No difference.

Before:

After:

- [X] **DONE**

## Documentation
- No documentation needed: just changing the testsuite settings

- [X] **DONE**

## Test coverage
No code changes

- [ ] **DONE**

## Links

Needed for debugging https://github.com/SUSE/spacewalk/issues/14057

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -383,6 +383,19 @@
   end
 end
 
+When(/^I increase Java log level to "([^"]*)" on server$/) do |log_level|
+  $server.run(
+    "echo -e >>/srv/tomcat/webapps/rhn/WEB-INF/classes/log4j.properties << CONFIG
```
Bischoff, line 388: no `-e` needed if you use `<<`

maybe `<<CONFIG` with no space (???)

testsuite/features/core/first_settings.feature
```diff
@@ -52,6 +52,10 @@ Feature: Very first settings
   Scenario: Wait for refresh of list of products to finish
     When I wait until mgr-sync refresh is finished
 
+  Scenario: Set up Java logging with level TRACE
+    When I increase Java log level to "trace" on "server"
```
Bischoff, line 56: no quotes around `"server"`

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -383,6 +383,19 @@
...
+log4j.logger.com.suse.manager.reactor=#{log_level.upcase}
+log4j.logger.com.suse.manager.webui.services.impl=#{log_level.upcase}
+CONFIG"
+  )
+end
+
+And(/^I restart the tomcat service$/) do
```
Bischoff, line 395: `When`

Bischoff (changes requested):
see remarks - one of them whould have prevented the thing from working - please test

testsuite/features/step_definitions/command_steps.rb
agraul: Tbh this is the most confusing part since I could find step implementations for both `When` and `And`, but I think they work the same. I'll change it to `When`, just wondered if you have a policy around that

testsuite/features/step_definitions/command_steps.rb
agraul: right, I "tested" this in a Ruby shell but without the space.... And you're right of course about the `-e`, no need to interpret backslashes if I don't have them in the string :D

testsuite/features/step_definitions/command_steps.rb
Bischoff: Yes, it works the same. Yes there's a policvy. `And ` is a shortcut for "same as previous line" in the feature test (in this case `When`), it has no reason to appear in the step definitions.

Bischoff (approved):

srbarrios:
Personally, I would prefer a parameter in Sumaform that allows handling the log level.

We need to keep in mind that all additional Cucumber scenarios added in the core stage increase the time of testing results (in all kinds of test executions) also it will generate some extra entries in the HTML report.

Also, I think that can be useful for other kind of users, to have this option in Sumaform.

srbarrios (approved):

agraul:
> Personally, I would prefer a parameter in Sumaform that allows handling the log level.

For me either way works. Generally non-qa sumaform users (read: devs) should be able to adjust logging as they go but having it inside sumaform is not something I'm against.

> We need to keep in mind that all additional Cucumber scenarios added in the core stage increase the time of testing results (in all kinds of test executions) also it will generate some extra entries in the HTML report.

You're right that this increases the core stage run time. I don't know if a sumaform approach would be quicker and how often you run sumaform in relation to test suite executions (is there a sumaform execution before every core stage?). 

The idea is not to leave this in for eternity, but rather to catch the heisenbug with additional logging and then decide if it makes sense to keep it or not. Again, if this should rather go into sumaform, we can move it there (now or after solving heisenbug)

srbarrios:
> > Personally, I would prefer a parameter in Sumaform that allows handling the log level.
> 
> For me either way works. Generally non-qa sumaform users (read: devs) should be able to adjust logging as they go but having it inside sumaform is not something I'm against.
> 
> > We need to keep in mind that all additional Cucumber scenarios added in the core stage increase the time of testing results (in all kinds of test executions) also it will generate some extra entries in the HTML report.
> 
> You're right that this increases the core stage run time. I don't know if a sumaform approach would be quicker and how often you run sumaform in relation to test suite executions (is there a sumaform execution before every core stage?).
> 
> The idea is not to leave this in for eternity, but rather to catch the heisenbug with additional logging and then decide if it makes sense to keep it or not. Again, if this should rather go into sumaform, we can move it there (now or after solving heisenbug)

Ah ok, I did not catch that this was temporary, to find an issue.
Ok, so let's keep it here, for now, just to avoid re-doing it in sumaform, but it might be interesting to create a card in sumaform, as that could be a "nice to have" there some day.

agraul:
@srbarrios https://github.com/uyuni-project/sumaform/issues/871

PR #3485, opened 2021-03-30
Remove SSL download from push package

maximenoel8:
https://github.com/SUSE/spacewalk/issues/14358

- [x] No changelog needed

After investigation, the certificate doesn't seem needed anymore. I could run `rhnpush` command to push the package on the server side without any certificate issue.

Test suite have been run without the ssl downloaded without new error.

The SSL download should not be a test by itself because it's not part of SUMA project.

Discution : https://github.com/uyuni-project/uyuni/pull/3318#issuecomment-800077153


Bischoff (approved):
LGTM. Don't forget the ports to 4.0 and 4.1 :-) .

PR #3490, opened 2021-03-30
Keep salt minion when installed

mcalmer:
## What does this PR change?

Keep salt-minion when it is installed to prevent update problems with dependend packages not available in the bootstrap repo.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/pull/14426

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
@@ -139,7 +139,7 @@ install_gnupg_debian:
 {% include 'channels/gpg-keys.sls' %}
 
 salt-minion-package:
-  pkg.latest:
+  pkg.installed:
     - name: salt-minion
     - install_recommends: False
     - require:
--- a/susemanager-utils/susemanager-sls/susemanager-sls.changes
+++ b/susemanager-utils/susemanager-sls/susemanager-sls.changes
@@ -1,3 +1,5 @@
+- keep salt-minion when it is installed to prevent update problems with
+  dependend packages not available in the bootstrap repo (bsc#1183573)
 - Add support for AlmaLinux 8
 - Provide Custom Info as Pillar data
 - Add support for Amazon Linux 2


susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -131,7 +131,7 @@ install_gnupg_debian:
 {% include 'channels/gpg-keys.sls' %}
 
 salt-minion-package:
-  pkg.latest:
+  pkg.installed:
```
srbarrios, line 134: @mcalmer I'm worried about how this could affect our side, where every machine has a salt-minion installed by sumaform, when configuring it after deployment. We will need to clean first those leftovers from sumaform, to avoid testing with an incorrect salt version?

We might have a similar behavior in our Build Validation pipelines, where we use MU repos to boostrap the minions.

Bischoff: I think this is on the contrary an opportunity to clean up the leftovers from sumaform. It's a long time we want to do that and we already do it in BV testsuite. So I see it as a good thing.

I am just wondering why we should not do it in sumaform instead of doing it in test suite. This way, developers would benefit from the cleanup too. @rjmateus ping ?

mcalmer: This time the return was quick. We got the positive feedback. We can delay merging some days if wanted.

srbarrios: I have taken the liberty of triggering a full test suite of this PR https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-pr-ci-tests/55/

srbarrios: CI test suite is not affected by this code change, so I guess we can merge it.

mcalmer: Cool, thanks.

PR #3492, opened 2021-03-30
Switch from kunihiko-t/review-request-action to necojackarc/auto-request-review

ktsamis:
## What does this PR change?

Switch from kunihiko-t/review-request-action to necojackarc/auto-request-review

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


srbarrios:
The current GitHub action, as far as I remember never worked for forks, and Kostas found this other, which at least seems maintained and easier to configure.
At that moment it seems to fail, for the recurrent issue in GitHub actions of not having permissions in the repository from a fork. @ktsamis I suggest that you create the same PR but from a branch coming from the upstream repo instead of using your fork, it should allow us to validate that this GitHub action works.

ktsamis:
Exactly why Oscar said, it's easier to maintain and it should work with forks as well I think. The error we have right now is because the PR changes the workflow itself, I think it will go away in the next PR, this should work with forks as well. If not we lost nothing and we are in the same situation.

ktsamis:
Superseded by https://github.com/uyuni-project/uyuni/pull/3495

PR #3529, opened 2021-04-08
Avoid conflicts with running ioloop on mgr_events engine (bsc#1172711)

meaksh:
## What does this PR change?

This PR fixes an issue with our custom `mgr_events` Salt engine at the time of dealing with the Tornado IOLoop.

Before this PR, the `mgr_events` engine is not properly instantiating the IOLoop to be used by this process, so it clashes when there is an already running tornado IOLoop, for example, when SaltStack Enterprise is running:

```console
Jan 16 15:01:59 suma-hostname salt-master[16536]: [CRITICAL] Engine 'mgr_events' could not be started!
Jan 16 15:01:59 suma-hostname salt-master[16536]: Traceback (most recent call last):
Jan 16 15:01:59 suma-hostname salt-master[16536]: File "/usr/lib/python3.6/site-packages/salt/engines/__init__.py", line 132, in run
Jan 16 15:01:59 suma-hostname salt-master[16536]: self.engine[self.fun](**kwargs)
Jan 16 15:01:59 suma-hostname salt-master[16536]: File "/usr/share/susemanager/modules/engines/mgr_events.py", line 214, in start
Jan 16 15:01:59 suma-hostname salt-master[16536]: io_loop.start()
Jan 16 15:01:59 suma-hostname salt-master[16536]: File "/usr/lib64/python3.6/site-packages/tornado/ioloop.py", line 754, in start
Jan 16 15:01:59 suma-hostname salt-master[16536]: raise RuntimeError("IOLoop is already running")
```

With this PR, we properly instantiate a new IOLoop and make it the current, as it's done for other engines included in Salt codebase.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **bugfix**

- [x] **DONE**

## Test coverage
- No tests: **bugfix**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/11767

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
+++ b/susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
@@ -70,12 +70,9 @@ except ImportError:
     HAS_PSYCOPG2 = False
 
 # Import salt libs
+import salt.ext.tornado
 import salt.utils.event
 import json
-
-# Import third-party libs
-import tornado
-from salt.utils.zeromq import ZMQDefaultLoop
 
 log = logging.getLogger(__name__)
 
@@ -167,7 +164,7 @@ class Responder:
     def _is_batch_mode(self, data):
         return data.get("metadata", {}).get("batch-mode")
 
-    @tornado.gen.coroutine
+    @salt.ext.tornado.gen.coroutine
     def add_event_to_queue(self, raw):
         tag, data = self.event_bus.unpack(raw, self.event_bus.serial)
         self._insert(tag, data)
@@ -177,7 +174,7 @@ class Responder:
             log.error("%s: Diconnected from database. Trying to reconnect...", __name__)
             self._connect_to_database()
 
-    @tornado.gen.coroutine
+    @salt.ext.tornado.gen.coroutine
     def add_token(self):
         self.tokens = min(self.tokens + 1, self.config['commit_burst'])
         self.attempt_commit()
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -204,7 +201,8 @@ def start(**config):
     '''
     Listen to events and write them to the Postgres database
     '''
-    io_loop = ZMQDefaultLoop.current()
+    io_loop = salt.ext.tornado.ioloop.IOLoop(make_current=False)
+    io_loop.make_current()
```
moio, line 205: Is this the same as the code below?

```suggestion
    io_loop = salt.ext.tornado.ioloop.IOLoop(make_current=True)
```

moio (approved):
I do not think I understand the deep difference, but I am fine with this patch given the pattern is used in https://github.com/openSUSE/salt/blob/openSUSE-3000/salt/engines/webhook.py

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
meaksh: I'm not 100% sure, but apparently the behavior might not be the same.

It seems creating the IOLoop instance with `make_current=True` might raise an exception when there is already another IOLoop: https://www.tornadoweb.org/en/stable/ioloop.html?tornado.ioloop.IOLoop#tornado.ioloop.IOLoop (bottom)

There is also this note for `IOLoop.make_current` method:

Makes this the IOLoop for the current thread.

*An IOLoop automatically becomes current for its thread when it is started, but it is sometimes useful to call make_current explicitly before starting the IOLoop, so that code run at startup time can find the right instance.*

PR #3533, opened 2021-04-08
QA: Capybara hold the web browser session

srbarrios:
## What does this PR change?

This code changes the actual behavior of our test suite, currently, we open a new web session for each scenario, so we require a log-in for each scenario. This PR aims to keep open the browser for the whole feature, resetting the session only when a new Cucumber feature starts.

Some scenarios rely on the current behavior, so they expect that we have a reset page on every scenario, but if we change that by holding the session, we must assure that the scenarios don't rely anymore on this.

So, this PR also includes a change in the KVM/XEN tests where we open a new tab, and we expect to have it closed when a new scenario starts. What we do in this case is to assure that we close the tab and we move to the first tab.

**Note:** 
This PR will speed up the test suite "per se", but holding the session by default will help to speed up the tests if we start removing the logout and log-in actions that we do every time we start a scenario. I will proceed to these refactors in next PRs.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.0
- Manager-4.1

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/minkvm_guests.feature
+++ b/testsuite/features/secondary/minkvm_guests.feature
@@ -82,6 +82,7 @@ Feature: Be able to manage KVM virtual machines via the GUI
     When I click on "Graphical Console" in row "test-vm"
     And I switch to last opened window
     Then I wait until I see the VNC graphical console
+    When I close the last opened window
 
 @virthost_kvm
   Scenario: Suspend a KVM virtual machine
@@ -223,6 +224,7 @@ Feature: Be able to manage KVM virtual machines via the GUI
     When I click on "Graphical Console" in row "test-vm2"
     And I switch to last opened window
     Then I wait until I see the spice graphical console
+    When I close the last opened window
 
 @virthost_kvm
   Scenario: Show the virtual storage pools and volumes for KVM
@@ -430,6 +432,7 @@ Feature: Be able to manage KVM virtual machines via the GUI
     And I wait until I see the VNC graphical console
     Then I wait at most 500 seconds until I see modal containing "Disconnected" text
     And I wait at most 500 seconds until Salt master sees "test-vm2" as "unaccepted"
+    When I close the last opened window
 
 @long_test
 @virthost_kvm
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -103,6 +103,16 @@ After do |scenario|
       debug_server_on_realtime_failure
     end
   end
+  page.instance_variable_set(:@touched, false)
+end
+
+# Reset the page if it's running the first scenario of a feature
+$first_scenario = true
+Before do
+  next unless $first_scenario
+
+  page.reset!
+  $first_scenario = false
 end
 
 AfterStep do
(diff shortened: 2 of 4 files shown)


srbarrios:
Trying the switch of tabs, to fix the corner case of not resetting the browser session when you are in a different tab.
As before holding the session, we reset the session, the second tab is closed and you starts from the Uyuni home page.

https://user-images.githubusercontent.com/2827771/114170324-0c14be80-9933-11eb-878c-ca83a58de2e3.mp4

srbarrios:
Testing PR:
https://ci.suse.de/blue/organizations/jenkins/uyuni-pr-ci-tests/detail/uyuni-pr-ci-tests/65/pipeline

testsuite/features/secondary/minkvm_guests.feature
```diff
@@ -82,6 +82,7 @@ Feature: Be able to manage KVM virtual machines via the GUI
     When I click on "Graphical Console" in row "test-vm"
     And I switch to last opened window
     Then I wait until I see the VNC graphical console
```
Bischoff, line 84: And

testsuite/features/secondary/minkvm_guests.feature
```diff
@@ -223,6 +224,7 @@ Feature: Be able to manage KVM virtual machines via the GUI
     When I click on "Graphical Console" in row "test-vm2"
     And I switch to last opened window
     Then I wait until I see the spice graphical console
```
Bischoff, line 226: And

testsuite/features/secondary/minkvm_guests.feature
```diff
@@ -430,6 +432,7 @@ Feature: Be able to manage KVM virtual machines via the GUI
     And I wait until I see the VNC graphical console
     Then I wait at most 500 seconds until I see modal containing "Disconnected" text
```
Bischoff, line 433: And

testsuite/features/support/env.rb
```diff
@@ -103,6 +103,16 @@ def enable_assertions
       debug_server_on_realtime_failure
     end
   end
+  page.instance_variable_set(:@touched, false)
+end
+
+# Reset the page if it's running the first scenario of a feature
+$first_scenario = true
+Before do
```
Bischoff, line 111: too bad we don't have hooks at feature level, it would avoid running this for every scenario :-(

Bischoff (changes requested):
minor nitpicks

testsuite/features/secondary/minkvm_guests.feature
srbarrios: Not part of my changes

testsuite/features/secondary/minkvm_guests.feature
srbarrios: Not part of my changes

testsuite/features/secondary/minkvm_guests.feature
srbarrios: Not part of my changes

testsuite/features/support/env.rb
srbarrios: yup, let's move to java, there they implemented it :trollface:

srbarrios:
Note:
I will port it only after we had a CI pass and I can assure it will not break other CIs

PR #3542, opened 2021-04-12
Fix insecure JXM configuration

witekest:
## What does this PR change?

The change restricts access to Java Remote Method Invocation to
`localhost` to prevent unauthorized remote code execution.

## GUI diff

No change.

## Documentation
- [ ] Release notes: JMX configuration hardening should be mentioned.

## Test coverage
- No tests: Tests updated

- [x] **DONE**

## Links

Fixes # https://bugzilla.suse.com/show_bug.cgi?id=1184617

- [x] **DONE**

## Changelogs

- [x] susemanager-sls.changes

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/srvmonitoring/removejmxprops.sls
+++ b/susemanager-utils/susemanager-sls/salt/srvmonitoring/removejmxprops.sls
@@ -1,3 +1,8 @@
+remove_{{remove_jmx_props.service}}_jmx_host:
+  cmd.run:
+    - name: sed -ri 's/JAVA_OPTS="(.*)-Dcom\.sun\.management\.jmxremote\.host=\S*(.*)"/JAVA_OPTS="\1 \2"/' {{remove_jmx_props.file}}
+    - onlyif: grep -F -- '-Dcom.sun.management.jmxremote.host=' {{remove_jmx_props.file}}
+
 remove_{{remove_jmx_props.service}}_jmx_port:
   cmd.run:
     - name: sed -ri 's/JAVA_OPTS="(.*)-Dcom\.sun\.management\.jmxremote\.port=[0-9]*(.*)"/JAVA_OPTS="\1 \2"/' {{remove_jmx_props.file}}
--- a/java/code/src/com/suse/manager/webui/services/impl/test/monitoring/disable.json
+++ b/java/code/src/com/suse/manager/webui/services/impl/test/monitoring/disable.json
@@ -26,6 +26,22 @@
             "duration": 286.435,
             "__id__": "postgres_exporter_service"
         },
+        "cmd_|-remove_tomcat_jmx_host_|-sed -ri 's/JAVA_OPTS=\"(.*)-Dcom\\.sun\\.management\\.jmxremote\\.host=localhost(.*)\"/JAVA_OPTS=\"\\1 \\2\"/' /etc/sysconfig/tomcat_|-run": {
+            "name": "sed -ri 's/JAVA_OPTS=\"(.*)-Dcom\\.sun\\.management\\.jmxremote\\.host=localhost(.*)\"/JAVA_OPTS=\"\\1 \\2\"/' /etc/sysconfig/tomcat",
+            "changes": {
+                "pid": 6335,
+                "retcode": 0,
+                "stdout": "",
+                "stderr": ""
+            },
+            "result": true,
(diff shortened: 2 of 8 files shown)


meaksh (commented):
Thanks @witekest - the changes on the SLS files look good to me :+1:

I only miss the necessary changelog entry for `susemanager-sls` at `susemanager-utils/susemanager-sls/susemanager-sls.changes` including the bsc reference.

mcalmer:
@witekest beside of the changelog problem: how we fix the customer configuration when he has monitoring enabled and update the package? AFAIK just installing the package does not change the config files, right?

witekest:
> how we fix the customer configuration when he has monitoring enabled and update the package? AFAIK just installing the package does not change the config files, right?

Good point. Disabling and enabling monitoring should work. Any other idea than documenting it?

juliogonzalez (changes requested):
Changelog missing, and same concerns as those from @mcalmer.

Looks to me that we'd need to document at the release notes that users/customers should apply the highstate to get fixed? Any other better approach?

juliogonzalez:
> Good point. Disabling and enabling monitoring should work. Any other idea than documenting it?

Applying the highstate is not enough :-?

mcalmer:
> > Good point. Disabling and enabling monitoring should work. Any other idea than documenting it?
> 
> Applying the highstate is not enough :-?

No, we talk about the "Server". You cannot call highstate for the server.

@witekest will "mgr-monitoring-ctl status" show it as enabled after the update when it was enabled before?

mcalmer:
This is how the Configuration page would look like:

image

susemanager-utils/susemanager-sls/salt/srvmonitoring/removejmxprops.sls
```diff
@@ -1,3 +1,8 @@
+remove_{{remove_jmx_props.service}}_jmx_host:
+  cmd.run:
+    - name: sed -i 's/JAVA_OPTS="(.*)-Dcom\.sun\.management\.jmxremote\.host=localhost(.*)"/JAVA_OPTS="\1 \2"/' {{remove_jmx_props.file}}
```
mcalmer, line 3: This failed with: 

  "sed: -e expression #1, char 90: invalid reference \\2 on `s' command's RHS"

Seems `(` needs to be escaped

mcalmer (changes requested):

mcalmer:
I am unsure what we discussed: do we want to fix the configuration via %post script or do we want just to document it?
Please also check, if the salt state restart the jmx exporters with the new config when it gets changed via the configuration.

witekest:
After setting `JAVA_OPTS` _Tomcat_ and _Taskomatic_ have to be restarted to apply the changes. I think `susemanager-sls.spec %post` is not the right place to implement it.

witekest:
@juliogonzalez has pointed to me that all the services have to be stopped during the update anyway.
@mcalmer, could you confirm that `susemanager-sls.spec %post` is the place you have meant to implement it?

mcalmer:
I think in %post we should do the following test:

```
if service prometheus_jmx_tomcat is enabled and /etc/sysconfig/tomcat does not contain "jmxremote.host="; then
  add ... jmxremote.host=locahost to config
endif
```

same for taskomatic.

Additionally we should review the `spacewalk-service start|stop` . AFAIK it does not handle monitoring which is not so good.

witekest:
> Additionally we should review the `spacewalk-service start|stop` . AFAIK it does not handle monitoring which is not so good.

We could be updating `/etc/rhn/service-list` file. Restarting exporters is not needed for this PR to take effect though. Do we have another reason for restarting exporters?
In any case I think we should handle it in a separate PR.

mcalmer (approved):
LGTM

mcalmer:
> > Additionally we should review the `spacewalk-service start|stop` . AFAIK it does not handle monitoring which is not so good.
> 
> We could be updating `/etc/rhn/service-list` file. Restarting exporters is not needed for this PR to take effect though. Do we have another reason for restarting exporters?
> In any case I think we should handle it in a separate PR.

Yes, for a separate PR. But be careful about service-list. AFAIK this is not used anymore. There is a systemd target and other funny things.
I would suggest to play around with systemd dependencies. What about "binding" the JMX exporter to the main service.
It does not make sense to run a tomcat jmx exporter when tomcat itself is not running.

juliogonzalez (approved):
Seems to be OK. If you tested that the change to susemanager-sls.spec are working fine, then go ahead and merge :-)

PR #3544, opened 2021-04-13
Supportconfig: Fix the IS_SUSE variable reading

hustodemon:
It seems it haven't worked for some time (I was missing files for inventory, for instance). This should make it work again.

In the end this rather obscure approach with grep was used.
Alternative (cleaner) approach with sourcing the os-release file was not
used as there is a danger of shadowing other env variables.

## Documentation
- No documentation needed: internal stuff

- [x] **DONE**

## Test coverage
- No tests: no tests here

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/backend/satellite_tools/spacewalk-debug
+++ b/backend/satellite_tools/spacewalk-debug
@@ -28,7 +28,7 @@ IS_SUSE=0
 NO_REPORTS=0
 NO_COMPRESSION=0
 MAX_LOG_AGE=10
-if [ -f /etc/SuSE-release ]; then
+if `grep -iq '^ID_LIKE=.*suse' /etc/os-release`; then
     IS_SUSE=1
 fi
 if [ "$SPACEWALK_DEBUG_NO_REPORTS" == 1 ]; then


backend/satellite_tools/spacewalk-debug
```diff
@@ -28,7 +28,7 @@ IS_SUSE=0
 NO_REPORTS=0
 NO_COMPRESSION=0
 MAX_LOG_AGE=10
-if [ -f /etc/SuSE-release ]; then
+if `grep -iq '^ID_LIKE=.*suse' /etc/os-release`; then
```
juliogonzalez, line 31: I understand that despite the variable name `IS_SUSE` you really want this to work for both SLE and openSUSE, right?

If so, then LGTM.

juliogonzalez (approved):

backend/satellite_tools/spacewalk-debug
hustodemon: Yes. If we have uyuni on opensuse, i also think we want this to be `1`.

PR #3560, opened 2021-04-14
Maintenance windows web calendar

parlt91:
## What does this PR change?

Replace the existing display of the ical file in `Maintenance Windows` -> `Calendars` -> `Details` and
the listing of the upcoming maintenance windows in `Maintenance Windows` -> `Schedules` -> `Details`
with a interactive calendar widget.

A list of remaining issues that need to be addressed once this PR is merged can be seen here:
https://github.com/SUSE/spacewalk/issues/12183

### Before:

**Schedule:**
schedule-before

**Calendar:**
calendar-before

### After:

**Schedule:**
schedule-month-after

**Calendar:**
calendar-month-after

**Day view:**
calendar-day-after

- [x] **DONE**

## Documentation

TODO
- Documentation issue was created: [Link for SUSE Manager contributors]

- [ ] **DONE**

## Test coverage
- No tests: React unit tests and cucumber tests will be provided in the future

- [ ] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
+import * as React from "react";
+import { useState, useEffect, useRef } from "react";
+
+import FullCalendar from "@fullcalendar/react";
+import dayGridPlugin from "@fullcalendar/daygrid";
+import timeGridPlugin from "@fullcalendar/timegrid";
+import interactionPlugin from "@fullcalendar/interaction";
+import allLocales from "@fullcalendar/core/locales-all";
```
Etheryte, line 8: While this isn't a huge consideration with Uyuni, this might add a lot of volume to the package while we only support a small number of locales. I don't know if we currently have a way to manage those in the frontend packaging? Perhaps @ncounter or @cbbayburt know?

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+  const calendarRef = useRef<any>();
+
+  useEffect(() => {
+    getEvents("initial");
+    setCurrentDate(getApi().currentDataManager.data.currentDate);
+    // Hack: Make Fullcalendar button icons show correctly
+    jQuery(".fc-backButton-button").html('<div class="fa fa-angle-left"></div>');
+    jQuery(".fc-nextButton-button").html('<div class="fa fa-angle-right"></div>');
+    jQuery(".fc-skipBackButton-button").html('<div class="fa fa-angle-double-left"></div>')
+      .prop('title', "Skip to the last maintenance window");
```
Etheryte, line 40: Here and below, needs `t()`.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: I am aware if this. AFAIK there is also the option to only import the locales that we support. Downside of that is that we would have to revisit this whenever we decide to support a new locale.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: Uff I missed those :sweat_smile:
Thanks for pointing out

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  const [events, setEvents] = useState<Event[]>([]);
+  const [currentDate, setCurrentDate] = useState<Date>();
+  const calendarRef = useRef<any>();
+
+  useEffect(() => {
+    getEvents("initial");
+    setCurrentDate(getApi().currentDataManager.data.currentDate);
+    // Hack: Make Fullcalendar button icons show correctly
+    jQuery(".fc-backButton-button").html('<div class="fa fa-angle-left"></div>');
```
Etheryte, line 37: What's the problem that's being solved here? I understand that the default DOM somehow doesn't suit us, but what's the exact issue? Without being familiar with it, might one of these be of value here?  

 - https://fullcalendar.io/docs/buttonIcons  
 - https://fullcalendar.io/docs/themeSystem

parlt91: For some reason that I couldn't figure out adding the icons this way does not work here. They are only displayed as empty square (like it can't access the texture somehow)

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+  const onSkipNext = () => {
+    // Get last date of currently displayed range
+    let date = getDate("next");
+    // Check if we need to update the events
+    if (!needsUpdate(date) && skipToNextEvent(date)) {
+      return;
+    }
+    // If we are in day view and need to update we want to make sure to always use the first day of the next month
+    if (getApi().currentDataManager.data.currentViewType === "timeGridDay" && moment(date).date() !== 1) {
+      date = Date.parse(moment(date).add(1, 'month').startOf('month'));
```
Etheryte, line 98: Here and everywhere else, we probably want to use `moment()` instead of `Date.parse()` if we already use it for some operations. If need be, you can also use `moment(input, formatString)` if the format isn't one of the default ones.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  const onSkipNext = () => {
+    // Get last date of currently displayed range
+    let date = getDate("next");
+    // Check if we need to update the events
+    if (!needsUpdate(date) && skipToNextEvent(date)) {
+      return;
+    }
+    // If we are in day view and need to update we want to make sure to always use the first day of the next month
+    if (getApi().currentDataManager.data.currentViewType === "timeGridDay" && moment(date).date() !== 1) {
```
Etheryte, line 97: `moment(date).date !== 1` looks odd, what's the intent here? I might be misunderstanding, but perhaps you meant `moment(...).isValid()`?

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: I want date to be in `epochmillis` format here. `Date.parse()` is just a convenient way to do this. Though I think moment should also be able to do this. I will look into it

web/html/src/manager/maintenance/calendar/web-calendar.tsx
Etheryte, line 97: Instead of repeatedly calling `moment(date)`, I recommend consider making `getDate(...)` return a moment instance instead.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+      date = Date.parse(moment(date).add(1, 'month').startOf('month'));
+    }
+    const endpoint = `/rhn/manager/api/maintenance/events/skipNext/${props.type}/${date}/${props.id}`;
```
Etheryte, line 100: Automatically stringifying a date this way may have unexpected consequences in different locales. I would instead recommend using [`moment(...).format(formatString)`](https://momentjs.com/docs/#/displaying/format/) to be sure we always pass the same format.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: `moment(date).date()` returns the day of the month. I only want to add one month here if `date` is not already the first of the month. Otherwise we would jump two months ahead in some cases which I want to prevent

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: date is formatted in milliseconds since epoch (meaning 1970-01-01) here. This way I make sure that the format is always the same.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  // Get the date used to fetch events based on the operation
+  const getDate = (operation) => {
+    switch (operation) {
+      case "next":
+        return Date.parse(getApi().currentDataManager.state.dateProfile.currentRange.end);
+      case "back":
+        // currentRange.start returns the first day of the displayed month, but we want to end up in
+        // the previous month so we subtract one day in this case
+        return Date.parse(getApi().currentDataManager.state.dateProfile.currentRange.start) - 24 * 60 * 60 * 1000;
```
Etheryte, line 149: Here and elsewhere, while this is not wrong per se, it will be easier to reason about and change if need be if we use moment to modify the underlying object as is already done in `onSkipNext `.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: Okay agreed I can change those instances to be in `moment` format already. In this case I only would need to format the date I pass to the backend to epoch millis

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+  };
+
+  const skipToNextEvent = (date) => {
+    // We are only interested in events from the same month that are later or equal to 'date'
+    const filteredEvents = events.filter(event => (
+      moment.parseZone(event.start).month() === moment(currentDate).month() &&
+      moment.parseZone(event.start).date() >= moment(date).date())
+    );
+    // Only perform action if there is an event to go to
+    filteredEvents.length > 0 && getApi().gotoDate(filteredEvents[0].start);
```
Etheryte, line 185: Not wrong per se, I would however recommend an `if` to make it easier to spot that a conditional function call is taking place here when skimming through the code.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: I have no preference here. This is just a leftover from before I used moment to format the dates. I can change it to be in line with the rest.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -0,0 +1,262 @@
...
+    return filteredEvents.length > 0;
+  }
+
+  return (
+    <div>
+      <FullCalendar
+        ref={calendarRef}
+        timeZone={""} // Prevent FullCalendar from using the browsers set timezone
+        locales={allLocales}
+        locale={window.preferredLocale.replace("_", "-")}
```
Etheryte, line 208: Would appreciate a comment here to be clear about what format we're going to and from, right now it's hard to say whether this is safe or not without going to the backend code.

parlt91: Sure I can add a comment. At the moment we use all the locales with `_` on the backend while Fullcalendar expects those with `-` instead

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: Haha I knew that one would come up. Although I expected Dario to be the one to complain about it.
Well I like that format better, since it takes less space, but no strong aversion against using an if instead :P

web/html/src/manager/maintenance/calendar/web-calendar.tsx
Etheryte: Ah okay, yeah, off the top of my head I wasn't sure what comes out of it here since I don't use moment that often day to day. I think a good small change here would be to extract it to a variable with a descriptive name a-la `isFirstDayOfMonth` or w/e.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
Etheryte: Fair enough. Imo it's not a showstopper to shipping this in any way, might be nice to create a separate ticket to follow up on this in the future.

Etheryte (approved):
Had some small questions and nitpicks, but overall this looks very good to me. Nice work. 👍

Etheryte:
As a separate comment from the existing code, would be nice to add some unit tests to this at some point, but that doesn't have to happen in this PR.

parlt91:
> As a separate comment from the existing code, would be nice to add some unit tests to this at some point, but that doesn't have to happen in this PR.

Agreed. I put it into the issue to track things that are still to do

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: I added the issue to the card. So that we don't forget to look into it later on.

java/code/src/com/suse/manager/webui/controllers/maintenance/MaintenanceController.java
```diff
@@ -114,4 +150,30 @@ static void handleRescheduleResult(List<RescheduleResult> results, RescheduleStr
...
+                        "title", event.getName(),
+                        "start", ZonedDateTime.ofInstant(
+                                Instant.ofEpochMilli(event.getFromMilliseconds()), zoneId
+                        ).toString().split("\\[")[0],
+                        "end", ZonedDateTime.ofInstant(
+                                Instant.ofEpochMilli(event.getToMilliseconds()), zoneId
+                        ).toString().split("\\[")[0]
+                ))
+        );
+        return json(response, ResultJson.success(data));
```
cbbayburt, line 177: Another nitpick: maybe you can create a `...Json` DTO class with `title`, `start` and `end` properties to make the API signature more clear (even a nested class would be ok)

cbbayburt (approved):
I just left a couple of nitpics below, but overall it looks great. Thanks @parlt91 

Let's not forget about JS unit tests also

java/code/src/com/suse/manager/maintenance/MaintenanceManager.java
```diff
@@ -512,6 +512,52 @@ public int retractScheduleFromSystems(User user, Set<Long> systemIds) {
...
+     * to perform starting from the given date.
+     *
+     * @param user the current user
+     * @param operation get previous, current or future maintenance windows based on the operation
+     * @param id the id of the calendar or schedule
+     * @param type determines whether the id belongs to a schedule or a calendar
+     * @param date the date to start looking for maintenance windows
+     * @return the resulting list of maintenance windows
+     */
+    public List<MaintenanceWindowData> preprocessMaintenanceWindows(User user, String operation, Long id, String type,
```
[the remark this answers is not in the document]
parlt91: Agreed this makes sense and looks much cleaner now

java/code/src/com/suse/manager/webui/controllers/maintenance/MaintenanceController.java
parlt91: Done

cbbayburt:
@parlt91 thanks for the quick response. FYI: The JUnit failure is from the `master` branch and has been fixed since. A rebase should make it green.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
ncounter: I would not see a bid deal to define one by one, if we add more it will not be dynamically added I guess, but we will have to add somewhere else the enablement. So, I'm fine with both approaches. If we go with the specifying one by one, I'd love to see one single point to configure it, instead of having it inside of a component, then maybe in another one, and then in Java backend, and maybe more. Question: how many locales we support now? Btw, it's okay if we keep them all for now, but take a note/issue to revisit maybe later.

branding/css/susemanager-theme.less
```diff
@@ -1314,3 +1314,13 @@ select.small-select {
...
+
+.fc button.fc-button, .fc button.fc-button-primary {
+  .btn;
+  .btn-default;
+}
+
+.fc-daygrid-day-frame:hover {
+  cursor: pointer;
+  background-color: fadeout(@gray, 70%);
+}
```
ncounter, line 1326: Please add a comment about the scope of these: these are HACKS to hijack the plugin style to behave like our instead, not classes to reuse.

branding/css/uyuni-theme.less
```diff
@@ -1317,3 +1317,13 @@ select.small-select {
...
+
+.fc button.fc-button, .fc button.fc-button-primary {
+  .btn;
+  .btn-default;
+}
+
+.fc-daygrid-day-frame:hover {
+  cursor: pointer;
+  background-color: fadeout(@gray, 70%);
+}
```
ncounter, line 1329: Same as above :wink:

java/code/src/com/suse/manager/maintenance/IcalUtils.java
```diff
@@ -145,6 +150,25 @@
...
     }
 
+    private PeriodList getInitialEvents(Collection<CalendarComponent> events, Period period) {
+        PeriodList periodList = new PeriodList();
+        events.forEach(event -> {
+                    try {
+                        DateTime start = new DateTime(event.getProperty(Property.DTSTART).getValue());
+                        DateTime end = new DateTime(event.getProperty(Property.DTEND).getValue());
+                        // Check if event is contained in given period
+                        if (period.getStart().compareTo(start) <= 0 && period.getEnd().compareTo(end) >= 0) {
```
ncounter, line 160: sounds like you want to use `period.includes(start) && period.includes(end)` ?

ncounter (approved):
I left very small comments but the overall LGTM already. Very good, thank you :+1:

java/code/src/com/suse/manager/maintenance/IcalUtils.java
parlt91: Yeah that's exactly what I wanted here :facepalm:
Over complicating things for no reason it seems :sweat_smile: 
Thanks

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: Sounds good I will add a point about it to the issue.

PR #3564, opened 2021-04-15
Testsuite: Make sure package is installed on/removed from client in retracted patches tests

hustodemon:
Previously, we were relying on the 'I wait until refresh package list on
"sle_minion" is finished' step, which didn't work very well in case
there was a lot of subsequent package installations/removals.

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_retracted_patches.feature
+++ b/testsuite/features/secondary/min_retracted_patches.feature
@@ -6,15 +6,14 @@ Feature: Retracted patches
 
   Scenario: Installed retracted package should show icon in the system packages list
     Given I am authorized as "admin" with password "admin"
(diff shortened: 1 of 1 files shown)


srbarrios:
What about... hide the fact that the call is synchron, and just have step definition as :

I install "x" package using spacecmd
I remove "x" package using spacecmd

And if you wait for it inside... it's fine.

srbarrios:
I see, you are just re-using a step already implemented. Ok, nevermind then.

srbarrios (approved):

hustodemon:
Thanks for reviewing @srbarrios !

PR #3572, opened 2021-04-16
Uyuni master delete archived

brejoc:
## What does this PR change?

This adds the possibility to delete archived actions from the database via spacecmd: `spacecmd -- schedule_deletearchived`

## GUI diff

Just cli.

## Documentation

- [ ] This still needs to be done.

- [ ] **DONE**

## Test coverage

- Unit tests were added

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14196

## Changelogs

- [ ] Changelog needed

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/src/spacecmd/schedule.py
+++ b/spacecmd/src/spacecmd/schedule.py
@@ -465,6 +465,7 @@ def do_schedule_deletearchived(self, args):
     """
     This method removes all of the archived actions.
     """
+    args = args.split() or []
     if args:
         begin_date = parse_time_input(args[0])
         logging.debug('Begin Date: %s' % begin_date)
--- a/spacecmd/src/spacecmd/schedule.py
+++ b/spacecmd/src/spacecmd/schedule.py
@@ -452,3 +452,49 @@ def help_schedule_list(self):
 
 def do_schedule_list(self, args):
     return self.print_schedule_summary('all', args)
+
+####################
+
+def help_schedule_deletearchived(self):
+    print(_('schedule_deletearchived: Delete all archived actions'))
+    print(_('usage: schedule_deletearchived [BEGINDATE] [ENDDATE]'))
+    print('')
+    print(self.HELP_TIME_OPTS)
+
+def do_schedule_deletearchived(self, args):
+    """
+    This method removes all of the archived actions.
+    """
+    if args:
+        begin_date = parse_time_input(args[0])
+        logging.debug('Begin Date: %s' % begin_date)
+    else:
+        begin_date = None
+
+    if len(args) > 1:
+        end_date = parse_time_input(args[1])
+        logging.debug('End Date:   %s' % end_date)
+    else:
+        end_date = None
+
+    while True:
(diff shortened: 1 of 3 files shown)


spacecmd/src/spacecmd/schedule.py
```diff
@@ -452,3 +452,49 @@ def help_schedule_list(self):
...
+                    if action.get('earliest') < begin_date:
+                        continue
+                if end_date:
+                    if action.get('earliest') > end_date:
+                        continue
+                action_ids.append(action.get('id'))
+
+            # Pass list of actions that should be deleted
+            if action_ids:
+                self.client.schedule.deleteActions(self.session, action_ids)
```
agraul, line 498: I haven't looked at `deleteActions` yet, do you know what it does then you pass a list that contains the same id twice? If that can be a problem, we could use a Set for `action_ids` instead of a list.

brejoc: I don't think this should happen, but also wouldn't harm. Thx!

spacecmd/src/spacecmd/schedule.py
```diff
@@ -452,3 +452,41 @@ def help_schedule_list(self):
...
+        actions = self.client.schedule.listArchivedActions(self.session)
+        logging.debug("actions: {}".format(actions))
+        if actions:
+            if not user_already_prompted:
+                if len(actions) >= 10000:
+                    action_length = ">10000"
+                else:
+                    action_length = len(actions)
+                user_answer = prompt_user(_("Do you want to delete all ({}) archived actions? [y/N]").format(action_length))
+                if user_answer not in ("y", "Y", "yes", "Yes", "YES"):
```
agraul, line 478: I'd just against the lower case version like this:
```suggestion
                if user_answer.lower() not in ("y", "yes",):
```

But I wonder if we can assume a "yes" as an answer. The prompt is getting translated, are we sure that the "[y/N]" is not changed to e.g. "[j/N]" (for "ja/Nein")?

spacecmd/src/spacecmd/schedule.py
```diff
@@ -452,3 +452,41 @@ def help_schedule_list(self):
...
+                    break
+                else:
+                    user_already_prompted = True
+            # Collect IDs of actions that should be deleted
+            action_ids = [action.get('id') for action in actions]
+
+            # Pass list of actions that should be deleted
+            if action_ids:
+                # set needs to be cast to list, since set cannot be marshalled
+                self.client.schedule.deleteActions(self.session, list(set(action_ids)))
```
agraul, line 488: We can save the conversion from a List to a Set by generating a Set in the first place
```suggestion
            action_ids = {action.get('id') for action in actions}

            # Pass list of actions that should be deleted
            if action_ids:
                # set needs to be cast to list, since set cannot be marshalled
                self.client.schedule.deleteActions(self.session, list(action_ids))
```

meaksh:
Closing in favor of https://github.com/uyuni-project/uyuni/pull/4007

Thanks!

PR #3577, opened 2021-04-19
AnsibleControlNode Configuration Paths Page

ncounter:
## TODO
- [x] Agree on the permissions model (any user having access to the system should be able to CRUD paths?)

## What does this PR change?

- DB: add the `suseAnsiblePath` table for storing paths to playbook directories and inventory files per ansible control node system
- Backend: `AnsiblePath` entity (+ subclasses for inventory and playbooks), Factory + Manager code for CRUD
- XMLRPC: ansible paths (inventory + playbook) CRUD
- Frontend:
  -  ansible tab for ansible control node systems
  - ansible paths (inventory + playbook) CRUD

Tracks https://github.com/SUSE/spacewalk/issues/14532

## GUI diff

No difference.

After:

ansible-control-node-configuration-paths

- [x] **DONE**

## Documentation
https://github.com/SUSE/spacewalk/issues/14571

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/14532

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
@@ -70,6 +70,9 @@ public class AnsibleController {
         // todo no CSRF?
         post("/manager/api/systems/details/ansible/paths/save",
                 withUser(AnsibleController::saveAnsiblePath));
+
+        post("/manager/api/systems/details/ansible/paths/delete",
+                withUser(AnsibleController::deleteAnsiblePath));
     }
 
     /**
@@ -138,6 +141,30 @@ public class AnsibleController {
         Map<String, Object> data = new HashMap<>();
         data.put("success", ResultJson.success());
         data.put("newPathId", currentPath.getId());
+        return json(res, data);
+    }
+    /**
+     * Delete an Ansible path
+     *
+     * @param req the request object
+     * @param res the response object
+     * @param user the authorized user
+     * @return the string with JSON response
+     */
+    public static String deleteAnsiblePath(Request req, Response res, User user) {
+        AnsiblePathJson json = GSON.fromJson(req.body(), AnsiblePathJson.class);
+
+        try {
+            SystemManager.removeAnsiblePath(SystemManager.lookupAnsiblePathById(json.getId(), user).get(), user);
+        }
+        catch (ValidatorException e) {
+            return json(res, ResultJson.error(
+                    ValidationUtils.convertValidationErrors(e),
+                    ValidationUtils.convertFieldValidationErrors(e)));
+        }
+
+        Map<String, Object> data = new HashMap<>();
+        data.put("success", ResultJson.success());
         return json(res, data);
     }
 
--- a/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
@@ -23,6 +23,7 @@ import static spark.Spark.get;
 import static spark.Spark.post;
 
 import com.redhat.rhn.common.localization.LocalizationService;
+import com.redhat.rhn.common.validator.ValidatorException;
 import com.redhat.rhn.domain.server.Server;
 import com.redhat.rhn.domain.server.ServerFactory;
 import com.redhat.rhn.domain.server.ansible.AnsiblePath;
@@ -31,6 +32,7 @@ import com.redhat.rhn.manager.system.SystemManager;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.suse.manager.webui.controllers.contentmanagement.handlers.ValidationUtils;
 import com.suse.manager.webui.utils.gson.AnsiblePathJson;
 import com.suse.manager.webui.utils.gson.ResultJson;
 
@@ -115,21 +117,26 @@ public class AnsibleController {
      * @param user the authorized user
      * @return the string with JSON response
      */
-    // todo path validation: path format + existence (in case of update) + constraints. do it in SystemManager
-    // and reuse in xmlrpc
     public static String saveAnsiblePath(Request req, Response res, User user) {
         AnsiblePathJson json = GSON.fromJson(req.body(), AnsiblePathJson.class);
 
-        if (json.getId() == null) {
-            SystemManager.createAnsiblePath(AnsiblePath.Type.fromLabel(json.getType()),
-                    json.getMinionId(),
-                    json.getPath(),
-                    user);
+        try {
+            if (json.getId() == null) {
+                SystemManager.createAnsiblePath(AnsiblePath.Type.fromLabel(json.getType()),
+                        json.getMinionId(),
+                        json.getPath(),
+                        user);
+            }
+            else {
+                SystemManager.updateAnsiblePath(json.getId(),
+                        json.getPath(),
+                        user);
+            }
         }
-        else {
-            SystemManager.updateAnsiblePath(json.getId(),
-                    json.getPath(),
-                    user);
+        catch (ValidatorException e) {
+            return json(res, ResultJson.error(
+                    ValidationUtils.convertValidationErrors(e),
+                    ValidationUtils.convertFieldValidationErrors(e)));
         }
         return json(res, ResultJson.success());
     }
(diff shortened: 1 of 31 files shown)


web/html/src/manager/minion/ansible/ansible-control-node.tsx
```diff
@@ -0,0 +1,230 @@
...
+        <p>
+          {t("Ansible Control Node Configuration: add Paths to discover Playbooks and Inventories.")}
+        </p>
+        <div className="col-md-6">
+          <Panel
+            headingLevel="h3"
+            title="Playbooks Paths"
+          >
+            {this.state.playbooksPaths.map(p =>
+                this.state.editPlaybookPath && this.state.editPlaybookPath.path == p.path ?
```
cbbayburt, line 165: Fancy alternative:
```suggestion
                this.state.editPlaybookPath?.path == p.path ?
```

cbbayburt (approved):
Thanks, guys, looks great. Just a couple of minor comments below:

- We need to somehow clearly distinguish for the users between an inventory **file** path, and a playbook **directory** path. It should be clear to them to input a **file** as an inventory path, but input a **directory** as a playbook path. Maybe name them as "Playbook directory" vs "Inventory path"
- We need to include `/etc/ansible/hosts` as an inventory path by default. I'm not sure if it's a good idea to add it in the DB, so maybe we can hardcode it wherever needed. Also, it'd be nice if it could be seen in the configuration UI as a "disabled" (can't be edited/deleted) entry.

web/html/src/manager/minion/ansible/ansible-control-node.tsx
```diff
@@ -0,0 +1,230 @@
...
+                  :
+                  <div className="d-block" key={p.id}>
+                    <pre className="pointer" onClick={() => this.setState({ editPlaybookPath: p })}>
+                      {p.path}<i className="fa fa-edit pull-right" />
+                    </pre>
+                  </div>
+            )}
+            <hr/>
+            <NewAnsiblePath
+              title={t("Add an Playbook path to discover")}
```
cbbayburt, line 184:
```suggestion
              title={t("Add a Playbook path to discover")}
```

ncounter:
>     * We need to somehow clearly distinguish for the users between an inventory **file** path, and a playbook **directory** path. It should be clear to them to input a **file** as an inventory path, but input a **directory** as a playbook path. Maybe name them as "Playbook directory" vs "Inventory path"
>

@cbbayburt on one hand I believe it's partially on the user knowledge side, but I see the point and I agree, and what I can think of to easily fix is to clearly mention the **File** word every time the Inventory is in the game and **Directory** or **Folder** in the Playbook case instead:
- `Playbooks Directory Path`
- `Inventory File Path`

Would it be enough? I am not sure if we also want to validate it properly checking if the value is really of a `dir` or of a `file` depending on the case, or just leave it up to the user to put the right value, because it can also be tricky to check without checking directly on the system.

hustodemon:
I think the wording distinction between "Playbooks Directory Path" and "Inventory File Path" is enough . As @ncounter said, we can't do much about it without reaching minion.

java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
```diff
@@ -52,6 +66,12 @@ private AnsibleController() { }
     public static void initRoutes(JadeTemplateEngine jade) {
         get("/manager/systems/details/ansible",
                 withCsrfToken(withDocsLocale(withUser(AnsibleController::view))), jade);
+
+        get("/manager/api/systems/details/ansible/paths/:minionId",
+                withUser(AnsibleController::listAnsiblePathsByMinion));
+        // todo no CSRF?
```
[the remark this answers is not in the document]
hustodemon: In other controllers, we have `withCsrfToken` decorator, so I don't think so...

ncounter:
> I think the wording distinction between "Playbooks Directory Path" and "Inventory File Path" is enough . As @ncounter said, we can't do much about it without reaching minion.


Good, for now I did [this change](https://github.com/uyuni-project/uyuni/pull/3577/commits/3bc7b9814258badc7b9483d6ad44080d19a28e07). If we need/want more, we can always do it later on.

web/html/src/manager/minion/ansible/ansible-control-node.tsx
```diff
@@ -0,0 +1,227 @@
...
+    }
+    else {
+      this.setState({ newInventoryPath: newPath });
+    }
+  }
+
+  deletePath(path: AnsiblePath) {
+    Network.post(
+      "/rhn/manager/api/systems/details/ansible/paths/delete",
+      JSON.stringify(path.id),
```
Etheryte, line 66: Small nitpick, if this is just a number, I would recommend `.toString()` instead since then it's clear that we're not dealing with any JSON here.

web/html/src/manager/minion/ansible/ansible-control-node.tsx
```diff
@@ -0,0 +1,227 @@
...
+      "application/json"
+    ).promise.then(data => {
+      if (data.success) {
+        if (path.type === "playbook") {
+          this.setState({ playbooksPaths: this.state.playbooksPaths.filter(p => p.id !== path.id) });
```
Etheryte, line 71: I'm not sure what the business logic is here, so checking just in case: is it intentional that `inventoriesPaths` is not set to an empty array in this branch and vice versa in the other one?

web/html/src/manager/minion/ansible/ansible-control-node.tsx
```diff
@@ -0,0 +1,227 @@
...
+    return (
+      <div>
+        {errors}
+        <p>
+          {t("Ansible Control Node Configuration: add paths for Playbook discovery and Inventory files introspection.")}
+        </p>
+        <div className="col-md-6">
+          <Panel
+            headingLevel="h3"
+            title="Playbook Directories"
```
Etheryte, line 157: Missing `t()`.

web/html/src/manager/minion/ansible/ansible-control-node.tsx
```diff
@@ -0,0 +1,227 @@
...
+              placeholder={t("e.g., /srv/playbooks")}
+              newPath={(path: string) => this.newPath("playbook", path)}
+              savePath={() => this.savePath("playbook")}
+            />
+          </Panel>
+        </div>
+        <div className="col-md-6">
+          <Panel
+            headingLevel="h3"
+            title="Inventory Files"
```
Etheryte, line 191: Missing `t()`.

web/html/src/manager/minion/ansible/edit-ansible-path.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { AsyncButton, Button } from "components/buttons";
+
+const EditAnsiblePath = (props) => {
+  return (
+    <>
+      <div className="d-block">
+        <div className="col-md-10">
+          <TextField
+            value={props.ansiblePath.path}
+            onChange={(e: any) => props.editPath(e.target.value.toString())}
```
Etheryte, line 12: Out of curiosity, is the type of `e` not automatically inferred here? You can make a ticket for me to look into this if that's the case, `onChange` should automatically be typed as `React.ChangeEvent<HTMLInputElement>` or something similar imo.

Etheryte (approved):
Frontend part looks good, only a few small questions but I think those are because I'm not familiar enough with the business logic at hand.

web/html/src/manager/minion/ansible/edit-ansible-path.tsx
Etheryte: I pushed a change that should fix this, the prop declaration in `TextField` was slightly wrong.

web/html/src/manager/minion/ansible/ansible-control-node.tsx
ncounter: Yes it is intentional: we are deleting only one entry which can be only in one or in the other set of paths, indeed we enter in an if condition. I could think about something more abstract where we filter both sets by dropping the path we deleted, but in the end it will be just a drop in one of the two, and the other one will not be affected at all. The logic maybe is too narrowed down when I set the state at the beginning by filtering the whole set into two subset sliced by the type. Anyway, this was intentional. Thanks for the heads-up :+1:

java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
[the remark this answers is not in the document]
hustodemon: Ok, it seems it's ok. we include the csrf token in our page and the js code makes use of it.

PR #3582, opened 2021-04-20
QA: Avoid using a console parameter inside a string parameter

srbarrios:
## What does this PR change?

I don't consider good practice to include parameters inside a string parameter.

Example:
When we use `rute-dummy=2.1-1.1` we are using a parameter that defines the version we use, appended by the base name of the package, this could be not supported by other package managers (or it might do). I think that the best we can do for the installation step is to use the full package name that includes the version, so `rute-dummy-2.1-1.1`

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.0
- Manager-4.1

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_retracted_patches.feature
+++ b/testsuite/features/secondary/min_retracted_patches.feature
@@ -6,7 +6,7 @@ Feature: Retracted patches
 
   Scenario: Installed retracted package should show icon in the system packages list
     Given I am authorized as "admin" with password "admin"
-    When I install package "rute-dummy=2.1-1.1" on this "sle_minion"
+    When I install package "rute-dummy-2.1-1.1" on this "sle_minion"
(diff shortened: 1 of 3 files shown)


calancha (approved):
It makes sense.

Bischoff (commented):
Please at least test. I think there was a specific reason for using `=` instead of `-`.

srbarrios:
> Please at least test. I think there was a specific reason for using `=` instead of `-`.

I tested on zypper, but not on apt-get, indeed there it seems we require `=`, my next proposal was to split the parameter in the step definition to use name and version params, but the problem is that this step also support multiple packages.
SO, I will draw a veil over this nitpick.

PR #3285, opened 2021-02-16
Add vendor change option to SP migration cucumber tests 

mseidl:
## What does this PR change?

Port of spacewalk PR#
This adds the allow vendor change option to the SP migration cucumber tests for the sle minion + ssh minion. 

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation
- No documentation needed:  
- 
- [ ] **DONE**

## Test coverage

- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #
Tracks # **add downstream PR, if any**

- [ ] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #3600, opened 2021-04-22
Testsuite: inspect the zypper rotated log as well

calancha:
## What does this PR change?

The zypper.log rotates at 00:00 if its size is >= 10 Mb.
At the time when we inspect this log, its size is >~ 10 Mb, thus
there are some chances that the log is rotated before we search for
the pattern inside it.

Inspect both, zypper.log and the rotated one zypper.log-%Y%m%d.xz.

## Links

Fixes https://github.com/uyuni-project/uyuni/pull/3285
### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/14642

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -190,7 +190,11 @@ end
 
 When(/^vendor change should be enabled for "(?:[^"]*)" on "([^"]*)"$/) do |host|
   node = get_target(host)
-  _result, return_code = node.run("grep -- --allow-vendor-change /var/log/zypper.log")
+  pattern = '--allow-vendor-change'
+  log = '/var/log/zypper.log'
+  rotated_log = "#{log}-#{Time.now.strftime('%Y%m%d')}.xz"
+  _, return_code = node.run("grep -- #{pattern} #{log} || xzdec #{rotated_log} | grep -- #{pattern}")
+
   raise 'Vendor change option not found in logs' unless return_code.zero?
 end
 


calancha:
I have confirmed this bug in my local machine.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -190,7 +190,11 @@
 
 When(/^vendor change should be enabled for "(?:[^"]*)" on "([^"]*)"$/) do |host|
   node = get_target(host)
-  _result, return_code = node.run("grep -- --allow-vendor-change /var/log/zypper.log")
+  pattern = '--allow-vendor-change'
+  log = '/var/log/zypper.log'
+  rotated_log = "#{log}-#{Time.now.strftime('%Y%m%d')}.xz"
+  _, return_code = node.run("grep -- #{pattern} #{log} || xzdec #{rotated_log} | grep -- #{pattern}")
```
srbarrios, line 196: If this method should work for all our clients, do you know if xzdec and grep are installed by default in centos, ubuntu, sle, etc?

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -190,7 +190,11 @@
 
 When(/^vendor change should be enabled for "(?:[^"]*)" on "([^"]*)"$/) do |host|
```
srbarrios, line 191: This method has two regular expressions, but only one parameter.
I guess we should have filter by the first text or something, not sure, but it seems wrong as it is.

calancha: The first regexp is not captured: it is inside a shy group.

testsuite/features/step_definitions/command_steps.rb
calancha: @mseidl He must know if this is only for SLES.

testsuite/features/step_definitions/command_steps.rb
srbarrios: I don't get why we need it then

calancha: The reason is that we use this step at 3 feature files: the matched string inside the shy group is different (it serves as documentation, in the sense that reflects the action we are doing in the context where the step is used). On the other hand, we use a shy group because we don't need such info inside the body of the step.
It's just a matter of style; we could instead capture the string and define a local var `_foo` in the block: `|_foo, host|`.
I prefer the shy group than pollute with an unused variable.

srbarrios: In that case, could we please don't use the double quotes?

calancha: Agreed.

srbarrios: We agreed on fixing it out of this PR, as this is out of scope.

srbarrios (approved):

testsuite/features/step_definitions/command_steps.rb
srbarrios: Done here https://github.com/uyuni-project/uyuni/pull/3605

PR #3609, opened 2021-04-23
Ansible path content for Playbooks and Inventories

ncounter:
## What does this PR change?

Add Playbooks and Inventories subtabs for Ansible to list all playbooks in paths and content of inventory files.

## GUI diff

After:

playbooksContent

Screenshot from 2021-04-26 12-35-10

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14523 https://github.com/SUSE/spacewalk/issues/14522

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/minion/ansible/accordion-path-content.tsx
+++ b/web/html/src/manager/minion/ansible/accordion-path-content.tsx
@@ -1,6 +1,8 @@
 import * as React from "react";
 import { AnsiblePath } from "./ansible-path-type";
 import { Panel } from "components/panels/Panel";
+import Network from "utils/network";
+import { Messages, Utils } from "components/messages";
 
 type PropsType = {
   path: AnsiblePath;
@@ -9,25 +11,52 @@ type PropsType = {
 type StateType = {
   open: boolean; 
   content: Map<string, any>;
+  errors: string[];
 };
+
+function getURL(path: AnsiblePath) {
+  let baseUrl: string;
+  if (path.type === "playbook") {
+    baseUrl = "/rhn/manager/api/systems/details/ansible/discover-playbooks/";
+  }
+  else {
+    baseUrl = "/rhn/manager/api/systems/details/ansible/introspect-inventory/";
+  }
+  return baseUrl + path.id;
+}
 
 class AccordionPathContent extends React.Component<PropsType, StateType> {
   constructor(props) {
     super(props);
 
-    this.state = { open: false, content: new Map<string, any>() };
+    this.state = {
+      open: false,
+      content: new Map<string, any>(),
+      errors: [],
+    };
   }
 
   onOpen() {
-    this.loadPathContent(this.props.path);
-    this.setState({ open: !this.state.open });
-  }
-
-  loadPathContent(path: AnsiblePath) {
-    // Network.get("/rhn/manager/api/systems/details/ansible/path/" + path.id)
-    // .promise.then(data => {
-      this.setState({ content: this.state.content.set(path.path, ["element 1", "element 2"]) });
-    // });
+    if (!this.state.open) {
+      const path: AnsiblePath = this.props.path;
+      if (Object.keys(this.state.content).includes(path.path)) {
+        this.setState({ open: true });
+      }
+      else {
+        Network.get(getURL(path))
+        .promise.then(data => {
+          if (data.success) {
+            this.setState({ content: this.state.content.set(path.path, ["element 1", "element 2"]), open: true});
+          }
+          else {
+            this.setState({ errors: data.messages });
+          }
+        });
+      }
+    }
+    else {
+      this.setState({ open: false });
+    }
   }
 
   render() {
@@ -35,9 +64,12 @@ class AccordionPathContent extends React.Component<PropsType, StateType> {
       <div className="accordion-toggle" onClick={() => this.onOpen()}>
         { this.props.path.path }
       </div>;
+
+    const errors = this.state.errors.length > 0 ? <Messages items={Utils.error(this.state.errors)} /> : null;
     return (
-      <Panel header={header} headingLevel="h5">   
+      <Panel header={header} headingLevel="h5">
         <div>
+          {errors}
           { this.state.open ? 
               <ul>
                 {
--- a/web/html/src/manager/minion/ansible/accordion-path-content.tsx
+++ b/web/html/src/manager/minion/ansible/accordion-path-content.tsx
@@ -60,8 +60,11 @@ class AccordionPathContent extends React.Component<PropsType, StateType> {
 
   render() {
     const header =
-      <div className="panel-heading accordion-toggle" onClick={() => this.onToggle()}>
-        <i className={this.state.open || this.state.loading ? "fa fa-chevron-up" : "fa fa-chevron-down"} />{ this.props.path.path }
+      <div className="panel-heading pointer" onClick={() => this.onToggle()}>
+        <h6>
+          <i className={this.state.open || this.state.loading ? "fa fa-chevron-up" : "fa fa-chevron-right"} />
+          { this.props.path.path }
+        </h6>
       </div>;
 
     const errors = this.state.errors.length > 0 ? <Messages items={Utils.error(this.state.errors)} /> : null;
--- a/web/html/src/manager/minion/ansible/accordion-path-content.tsx
+++ b/web/html/src/manager/minion/ansible/accordion-path-content.tsx
@@ -1,8 +1,8 @@
 import * as React from "react";
 import { AnsiblePath } from "./ansible-path-type";
-import { Panel } from "components/panels/Panel";
 import Network from "utils/network";
 import { Messages, Utils } from "components/messages";
+import { Loading } from "components/utils/Loading";
 
 type PropsType = {
   path: AnsiblePath;
@@ -10,8 +10,9 @@ type PropsType = {
 
 type StateType = {
   open: boolean; 
-  content: Map<string, any>;
+  content: any;
   errors: string[];
+  loading: boolean;
 };
 
 function getURL(path: AnsiblePath) {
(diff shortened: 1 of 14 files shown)


web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -0,0 +1,182 @@
...
+import { Loading } from "components/utils/Loading";
+import { AceEditor } from "components/ace-editor";
+
+type PropsType = {
+  path: AnsiblePath;
+};
+
+type StateType = {
+  open: boolean; 
+  content: any;
```
Etheryte, line 14: I'm guessing this is actually `PlaybookDetails[] | null`?

web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -0,0 +1,182 @@
...
+import { Messages, Utils } from "components/messages";
+import { Loading } from "components/utils/Loading";
+import { AceEditor } from "components/ace-editor";
+
+type PropsType = {
+  path: AnsiblePath;
+};
+
+type StateType = {
+  open: boolean;
```
Etheryte, line 13: We use both `open` and `isOpen` mixed in different contexts for values like this so either works. I would recommend considering `isOpen` though since it's easier to grep for.

web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -0,0 +1,182 @@
...
+      loading: false,
+    };
+  }
+
+  onToggle() {
+    const path: AnsiblePath = this.props.path;
+    if (!this.state.open) {
+      if (this.state.content === null) {
+        this.setState({ loading: true });
+        Network.get(getURL(path))
```
Etheryte, line 54: FYI this will lead to issues if the user a-la double-clicks on the header and then the response will arrive after another change state has already triggered. We have issues like this in a lot of places though so I don't think it's worth solving here right now, we can tackle that somewhere in the undefined future.

web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -0,0 +1,182 @@
...
+        this.setState({ open: true });
+      }
+    }
+    else {
+      this.setState({ open: false, errors: [] });
+    }
+  }
+
+  digestPlaybookPathContent(blob: any) {
+    const content: Map<string, {}> = blob[this.props.path.path];
```
Etheryte, line 81: You probably meant a `Record` here, unless `blob[this.props.path.path]` is an actual [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -0,0 +1,182 @@
...
+        }
+        playbookDetailsList.push(newPlaybookObject);
+      }
+    );
+
+    return playbookDetailsList;
+  }
+
+  renderPlaybookPathContent() {
+    if (this.state.content === null || this.state.content.length === 0) {
```
Etheryte, line 102: Could also do a-la `if (!this.state.content?.length) {}`, but this works too.

web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -0,0 +1,182 @@
...
+              maxLines={40}
+              readOnly={true}
+              mode="yaml"
+              content={this.state.content}
+            ></AceEditor>;
+  }
+
+  render() {
+    const header =
+      <div className="panel-heading pointer" onClick={() => this.onToggle()}>
```
Etheryte, line 149: Would be nice if this was a `<button />` element instead, but this is a small nitpick that can be revisited later.

web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -0,0 +1,182 @@
...
+            this.state.loading?
+              <Loading text={t("Loading content..")} />
+              :
+              this.state.open ?
+                <>
+                  {
+                    this.state.errors.length > 0 ?
+                      errors
+                      :
+                      this.props.path.type === "playbook" ? this.renderPlaybookPathContent() : this.renderInventoryPathContent()
```
Etheryte, line 171: It's been quite a while since I've seen this many nested ternaries. 😄 I don't mind though, this works and I don't really have a better recommendation. If you want to refactor this in the future you could move some of it into a standalone function where you can return early and then have fewer branches. This works as is though.

web/html/src/manager/minion/ansible/ansible-path-type.tsx
```diff
@@ -0,0 +1,11 @@
+export type AnsiblePath = {
+  id: number;
+  minionServerId: number;
+  type: string;
+  path: string;
+}
+
+export function createNewAnsiblePath(fieldsToUpdate: Partial<AnsiblePath>) {
+  const todo: AnsiblePath = {id: -1, minionServerId: -1, type: "", path: ""};
```
Etheryte, line 9: Just in case you missed this, a todo here, but I suspect this is for the future.

Etheryte (approved):
Looks good, I only found some optional suggestions for improvements. 👍

web/html/src/manager/minion/ansible/accordion-path-content.tsx
ncounter: Actually `content` depending on the `pathType` can be a `PlaybookDetails[]` if the type is `playbook` but also a `YAML` formatted `String` if the type is `inventory`, and also it can be `null` if it is not yet assigned.

```

File: web/html/src/manager/minion/ansible/accordion-path-content.tsx (hunk 4 above)
Comment (ncounter) [reply web/html/src/manager/minion/ansible/accordion-path-content.tsx:81]: Actually it is a `Map`, yes :/

```
{
  "/root/ansible-examples/lamp_haproxy": {
		"aws/demo-aws-launch.yml": {
			"fullpath": "/root/ansible-examples/lamp_haproxy/aws/demo-aws-launch.yml"
		},
		"aws/rolling_update.yml": {
			"fullpath": "/root/ansible-examples/lamp_haproxy/aws/rolling_update.yml"
		},
		"aws/site.yml": {
			"fullpath": "/root/ansible-examples/lamp_haproxy/aws/site.yml"
		},
		"provision.yml": {
			"custom_inventory": "/root/ansible-examples/lamp_haproxy/hosts",
			"fullpath": "/root/ansible-examples/lamp_haproxy/provision.yml"
		},
		"rolling_update.yml": {
			"custom_inventory": "/root/ansible-examples/lamp_haproxy/hosts",
			"fullpath": "/root/ansible-examples/lamp_haproxy/rolling_update.yml"
		},
		"site.yml": {
			"custom_inventory": "/root/ansible-examples/lamp_haproxy/hosts",
			"fullpath": "/root/ansible-examples/lamp_haproxy/site.yml"
		}
	}
}
```
and I expect the response to be like the one above because I need to parse properly. Do you still think `Record` makes more sense here? Should I adapt it?

File: web/html/src/manager/minion/ansible/accordion-path-content.tsx (hunk 1 above)
Comment (Etheryte) [reply web/html/src/manager/minion/ansible/accordion-path-content.tsx:14]: I think anything that offers information beyond `any` is good, yes. With `any` we don't get any benefits of type checking.

File: web/html/src/manager/minion/ansible/accordion-path-content.tsx (hunk 4 above)
Comment (Etheryte) [reply web/html/src/manager/minion/ansible/accordion-path-content.tsx:81]: If `content` is a `Map` then later accesses into it should probably be with `const playbookObj = content.get(playbookName)` (not `const playbookObj = content[playbookName]`) etc. Or perhaps I don't understand what you mean?

File: web/html/src/manager/minion/ansible/ansible-path-type.tsx (hunk 8 above)
Comment (ncounter) [reply web/html/src/manager/minion/ansible/ansible-path-type.tsx:9]: LOL, a leftover of a copy-paste from [Partial<Type> example](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype)

File: web/html/src/manager/minion/ansible/accordion-path-content.tsx [hunk 9]
```diff
@@ -0,0 +1,182 @@
...
+  else {
+    baseUrl = "/rhn/manager/api/systems/details/ansible/introspect-inventory/";
+  }
+  return baseUrl + path.id;
+}
+
+interface PlaybookDetails {
+  path: AnsiblePath,
+  fullPath: string,
+  customInventory?: string
```
hustodemon, line 33: Missing `,`.

web/html/src/manager/minion/ansible/accordion-path-content.tsx
hustodemon: +1ed

I find the branching in the component quite complex. I think this component has maybe too much responsibilities (as it takes care of both inventories and playbooks in very specific way. To be concrete: `if (type=='playbook') do something else do something else` is present multiple times in the component.

I'd have some suggestion for improvement (not sure if it's react-ish enough): I think the branching could be avoided if some pieces of logic was injected from outside (namely the `digestXYZ` and `renderXYZ`functions.). In the end, we would use the component like `<AccordionPathContent digestFn=digestPlaybook renderFn=renderPlaybook />`, alternatively in one piece: `<AccordionPathContent contentProcessor=someComponentThatKnowsHowToDigestAndRenderPlaybook />`. Then the accordion component wouldn't care, if it's rendering a playbook or an inventory. Then the user of the component would decide, which logic to inject and the code in the `AccordionPathContent` should be cleaner, in theory. If this sounds confusing, I can try to come up with a patch.

Even if you think the above is a good idea, I don't insist on implementing this as part of the PR (the schedule is very tight), but I'd like to create an issue that would be addressed after the SUMA release deadline.

hustodemon (approved):
I have 2 minor comments and 1 task that should be addressed afterwards, if you agree.

web/html/src/manager/minion/ansible/accordion-path-content.tsx
ncounter: :heart: @hustodemon because you make me feel I was on the right mood at the beginning. I started willing to have a completely ignorant Accordion component without taking care of the content, but (my bad and I am the one to blame) because of time pressure and being in a rush I didn't wanted to spend time thinking about it with the perfect approach and, instead, I didn't wanted to come up with something in the middle that needs to be refactored anyway, so I went for the fast-line implementation, trying just to keep it refactorable.

Initial idea was to have the Accordion to render given children passed by the parent, but I'd like also to come up with an Accordion component able to just do the open/close behavior and also  the loading part. I mean, it would be nice to pass an URL to the Accordion component if it has to load the content once it is called to open, but then it comes the management with the children to render.

@Etheryte : I actually don't like it that much as well, but I tend (not always succeeding though) to extract what has business logic or reusable piece in separate functions, and try to keep everything that is only related to rendering in the `render` method even with some condition (if it does not lead to an unreadable chunk). This is a very borderline case, and I agree it could be written better. Indeed Ithe idea was I extracted the shaping of the content in `renderPlaybookPathContent()` and `renderInventoryPathContent, right because I didn't want to put the two completely different output one next the other, but still the two methods are only rendering pieces, so the ternary condition is visible I can realize the output shape depends on that forking path.

Nevertheless I would love a more readable code (extracting reusable or non context related parts, shorter and maintainable components, ignorant context related components, etc), and I think it would be nice if we can think and plan something about it for the whole codebase (on the React side I mean). @Etheryte would you be interested to take part into some task like this?

cbbayburt (approved):
LGTM :+1:

PR #3615, opened 2021-04-26
Testsuite: Fix union syntaxis for tags

calancha:
## What does this PR change?
It fixes the union of the tags.

The Cucumber syntaxis for tag union (i.e. OR) is:
```
--tags @foo,@bar,@baz
```

While the syntaxis for tag intersection (i.e. AND) is:
```
--tags @foo --tags @bar --tags @baz.
```

- testsuite/Rakefile:
Prefix tags with '@'.
Use the right syntaxis for their union.

## Links

Fixes https://github.com/uyuni-project/uyuni/pull/3614
### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/14665
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/14666

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/Rakefile
+++ b/testsuite/Rakefile
@@ -17,8 +17,8 @@ namespace :cucumber do
       profiles = ENV['PROFILE']
       include_profiles = profiles.nil? ? '--profile default' : '--profile ' + profiles.gsub(',', ' --profile ')
       tags = ENV['TAGS']
-      include_tags =  tags.nil? ? '' : '--tags ' + tags.gsub(',', ' or ')
-      cucumber_opts = %W[#{html_results} #{json_results} #{junit_results} -f rerun #{include_profiles} #{include_tags} --out failed.txt -f pretty -r features]
+      include_tags = tags.nil? ? [] : ['--tags'] + [tags.split(',').map {|x| '@' + x}.join(',')]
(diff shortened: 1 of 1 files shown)


srbarrios:
Could you share a link to the documentation where you learn that?

As in the official documentation we have for Cucumber Ruby, it uses `and` and `or` `(@smoke or @ui) and (not @slow)`
https://cucumber.io/docs/cucumber/api/#tags

calancha:
This will be superseded by a future PR; we will amend susemanager-ci as well, thus some changes here won't be necessary.

srbarrios:
Indeed we need to use the old style yet:
``` 
   -t, --tags TAG_EXPRESSION        Only execute the features or scenarios with tags matching TAG_EXPRESSION.
                                     Scenarios inherit tags declared on the Feature level. The simplest
                                     TAG_EXPRESSION is simply a tag. Example: --tags @dev. When a tag in a tag
                                     expression starts with a ~, this represents boolean NOT. Example: --tags ~@dev.
                                     A tag expression can have several tags separated by a comma, which represents
                                     logical OR. Example: --tags @dev,@wip. The --tags option can be specified
                                     several times, and this represents logical AND. Example: --tags @foo,~@bar --tags @zap.
                                     This represents the boolean expression (@foo || !@bar) && @zap.
```

PR #3618, opened 2021-04-26
QA: Use the correct tags expressions for Cucumber 2.0.0

srbarrios:
## What does this PR change?

As @calancha pointed out here https://github.com/uyuni-project/uyuni/pull/3615 we are using a tags expression that only works on a Cucumber version higher than 2.0.0, and we still need to use the previous style in the command line.

The Rake task expects a list of tags, separated by a comma, including any prepend necessary for the behavior the user wants to have, for instance: `TAGS=@scope_salt,@scope_traditional_client` which will run all the tests having one of those tags at least, another example could be `TAGS=~@scope_traditional_client` which will run all the tests not having this tag.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.0
- Manager-4.1

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/Rakefile
+++ b/testsuite/Rakefile
@@ -15,10 +15,15 @@ namespace :cucumber do
       json_results = "-f json -o output_#{timestamp}-#{filename}.json"
       html_results = "-f html -o output_#{timestamp}-#{filename}.html"
       profiles = ENV['PROFILE']
+      # Our profiles include a --tags keyword in all of them, if we have multiple profiles in the list
+      # it will act as an intersection of tags, not as an union of them.
       include_profiles = profiles.nil? ? '--profile default' : '--profile ' + profiles.gsub(',', ' --profile ')
       tags = ENV['TAGS']
-      include_tags =  tags.nil? ? '' : '--tags ' + tags.gsub(',', ' or ')
-      cucumber_opts = %W[#{html_results} #{json_results} #{junit_results} -f rerun #{include_profiles} #{include_tags} --out failed.txt -f pretty -r features]
+      # We provide the list of tags including the '@', so if necessary we can provide a negative tag '~@' on this list.
+      # We generate the tags list as an union of tags, not as an intersection of them (that would be using multiple times `--tags`)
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

testsuite/Rakefile
```diff
@@ -15,9 +15,14 @@ namespace :cucumber do
...
       profiles = ENV['PROFILE']
+      # Our profiles include a --tags keyword in all of them, if we have multiple profiles in the list
+      # it will act as an intersection of tags, not as an union of them.
       include_profiles = profiles.nil? ? '--profile default' : '--profile ' + profiles.gsub(',', ' --profile ')
       tags = ENV['TAGS']
-      include_tags =  tags.nil? ? '' : '--tags ' + tags.gsub(',', ' or ')
+      # We provide the list of tags including the '@', so if necessary we can provide a negative tag '~@' on this list.
+      # We generate the tags list as an union of tags, not as an intersection of them (that would be using multiple times `--tags`)
+      # WARNING: When we upgrade Cucumber 2.0.0 to a higher version, we need to use the new Tags expression implementation
+      include_tags =  tags.nil? ? '' : "--tags #{tags}"
```
calancha, line 25: I would prefer passing a list in the non-parallel case (jus a feeling, I am not saying that it won't work):
```
include_tags = tags.nil? ? [] '' : ["--tags", tags]
```
and then pass:
```
cucumber_opts = %W[#{html_results} #{json_results} #{junit_results} -f rerun #{include_profiles} --out failed.txt -f pretty -r features] + include_tags
```

calancha (approved):
I've made a small suggestion.

testsuite/Rakefile
srbarrios: I like to keep both tasks with closer code as possible, but you are right, let's use a list.

PR #3643, opened 2021-04-28
CI: add option for not expanding links

jordimassaguerpla:
## What does this PR change?

when comparing if packages need to be updated, by default we were not
expanding links, and as result, all packages were found to need updates.

Let's add an option so we can expand links and we only update packages
that need to. This is needed for testing Pull Requests because
systemsmanagement:Uyuni:Master:TEST:X:CR are links to
systemsmanagement:Uyuni:Master, and without this option, all packages
are found that need to be updated and so the task takes longer.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed

- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14706

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/rel-eng/push-packages-to-obs.sh
+++ b/rel-eng/push-packages-to-obs.sh
@@ -17,7 +17,7 @@ OSCAPI=${OSCAPI:-https://api.suse.de}
 
 OSC="osc ${OSCRC} -A ${OSCAPI}"
 if [ "$OSC_EXPAND" == "TRUE" ];then
-    OSC_CHECKOUT="$OSC checkout"
+    OSC_CHECKOUT="$OSC checkout -e"
 else
     OSC_CHECKOUT="$OSC checkout -u"
 fi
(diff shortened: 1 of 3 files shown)


rel-eng/push-packages-to-obs.sh
```diff
@@ -16,6 +16,11 @@ OSCRC=${OSCRC:+-c $OSCRC}
 OSCAPI=${OSCAPI:-https://api.suse.de}
 
 OSC="osc ${OSCRC} -A ${OSCAPI}"
+if [ "$OSC_EXPAND" == "TRUE" ];then
+    OSC_CHECKOUT="$OSC checkout"
```
juliogonzalez, line 20: What if we add `-e` here, to make it explicit?

juliogonzalez (commented):
Minor comment. Other than that, LGTM.

PR #3645, opened 2021-04-29
Automatic cookie file selection in pkgset beacon and salt bundle support

vzhestkov:
## What does this PR change?

1. `cookie` pillar settings is not required for pkgset beacon anymore, it selects the cookie file path by presence of pkg notify plugin (zyppnotify, yumnotify.py or dpkgnotify)
2.  Implemented support for salt bundle minion version (`venv-salt-minion`) it's preferable if available on the system.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: already covered

## Changelogs
- Add support of salt bundle to pkgset notify beacon
- Add automatic cookie file selection for pkgset beacon

## Re-run a test


mcalmer (approved):

meaksh (commented):
The changes look good to me, but I think we should also change the Java backend to not include `cookie` in the pkgset pillar data as it's not going to be used anymore:

https://github.com/uyuni-project/uyuni/blob/76a121bb820fc0083e330d2285f61ff91afc652d/java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java#L49-L59

@vzhestkov maybe you can also fix that in this PR (the Java part is not complex)

vzhestkov:
@meaksh  Yes, I saw this java part, but I have one fear about it.
If someone will apdate java part prior the susemanager-sls it could lead to a lot of noise in the minions logs. Probably it could be better to leave this part of the code in java for a while?

meaksh:
> @meaksh Yes, I saw this java part, but I have one fear about it.
> If someone will apdate java part prior the susemanager-sls it could lead to a lot of noise in the minions logs. Probably it could be better to leave this part of the code in java for a while?

AFAIK, we release all new packages at the same time during a MU, then customer are expected to upgrade all packages in one shot, so it's not expected the customers only installs the newer version of `susemanager-sls` and not all the others (include spacewalk-java)

java/spacewalk-java.changes
```diff
@@ -1,3 +1,4 @@
+- 'cookie' propertie for pkgset beacon was removed as no longer required
```
meaksh, line 1: Small typo here:

```suggestion
- The 'cookie' property for pkgset beacon was removed as no longer required
```

vzhestkov: yup, now it's better :-) Thanks

PR #3654, opened 2021-04-29
QA: Fix flaky tests waiting for a pick-up of an event of a page without ajax

srbarrios:
## What does this PR change?

Reduce from 10 to 3 the seconds (enough to re-load the page) to wait for a text in a list that will not be updated unless we refresh the page

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.0
- Manager-4.1

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


calancha (approved):
Two consecutive runs at 4.1 w/ this fix don't show the issue.  It looks like a valid fix.

PR #3660, opened 2021-04-29
WIP: Build sources (tito) in parallel

jordimassaguerpla:
## What does this PR change?

Building the sources of all the packages in the CI takes 20 minutes.
If we run this in parallel, it should take 5 minutes instead.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed:

- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/rel-eng/build-one-package-for-obs.sh
+++ b/rel-eng/build-one-package-for-obs.sh
@@ -1,0 +1,121 @@
+#! /bin/bash
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+set -e
+#
+# For all packages in git:/rel-eng/packages (or defined in $PACKAGES)
+# provide tarball, spec and changes in $WORKSPACE/SRPMS/<package>
+#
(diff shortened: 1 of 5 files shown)


jordimassaguerpla:
Do not review yet.. I am testing this idea by running the e2e tests with this PR.

jordimassaguerpla:
It goes "SUPER FAST" but there is an issue when running yarn in parallel for those packages that run yarn install in the setup.sh

jordimassaguerpla:
This is the error https://github.com/yarnpkg/yarn/issues/6312

jordimassaguerpla:
This only saves us 3 minutes. The issue is that what takes most of the time is running yarn. Given we cannot run yarn in parallel, we can not really save more than 3 minutes of 20 of the build step. However, given the total time is 3 hours, this is a gain of 1.7% ... I don't think is worth introducing this.

PR #3664, opened 2021-04-30
Backport #3662 and tagging to master

juliogonzalez:
## What does this PR change?

Backport #3662 and tagging to master


juliogonzalez:
>  The changelog has not been updated. Please update it @juliogonzalez.

image

@hustodemon: not now, but can you have a look at the action after the ansible stuff is done? Seems it's causing a lot of false warnings lately.

PR #3668, opened 2021-04-30
Ansible: Add missing cucumber tests and fix issue on SLS

meaksh:
## What does this PR change?

This PR implements missing cucumber tests for Ansible feature in normal and SSH minion.  

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **tests**

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Tracks: https://github.com/SUSE/spacewalk/issues/14644

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


meaksh:
Unassigning QA people since this PR is targetting `ansible-integration` atm.

testsuite/features/secondary/min_ansible_control_node.feature
```diff
@@ -4,6 +4,9 @@
 @scope_ansible
 Feature: Operate an Ansible control node in a normal minion
 
+  Scenario: Pre-requisite: Deploy test playbooks and inventory file
```
srbarrios, line 7: I noticed that the full feature gives 3 whitespaces instead of 2 in the steps of the scenario.
Could you fix it, please?

testsuite/features/secondary/min_ansible_control_node.feature
```diff
@@ -12,31 +15,62 @@ Feature: Operate an Ansible control node in a normal minion
...
      Then I should see a "Ansible Control Node Configuration" text
 
-  Scenario: Pre-requisite: Deploy test playbooks and inventory file
-     Given I deploy testing playbooks and inventory files to "sle_minion" 
+  Scenario: Configure some inventory and playbooks path
+     Given I am on the Systems overview page of this "sle_minion"
+     When I follow "Ansible" in the content area
+     Then I should see a "Ansible Control Node Configuration" text
+     And I enter "/srv/playbooks/" as "new_playbook_path_input"
+     And I click on button with id "new_playbook_path_save"
```
srbarrios, line 34: Could we use the text of the button instead of the ID?
It's more readable for a user following these steps.

testsuite/features/secondary/min_ansible_control_node.feature
```diff
@@ -12,31 +15,62 @@ Feature: Operate an Ansible control node in a normal minion
...
+     And I enter "/srv/playbooks/example_playbook2_orion_dummy/hosts" as "new_inventory_path_input"
+     And I click on button with id "new_inventory_path_save"
+
+  Scenario: Display inventories
+     Given I am on the Systems overview page of this "sle_minion"
+     When I follow "Ansible" in the content area
+     And I follow "Inventories" in the content area
+     And I wait until I see "/srv/playbooks/example_playbook2_orion_dummy/hosts" text
+     Then I click on "/srv/playbooks/example_playbook2_orion_dummy/hosts" text in Ansible paths
```
srbarrios, line 43: I don't care too much about this, but there are team members that could not like using a "Then" here :)

testsuite/features/secondary/min_ansible_control_node.feature
```diff
@@ -12,31 +15,62 @@ Feature: Operate an Ansible control node in a normal minion
...
+     Then I wait until I see "/srv/playbooks/example_playbook2_orion_dummy/example_playbook2_orion_dummy.yml" text
+
+  Scenario: Run a playbook using custom inventory
+     Given I am on the Systems overview page of this "sle_minion"
+     When I follow "Ansible" in the content area
+     And I follow "Playbooks" in the content area
+     And I wait until I see "/srv/playbooks" text
+     And I click on "/srv/playbooks" text in Ansible paths
+     And I wait until I see "/srv/playbooks/example_playbook2_orion_dummy/example_playbook2_orion_dummy.yml" text
+     Then I click on "example_playbook2_orion_dummy/example_playbook2_orion_dummy.yml"
```
srbarrios, line 61: So we use "When" for all the actions triggered to the system provoking a "reaction", and "Then" when we expect some result from this action.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1292,6 +1292,9 @@ def server_secret
   source = File.dirname(__FILE__) + '/../upload_files/ansible/playbooks/example_playbook2_orion_dummy/hosts'
   return_code = file_inject(target, source, dest + "hosts")
   raise 'File injection failed' unless return_code.zero?
+  source = File.dirname(__FILE__) + '/../upload_files/ansible/playbooks/example_playbook2_orion_dummy/example_file.txt'
```
srbarrios, line 1295: What about having this other structure?
```
.
└── ansible
    └── playbooks
        ├── orion_dummy
        │   ├── file.txt
        │   ├── hosts
        │   └── playbook_orion_dummy.yml
        └── ping
            └── playbook_ping.yml
```

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -244,6 +244,21 @@
   end
 end
 
+#
+# Click on a button with a given "id"
+When(/^I click on button with id "([^"]*)"$/) do |button_id|
+  find(:xpath, ".//button[@id=\"#{button_id}\"]").click
+end
+
```
srbarrios, line 252: You don't need a special step to click by Id.

The step `I click on "zzz"` will check in both text and ids.

image

srbarrios: In any case, as I commented in another part of the PR, I strongly recommend to search by text and not by ID.
As it's easier to follow the steps by someone that reads the .feaure file.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -244,6 +244,21 @@
...
+#
+# Click on a text which appears inside of <div> with
+# the given "id"
+When(/^I click on "([^"]*)" text in Ansible paths$/) do |text|
+  within(:xpath, '//div[@id="ansible-path-content"]') do
+    find('h6', text: text).click
+  end
+end
+
```
srbarrios, line 261: I would say we don't need this special step.

Could you instead try to use this step in your scenario?

``` #
# Click on a button which appears inside of <div> with
# the given "id"
When(/^I click on "([^"]*)" in element "([^"]*)"$/) do |text, element_id|
  within(:xpath, "//div[@id=\"#{element_id}\"]") do
    click_button_and_wait(text, match: :first)
  end
end
```

testsuite/features/secondary/min_ansible_control_node.feature
srbarrios, line 43: Try to use this step instead:
`I click on "/srv/playbooks/example_playbook2_orion_dummy/hosts" in element "ansible-path-content"`

Not sure if it will work, due to the extra filter 'h6' you added, but it's worth trying it, and safe an additional step.

testsuite/features/secondary/min_ansible_control_node.feature
meaksh: Okk :+1:

testsuite/features/secondary/min_ansible_control_node.feature
meaksh: No, I think we cannot. 

Here we have two different "Save" buttons (with same text) in the same page, and the `<div>` where they are contained have no ids to distinguise between them. So I cannot reuse one of the existing step and had to implement this new one.

testsuite/features/secondary/min_ansible_control_node.feature
meaksh: Got it

testsuite/features/secondary/min_ansible_control_node.feature
meaksh: Thanks for clarifying!

testsuite/features/step_definitions/common_steps.rb
meaksh: Yeah, It makes sense to simplify this a bit. In any case, I'm keeping `playbook_ping.yml` in the main `playbooks` as I don't want that one to be contained in a custom folder as the `orion_dummy` one.

testsuite/features/step_definitions/navigation_steps.rb
meaksh: Ahhh! I can pass the `id` directly to the "I click on" step.

Ok. I'm fixing this - Thanks!

testsuite/features/step_definitions/navigation_steps.rb
meaksh: I tried that one but it didn't work since it's expecting a button.

In this case, what we have here is not a button but a "collapsible" div that it's only containing `<h6>text</h6>` where I need to click to get the content loaded.

That's why I had to implement this particular step to deal with those "paths" boxes in the Ansible UI.

testsuite/features/secondary/min_ansible_control_node.feature
meaksh: As mentioned above, that didn't work for me

ktsamis (approved):
Approving to get some runs over the weekend, we can take another look Monday and adjust as needed.

meaksh:
@srbarrios I addressed your comments and tackled some of your suggestions. I'm merging this already to have testsuite executions by the weekend. 

Let's talk on Monday in case there is something else to be tackled here. Thanks!

PR #3670, opened 2021-04-30
CI: Build each package in a docker container in parallel

jordimassaguerpla:
## What does this PR change?

Build packages in parallel, going from 18-20 minutes to 5 minutes!

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed

- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/14762

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/testing/automation/push-to-obs.sh
+++ b/susemanager-utils/testing/automation/push-to-obs.sh
@@ -70,7 +70,6 @@ test -n "$PACKAGES" || {
     PACKAGES=$(ls "$GITROOT"/rel-eng/packages/)
 }
 echo "Starting building and submission at $(date)"
-date
 [ -d ${GITROOT}/logs ] || mkdir ${GITROOT}/logs
 for p in ${PACKAGES};do
     CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${p}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
--- a/susemanager-utils/testing/automation/push-to-obs.sh
+++ b/susemanager-utils/testing/automation/push-to-obs.sh
@@ -22,11 +22,14 @@ help() {
   echo "  -n  If used, update PROJECT instead of the projects specified with -d,"
   echo "      for example, if you want to package only the changes from a PR on"
   echo "      a separate project"
-  echo "  -e  If used, when checking out projects from obs, links will be expanded. Useful for comparing packages that are links" 
+  echo "  -e  If used, when checking out projects from obs, links will be expanded. Useful for comparing packages that are links"
+  echo "  -x  Enable parallel builds"
   echo ""
 }
 
-while getopts ":d:c:p:n:vthe" opts; do
+PARALLEL_BUILD="FALSE"
+
+while getopts ":d:c:p:n:vthex" opts; do
   case "${opts}" in
     d) DESTINATIONS=${OPTARG};;
     p) PACKAGES=${OPTARG};;
@@ -35,6 +38,7 @@ while getopts ":d:c:p:n:vthe" opts; do
     t) TEST="-t";;
     n) OBS_TEST_PROJECT="-n ${OPTARG}";;
     e) EXTRA_OPTS="-e";;
+    x) PARALLEL_BUILD="TRUE";;
     h) help
        exit 0;;
     *) echo "Invalid syntax. Use ${SCRIPT} -h"
@@ -59,8 +63,22 @@ if [ ! -f ${CREDENTIALS} ]; then
 fi
 
 INITIAL_CMD="/manager/susemanager-utils/testing/automation/initial-objects.sh"
-CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${PACKAGES}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
 CHOWN_CMD="/manager/susemanager-utils/testing/automation/chown-objects.sh $(id -u) $(id -g)"
-
 docker pull $REGISTRY/$PUSH2OBS_CONTAINER
-docker run --rm=true -v "$GITROOT:/manager" -v "/srv/mirror:/srv/mirror" --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${INITIAL_CMD}; ${CMD}; RET=\${?}; ${CHOWN_CMD} && exit \${RET}"
+
+test -n "$PACKAGES" || {
+    PACKAGES=$(ls "$GITROOT"/rel-eng/packages/)
+}
+echo "Starting building and submission at $(date)"
+date
+[ -d ${GITROOT}/logs ] || mkdir ${GITROOT}/logs
+for p in ${PACKAGES};do
+    CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${p}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
+    if [ "$PARALLEL_BUILD" == "TRUE" ];then
+        docker run --rm=true -v $GITROOT:/manager -v /srv/mirror:/srv/mirror --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${CMD}; RET=\${?}; exit \${RET}" | tee ${GITROOT}/logs/${p}.log &
+    else
+        docker run --rm=true -v $GITROOT:/manager -v /srv/mirror:/srv/mirror --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${CMD}; RET=\${?}; exit \${RET}" | tee ${GITROOT}/logs/${p}.log
+    fi
+done
+echo "End of task at ($(date). Logs for each package at ${GITROOT}/logs/"
+wait
--- a/web/.yarnrc
+++ b/web/.yarnrc
@@ -1,0 +1,5 @@
+yarn-path "../susemanager-frontend/susemanager-nodejs-sdk-devel/build/yarn/yarn-1.22.4.js"
+--install.cwd .
+--clean.cwd .
+--audit.cwd .
+--autoclean.cwd .
--- a/susemanager-utils/testing/automation/publish-rpms.sh
+++ b/susemanager-utils/testing/automation/publish-rpms.sh
@@ -15,10 +15,10 @@ usage()
 
 while getopts ":p:r:a:d:" opts;do
     case "${opts}" in
-        p) echo "PPPP";obs_project=${OPTARG};;
-        r) echo "RRRR";obs_repo=${OPTARG};;
-        a) echo "AAA";obs_arch=${OPTARG};;
-        d) echo "DDDD";repo_dir=${OPTARG};;
+        p) obs_project=${OPTARG};;
+        r) obs_repo=${OPTARG};;
+        a) obs_arch=${OPTARG};;
+        d) repo_dir=${OPTARG};;
         \?) usage;exit -1;;
     esac
 done


jordimassaguerpla:
NO NEED TO REVIEW, This is just a test.

jordimassaguerpla:
and voila! we went from 18 minutes to 5 minutes on building the RPMs!

The trick, the .yarnrc: https://github.com/uyuni-project/uyuni/pull/3670/files#diff-0eaf1805c221f60b902f23b298b89f8d55dd1c2f2a6771e41a8103dd3fc3879f

I will prepare a cleaner Pull Request

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +63,20 @@ if [ ! -f ${CREDENTIALS} ]; then
...
 
 INITIAL_CMD="/manager/susemanager-utils/testing/automation/initial-objects.sh"
-CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${PACKAGES}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
 CHOWN_CMD="/manager/susemanager-utils/testing/automation/chown-objects.sh $(id -u) $(id -g)"
-
 docker pull $REGISTRY/$PUSH2OBS_CONTAINER
-docker run --rm=true -v "$GITROOT:/manager" -v "/srv/mirror:/srv/mirror" --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${INITIAL_CMD}; ${CMD}; RET=\${?}; ${CHOWN_CMD} && exit \${RET}"
+
+test -n "$PACKAGES" || {
+  PACKAGES=$(ls "$GITROOT"/rel-eng/packages/ \
```
juliogonzalez, line 70: Unless @mcalmer says otherwise, I think the packages you are filtering should be removed. Those for oracle, for sure, and the others I don't remember tagging them in four years.

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +63,20 @@ if [ ! -f ${CREDENTIALS} ]; then
...
+          | grep -v -x -e heirloom-pkgtools -e rhnclient -e smartpm -e jabberd-selinux -e oracle-rhnsat-selinux -e oracle-selinux -e oracle-xe-selinux -e spacewalk-monitoring-selinux -e spacewalk-proxy-selinux -e spacewalk-selinux -e cx_Oracle -e apt-spacewalk -e perl-DBD-Oracle -e spacewalk-jpp-workaround)
+}
+echo "Building packages in parallel"
```
juliogonzalez, line 73: You are echoing this, even with you are not building (and submitting) in parallel.
```suggestion
echo "Starting building and submission at $(date)"
```

Feel free to improve it if you really want to show when you are building in parallel. Easiest way is by modifying how you are managing DOCKER_CMD.:

```
x) DOCKER_BG="-d";;
```
and
```
if [ "${DOCKER_BG}" == "-d" ]; then
  echo "Starting building and submission (parallel) at $(date)"
else
  echo "Starting building and submission (serial) at $(date)"
fi
```
Of course you will need to take care of the other mentions to `DOCKER_CMD`.

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +63,20 @@ if [ ! -f ${CREDENTIALS} ]; then
...
+date
+for p in ${PACKAGES};do
+    CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${p}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
+    $DOCKER_CMD run --rm=true -v "$GITROOT:/manager" -v "/srv/mirror:/srv/mirror" --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${CMD}; RET=\${?}; exit \${RET}" &
+done
+echo "Done"
```
juliogonzalez, line 79: Done what? A package or all packages? Since this before the wait, I am unsure. Looks to me this will print when all builds are launched, not when they are complete. Maybe this should be moved to line 82?

And maybe:

```suggestion
echo "End of task at ($(date). Logs for each package at ${GITROOT}/logs/"
```

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +63,20 @@ if [ ! -f ${CREDENTIALS} ]; then
...
+
+test -n "$PACKAGES" || {
+  PACKAGES=$(ls "$GITROOT"/rel-eng/packages/ \
+          | grep -v -x -e heirloom-pkgtools -e rhnclient -e smartpm -e jabberd-selinux -e oracle-rhnsat-selinux -e oracle-selinux -e oracle-xe-selinux -e spacewalk-monitoring-selinux -e spacewalk-proxy-selinux -e spacewalk-selinux -e cx_Oracle -e apt-spacewalk -e perl-DBD-Oracle -e spacewalk-jpp-workaround)
+}
+echo "Building packages in parallel"
+date
+for p in ${PACKAGES};do
+    CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${p}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
+    $DOCKER_CMD run --rm=true -v "$GITROOT:/manager" -v "/srv/mirror:/srv/mirror" --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${CMD}; RET=\${?}; exit \${RET}" &
```
juliogonzalez, line 77: How is this going to print the logs? I understand that if running with `-x` it will be a mess, impossible to debug. I suggest you save logs for each package, but that will require some work.

First of all, this script should create a folder at `$GITROOT`, say for example `logs` (the folder must be removed if it exists from a previous run).

Then this line should look more or less like:

```suggestion
    $DOCKER_CMD run --rm=true -v "$GITROOT:/manager" -v "/srv/mirror:/srv/mirror" --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${CMD}; RET=\${?}; exit \${RET}" | tee ${GITROOT}/logs/${p}.log &
```
This way, if something fails we can at least inspect the log for each package.

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +63,20 @@ if [ ! -f ${CREDENTIALS} ]; then
...
+done
+echo "Done"
+date
```
juliogonzalez, line 80:
```suggestion
```

web/.yarnrc
```diff
@@ -0,0 +1,5 @@
+yarn-path "../susemanager-frontend/susemanager-nodejs-sdk-devel/build/yarn/yarn-1.22.4.js"
```
juliogonzalez, line 1: Why do we need this?

juliogonzalez (changes requested):
Some comments.

About PR metadata:
- Is this really WIP? If not, please remove it from the title.
- Please complete the full template. All "DONE" checkboxes must be marked before the PR is merged even if nothing was done (we should probably change the text), and things that don't apply must be removed (`GUI diff` should say `No difference.`)

susemanager-utils/testing/automation/push-to-obs.sh
mcalmer: I think most of them are already removed. This list is very old and comes from the time where we were based on spacewalk and we did not want to differ too much. Time to cleanup

jordimassaguerpla:
@juliogonzalez thanks for your suggestions. I will address them :) . One doubt that I have is if we should change the existing push-to-obs or create a new push-to-obs-parallel . What do you think?

My concern is changing what we have running for master at this moment or doing it only for PR and keeping it as it is for master tests.

jordimassaguerpla:
@mcalmer @juliogonzalez I have addressed your revisions. However, when running the tests afterwards, the metadata is wrong, so I need to address it before merging this one (https://github.com/SUSE/spacewalk/issues/14626#issuecomment-832126417) .

Then, there is one remaining doubt from my side, and it is the strategy we want to use. Should we merge this to master? Or would you prefer me to add an option to not run in parallel or create a new script so we keep what we have in master?

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +60,18 @@ if [ ! -f ${CREDENTIALS} ]; then
...
 CHOWN_CMD="/manager/susemanager-utils/testing/automation/chown-objects.sh $(id -u) $(id -g)"
-
 docker pull $REGISTRY/$PUSH2OBS_CONTAINER
-docker run --rm=true -v "$GITROOT:/manager" -v "/srv/mirror:/srv/mirror" --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${INITIAL_CMD}; ${CMD}; RET=\${?}; ${CHOWN_CMD} && exit \${RET}"
+
+test -n "$PACKAGES" || {
+    PACKAGES=$(ls "$GITROOT"/rel-eng/packages/)
+}
+echo "Starting building and submission at $(date)"
+date
```
juliogonzalez, line 70:
```suggestion
```

web/.yarnrc
juliogonzalez: Why was this marked as resolved, but without a reply? :-)

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -22,10 +22,11 @@ help() {
   echo "  -n  If used, update PROJECT instead of the projects specified with -d,"
   echo "      for example, if you want to package only the changes from a PR on"
   echo "      a separate project"
-  echo "  -e  If used, when checking out projects from obs, links will be expanded. Useful for comparing packages that are links" 
+  echo "  -e  If used, when checking out projects from obs, links will be expanded. Useful for comparing packages that are links"
```
juliogonzalez, line 25: I am now completely confused. Where did the option for parallel is?

I don't see `-d` at the docker calls either, what happened :-?

juliogonzalez (changes requested):
I am confused. Wasn't this PR supposed to be about parallelism? That's what the description still has. What happened to the code? :-?

And maybe you should update the title if this is about parallelism in the end?

susemanager-utils/testing/automation/push-to-obs.sh
jordimassaguerpla: Hi,

the parallelism is accomplished with using amperstand and wait. Using the -d was redundant, was not printing the logs in stderr nor stdout, and, overall, wait was not waiting for them to finish.

I want to use wait because I want to ensure the process does not end until all packages have been pushed to obs.

I updated the title to reflect this is about parallelism.

Sorry I removed the option "x" and made parallelism mandatory cause the "x" option was not working well for testing this PR with the current jenkins job implementation.

My question remains, should we make it default, or should the parallelism be optional?

jordimassaguerpla:
@juliogonzalez About yarn.js ..I did wrote a reply .... but can't see it :) so I guess something went wrong. Anyway, this is where yarn.js is in our code, I don't think we need to duplicate that ... Alternatively, I could move the yarn.js to another directory if you prefer, would you like me to do that?

susemanager-utils/testing/automation/push-to-obs.sh
juliogonzalez: I'd certainly make it optional with `-x`. What's the problem with the Jenkins job implementation?

juliogonzalez:
> @juliogonzalez About yarn.js ..I did wrote a reply .... but can't see it :) so I guess something went wrong. Anyway, this is where yarn.js is in our code, I don't think we need to duplicate that ... Alternatively, I could move the yarn.js to another directory if you prefer, would you like me to do that?

But why do we need to add this to this PR? What does it have to do with parallelism?

jordimassaguerpla:
> > @juliogonzalez About yarn.js ..I did wrote a reply .... but can't see it :) so I guess something went wrong. Anyway, this is where yarn.js is in our code, I don't think we need to duplicate that ... Alternatively, I could move the yarn.js to another directory if you prefer, would you like me to do that?
> 
> But why do we need to add this to this PR? What does it have to do with parallelism?

You can't build 2 packages with yarn the way it is setup right now, because they both share the same git checkout and so the same yarnrc. Thus, yarn uses the same cache and when we run the setup.sh for the susemanager-frontend and the one for the susemanager-web package, one removes the cache the other is using. This happens always because these are the 2 packages that take more time, so they always end up being run in parallel. We need to add a yarnrc in the web directory for this reason, so when building the susemanager-web package, this one does not use the other one cache.

susemanager-utils/testing/automation/push-to-obs.sh
jordimassaguerpla: Ok, I'll do that. The problem is that the current job can't have the "-x" without having this code merged first, it is a "fish eating its tail" problem ;) . But I can certainly do this once we are ready to merge it.

juliogonzalez:
> > > @juliogonzalez About yarn.js ..I did wrote a reply .... but can't see it :) so I guess something went wrong. Anyway, this is where yarn.js is in our code, I don't think we need to duplicate that ... Alternatively, I could move the yarn.js to another directory if you prefer, would you like me to do that?
> > 
> > 
> > But why do we need to add this to this PR? What does it have to do with parallelism?
> 
> You can't build 2 packages with yarn the way it is setup right now, because they both share the same git checkout and so the same yarnrc. Thus, yarn uses the same cache and when we run the setup.sh for the susemanager-frontend and the one for the susemanager-web package, one removes the cache the other is using. This happens always because these are the 2 packages that take more time, so they always end up being run in parallel. We need to add a yarnrc in the web directory for this reason, so when building the susemanager-web package, this one does not use the other one cache.

Good, this is what I wanted to know :-)

susemanager-utils/testing/automation/push-to-obs.sh
juliogonzalez: I don't understand, what's wrong with the jobs? If we make parallelism optional with `-x`, the jobs will work before the PR is merged and after the PR is merged, as parallelism won't be used because `-x` is not used.

When the PR is merged, then you can configure them to use `-x`.

Where's the "fish eating its tail" problem :-?

jordimassaguerpla:
@juliogonzalez I've been testing this. I think we are good to go now :+1: Can we merge this one?

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +63,22 @@ if [ ! -f ${CREDENTIALS} ]; then
...
 CHOWN_CMD="/manager/susemanager-utils/testing/automation/chown-objects.sh $(id -u) $(id -g)"
-
 docker pull $REGISTRY/$PUSH2OBS_CONTAINER
-docker run --rm=true -v "$GITROOT:/manager" -v "/srv/mirror:/srv/mirror" --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${INITIAL_CMD}; ${CMD}; RET=\${?}; ${CHOWN_CMD} && exit \${RET}"
+
+test -n "$PACKAGES" || {
+    PACKAGES=$(ls "$GITROOT"/rel-eng/packages/)
+}
+echo "Starting building and submission at $(date)"
+date
```
juliogonzalez, line 73: Why is this needed? We already have the date at the echo.

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -59,8 +63,22 @@ if [ ! -f ${CREDENTIALS} ]; then
...
+[ -d ${GITROOT}/logs ] || mkdir ${GITROOT}/logs
+for p in ${PACKAGES};do
+    CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${p}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
+    if [ "$PARALLEL_BUILD" == "TRUE" ];then
+        docker run --rm=true -v $GITROOT:/manager -v /srv/mirror:/srv/mirror --mount type=bind,source=${CREDENTIALS},target=/tmp/.oscrc $REGISTRY/$PUSH2OBS_CONTAINER /bin/bash -c "${CMD}; RET=\${?}; exit \${RET}" | tee ${GITROOT}/logs/${p}.log &
```
juliogonzalez, line 78: I wonder if `&` could not be a variable, so you don't need to duplicate the whole command. But maybe you already tried and it didn't work, so feel free to ignore.

juliogonzalez (commented):
Some minor things

susemanager-utils/testing/automation/push-to-obs.sh
jordimassaguerpla: I already tried :( and didn't worked. I also tried having the whole command into a variable, but it was a mess with the `"` and the `\"` and the `\\` .... and didn't get it working ... too many indirections I guess

susemanager-utils/testing/automation/push-to-obs.sh
jordimassaguerpla: fixed.

jordimassaguerpla:
@juliogonzalez I fixed the minor issue and commented the other one which you said I could ignore. Anything else or can this be merged?thanks

juliogonzalez (approved):
LGTM

juliogonzalez:
Merging. No need to wait for `java_pgsql_tests` (your next task BTW :-D)

PR #3679, opened 2021-05-03
Fix for pkgset beacon test

vzhestkov:
## What does this PR change?

Fix for test unit of `pkgset` beacon from PR https://github.com/uyuni-project/uyuni/pull/3645

## GUI diff

No difference.

## Documentation

- [ ] **DONE**

## Test coverage
- Unit tests were fixed

## Links

Fix for test unit related to this PR https://github.com/uyuni-project/uyuni/pull/3645

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
+++ b/susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
@@ -22,12 +22,12 @@ def test_validate():
     '''
     Test validate() function
     '''
-    res, msg = pkgset.validate({'cookie': '/bogus/path'})
-    assert res is True
-    assert msg == 'Configuration validated'
-
-    for cfg in [{}, {'bogus': 'data'}]:
-        res, msg = pkgset.validate(cfg)
+    with patch.object(pkgset.os.path, "exists", MagicMock(return_value=True)):
+        res, msg = pkgset.validate({})
+        assert res is True
+        assert msg == 'Configuration validated'
+    with patch.object(pkgset.os.path, "exists", MagicMock(return_value=False)):
+        res, msg = pkgset.validate({})
         assert res is False
         assert msg == 'Cookie path has not been set.'
 


susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
```diff
@@ -22,12 +22,12 @@ def test_validate():
...
     Test validate() function
     '''
-    res, msg = pkgset.validate({'cookie': '/bogus/path'})
-    assert res is True
-    assert msg == 'Configuration validated'
-
-    for cfg in [{}, {'bogus': 'data'}]:
-        res, msg = pkgset.validate(cfg)
+    with patch.object(pkgset.os.path, "exists", MagicMock(return_value=True)):
+        res, msg = pkgset.validate({})
```
agraul, line 26: I didn't see the original PR until now, but since `config` is now unused, I think it shouldn't be removed from `pkgset.validate`'s  API.

susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
```diff
@@ -22,12 +22,12 @@ def test_validate():
     '''
     Test validate() function
     '''
-    res, msg = pkgset.validate({'cookie': '/bogus/path'})
-    assert res is True
-    assert msg == 'Configuration validated'
-
-    for cfg in [{}, {'bogus': 'data'}]:
-        res, msg = pkgset.validate(cfg)
+    with patch.object(pkgset.os.path, "exists", MagicMock(return_value=True)):
```
agraul, line 25: nitpick, no need to change: I think a regular `Mock` would be enough here as `os.path.exists` is a function and not a class with dunder methods.

susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
agraul: Or is this part of Salt's api? Then I'd rename it to `_config`, but no need to change anything here.

vzhestkov: There is no need to check the config now, the only `interval` value is make sense now, but it's managed outside the exact beacon. The test unit previously just contains the checks for empty config or confing which doesn't contain `cookie`. In current state the only important point if the plugin itself is available on the system or not.

mcalmer (approved):

susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
agraul: I understand that. The "issue" I have is that it is unclear why the function `validate` accepts a parameter `config` which it then ignores. If Salt itself passes a `dict` to `validate()` as part of its beacon API, we can't remove `config` from the function definition. In that case, I would mark it as "a parameter that is expected to be ignored", either by prefixing it with an `_` or by "deleting" it as the first thing in the function body:

```python
def validate(config):
    del config # passed by Salt, not needed in validate
    ...
```
The upside of doing this is that it is clear that `config` was not forgotten and linters don't complain.


If the `config` parameter is not required by Salt, I would remove it.

vzhestkov: I see what you mean. Will fix it in the next PR. Most probably need to change the messages as current messages in validate is not reflecting real meaning.

PR #3680, opened 2021-05-03
Ansible fixes

hustodemon:
See the commit msgs.

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/templates/minion/ansible-path-content.jade
+++ b/java/code/src/com/suse/manager/webui/templates/minion/ansible-path-content.jade
@@ -6,7 +6,10 @@ script(type='text/javascript').
     window.csrfToken = "#{csrf_token}";
     window.timezone = "#{h.renderTimezone()}";
     window.localTime = "#{h.renderLocalTime()}";
-    window.actionChains = !{actionChains};
+
+if actionChains
+    script(type='text/javascript').
+        window.actionChains = !{actionChains};
 
 script(src='/javascript/ace-editor/src-min-noconflict/ace.js?cb=#{webBuildtimestamp}')
 script(src='/javascript/ace-editor/src-min-noconflict/ext-modelist.js?cb=#{webBuildtimestamp}')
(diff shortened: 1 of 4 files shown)


java/code/src/com/suse/manager/webui/templates/minion/ansible-path-content.jade
```diff
@@ -6,7 +6,10 @@ script(type='text/javascript').
     window.csrfToken = "#{csrf_token}";
     window.timezone = "#{h.renderTimezone()}";
     window.localTime = "#{h.renderLocalTime()}";
-    window.actionChains = !{actionChains};
+
+if actionChains
```
cbbayburt, line 10: How does this happen? I thought it would be an empty array in that case

hustodemon: Apparently, when `actionChains` is empty in the model, then it renders to a whitespace, leading to invalid JS:

```javascript
window.actionChains = ;
```

If you know a better trick, I'm all ears :smile:

ncounter: It would be, if we at least add the `key` in the `result Map`, like we do [here](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java#L145) --> [that means here](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/controllers/MinionController.java#L425). Instead in [`Inventories` call](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java#L162) we are completely missing the `key`

hustodemon: In the end I'm starting to incline that this would be likely the best thing: have 2 separate `jade` templates for these pages..

cbbayburt: Ah of course, I forgot that this is used for both pages.

cbbayburt: In this case, I think this makes sense as it is. So it's fine by me :+1:

ncounter (approved):

PR #3682, opened 2021-05-03
Master: ansible inventory parsing to show registered systems

ncounter:
## What does this PR change?

Parse Inventory content to distinguish already registered and managed systems against Uyuni or unknown hostnames.

## GUI diff

No difference.

Before:
inventoryContent

After:
Screenshot from 2021-05-04 12-30-19

- [x] **DONE**

## Documentation
- No documentation needed: included in ansible documentations

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes # https://github.com/SUSE/spacewalk/issues/14753

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
@@ -40,16 +40,18 @@ import com.suse.manager.webui.controllers.contentmanagement.handlers.ValidationU
 import com.suse.manager.webui.utils.gson.AnsiblePathJson;
 import com.suse.manager.webui.utils.gson.AnsiblePlaybookExecutionJson;
 import com.suse.manager.webui.utils.gson.AnsiblePlaybookIdJson;
+import com.suse.manager.webui.utils.gson.SimpleMinionJson;
 import com.suse.utils.Json;
-
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
-
 import org.apache.log4j.Logger;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.SafeConstructor;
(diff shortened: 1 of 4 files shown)


java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
```diff
@@ -345,6 +372,20 @@ public static String introspectInventory(Request req, Response res, User user) {
...
 
+    /**
+     * Parse Ansible Inventory content to look for host names
+     *
+     * @param inventoryMap the Ansible Inventory content
+     * @return the Set of hostnames
+     */
+    public static Set<String> parseInventoryAndGetHostnames(Map<String, Map<String, Object>> inventoryMap) {
+        // Assumption: "_meta" and the nested "hostvars" keys always present in the map to contains all hostnames
+        if (inventoryMap.keySet().contains("_meta") && inventoryMap.get("_meta").keySet().contains("hostvars")) {
```
hustodemon, line 383: - Nit: you can use `containsKey` method instead of getting the `keySet`.
- I think it'd be also worth checking if the nested thing is a map, just to be more sure, that the casting doesn't blow up.

```suggestion
        if (inventoryMap.containsKey("_meta") && inventoryMap.get("_meta").containsKey("hostvars") &&
                inventoryMap.get("_meta").get("hostvars") instanceof Map) {
```

java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
```diff
@@ -332,7 +334,32 @@ public static String introspectInventory(Request req, Response res, User user) {
...
+                                serverName -> {
+                                    if (ServerFactory.findByFqdn(serverName).isPresent()) {
+                                        Server server = ServerFactory.findByFqdn(serverName).get();
+                                        registeredServers.add(new SimpleMinionJson(server.getId(), server.getName()));
+                                    }
+                                    else {
+                                        unknownHostNames.add(serverName);
+                                    }
+                                }
+                        );
```
hustodemon, line 355: This could be simplified by using `ifPresentOrElse`. As a bonus you avoid
- multiple calls to expensive `findByFqdn`
- `Optional.get()` call.

```suggestion
                        hostvars.forEach(serverName -> ServerFactory.findByFqdn(serverName).ifPresentOrElse(
                                s -> registeredServers.add(new SimpleMinionJson(s.getId(), s.getName())),
                                () -> unknownHostNames.add(serverName))
                        );
```

hustodemon (approved):
I've just added 2 comments about the java code (I'd suggest that you don't just commit the suggestions using github, but you use your IDE. I'm not sure about the formatting :grimacing: ). Otherwise it looks good to me! :+1:

java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
ncounter: It was exactly my intention but I always forget how to use it and the fact that `ifPresentOrElse` exists :sweat_smile: 
Thank you very much for the pointer :+1:

java/code/src/com/suse/manager/webui/controllers/AnsibleController.java
ncounter: It makes definitely sense, thank you.

PR #3696, opened 2021-05-04
Accept package list refresh scheduled by (none)

Bischoff:
## What does this PR change?

When the user adds/removes a package on a client, Salt automatically triggers a package list refresh.

But in that case, the action is triggered by `(none)`. Our wait step was waiting for an action triggered by `admin`.

This PR makes the test not care about who triggered the action.

## Links

Ports:
* 4.0: SUSE/spacewalk#14770
* 4.1: SUSE/spacewalk#14771

## Changelogs

- [x] No changelog needed


hustodemon (approved):
LGTM!

lkotek (approved):
LGTM. This change should be fine.

PR #3702, opened 2021-05-05
Ubuntu bootstrap repo fixes

mcalmer:
## What does this PR change?

reprepro is picky with packages having the same name, but different checksums.
To avoid aborting, we copy the old packages to debs, add or overwrite them
with new found packages and run reprepro again (bsc#1184330)

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **manual**

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14499
Tracks https://github.com/SUSE/spacewalk/pull/14777

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager/src/mgr-create-bootstrap-repo
+++ b/susemanager/src/mgr-create-bootstrap-repo
@@ -2,6 +2,7 @@
 
 import sys
 import os
+import glob
 import shutil
 import subprocess
 import time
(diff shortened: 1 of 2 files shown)


susemanager/src/mgr-create-bootstrap-repo
```diff
@@ -489,6 +490,18 @@ def create_repo(label, options, mgr_bootstrap_data, additional=[]):
     log("The bootstrap repo should contain the following packages: {0}".format(packagelist), 2)
 
     debs_dir = os.path.join(destdirtmp, 'debs')
+    if repotype == 'deb' and not options.dryrun:
+        # reprepro is picky with packages having the same name, but different checksums.
+        # To avoid aborting, we copy the old packages to debs and add or overwrite them
+        # with new found packages and run reprepro again (bsc#1184330)
```
agraul, line 496: I don't really understand the last part of the comment "run reprepro again". I think `reprepro` was not executed before getting to this point.

agraul (approved):
I think the script could be refactored a bit, currently it looks a lot like deb support was forced on top of the existing rpm support. But that's outside of this PR.

I haven't run the code, but from reading the bug report and the code I think the PR works correctly. By not letting `rerepro` itself deal with incremental updates the checksum error should go away.

susemanager/src/mgr-create-bootstrap-repo
mcalmer: Every reposync trigger this script and we operate on the already created repository from the run before.
For rpm with createrepo_c this is not a problem, but reprepro is picky and has conditions where it fail.
We hit such a condition in the bug.

This change is to workaround such problems.

PR #3704, opened 2021-05-05
Testuite: Wait for package list refresh rigth after the SP migration

calancha:
## What does this PR change?
Wait for the package list refresh right after the SP migration.

The SP migration fully completes once a package list refresh finish; only
after that, the client overview page will show the migrated products.

### Ports
- Manager-4.1: 

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/init_clients/sle_minion.feature
+++ b/testsuite/features/init_clients/sle_minion.feature
@@ -54,9 +54,10 @@ Feature: Bootstrap a Salt minion via the GUI
 
   Scenario: Check the migration is successful for this minion
     Given I am on the Systems overview page of this "sle_spack_migrated_minion"
-    When I follow "Events"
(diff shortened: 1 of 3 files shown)


Bischoff (changes requested):
I don't like it much.

First, the "package list refresh" should be caused by something else, before.

Second, the logic is obscure. Why don't we keep waiting for the package migration first?

Third, do we really need to wait for the second refresh ?

What about putting two steps:
 1) wait for the migration event
 2) wait for one "package list refresh"

calancha:
Tested and it works.  Let's merge it and think about how to refactor it later.

Bischoff:
> Tested and it works. Let's merge it and think about how to refactor it later.

Please don't. I will do another PR and test it.

Bischoff:
New version under test: #3708

calancha:
It will be superseded by Eric PR.

PR #3722, opened 2021-05-07
Add support of DISABLE_LOCAL_REPOS=0 in the bootstrap script for salt minions (bsc#1185568)

vzhestkov:
## What does this PR change?

Implements the possiblity to set DISABLE_LOCAL_REPOS=0 in the bootstrap script to prevent disabling local repos on the salt minions like it works for the traditional clients.

Adds missing SALT_RUNNING environment variable to salt minion config to prevent duplicated package list refreshes.

## GUI diff

No difference.

## Documentation

## Links

Fixes bsc#1185568
https://jira.suse.com/browse/PM-2615

## Changelogs
- Add support of DISABLE_LOCAL_REPOS=0 for salt minions (bsc#1185568)
- Add missing environment variable SALT_RUNNING for pkg module
  to the minion configuration

- Add support for 'disable_local_repos' salt minion config parameter
  (bsc#1185568)

## Re-run a test

### DIFF
--- a/spacewalk/certs-tools/rhn_bootstrap_strings.py
+++ b/spacewalk/certs-tools/rhn_bootstrap_strings.py
@@ -1052,6 +1052,14 @@ enable_legacy_startup_events: False
 enable_fqdns_grains: False
 start_event_grains: [machine_id, saltboot_initrd, susemanager]
 mine_enabled: False
+EOF
+    if [ "$DISABLE_LOCAL_REPOS" -eq 0 ]; then
+        echo "Do not disable local repos"
+        cat <<EOF >>"$SUSEMANAGER_MASTER_FILE"
+disable_local_repos: False
+EOF
+    fi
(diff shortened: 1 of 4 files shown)


mcalmer (approved):

meaksh (approved):

Shirocco88:
Hallo,
would it be possible to add this feature to the bootstrap web-ui:
https://uyuni.srv.local/rhn/manager/systems/bootstrap

Best regards
Shirocco88

vzhestkov:
@Shirocco88, what do you mean by adding this feature to web UI? Do you want to add a kind of checkbox indicating that disabling local repos is not required? It's easier to implement it with pillar data, but I'm not sure anyway that we can recommend this way of using repos on the managed systems. In general it means that managed system has an access to the repos which are out of control of SUSE Manager Server. Such usecase can lead to some side effects on using CLM for example or some others.

PR #3725, opened 2021-05-07
Fix SP migration

Bischoff:
## What does this PR change?

This PR:
* SP migrates repositories inherited from sumaform
* does cosmetic changes, removes dead code, and synchronizes branches

Port of SUSE/spacewalk#14813, SUSE/spacewalk#14820, and SUSE/spacewalk#14836

## Links

Ports:
* 4.0: none
* 4.1: SUSE/spacewalk#14836

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -772,6 +772,15 @@ When(/^I register this client for SSH push via tunnel$/) do
 end
 
 # Repositories and packages management
+When(/^I migrate the non-SUMA repositories on "([^"]*)"$/) do |host|
+  node = get_target(host)
+  # use sumaform states to migrate to latest SP the system repositories:
+  node.run('salt-call --local --file-root /root/salt/ state.apply repos')
+  # disable again the non-SUMA repositories:
+  node.run("for repo in $(zypper lr | awk 'NR>7 && !/susemanager:/ {print $3}'); do zypper mr -d $repo; done")
+  # node.run('salt-call state.apply channels.disablelocalrepos') does not work
+end
+
 When(/^I (enable|disable) (the repositories|repository) "([^"]*)" on this "([^"]*)"((?: without error control)?)$/) do |action, _optional, repos, host, error_control|
   node = get_target(host)
   _os_version, os_family = get_os_version(node)
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -536,29 +536,6 @@ When(/^I check test channel$/) do
   step %(I check "Test Base Channel" in the list)
 end
 
-When(/^I check the child channel "([^"]*)"$/) do |channel|
-  find(:xpath, "//i[@class='fa fa-angle-right']").click unless has_xpath?("//i[@class='fa fa-angle-down']", wait: 60)
-  checkbox = find(:xpath, "//label[contains(.,'#{channel}')]/..//input", match: :first, wait: 60)
-  checkbox.set(true)
-end
-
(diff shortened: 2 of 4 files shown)


testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -536,27 +536,13 @@
...
-  _os_version, os_family = get_os_version(node)
-  if os_family =~ /^ubuntu/
-    steps %(
-      When I check the child channel "main"
-      And I check the child channel "main-updates"
-    )
-  elsif os_family =~ /^centos/
-    step %(I check the child channel "DVD")
+When(/^I add "([^"]*)" to the child channels to migrate$/) do |channel|
+  within(:xpath, "//ul[@class='list-channel']/li[strong/text()='Optional Child Channels:']") do
```
srbarrios, line 540: I would prefer something simpler in this xpath. First, if I'm not wrong there is only one place in the section where we can match a text '#{channel}' followed by an input element, so there is no need for this within statement. Second, it would be enough to filter using the class name 'list-channel' without including the text part, as this text has more probabilities to be changed.

Bischoff: 1) Good remark. Right, but I would remove the "list-channel" instead

2) As Can said, relying on internals is bad practice, it's better to rely on something visible by the user
  (and internals change too)

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -536,27 +536,13 @@
...
+    path = "./ul/li[a/text()='#{channel}']/input"
+    if has_xpath?(path)
```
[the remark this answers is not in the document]
srbarrios: I don't agree with that, as the text in the step definition don't mention that extra behavior.
We should let that extra behavior outside this method, and just check the child channel to migrate.
Also I don't like to use the word "I add" we should instead use "I check" to keep some homogeneity on the actions we make on the steps we call.

testsuite/features/step_definitions/navigation_steps.rb
srbarrios: I prefer internals than text for filtering like here, I don't consider a bad practice to rely on ids and classnames, anyhow as said, this is a completely unnecessary filter.

testsuite/features/step_definitions/navigation_steps.rb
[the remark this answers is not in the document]
Bischoff: Hmmm, right, then it should be removed at feature level then. You convinced me.

But I have a feeling something is wrong. As a started, we don't get same lists of optional channels you and me. I need to take the issue from the beginning again.  

add => check  OK

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: yes it's unnecessary.

Bischoff: done

testsuite/features/step_definitions/navigation_steps.rb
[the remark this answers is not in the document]
Bischoff: "check" done, for the rest it will depend on my testing now.

srbarrios:
Converting to Draft ;)

srbarrios (approved):

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -772,6 +772,13 @@
 end
 
 # Repositories and packages management
+When(/^I migrate the non-SUMA repositories on "([^"]*)"$/) do |host|
```
srbarrios, line 775: I would say "update" instead of "migrate".

PR #3726, opened 2021-05-07
Master - ansible frontend tests

ncounter:
## What does this PR change?

Add frontend tests for the new Ansible Integration code

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: 

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/minion/ansible/accordion-path-content.test.tsx
+++ b/web/html/src/manager/minion/ansible/accordion-path-content.test.tsx
@@ -28,7 +28,7 @@ describe("AccordionPathContent summary", () => {
             name: "minion.tf.local"
           }
         ],
-        dump: "_meta:\n  hostvars:\n    minion.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-1.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-2.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-3.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-4.tf.local: {ansible_ssh_private_key_file: /etc/ansible/some_ssh_key}\nall:\n  children: [mygroup1, mygroup2, ungrouped]\nmygroup1:\n  hosts: [my-ansible-managed-client-1.tf.local, my-ansible-managed-client-2.tf.local]\nmygroup2:\n  hosts: [my-ansible-managed-client-3.tf.local]\nungrouped:\n  hosts: [minion.tf.local, my-ansible-managed-client-4.tf.local]\n"
+        dump: "The Inventory dump content for ACE Editor"
       }
     };
     // server loading inventory details on click
--- a/web/html/src/manager/minion/ansible/accordion-path-content.test.tsx
+++ b/web/html/src/manager/minion/ansible/accordion-path-content.test.tsx
@@ -1,0 +1,53 @@
+import { render, server, click, waitFor } from "utils/test-utils";
+import AccordionPathContent from "./accordion-path-content";
+import { createNewAnsiblePath } from "./ansible-path-type";
+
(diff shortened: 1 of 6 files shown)


web/html/src/manager/minion/ansible/accordion-path-content.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+          "my-ansible-managed-client-3.tf.local",
+          "my-ansible-managed-client-4.tf.local"
+        ],
+        knownSystems: [
+          {
+            id: 10000,
+            name: "minion.tf.local"
+          }
+        ],
+        dump: "_meta:\n  hostvars:\n    minion.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-1.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-2.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-3.tf.local: {ansible_ssh_private_key_file: /etc/ansible/my_ansible_private_key}\n    my-ansible-managed-client-4.tf.local: {ansible_ssh_private_key_file: /etc/ansible/some_ssh_key}\nall:\n  children: [mygroup1, mygroup2, ungrouped]\nmygroup1:\n  hosts: [my-ansible-managed-client-1.tf.local, my-ansible-managed-client-2.tf.local]\nmygroup2:\n  hosts: [my-ansible-managed-client-3.tf.local]\nungrouped:\n  hosts: [minion.tf.local, my-ansible-managed-client-4.tf.local]\n"
```
Etheryte, line 31: This makes me a bit wary, if this test becomes outdated, how would I generate this long string from scratch? Or what parts of it are needed etc? If all of this is required, a comment would help a lot please.

web/html/src/utils/test-utils/mock.tsx
```diff
@@ -9,3 +9,13 @@ jest.mock("components/datetimepicker", () => {
     },
   };
 });
+
+// Mock the datetime picker to avoid it causing issues due to missing jQuery/Bootstrap parts
```
Etheryte, line 13: Out of date comment.

web/html/src/manager/minion/ansible/accordion-path-content.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+import AccordionPathContent from "./accordion-path-content";
+import { createNewAnsiblePath } from "./ansible-path-type";
+
+const API_INVENTORY_DETAILS = "/rhn/manager/api/systems/details/ansible/introspect-inventory/2";
+
+describe("AccordionPathContent summary", () => {
+  test("Render accordion collapsed element", async () => {
+    const path = createNewAnsiblePath({id: 1, minionServerId: 1000, path: "/srv/ansible/playbooks", type: "playbook"});
+    const { getByText } = render(<AccordionPathContent path={path} onSelectPlaybook={() => {}} />);
+    getByText("/srv/ansible/playbooks");
```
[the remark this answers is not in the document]
ncounter: Makes sense, I like it even better. I actually copy-pasted it so I'll fix also other places where I will find it. Thanks :+1:

web/html/src/manager/minion/ansible/accordion-path-content.test.tsx
ncounter: This is true, and it's a good corner-case catch. This piece of data was copy-pasted from the real-life response from the server on a working environment, but actually this exact value (the `dump`) it is not part of the test as it is the value that is passed to the `ACE Editor`, the one that we mock up because we don't want to test. I will just remove the content and use a simple mock string like `dump: "The Inventory dump content for ACE Editor"`

Etheryte: Ah okay I see yeah. Thanks for the explanation. 👍

ncounter:
@juliogonzalez this PR adds test coverage only, so as soon as you want (RC1/GMC etc) it can be merged, it does not affect the product with any change.

PR #3728, opened 2021-05-07
Sort mgr sync products

mcalmer:
## What does this PR change?

Sort the product list in mgr-sync output

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager/src/mgr_sync/product.py
+++ b/susemanager/src/mgr_sync/product.py
@@ -47,11 +47,11 @@ class Product(object):
         return not (self == other)
 
     def __lt__(self, other):
-        selfIsSUSE = self.friendly_name.startswith("SUSE")
-        otherIsSUSE = other.friendly_name.startswith("SUSE")
-        if selfIsSUSE and not otherIsSUSE:
+        self_is_SUSE = self.friendly_name.startswith("SUSE")
+        other_is_SUSE = other.friendly_name.startswith("SUSE")
+        if self_is_SUSE and not other_is_SUSE:
             return False
-        if not selfIsSUSE and otherIsSUSE:
+        if not self_is_SUSE and other_is_SUSE:
             return True
         return (self.friendly_name.lower() < other.friendly_name.lower())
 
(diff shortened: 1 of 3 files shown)


susemanager/src/mgr_sync/product.py
```diff
@@ -39,6 +40,21 @@ def __init__(self, data):
...
+
+    def __ne__(self, other):
+        return not (self == other)
+
+    def __lt__(self, other):
+        selfIsSUSE = self.friendly_name.startswith("SUSE")
+        otherIsSUSE = other.friendly_name.startswith("SUSE")
+        if selfIsSUSE and not otherIsSUSE:
+            return False
+        if not selfIsSUSE and otherIsSUSE:
```
agraul, line 54: [nitpick]: I would prefer snake_case to be consistent with Python's style:
```suggestion
        self_is_SUSE = self.friendly_name.startswith("SUSE")
        other_is_SUSE = other.friendly_name.startswith("SUSE")
        if self_is_SUSE and not other_is_SUSE:
            return False
        if not self_is_SUSE and other_is_SUSE:
```

agraul (approved):
Looks good in general, I only have a small nitpick. 

As always, feel free to ignore my nitpicks :).

susemanager/src/mgr_sync/product.py
```diff
@@ -39,6 +40,21 @@ def __init__(self, data):
     def __repr__(self):
         return self.to_ascii_row()
 
+    def __eq__(self, other):
+        return (self.friendly_name == other.friendly_name)
+
+    def __ne__(self, other):
+        return not (self == other)
```
agraul, line 47: As far as I know this is redundant, the default for `__ne__` is already to take the inverse of `__eq__` ([doc](https://docs.python.org/3/reference/datamodel.html#object.__ne__))

mcalmer: I copied an example and they also said this, but implemented it anyway :-) I will try to remove it

mcalmer: removed

PR #3757, opened 2021-05-13
Drop compatibility support of IE8 and olders

ncounter:
## What does this PR change?

Drop compatibility support of IE8 and olders

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: 

- [x] **DONE**

## Test coverage
- No tests: 

- [x] **DONE**

## Links

Fixes #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


ncounter:
@moio is it okay if we do this now? IE8 is pretty old and we declare we do not support other than Edge and latest browser versions so, anything against?

Etheryte:
I think the changelog entry may be a bit misleading — realistically we already don't support IE8 etc, we simply have the old shims in place. I've outlined some technical details about this below.  

The changeset looks good otherwise, thanks a lot for taking the time to clean this up. 👍

<hr>

Context:  

Not speaking for JSP pages, React-based pages are currently built based on [`@babel/preset-env`](https://babeljs.io/docs/en/babel-preset-env) which we don't configure so it uses the default values from [`browserslist`](https://github.com/browserslist/browserslist).  

The default value is `> 0.5%, last 2 versions, Firefox ESR, not dead` which means IE11 is the oldest one we happen to support (run `npx browserslist 'defaults'` to see the full list). This is unofficial though and just happens to be included, the official support list [is here](https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/installation/general-requirements.html#installation-general-supportedbrowsers) and reads  

> SUSE Manager is supported on:
> 
> - Latest Firefox browser shipped with SLES
> - Latest Chrome browser on all operating systems
> - Latest Edge browser shipped with Windows

There was a ticket discussing all of this somewhere but I can't find it right now, if I recall correctly then the conclusion was that the current setup is completely fine for all involved. In short, we don't officially support IE but IE11 might work just as a happenstance.  

This is all simply for context though so everyone is up to date (and also so I have something to refer to the next time when this topic comes up).

Etheryte (approved):

moio:
We do not support IE since very long time, before SUSE Manager 4.0 actually.

All Uyuni development should concentrate on latest Edge, Firefox and Chrome by and large.

https://github.com/uyuni-project/uyuni-docs/blob/master/modules/installation/pages/general-requirements.adoc#supported-browsers-for-the-susemgr-webui

ncounter:
@moio :+1: thanks for the feedback.

@Etheryte thanks for bringing up a bit of context to make it clear, I see we are all on the same page. I'll adapt a bit the changelog entry and I'd say we are good to go merging this. We are in Spring, that's the time clean-ups are meant to happen :smile:

juliogonzalez:
Strictly speaking this is not bugfixing. Are we sure we want to merge this and risk more failures at the testsuite, or can we leave it for post-GM?

ncounter:
> Strictly speaking this is not bugfixing. Are we sure we want to merge this and risk more failures at the testsuite, or can we leave it for post-GM?

@juliogonzalez I knew you would say that :) 

It also does not change anything actually.
My point is: I want to maintain the less code as possible. Shipping this means it will stay for the whole lifecycle, and this is stale :/

Context: it is not bugfixing but it is stale code we never reach **unless** someone is using/testing it with IE8 or earlier. If this is the case, I'd like to see the failure anyway and **do not** use or test it with IE8 or earlier.

From my POV we should stay confident and stable merging this, but you know, the :crystal_ball: is not under my hand :sweat_smile:

juliogonzalez:
> > Strictly speaking this is not bugfixing. Are we sure we want to merge this and risk more failures at the testsuite, or can we leave it for post-GM?
> 
> @juliogonzalez I knew you would say that :)
> 
> It also does not change anything actually.
> My point is: I want to maintain the less code as possible. Shipping this means it will stay for the whole lifecycle, and this is stale :/
> 
> Context: it is not bugfixing but it is stale code we never reach **unless** someone is using/testing it with IE8 or earlier. If this is the case, I'd like to see the failure anyway and **do not** use or test it with IE8 or earlier.
> 
> From my POV we should stay confident and stable merging this, but you know, the  is not under my hand

Ok. Keep an eye on the testsuite, just in case (only 20 failures now)

PR #3773, opened 2021-05-17
Fix get_os_version return nil instead of raise

ktsamis:
## What does this PR change?

Fix get_os_version return nil instead of raise because when we call the method we expect nil, nil instead of a raise. Without this it doesn't work. Also the `false` is needed to use it as array instead of string

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/client_stack.rb
+++ b/testsuite/features/support/client_stack.rb
@@ -50,12 +50,13 @@ end
 # We get these data decoding the values in '/etc/os-release'
 # rubocop:disable Metrics/AbcSize
 def get_os_version(node)
-  os_family_raw, code = node.run('grep "^ID=" /etc/os-release')
-  raise "No os-release found in the node: #{node.hostname}" unless code.zero?
+  os_family_raw, code = node.run('grep "^ID=" /etc/os-release', false)
+  return nil,nil unless code.zero?
   os_family = os_family_raw.strip.split('=')[1]
   return nil, nil if os_family.nil?
   os_family.delete! '"'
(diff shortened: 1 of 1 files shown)


Bischoff (approved):
LGTM, thanks

lkotek (approved):
LGTM, thanks :-)

testsuite/features/support/client_stack.rb
```diff
@@ -50,12 +50,13 @@ def check_restart(host, node, time_out)
...
+  os_family_raw, code = node.run('grep "^ID=" /etc/os-release', false)
+  return nil, nil unless code.zero?
   os_family = os_family_raw.strip.split('=')[1]
   return nil, nil if os_family.nil?
   os_family.delete! '"'
-  os_version_raw = node.run('grep "^VERSION_ID=" /etc/os-release')
+  os_version_raw, code = node.run('grep "^VERSION_ID=" /etc/os-release', false)
+  return nil, nil unless code.zero?
   os_version = os_version_raw.strip.split('=')[1]
   return nil, nil if os_version.nil?
```
srbarrios, line 61: We could at least return the os_family, which is not nil at that point of the code.
But ok, as we were in a hurry.

ktsamis: I don't think it would be useful half the info, either everything or nil

PR #3781, opened 2021-05-19
Do not render the section toolbar if nothing is expected

ncounter:
## What does this PR change?

Hide the `Spacewalk Section Toolbar` if it has no children to render

## GUI diff

Before:
image

After:
Screenshot from 2021-05-19 12-29-02

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: UI bugfix

- [x] **DONE**

## Links

Fixes # https://github.com/SUSE/spacewalk/issues/13756

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/components/panels/InnerPanel.tsx
+++ b/web/html/src/components/panels/InnerPanel.tsx
@@ -5,26 +5,41 @@ import { SectionToolbar } from "components/section-toolbar/section-toolbar";
 type Props = {
   title: string;
   icon: string;
-  buttons: React.ReactNode;
-  buttonsLeft?: React.ReactNode;
+  buttons: React.ReactNode[];
+  buttonsLeft?: React.ReactNode[];
   children: React.ReactNode;
 };
 
 function InnerPanel(props: Props) {
+  let toolbar;
+
+  if (props.buttons?.length > 0 || props.buttonsLeft?.length !== 0) {
+    toolbar =
+      <SectionToolbar>
+        {
+          props.buttonsLeft?.length !== 0 ?
+            <div className="selector-button-wrapper">
+              <div className="btn-group">{props.buttonsLeft}</div>
+            </div>
+            : null
+        }
+        {
+          props.buttons?.length > 0 ?
+            <div className="action-button-wrapper">
+              <div className="btn-group">{props.buttons}</div>
+            </div>
+            : null
+        }
+      </SectionToolbar>;
+  }
+
   return (
     <div>
       <h2>
         <i className={`fa ${props.icon}`} />
         {props.title}
       </h2>
-      <SectionToolbar>
-        <div className="selector-button-wrapper">
-          <div className="btn-group">{props.buttonsLeft}</div>
-        </div>
-        <div className="action-button-wrapper">
-          <div className="btn-group">{props.buttons}</div>
-        </div>
-      </SectionToolbar>
+      { toolbar }
       <div className="row">
         <div className="panel panel-default">
           <div className="panel-body">{props.children}</div>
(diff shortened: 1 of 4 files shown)


juliogonzalez:
@ncounter add the label only when:
- Tests are passing
- At least a reviewer approves the PR (if reviewers are required).

I am removing the label for now, as otherwise I can merge it by mistake.

web/html/src/components/panels/InnerPanel.tsx
```diff
@@ -5,26 +5,41 @@ import { SectionToolbar } from "components/section-toolbar/section-toolbar";
...
 };
 
 function InnerPanel(props: Props) {
+  let toolbar;
+
+  if (props.buttons?.length > 0 || (props.buttonsLeft !== undefined && props.buttonsLeft?.length > 0)) {
+    toolbar =
+      <SectionToolbar>
+        {
+          props.buttonsLeft !== undefined && props.buttonsLeft?.length > 0 ?
```
Etheryte, line 20: The undefined check should be superfluous with the optional chaining in the next part of the condition.

web/html/src/components/panels/InnerPanel.tsx
```diff
@@ -5,26 +5,41 @@ import { SectionToolbar } from "components/section-toolbar/section-toolbar";
 type Props = {
   title: string;
   icon: string;
-  buttons: React.ReactNode;
-  buttonsLeft?: React.ReactNode;
+  buttons: React.ReactNode[];
```
Etheryte, line 8: Is there a reason to only limit this to arrays? If we want to check for non-empty arrays we could do something along the lines of

```tsx
if (typeof buttons === 'undefined' || (Array.isArray(buttons) && buttons.length === 0)) {
  return null;
}
return ...;
```

web/html/src/components/panels/InnerPanel.tsx
ncounter: It way Typescript complaining about it: `Object is possibly undefined` :thinking:

web/html/src/components/panels/InnerPanel.tsx
ncounter: I kind of wanted to limit it because the props is `buttons` so I expect a collection (`Array`) of items. And we also render them already into a `<div className="btn-group">` button wrapper. So I want to make sure we shape the prop in the right way otherwise we do not know what to expect and what will be rendered in the end.

Etheryte: I more so meant that code-wise there will be no difference, both will still be rendered using `{props.buttons}`, so I was wondering if there was something I was missing. I don't have a strong preference though, this works well too.

web/html/src/components/panels/InnerPanel.tsx
Etheryte: Ah okay I see, my bad sorry. When `buttonsLeft` is `undefined`, `props.buttonsLeft?.length` will be undefined and Typescript is unhappy about the `undefined > 0` comparison. The problem will go away if you check for `!== 0` instead.

Etheryte (approved):
Looks good, thanks for fixing this. 👍

parlt91 (approved):
It looks good to me. Only thing I wondered about was the `createButton` -> `buttons` rename. Did it bother you that much that it's an array but named as a singular button? :-) (no objection though)

ncounter:
> It looks good to me. Only thing I wondered about was the `createButton` -> `buttons` rename. Did it bother you that much that it's an array but named as a singular button? :-) (no objection though)

Ehehe, you caught me. Yes I am a picky guy sometimes :sweat_smile: 
No actually it was not bothering me that much but, for consistency, I fixed that because otherwise one can just shape the variable as a single element instead, and pass it to the component as a prop but it should not be a single one, it expects an Array. Or the other way around, you see the singular button variable name, but then you see an Array of buttons and you (where _you_ means someone new to the code) don't understand how it will work eventually through the component.

web/html/src/components/panels/InnerPanel.tsx
ncounter: Thanks for the hint, I fixed it accordingly :wink:

juliogonzalez:
Merging.

PR #3787, opened 2021-05-20
Install proxy after bootstrap and after refreshing metadata

Bischoff:
## What does this PR change?

Follow-up of #3487

Installing proxy's software should happen after the bootstrap, so we pick the packages from the server.

## Changelogs

- [x] No changelog needed


srbarrios (approved):

Bischoff:
won't work, closing

PR #3789, opened 2021-05-20
Web calendar fix

parlt91:
## What does this PR change?

Don't display the web calendar in schedule details if the schedule has no maintenance calendar selected.
Remove some unused code from `MaintenanceWindowData`

## GUI diff

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: **Bugfix**

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/maintenance/details/schedule-details.tsx
+++ b/web/html/src/manager/maintenance/details/schedule-details.tsx
@@ -23,9 +23,9 @@ import { WebCalendar } from "manager/maintenance/calendar/web-calendar";
 
 type MaintenanceScheduleDetailsProps = {
   id: number;
-  name: string;
+  name: string; // Name of the maintenance schedule
   type: "SINGLE" | "MULTI";
-  calendarName: string;
+  calendarName: string | undefined;
   onDelete: (item: { name: string }) => Promise<any>;
   onMessage: (messages: MessageType[]) => void;
   clearMessages: (messages: void) => void;
@@ -79,8 +79,8 @@ const MaintenanceScheduleDetails = (props: MaintenanceScheduleDetailsProps) => {
 
 type OverviewProps = {
   id: number,
-  name: string;
-  calendarName: string;
+  name: string; // Name of the maintenance schedule
+  calendarName: string | undefined;
   type: "SINGLE" | "MULTI";
   onMessage: (messages: MessageType[]) => void;
   clearMessages: (messages: void) => void;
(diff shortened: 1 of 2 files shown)


web/html/src/manager/maintenance/details/schedule-details.tsx
```diff
@@ -102,22 +102,24 @@ const MaintenanceScheduleOverview = (props: OverviewProps) => {
...
-        </div>
-        <div className="panel-body">
-          <WebCalendar
-            id={props.id}
-            type={"schedule"}
-            messages={props.onMessage}
-            clearMessages={props.clearMessages}
-            responseError={props.responseError}
-          />
+      { props.calendarName &&
```
Etheryte, line 105: Is this checking for an empty string or can `calendarName` be undefined too? Right now the prop says there will always be a value, just a heads up it might be out of date. But if we're only checking for an empty string then this is fine.  

One additional comment regarding this prop, from simply reading the code it isn't very clear what the difference between `name` and `calendarName` is?

Etheryte (approved):
One small question about ambiguity. Looks good. 👍

web/html/src/manager/maintenance/details/schedule-details.tsx
parlt91: It's actually either a string (if there is a related calendar) or undefined (if there is none).
During my testing this worked fine. Do you think I should explicitly check for undefined so
`props.calendarName !== undefined`

As for the names `name` is in this case the name of the maintenance schedule while `calendarName` is the name of the maintenance calendar

ncounter (approved):

web/html/src/manager/maintenance/details/schedule-details.tsx
Etheryte: This check is fine, if it can be `undefined` then the prop definition should be either `calendarName?: string` or `calendarName: string | undefined`, depending on which case is correct.  

As for `name`, a comment on the prop declaration could help, because personally I got confused. :D

parlt91: Ok makes sense. Updated the PR accordingly and added a comment for the name prop

jordimassaguerpla:
Congratulations! Acceptance tests passed :woman_dancing: :man_dancing: 

* core tests :heavy_check_mark: 
* visualization functional scope :heavy_check_mark: 

reference: ci#283

PR #3811, opened 2021-05-26
Fix generation of unique repository name

Bischoff:
## What does this PR change?

This PR fixes the function that generates a unique repository name:

It transforms *all* `/` to `_`, not only the *first one* (`gsub!` instead of `sub!`).

## Changelogs

- [x] No changelog needed


calancha (approved):

srbarrios (commented):
Cool, make sense in some case indeed.

PR #3816, opened 2021-05-27
Testsuite: More robust client IP extraction

calancha:
## What does this PR change?
We parse the output of `ip address show dev eth0`
assuming that the IP appears in the second line.

The following piece of output is from the server at the head branch in the CI:
```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether aa:b2:93:01:00:b2 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    altname ens3
    inet 10.162.210.178/24 brd 10.162.210.255 scope global eth0
```

The above lines with `altname` break our assumption: we'd extract `enp0s3` as the wrong IP.

retail-wrong-public-ip-at-formula-Screenshot from 2021-05-19 10-44-59

- testsuite/features/step_definitions/common_steps.rb
- testsuite/features/step_definitions/salt_steps.rb
- testsuite/features/support/lavanda.rb
- testsuite/features/support/twopence_init.rb:
Grep for 'inet ' to select the right line.
Store the public IP as a new attribute of the clients.

## Links
This comes from https://github.com/SUSE/spacewalk/issues/5838

### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/15011
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/15012

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -696,9 +696,7 @@ end
 Then(/^I add proxy record into hosts file on "([^"]*)" if avahi is used$/) do |host|
   node = get_target(host)
   if node.full_hostname.include? 'tf.local'
-    output, _code = $proxy.run("ip address show dev eth0")
-    ip = output.split("\n")[2].split[1].split('/')[0]
-    node.run("echo '#{ip} #{$proxy.full_hostname} #{$proxy.hostname}' >> /etc/hosts")
+    node.run("echo '#{$proxy.public_ip} #{$proxy.full_hostname} #{$proxy.hostname}' >> /etc/hosts")
   else
     puts 'Record not added - avahi domain is not detected'
   end
--- a/testsuite/features/support/twopence_init.rb
+++ b/testsuite/features/support/twopence_init.rb
@@ -134,13 +134,6 @@ $nodes.each do |node|
   puts "Host '#{$named_nodes[node.hash]}' is alive with determined hostname #{hostname.strip} and FQDN #{fqdn.strip}" unless $build_validation
 end
 
-# Initialize IP address or domain name
-$nodes.each do |node|
-  next if node.nil?
-
-  node.init_ip(node.full_hostname)
-end
-
 # This function is used to get one of the nodes based on its type
 def get_target(host)
   node = $node_by_host[host]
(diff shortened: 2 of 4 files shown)


Bischoff (changes requested):
Yes, the wrong assumption was mine, sorry. It's much better now.

I would just move the new function to e.g. `support/commonlib.rb`.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -696,8 +696,7 @@
 Then(/^I add proxy record into hosts file on "([^"]*)" if avahi is used$/) do |host|
   node = get_target(host)
   if node.full_hostname.include? 'tf.local'
-    output, _code = $proxy.run("ip address show dev eth0")
-    ip = output.split("\n")[2].split[1].split('/')[0]
+    ip = client_public_ip 'proxy'
```
srbarrios, line 699: I personally prefer to call the method as you are doing here, it's more Ruby style.
But in the majority of our code we use parenthesis to call our methods, so, maybe better to use parenthesis?
Really just a matter of taste, but to try to keep some code style.

testsuite/features/support/commonlib.rb
```diff
@@ -219,6 +219,21 @@ def get_client_type(name)
   end
 end
 
+def client_public_ip(host)
+  node = get_target(host)
+  interfaz = 'eth0'
```
srbarrios, line 224: I personally don't like this variable name :P 
What about `inet`

srbarrios, line 224: Instead of having this initial value, I would have an `else` the `case` statement.

srbarrios, line 224: Instead of having this initial value, I would have an `else` the `case` statement.

testsuite/features/support/commonlib.rb
```diff
@@ -219,6 +219,21 @@ def get_client_type(name)
...
+  case host
+  when /^sle/, /^ceos/
+    interfaz = 'eth0'
+  when /^ubuntu/
+    interfaz = 'ens3'
+  end
```
srbarrios, line 230:
```suggestion
inet = case host
  when /^ubuntu/
    'ens3'
  else
    'eth0'
  end
```

testsuite/features/support/commonlib.rb
```diff
@@ -219,6 +219,21 @@ def get_client_type(name)
...
+  output, code = node.run("ip address show dev #{interfaz} | grep 'inet '")
```
srbarrios, line 231: `hostname -i` can be good enough? Or it might be the case that the first IP is not the IP we want? :thinking:

Bischoff (changes requested):
some last changes

testsuite/features/step_definitions/common_steps.rb
calancha: I like this idea.

Bischoff: maybe it even already exists.... who knows

testsuite/features/support/twopence_init.rb
```diff
@@ -139,6 +139,7 @@ def twopence_init(target)
   next if node.nil?
 
   node.init_ip(node.full_hostname)
+  node.init_public_ip(client_public_ip(node))
```
Bischoff, line 142: I thought Lavanda would provide things like hostname and IP but it does not. It only provides `run()`

Bischoff: It looks like `init_ip` is named badly, as it registers the hostname ???

Anyway....

srbarrios (approved):
I love it!

testsuite/features/support/twopence_init.rb
```diff
@@ -288,3 +281,51 @@ def file_inject(node, local_file, remote_file)
...
+  next if node.is_a?(String) && node.empty?
+
+  node.init_ip(node.full_hostname)
+
+  host = $host_by_node[node]
+  skipped_hosts = 'localhost'
+  next if skipped_hosts.include? host
+
+  ip = client_public_ip host
+  node.init_public_ip ip
```
srbarrios, line 330: nitpick, I would always call the methods in the same way for code styling, but ok, fine `node.init_public_ip(ip)`

testsuite/features/support/twopence_init.rb
```diff
@@ -288,3 +281,51 @@ def file_inject(node, local_file, remote_file)
...
+
+  output.split[1].split('/')[0]
+end
+
+# Initialize IP address or domain name
+$nodes.each do |node|
+  next if node.nil?
+  next if node.is_a?(String) && node.empty?
+
+  node.init_ip(node.full_hostname)
```
srbarrios, line 323: Not for this PR, but I think we need a card to research and address if this instantiation is ok, where we use .ip attribute, etc.

testsuite/features/support/twopence_init.rb
```diff
@@ -288,3 +281,51 @@ def file_inject(node, local_file, remote_file)
...
+  next if ignored_hosts.include? host
+
+  $host_by_node[node] = host
+  puts "Added #{host} as key for node: '#{node}'"
+end
+
+def client_public_ip(host)
+  node = $node_by_host[host]
+  interface = case host
+              when /^sle/, /^ceos/, /^debian/, 'server', 'proxy', 'build_host'
```
srbarrios, line 305: Be caution here, I would say it covers all hosts in BV but just be sure about it :)

testsuite/features/support/twopence_init.rb
calancha: I was forced to add this guard after realizing that some nodes were passing with value the empty string.
The code creating those nodes should be responsible to change those empty strings to `nil`.

srbarrios: OK, but I'm not talking about guard :)
I'm talking about setting inside the .ip attribute the string .full_hostname

PR #3818, opened 2021-05-27
Fail early - ensure minimal performance

Bischoff:
## What does this PR change?

This PR changes the default timeout for build validations from half an hour (!) to 8 minutes.

Rationale:
 * fail early
 * ensure minimal performance
 
**Fail early**: the less we wait for failures, the less we lose our time. Today I lost one hour and a half for 3 failed bootstraps...

**Ensure minimal performance**: with a default timeout of half an hour, any action that takes, say, 25 minutes, will be considered as acceptable. Our customers will certainly not be of that opinion.

If SUSE Manager can't cope with parallel bootstraps in a reasonable delay, then it is a product bug, that should be addressed as such. We should not hide that problem in the test suite.

The comment says:
```
  # Build Validations will require longer timeouts due to the low performance of our VMs
  # if we ever improve this fact, we can reduce these timeouts.
```
If that is really the cause for longer timeouts, then accepting VMs that are twice as slow is generous enough, I think.

## Links

Ports:
* 4.0:
* 4.1:

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -37,8 +37,8 @@ if File.exist?(custom_repos_path)
   # HACK
   # Build Validations will require longer timeouts due to the low performance of our VMs
   # if we ever improve this fact, we can reduce these timeouts.
-  Capybara.default_max_wait_time = 30
-  DEFAULT_TIMEOUT = 1800
+  Capybara.default_max_wait_time = 20
+  DEFAULT_TIMEOUT = 500
 end
 
 def enable_assertions


calancha (approved):

srbarrios (commented):
I'm fine to merge it ONLY if we first do a full pass of the BV test suite in parallel and this change doesn't impact.
If it does, failing applying states on bootstrap in any client, I would not agree on merging it until we don't address the product issue causing this. 

The issue we tried to fix with a bigger timeout was reported [here]( https://github.com/SUSE/spacewalk/issues/12775#issuecomment-737962684) and more than one developer debug it, but we could not found the root cause. So we took that decision to unblock the test of testsuite.

If you really want to merge this timeout in BV without further considerations, then make this value parametrizable and set up the value from the Jenkins Pipeline, depending on the mode of running (sequential/parallel) and being able to setup it when we run a feature manually with cucumber comand.

testsuite/features/support/env.rb
```diff
@@ -37,8 +37,8 @@
   # HACK
   # Build Validations will require longer timeouts due to the low performance of our VMs
   # if we ever improve this fact, we can reduce these timeouts.
-  Capybara.default_max_wait_time = 30
-  DEFAULT_TIMEOUT = 1800
+  Capybara.default_max_wait_time = 20
+  DEFAULT_TIMEOUT = 500
```
srbarrios, line 41:
```suggestion
  DEFAULT_TIMEOUT = ENV['CUSTOM_TIMEOUT'].to_i if ENV['CUSTOM_TIMEOUT']
```

lkotek (approved):
Very reasonable :-)

srbarrios (changes requested):
Please don't merge this without my suggestion.
Once done, I can do another PR to take care of Pipeline part.

PR #3820, opened 2021-05-28
Testsuite: Speed up a slow step

calancha:
## What does this PR change?
Speed up a slow step.

We often the step `I check for failed events on history event page`.
This step should be quite fast (< 30 s); instead, it's taking
about 2-3 minutes on average.

This is because the `has_css?` inside the loop uses the default timeout,
i.e. `Capybara.default_max_wait_time` (10 s).

Check if the elemnent is in the row with no waiting time.

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15019
### Ports
- Manager-4.1: https://github.com/SUSE/spacewalk/pull/15020
- Manager-4.0: https://github.com/SUSE/spacewalk/pull/15021

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -1150,7 +1150,7 @@ When(/^I check for failed events on history event page$/) do
   event_table_xpath = "//div[@class='table-responsive']/table/tbody"
   rows = find(:xpath, event_table_xpath)
   rows.all('tr').each do |tr|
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

srbarrios:
I'd merge this, better don't let it here long time, if it worked fine in 4.0 and 4.1

Bischoff (approved):
I don't fully understand the fix, but approving

Bischoff (approved):
ah now i understand the change

PR #3821, opened 2021-05-28
Fix salt proxy cleanup

Bischoff:
## What does this PR change?

This PR:
* cleans up the package `spacewalk-proxy-salt` holding the salt broker when we do a full salt cleanup
* handles the debian case
* tests that the `salt-broker` service is properly started

## Links

Ports:
* 4.0: SUSE/spacewalk#15027
* 4.1: SUSE/spacewalk#15026

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/salt_steps.rb
+++ b/testsuite/features/step_definitions/salt_steps.rb
@@ -674,10 +674,10 @@ When(/^I perform a full salt minion cleanup on "([^"]*)"$/) do |host|
   node = get_target(host)
   if host.include? 'ceos'
     node.run('yum -y remove --setopt=clean_requirements_on_remove=1 salt salt-minion', false)
-  elsif host.include? 'ubuntu'
+  elsif (host.include? 'ubuntu') || (host.include? 'debian')
     node.run('apt-get --assume-yes remove salt-common salt-minion && apt-get purge salt-common salt-minion && apt-get autoremove', false)
   else
-    node.run('zypper --non-interactive remove --clean-deps -y salt salt-minion', false)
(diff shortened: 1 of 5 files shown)


testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -674,10 +674,10 @@ def pillar_get(key, minion)
   node = get_target(host)
   if host.include? 'ceos'
     node.run('yum -y remove --setopt=clean_requirements_on_remove=1 salt salt-minion', false)
-  elsif host.include? 'ubuntu'
+  elsif host.include? 'ubuntu' or host.include? 'debian'
     node.run('apt-get --assume-yes remove salt-common salt-minion && apt-get purge salt-common salt-minion && apt-get autoremove', false)
   else
-    node.run('zypper --non-interactive remove --clean-deps -y salt salt-minion', false)
+    node.run('zypper --non-interactive remove --clean-deps -y salt salt-minion spacewalk-proxy-salt', false)
```
ktsamis, line 680: This should only run in proxy right? Otherwise does it throw an error or a different exit code than expected? Since we run this step to all minions

ktsamis (changes requested):

testsuite/features/step_definitions/salt_steps.rb
Bischoff: It indeed runs everywhere, but it does not throw an error if one of the packages is missing - the `false` argument makes it ignore errors.

In practice, error 140 appears when some packages are removed and others not, and it is then ignored.

ktsamis: all good then

PR #3822, opened 2021-05-28
Missing Debian test

Bischoff:
## What does this PR change?

This PR adds a test for debian instead of only ubuntu.

It is likely we have other such problems.

## Links

Ports:
* 4.0: SUSE/spacewalk#15031
* 4.1: SUSE/spacewalk#15030

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -1172,12 +1172,13 @@ When(/^I create the MU repositories for "([^"]*)"$/) do |client|
     if repository_exist? unique_repo_name
       puts "The MU repository #{unique_repo_name} was already created, we will reuse it."
     else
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1172,12 +1172,13 @@ def server_secret
     if repository_exist? unique_repo_name
       puts "The MU repository #{unique_repo_name} was already created, we will reuse it."
     else
+      content_type = (client.include? 'ubuntu') || (client.include? 'debian') ? 'deb' : 'yum'
```
srbarrios, line 1175: Ah, so we can use both operators! `or` and `||` :)

srbarrios (approved):

PR #3824, opened 2021-05-29
cleanup code from legacy things

mcalmer:
## What does this PR change?

- remove mirror credential settings from rhn.conf - moved to DB since a long time
- remove `satellite.rhn.redhat.com` as default for rhn_parent. We do not set a default anymore
- remove iss_ca_chain. We use now just ca_cert as fallback
- set ca_chain to empty value or to /usr/share/rhn/RHN-ORG-TRUSTED-SSL-CERT on managed clients
- remove expired RHNS-CA-CERT certificate
- do not install RedHat's gpg key

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/backend/rhn-conf/rhn_server_satellite.conf
+++ b/backend/rhn-conf/rhn_server_satellite.conf
@@ -13,8 +13,7 @@ mount_point = /var/spacewalk
 rhn_metadata_handler = /SAT-DUMP
 rhn_iss_metadata_handler = /SAT-DUMP-INTERNAL
 rhn_xmlrpc_handler = /SAT
-ca_chain = /usr/share/rhn/RHNS-CA-CERT
-rhn_parent = satellite.rhn.redhat.com
+server.satellite.rhn_parent =
 
 sync_cache_dir = /var/cache/rhn/
 http_proxy =
--- a/backend/rhn-conf/rhn_server_satellite.conf
+++ b/backend/rhn-conf/rhn_server_satellite.conf
@@ -13,7 +13,7 @@ mount_point = /var/spacewalk
 rhn_metadata_handler = /SAT-DUMP
 rhn_iss_metadata_handler = /SAT-DUMP-INTERNAL
 rhn_xmlrpc_handler = /SAT
-server.satellite.rhn_parent =
+rhn_parent =
 
 sync_cache_dir = /var/cache/rhn/
 http_proxy =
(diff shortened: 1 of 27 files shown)


spacewalk/config/var/lib/rhn/rhn-satellite-prep/etc/rhn/rhn.conf
```diff
@@ -2,7 +2,6 @@ traceback_mail = @@traceback_mail@@
 mount_point = @@mount_point@@
 kickstart_mount_point = @@kickstart_mount_point@@
 repomd_cache_mount_point = /var/cache 
-server.satellite.rhn_parent = satellite.rhn.redhat.com
```
meaksh, line 5: Are we dropping `server.satellite.rhn_parent` from `rhn.conf` but not from other conf files like proxy?

meaksh (commented):
Thanks - Just two questions to clarify

backend/rhn-conf/rhn_server_satellite.conf
```diff
@@ -13,8 +13,7 @@ mount_point = /var/spacewalk
 rhn_metadata_handler = /SAT-DUMP
 rhn_iss_metadata_handler = /SAT-DUMP-INTERNAL
 rhn_xmlrpc_handler = /SAT
-ca_chain = /usr/share/rhn/RHNS-CA-CERT
-rhn_parent = satellite.rhn.redhat.com
+server.satellite.rhn_parent =
```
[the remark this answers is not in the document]
mcalmer: correct. I wonder how this sneaked in

spacewalk/config/var/lib/rhn/rhn-satellite-prep/etc/rhn/rhn.conf
mcalmer: This is the setup program for the server. So we drop this from the server as the server to server connection is the old redhat cdn connection which we never used.
On a proxy the rhn_parent is used in case of ISS and refer to the Uyuni/SUSE Manager Server.
So we need to keep it.

meaksh (approved):
LGTM. Thanks for doing this cleanup!

PR #3815, opened 2021-05-27
Make spec license canonical

Etheryte:
## What does this PR change?

The frontend build step currently checks licenses and joins them using `and`, whereas OBS tools use `AND`. This means that pulling master will result in a broken frontend project build. This PR fixes the issue.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes

- [x] **DONE**

## Test coverage
- No tests: only internal and user invisible changes

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15002

- [x] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #3829, opened 2021-05-31
Deprecate `objectDefaultValueHandler`

Etheryte:
## What does this PR change?

[`objectDefaultValueHandler`](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/core/utils/objects.ts) is currently imported in only two views and is practically not used in either of them. This is evidenced by the fact that the util is actually bugged but the branch is never hit because all the values are defined in all the inputs. This PR removes the bugged utilty, removing unnecessary complexity.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: only internal and user invisible changes

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13648

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/spacewalk-web.spec
+++ b/web/spacewalk-web.spec
@@ -63,7 +63,7 @@ but it does generate a number of sub-packages.
 
 %package -n susemanager-web-libs
 Summary:        Vendor bundles for spacewalk-web
-License:        0BSD AND BSD-3-Clause AND LGPL-3.0-or-later AND MIT and MPL-2.0
+License:        0BSD and BSD-3-Clause and LGPL-3.0-or-later and MIT and MPL-2.0
 Group:          Applications/Internet
 
 BuildArch:      noarch
--- a/web/spacewalk-web.spec
+++ b/web/spacewalk-web.spec
@@ -63,7 +63,7 @@ but it does generate a number of sub-packages.
 
 %package -n susemanager-web-libs
 Summary:        Vendor bundles for spacewalk-web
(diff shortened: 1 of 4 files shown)


web/spacewalk-web.spec
```diff
@@ -63,7 +63,7 @@ but it does generate a number of sub-packages.
 
 %package -n susemanager-web-libs
 Summary:        Vendor bundles for spacewalk-web
-License:        0BSD AND BSD-3-Clause AND LGPL-3.0-or-later AND MIT AND MPL-2.0
+License:        0BSD and BSD-3-Clause and LGPL-3.0-or-later and MIT and MPL-2.0
```
parlt91, line 66: At first I thought this is weird but then I realized this is due to my recent change to the script that fills the spec file. In the end the case doesn't matter here, so I think it's okay

parlt91 (approved):
It looks good to me. I don't know the original reasoning behind using a Proxy here, but as far as I can tell it's not needed here, but better to wait for @ncounter to approve as well.

web/spacewalk-web.spec
Etheryte: I have a PR that addresses this, for now I needed to rebuild it so the Github Action validation passes. https://github.com/uyuni-project/uyuni/pull/3815

ncounter (approved):
I believe the purpose of it was exactly the replacement of the `Partial<T>` of Typescript. Now we have it so we don't really need the utility anymore. :+1: Thank you

PR #3844, opened 2021-06-02
Detect free products without ping

mcalmer:
## What does this PR change?

In the Product Wizard we only show products which are really available and can be synced.
To very this we tested with a HEAD request if the mandatory product repositories are available.
A lot of people configure their firewall to drop requests also from the internal network.
This result in the problem, that these tests need to run into the timeout before they are marked as "not available".
As we add more and more of these free products which need this test, the repository check takes longer and longer and fail more often because of the XMLRPC timeout.

It was decided to identify "free products" and do not perform the HEAD requests there. We always assume that they have to be available.

In case they are not and a customer add the channels for syncing, it will fail "later".

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- https://github.com/uyuni-project/uyuni-docs/pull/989

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14395
Tracks https://github.com/SUSE/spacewalk/pull/15191

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


paususe (approved):
Thank you!

backend/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -630,15 +631,21 @@ def setup_repo(self, repo, uln_repo=False):
...
+        zypper_cmd = "{} --root {} --reposd-dir {} --cache-dir {} --raw-cache-dir {} --solv-cache-dir {} ref".format(
             zypper_cmd,
             REPOSYNC_ZYPPER_ROOT,
             os.path.join(repo.root, "etc/zypp/repos.d/"),
             REPOSYNC_ZYPPER_RPMDB_PATH,
             os.path.join(repo.root, "var/cache/zypp/raw/"),
             os.path.join(repo.root, "var/cache/zypp/solv/")
-        ))
+        )
+        process = subprocess.Popen(zypper_cmd.split(' '), stdout=subprocess.DEVNULL, stderr=subprocess.PIPE)
```
meaksh, line 642: Hmmm, I think this is going to cause some problems.

IIUC this will hide the stdout from "zypper ref" execution. Currently we see Zypper output refreshing the repo metadata when running `spacewalk-repo-sync`.

And even more problematic is that with this way I think we won't be able to accept the GPG keys in case Zypper is asking for it.

Maybe what you want is something like:

```python
   process = subprocess.run(zypper_cmd.split(' '), stderr=subprocess.PIPE)
```

This way, the output from Zypper will be shown and you can also capture retcode and stderr at `process.returncode` and `process.stderr`.

Does it make more sense?

mcalmer: changed, please check

java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
```diff
@@ -754,6 +759,18 @@ public void refreshRepositoriesAuthentication(
             if (tokenOpt.isPresent()) {
                 newAuth = new SCCRepositoryTokenAuth(tokenOpt.get());
             }
+            else if (repo.getProducts().isEmpty()) {
+                log.debug("Repo '" + repo.getUrl() + "' not in the product tree. Skipping");
+                continue;
+            }
```
ncounter, line 765: I'd personally prefer better not to have the `continue;` jump, but wrapping instead with the condition criteria, something like

```java
if (repo.getProducts().isEmpty(){
  log.debug("Repo '" + repo.getUrl() + "' not in the product tree. Skipping");
}
else {
  if (tokenOpt.isPresent()) {
    newAuth = new SCCRepositoryTokenAuth(tokenOpt.get());
  }
  else if {
    ....
  }
}
```

just from the logic perspective and the code readability, so I can understand the flow and the reasoning (if I understood the logic correctly).

ncounter (approved):
I just left a small comment, otherwise from the pure code shape perspective of Java and the CSS part it LGTM :+1: 

I cannot comment about the overall logic implementation because I don't know how it works.

backend/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -630,15 +631,20 @@ def setup_repo(self, repo, uln_repo=False):
...
             REPOSYNC_ZYPPER_RPMDB_PATH,
             os.path.join(repo.root, "var/cache/zypp/raw/"),
             os.path.join(repo.root, "var/cache/zypp/solv/")
-        ))
+        )
+        process = subprocess.run(zypper_cmd.split(' '), stderr=subprocess.PIPE)
+        ret_error = process.returncode
+
         if ret_error:
+            if errs:
```
meaksh, line 646: AFAICS this `errs` variable here is not defined at this point. Should this be `process.stderr` ?

Also, maybe could we do something like:

```python
    if process.returncode:
        if process.stderr:
            raise RepoMDError("Cannot access repository.\n{}".format(sstr(process.stderr)))
        raise RepoMDError("Cannot access repository. Maybe repository GPG keys are not imported")
```

to avoid declaring `ret_error = process.returncode` which seems no to be really necessary.

meaksh (changes requested):

java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
mcalmer: Well, I am a fan of not indent things too much.
In this case there is also another reason why we getter not call repo.getProducts() on all repos. This cause a single SQL query which result in performance issues when we do it on all of our repos. When we get repos from SCC with token auth I expect that we have them in the tree. Even if not, we do not more operations on them.

For the other we might trigger a head request which slow down things and with this I try everything to minimize the cases where we have to do it. But at thi splace it does not hurt as in the next "if" we call the same to find out if we need to do more checks. The SQL query will be done anyway. So we do not loose anything.

I think I will keep it as it is now.

backend/satellite_tools/repo_plugins/yum_src.py
mcalmer: changed

meaksh (approved):

java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
ncounter: As long as there is a reason behind and it affects performances or whatever else, I'm just fine with it. Thanks for taking the time to explain it here. :+1:

susemanager-sync-data/additional_products.json
```diff
@@ -10,7 +10,7 @@
         "friendly_name" : "AlmaLinux 8 x86_64",
         "product_class" : "SLE-M-T",
         "cpe" : null,
-        "free" : false,
+        "free" : true,
```
lucidd, line 13: since this file is generated we need to also adjust the data in the fake products in suma toolbox.

mcalmer: I think I did it (MR) but I need to double check.

mcalmer: Seems this is missing. I will do a MR

mcalmer: I had the changes already on my disk. Now is the merge request created `merge_requests/281`

lucidd (approved):

PR #3854, opened 2021-06-07
Fix alma amazon product matching

mcalmer:
## What does this PR change?

When parsing release files of Amazon and Alma linux the resulting string does not match the SCC/sumatoolbox product setup.
This change adapt the parser to match the expected strings.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were adapted

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15089

SUMA 4.2 only. No need to backport to 4.1

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


renner (approved):
This looks good to me! :+1:

java/code/src/com/suse/manager/reactor/utils/RhelUtils.java
```diff
@@ -156,11 +157,18 @@ public String getRelease() {
...
+        // We match here data from the system and try to find the product
+        // how it is named in SCC or sumatoolbox. This requires sometimes
+        // some changes on the string we parse.
+        //
         // AlmaLinux and AmazonLinux are also matched by the RHEL matcher
         Matcher matcher = RHEL_RELEASE_MATCHER.matcher(releaseFile);
         if (matcher.matches()) {
             String name =
                     matcher.group(1).replaceAll("(?i)linux", "").replaceAll(" ", "");
+            if (name.startsWith("Alma") || name.startsWith("Amazon")) {
```
srbarrios, line 169: I wonder if we considered AlibabaLinux or we should let it for later.

mcalmer: AFAIK Alibaba has a different matcher as the string is not comparable. So we should be fine for Alibaba in case we support it. Currently it is not in the toolbox/SCC.

PR #3861, opened 2021-06-08
Add reverse resolution sanity check

Bischoff:
## What does this PR change?

This PR adds a check for reverse resolution of names.

Technical details:
* I used `getent` because it had to work with both DNS and avahi

## Links

Fixes SUSE/spacewalk#15037

Ports:
* 4.0: SUSE/spacewalk#15134
* 4.1: SUSE/spacewalk#15133, SUSE/spacewalk#15167

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -9,10 +9,18 @@ require 'nokogiri'
 
 Then(/^"([^"]*)" should have a FQDN$/) do |host|
   node = get_target(host)
-  result, return_code = node.run('hostname -f')
+  result, return_code = node.run('hostname -f', false)
   result.delete!("\n")
   raise 'cannot determine hostname' unless return_code.zero?
   raise 'hostname is not fully qualified' unless result == node.full_hostname
+end
+
+Then(/^reverse resolution should work for "([^"]*)"$/) do |host|
+  node = get_target(host)
+  result, return_code = node.run("getent hosts #{node.ip}", false)
+  result.delete!("\n")
+  raise 'cannot do reverse resolution' unless return_code.zero?
+  raise "reverse resolution returned #{result}, expected to see #{node.full_hostname}" unless result.include? node.full_hostname
 end
 
 Then(/^"([^"]*)" should communicate with the server$/) do |host|
--- a/testsuite/features/core/allcli_sanity.feature
+++ b/testsuite/features/core/allcli_sanity.feature
@@ -7,6 +7,7 @@ Feature: Sanity checks
 
   Scenario: The server is healthy
     Then "server" should have a FQDN
+    And reverse resolution should work for "server"
     And the clock from "server" should be exact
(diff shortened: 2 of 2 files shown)


ktsamis (approved):

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -9,12 +9,23 @@
...
 end
 
+Then(/^reverse resolution should work for "([^"]*)"$/) do |host|
+  node = get_target(host)
+  result, _rc = node.run('hostname --version', false)
+  if result =~ /^hostname 3\./
+    result, return_code = node.run('hostname -A | tr " " "\n" | sort -u', false)
+    result.delete!("\n")
+    raise 'cannot do reverse resolution' unless return_code.zero?
+    raise 'reverse resolution does not give expected results' unless result == node.full_hostname
```
srbarrios, line 25:
```suggestion
    raise "reverse resolution #{result} does not give expected results #{node.full_hostname}" unless result == node.full_hostname
```

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -9,12 +9,23 @@
...
+  result, return_code = node.run('hostname -f', false)
   result.delete!("\n")
   raise 'cannot determine hostname' unless return_code.zero?
   raise 'hostname is not fully qualified' unless result == node.full_hostname
 end
 
+Then(/^reverse resolution should work for "([^"]*)"$/) do |host|
+  node = get_target(host)
+  result, _rc = node.run('hostname --version', false)
+  if result =~ /^hostname 3\./
```
srbarrios, line 21: I recommend an `else` on this statement, so we report when this step is actually not checking anything.
That will be the case for clients like sles11sp4 where the hostname version is 1.100

testsuite/features/step_definitions/command_steps.rb
Bischoff: ok

testsuite/features/step_definitions/command_steps.rb
Bischoff: yes, the problem is that `hostname` version 1.100 does not provide `-A` option.

will do.

testsuite/features/step_definitions/command_steps.rb

testsuite/features/step_definitions/command_steps.rb
Bischoff: done

one of the problems of this step is that it has to work for avahi too, so we cannot use `host`, `nslookup`, or `dig`

Bischoff: wait, getent could do the trick

Bischoff: changed to use `getent`

Bischoff: new version tested locally to work.

srbarrios (approved):

PR #3864, opened 2021-06-09
Retracted patches spacecmd

hustodemon:
## What does this PR change?

Make spacecmd aware of the retracted attribute exposed for packages/patches via the XMLRPC API and display this information for the user, if it's available.

This PR doesn't break spacecmd when it consumes older API which does not expose the retracted attributes. Smoke testing was done with SUSE Manager 4.0.

Review commit-by-commit.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: existing spacecmd documentation does not contain detailed information about the commands output

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links
https://github.com/SUSE/spacewalk/issues/14135

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mcalmer (approved):
Looks good so far. I am not sure if all cases are ready to work with a SUMA API version which does not have retracted patches implemented. But if this goes out after SUMA 4.0 is EOL we do not have a problem.

hustodemon:
thanks for the review!

> Looks good so far. I am not sure if all cases are ready to work with a SUMA API version which does not have retracted patches implemented.

@mcalmer  I tried to run all `spacecmd` subcommands that contain `package` or `errata` word against `4.0` as a smoke test and it worked, so we should be safe for 4.0 too...

meaksh (approved):
LGTM :+1:

meaksh:
@hustodemon I just noticed the `java_pgsql_tests` job is failing in this PR but seems it was due an error when `obs-to-maven` was running.

I'm retriggering those so hopefully they get green this time! :)

hustodemon:
Hmm, still red (but the error is still unrelated). I re-run again before investigating deeper.

spacecmd/src/spacecmd/utils.py
```diff
@@ -434,9 +434,10 @@ def print_errata_summary(erratum):
     if len(date_parts) > 1:
         erratum['date'] = date_parts[0]
 
-    print('%s  %s  %s' % (
+    print('%s  %s  %s  %s' % (
         erratum.get('advisory_name').ljust(14),
-        wrap(erratum.get('advisory_synopsis'), 50)[0].ljust(50),
+        wrap(erratum.get('advisory_synopsis'), 49)[0].ljust(49),
+        (erratum.get('advisory_status') or '').ljust(9),
```
agraul, line 440: Instead of checking if `erratum.get` returns `None` and then setting the default value, it can be passed to `dict.get()` as the second parameter like this
```suggestion
        erratum.get('advisory_status', '').ljust(9),
```

I see the same is done further up, is this something that wasn't in Python 2 or is there another reason we've been using `dict.get() or X` until now?

hustodemon: The reason is possible behavior in case the key exists in the dictionary, but the value is `null`: the `dict.get('key', '') would return `null` in such case.

agraul: Thank you for the explanation, I didn't think about that case. I was expecting that `erratum.get()` would return `None` because the key might not be in the `dict`, not because the value is `None`. When the key is always in the dict, I favor `dict[key]` (i.e. `erratum['advisory_status']`) which might explain why I didn't not understand this immediately.

PR #3867, opened 2021-06-09
Try to install spacecmd

Bischoff:
## What does this PR change?

`spacecmd` is an important utility that must be installable.

As `spacecmd` comes from the client tools, this PR also tests
the GPG key for the client tools at the same time.

## Links

Ports:
* 4.0:
* 4.1:

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/smoke_tests/smoke_tests.template
+++ b/testsuite/features/build_validation/smoke_tests/smoke_tests.template
@@ -14,6 +14,7 @@ Feature: Smoke tests for <client>
   - Schedule Software package refresh
   - Schedule Hardware refresh
   - Reboot the client via Web UI
+  - Install spacecmd from client tools
 
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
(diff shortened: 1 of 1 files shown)


Bischoff:
tested in 4.2 BV existing testbed

testsuite/features/build_validation/smoke_tests/smoke_tests.template
```diff
@@ -137,3 +138,16 @@ Feature: Smoke tests for <client>
     When I force picking pending events on "<client>" if necessary
     And I wait at most 600 seconds until event "System reboot scheduled by admin" is completed
     Then I should see a "This action's status is: Completed" text
+
+  Scenario: Install spacecmd from the client tools on the <client>
+    Given I am on the Systems overview page of this "<client>"
```
srbarrios, line 143: In theory we don't need this first line.
If `Then I should see a "This action's status is: Completed" text` ends in a state where "Software" tab is visible.

testsuite/features/build_validation/smoke_tests/smoke_tests.template
```diff
@@ -137,3 +138,16 @@ Feature: Smoke tests for <client>
...
+    When I follow "Software" in the content area
+    And I follow "Install"
+    And I enter "spacecmd" as the filtered package name
```
srbarrios, line 146: I wonder if we can replace our package install scenario, by this one :thinking: 
This is installing something from client tools, so better keep both?

srbarrios (approved):

srbarrios (approved):

testsuite/features/build_validation/smoke_tests/smoke_tests.template
Bischoff: ok, removing it

i needed it for testing, so i was unsure (i was testing the scenario in isolation)

testsuite/features/build_validation/smoke_tests/smoke_tests.template
Bischoff: no we can't i already had same idea and discussed it with julio

yes because of the client tools thingie, and also because it tests installation still works after reboot

juliogonzalez:
Removing me as reviewer. Sorry, can't handle that many PR reviews, so I will handle only those that are strictly needed :(

lkotek (approved):
LGTM :+1:

PR #3869, opened 2021-06-10
Fix product migration when scheduled from the event page (bsc#1187066)

cbbayburt:
Problems fixed:
 - Pass 'allowVendorChange' correctly when the migration is scheduled from the completed event page of a dry-run
 - ~~Revert client's assigned channels when migration is failed for some reason~~ (needs more consideration in the future)

## Test coverage
- Cucumber tests needed

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15131

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/utils/SaltUtils.java
+++ b/java/code/src/com/suse/manager/utils/SaltUtils.java
@@ -652,8 +652,7 @@ public class SaltUtils {
         else if (action.getActionType().equals(ActionFactory.TYPE_DIST_UPGRADE)) {
             DistUpgradeAction dupAction = (DistUpgradeAction) action;
             DistUpgradeActionDetails actionDetails = dupAction.getDetails();
-            if (serverAction.getStatus().equals(ActionFactory.STATUS_FAILED) || actionDetails.isDryRun()) {
-                // Collect changed channel list to revert the channel change
+            if (actionDetails.isDryRun()) {
                 Map<Boolean, List<Channel>> collect = actionDetails.getChannelTasks()
                         .stream().collect(Collectors.partitioningBy(
                                 ct -> ct.getTask() == DistUpgradeChannelTask.SUBSCRIBE,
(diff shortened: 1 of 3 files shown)


java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -652,7 +652,8 @@ else if (action.getActionType().equals(ActionFactory.TYPE_HARDWARE_REFRESH_LIST)
         else if (action.getActionType().equals(ActionFactory.TYPE_DIST_UPGRADE)) {
             DistUpgradeAction dupAction = (DistUpgradeAction) action;
             DistUpgradeActionDetails actionDetails = dupAction.getDetails();
-            if (actionDetails.isDryRun()) {
+            if (serverAction.getStatus().equals(ActionFactory.STATUS_FAILED) || actionDetails.isDryRun()) {
```
mcalmer, line 655: Uhhh I think this is wrong.
There is a reason why we do not revert the channels after a failed DUP.

Reason: we do not know "when" it failed. In case the migration started already and you abort becuase of "disk full" or similar error, you have no chance to fix it and continue the migration.

mcalmer (changes requested):

java/code/src/com/suse/manager/utils/SaltUtils.java
cbbayburt: > Reason: we do not know "when" it failed. In case the migration started already and you abort becuase of "disk full" or similar error, you have no chance to fix it and continue the migration.

In that case, does it even matter if the channels are reverted or not?

cbbayburt: Ok, then let me revert this one anyway.
Maybe we can further inspect the result and revert only in case of some exceptional cases. But that's a bigger thing for later.

mcalmer (approved):

hustodemon (approved):
LGTM

PR #3870, opened 2021-06-10
Use a different repository for Salt for Uyuni

Bischoff:
## What does this PR change?

On Uyuni, Salt was downgraded after SP migration. This was because the repositories where latest Salt is are not the same for Uyuni.

## Links

Fixes  SUSE/spacewalk#15117

Ports:
* 4.0:
* 4.1:

Note: it could be made as uyuni-branch only, I'm porting only to avoid branches diverge.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/salt_steps.rb
+++ b/testsuite/features/step_definitions/salt_steps.rb
@@ -650,11 +650,13 @@ When(/^I install Salt packages from "(.*?)"$/) do |host|
 end
 
 When(/^I enable repositories before installing Salt on this "([^"]*)"$/) do |host|
-  step %(I enable repository "tools_additional_repo" on this "#{host}" without error control)
+  repo = $product == 'Uyuni' ? 'salt3002_repo_repo' : 'tools_additional_repo'
+  step %(I enable repository "#{repo}" on this "#{host}" without error control)
 end
 
 When(/^I disable repositories after installing Salt on this "([^"]*)"$/) do |host|
-  step %(I disable repository "tools_additional_repo" on this "#{host}" without error control)
+  repo = $product == 'Uyuni' ? 'salt3002_repo_repo' : 'tools_additional_repo'
+  step %(I disable repository "#{repo}" on this "#{host}" without error control)
 end
 
 # minion bootstrap steps


testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -650,11 +650,13 @@ def pillar_get(key, minion)
 end
 
 When(/^I enable repositories before installing Salt on this "([^"]*)"$/) do |host|
-  step %(I enable repository "tools_additional_repo" on this "#{host}" without error control)
+  repo = $product == 'Uyuni' ? 'salt3002_repo_repo' : 'tools_additional_repo'
```
ktsamis, line 653: not sure that salt3002_repo_repo would be there forever or it's an additional_repo

srbarrios, line 653: Silly question, should we just fix sumaform to have as name `tools_additional_repo` instead of `salt3002_repo_repo` when we deploy for Uyuni?

We had the same s--- with Uyuni for Ubuntu, and I fixed it here https://github.com/uyuni-project/sumaform/pull/885

Bischoff: you are right, it won't be here forever... it means we'll need to adapt this after a while :-(

calancha (approved):
LGTM.  Thank you!

testsuite/features/step_definitions/salt_steps.rb
Bischoff: They probably have different names because `salt3002_repo_repo` comes from the main.tf and `tools_additional_repo` comes from sumaform.

I am not sure about the reason for that - maybe because `salt3002_repo_repo` won't be there forever as Kostas said?

Please notice that in the test suite code, there are several places where `tools_additional_repo` is skipped for the Uyuni branch.

I don't mind moving this PR to sumaform and having the same name... but we would have to fix all those occurrences, and I am not sure about the consequences.

I would need the help of @juliogonzalez to really understand the difference between `tools_additional_repo` and `salt3002_repo_repo`...

srbarrios: Aha, so `salt3002_repo_repo` comes from the main.tf, I though it was on sumaform, then nothing, then we are fine with this PR.

srbarrios (approved):

juliogonzalez (approved):
OK for the short term, but but please: prepare a PR for Uyuni's `main.tf` at  `susemanager-ci` so the repos point to to this PR. We'll need to rollback this PR when we remove those repos around June 22nd.

And remember: we should have a general solution for this.
That's the name we gave, but FMPOV we need a general solution, not something adhoc.

FMPOV we need to document at sumaform a way to specify repos at the `main.tf` with a certain prefix, that then are enable by the testsuite.

Do we already have a card?

Bischoff:
> FMPOV we need to document at sumaform a way to specify repos at the `main.tf` with a certain prefix, that then are enable by the testsuite.
> 
> Do we already have a card?

Feel free to create one. I like the idea. However it might not work when we install docker (that one comes from a "regular" repo).

juliogonzalez:
> > FMPOV we need to document at sumaform a way to specify repos at the `main.tf` with a certain prefix, that then are enable by the testsuite.
> > Do we already have a card?
> 
> Feel free to create one. I like the idea. However it might not work when we install docker (that one comes from a "regular" repo).

@ktsamis all yours. I seem to remember you created a card, or started a discussion, or something.

PR #3873, opened 2021-06-11
Check the reposync went smoothly

Bischoff:
## What does this PR change?

We are currently waiting that the reposyncs finish, but then we don't check they went well.

This PR checks for the absence of `ERROR:` in reposync logs.

For now, it's BV only. We don't use mirrors in NUE's CI, so it's very easy to get errors of package checksum that does not match the one in the metadata.

## Links

Ports:
* 4.0: SUSE/spacewalk#15161
* 4.1: SUSE/spacewalk#15160, SUSE/spacewalk#15167

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -328,6 +328,11 @@ When(/^I kill all running spacewalk\-repo\-sync, excepted the ones needed to boo
   end
 end
 
+Then(/^the reposync logs should not report errors$/) do
+  result, code = $server.run('grep "ERROR:" /var/log/rhn/reposync/*.log', false)
+  raise "Errors during reposync:\n#{result}" if code.zero?
+end
+
 Then(/^"([^"]*)" package should have been stored$/) do |pkg|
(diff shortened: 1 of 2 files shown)


Bischoff:
Tested to work fine on 4.2 BV.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -328,6 +328,11 @@
   end
 end
 
+Then(/^the reposync logs should not report errors$/) do
+  result, code = $server.run('grep "ERROR:" /var/log/rhn/reposync/*.log', false)
```
srbarrios, line 332:
```suggestion
  result, code = $server.run('grep -H "ERROR:" /var/log/rhn/reposync/*.log', false)
```

srbarrios (approved):

testsuite/features/step_definitions/command_steps.rb
Bischoff: according to `man grep` " This is the default when there is more than one file to search." so it won't change much

PR #3876, opened 2021-06-14
fix db migrations script from 12 to 13

jordimassaguerpla:
## What does this PR change?

This fixes the db migration script from 12 to 13. There was a hardcoded 10 in the path which should have been OLDVERSION variables instead.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed
- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager/bin/pg-migrate-12-to-13.sh
+++ b/susemanager/bin/pg-migrate-12-to-13.sh
@@ -92,8 +92,8 @@ else
     echo "`date +"%H:%M:%S"`   Initialization of new postgresql $NEW_VERSION database failed!"
     echo "`date +"%H:%M:%S"`   Trying to restore previous state..."
     mv /var/lib/pgsql/data /var/lib/pgsql/data-new-failed
-    mv /var/lib/pgsql/data-pg10 /var/lib/pgsql/data
-    /usr/sbin/update-alternatives --set postgresql /usr/lib/postgresql10
+    mv /var/lib/pgsql/data-pg${OLD_VERSION} /var/lib/pgsql/data
+    /usr/sbin/update-alternatives --set postgresql /usr/lib/postgresql${OLD_VERSION}
     exit 1
 fi
 
--- a/susemanager/susemanager.changes
+++ b/susemanager/susemanager.changes
@@ -1,3 +1,5 @@
+- Fix database database migration script from PostgreSQL 12 to 13
+
 -------------------------------------------------------------------
 Wed Jun 02 11:08:38 CEST 2021 - jgonzalez@suse.com
 


susemanager/bin/pg-migrate-12-to-13.sh
```diff
@@ -92,8 +92,8 @@ else
     echo "`date +"%H:%M:%S"`   Initialization of new postgresql $NEW_VERSION database failed!"
     echo "`date +"%H:%M:%S"`   Trying to restore previous state..."
     mv /var/lib/pgsql/data /var/lib/pgsql/data-new-failed
-    mv /var/lib/pgsql/data-pg10 /var/lib/pgsql/data
-    /usr/sbin/update-alternatives --set postgresql /usr/lib/postgresql10
+    mv /var/lib/pgsql/data-$OLD_VERSION /var/lib/pgsql/data
```
juliogonzalez, line 95:
```suggestion
    mv /var/lib/pgsql/data-${OLD_VERSION} /var/lib/pgsql/data
```

susemanager/bin/pg-migrate-12-to-13.sh
```diff
@@ -92,8 +92,8 @@ else
...
+    /usr/sbin/update-alternatives --set postgresql /usr/lib/postgresql$OLD_VERSION
```
juliogonzalez, line 96:
```suggestion
    /usr/sbin/update-alternatives --set postgresql /usr/lib/postgresql${OLD_VERSION}
```

susemanager/susemanager.changes
```diff
@@ -1,3 +1,5 @@
+- fix db migration script from 12 to 13
```
juliogonzalez, line 1:
```suggestion
- Fix database database migration script from PostgreSQL 12 to 13
```

susemanager/bin/pg-migrate-12-to-13.sh
aaannz, line 95:
```suggestion
    mv /var/lib/pgsql/data-pg$OLD_VERSION /var/lib/pgsql/data
```

juliogonzalez (changes requested):
Each time we use variables without `{}` I am afraid things will break at any moment. I understand we don't change all the rest of the script, but at least lest's start doing it here.

susemanager/bin/pg-migrate-12-to-13.sh
juliogonzalez: I'd not change this, to keep it as we have for the other packages. We don't use `data-pg` anywhere else, but up to Jordi.

aaannz: On a contrary. This is what we use everywhere. `OLD_VERSION` is just a number.

aaannz: See https://github.com/uyuni-project/uyuni/blob/master/susemanager/bin/pg-migrate-12-to-13.sh#L79

jordimassaguerpla: @aaannz Thanks! That was a mistake from my side. Fixed now.

juliogonzalez: Right,. I basically blind :-D

juliogonzalez (approved):

PR #3884, opened 2021-06-16
Improve rhel like product detection

mcalmer:
## What does this PR change?

There is a conflict between Amazon Linux and native RHEL when detecting products.
This PR improve the state to get the data and enhance the unit tests to check if SUSE Products could be found
with what we are parsing.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were improved

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15189
Tracks 

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mcalmer:
@lucidd @hustodemon - best to review commit by commit.
- The first is the real fix.
- The second is a fix to run the testsuite as a user. @lucidd This skip the permission change you implemented. Better ideas are welcome.
- The third is to enhance test tooling we need for the next commit
- The forth is about enhancing the test to really find the SUSE Product in the DB. This is often the bug reported, that a product could not be found.

The last is just the changelog

susemanager-utils/susemanager-sls/salt/packages/redhatproductinfo.sls
```diff
@@ -18,7 +18,7 @@ oraclerelease:
 amazonrelease:
   cmd.run:
     - name: cat /etc/system-release
-    - onlyif: test -f /etc/system-release
+    - onlyif: test -f /etc/system-release -a ! -L /etc/system-release
```
hustodemon, line 21: I'm quite scared about this as I think this can be unreliable:
- this might be true now, but not in the future
- this might be true for other linux distributions, not only amazon linux

Wouldn't it make more sense to look into the `system-release` file and just do a simple grepping, for instance? We could use `file.find` modules with grep parameter (e.g. `^NAME.\*Amazon)`. Quite fishy, but I still think it's much more reliable.

hustodemon (commented):
I've left one comment/question.

lucidd (approved):

susemanager-utils/susemanager-sls/salt/packages/redhatproductinfo.sls
```diff
@@ -18,7 +18,7 @@ oraclerelease:
 amazonrelease:
   cmd.run:
     - name: cat /etc/system-release
-    - onlyif: test -f /etc/system-release -a ! -L /etc/system-release
+    - onlyif: test -f /etc/system-release && grep -i Amazon /etc/system-release >/dev/null
```
hustodemon, line 21: Nit: I think the `-q` (quiet) switch has the same effect as piping grep into `/dev/null` and is a bit nicer: 
```bash
grep -iq Amazon /etc/system-release
```

mcalmer: changed

hustodemon (approved):
LGTM!

Shirocco88 (commented):
public void setSkipSetOwner(boolean skip) {

PR #3885, opened 2021-06-16
Bootstrap via Ansible

hustodemon:
## What does this PR change?

Bootstrapping systems that are behind Ansible controller registered in Uyuni. The bootstrap can be triggered from the Inventories page of an Ansible Control Node.

### How does it work?

We make use of the fact that the ansible controller can run arbitrary playbooks on its systems. Before the bootstrapping itself, we run an ansible playbook (via salt) that copies over the ssh public key of the Uyuni server to the bootstrapped system. Then, the standard bootstrap procedure via salt-ssh is run (the authorization between the Uyuni server and the bootstrap client is done via the copied ssh public key).

Currently, we support 2 authentication methods: pwd and ssh-key, but the current scenario does not fit to any of these. I've implemented a "ansible-preauth" method that tells the bootstrap mechanism, that it doesn't need to care about authentication.

The pre-bootstrap procedure is also depicted here:
graphgenerator svg

## Review

I recommend review commit-by-commit and read the commit messages. Please focus on the authentication mechanism during the review.

## TODO
- [x] all todos in the code
- [x] cleaning up the `authorized_keys` post-bootstrap in case the minion is not bootstrapped as salt-ssh minion

## GUI diff

ss

ss2

- [x] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/14524

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/code/src/com/suse/manager/webui/controllers/utils/RegularMinionBootstrapper.java
```diff
@@ -113,4 +124,49 @@ else if (weRejectedIt) {
         LOG.info("Minion bootstrap success: " + result.isSuccess());
         return result;
     }
+
+    /**
+     * We want to cleanup the authorized ssh key, which was copied by Ansible and was used for authentication of the
+     * primal bootstrap call.
+     *
+     * Since we don't have the minion id of the bootstrap minion yet, we need to use Ansible for removing the authorized
+     * key.
```
mcalmer, line 133: Not sure if we should do it this way.

1. what if they want to do salt-ssh management? AFAIK we need this key where it is.
2. We could also write a salt state which does the cleanup. As the client is onboarded we could do it there as well

Or did I miss something?

susemanager-utils/susemanager-sls/salt/ansible/mgr-ssh-pubkey-authorized.sls
```diff
@@ -0,0 +1,18 @@
+/tmp/mgr-ssh-pubkey-authorized.yml:
+  file.managed:
+    - source: 'salt://ansible/mgr-ssh-pubkey-authorized.yml'
```
mcalmer, line 3: I do not like to use the path as ID.

```suggestion
mgr_ssh_pubkey_authorize_playbook_deploy:
  file.managed:
    - name: /tmp/mgr-ssh-pubkey-authorized.yml
    - source: 'salt://ansible/mgr-ssh-pubkey-authorized.yml'
```

mcalmer (changes requested):

mcalmer:
A bit OT: we had already the question why we cannot bootstrap using our own SSH key. They were taking it and use cloud-init or similar things to deploy it on all instances and just want to "onboard" the client. This is the same as now pre-auth but without the need to deploy the key.

java/code/src/com/suse/manager/webui/controllers/utils/RegularMinionBootstrapper.java
hustodemon: > 1. what if they want to do salt-ssh management? AFAIK we need this key where it is.

Yes, that's why this method is overriden as a no-op in the `SSHMinionBootstrapper`.

> 2. We could also write a salt state which does the cleanup. As the client is onboarded we could do it there as well

This could also do the thing and maybe it'd be more straightforward.

susemanager-utils/susemanager-sls/salt/ansible/mgr-ssh-pubkey-authorized.sls
hustodemon: I do :wink:

java/code/src/com/suse/manager/webui/controllers/utils/RegularMinionBootstrapper.java
hustodemon: I moved the cleanup to salt bootstrap state. I'm going to test it now.

hustodemon:
> A bit OT: we had already the question why we cannot bootstrap using our own SSH key. They were taking it and use cloud-init or similar things to deploy it on all instances and just want to "onboard" the client. This is the same as now pre-auth but without the need to deploy the key.

That sounds reasonable and should be done in a separate issue.

hustodemon:
Hi @mcalmer  , @meaksh , @cbbayburt and @ncounter , I've updated the PR with the following improvements:

- backend + salt: cleanup of the ssh key used for bootstrapping now happens in salt (for non-ssh minions)
- backend (minor): log more details in case of failed pre-auth step
- salt: use temp directory for deploying the temporary ansible playbook for authorizing ssh key. I haven't seen any race condition, but I can imagine when doing parallel bootstraps, this could become an issue.
- frontend: better placement of the bootstrap link (screenshot updated)

The new code starts with the commit "Remove the unused cleanup state". Feel free to review only the parts that are in your scope.

java/code/src/com/suse/manager/webui/controllers/utils/AbstractMinionBootstrapper.java
```diff
@@ -214,8 +223,62 @@ protected BootstrapResult bootstrapInternal(BootstrapParameters params, User use
...
+                    saltApi.callSync(call, minionId).ifPresentOrElse(
+                            res -> {
+                                // all results must be successful, otherwise we throw an exception
+                                List<String> failedStates = res.entrySet().stream()
+                                        .filter(r -> !r.getValue().isResult())
+                                        .map(r -> r.getKey())
+                                        .collect(Collectors.toList());
+
+                                if (!failedStates.isEmpty()) {
+                                    LOG.error("Ansible pre-authentication failed: " + failedStates);
```
ncounter, line 259: Could we add the `minionId` value to this `LOG.error` line? I believe it could be helpful understanding and investigating what went wrong. Or is this information present already in the `failedStates` content?

web/html/src/manager/minion/ansible/accordion-path-content.tsx
```diff
@@ -178,10 +178,15 @@ class AccordionPathContent extends React.Component<PropsType, StateType> {
...
             <ul>
               { content?.unknownSystems.map(s => 
-                <div>
-                  <li key={s + "_hostname"}>{s}</li>
-                  <a href={"/rhn/manager/systems/bootstrap?ansibleInventoryId=" + this.props.path.id  + "&targetHost=" + s}>Bootstrap</a>
-                </div>)}
+                <li key={s + "_hostname"}>
+                  {s}
+                  &nbsp;(
+                  <a href={"/rhn/manager/systems/bootstrap?ansibleInventoryId=" + this.props.path.id  + "&targetHost=" + s}>
```
ncounter, line 184: What about adding a `title={t("Redirect to the Bootstrap page in order to register and manage this system with the Uyuni server by the bootstrap process.")}` (where Uyuni is a parameter for the "SUSE Manager" case).

ncounter (approved):
I left two very small comments.

The overall implementation and the way it works LGTM. Thank you very much for the great effort. :+1:

hustodemon:
Flagging as ready to merge. I run cucumber tests (only the base + salt and traditional "scopes") and it only failed on ubuntu clients (with an unrelated error). Unit tests look good.

I still haven't received the requested reviews for the salt and backend parts and at this time I don't have the capacity to setup everything and do another round of manual testing. Merge on your own risk.

mcalmer (approved):

PR #3890, opened 2021-06-17
Filter template: Live patching based on a system

cbbayburt:
New filter template: "Live patching based on a system" in CLM

## GUI diff
system-filter-template-1
system-filter-template-2

## Documentation
- Documentation issue was created: https://github.com/SUSE/spacewalk/issues/15205

## Test coverage
- Unit tests were added

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14098

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/product/SUSEProductFactory.java
+++ b/java/code/src/com/redhat/rhn/domain/product/SUSEProductFactory.java
@@ -668,6 +668,10 @@ public class SUSEProductFactory extends HibernateFactory {
 
     /**
      * Get all root SUSE products that support live patching
+     *
+     * The query selects all the root products that has a live patching extension product.
+     * It selects only installed products by checking if they have a channel present.
+     *
      * @return the stream of products
      */
     public static Stream<SUSEProduct> getLivePatchSupportedProducts() {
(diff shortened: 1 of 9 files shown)


Etheryte:
One thing we could consider here is whether we want to keep a distinction between a sync and an async select component. I haven't looked into the implications right now, but I would wager that we could use only `AsyncSelect` and simplify the logic we need to maintain.  

One way to do this would be to change `options: Array<Object | string>;` to `options: Array<Object | string> | Promise<Array<Object | string>>;`, and then internally make the regular data usage into an async function that resolves with the data and sets caching to true.  

There may be downsides to this which are not immediately visible purely from code, but I think it would make both using and maintaining this component a lot easier. This of course hinges on the question whether we can provide all of the functionality we need with that approach. I couldn't quite figure out why `value` and `defaultValue` could not be used in the async variant?

cbbayburt:
> I couldn't quite figure out why `value` and `defaultValue` could not be used in the async variant?

In the sync variant, we do some conversion of the options and the value between `string` and `Object` since our `InputBase` context works with strings, but `react-select` works with Objects of value-label kind. For this to work, the options need to be known initially. I didn't spend much time and the async docs are pretty basic, but I'm not sure if the same is possible via callbacks in async version.

If we can make this work, I don't see any other obstacles against using async exclusively :+1:

ncounter (approved):

Etheryte (approved):
Looks good. 👍

PR #3895, opened 2021-06-18
Add salt bundle bootstrapping support and Raspbian 10 and 9

vzhestkov:
## What does this PR change?

Adds support for bootstrapping the minions with Salt Bundle (venv-salt-minion package).
Also implements the bootstrapping of Raspbian 10 and 9.

## Description

Additional files `venv-enabled-$OSARCH$.txt` at the bootstrap repo root are created on creating bootstrap repo with `mgr-create-bootstrap-repo` if there is `venv-salt-minion` package inside. For example `venv-enabled-x86_64.txt` or `venv-enabled-aarch64.txt`. If the repo contains more than one package for different architectures the relevant files are created.

Bootstrap state and bootstrapt script are checking for these files. In case if the file relevant for OS architecture exists, installs the salt bundle package (`venv-salt-minion`) instead of oroginal `salt-minion`. All the config files are readjusted for salt bundle in this case.

It's possible to avoid installing salt bundle with using `--no-bundle` for `mgr-bootstrap` or changing `AVOID_VENV_SALT_MINION` to `1` inside the generated bootstrap script.

For bootstrap state it's possible to use `mgr_avoid_venv_salt_minion` pillar set to `True`

Implemented the state to switch from `salt-minion` to `venv-salt-minion` - `util.mgr_switch_to_venv_minion`
It's better to run it in two steps:
1. Apply it with no pillar specified first to switch to `venv-salt-minion` (copy config files etc., but not cleaning up the original salt-minion confgs and packages)
2. Apply it with `mgr_purge_non_venv_salt` set to `True` to remove `salt-minion` and `mgr_purge_non_venv_salt_files` set to `True` to remove all the files related to `salt-minion`

## GUI diff

No difference.

## Test coverage
- No tests: as the packages itstelf not published yet, it's not possible to test

## Links

https://github.com/SUSE/spacewalk/issues/15041
https://github.com/SUSE/spacewalk/issues/14325

## Changelogs

- add the salt bundle support to mgr-create-bootstrap-repo
- added support of bootstrapping with salt bundle
- added support of bootstrapping Raspbian 10 and 9
  with bootstrap script

## Re-run a test


spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -310,41 +321,60 @@ def getRegistrationStackSh(saltEnabled):
...
+function test_venv_enabled() {{
+    if [ $AVOID_VENV_SALT_MINION -ne 1 ]; then
+        local repourl="$CLIENT_REPO_URL"
+        if [ "$INSTALLER" == "zypper" ] || [ "$INSTALLER" == "yum" ]; then
+            ARCH=$(rpm --eval "%{{_host_cpu}}")
+        else
+            ARCH=$(dpkg --print-architecture)
+        fi
+        VENV_FILE="venv-enabled-$ARCH.txt"
+        $FETCH $repourl/$VENV_FILE
```
mcalmer, line 344: The bootstrap repo is multiarch. This works only, if all architectures of a product has the venv or not.
Shipping venv for x86_64 but not for s390x would not work.
We need to check if this is the case.

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -407,6 +437,9 @@ def getRegistrationStackSh(saltEnabled):
 
     function getY_MISSING() {{
         local NEEDED="{PKG_NAME_YUM}"
+        if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 440: You are comparing Integer with String. This may lead into warnings

```suggestion
        if [ $VENV_ENABLED  -eq 1 ]; then
```

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -450,155 +486,166 @@ def getRegistrationStackSh(saltEnabled):
...
-    Z_CLIENT_CODE_VERSION="${{VERSION:-unknown}}"
-    Z_CLIENT_CODE_PATCHLEVEL="${{PATCHLEVEL:-0}}"
-  }}
-
-  function getZ_MISSING() {{
-    local NEEDED="{PKG_NAME}"
-    if [ "$Z_CLIENT_CODE_BASE" == "sle" -a "$Z_CLIENT_CODE_VERSION" == "10" ]; then
-      # (bnc#789373) Code 10 product migration requires 'xsltproc' being installed
-      which 'xsltproc' || NEEDED="$NEEDED libxslt"
+    if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 489: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -450,155 +486,166 @@ def getRegistrationStackSh(saltEnabled):
...
+        fi
     fi
-  fi
 
-  get_rhnlib_pkgs
-  # try update main packages for registration from any repo which is available
-  zypper --non-interactive up {PKG_NAME_UPDATE} $RHNLIB_PKG ||:
+    get_rhnlib_pkgs
+    # try update main packages for registration from any repo which is available
+    if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 644: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -627,71 +674,86 @@ def getRegistrationStackSh(saltEnabled):
 
     function getA_MISSING() {{
         local NEEDED="salt-common salt-minion"
+        if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 677: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -703,7 +765,11 @@ def getRegistrationStackSh(saltEnabled):
         done
     fi
     # try update main packages for registration from any repo which is available
-    apt-get --yes install --no-install-recommends --only-upgrade salt-common salt-minion ||:
+    if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 768: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -1033,15 +1102,22 @@ def getRegistrationSaltSh(productName):
...
+if [[ $ACTIVATION_KEYS =~ , ]]; then
     echo "*** ERROR: Multiple activation keys are not supported with salt!"
     exit 1
 fi
 
 MINION_ID_FILE="/etc/salt/minion_id"
 SUSEMANAGER_MASTER_FILE="/etc/salt/minion.d/susemanager.conf"
+MINION_SERVICE="salt-minion"
+
+if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 1114: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -1163,102 +1241,126 @@ def getUp2dateTheBoxSh(productName, saltEnabled):
...
     else
         echo "up2date up2date; up2date -p; up2date -uf (conditional)"
     fi
 else
-    if [ "$INSTALLER" == zypper ] ; then
-        echo "zypper --non-interactive up zypper {PKG_NAME_ZYPPER_SYNC}"
-    elif [ "$INSTALLER" == yum ] ; then
-        echo "yum -y upgrade yum {PKG_NAME_YUM_SYNC}"
+    if [ "$INSTALLER" == zypper ]; then
+        if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 1299: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -1163,102 +1241,126 @@ def getUp2dateTheBoxSh(productName, saltEnabled):
...
+            echo "zypper --non-interactive up zypper {PKG_NAME_VENV_ZYPPER_SYNC}"
+        else
+            echo "zypper --non-interactive up zypper {PKG_NAME_ZYPPER_SYNC}"
+        fi
+    elif [ "$INSTALLER" == yum ]; then
+        if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 1305: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -1163,102 +1241,126 @@ def getUp2dateTheBoxSh(productName, saltEnabled):
...
+if [ "$INSTALLER" == zypper ]; then
     zypper lr -u
-    if [ $SALT_ENABLED -eq 0 ] ; then
+    if [ $SALT_ENABLED -eq 0 ]; then
         zypper --non-interactive ref -s
     fi
-    zypper --non-interactive up zypper {PKG_NAME_ZYPPER}
-    if [ $SALT_ENABLED -eq 0 ] ; then
-        if [ -x /usr/sbin/rhn-profile-sync ] ; then
+    if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 1326: same as above

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -1163,102 +1241,126 @@ def getUp2dateTheBoxSh(productName, saltEnabled):
...
+    if [ $FULLY_UPDATE_THIS_BOX -eq 1 ]; then
         zypper --non-interactive up
     fi
-elif [ "$INSTALLER" == yum ] ; then
+elif [ "$INSTALLER" == yum ]; then
     yum repolist
-    /usr/bin/yum -y upgrade yum {PKG_NAME_YUM}
-    if [ $SALT_ENABLED -eq 0 ] ; then
-        if [ -x /usr/sbin/rhn-profile-sync ] ; then
+    if [ "0$VENV_ENABLED" -eq 1 ]; then
```
mcalmer, line 1344: same as above

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -127,16 +129,39 @@ bootstrap_repo:
...
+    - name: rpm --import https://{{ salt['pillar.get']('mgr_server') }}/pub/{{ salt['pillar.get']('gpgkeys:res:file') }}
+    - unless: rpm -q {{ salt['pillar.get']('gpgkeys:res:name') }}
+    - runas: root
+{%- elif grains['os_family'] == 'Debian' %}
+install_gnupg_debian:
+  pkg.latest:
+    - pkgs:
+      - gnupg
+{%- endif %}
+
```
mcalmer, line 144: This duplicate states. One line below we include "channels/gpg-keys.sls". That is the place where it should happen and afaik this is already implemented.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -127,16 +129,39 @@ bootstrap_repo:
...
 
+{%- set salt_minion_name = 'salt-minion' %}
+{%- set salt_config_dir = '/etc/salt' %}
+{% set venv_available_request = salt['http.query'](bootstrap_repo_url + 'venv-enabled-' + grains['osarch'] + '.txt', status=True, verify_ssl=False) %}
+{# Prefer venv-salt-minion if available and not disabled #}
+{%- set use_venv_salt = (0 < venv_available_request.get('status', 404) < 300) and not salt['pillar.get']('mgr_avoid_venv_salt_minion') %}
+{%- if use_venv_salt %}
+{%- set salt_minion_name = 'venv-salt-minion' %}
+{%- set salt_config_dir = '/etc/opt/venv-salt-minion' %}
+{%- endif -%}
```
mcalmer, line 155: I wonder if it is possible to move this to the top of the file.
As this is living in the "general" section without limit to a special OS it would make sense to have all the "set" rules at the top.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -127,16 +129,39 @@ bootstrap_repo:
...
 salt-minion-package:
   pkg.installed:
-    - name: salt-minion
+    - name: {{ salt_minion_name }}
     - install_recommends: False
     - require:
       - file: bootstrap_repo
 
-/etc/salt/minion.d/susemanager.conf:
+{{ salt_config_dir }}/minion.d/susemanager.conf:
```
mcalmer, line 164: nitpick

I am not a fan of filenames as "ids". As we touch them anyway, would it make sense to specify them as "name" and add a real id (mgr_something) to all of these rules?

susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,100 @@
...
+                                                dry_run=True, show_changes=False, ignore_if_missing=True) %}
+{%- set venv_managed_minion=salt['file.file_exists'](venv_susemanager_conf) and
+                            not salt['file.replace'](venv_susemanager_conf, '^master: .*', 'master: ' + pillar['mgr_server'],
+                                                     dry_run=True, show_changes=False, ignore_if_missing=True) %}
+{%- if managed_minion or venv_managed_minion %}
+{%- set pkgs_installed = salt['pkg.info_installed']() %}
+{%- set venv_minion_installed = pkgs_installed.get('venv-salt-minion', False) and True %}
+{%- set venv_minion_available = venv_minion_installed or salt['pkg.latest_version']('venv-salt-minion') or False %}
+{%- if venv_minion_available %}
+venv-salt-minion-pkg:
```
mcalmer, line 14: In theory we have the rule that all ids of states we write should start with `mgr_` . We do not follow this rule in all cases, but when we remember this, we should try to do it :-)

susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,100 @@
...
+  cmd.run:
+    - name: cp -r /etc/salt/minion.d /etc/opt/venv-salt-minion/
+    - require:
+      - pkg: venv-salt-minion-pkg
+    - onlyif:
+      - ([ {{ venv_managed_minion }} = "False" ])
+
+copy-salt-minion-keys:
+  cmd.run:
+    - name: cp -r /etc/salt/pki/minion/minion* /etc/opt/venv-salt-minion/pki/minion/
```
mcalmer, line 39: Are we sure that the destination directory exists? For the normal salt package it does not:
```
$> rpm -qf /etc/salt/pki/minion/
file /etc/salt/pki/minion is not owned by any package
```

mcalmer (changes requested):
In addition to the comments you may want to modify also the package list for generating the bootstrap repo and add the text file there?

spacewalk/certs-tools/rhn_bootstrap_strings.py
vzhestkov: The package itself is arch dependant and in current state we are adding venv-salt-minion to the bootstrap repo almost manually and it's possible in this case that venv-salt-minion for x86_64 is in the repo, but aarch64 is not. I'm now trying to avoid such situation.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
vzhestkov: Unfortunately `bootstrap_repo_url` is calculated far from the top of the sls file, in this case only first 2 sets can e moved to the top. Not so good idea to move just to lines far from the rest logic

susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
vzhestkov: On my test systems it works for openSUSE/CentOS and Raspbian, it's possible that I've appended the dir itself to the package, but will check it anyway.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
vzhestkov: in case of using ids for each file we will have 4 vars instead of one and probably make sls less readable as have to check what each id should contain, for this case `salt_config_dir` defines the root of salt config and we see the relative path of the file. It's looks more clear for me.

spacewalk/certs-tools/rhn_bootstrap_strings.py
vzhestkov: it's quite common solution for the case when the variable is not defined on time of checking.
I can fix it, but in this case need to add explicit setting of this variable to the top of the file otherwise we will get an error on evaluating `if [ -eq 1 ]; then`

mcalmer: This is not correct as typical is, that you compare string with string.

```
   if ["x$VAR" = "x1"]; then
```

Is often used. I am also fine with this solution.

spacewalk/certs-tools/rhn_bootstrap_strings.py
mcalmer: I overlooked that the "arch" is in the filename. So you will have for every arch a dedicated file.
So it looks ok.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
mcalmer: ahh, ok.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
mcalmer: What I mean is:
```
mgr_susemanager_conf:
  file.managed:
    - name: {{ salt_config_dir }}/minion.d/susemanager.conf
    - source:
      - salt://bootstrap/susemanager.conf
   - template: jinja
   - mode: 644
    - require:
      - pkg: salt-minion-package
```

Why do we need 4 vars for this change?
Of cause all the states below changed in the same way and the require statements need to name the id instead of the file path.
But as I said, this is optional

spacewalk/certs-tools/rhn_bootstrap_strings.py
vzhestkov: yup, this is one of the options, but in this case I would prefer to use `if [ $VENV_ENABLED  -eq 1 ]; then` with explicit setting of `VENV_ENABLED` to `0` on top of the file, but it's better to check on the test system first.

susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
mcalmer: In the original salt package the minion directory is not part of the package. This can be different in the new venv package.
If this is the case, fine. Otherwise I would suggest to set the makedirs option.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
vzhestkov: There are few more files handled inside `salt_config_dir` the same way in case of using distinct id to exact this file it's better to use the same approach for all of the files and there are 4 files like this I found on the first glance

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
vzhestkov: I think it was side effect with rebasing the commit, fixed

susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
vzhestkov: Fixed

susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
vzhestkov: `%dir %{_sysconfdir}/opt/%{venv}/pki/minion/` is in the `%files` section of the spec and it's reflected to `deb` with `debbuild` also. I beleive we are safe with it.

mcalmer (approved):
Ok, we could try it

PR #3896, opened 2021-06-18
Use the right repositories for the AlmaLinux 8 Uyuni client tools

juliogonzalez:
## What does this PR change?

We were supposed to use the EL8 client tools, not CentOS8, but since EL8 arrrived later to OBS, we didn't remember to change it.

This PR takes care of this, for new installations, and for existing installations.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: We tested AlmaLinux8, thing is that the CentOS8 client tools work as well, as they use the same sources.

- [x] **DONE**

## Links

None

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


schema/spacewalk/upgrade/susemanager-schema-4.2.14-to-susemanager-schema-4.2.15/000-almalinux-fix-uyuni-tools-repositories.sql
```diff
@@ -0,0 +1,6 @@
+UPDATE rhnContentSource
```
juliogonzalez, line 1: I tested the script at an Uyuni server that had the AlmaLinux 8 client tools added. Did the replacement as expected.

mcalmer (approved):

PR #3898, opened 2021-06-18
Add sanity checks when migrating database and infere values from system

jordimassaguerpla:
## What does this PR change?

We have this request from support to ensure the right version parameters are
used when migrating the database.

We are infering that the original version is the configured one, and the
new version is the latest version installed.

We can assume the new version is the latest installed because
spacewalk-postgreslq requires a specific postgresql version, and given
you are expected to run the migration script after having run "zypper
dup", it is safe to assume you have the right version installed.

Then, before migrating the database, the previous postgresql version is
the one running and the one configured in /var/lib/pgql/data/PG_VERSION.

This also prevents the user from running the migration twice, given
after running the migration, the value in /var/lib/pgsql/data/PG_VERSION
will be updated.

In order to do this, I have extracted the common code from
pg-migrate-10-to-13.sh and pg-migrate-12-to-13.sh to
pg-migrate-x-to-y.sh and made the two be wrappers of the latest.

Then, I added the infere code and the sanity checks into
pg-migrate-x-to-y.sh

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- Doc needed: Add the issue/PR here.
https://github.com/SUSE/spacewalk/issues/15217

- [x] **DONE**

## Test coverage
- No tests

I have run manual testing

- [X] **DONE**

## Links

Fixes # https://github.com/SUSE/spacewalk/issues/15170

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


susemanager/bin/pg-migrate-x-to-y.sh
```diff
@@ -0,0 +1,174 @@
...
+        d)  
+            NV=${OPTARG}
+            ;;
+       f)
+           FAST_UPGRADE=" --link"
+           ;;   
+       *)
+           echo "Usage: $0 [-s initial postgresql to migrate from] [-d destination postgresql to migrate to] [-f fast upgrade]"
+           echo "For example:"
+           echo "  $0 -s10 -d13"
```
mcalmer, line 21: The other scripts call it with a space between option and parameter.
I am not a bash getopts guru, but does it also work when you call `$0 -s 10 -d 13` ?
I would think it is better to document it this way.
If this does not work, you call this script wrong from the others.

jordimassaguerpla: It works with or without space. I just adjusted line 21 to match line 19 for consistency.

mcalmer (approved):

jordimassaguerpla:
> If we are going this **big** rewrite, shouldn't we just detect the current version and upgrade to the new one, without asking the user to specify the versions?

This is what is happening if you do not use the "-s" and "-d" parameters. Without those parameters, the script basically detects which version is configured and which is the latest version installed, and performs the upgrade. The "-s" and "-d" are for maintaining compatibility with what we had released already.

juliogonzalez:
> > If we are going this **big** rewrite, shouldn't we just detect the current version and upgrade to the new one, without asking the user to specify the versions?
> 
> This is what is happening if you do not use the "-s" and "-d" parameters. Without those parameters, the script basically detects which version is configured and which is the latest version installed, and performs the upgrade. The "-s" and "-d" are for maintaining compatibility with what we had released already.

So I understand that:

a) For what you describe, the user would call `pg-migrate-x-to-y.sh`
b) We should document this procedure, at least for `master`, and `Manager-4.2`. (which means you should create an issue for the doc squad, or PRs).

In the end I see a lot of benefits of not asking the user check what versions he should be using:
- Less mistakes (even if now we have checks)
- Easier doc (@0rnela can tell you what a big mess this is :-D)

jordimassaguerpla:
> > > If we are going this **big** rewrite, shouldn't we just detect the current version and upgrade to the new one, without asking the user to specify the versions?
> > 
> > 
> > This is what is happening if you do not use the "-s" and "-d" parameters. Without those parameters, the script basically detects which version is configured and which is the latest version installed, and performs the upgrade. The "-s" and "-d" are for maintaining compatibility with what we had released already.
> 
> So I understand that:
> 
> a) For what you describe, the user would call `pg-migrate-x-to-y.sh`


The user can call pg-migrate-x-to-y.sh or pg-migrate-12-to-13.sh or pg-migrate-10-to-13.sh, the three would work.

> b) We should document this procedure, at least for `master`, and `Manager-4.2`. (which means you should create an issue for the doc squad, or PRs).
> 
> In the end I see a lot of benefits of not asking the user check what versions he should be using:
> 
>     * Less mistakes (even if now we have checks)
> 
>     * Easier doc (@0rnela can tell you what a big mess this is :-D)

Basically depends if we want to make calling pg-migrate-x-to-y.sh the default. I see it a good idea for master. My doubt is for Manager-4.2, given this change is not part of GMC. Should we make calling pg-migrate-x-to-y.sh the default for 4.2 and so update the docs?

juliogonzalez:
> Basically depends if we want to make calling pg-migrate-x-to-y.sh the default. I see it a good idea for master. My doubt is for Manager-4.2, given this change is not part of GMC. Should we make calling pg-migrate-x-to-y.sh the default for 4.2 and so update the docs?

I'd say we should have it for 4.2.1. Please join the doc daily today at 11.20.

jordimassaguerpla:
Docs have been merged https://github.com/uyuni-project/uyuni-docs/pull/1025

jordimassaguerpla:
we can merge this one now.

PR #3905, opened 2021-06-22
Add support for Kiwi options

nadvornik:
## What does this PR change?

This PR adds support for Kiwi options, including --profile. This is necessary for building unmodified image sources from Build Service.

--profile option can be specified multiple times, there is also --type option with similar use, so this PR stores the unmodified Kiwi command line in DB, to allow more flexibility.

Additional validation can be implemented later in front end.

## GUI diff

Before:

After:

- Kiwi options field has been added to kiwi profile page

- [X] **DONE**

## Documentation

WIP

- [ ] **DONE**

## Test coverage

- Added to unit tests

- [x] **DONE**

## Links

This is minimal implementation of https://github.com/SUSE/spacewalk/issues/9264
Part of https://github.com/SUSE/spacewalk/issues/14570

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/image/profile/ImageProfileHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/image/profile/ImageProfileHandler.java
@@ -107,6 +107,7 @@ public class ImageProfileHandler extends BaseHandler {
      * @param storeLabel the image store label
      * @param path the path or git uri to the source
      * @param activationKey the activation key which defines the channels
+     * @param kiwiOptions command line options passed to Kiwi
      * @return 1 on success
      *
      * @xmlrpc.doc Create a new Image Profile
@@ -116,10 +117,11 @@ public class ImageProfileHandler extends BaseHandler {
      * @xmlrpc.param #param("string", "storeLabel")
      * @xmlrpc.param #param("string", "path")
      * @xmlrpc.param #param_desc("string", "activationKey", "Optional")
+     * @xmlrpc.param #param("string", "kiwiOptions")
      * @xmlrpc.returntype #return_int_success()
      */
     public int create(User loggedInUser, String label, String type, String storeLabel,
-            String path, String activationKey) {
+            String path, String activationKey, String kiwiOptions) {
         ensureImageAdmin(loggedInUser);
 
         if (StringUtils.isEmpty(label)) {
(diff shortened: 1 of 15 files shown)


java/code/src/com/redhat/rhn/frontend/xmlrpc/image/profile/ImageProfileHandler.java
```diff
@@ -116,10 +117,11 @@ public ImageProfile getDetails(User loggedInUser, String label) {
      * @xmlrpc.param #param("string", "storeLabel")
      * @xmlrpc.param #param("string", "path")
      * @xmlrpc.param #param_desc("string", "activationKey", "Optional")
+     * @xmlrpc.param #param_desc("string", "kiwiOptions", "Optional")
      * @xmlrpc.returntype #return_int_success()
      */
     public int create(User loggedInUser, String label, String type, String storeLabel,
-            String path, String activationKey) {
+            String path, String activationKey, String kiwiOptions) {
```
aaannz, line 124: This does not seems really optional to me.
How will this behave when you call create from xmlrpc and omit `kiwiOptions` option?

Seems to me, because of all the changed unit tests that it will not work -> breaking change in API.
I would rather create new overloaded create function with added kiwiOptions param.

schema/spacewalk/common/tables/suseKiwiProfile.sql
```diff
@@ -19,7 +19,8 @@ CREATE TABLE suseKiwiProfile
                   CONSTRAINT suse_kiwi_prid_fk
                      REFERENCES suseImageProfile (profile_id)
                      ON DELETE CASCADE,
-    path        VARCHAR(1024) NOT NULL
+    path        VARCHAR(1024) NOT NULL,
+    kiwi_options VARCHAR(256)
```
aaannz, line 23: Just wondering, Is 256 enough for potentially lengthy options? Considering this is not only for profiles, but generic options.

aaannz (changes requested):
I am concerned about API change.
Do not forget to remove WIP: prefix from the PR name.

nadvornik:
Updated according to the feedback.

aaannz (approved):
LGTM. Remember to add all three changelogs - java, schema, html? - and remove WIP: prefix from PR name before merging.

mbologna (approved):

PR #3909, opened 2021-06-23
Regen sls files when chaning machine

mcalmer:
## What does this PR change?

When a system change its machine id, we need to remove old SLS files and regenerate them as they have the machine id in their name.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **unable to test file changes as they are disabled in tests**

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15231
Tracks https://github.com/SUSE/spacewalk/pull/15278

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/StatesAPI.java
+++ b/java/code/src/com/suse/manager/webui/controllers/StatesAPI.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015 SUSE LLC
+ * Copyright (c) 2015--2021 SUSE LLC
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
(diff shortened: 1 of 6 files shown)


java/code/src/com/suse/manager/webui/controllers/StatesAPI.java
```diff
@@ -657,13 +656,32 @@ else if (state.getPackageState() == PackageStates.REMOVED) {
...
+        Path baseDir = Paths.get(
+                SaltConstants.SUMA_STATE_FILES_ROOT_PATH, SaltConstants.SALT_PACKAGES_STATES_DIR);
+        Path filePath = baseDir.resolve(getPackagesSlsName(new MinionSummary(minion)));
+
+        try {
+            Files.deleteIfExists(filePath);
+        }
+        catch (IOException e) {
+            LOG.error(e.getMessage(), e);
+            throw new RuntimeException(e);
```
cbbayburt, line 674: Nitpick: Probably this won't be hit much but in this case, do we really want to cut the execution? Maybe just logging the error and continuing is a better option. We can even propagate the `IOException` and decide what to do with it from the calling code.

cbbayburt (approved):
LGTM :+1: 
Just one nitpick below, feel free to ignore :)

moio (approved):
Changes LGTM, I'd still recommend to make sure unit tests cover this case well.

mcalmer:
In unit tests we cannot check this, as the file changes are disabled. Could be a candidate for cucumber tests.
I will open an issue

PR #3918, opened 2021-06-28
Testsuite - workaround to assure no dhcp packages are installed on proxy

lkotek:
## What does this PR change?

Workaround for Leap 15.3 JeOS OpenStack image, which contains dhcp + dhcp-client in version 4.3.6-P1 (differs from the version in repository causing highstate to fail).

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/core/proxy_branch_network.feature
+++ b/testsuite/features/core/proxy_branch_network.feature
@@ -15,6 +15,13 @@ Feature: Setup SUSE Manager for Retail branch network
 
 @proxy
 @private_net
+  # WORKAROUND
+  Scenario: Remove dhcp packages on the proxy
+    When I remove package "dhcp dhcp-client" from this "proxy"
+  # End of WORKAROUND
+
+@proxy
+@private_net
   Scenario: Install or update branch network formulas on the server
     When I manually install the "branch-network" formula on the server
     And I manually install the "dhcpd" formula on the server


testsuite/features/core/proxy_branch_network.feature
```diff
@@ -21,6 +21,14 @@ Feature: Setup SUSE Manager for Retail branch network
     And I manually install the "bind" formula on the server
     And I synchronize all Salt dynamic modules on "proxy"
 
+@proxy
+@private_net
+  # WORKAROUND
+  Scenario: Remove dhcp packages on the proxy
+    When I remove package "dhcp" from this "proxy"
+    And I remove package "dhcp-client" from this "proxy"
```
Bischoff, line 29: maybe `When I remove package "dhcp dhcp-client" from this "proxy"` would have worked. If i remember correctly, you can pass a packages list.

testsuite/features/core/proxy_branch_network.feature
```diff
@@ -21,6 +21,14 @@ Feature: Setup SUSE Manager for Retail branch network
     And I manually install the "bind" formula on the server
     And I synchronize all Salt dynamic modules on "proxy"
 
+@proxy
+@private_net
+  # WORKAROUND
```
Bischoff, line 26: ok please put bsc# if you create a bug

Bischoff (approved):
lgtm, just small suggestions

testsuite/features/core/proxy_branch_network.feature
lkotek: Done, thanks :-)

testsuite/features/core/proxy_branch_network.feature
```diff
@@ -21,6 +21,13 @@ Feature: Setup SUSE Manager for Retail branch network
...
+  Scenario: Remove dhcp packages on the proxy
+    When I remove package "dhcp dhcp-client" from this "proxy"
+  # End of WORKAROUND
```
srbarrios, line 29: Be sure the Build Validation code is also working correctly if the same case comes up there.

testsuite/features/core/proxy_branch_network.feature
```diff
@@ -21,6 +21,13 @@ Feature: Setup SUSE Manager for Retail branch network
     And I manually install the "bind" formula on the server
     And I synchronize all Salt dynamic modules on "proxy"
 
+@proxy
+@private_net
+  # WORKAROUND
+  Scenario: Remove dhcp packages on the proxy
```
srbarrios, line 27: I would personally do this scenario as the first scenario, before the installation of formulas, as it is more like a pre-requisite.

srbarrios (changes requested):

srbarrios (approved):
Some remarks, approving, but take into account, please.

testsuite/features/core/proxy_branch_network.feature
lkotek: No problem with that, moving :-)

testsuite/features/core/proxy_branch_network.feature
lkotek: It shouldn't be necessary. This fix is intended as quick workaround for temporary issue. At the end (and I hope quickly) there are expected to be same package versions at JeOS level and also in the repos.

testsuite/features/core/proxy_branch_network.feature
lkotek: From what I know it shouldn't be necessary, the package version is already released for all other relevant systems (Leap 15.2, SLE 15 SP3). I haven't find reason for Leap 15.3, but the whole situation currently doesn't sound like bug to me.

Issue #3559, opened 2021-04-14
Unable to sync flat repository format for Debian repositories


lumarel:
# Additional Information

I tried to add and sync the Jenkins Debian Repository ([pkg.jenkins.io](https://pkg.jenkins.io/debian-stable/)) to our Uyuni Server, but I was unfortunately not able to find the correct way to insert the url so it gets downloaded.
Maybe I'm just missing something or maybe it is not possible.

## Version of Uyuni Server
`2021.02-116.3` on `OpenSUSE Leap 15.2` (this system has been setup lately so the packages are new)

## Details about issue

I'm following the guide in the documentation about the [Custom Channel Creation](https://www.uyuni-project.org/uyuni-docs/uyuni/administration/custom-channels.html#_creating_custom_channels_and_repositories), so I created a Channel via the UI and added a Repository with the deb type.

Configuration of the softwarechannel:
```
Label:              ubuntu-1804-amd64-jenkins
Name:               Jenkins repository for Ubuntu 18.04 AMD64
Architecture:       AMD64 Debian
Parent:             ubuntu-18.04-pool-amd64-uyuni
Systems Subscribed: 3
Number of Packages: 0

Summary
-------
Jenkins repository for Ubuntu 18.04 AMD64

GPG Key:            45F2C3D5
GPG Fingerprint:    62A9 756B FD78 0C37 7CF2  4BA8 FCEF 32E7 45F2 C3D5
GPG URL:            https://pkg.jenkins.io/debian-stable/jenkins.io.key
GPG CHECK:          True

Repos
-----
External - Jenkins repository AMD64
```

Configuration of the repository:
```
Repository Label:                  External - Jenkins repository AMD64
Repository URL:                    https://pkg.jenkins.io/debian-stable/binary/
Repository Type:                   deb
Repository SSL Ca Certificate:     None
Repository SSL Client Certificate: None
Repository SSL Client Key:         None
```

And here is the current output of the spacewalk-repo-sync:
```
2021/04/14 14:02:32 +02:00 Command: ['/usr/bin/spacewalk-repo-sync', '--channel', 'ubuntu-1804-amd64-jenkins', '--type', 'deb', '--non-interactive']
2021/04/14 14:02:32 +02:00 Sync of channel started.
2021/04/14 14:02:36 +02:00 Repo URL: https://pkg.jenkins.io/debian-stable/binary/
2021/04/14 14:02:36 +02:00     Packages in repo:               108
[...]

PR #3921, opened 2021-06-28
Non-flat Debian repositories handling improved

vzhestkov:
## What does this PR change?

According to [Debian repository specification](https://wiki.debian.org/DebianRepository/Format#Overview) the repository definition format is the following:
```
deb uri distribution [component1] [component2] [...]
```
For example:
```
deb https://deb.debian.org/debian/dists stable main
```
or 
```
deb https://pkg.jenkins.io/debian-stable binary/
```

For each pair of distribution+component it defines distinct URL:
https://deb.debian.org/debian/dists/stable/main

For `spacewalk-repo-sync` we are providing URL only with no `distribution` and `component`. But the other important URL part is the root of the repository while the URL path, we are pointing to, contains `Release` file, 'Packages(.xz|.gz|)'. It means we have no root of the distro specified in the repo and have no any field to store it.

This change is providing the following way of specification a repository URL:
```
https://pkg.jenkins.io/debian-stable/binary/../
```
In this case the URL path to the `Release` and 'Packages(.xz|.gz|)' defined as `https://pkg.jenkins.io/debian-stable/binary/` while the repository root is set to `https://pkg.jenkins.io/debian-stable/` with the specified steps to the parent with `../` as the root of the repo and the URL path to the `Release` can be on the other levels of depth.

The solution changes only handling the URLs with any number of trailing `../`, the previous behavior for other URL formats is not affected. URL format kept safe.

## GUI diff

No difference.

## Documentation
required

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/3559

## Changelogs

- Non-flat Debian repositories handling improved

## Re-run a test


mcalmer (changes requested):
I do not like the .. in the repo path. This create all kind of trouble when something else want to use the the URL. Let's talk about other solutions.

vzhestkov:
@mcalmer the other solution is to implement handling apt source list entry as `deb uri distribution [component1] [component2] [...]` but it's way much more effort and will lead to harm URL format if use the same field `Repository URL*:`, or add `Repository Type` different than `deb`. `..` usage in the URL keeps the URL format valid and URL is still accessible with this format, but yes looks confusing and requires expalanations. This is the easies way I found to pass two distinct URL pathes with the only URL value.

mcalmer (changes requested):
Is it really needed to introduce apt_src? This would require a lot of changes also in the java code and a lot of guessing if we should use "deb" or "apt". 
Can't we just implement this in "deb" ?

vzhestkov:
@mcalmer hmm, we can. Probably I missunderstood you when we discussed it. I supposed you mean that the fix shouldn't affect deb plugin. I can just move all the logic to deb and process URL different way if `uyuni_suite` query parameter is in the URL.

backend/satellite_tools/repo_plugins/apt_src.py
```diff
@@ -0,0 +1,102 @@
...
+from spacewalk.satellite_tools.repo_plugins.deb_src import ContentSource as deb_ContentSource
+from spacewalk.satellite_tools.syncLib import log2
+from spacewalk.server import rhnSQL
+
+try:
+    #  python2
+    from urllib import unquote
+    import urlparse
+except ImportError:
+    #  python3
```
agraul, line 27: Is Python2 compatibility needed?

vzhestkov: After discussion with @mcalmer decided to move the logic to `deb` plugin as additional plugin requre a lot of changes outside plugin itself.

vzhestkov:
@mcalmer I've moved all the logic back to `deb_src.py`

mcalmer (approved):
Looks good, but please open a doc issue and link it here. 
Maybe you can also check at which place we can document it.
In best case you can start a PR.

Thanks

backend/satellite_tools/repo_plugins/deb_src.py
```diff
@@ -94,6 +94,66 @@ def __init__(
...
+            if "/" not in suite:
+                if arch is None:
+                    rhnSQL.initDB()
+                    h = rhnSQL.prepare("""
+                                       SELECT ca.label AS arch_label
+                                       FROM rhnChannel AS c
+                                       LEFT JOIN rhnChannelArch AS ca
+                                           ON c.channel_arch_id = ca.id
+                                       WHERE c.label = :channel_label
+                                       """)
```
agraul, line 134: BTW, this string contains the (common) leading white space. There is `textwrap.dedent()` if that's not wanted .Here I think it's not a big deal, just caught my attention now.

```python
>>> if True:
...     if True:
...         if True:
...             print("""
...                   SELECT something
...                   FROM somewhere
...                   WHERE condition
...                   """)
... 

                  SELECT something
                  FROM somewhere
                  WHERE condition
>>> import textwrap
>>> if True:
...     if True:
...         if True:
...             print(textwrap.dedent("""
...                                   SELECT something
...                                   FROM somewhere
...                                   WHERE condition
...                                   """))
... 

SELECT something
FROM somewhere
WHERE condition


```

PR #3931, opened 2021-07-01
Reduce the time to 

jordimassaguerpla:
## What does this PR change?

I am testing to reduce the timeouts for syncing SCC products so that the acceptance tests for PR do not take that long.

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -306,11 +306,15 @@ end
 #
 # This function is written as a state machine. It bails out if no process is seen during
 # 30 seconds in a row, or if the whitelisted reposyncs last more than 7200 seconds in a row.
+# 7200 is the default value but it can be shortened by setting the TEST_REPOSYNC_TIMEOUT
+# environment variable.
 When(/^I kill all running spacewalk\-repo\-sync, excepted the ones needed to bootstrap$/) do
   do_not_kill = compute_list_to_leave_running
   reposync_not_running_streak = 0
   reposync_left_running_streak = 0
-  while reposync_not_running_streak <= 30 && reposync_left_running_streak <= 7200
+  reposync_timeout = 7200
+  reposync_timeout = 600 if $type_environment && $type_environment == 'PULL_REQUEST_TESTING'
+  while reposync_not_running_streak <= 30 && reposync_left_running_streak <= reposync_timeout
     command_output, _code = $server.run('ps axo pid,cmd | grep spacewalk-repo-sync | grep -v grep', false)
     if command_output.empty?
       reposync_not_running_streak += 1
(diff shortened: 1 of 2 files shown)


jordimassaguerpla:
Reducing to 3600 made no difference.

Let's be more aggressive and try with 600 seconds.

jordimassaguerpla:
> Reducing to 3600 made no difference.
> 
> Let's be more aggressive and try with 600 seconds.

With 600, the core setup takes 40 minutes instead of 1h and 15 minutes. The "Initialize clients" phase passed after the core setup.

@srbarrios @Bischoff the question is, how can I make 600 be the timeout instead of 7200 when running this test at the PR? I assume you have 7200 for a reason and you want to keep that when running the test for master branch. Can we make it configurable somehow? What is the simplest solution here?

srbarrios:
> > Reducing to 3600 made no difference.
> > Let's be more aggressive and try with 600 seconds.
> 
> With 600, the core setup takes 40 minutes instead of 1h and 15 minutes. The "Initialize clients" phase passed after the core setup.
> 
> @srbarrios @Bischoff the question is, how can I make 600 be the timeout instead of 7200 when running this test at the PR? I assume you have 7200 for a reason and you want to keep that when running the test for master branch. Can we make it configurable somehow? What is the simplest solution here?

This timeout waits until we can assure that the channel is fully sync, checking the existence of a file in a expected folder. But in case of a timeout, we are catching the exception and continue (see line 346), without returning an exception, this was done just in case we pass wrong MU URLs at some point (with a wrong channel that will never be found on that folder) but we must revisit this (@Bischoff fyi), and the good practice might be to don't hide this exception.

If the core setup was sped up from 75min to 40min is due to failing faster, but it doesn't mean it's stable to continue the tests as it must be that we still having pending synchronization processes.

To summarize, we can fail fast ofc, but that will speed up results (failed results) but not the tests themselves.

We can also parametrize this depending on the environment where we run it, for that, I would create a global variable as we do in this example, but specifying as value the kind of environment we run:
``` 
$long_tests_enabled = true if ENV['LONG_TESTS'] == 'true'
``` 
For instance, something like:
``` 
$type_environment = ENV['TYPE_ENVIRONMENT']
or
$kind_of_env = ENV['TYPE_ENVIRONMENT']
```

jordimassaguerpla:
Do we need to have the channel full synced for the tests to run? Or just the metadata?

jordimassaguerpla:
@srbarrios I just introduced the use of an environment variable as you suggested. This is to overwrite the default, so nothing should change for the master acceptance tests. This way I can set this for the acceptance tests for the PR. We can set this as experimental and give it a try for some weeks and see.

srbarrios (approved):

Bischoff:
ok. `type_environment` is not really self-explanatory. What do you mean @jordimassaguerpla ? If you meant "type of environment" it would rather be `environment_type`, in correct English.

Bischoff:
We want to get rid of `LONG_TESTS`, so it's not really one of the best examples.

I'd also like to have a common value for all environments - no env variable.

But something's deeply wrong with this PR. Hold on.

Bischoff:
After discussing with Jordi, I asked him to simply revert @srbarrios .

We can do 3 things with reposync:
1. let it running during the rest of the test suite
2. kill it
3. wait until it finishes

1. is what we did before introducing those two functions. **we do not want that** anymore as it generated race conditions. Unfortunately, Jordi's PR returned to this situation.

If you really don't want to wait for reposyncs nor to kill them, then you should simply not call those two methods. But that's dangerous.

This needs more discussion. For now, let's revert.

PR #3937, opened 2021-07-02
Add some delay after SP migration before scheduling package refresh action (bsc#1187963)

meaksh:
## What does this PR change?

This PR adds a 30 seconds delay before triggering a "Package List Refresh" action after a SP migration action is performed on a minion.

The problem before this PR is that the "Package List Refresh" action is scheduled just after the Salt return event is received from the minion when the "distupgrade" state has been executed, and it's as this time when the "salt-minion" services is restarted after the Salt update done during the SP migration.

Therefore, there is a high probability that "Package List Refresh" action is triggered when the minion is not fully reloaded and then the action will be marked as "Failed" due minion is not reachable at the moment.

To workaround this situation, this PR simply schedule "Package Refresh List" action to be triggered 30 seconds in the future, so we give some time for the minion to be fully restarted and accepting jobs.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered in cucumber

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/15284

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/action/ActionManager.java
+++ b/java/code/src/com/redhat/rhn/manager/action/ActionManager.java
@@ -1145,8 +1145,8 @@ public class ActionManager extends BaseManager {
      */
     public static PackageAction schedulePackageRefresh(Org schedulerOrg, Server server)
             throws TaskomaticApiException {
-	Date earliest = new Date();
-	return schedulePackageRefresh(schedulerOrg, server, earliest);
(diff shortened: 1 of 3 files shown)


admd (approved):
:+1:

PR #3945, opened 2021-07-06
Maintenance windows web calendar improvements

parlt91:
## What does this PR change?

Adds some improvements to the maintenance windows calendar view as well as some `junit` tests.
**Note:** Still need to provide some more event colors

## GUI diff

After:
start-monday
start-sunday
event-coloring
show-timezone

- [x] **DONE**

## Documentation
- No documentation needed: 

- [ ] **DONE**

## Test coverage

- Unit tests were added

- [x] **DONE**

## Links

Tacks https://github.com/SUSE/spacewalk/issues/14744

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -24,6 +25,15 @@ type Event = {
   title: string;
 }
 
+// TODO: Add more colors
```
Etheryte, line 28: I would recommend picking something like the LCH color space and generating a long list of colors with a comment on how to add more colors if needed later. See https://darker.ink/writings/Towards-richer-colors-on-the-Web#consistent-and-predictable-colors for background on why LCH is a good choice for this. Could also pick up a library for this and generate them on the fly, but I prefer the static pre-built colors variant.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -203,6 +216,39 @@ const WebCalendar = (props: WebCalendarProps) => {
...
+    if (props.eventNames && props.eventNames.length > 0) {
+      props.eventNames.forEach((name, i) =>
+        events.filter(event => event.title === name).map(event =>
+          Object.assign(event, {color: colors[(i + colors.length) % colors.length]})
+        )
+      );
+    }
+    events.forEach(event =>
+      Object.assign(event, {
+        start: event.start.split(" ")[0],
```
Etheryte, line 230: I'm guessing `event.start` is a string a-la `9:00am CEST`? I would recommend passing an ISO datetime string or a Date object and then using a-la `moment.format()` instead. This will allow us to easily accommodate for user-defined formats in the future for https://github.com/SUSE/spacewalk/issues/9049

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -203,6 +216,39 @@ const WebCalendar = (props: WebCalendarProps) => {
...
+  }
+
+  const setEventContent = (eventContent) => {
+    if (eventContent.view.type === "dayGridMonth") {
+      return <>{eventContent.timeText} {eventContent.event.extendedProps.startTimezone} {eventContent.event.title}</>;
+    }
+    else {
+      // Add the timezone to the timeText string e.g. '9:00am - 12:00pm' becomes '9:00am CEST - 12:00pm CEST'
+      return <>
+        {eventContent.timeText.replace(/m/, "m " + eventContent.event.extendedProps.startTimezone)
```
Etheryte, line 246: I'm not sure if this is due to how the calendar library handles data or something else, but this looks a bit off. Would it be possible to get the correct value from the data we know instead of doing a regex replace?

Etheryte (changes requested):
The frontend part looks good in general. 👍  
Had a few small comments and questions.

java/code/src/com/suse/manager/maintenance/MaintenanceManager.java
```diff
@@ -580,18 +583,24 @@ else if (operation.equals("skipNext")) {
...
-        Map<String, Long> activeRange = getActiveRange(date);
+        Map<String, Long> activeRange = getActiveRange(user, date, startOfWeek);
         Long start = activeRange.get("start");
         Long end = activeRange.get("end");
 
         return icalUtils.getCalendarEvents(calendar, eventName, start, end);
     }
 
-    private Map<String, Long> getActiveRange(Long date) {
+    private Map<String, Long> getActiveRange(User user, Long date, Long startOfWeek) {
```
hustodemon: The parameter `user` does not seem to be used.

web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -49,10 +50,11 @@ const WebCalendar = (props: WebCalendarProps) => {
   const getEvents = (operation) => {
     const date = getDate(operation);
     if (operation === "initial" || needsUpdate(date)) {
-      const endpoint = `/rhn/manager/api/maintenance/events/${operation}/${props.type}/${date.valueOf()}/${props.id}`;
+      const startOfWeek = getApi().currentDataManager.data.dateEnv.weekDow;
```
hustodemon, line 78: I couldn't find any reference to `weekDow` in the full calendar api docs. Is this safe to use? What about https://fullcalendar.io/docs/firstDay ?

hustodemon: Moreover: is this passing of the first day necessary? Wouldn't it be enough to compute the first day of week based on the users preferred locale on the java side?

java/code/src/com/suse/manager/maintenance/MaintenanceManager.java
```diff
@@ -528,12 +528,13 @@ public int retractScheduleFromSystems(User user, Set<Long> systemIds) {
      * @param date the date to start looking for maintenance windows
      * @return the resulting list of maintenance windows
      */
-    public List<MaintenanceWindowData> preprocessCalendarData(User user, String operation, Long id, Long date) {
+    public List<MaintenanceWindowData> preprocessCalendarData(User user, String operation, Long id, Long date,
+                                                              Long startOfWeek) {
```
hustodemon, line 533: Nit: using 64 bits for passing 1 bit of information is a bit overkill :smile: \scnr

hustodemon (commented):
Left couple of comments and questions.

java/code/src/com/suse/manager/maintenance/MaintenanceManager.java
parlt91: Yeah it was removed in a later commit.

java/code/src/com/suse/manager/maintenance/MaintenanceManager.java
parlt91: Haha well yeah you are absolutely right. A boolean would do here as well :sweat:
Dunno what rid me there :rofl:

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: That one let's you set the first day of the week that `fullcalendar` should use, but it actually has built in support for the default start of the week depending on locale.

The second one is a bit of a difficult one. Let me try to explain.
The java class that spits out the first day of the week expects locales passed to it to have both language and country defined (meaning e.g. de_DE) but for a lot of the locales we use in suma we do not define the country (e.g. de, fr) and just passing the language (e.g. de) will always result in sunday being the first day of the week.

Since I didn't want to go through guessing the country we are using for these cases I decided to take `fullcalendars` build in support for this as the source of truth.
I actually went through all the languages we use and compared `fullcalendars` built in default with what's supposed to be the default for the country using the language and they matched in all the cases. So I think we are on the safe side in this case.

Etheryte (approved):
Looks great, thanks for going through the comments. 👍

web/html/src/manager/maintenance/calendar/web-calendar.tsx
hustodemon: Ok, thanks for explaining.

hustodemon (approved):
LGTM, thanks!

PR #3950, opened 2021-07-07
Speed up adding products via Web UI (bsc#1187727)

hustodemon:
## What does this PR change?

Adding multiple products via the product wizard was too slow and lead to http timeout (10 minutes).

Products are added in a sequence and each step took approximately one second longer than the previous one, making the complexity `O(n^2)` (n = number of added products), which made it impossible to add more than ~35 products. This was caused by flushing hibernate cache, that grew with each step.

Solved by executing part of the code without auto flushing. Now adding a product takes approximately 3 seconds and this number doesn't grow with the number of products added (`O(n)`). Roughly said, the 10-minute timeout now allows adding ~200 products with a single UI action.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
bugfix

- [x] **DONE**

## Test coverage
- No tests: only affects performance

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/15245

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
+++ b/java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
@@ -16,6 +16,7 @@ package com.redhat.rhn.manager.setup;
 
 import com.redhat.rhn.common.db.datasource.ModeFactory;
 import com.redhat.rhn.common.db.datasource.SelectMode;
+import com.redhat.rhn.common.hibernate.HibernateFactory;
 import com.redhat.rhn.common.messaging.MessageQueue;
 import com.redhat.rhn.domain.channel.ChannelFactory;
 import com.redhat.rhn.domain.product.MgrSyncChannelDto;
@@ -32,13 +33,18 @@ import com.redhat.rhn.taskomatic.domain.TaskoRun;
 import com.redhat.rhn.taskomatic.domain.TaskoSchedule;
 import com.redhat.rhn.taskomatic.task.RepoSyncTask;
 import com.redhat.rhn.taskomatic.task.TaskConstants;
+
 import com.suse.manager.model.products.Channel;
 import com.suse.manager.model.products.MandatoryChannels;
 import com.suse.manager.model.products.OptionalChannels;
 import com.suse.mgrsync.MgrSyncStatus;
+import com.suse.salt.netapi.utils.Xor;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
```diff
@@ -157,6 +166,33 @@ public void addProduct(String productIdent, User user) throws ProductSyncExcepti
...
+     * @return Xor with the result on the left side, or with an exception on the right side
+     * @throws ProductSyncException in case of error
+     */
+    private SetupWizardProductDto findProductWithoutAutoFlush(String productIdent) throws ProductSyncException {
+        HibernateFactory.getSession().flush();
+        Xor<SetupWizardProductDto, ProductSyncException> result = HibernateFactory.doWithoutAutoFlushing(() -> {
+            try {
+                return Xor.left(findProductByIdent(productIdent));
+            }
+            catch (ProductSyncException e) {
```
moio, line 183: `ProductSyncException` seems to be used in this class exclusively, and only in a few places. Outside of this class, the only thing that is consumed out of it is the string message (in the one caller of `addProducts`).

What would you think about turning `ProductSyncException` into a `RuntimeException`, and to return an (`Optional`) `String` message instead of a `Throwable` from `addProducts`, thereby removing the `Xor` contortion here (which is by the way internal to netapi), and the wrapping of `Throwable` into `Optional` (aka "poor-man's Xor") above?

hustodemon: > ProductSyncException seems to be used in this class exclusively

That's a great catch which I did not see. Let's make use of it! I don't like the structure of the code (that's why I hid it in the separate method `findProductWithoutAutoflush`).

> (which is by the way internal to netapi)

Yes, I always feel a bit guilty of using it in Uyuni, but I also see other usages in our codebase.

rjmateus (commented):
The code change looks good to me, and agree with Moio's comments.
But there is something that I didn't get. you referred that "This was caused by flushing hibernate cache". 
But if I remember correctly is the equals method on all database entities is correctly implemented, each flush should only save to the database the objects that are not yet flushed.
 Did you check which inserts/updates were constantly running in the database?

hustodemon:
@rjmateus I haven't checked the low level db operations. I only used profiler. I'm attaching a screenshot and the dump. Let me know, if this answers your questions.

ss

rjmateus:
@hustodemon the print shows the problem is when we perform the flush, but doesn't tell us what are entities/queries are causing this flush explosion. 
My concern is that we may be solving the problem in this case, but we are not addressing the root cause which is the flush explosion.
If we call multiple times the flush, entities already synced with the database should not be synced (flushed) again. If that is happening it may be telling us that something else is not working as expected.
The only way I know to check what entities are being constantly flushed to the database is to increase the logs and see the queries that are being executed. But maybe @moio knows other ways.

hustodemon:
The user behind this bug haven't responded. Closing.

PR #3958, opened 2021-07-07
Add timezone support to datetime pickers

Etheryte:
## What does this PR change?

This is the frontend part for the timezone update work.  

The central logic for datetime handling is in [`localizedMoment.ts`](https://github.com/uyuni-project/uyuni/pull/3958/files#diff-d14bb0f8bb7fe7ad2cfcb5799a0b6e9e9f6ab273fb925f4a723c5b8013369069). The updated datetime picker can be found in [`DateTimePicker.tsx`](https://github.com/uyuni-project/uyuni/pull/3958/files#diff-ba7deb3def321c49b4cfc1d24685030d73890e5269ce6ea047a20fb33ed0c41f) which is a restructure and refacto of the old picker component. I've also updated the directory structure to accommodate for tests etc.  

Additionally, this PR includes bindings to use the same picker component on the legacy pages, see [`legacy/DateTimePicker.tsx`](https://github.com/uyuni-project/uyuni/pull/3958/files#diff-192c0cebf4375d76fc867cbfebadf89f3aaea4db538331a7d1e80fd79bc0b5e2) and [`DateTimePicker.java`](https://github.com/uyuni-project/uyuni/pull/3958/files#diff-7656344eff59a2ac36b5492712e2c2b81bddfb6c6f451f5a59d6ca3169b389bf) for more details on that.  

Don't try to review this commit-by-commit because I tried a number of different approaches before settling on the final one.  

This is currently blocked by https://github.com/SUSE/spacewalk/issues/15111  

## GUI diff

Different time zone name format, was `CEST`, is `Europe/Berlin`.

- [x] **DONE**

## Documentation
- No documentation needed: Only textual difference, doesn't affect any examples afaik.

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15112 and https://github.com/SUSE/spacewalk/issues/15158

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


cbbayburt (approved):
LGTM :+1: very neat.

I believe most of the TODOs can already be addressed so please remember them before merging.

Just one nitpick: some of the files are missing the newline at the end. I think this is the default behavior with vscode, but you can change it with the "Insert Final Newline" setting.

susemanager-frontend/yarn.lock
```diff
@@ -11594,6 +11594,18 @@ mkdirp@^0.5.3, mkdirp@^0.5.5, mkdirp@~0.5.1:
...
     minimist "^1.2.5"
 
+moment-timezone@^0.5.33:
+  version "0.5.33"
+  resolved "https://registry.yarnpkg.com/moment-timezone/-/moment-timezone-0.5.33.tgz#b252fd6bb57f341c9b59a5ab61a8e51a73bbd22c"
+  integrity sha512-PTc2vcT8K9J5/9rDEPe5czSIKgLoGsH8UNpA4qZTVw0Vd/Uz19geE9abbIOQKaAQFcnQ3v5YEXrbSc5BpshH+w==
+  dependencies:
+    moment ">= 2.9.0"
+
+"moment@>= 2.9.0", moment@^2.29.1:
```
parlt91, line 11604: Since we're adding this here we should remove the other source of moment that we were using before (in case this is not already on your plate).

parlt91 (approved):
It looks good to me. There is only one thing it noticed: In the details display of a recurring action we display the frequency of the recurrence in human readable form. This one is still displayed in server time. It's the `getExecutionText()` method in `recurring-states-details.tsx`, but this can be addressed separately.

web/html/src/utils/functions.ts
```diff
@@ -33,33 +33,6 @@ function cancelable<T = any>(promise: Promise<T>, onCancel?: (arg0: Error | void
...
-      : parseInt(originalDateString.substring(originalDateString.length - 6).replace(":", ""), 10);
-  const serverOffset = Math.trunc(offsetNum / 100) * 60 + (offsetNum % 100);
-  const dateToTransform = new Date(dateStringToTransform);
-  const clientOffset = -dateToTransform.getTimezoneOffset();
-
-  const final = new Date(dateToTransform.getTime() - (serverOffset - clientOffset) * 60000);
-  return final;
-}
-
 function LocalDateTime(date: Date): string {
```
ncounter, line 36: Partially out of scope of this PR: does this function and especially its name still makes sense or we can get rid of it as well or substitute it with something from the library?

ncounter (approved):
I just left a comment which is in and out of the scope, just for tackling the topic widely.
After this PR is merge and we fix the entire thing, we could also think about removing [this hack](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/utils/functions.ts#L109) if possible.

Thanks a lot for the PR :+1:

susemanager-frontend/yarn.lock
Etheryte: Will be addressed in https://github.com/SUSE/spacewalk/issues/13145 which I can hopefully do once 4.2 is unlocked. 👍

web/html/src/utils/functions.ts
Etheryte: Created https://github.com/SUSE/spacewalk/issues/15389

PR #3681, opened 2021-05-03
Basic support for Rocky Linux 8

digdilem:
## What does this PR change?

Add support for Rocky Linux as a supported distro.

Packages building for tests on top of Uyuni Master: https://build.opensuse.org/project/show/home:juliogonzalezgil:branches:systemsmanagement:Uyuni:Master

Java code: https://github.com/uyuni-project/uyuni/pull/3959

# Tests 

## WebUI, regular minion

- [x] Bootstrap
- [x] Hardware refresh
- [x] Enabling a formula (locale) and adjusting data
- [x] Applying the highstate with the formula
- [x] Detect pending updates
- [x] Install a package
- [x] Remove a package
- [ ] Errata detection: Not available -> https://forums.rockylinux.org/t/will-there-be-errata/1806
- [x] Run a remote command

## WebUI, SSH minion

- [x] Bootstrap
- [x] Hardware refresh
- [x] Enabling a formula (locale) and adjusting data
- [x] Applying the highstate with the formula
- [x] Detect pending updates
- [ ] Install a package: Fails, known issue (https://github.com/SUSE/spacewalk/issues/14400). The GPG key for Rocky must be trusted before the onboarding.
- [x] Remove a package
- [ ] Errata detection: Not available -> https://forums.rockylinux.org/t/will-there-be-errata/1806
- [x] Run a remote command

## Bootstrap script, regular minion

- [x] Bootstrap
- [x] Hardware refresh
- [x] Enabling a formula (locale) and adjusting data
- [x] Applying the highstate with the formula
- [x] Detect pending updates
- [x] Install a package
- [x] Remove a package
- [ ] Errata detection: Not available -> https://forums.rockylinux.org/t/will-there-be-errata/1806
- [x] Run a remote command

## GUI diff

No difference.

## Documentation
- Pending (https://github.com/SUSE/spacewalk/issues/14220)

PR #3959, opened 2021-07-07
Rocky Linux 8 Java enablement

juliogonzalez:
## What does this PR change?

Java enablement for Rocky Linux 8.

## Works
- package provider detection:
image

 ## Does not work
- product detection:
image

**NOTE:** This will only work when we enable Rocky Linux as a SUSE Product it seems. Honestly I think this should work even for things that are not defined as SUSE Products (but that's not to be fixed at this PR)

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Issue: https://github.com/SUSE/spacewalk/issues/15322

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Related:
- https://github.com/uyuni-project/uyuni/pull/3681
- https://github.com/uyuni-project/uyuni/pull/3959

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java
+++ b/java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java
@@ -1337,7 +1337,7 @@ public class ChannelFactory extends HibernateFactory {
     public static ContentSource findVendorContentSourceByRepo(String repoUrl) {
         Criteria criteria = getSession().createCriteria(ContentSource.class);
         criteria.add(Restrictions.isNull("org"));
-        if (repoUrl.contains("mirrorlist.centos.org")) {
+        if (repoUrl.contains("mirrorlist.centos.org") || repoUrl.contains("mirrors.rockylinux.org")) {
             criteria.add(Restrictions.eq("sourceUrl", repoUrl));
         }
         else {
--- a/java/code/src/com/redhat/rhn/domain/server/MinionServer.java
+++ b/java/code/src/com/redhat/rhn/domain/server/MinionServer.java
@@ -152,7 +152,7 @@ public class MinionServer extends Server implements SaltConfigurable {
     @Override
     public boolean doesOsSupportsMonitoring() {
         return isSLES12() || isSLES15() || isLeap15() || isUbuntu1804() || isUbuntu2004() || isRedHat6() ||
-                isRedHat7() || isRedHat8() || isAlibaba2() || isAmazon2();
+                isRedHat7() || isRedHat8() || isAlibaba2() || isAmazon2() || isRocky8();
     }
 
     /**
@@ -216,6 +216,10 @@ public class MinionServer extends Server implements SaltConfigurable {
 
     private boolean isAmazon2() {
         return ServerConstants.AMAZON.equals(getOsFamily()) && getRelease().equals("2");
+    }
+
+    private boolean isRocky8() {
+        return ServerConstants.ROCKY.equals(getOs()) && getRelease().startsWith("8.");
     }
 
     /**
--- a/java/code/src/com/suse/manager/reactor/utils/RhelUtils.java
+++ b/java/code/src/com/suse/manager/reactor/utils/RhelUtils.java
@@ -98,7 +98,7 @@ public class RhelUtils {
     }
 
     /**
-     * The content of the /etc/redhat|centos|oracle|almalinux|alinux|system-release file.
+     * The content of the /etc/redhat|centos|oracle|almalinux|alinux|system|rocky-release file.
      */
     public static class ReleaseFile {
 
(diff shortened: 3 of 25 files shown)


java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java
```diff
@@ -1337,7 +1337,7 @@ public static void addErrataToChannel(Set<Long> eids, Long cid) {
     public static ContentSource findVendorContentSourceByRepo(String repoUrl) {
         Criteria criteria = getSession().createCriteria(ContentSource.class);
         criteria.add(Restrictions.isNull("org"));
-        if (repoUrl.contains("mirrorlist.centos.org")) {
+        if (repoUrl.contains("mirrorlist.centos.org") || repoUrl.contains("mirrors.rockylinux.org")) {
```
juliogonzalez, line 1340: My understanding his that this is needed if the URLs are mirrorlists and not repositories. This is the case for RockyLinux.

java/code/src/com/suse/manager/reactor/utils/test/RhelUtilsTest.java
```diff
@@ -288,6 +303,17 @@ public void testDetectRhelProductAmazon() throws Exception {
                 });
     }
 
+    public void testDetectRhelProductRocky() throws Exception {
+        doTestDetectRhelProduct("dummy_packages_redhatprodinfo_rockylinux.json",
+                null,
+                prod -> {
+                    assertFalse(prod.get().getSuseProduct().isPresent());
```
juliogonzalez, line 310: I think this should change to:
```
                    assertTrue("SUSE Product not found", prod.get().getSuseProduct().isPresent());
```
As soon as we have SCC products and the definitions as we did for Alma at https://github.com/uyuni-project/uyuni/commit/915d51c0a5e8ab434ace2310714b71ccb7d256df

java/code/src/com/redhat/rhn/domain/server/MinionServer.java
```diff
@@ -218,6 +218,10 @@ private boolean isAmazon2() {
         return ServerConstants.AMAZON.equals(getOsFamily()) && getRelease().equals("2");
     }
 
+    private boolean isRocky8() {
+        return ServerConstants.ROCKY.equals(getOs()) && getRelease().equals("8");
```
juliogonzalez, line 222: I think using `getOsFamily()` here was what caused the problem with the product detection. It's `RedHat` for Rocky Linux.

juliogonzalez: Second problem: `getRelease().equals("8");` does not work, as it seems the release is `8.4' right now. Changed to match `8.*`

java/code/src/com/suse/manager/reactor/utils/RhelUtils.java
```diff
@@ -280,6 +282,11 @@ public String getRelease() {
             return amazonReleaseFile.map(v -> detectPlainRHEL(v, arch, "AmazonLinux"));
         }
 
+        // next check if Rocky Linux
+        if (rockyReleaseFile.filter(StringUtils::isNotBlank).isPresent()) {
+            return rockyReleaseFile.map(v -> detectPlainRHEL(v, arch, "RockyLinux"));
```
cbbayburt, line 287: Following the code path, the `RockyLinux` value here has no effect at all in the normal case, (it's overridden by the name in the release file) but for consistency, it could be changed to `Rocky`.

The values in the SUSE product for product matching lookup shall be:
Name: `Rocky`
Version: `8` or null
Release: `Green Obsidian` or null
Arch: `x86_64` or null

java/code/src/com/suse/manager/reactor/utils/test/RhelUtilsTest.java
cbbayburt: :+1:

cbbayburt (approved):
LGTM :+1: Just added a tiny suggestion below.

java/code/src/com/suse/manager/reactor/utils/RhelUtils.java
juliogonzalez: Does it mean we have it wrong for `AmazonLinux` as well?

cbbayburt: probably, but like I said, as long as the release file is there, it'll be overridden so it's not a big deal.

cbbayburt: That parameter means "look up this name in SUSE products if I cannot parse the release file for some reason".

schema/spacewalk/upgrade/susemanager-schema-4.3.0-to-susemanager-schema-4.3.1/100-rocky.sql
```diff
@@ -0,0 +1,9 @@
+insert into rhnPackageProvider (id, name)
+    (select sequence_nextval('rhn_package_provider_id_seq'), 'Rocky Linux' from dual
+    where not exists (select 1 from rhnPackageProvider where name = 'Rocky Linux'));
+
+-- Alma Linux 8
```
vzhestkov, line 5: Just  nitpick: comment points Alma instead of Rocky

PR #3969, opened 2021-07-08
Add findutils to Kiwi bootstrap packages

nadvornik:
## What does this PR change?

rhn-org-trusted-ssl-cert-osimage requires findutils for correct function of post-install script.
This pull request adds the package explicitly, in case it is missing from bootstrap section of Kiwi config.xml.

Adding it to sls file allows straightforward update, without need to regenerate any package.

## GUI diff

No difference.

## Documentation
- No documentation needed: no user visible change, it just makes the image building more robust

## Test coverage
- No tests: already covered by cucumber tests

## Links

Part of implementation of https://github.com/SUSE/spacewalk/issues/14570

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/images/kiwi-image-build.sls
+++ b/susemanager-utils/susemanager-sls/salt/images/kiwi-image-build.sls
@@ -43,9 +43,13 @@ mgr_buildimage_prepare_activation_key_in_source:
 {%- set kiwi = 'kiwi-ng' %}
 
 {%- set kiwi_options = pillar.get('kiwi_options', '') %}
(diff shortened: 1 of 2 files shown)


aaannz (approved):
I understand need for this change because of migrations, right?
The package itself (ca-certificates dependency) should be fixed because of https://bugzilla.suse.com/show_bug.cgi?id=1183680

nadvornik:
I tested it with SLE Micro 5.0 which is based on SP2 and it is not fixed there.
So I think we need it anyway.

PR #3974, opened 2021-07-12
Use correct API endpoint in spacecmd list_proxies

agraul:
## What does this PR change?

The "Satellite" XMLRPC handler was removed, but spacecmd was still using it. The correct endpoint is "proxy.listProxies".

## GUI diff

No difference.

## Documentation
- No documentation needed: user invisible change

## Test coverage

There is an existing [unit test](https://github.com/uyuni-project/uyuni/blob/master/spacecmd/tests/test_misc.py#L46-L57) for this function, but I think it is rather pointless. I would prefer to remove it, but first I want to hear your opinion.

- [x] **DONE**

## Links

Related to https://github.com/SUSE/spacewalk/issues/15313
Fixes bsc#1188042

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/spacecmd.changes
+++ b/spacecmd/spacecmd.changes
@@ -1,3 +1,4 @@
+- Use correct API endpoint in list_proxies (bsc#1188042)
 - Add schedule_deletearchived to bulk delete archived actions (bsc#1181223)
 - make spacecmd aware of retracted patches/packages
 - Bump version to 4.3.0
--- a/spacecmd/src/spacecmd/misc.py
+++ b/spacecmd/src/spacecmd/misc.py
@@ -174,7 +174,7 @@ def help_list_proxies(self):
 
 
 def do_list_proxies(self, args):
-    proxies = self.client.satellite.listProxies(self.session)
+    proxies = self.client.proxy.listProxies(self.session)
     print(proxies)
     return 0
 
--- a/spacecmd/tests/test_misc.py
+++ b/spacecmd/tests/test_misc.py
@@ -51,10 +51,14 @@ class TestSCMisc:
         :return:
         """
         mprint = MagicMock()
-        with patch("spacecmd.misc.print", mprint) as prt:
+        shell.client.proxy.listProxies.return_value = "Print me"
+
+        with patch("spacecmd.misc.print", mprint):
             spacecmd.misc.do_list_proxies(shell, "")
+
         assert mprint.called
-        assert shell.client.satellite.listProxies.called
+        assert_expect(mprint.call_args_list, "Print me")
+        assert shell.client.proxy.listProxies.called
 
     def test_get_session(self, shell):
         """


vzhestkov (approved):
:+1:

agraul:
@vzhestkov what do you think should we do about the unit test? Personally, I think we can remove it.

vzhestkov:
In case if it checks just a call for exact function it looks useless for such a case, but it's a unit test in case of API it just ensures that the correct function is called, probably it's better to fix to the right one, but exact testing should be performed in Cucumber.

agraul:
It's just that a unit only duplicates the code in this case. There is no logic that we can test, all we check is that we copy the line with the API call from the code to the test module. Chances are high that either the test always passes (copy&paste successful), or the test fails because the endpoint changed (copy&paste failure), but not because we change the logic.

 I think it would make more sense to have an integration test for this, then we cat realize that something is wrong after the API changed.

meaksh (approved):

agraul:
@meaksh what do you think about the failing test?

agraul:
I see two approaches we can chose to get this into merge-ready:

1) "Fix" the unit test by copy & paste the API call from the code to the test. This will make the testsuite green, can be done quickly, but it does not tell us a lot. The unit test only does two things: compare the api call in the code to a (hardcoded) value and check that `print()` is called. Personally, I think this is a pretty useless test.
2) Remove the unit test and add an integration test. The integration test could be part of the cucumber testsuite where `spacecmd list_proxies` is called and its output is checked. The downside is that this would take more time to implement, I asked the QA people in SUSE's RC about it and they like the idea, but we need to coordinate with them about it.

There is also the option 3) Have both a unit and integration test, but I really don't see the point.

What do you think @vzhestkov @meaksh @SchoolGuy ?

vzhestkov:
Basically unit test here should just check if correct endpoint called, and it should fail on deprecating the endpoint, but the main functionality should be tested with cucumber (more likely option 3 is better in this case).

agraul:
> Basically unit test here should just check if correct endpoint called, and it should fail on deprecating the endpoint, but the main functionality should be tested with cucumber (more likely option 3 is better in this case).

Why do you feel this way? What's the purpose of the unit test in this case?

vzhestkov:
Only to test if it calls the correct endpoint.

SchoolGuy:
I agree with @vzhestkov and would love to see that there is no change in behavior of the endpoint. The integration test is - as you said - time consuming and thus I would go for the easy solution for now. If QA desires to add this to the testsuite we can still remove the "quickly" implemented test at a later point. Better to have one test, then no test in my eyes.

agraul:
> Better to have one test, then no test in my eyes.

I think that having a test is not something positive in of itself. Having a test has a cost attached to it, we need to maintain it. So, in my eyes, a test needs to serve a particular purpose. When we have a bug, I'm always in favor of adding a test that checks that this bug is fixed in the code. That's a pretty good safeguard that stops us from re-introducing a bug when we refactor code. Another example is testing logic or manipulating data. A test can bring clarity for how a function is supposed to be used (examples), and show if the edge-cases are taken into account correctly.

I just think that it is not a net-win to have this test:
```python
        mprint = MagicMock()
        with patch("spacecmd.misc.print", mprint) as prt:
            spacecmd.misc.do_list_proxies(shell, "")
        assert mprint.called
        assert shell.client.proxy.listProxies.called
```
for this function:
```python
    proxies = self.client.proxy.listProxies(self.session)
    print(proxies)
    return 0
```
All we do is replicating the code 1:1 in a test framework. We don't check any logic. The `shell` in the test is a Mock, we don't know if the API call we use in the code and the test is correct. The test would pass even if the endpoint is wrong, as long is we copy&paste correctly between the two files. This is not a useful test in my opinion.

I know that discussing this single test now takes a lot more time than just changing the unit test, but I think this is a good opportunity to discuss our test strategy in general (what do we test/what kind of test do we chose/when do we decide not to test something etc.).

agraul:
> Only to test if it calls the correct endpoint.

But that's not what this test is doing. We had this test and the code passed the test. That didn't help us, we still have a broken `spacecmd` because the test was as outdated as the code.

vzhestkov:
In case if it's not possible to check if it tries to reach correct endpoint in API it becomes totally useless.

meaksh:
IMHO this current unit test is not doing much as you already mentioned, but this is also because the function to test is really too simple. The test only ensures that the funcions calls a defined endpoint and do a print. Nothing else. But for me this is fine.

For a better unit test, maybe we should make the endpoint mock to return some value and then check whether this value is passed properly to the print or not, which is what this function is all about. In the context of a unit test, not sure what else we could check here.

I do understand that the value of the current test is almost nothing in this case. But it brings a small value already (it would fail if the `print` function is not called, or if the called endpoint changed). That's ok I would say, and we do not really sacrify lot of resources for this AFAIK.

Do you think these tests are really causing troubles atm or eating too many resources that we should consider dropping them? If this is really problem then I would consider it. 

What we may want to consider is to also cover this via integration tests, to really check the consistency between spacecmd and the actual Java endpoints on the XMLRPC api, which makes lot of sense. But this would be for cucumber.

agraul:
> IMHO this current unit test is not doing much as you already mentioned, but this is also because the function to test is really too simple. The test only ensures that the funcions calls a defined endpoint and do a print. Nothing else. But for me this is fine.
>
> For a better unit test, maybe we should make the endpoint mock to return some value and then check whether this value is passed properly to the print or not, which is what this function is all about. In the context of a unit test, not sure what else we could check here.
>
> I do understand that the value of the current test is almost nothing in this case. But it brings a small value already (it would fail if the `print` function is not called, or if the called endpoint changed). That's ok I would say, and we do not really sacrify lot of resources for this AFAIK.
>

I think the small value is not enough to warrant a unit test for this function. An integration test would cover the same thing, and it would ensure that the endpoint is real. I'm not saying we should drop the test and not replace it by something else, just that the unit test is not adding much on top of the (to-be-written) integration test. One point in favor of the unit test would be a shorter feedback loop, but I think it's not really worth having to update the endpoints. This is not restricted to this one endpoint, I would prefer to have unit tests only for functions that include logic and use integration tests to verify endpoints.

> Do you think these tests are really causing troubles atm or eating too many resources that we should consider dropping them? If this is really problem then I would consider it.
>

Troubles is too much, the effort to maintain them is not high. I think it is better to drop them in favor of integration tests, that's all.

> What we may want to consider is to also cover this via integration tests, to really check the consistency between spacecmd and the actual Java endpoints on the XMLRPC api, which makes lot of sense. But this would be for cucumber.

It is done in cucumber, but that is not a problem in my eyes. We need to coordinate with QA, the hardest question is probably where to put it, but that should not stop us from having tests that will warn us when spacecmd breaks like this.

---

But given all three of you would prefer to keep it, I'll fix the unit test and we keep it.

vzhestkov (approved):

meaksh (approved):

meaksh:
@agraul I agree with you that the real value in this case would be to have proper integration tests for "spacecmd". 

Currently the cucumber testsuite is already doing some testing around "spacecmd" in some scenarios, but of course only few commands are in the end tested. I'm totally in favor of creating proper integration tests in cucumber.

Maybe introducing a new "spacecmd" feature file in cucumber to extensive test different commands would be nice to have.

agraul:
> Maybe introducing a new "spacecmd" feature file in cucumber to extensive test different commands would be nice to have.

That's one way, but then the "spacecmd" feature must be run at the correct time. A test for `list_proxies` would benefit from having proxies to list.

PR #3985, opened 2021-07-13
Stop using client name aliases

Bischoff:
## What does this PR change?

In the test suite, we introduced those aliases to be very precise about the role of each client at any time.

While this was nice, it made the code more complicated, and sometimes hid which client we were really using.

## Links

Fixes SUSE/spacewalk#15152

Ports:
* 4.1: SUSE/spacewalk#15392
* 4.2: SUSE/spacewalk#15391

## Changelogs

- [x] No changelog needed


calancha (approved):
I like this change.

Bischoff:
> I like this change.

... and I think I know why :smile_cat: .

PR #3987, opened 2021-07-14
CLM cleanup targets on project removal

hustodemon:
## What does this PR change?

This was a long-time missing feature. It allows to delete the project targets (currently only SW channels) on project removal. This is optional and needs to be confirmed (checkbox in the UI, parameterer in the API). The default behavior is unchanged (targets are not cleaned).

## TODO
- [ ] get some UI help, so that the dialog does not looks so ugly
- [ ] Implement the detection of systems possible to affected channels and reflect this in the UI accordingly
  - [ ] for project removal screen
  - [ ] for environment removal screen
- [x] make sure it behaves correctly on projects with many channels and many assignments
- [x] unify the names (clean vs. remove)

## GUI diff

After:
ss

- [x] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
@@ -179,7 +179,7 @@ public class ContentManagementHandler extends BaseHandler {
     }
 
     /**
-     * Remove Content Project
+     * Remove Content Project. Do not purge the underlying environment targets (e.g. sw channels).
      *
      * @param loggedInUser the logged in user
      * @param label the label
@@ -194,7 +194,33 @@ public class ContentManagementHandler extends BaseHandler {
     public int removeProject(User loggedInUser, String label) {
         ensureOrgAdmin(loggedInUser);
         try {
-            return contentManager.removeProject(label, loggedInUser);
+            return contentManager.removeProject(label, false, loggedInUser);
+        }
+        catch (EntityNotExistsException e) {
+            throw new EntityNotExistsFaultException(e);
+        }
+    }
+
+    /**
+     * Remove Content Project
+     *
+     * @param loggedInUser the logged in user
+     * @param label the label
+     * @param cleanTargets true if the underlying targets (e.g. sw channels) should be removed
+     * @throws EntityNotExistsFaultException when Project does not exist
+     * @return the number of removed objects
+     *
+     * @xmlrpc.doc Remove Content Project
+     * @xmlrpc.param #session_key()
+     * @xmlrpc.param #param_desc("string", "label", "Content Project label")
+     * @xmlrpc.param #param_desc("boolean", "cleanTargets", "Whether the underlying targets (e.g. sw channels)
+     * should be removed too")
+     * @xmlrpc.returntype #return_int_success()
+     */
+    public int removeProject(User loggedInUser, boolean cleanTargets, String label) {
+        ensureOrgAdmin(loggedInUser);
+        try {
+            return contentManager.removeProject(label, cleanTargets, loggedInUser);
         }
         catch (EntityNotExistsException e) {
             throw new EntityNotExistsFaultException(e);
--- a/java/code/src/com/redhat/rhn/manager/contentmgmt/ContentManager.java
+++ b/java/code/src/com/redhat/rhn/manager/contentmgmt/ContentManager.java
@@ -195,16 +195,21 @@ public class ContentManager {
      * Remove Content Project
      *
      * @param label the label
+     * @param cleanTargets true if the linked environment targets (e.g. sw channels) should be removed
      * @param user the user
      * @throws PermissionException if given user does not have required role
      * @throws EntityNotExistsException if Content Project with given label is not found
      * @return the number of objects affected
      */
-    public int removeProject(String label, User user) {
+    public int removeProject(String label, boolean cleanTargets, User user) {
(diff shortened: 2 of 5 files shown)


java/code/src/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
```diff
@@ -194,7 +194,33 @@ public ContentProject updateProject(User loggedInUser, String label, Map<String,
...
+     * @return the number of removed objects
+     *
+     * @xmlrpc.doc Remove Content Project
+     * @xmlrpc.param #session_key()
+     * @xmlrpc.param #param_desc("string", "label", "Content Project label")
+     * @xmlrpc.param #param_desc("boolean", "cleanTargets", "Whether the underlying targets (e.g. sw channels)
+     * should be removed too")
+     * @xmlrpc.returntype #return_int_success()
+     */
+    public int removeProject(User loggedInUser, boolean cleanTargets, String label) {
```
cbbayburt, line 220: The parameter order is different than the XMLRPC docs. Does it matter? Also, by convention, the optional parameter could go to the end of the parameter order.

cbbayburt (approved):
LGTM :+1: 
Just one question below.

Adding @mcalmer as a reviewer to ask if he can think of any unwanted consequences.

java/code/src/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
hustodemon: Good catch! It matters. I'll change it. Thanks!

java/code/src/com/redhat/rhn/manager/contentmgmt/ContentManager.java
```diff
@@ -195,16 +195,21 @@ public ContentProject updateProject(String label, Optional<String> newName, Opti
...
-    public int removeProject(String label, User user) {
+    public int removeProject(String label, boolean cleanTargets, User user) {
         ensureOrgAdmin(user);
-        return lookupProject(label, user)
-                .map(cp -> ContentProjectFactory.remove(cp))
-                .orElseThrow(() -> new EntityNotExistsException(label));
+        ContentProject project = lookupProject(label, user).orElseThrow(() -> new EntityNotExistsException(label));
+
+        if (cleanTargets) {
+            listProjectEnvironments(label, user).forEach(e -> removeEnvironment(e.getLabel(), label, user));
```
mcalmer, line 209: Hmm, but `removeEnvironment()` seems to remove only the environment and not the software channels which are part of the env.
At least I cannot find code doing it.

mcalmer (changes requested):

java/code/src/com/redhat/rhn/manager/contentmgmt/ContentManager.java
mcalmer: Additionally what should happen with clients using the channels which should be removed? 
We have other code which handle it. In that case a question come up to unsubscribe all clients from the channel which should be removed.

ncounter:
Orthogonal question: is the removal of channels allowed and wanted in case of systems subscribed to them? Should we at least warn the user about how many systems are subscribed to (or that there are any) within the context of asking this PR implementation checkbox option?

java/code/src/com/redhat/rhn/manager/contentmgmt/ContentManager.java
cbbayburt: > In that case a question come up to unsubscribe all clients from the channel which should be removed.

If we do this, it should be mentioned in the popup dialog.

ncounter: I think my question is already under discussion here so ---> https://github.com/uyuni-project/uyuni/pull/3987#issuecomment-880660952

I completely agree we should at least tell the user there are clients subscribed to channels (even better to have a link to point to a selection of them and list them in a proper view). Telling how many is an option but should not be a big deal once we know the info. For sure we should enhance the current option to be checked and warn the user about it within the same moment and context.

mcalmer: I would suggest:

Check if there are clients using the channels. Depending on the result I would add one additional checkbox:
- If no clients are using the channels, provide the option to remove the channels as well
- If there are clients, tell the number of clients with the option to unsubscribe the clients and remove channels. The number can be a link pointing to a system list page with all the clients which are affected. E.g. ` There are 20 Clients still using the Channels (Targets or Environments?). Unsubscribe all clients from this project before removing it?`

I would not support the option "unsubscribe but keep the channels". I think this would be too much.

hustodemon: Ok, the discussion got pretty hot here. Good :smile: 

@mcalmer removing an environment already removes channels as of today (it's done in the `ContentProjectFactory.purgeTarget`). I like your suggestion and I'll improve it that way. I'll also improve the screen for removing environments.

Thanks for the discussion!

hustodemon:
@ncounter Thanks for your point. I agree and I tried to sum it up in my comment above ^.

hustodemon:
Closing for now. No time/mood to work on that. Will re-open once this gets prioritized.

Issue #3067, opened 2021-01-04
Modify smdba to work on all Postgresql versions. [RHEL8]


sbluhm:
# Additional Information
The current SMDBA implementation requires the SUSE Postgresql package installation and will fail with the error message on other installations or OS.

```
# /usr/bin/smdba
General error:
	Unknown database backend. Please check /etc/rhn/rhn.conf configuration file.
```

## Version of SMDBA
smdba-1.7.7.tar.bz2 (2021-01-04)

## Details about issue
```
<module 'smdba.postgresqlgate' from '/usr/lib/python3.6/site-packages/smdba/postgresqlgate.py'>
General error:
[Errno 2] No such file or directory: '/etc/sysconfig/postgresql'
(L231)
```
This seems to boil down to `/etc/sysconfig/postgresql` not being found on all systems (i.e. RHEL).
See line 231 from `/usr/lib/python3.6/site-packages/smdba/postgresqlgate.py` (lines 227 to 233):
```
    def _get_sysconfig(self) -> None:
        """
        Read the system config for the postgresql.
        """
        for line in filter(None, map(lambda line: line.strip(), open('/etc/sysconfig/postgresql').readlines())):
            if line.startswith('#'):
                continue

```

For example RHEL has the options from that file directly in  `/etc/systemd/system/multi-user.target.wants/postgresql.service`. So that file should also be processed. (but has a different format.)
(See: [https://superuser.com/questions/1424293/where-to-set-pgopts-variable-in-centos-7](https://superuser.com/questions/1424293/where-to-set-pgopts-variable-in-centos-7))

SMDBA lives at:
https://build.opensuse.org/package/show/systemsmanagement:Uyuni:Master/smdba
https://github.com/SUSE/smdba

mcalmer:
CC @jordimassaguerpla

sbluhm:
Same goes for susemanager.spec:
https://github.com/uyuni-project/uyuni/blob/c8f35edc9b69e28e2e36151e93b8d4d8af89b4b0/susemanager/susemanager.spec#L270

jordimassaguerpla:
Hi @sbluhm ! Thanks for the report and sorry I have not answered for a while. I can't commit to look into that in the short term, but if you could provide us with a patch and how you tested it, it would certainly speed up things. Otherwise, I'll put it on my backlog and when time/priority permits I will look into it.

sbluhm:
[...]

PR #3994, opened 2021-07-16
susemanager: Modified postgresql configuration handling

sbluhm:
## What does this PR change?

**Moved postgresql config from spec to mgr-setup**:
The involved variable POSTGRES_LANG is only used during database initalisation. So the changes in %posttrans would only be relevant for new databases/upgrades. The Uyuni setup now checks for the right locale.

**Improved non-SUSE postgresql config handling**:

1. Set the environment locale for the postgres user during Uyuni setup.
2. Explicitly initialise the database with the required locale.
3. Load and double check the locale during a postgres upgrade.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

Build tested successfully on
 epel-8-x86_64
 opensuse-leap-15.3-x86_64

Manually tested upgrade from Leap 15.2 to 15.3 and Postgresql 13 using the scripts.
Manually tested installation on Alma8

- [X] **DONE**

## Links
Fixes: #3067 (one part of it)

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager/susemanager.spec
+++ b/susemanager/susemanager.spec
@@ -274,16 +274,6 @@ sed -i 's/su wwwrun www/su apache apache/' /etc/logrotate.d/susemanager-tools
 %endif
 
 %posttrans
-# make sure our database will use correct encoding
-. /etc/sysconfig/postgresql
-if [ -z $POSTGRES_LANG ]; then
-    grep "^POSTGRES_LANG" /etc/sysconfig/postgresql > /dev/null 2>&1
-    if [ $? = 0 ]; then
-        sed -i -e "s/^POSTGRES_LANG.*$/POSTGRES_LANG=\"en_US.UTF-8\"/" /etc/sysconfig/postgresql
-    else
-        echo "POSTGRES_LANG=\"en_US.UTF-8\"" >> /etc/sysconfig/postgresql
-    fi
-fi
 
 %postun
 %if 0%{?suse_version}
--- a/susemanager/bin/mgr-setup
+++ b/susemanager/bin/mgr-setup
@@ -204,13 +204,29 @@ setup_db_postgres() {
         # Create the PostgreSQL data folder, should it not exist.
         if [ ! -f $DATADIR/PG_VERSION ]; then
             rm -Rf ${DATADIR}
+            POSTGRES_LANG=en_US.UTF-8
+
+            # Define LC_CTYPE in postgres user profile. (Used during any DB creation)
+            PGHOME=$(getent passwd postgres | awk -F: '{print $6}')
+            . $PGHOME/.i18n 2>/dev/null
+            if [ -z $LC_CTYPE ]; then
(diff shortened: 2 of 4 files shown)


juliogonzalez:
Looks good, but I want a confirmation from @jordimassaguerpla, just in case.

susemanager/susemanager.spec
```diff
@@ -25,19 +25,21 @@
 %if 0%{?rhel}
 %global apache_user root
 %global apache_group root
-%global tftp_group root
+%global pgconfig /var/lib/pgsql/data/postgresql.conf
```
mcalmer, line 28: does this really exists when there is no database initialized?
For SUSE we use `/etc/sysconfig/postgresql` to specify some values which define how the DB is started (commandline options) and in which directory the DB should store the files (e.g. /var/lib/pgsql/data/).
One of the values is the language and encoding. 

On a fresh installed SUSE system `/var/lib/pgsql/data/postgresql.conf` does not exist. Only after postgresql was started for the first time.
Is this different on RHEL?

sbluhm: hmm, you are right. It will not exist and created by spacewalk-setup. Seems I misunderstood the purpose of `/etc/sysconfig/postgresql`. Let me see if there is something similar on EL.

sbluhm: @mcalmer I just cross checked and couldn't find a similar mechanism on AlmaLinux. Can we remove the whole  %posttrans section and put the logic into spacewalk-setup? This is where we create the DB in the first place.

mcalmer: I think this section was made to "fix" existing DB configurations. I would suggest to simply write a 

```
%if 0%{suse_version}
...
%endif
```
macro around the whole %posttrans section.
We can also think about to remove this completely as I think we changed the setup to initialize always with `en_US.UTF8`.
In the meanwhile all customers should have this.
But we should double check if we really initialize correctly now.

sbluhm: ok, then I will

1. put an if statement and a comment around the posttrans section.
2. Verify the setup script.
And then resubmit.

sbluhm: I have now moved the `%posttrans` section from the spec file to `bin/mgr-setup` and verified the code in `bin/pg-migrate-x-to-y.sh`
Rationale:
The involved variable `POSTGRES_LANG `is only used during database initalisation. So the changes in `%posttrans` would only be relevant for new databases/upgrades. The Uyuni setup now checks for the right locale; the upgrade script did check for the right locale setting already.

mcalmer (approved):
looks good, but Jordi should have a look as well.

jordimassaguerpla (approved):

juliogonzalez:
Adding the label to merge later today, or tomorrow morning, after the translations are refreshed

PR #3997, opened 2021-07-19
WIP: Fix disabling monitoring if exporters could not be extracted

parlt91:
## What does this PR change?

**Note:** We are looking into another way to address the underlying issue, so do not review for now!
Fixes being unable to disable monitoring if the exporters could not be read from the form data file.

## GUI diff

No difference.

- [x] **DONE**

## Test coverage

- No tests: **Bugfix**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/15350

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
+++ b/java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
@@ -911,6 +911,13 @@ public class FormulaFactory {
     @SuppressWarnings("unchecked")
     public static Map<String, Object> disableMonitoring(Map<String, Object> formData) {
         Map<String, Object> exporters = (Map<String, Object>) formData.get("exporters");
+        // If we cannot extract the exporters from the form data the form data file might
+        // come from an older version of the prometheus-exporters formula package and we will
+        // use the example file instead
+        if (exporters == null) {
+            formData = FormulaFactory.getPillarExample(PROMETHEUS_EXPORTERS);
+            exporters = (Map<String, Object>) formData.get("exporters");
+        }
         for (Object exporter : exporters.values()) {
             Map<String, Object> exporterMap = (Map<String, Object>) exporter;
             exporterMap.put("enabled", false);
@@ -943,6 +950,12 @@ public class FormulaFactory {
     @SuppressWarnings("unchecked")
     private static boolean hasMonitoringDataEnabled(Map<String, Object> formData) {
         Map<String, Object> exporters = (Map<String, Object>) formData.get("exporters");
+        // If we cannot extract the exporters from the form data the form data file might
+        // come from an older version of the prometheus-exporters formula package and we will
+        // use the example file instead
+        if (exporters == null) {
+            exporters = (Map<String, Object>) FormulaFactory.getPillarExample(PROMETHEUS_EXPORTERS).get("exporters");
+        }
         return exporters.values().stream()
                 .map(exporter -> (Map<String, Object>) exporter)
                 .anyMatch(exporter -> (boolean) exporter.get("enabled"));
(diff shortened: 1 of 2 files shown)


parlt91:
@cbbayburt Better wait until you review until I looked into the failing tests

parlt91:
@cbbayburt Ready to review now :smiley:

cbbayburt (commented):
Looks like the following method also has the same vulnerability (I'm not sure how likely it is to hit, though)

Maybe we can catch the case and save new defaults here as well.

https://github.com/uyuni-project/uyuni/blob/d1ca8c36542b88968202e6e0d5f042bd9d9d1b31/java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java#L950-L956

parlt91:
> Looks like the following method also has the same vulnerability (I'm not sure how likely it is to hit, though)
> 
> Maybe we can catch the case and save new defaults here as well.
> 
> https://github.com/uyuni-project/uyuni/blob/d1ca8c36542b88968202e6e0d5f042bd9d9d1b31/java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java#L950-L956

Good catch. I looked into it and I think there is actually one case where we might run into the same issue. I updated the PR accordingly.

java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
```diff
@@ -943,6 +950,12 @@ public static boolean hasMonitoringDataEnabled(ServerGroup group) {
     @SuppressWarnings("unchecked")
     private static boolean hasMonitoringDataEnabled(Map<String, Object> formData) {
         Map<String, Object> exporters = (Map<String, Object>) formData.get("exporters");
+        // If we cannot extract the exporters from the form data the form data file might
+        // come from an older version of the prometheus-exporters formula package and we will
+        // use the example file instead
+        if (exporters == null) {
+            exporters = (Map<String, Object>) FormulaFactory.getPillarExample(PROMETHEUS_EXPORTERS).get("exporters");
+        }
```
witekest, line 958: This will change the function logic and return `true` even if exporters has been explicitly disabled in the old `formData`.

java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
```diff
@@ -911,6 +911,13 @@ else if (pillarExampleFileCustom.isFile()) {
     @SuppressWarnings("unchecked")
     public static Map<String, Object> disableMonitoring(Map<String, Object> formData) {
         Map<String, Object> exporters = (Map<String, Object>) formData.get("exporters");
+        // If we cannot extract the exporters from the form data the form data file might
+        // come from an older version of the prometheus-exporters formula package and we will
+        // use the example file instead
+        if (exporters == null) {
+            formData = FormulaFactory.getPillarExample(PROMETHEUS_EXPORTERS);
+            exporters = (Map<String, Object>) formData.get("exporters");
+        }
```
witekest, line 920: This will overwrite the user's configuration with defaults.

witekest (changes requested):
I think this is not the right approach. The new version of _prometheus-exporters-formula_ migrates all existing formula data files to the new schema, so upgrading this package should solve the issue.

If you want to make the code more robust I'd look at the old logic of these methods instead of taking the defaults from the pillar example.

java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
parlt91: Yes this is intended, if the files contents cannot be read by our codebase, we override the file with the defaults. This is to prevent NPE if exporters could not be extracted see: https://github.com/SUSE/spacewalk/issues/15350

parlt91: In this case the customer already had an up-to-date `prometheus-exporters-formula` package. So if there is a migration for these files something must have gone wrong

java/code/src/com/redhat/rhn/domain/formula/FormulaFactory.java
parlt91: You are right. Thinking about it I don't see a way to properly react to the exporters being null in this case. Any suggestions? I dislike using old logic just to get the data in case the file might be outdated.

PR #4007, opened 2021-07-20
spacecmd: add 'schedule_deletearchived' to bulk delete archived actions (bsc#1181223)

meaksh:
## What does this PR change?

This PR introduces a new command, `schedule_deletearchived` for `spacecmd`, to allow the user to bulk remove archived actions. Also fixes some related issues we figured out while testing this:

As you can see in the implementation, the new `schedule_deletearchived` command is making use of the already existing `schedule.deleteActions` endpoint from the XMLRPC api.

While making use of `schedule.deleteActions` endpoint we've realized of some problems regarding to removing actions:

1) Hibernate exception about missing default constructor:
```
ERROR: redstone.xmlrpc.XmlRpcFault: unhandled internal exception: org.hibernate.InstantiationException: No default constructor for entity:  : com.redhat.rhn.domain.product.SUSEProductUpgrade
```

2) Removing certain actions using `ActionFactory.remove` leads into 500 error due cascade deletion: 

Easy to reproduce:
1) Archive a "Subscribe to channel" action
2) Use spacecmd to delete this action like this:

```
spacecmd {SSM:0}> api schedule.deleteActions --args "[ACTION_ID]"
ERROR: <ProtocolError for myserver/rpc/api: 500 500>
```

If you then look at the `rhn_web_ui.log` you will see an error like this:

```
Caused by: org.postgresql.util.PSQLException: ERROR: update or delete on table "rhnchannel" violates foreign key constraint "rhn_sc_cid_fk" on table "rhnserverchannel"
  Detail: Key (id)=(107) is still referenced from table "rhnserverchannel".
```

So, aparently it's trying to do a deletion from the referenced "rhnchannel" table, which in this case is completely wrong.

We do have a custom SQL queries that we use to delete actions, which by the way are the ones used when we delete the archived actions from the UI, so this PR fixes this problem by switching the `schedule.deleteActions` endpoint to not use `HibernateFactory.remove` but instead this custom SQL queries as done by the UI.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/14196

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/product/SUSEProductUpgrade.java
+++ b/java/code/src/com/redhat/rhn/domain/product/SUSEProductUpgrade.java
@@ -31,6 +31,12 @@ public class SUSEProductUpgrade extends BaseDomainHelper implements Serializable
     private SUSEProduct toProduct;
 
     /**
+     * Default constructor
+     */
+    public SUSEProductUpgrade() {
+    }
+
+    /**
      * Constructor taking two {@link SUSEProduct}s.
      * @param from original product
      * @param to target product
--- a/java/code/src/com/redhat/rhn/manager/action/ActionManager.java
+++ b/java/code/src/com/redhat/rhn/manager/action/ActionManager.java
@@ -438,7 +438,7 @@ public class ActionManager extends BaseManager {
         }
         // now, delete them
         for (Action action : actions) {
-            ActionFactory.remove(action);
+            deleteActionsByIdAndType(action.getId(), action.getActionType().getId());
         }
     }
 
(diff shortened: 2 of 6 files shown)


hustodemon (approved):
LGTM, thanks @meaksh ! One question though: the deletion happens for actions 1-by-1 on the server: does it work "reasonably quick" for a big number of actions (say >10K)? Had anyone tested that?

vzhestkov (approved):
Looks good

meaksh:
Thanks for the reviews!

@hustodemon I ran some tests here and removing 11K archived actions took 1 minute (at least for my fake hardware refresh actions). So I think we're good here.

PR #4012, opened 2021-07-21
Testsuite: Catch cobbler test failures earlier

calancha:
## What does this PR change?

Raise an error as soon as we find a problem with this step.

This step fails silently at head.  Instead, we want to report this failure as soon as it happens.

## Links

https://github.com/SUSE/spacewalk/issues/5838
### Ports:
- Manager-4.1:
- 
## Changelogs

- [x] No changelog needed


Bischoff (approved):
LGTM

PR #4019, opened 2021-07-21
Skip 'update-ca-certificates' run if the certs are updated automatically

cbbayburt:
In SLES15SP3, the directory `/etc/pki/trust/anchors` is watched via the `ca-certificates.path` systemd unit, and as soon as its contents changed, the system-wide certificates are updated automatically. In `certs` state, we run `update-ca-certificates` right after deploying certs. These two processes create a race condition that in rare cases leads to an error.

See https://bugzilla.suse.com/1188500

This PR updates the `certs` state to run `update-ca-certificates` only if this systemd unit does not exist or is not running.

Corresponding sumaform fix for the testsuite: https://github.com/uyuni-project/sumaform/pull/926

## Example "certs" state output
**`ca-certificates.path` running (the state is skipped):**
```
----------
          ID: update-ca-certificates
    Function: cmd.run
        Name: /usr/sbin/update-ca-certificates
      Result: True
     Comment: unless condition is true
     Started: 00:23:46.289726
    Duration: 1996.236 ms
     Changes:   
```
**`ca-certificates.path` is dead or does not exist:**
```
          ID: update-ca-certificates
    Function: cmd.run
        Name: /usr/sbin/update-ca-certificates
      Result: True
     Comment: Command "/usr/sbin/update-ca-certificates" run
     Started: 00:24:28.358217
    Duration: 2016.635 ms
     Changes:   
              ----------
              pid:
                  11368
              retcode:
                  0
              stderr:
              stdout:
```

## Documentation
- No documentation needed: internal change

## Test coverage
- No tests: already covered

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15453

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/certs/SLES12.sls
+++ b/susemanager-utils/susemanager-sls/salt/certs/SLES12.sls
@@ -9,3 +9,7 @@ update-ca-certificates:
     - runas: root
     - onchanges:
       - file: /etc/pki/trust/anchors/RHN-ORG-TRUSTED-SSL-CERT
+    - unless:
+      - fun: service.status
+        args:
+          - ca-certificates.path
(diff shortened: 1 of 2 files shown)


mcalmer (approved):
Ok for me to give it a try. If we find out that we cannot trust that ca-certificates.path service, we need a different solution.

rjmateus (approved):
looks good to me. I think this is the more natural solution we could get :+1:

PR #4031, opened 2021-07-23
Proxy: Add second refresh pattern for repodata

SchoolGuy:
## What does this PR change?

This is required to not fetch out of data repodata. Change was suggested by @jmozd.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14875

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/proxy/installer/spacewalk-proxy-installer.changes
+++ b/proxy/installer/spacewalk-proxy-installer.changes
@@ -1,3 +1,5 @@
+- Add new refresh_pattern to the squid.conf to fix a case where the repodata
+  was invalid due to being cached (bsc#1186026)
 - Bump version to 4.3.0
 
 -------------------------------------------------------------------
--- a/proxy/installer/squid.conf
+++ b/proxy/installer/squid.conf
@@ -24,6 +24,7 @@ memory_replacement_policy heap GDSF
 
 # cache repodata only few minutes and then query parent whether it is fresh
 refresh_pattern /XMLRPC/GET-REQ/.*/repodata/.*$ 0 1% 1440 reload-into-ims refresh-ims
+refresh_pattern /ks/.*/repodata/.*$ 0 1% 1440 reload-into-ims refresh-ims
 # salt minions get the repodata via a different URL
 refresh_pattern /rhn/manager/download/.*/repodata/.*$ 0 1% 1440 reload-into-ims refresh-ims
 # bootstrap repos needs to be handled as well


vzhestkov (changes requested):
@SchoolGuy `uyuni/proxy/installer/spacewalk-proxy-installer.changes` should be updated

proxy/installer/spacewalk-proxy-installer.changes
```diff
@@ -1,4 +1,6 @@
 - Bump version to 4.3.0
+- Add new refresh_pattern to the squid.conf to fix a case where the repodata
+  was invalid due to being cached (bsc#1186026)
```
meaksh, line 3:
```suggestion
- Add new refresh_pattern to the squid.conf to fix a case where the repodata
  was invalid due to being cached (bsc#1186026)
- Bump version to 4.3.0
```

proxy/installer/squid.conf
```diff
@@ -24,6 +24,7 @@ memory_replacement_policy heap GDSF
 
 # cache repodata only few minutes and then query parent whether it is fresh
 refresh_pattern /XMLRPC/GET-REQ/.*/repodata/.*$ 0 1% 1440 reload-into-ims refresh-ims
+refresh_pattern /ks/.*/repomd.xml 0 1% 1440 ignore-no-cache reload-into-ims refresh-ims
```
meaksh, line 27: Why not including the entire `repodata` directory here? As we do for other repository paths here.

With your current proposal only `repomd.xml` file will be refreshed but not other metadata files. Not sure if I'm missing something.

I would suggest:

```suggestion
refresh_pattern /ks/.*/repodata/.*$ 0 1% 1440 ignore-no-cache reload-into-ims refresh-ims
```

jmozd: > Why not including the entire `repodata` directory here? As we do for other repository paths here.
> [...]
> I would suggest:
> [...]

I'm fine with your suggestion - my initial approach is usually to be the least "intrusive" and since I noticed the problem only for repomd.xml, I used the most limiting pattern.

SchoolGuy: I am not expert here. I just went with the suggestion from Jens here. Thus I will just apply your suggestion.

meaksh (approved):

vzhestkov (approved):
Looks good now

PR #4032, opened 2021-07-23
Ensure XMLRPC returns "issue_date" in ISO format when listing erratas (bsc#1188260)

meaksh:
## What does this PR change?

We got a report of a traceback when calling `softwarechannel_errata_merge` command of "spacecmd":

```
# spacecmd softwarechannel_errata_merge sle-module-basesystem15-sp3-updates-x86_64 testchannel
ERROR: time data '1/24/19' does not match format '%Y-%m-%d'
```

The logic on "spacecmd" is expecting the "issue_date" for the erratas listed using the XMLRPC are returned in ISO format, but this is not the currently the case. At the moment the "issue_date" format is different depending on the locale configuration of the Uyuni web UI.

This is not what we should expect from the XMLRPC where we probably want a fixed format for the dates.

So, this PR fixes the `ErrataOverviewSerializer` to ensure the "Date" attributes are returned in ISO format and not as the format defined by the locale configuration on the Uyuni server. This way, `spacecmd softwarechannel_errata_merge` is able to run successfully.

Additonally this PR fixes another issue with a wrong format in `Errata.setIssueDate` where "month" was messed with "minute".

Java / XMLRPC API experts, do you think this is the way to do it? Thanks

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/15383

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/ErrataOverviewSerializer.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/ErrataOverviewSerializer.java
@@ -61,8 +61,8 @@ public class ErrataOverviewSerializer extends RhnXmlRpcCustomSerializer {
 
         helper.add("id", errata.getId());
         helper.add("issue_date", errata.getIssueDateIsoFormat());
-        helper.add("date", errata.getUpdateDate());
-        helper.add("update_date", errata.getUpdateDate());
+        helper.add("date", errata.getUpdateDateIsoFormat());
+        helper.add("update_date", errata.getUpdateDateIsoFormat());
         helper.add("advisory_synopsis", errata.getAdvisorySynopsis());
         helper.add("advisory_type", errata.getAdvisoryType());
         helper.add("advisory_status", errata.getAdvisoryStatus().getMetadataValue());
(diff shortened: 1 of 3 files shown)


java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/ErrataOverviewSerializer.java
```diff
@@ -60,9 +60,9 @@ protected void doSerialize(Object value, Writer output, XmlRpcSerializer seriali
         SerializerHelper helper = new SerializerHelper(serializer);
 
         helper.add("id", errata.getId());
-        helper.add("issue_date", errata.getIssueDate());
-        helper.add("date", errata.getUpdateDate());
-        helper.add("update_date", errata.getUpdateDate());
+        helper.add("issue_date", errata.getIssueDateIsoFormat());
+        helper.add("date", errata.getUpdateDateIsoFormat());
+        helper.add("update_date", errata.getUpdateDateIsoFormat());
```
mcalmer, line 65: We better would return real date objects which are known by XMLRPC. But that would break all existing code which use this API.
So I think this is correct for now.

mcalmer (approved):

hustodemon (approved):

PR #4037, opened 2021-07-25
spacecmd: Improved and modified python code to pass pylint

sbluhm:
## What does this PR change?

1. Update all relevant python files to pass pylint checks. Either via minor code improvements or via adding pylint exceptions.
2. (Minor fix to avoid duplicate group definition in spec file)

This change enhances testing during build. Manual spacewalk-python3-pylint runs pass. Also creating a package on AlmaLinux8 via bash. Pylint fails on automated copr build for epel8 and is automatically skipped for Leap as "not supported" (and the package builds without checks). Therefore this feature is not enabled in the spec file yet until the root cause is identified on copr (or build is confirmed working on obs).

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
This change enhances testing during build. Manual spacewalk-python3-pylint runs as well as creating a package on AlmaLinux8 work. Pylint fails on automated build. Therefore this feature is not enabled in the spec file yet.

Build with current changes has been tested to work for:
 epel-8-x86_64 
 opensuse-leap-15.3-x86_64 

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


juliogonzalez:
@sbluhm seems some unit tests are failing:
```
=================================== FAILURES ===================================
______________ TestSCDistribution.test_distribution_update_noargs ______________
  File "/manager/spacecmd/tests/test_distribution.py", line 554, in test_distribution_update_noargs
    assert not shell.do_distribution_create.called
AssertionError: assert not True
 +  where True = <MagicMock id='140714441392816'>.called
 +    where <MagicMock id='140714441392816'> = <MagicMock id='140714442063096'>.do_distribution_create
_____________ TestSCMisc.test_generate_package_cache_duplicate_ids _____________
  File "/manager/spacecmd/tests/test_misc.py", line 1061, in test_generate_package_cache_duplicate_ids
    'Taking "vim-1-2" instead of "gedit-1-2"',), {})])
  File "/manager/spacecmd/tests/helpers.py", line 114, in assert_args_expect
    assert args == _args, "{} is not as expected {}".format(str(args), str(_args))
AssertionError: ('Non-unique package id "%s" is detected. Taking "%s" instead of "%s"', 69, 'vim-1-2', 'gedit-1-2') is not as expected ('Non-unique package id "69" is detected. Taking "vim-1-2" instead of "gedit-1-2"',)
==================== 2 failed, 662 passed in 11.04 seconds =====================
```

sbluhm:
@juliogonzalez, thank you. fixed them.
I wrongly assumed the unit tests are run during package build. Have to figure out how to run them manually...

mcalmer:
@sbluhm to run the tests you need a docker daemon running and then call:
```
   ./susemanager-utils/testing/automation/backend-unittest-pgsql.sh
```
from the root of the git repo.

mcalmer:
Sorry, that script was for backend unit tests. The spacecmd seems to be:

`susemanager-utils/testing/automation/spacecmd-unittest.sh`

sbluhm:
> Sorry, that script was for backend unit tests. The spacecmd seems to be:
> 
> `susemanager-utils/testing/automation/spacecmd-unittest.sh`

It tries to call "registry.mgr.suse.de" which is not accessible. Will get into this topic in a while. Not really an expert in this area... Thank you for the pointer.

mcalmer:
Hmm, ok. Seems this is not yet changed to use the opensuse registry. Another point on our todo list.

spacecmd/src/spacecmd/utils.py
```diff
@@ -633,7 +634,7 @@ def parse_list_str(list_s, sep=","):
     >>> assert parse_list_str("a,,b,c") == ["a", "", "b", "c"]
     >>> assert parse_list_str("a:b:", ":") == ["a", "b", ""]
     """
-    return [p for p in list_s.split(sep)]
+    return [p for p in list_s.split(sep)] # pylint: disable=bad-option-value,unnecessary-comprehension
```
agraul, line 637: Instead of disabling the pylint warning, this could be as simple as:
```
>>> def parse_list_str(list_s, sep=","):
...     return list_s.split(sep)
... 
>>> assert parse_list_str("") == [""]
>>> assert parse_list_str("a,b") == ["a", "b"]
>>> assert parse_list_str("a,b,") == ["a", "b", ""]
>>> assert parse_list_str("a,,b,c") == ["a", "", "b", "c"]
>>> assert parse_list_str("a:b:", ":") == ["a", "b", ""]

```

spacecmd/src/spacecmd/system.py
```diff
@@ -157,6 +157,8 @@ def manipulate_child_channels(self, args, remove=False):
                                             system_id,
                                             child_channels)
 
+    return
```
agraul, line 160: Shouldn't this be a `return None` like the other explicit `return None` statements that you've added?

sbluhm: Pylint didn't complain about it :)
Reason for the other explicit `return None`s is, that other `return` statements in the **same** function return a value. So the pylint statement is: Either none or all return statements return should a value (within the same function).

spacecmd/src/spacecmd/utils.py
sbluhm: Will modify it. I don't want to risk breaking code if I don't know what I am doing. So in those cases I rather disabled the warning for someone to revisit at a later stage.

agraul: Fair enough, that's a good approach.

spacecmd/src/spacecmd/system.py
agraul: I see, I didn't expand any of the views and I don't (yet) know all the pylint expectations. But if pylint does not complain, then I won't complain either :)

spacecmd/src/spacecmd/utils.py
sbluhm: Updated.

juliogonzalez:
> Hmm, ok. Seems this is not yet changed to use the opensuse registry. Another point on our todo list.

True, but this is becoming a huge mess. We *really* need to reconsider a refactor of the tests, so we don't need to change 100 places each time we branch a new SUSE Manager version, or if we ever consider changing something on the image names/location.

I will handle this tomorrow, with a PR.

juliogonzalez:
> > Sorry, that script was for backend unit tests. The spacecmd seems to be:
> > `susemanager-utils/testing/automation/spacecmd-unittest.sh`
> 
> It tries to call "registry.mgr.suse.de" which is not accessible. Will get into this topic in a while. Not really an expert in this area... Thank you for the pointer.

@sbluhm I just merged https://github.com/uyuni-project/uyuni/pull/4049 to master.

Please rebase the changes from master into your branch, and you will be able to call the unit tests locally.

sbluhm:
> 
> Please rebase the changes from master into your branch, and you will be able to call the unit tests locally.

@juliogonzalez , do I need to rebase? Code is not touching each other and current tests are passing. So I don't think PR #4049  is required for this PR (#4037). (or I misunderstood your request)

agraul (commented):
I think the code changes are fine, but I'm not sure if adding all the comments to locally disable pylint is the correct approach. I would prefer to also use pylint-ignore for spacecmd, which would avoid having to add inline comments everywhere. But we don't have it packaged in a way that can be used at build-time yet, so I don't know what to do for now.

sbluhm:
Not sure how it ties in with side discussions.
My thought is to intentionally disable locally to point out where exactly improvements/changes are required. Global ignores should be made for design decisions.
Also my other thought is: Go two steps in the right direction; instead of none. We can still go back one step later.

agraul:
> My thought is to intentionally disable locally to point out where exactly improvements/changes are required. 

That's exactly the idea behind `pylint-ignore` except that the comment is done in a central file, pointing to the line that will be ignored, and not inside the python module itself. The main upside is that it's less work (does not apply in this case because you already did it!) and a nicer Git history.

> Also my other thought is: Go two steps in the right direction; instead of none. We can still go back one step later.

Sure, we can go ahead with this as you already put the work into it and we can't *just enable* python-pylint for build time checks, but in general I prefer not to touch every file to add those comments.

sbluhm:
Can we actually merge this one? I removed the pylint check itself again so there should be no packaging impact. Only code changes.

sbluhm:
@agraul can you please review my last changes? Thank you.

spacecmd/src/spacecmd/configchannel.py
```diff
@@ -809,12 +814,11 @@ def do_configchannel_addfile(self, args, update_path=''):
...
                     break
-                else:
-                    print('')
-                    logging.warning(_N('%s is not a valid channel') %
-                                    options.channel)
-                    print('')
-                    failures += 1
+                print('')
+                logging.warning(_N('%s is not a valid channel') %
+                                options.channel)
```
agraul, line 819: Isn't this triggering a "logging not lazy" warning?

spacecmd/src/spacecmd/configchannel.py
```diff
@@ -915,24 +919,23 @@ def do_configchannel_updateinitsls(self, args, update_path=''):
...
                 if options.channel in self.do_configchannel_list('', True):
                     break
-                else:
-                    print('')
-                    logging.warning(_N('%s is not a valid channel') %
-                                    options.channel)
-                    print('')
+                print('')
+                logging.warning(_N('%s is not a valid channel') %
+                                options.channel)
```
agraul, line 924: Same as above.

spacecmd/src/spacecmd/ssm.py
```diff
@@ -92,9 +92,8 @@ def do_ssm_add(self, args):
         if system in self.ssm:
             logging.warning(_N('%s is already in the list') % system)
             continue
-        else:
-            self.ssm[system] = self.get_system_id(system)
-            logging.debug('Added %s' % system)
+        self.ssm[system] = self.get_system_id(system)
+        logging.debug('Added %s' % system)
```
agraul, line 96: This string can also be constructed lazily (i.e. only if debug messages are actually logged):
```suggestion
        logging.debug('Added %s', system)
```

spacecmd/src/spacecmd/ssm.py
```diff
@@ -92,9 +92,8 @@ def do_ssm_add(self, args):
...
 
     if self.ssm:
         logging.debug('Systems Selected: %i' % len(self.ssm))
```
agraul, line 99: Same here

spacecmd/src/spacecmd/system.py
```diff
@@ -2755,7 +2769,8 @@ def _crashes_api_removed(check_api_version):
         logging.error(_("This method was removed since API version %s") % crashes_api_removed_version)
         return True
     else:
-        logging.warning(_("This method is deprecated and will be removed in API version %s") % crashes_api_removed_version)
+        logging.warning(_("This method is deprecated and will be removed in API version %s")
+                        % crashes_api_removed_version)
```
agraul, line 2773: This one could also be lazy.

agraul (commented):
I saw a few logging statements that aren't done lazily. Don't they cause warning? I ask because you fixed other eagerly formatted logging strings.

Other than that I'm okay with the code changes, at least for now (eventually the in-line pylint disable comments should be resolved, but that does not need to happen now).

Last thing is about the changelog, @juliogonzalez do you think it makes sense to include those changelog items? I'm not sure if users are really interested in those and we might be able to skip adding them.

juliogonzalez:
> Last thing is about the changelog, @juliogonzalez do you think it makes sense to include those changelog items? I'm not sure if users are really interested in those and we might be able to skip adding them.

Right, this is not useful for users, and we have other entries so we can submit/promote next time. 

@sbluhm the changelog entries can be safely removed, remember to mark the checkbox `No changelog needed` before you push the removal, so the test doesn't fail.

spacecmd/src/spacecmd/ssm.py
sbluhm: @agraul , `logging.debug('Added %s', system)` causes the unit test to fail.

spacecmd/src/spacecmd/configchannel.py
sbluhm: Not that I noticed (which does not mean anything). I have to admit that I am not confident on what I am doing here. So I just took your comments as is.

spacecmd/src/spacecmd/ssm.py
agraul: Seems that we're comparing log messages in code and tests, that's not really nice. I'd say just keep it as it is then, if you don't get pylint complains everything is fine.

spacecmd/src/spacecmd/group.py
```diff
@@ -550,14 +560,12 @@ def do_group_addconfigchannels(self, args):
         return 1
 
     arg_parser = get_argument_parser()
-    (args, options) = parse_command_arguments(args, arg_parser)
+    (args,) = parse_command_arguments(args, arg_parser)
```
agraul, line 563: This can't work, at least not if the previous version worked:
```python
In [1]: def returns_tuple_of_two():
   ...:     return (100, "100")
   ...: 

In [2]: (args,) = returns_tuple_of_two()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-627d73f70b4f> in <module>
----> 1 (args,) = returns_tuple_of_two()

ValueError: too many values to unpack (expected 1)
```

The reason is that `(args, )` is a tuple of length 1:
```python
In [4]: len((args,))
Out[4]: 1
```

I assume the change is done because `options` was never used. In that case, I would use `_`. It's the go-to placeholder variable name for variables that are not accessed. I don't know if `pylint` will still complain, haven't tried it yet.

```python
In [3]: (args,_) = returns_tuple_of_two()
# no error, values are saved in args and _
```

spacecmd/src/spacecmd/group.py
```diff
@@ -589,16 +598,13 @@ def do_group_removeconfigchannels(self, args):
...
```
agraul, line 601: same as above

spacecmd/src/spacecmd/shell.py
```diff
@@ -119,10 +119,9 @@ def __init__(self, options, conf_dir, config_parser):
                                     self.history_file)
                 except IOError:
                     logging.error(_N('Could not read history file'))
-        # pylint: disable=W0702
+        # pylint: disable=W0703
```
agraul, line 122: what is W0703? Please use the symbolic name instead of the id. With just the id, everyone needs to look it up to understand what is being disabled.

agraul (changes requested):
Sorry, it looks like I missed two tuple unpackings (is that even a word? :D) that change behavior earlier. Once those are resolved, and the pylint comment uses the name instead of the id, I'm fine with merging this. Thank you for your patience and resilience @sbluhm

spacecmd/src/spacecmd/group.py
sbluhm: Thank you for giving the explanation and examples!

agraul (approved):
Thank you for the quick changes!

PR #4043, opened 2021-07-26
Wait until file is created by cobbler

Bischoff:
## What does this PR change?

Sometimes the cobbler feature fails testing the presence of `/srv/tftpboot/pxelinux.cfg/default`.

Our hypothesis is that we simply don't wait long enough. This PR addresses this problem.

The feature `srv_distro_cobbler.feature` will soon be replaced with a better one, but we would like to make sure this is not a product issue before we remove it.

## Links

Ports:
* 4.1: SUSE/spacewalk#15525
* 4.2: SUSE/spacewalk#15524

## Changelogs

- [x] No changelog needed


calancha (approved):
LGTM

PR #4039, opened 2021-07-26
spacewalk-X-pylint: Detect pylint executable with minor python version in name

sbluhm:
## What does this PR change?

Currently, pylint is not detected if it does not exist with only the python major version in the executable filename.
Example:

- pylint-2: Detected
- pylint-3: Detected
- pylint-36: Not detected

This change also picks up minor versions or other variations after the major version.

Background: spacewalk-python3-pylint does not run on SUSE Leap.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes

- [X] **DONE**

## Test coverage
- No tests: Manually tested. Low risk, minor change, easily revertible.

Packages spacewalk-pylint, spacewalk-python2-pylint and spacewalk-python3-pylint build tested successfully on:
 epel-7-x86_64 
 epel-8-x86_64 
 fedora-33-x86_64 
 fedora-rawhide-x86_64 
 opensuse-leap-15.3-x86_64 

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #4046, opened 2021-07-26
spacewalk-setup: Different changes

sbluhm:
## What does this PR change?

1. Updated to pass pylint checks.
2. Enabled pylint for all builds but do not fail on error.
3. Fixed an issue on package update on Enterprise Linux 8 (post installation script adds unwanted parameter)
4. Removed perl-DateTime package dependency in favour of perl core Time::Piece

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

Build has been tested successfully with PR #4039 (running pylint) for
 epel-8-x86_64 
 opensuse-leap-15.3-x86_64 

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mcalmer:
Not sure about this. We were discussing this already longer time ago and as pylint is known to produce different results depending on the python interpreter we do not trust it to run with many different one.

I think we wanted to have a container with one dedicated pylint installation (python + pylint) and check all code against it outside of the build process.

On the other hand the pylint code fixes might be required.

@meaksh @agraul - what is your take on this?

sbluhm:
I can remove the "pylint for all builds" commit as I have done earlier for the other packages. Then we just have the code changes.
Either way, the tests will pass in this case.
In future, changes need to be done to the packages to line them up with the proposed pylint-ignore solution so I don't see this change as harmful.

sbluhm:
Included an additional fix https://github.com/uyuni-project/uyuni/pull/4046/files#diff-0ccbf54ec6c63979ce854bc1f3bfff6928637d55049132b6f3e661280ef9b72aR244
com.sun.xml.bind gets added to the tomcat startup. This is unwanted as this is not a standard Java 11 module.

agraul:
> Not sure about this. We were discussing this already longer time ago and as pylint is known to produce different results depending on the python interpreter we do not trust it to run with many different one.
> 
> I think we wanted to have a container with one dedicated pylint installation (python + pylint) and check all code against it outside of the build process.
> 

That is correct. For the beginning we will use `pip` to install a fixed versions of pylint-ignore and pylint in a container at runtime. I already have changed in a Makefile but I need to clean it up/copy it to other Makefiles and add initial pylint-ignore.md files before I can push it. pylint-ignore isn't available at build-time without quite a bit of packaging work that we want to avoid for now.

> On the other hand the pylint code fixes might be required.
> 

With pylint-ignore, adding a `# pylint:` comment won't be needed, but the removal of `or long` makes sense to me.

> @meaksh @agraul - what is your take on this?

I would not put too much more effort into adding comments that make pylint pass. We want to use pylint-ingore which does not need those comments and I'd prefer to have as few of them as possible. 

In general I think it makes more sense to wait with more pylint-related fixes until we have pylint-ingore ready. This one is so small I don't mind merging.

sbluhm:
> In general I think it makes more sense to wait with more pylint-related fixes until we have pylint-ingore ready.

The reason why I was adding these fixes is because Pylint is more or less enabled for the RHEL builds. So this is really just a by-product for LEAP (if at all)

sbluhm:
Updated to not fail spec test if pylint throws error. This allows reviewing pylint output whilst not stopping on different pylint versions/systems.

sbluhm:
@meaksh / @agraul can you please review my last changes?

meaksh (approved):
LGTM

sbluhm:
Can someone please merge? Thank you.

PR #4054, opened 2021-07-28
Add react unit tests to web-calendar

parlt91:
## What does this PR change?

Add react unit test for the maintenance windows web calendar

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **unit tests**

- [x] **DONE**

## Test coverage

- Unit tests were added

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/manager/maintenance/calendar/web-calendar.tsx
```diff
@@ -60,12 +63,14 @@ const WebCalendar = (props: WebCalendarProps) => {
...
     // Hack: Make Fullcalendar button icons show correctly
-    jQuery(".fc-backButton-button").html('<div class="fa fa-angle-left"></div>');
-    jQuery(".fc-nextButton-button").html('<div class="fa fa-angle-right"></div>');
+    jQuery(".fc-backButton-button").attr("aria-label", "back")
+      .html('<div class="fa fa-angle-left"></div>');
+    jQuery(".fc-nextButton-button").attr("aria-label", "next")
+      .html('<div class="fa fa-angle-right"></div>');
     jQuery(".fc-skipBackButton-button").html('<div class="fa fa-angle-double-left"></div>')
-      .prop('title', t("Skip to the last maintenance window"));
+      .attr("aria-label", "skip-back").prop('title', t("Skip to the last maintenance window"));
```
Etheryte, line 71: I'm not very sure about the content of the aria-labels here. Aria-label is what you expect the screen reader to verbally read out to the user and `"skip-back"` is not necessarily the clearest message to the user. I would recommend extracting the translated `t("Skip to the last maintenance window")` into a variable and use it as both the title and the aria-label.

Etheryte (approved):
Aside from one small nit pick, this looks very good. Thanks a lot for adding test coverage to this view. 👍

web/html/src/manager/maintenance/calendar/web-calendar.tsx
parlt91: Yeah I am aware that this one is a bit weird, but I needed a way to identify the button somehow in the tests (since it's not a text button). This way just worked. But I can provide a more clear message here of course

Etheryte: In this specific case, it might make sense to consider a [test id](https://testing-library.com/docs/queries/bytestid/) for it since the icon, the title and the label can change without any affected functionality.

parlt91: I see yeah makes sense. I will make the change. Thanks ;)

PR #4060, opened 2021-07-29
Application Streams tab for modular channels

mbussolotto:
## What does this PR change?

Add a new tab in the channel details page exclusively for modular channels. The tab lists all the AppStreams (module names and stream names) that the channel provides.

## GUI diff

Before:
Screenshot from 2021-07-30 16-49-00
After:
New tab called "AppStreams"
Screenshot from 2021-08-09 17-25-21

- [x] **DONE**

## Documentation
- Documentation issue was created: https://github.com/uyuni-project/uyuni-docs/issues/1094

- [x] **DONE**

## Test coverage

No tests: code with no test

- [x] **DONE**

## Links
Issue: https://github.com/SUSE/spacewalk/issues/10098
EPIC: [AppStream filters in CLM for modular repository support](https://github.com/SUSE/spacewalk/issues/10093)
RFC:[ Modular Repositories with CLM](https://github.com/uyuni-project/uyuni-rfc/pull/22)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mbussolotto:
Changes applied as discussed during the review:
- checkbox are removed
- default stream marked in bold

java/code/src/com/redhat/rhn/domain/contentmgmt/modulemd/ModuleStreams.java
```diff
@@ -45,4 +46,13 @@ public String getDefaultStream() {
...
     }
+
+    @Override
+    public String toString() {
+        streams.sort(Comparator.naturalOrder());
+        if (defaultStream == null) {
+            return String.join(", ", streams);
+        }
+        return String.join(", ", streams).replaceAll(defaultStream, "<b>" + defaultStream + "</b>");
+    }
```
cbbayburt, line 57: With this PR, we're introducing a feature that is specific to the web UI, but this class and `toString` in general has a much wider context, so it'd be better to move this formatting elsewhere. `toString` is expected to be used in many different contexts like APIs, CLI output, logs, stack traces, etc.

I think it'd be much better to do this formatting in the controller + JSP.

For example, if the sorting is required, it should be done in `ApplicationStreamsAction` to prepare it for JSP, and the `<b>` tags should be added inside the JSP file directly. Otherwise, for example, the `<b>` tags would also be searched with filtering. This can be done using some conditionals in JSP using the [core tag set](https://docs.oracle.com/javaee/5/jstl/1.1/docs/tlddocs/c/tld-summary.html).

My suggestion would be to add a `<c:forEach>` loop in the JSP and loop through the `streams` list and print each, and conditionally print with the `<b>` tags if the current item is equal to `defaultStream`. This approach separates the concern of formatting from the rest of the code.

For an even nicer UI, you can print each stream as a [Bootstrap label](https://getbootstrap.com/2.3.2/components.html#labels-badges) (blue for default, gray for others).

java/code/src/com/redhat/rhn/frontend/action/channel/ApplicationStreamsAction.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * ChannelPackagesAction
```
cbbayburt, line 41:
```suggestion
 * ApplicationsStreamsAction
```

cbbayburt (changes requested):
Thanks, @mbussolotto for the PR. It looks good overall.

I commented some changes below, but other than that, there's one more thing:

This is mostly my fault, and because this issue has existed since the dawn of time :see_no_evil: 
The official word for the module streams is actually "AppStreams" by RedHat. So we try to use this exact word in the UI and the documentation to keep the confusion to a minimum.

So let's change every occurrence of the word to "AppStreams", including UI, filenames and code, like:

`"Application Streams"` -> `"AppStreams"`
`ApplicationStreamsAction.java` -> `AppStreamsAction.java`
`applicationstreams.jsp` -> `appstreams.jsp` etc..

cbbayburt (approved):
Looks great! Exactly what I had in mind :+1: 
Thanks, @mbussolotto great work!

moio:
Hey @mbussolotto, first of all thanks for this PR!

I was wondering a bit - what are the pros and cons to implement this new page with the classic Struts framework, instead of adding a React-based page/table?

mbussolotto:
Hi @moio ! Probably @ncounter or @cbbayburt can give you a better answer for your question :) . The reasons why I prefer to do it in struts are:
- the other tabs of the page are in struts, so in this way there's a sort of homogeneity
- I have zero knowledge of react :)

Do you think it should be done in React-based page/table? If yes we will find a way to deal with that :)

ncounter:
@moio while we have the agreement of "whatever is new or it is largely touched should be in the new stack (AKA React), for this specific scenario reasons are:

**Pros**
- it is just a list page with no interaction, no huge benefit of having it on one or the other way
- all the related tabs and pages are alive on the Struts framework, therefore the business logic Model, the Routing, the Controllers and the templates were to be created for the React path instead
- no "free people" among the frontend guys to do this now or any time soon
- @mbussolotto was free instead, but he knows more or less nothing about the frontend and this is fine as we are not looking for him to learn it from scratch as an initial development issue

**Cons**
- yet another list page to be ported if we will ever do, but still a simple/trivial one

In the end the balance is not pushing for on or the other way, so we took the simplest and quickest one anyway. @moio : do you see any other drawback we should not merge this as @mbussolotto implemented it?
Anyway, I am one of the most pushing for "no more JSPs, only new stuff in new tech". I would like to be clear this is just "one special case" we did this the old way though.

moio:
Hmm, I can turn a blind eye for this special case, but let's really try to stick to the practice of introducing new code with the new frameworks or the old ones will stick around for long.

I will recommend to get @mbussolotto started with React at the first useful chance :-)

Q-Storm:
thank you very much for this feature in Uyuni, this is exactly what I am looking for. Do you know in which Version this will be implemented ? I am testing 2021.12 right now and this feature is not implemented yet.

PR #4073, opened 2021-08-02
Fix yum reposync plugin for Fedora 33 and 34 repos

vzhestkov:
## What does this PR change?

Fixes reposync for Fedora 33 and 34 repos.

`spacewalk-repo-sync` produces the traceback on synchronization of Fedora 33 and 34 repos as `zypper` is not downloading plain XML file, but only `xz` file with `group_xz` tag instead of `group`.
```
2021/07/31 22:57:56 +03:00 Unexpected error: <class 'FileNotFoundError'>
2021/07/31 22:57:56 +03:00 Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/reposync.py", line 593, in sync
    self.import_groups(plugin)
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/reposync.py", line 841, in import_groups
    abspath = self.copy_metadata_file(plug, groupsfile, 'comps', relative_comps_dir)
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/reposync.py", line 795, in copy_metadata_file
    src = fileutils.decompress_open(filename)
  File "/usr/lib/python3.6/site-packages/uyuni/common/fileutils.py", line 511, in decompress_open
    file_obj = codecs.open(filename, 'r', encoding="utf8")
  File "/usr/lib64/python3.6/codecs.py", line 897, in open
    file = builtins.open(filename, mode, buffering)
FileNotFoundError: [Errno 2] No such file or directory: '/var/cache/rhn/reposync/1/External___Fedora_33__aarch64_/var/cache/zypp/raw/fedora33-aarch64/repodata/e68b2e6c0a5485dfb5c53e252d32fe64877fea162639b5309bb24b77971520a0-comps-Everything.aarch64.xml'
```

## GUI diff

No difference.

## Documentation

## Changelogs

- Fix yum reposync plugin for Fedora 33 and 34 repos

## Re-run a test


mcalmer (commented):
Please check 
https://github.com/uyuni-project/uyuni/blob/c753a55bcbf997eb61852c82e274e075b8fb3d63/backend/satellite_tools/repo_plugins/yum_src.py#L1150-L1153
if we need to change there also something.

mcalmer:
And to we need to use `_decompress()` or how is the result used?

vzhestkov:
Will wait for full sync of the repo and check.

vzhestkov:
Need to add `zchunk` as required package for `uyuni-common-libs`, but this package is available in Leap 15.3, but not in SLE 15

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
```diff
@@ -13,13 +13,13 @@
 -- in this software or its documentation.
 --
 
-CREATE UNIQUE INDEX rhn_pkg_cap_name_version_uq
-    ON rhnPackageCapability (name, version)
- where version is not null;
+CREATE INDEX rhn_pkg_cap_name_version_uq
```
moio, line 16: What is the reason for the removal of the `UNIQUE` clause?

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
```diff
@@ -13,13 +13,13 @@
...
+    ON rhnPackageCapability USING btree (sha512(replace(name, '\', '\\')::bytea), version)
```
moio, line 17: What is the reason for the indexing of the column hash instead of the column?

moio (changes requested):
I am not sure about some aspects, please let's go through the questions before this is merged.

Maybe it's nothing really - but this is a delicate area, I am proceeding with some extra care :wink:

uyuni/common-libs/common/fileutils.py
```diff
@@ -502,13 +502,15 @@ def decompress_open(filename):
...
         except ImportError: # No LZMA lib - be sad
-            # xz uncompresses foo.xml.xz to foo.xml
-            # uncompress, keep both, return uncompressed file
-            subprocess.call(['xz', '-d', '-k', filename])
-            uncompressed_path = filename.rsplit('.', 1)[0]
-            file_obj = codecs.open(uncompressed_path, 'rb', encoding="utf8")
+            file_obj = subprocess.Popen(["xz", "-d", "-k", filename],
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.DEVNULL).stdout
+    elif filename.endswith('.zck'):
```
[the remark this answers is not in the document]
vzhestkov: Yes, Fedora 34 Updates and Updates Modular. In my demo from Monday you may notice the errata info from Fedora 34 Updates. The only fear I have that there is no `zchunk` package in SLE 15 yet, but it's in Leap 15.3 and as @mcalmer we can add `RecommendsL zchunk` for it to the spec.

uyuni/common-libs/common/fileutils.py
```diff
@@ -502,13 +502,15 @@ def decompress_open(filename):
             import lzma
             file_obj = lzma.LZMAFile(filename, 'rb')
         except ImportError: # No LZMA lib - be sad
```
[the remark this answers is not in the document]
vzhestkov: Only in case if it's not included to `python-base`, I think it could happen if it was configured this way during package building, but not for openSUSE/SLE as it's there by default.

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
vzhestkov: It was done by accident, will put it back. Thanks for noticing it.

schema/spacewalk/common/tables/rhnPackageCapability.sql
```diff
@@ -19,7 +19,7 @@ CREATE TABLE rhnPackageCapability
     id        NUMERIC NOT NULL
                   CONSTRAINT rhn_pkg_capability_id_pk PRIMARY KEY
                   ,
-    name      VARCHAR(4000) NOT NULL,
+    name      TEXT NOT NULL,
```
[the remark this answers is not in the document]
vzhestkov: I tested it on my Raspberry Pi and there is no difference in VARCHAR->TEXT change and there are a lot of performance tests I found in google reporting the same.

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
vzhestkov: In case of using UNIQUE only `btree` can be used, but in case of large values it could lead to the following issue:
```
2021/08/03 14:41:21 +03:00 index row size 3488 exceeds btree version 4 maximum 2704 for index "rhn_pkg_cap_name_uq"
DETAIL:  Index row references tuple (215928,89) in relation "rhnpackagecapability".
HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
Consider a function index of an MD5 hash of the value, or use full text indexing.
```
Using the hash is the recommended solution for such cases.
Will do some extra checks with the speed of reindexing.

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
vzhestkov: Put `UNIQUE` back with https://github.com/uyuni-project/uyuni/pull/4073/commits/3902deae64cb7e9c4a0b244f65add7103b4f299c

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
vzhestkov: I did some more tests to check index creation speed and reindex and it seems reindexing with hash runs almost twice faster (I tested it with Uyuni installed on Raspberry Pi 4), but as I said before full text indexing is one of the recommendations, but it can't be used with `btree` which is required for `UNIQUE`. MD5 is not reliable, sha512 used instead, but it requres additionally replace `/` with `//`

moio: There are two sides to consider here:
 1- the `UNIQUE` part - this guarantees there are no duplicates. From that perspective, taking the hash of values and index hashes works (although it is a bit redundant, one could in principle use a hash index instead of btree of hashes - but that's unimportant)
 2- the `INDEX` part - this is about speeding up queries that search by a certain column. If you apply a function to the indexed column (`sha512(...)` in this case), then _only queries which search by that function will get performance improvements_.

IOW:

```sql
CREATE INDEX fast_index ON table (column);
```

speeds up queries like:
```sql
SELECT * FROM table WHERE colum = 'value';
```

While

```sql
CREATE INDEX fast_index ON table (sha512(column));
```

**only** speeds up queries like:
```sql
SELECT * FROM table WHERE column = sha512('value');
```

So queries like:
```sql
SELECT * FROM table WHERE column = 'value';
```

will **not** benefit from the index.


In this case, I fear we do have queries looking up capabilities by their name around the codebase, can you please check?

uyuni/common-libs/common/fileutils.py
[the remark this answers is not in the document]
moio: Then I'd say such a fall-back mechanism is not needed. We only support openSUSE/SLE at this point and the requirements (in terms of packages) are well defined.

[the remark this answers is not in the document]
agraul: uyuni-libs/common are also used on the clients, not only on the server. `lzma` in the standard library of Python 3, but not in Python 2.

If we want to keep Python2 compatibility, which I think we do, I would move the import statement to the top where it belongs and use one of the two common import guards:
```python
try:
    import lzma
except ImportError:
    lzma = None

if lzma is not None:
    file_obj = lzma.LZMAFile(...)
else:
    ...

# or

try:
    import lzma
    HAS_LZMA = True # could also be in an else clause
except ImportError:
    HAS_LZMA = False

if HAS_LZMA:
    file_obj = lzma.LZMAFile(...)
else:
    ...
```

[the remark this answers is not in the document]
vzhestkov: `decompress_open` is used by `reposync` and it's plugins only, I think it's quite safe if not to put import for this module to the global scope.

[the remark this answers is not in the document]
agraul: > `decompress_open` is used by `reposync` and it's plugins only, 

I'm surprised this is in uyuni.common if it's only used on the server, afaik it was created for functionality common to server and clients.

> I think it's quite safe if not to put import for this module to the global scope.

It is safe in the sense that nothing else needs it and there won't be a traceback or anything like that, I agree. For compliance with [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) I would still move it to the top.

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
vzhestkov: Just finished the test of uploading `16 542 013` records (`15 530 913` uniq records) to test tables with the same stracture and indexes as `rhnPackageCapability` has now, and with modified table. It takes `38m37.023s` minutes to upload all the data with old table structure and indexes and `28m6.253s` for modified table. Upload was made the same way as it's done in `reposync`. No any side effect from the IO, the tests were performed on NVMe drive. I think full text index helps for cases like:
```
SELECT * FROM table WHERE column = 'value';
```
as Postgres using indexes in this case the proper way by getting the hash and comparing with the indexes it has.

uyuni/common-libs/common/fileutils.py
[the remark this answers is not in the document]
moio: In principle, anything which is not common should be moved out of the common library/package - but maybe in a different PR.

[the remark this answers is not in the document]
vzhestkov: I will discuss it with @mcalmer as he moved this part to common library, probably there are some reasons for it. I'll move the import to the top and handle the case if the import is failing

schema/spacewalk/postgres/tables/rhnPackageCapability_index.sql
moio: Full text index is for searching by keyword, ie. for natural language and search-engine kinds of queries, which is not what we are looking for here.

Problem is in any code piece that `SELECT`s `rhnPackageCapability` by `name`, or `ORDER BY name` or `JOIN` (potentially). I did a quick search and could find the following:

https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/common/db/datasource/xml/Package_queries.xml#L138-L274
https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/rhnpackage/Package_satellite.hbm.xml#L237-L263
https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/postgres/procs/lookup_package_capability.sql#L33
https://github.com/uyuni-project/uyuni/blob/master/backend/server/rhnServer/server_packages.py#L347-L371
https://github.com/uyuni-project/uyuni/blob/master/backend/server/importlib/backend.py#L111
https://github.com/uyuni-project/uyuni/blob/master/backend/satellite_tools/updatePackages.py#L251-L259

I understand your tests look promising, but I would not want to really make sure each and every one of these code paths (plus any that have escaped my attention) use the hash instead of the string, or perform equally well with the modification. Also, we would have to remember this particular column of this particular table works differently from all others in any future code change - whilst `INDEX`ing relevant columns as-is is the norm.

Do we have the possibility of enlarging the buffer page instead?

uyuni/common-libs/common/fileutils.py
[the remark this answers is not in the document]
mcalmer: fileutils are also used on clients. This special function (decompress_open) not (yet). In the past it was part of spacewalk-backend which required to build spacewalk-backend for all client OSes which was very hard as spacewalk-backend include a lot of other dependencies.
I just moved complete files over to uyuni-common and did not check every single call if it is used on clients or not.

I am fine with ignoring the python2 compatibility for this function or move this function back to spacewalk-backen.

uyuni/common-libs/uyuni-common-libs.spec
```diff
@@ -67,6 +67,7 @@ BuildRequires:  python-devel
 %else
 BuildRequires:  python2-devel
 %endif
+Recommends:     zchunk
```
moio, line 70: I'm curious / may lack context: why `Recommends` and not `Requires`?

vzhestkov: It's not available on SLE yet, but it's present in Leap

vzhestkov: https://github.com/uyuni-project/uyuni/pull/4073#issuecomment-896741633

moio: Okay thanks for the clarification

moio (approved):
I do not feel informed enough to have an in-depth opinion on the Python parts but what I see LGTM.

Nice job!

mcalmer (approved):

PR #4075, opened 2021-08-02
Add Raspbian 9/10 and Fedora 33/34 repos

vzhestkov:
## What does this PR change?

Adds Raspian 9 and 10 and Fedora 33 and 34 repos to `spacewalk-common-channells`

## GUI diff

No difference.

## Documentation

## Changelogs

- Add Fedora 33 and 34 repositories
- Add Raspbian 9 and 10 repositories

## Re-run a test

### DIFF
--- a/utils/spacewalk-common-channels.ini
+++ b/utils/spacewalk-common-channels.ini
@@ -3020,3 +3020,165 @@ gpgkey_url = https://download.opensuse.org/repositories/systemsmanagement:/Uyuni
 gpgkey_id = %(_uyuni_gpgkey_id)s
 gpgkey_fingerprint = %(_uyuni_gpgkey_fingerprint)s
 repo_url = https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/EL8-Uyuni-Client-Tools/EL_8/
+
+[raspbian-9-pool-armhf-uyuni]
+label    = raspbian-9-pool-armhf-uyuni
+checksum = sha256
+archs    = arm-deb
+repo_type = deb
+name     = Raspbian 9 (stretch) pool for armhf for Uyuni
+gpgkey_url =
+gpgkey_id =
+gpgkey_fingerprint =
+repo_url = https://archive.raspbian.org/raspbian/dists/stretch/main/binary-armhf/
+
+[raspbian-9-armhf-contrib-uyuni]
+label    = raspbian-9-armhf-contrib-uyuni
+name     = Raspbian 9 (stretch) armhf Contrib for Uyuni
+archs    = arm-deb
+repo_type = deb
+checksum = sha256
+base_channels = raspbian-9-pool-armhf-uyuni
+repo_url = https://archive.raspbian.org/raspbian/dists/stretch/contrib/binary-armhf/
+
+[raspbian-9-armhf-firmware-uyuni]
+label    = raspbian-9-armhf-firmware-uyuni
+name     = Raspbian 9 (stretch) armhf Firmware for Uyuni
+archs    = arm-deb
+repo_type = deb
+checksum = sha256
+base_channels = raspbian-9-pool-armhf-uyuni
+repo_url = https://archive.raspbian.org/raspbian/dists/stretch/firmware/binary-armhf/
+
+[raspbian-9-armhf-non-free-uyuni]
+label    = raspbian-9-armhf-non-free-uyuni
+name     = Raspbian 9 (stretch) armhf Non-Free for Uyuni
+archs    = arm-deb
+repo_type = deb
+checksum = sha256
+base_channels = raspbian-9-pool-armhf-uyuni
+repo_url = https://archive.raspbian.org/raspbian/dists/stretch/non-free/binary-armhf/
+
+[raspbian-9-armhf-rpi-uyuni]
+label    = raspbian-9-armhf-rpi-uyuni
+name     = Raspbian 9 (stretch) armhf RPi for Uyuni
+archs    = arm-deb
+repo_type = deb
+checksum = sha256
+base_channels = raspbian-9-pool-armhf-uyuni
+repo_url = https://archive.raspbian.org/raspbian/dists/stretch/rpi/binary-armhf/
+
+[raspbian-10-pool-armhf-uyuni]
+label    = raspbian-10-pool-armhf-uyuni
+checksum = sha256
+archs    = arm-deb
+repo_type = deb
+name     = Raspbian 10 (buster) pool for armhf for Uyuni
+gpgkey_url =
+gpgkey_id =
+gpgkey_fingerprint =
+repo_url = https://archive.raspbian.org/raspbian/dists/buster/main/binary-armhf/
+
+[raspbian-10-armhf-contrib-uyuni]
+label    = raspbian-10-armhf-contrib-uyuni
+name     = Raspbian 10 (buster) armhf Contrib for Uyuni
+archs    = arm-deb
+repo_type = deb
+checksum = sha256
+base_channels = raspbian-10-pool-armhf-uyuni
+repo_url = https://archive.raspbian.org/raspbian/dists/buster/contrib/binary-armhf/
+
+[raspbian-10-armhf-firmware-uyuni]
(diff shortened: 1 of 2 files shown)


utils/spacewalk-common-channels.ini
```diff
@@ -3020,3 +3020,165 @@ gpgkey_url = https://download.opensuse.org/repositories/systemsmanagement:/Uyuni
 gpgkey_id = %(_uyuni_gpgkey_id)s
 gpgkey_fingerprint = %(_uyuni_gpgkey_fingerprint)s
 repo_url = https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/EL8-Uyuni-Client-Tools/EL_8/
+
+[raspbian-9-pool-armhf-uyuni]
```
juliogonzalez, line 3024: Debian 9 is only on LTS and goes 100% EoL next year (https://wiki.debian.org/DebianReleases). Do we really want to add it?

Raspbian is now renamed: https://www.raspberrypi.org/software/ Shouldn't we use the new name?

Also remember we need to add client tools repos as well (can come later)

utils/spacewalk-common-channels.ini
```diff
@@ -3020,3 +3020,165 @@ gpgkey_url = https://download.opensuse.org/repositories/systemsmanagement:/Uyuni
...
+archs    = x86_64, s390x, aarch64, ppc64le
+checksum = sha256
+name     = Fedora 33 (%(arch)s)
+gpgkey_url = https://getfedora.org/static/fedora.gpg
+gpgkey_id = 9570FF31
+gpgkey_fingerprint = 963A 2BEB 0200 9608 FE67  EA42 49FD 7749 9570 FF31
+repo_url = https://mirrors.fedoraproject.org/metalink?repo=fedora-33&arch=%(arch)s
+dist_map_release = 33
+
+[fedora33-modular]
```
juliogonzalez, line 3128: Last time I checked, Fedora was not working with Uyuni at all because it provided a newer salt version, incompatible with our master. Do I understand we plan to use Fedora to test the bundle?

If so, this LGTM. But keep in mind we still need to add client tools repositories.

juliogonzalez (changes requested):
Some comments.

utils/spacewalk-common-channels.ini
vzhestkov: Raspberry Pi OS is for x86 only, it's not the OS for Raspberry Pi itself

utils/spacewalk-common-channels.ini
vzhestkov: It's for Salt Bundle only. Pau asked to add Fedora 33/34 for Salt Bundle, most probably someone asked for it, I don't remember the details.

utils/spacewalk-common-channels.ini
vzhestkov: About the version I don't really know if we need to add version 9, but there is no much difference in case of Salt Bundle, and venv-salt-minion was build for version 9 already.

juliogonzalez: > Raspberry Pi OS is for x86 only, it's not the OS for Raspberry Pi itself

I think you are mixing the image installer with the OS?

Raspberry Pi OS is indeed the former Raspbian:


> Your Raspberry Pi needs an operating system to work. This is it. Raspberry Pi OS (previously called Raspbian) is our official supported operating system.
> https://www.raspberrypi.org/software/operating-systems/#raspberry-pi-os-32-bit

juliogonzalez: > About the version I don't really know if we need to add version 9, but there is no much difference in case of Salt Bundle, and venv-salt-minion was build for version 9 already.

Ok, let's add it, and we'll see.

vzhestkov: Oh, sorry. I see now `Raspberry Pi OS`. Some time ago I remember this name was used for this only https://www.raspberrypi.org/software/raspberry-pi-desktop/
I just missed the plans to change the name for the distro. Not sure which one is better to use in such case. Probably it's better to use Raspbian as it was originally named this way.

juliogonzalez: I am not sure. Following that rule could mean people is still calling us `SuSE` :-)

I'd suggest we use the official naming at the code and the doc, even if day-to-day we still call it "rasbpian".

utils/spacewalk-common-channels.ini
Shirocco88: Salt Bundle for Fedora-33 see: https://build.opensuse.org/project/show/systemsmanagement:saltstack:bundle:Fedora-33

utils/spacewalk-common-channels.ini
Shirocco88: Hint: the salt grains shows "Raspbian"
https://github.com/saltstack/salt/blob/master/salt/grains/core.py#L1555

utils/spacewalk-common-channels.ini
vzhestkov: @Shirocco88 , it's a build project of Salt Bundle, there are a list of packages which shouldn't be published for the channels synchronized to Uyuni as these packages are build dependencies only and cant break the dependencies of the OS.

utils/spacewalk-common-channels.ini
vzhestkov: Raspbian is everywhere in the distro, os-release also shows Raspbian. I think it's better to add these channels as `raspbian-...` as most of the users know it with this name and the distro is also self identifying as Raspbian

utils/spacewalk-common-channels.ini
Shirocco88: @vzhestkov :
thanks for your answer;
I want to evaluate Salt Bundle, for example with Debian-10;
what is the recommended way ?
i.) Download venv-salt-minion.deb 
or ii.) is there a possibility to mirror a repository with venv-salt-minion, build repo, bootstrap

any help welcome.

vzhestkov: @Shirocco88 `venv-salt-minion` package is only usable package from Salt Bundle project repos. So you need to sync only this package (in simple way just to sync all the repo and delete everything except venv-salt-minion). `venv-salt-minion` package is the only package required for bootstrapping, but please note that only Uyuni Master contains all the states aligned for `venv-salt-minion`. `/usr/share/susemanager/mgr_bootstrap_data.py` should be updated to use `venv-salt-minion` for creating bootstrap repo.

aric89: Are Fedora 36 and future builds going to be supported by chance? Sadly, the previous admin used Fedora for some of the server builds and it would be nice to manage these alongside our other Linux distros.

vzhestkov: I suppose, no. It requires significant effort for building the Salt Bundle for it, but almost no requests to add these distros.

aric89: Would I be able to add the repos to the spacewalk-common-channels.ini and use the saltstack installer instead? I was reading that Fedora needed to use the venv-minion instead so maybe that wouldn't be doable then?

vzhestkov: you can try to add the repos manually there and most probably these repos will be synched with no any issues, but the problem is salt here. We can't guaranee that Uyuni Server will work with salt-minion from upstream. Maybe better to try building venv-salt-minion for Fedora 36 by branching these project in OBS: https://build.opensuse.org/project/show/systemsmanagement:saltstack:bundle:Fedora35

aric89: Honestly, I wouldn't know where to begin trying to make a build. If it was a config/yaml file and say build this with these packages I could do it but if it's doing programming like Python then I would have difficulties. Do you have links I could try to learn from?

juliogonzalez:
@vzhestkov should we close this PR? All OS from this PR are EoL or close to EoL (Raspbian 10). We can always creare a new one for newer versions later.

vzhestkov:
@juliogonzalez I'm ok to close it if it's not really required :+1:

PR #4092, opened 2021-08-04
Revert moving import new statement to the top

jordimassaguerpla:
## What does this PR change?

This is a bit ugly... "import new" is not valid anymore for python3, but
instead we should use the types module and adapt the code.

The "import new" was "hidden" into a "try" block, so we were not seeing
the error.

This commit
https://github.com/uyuni-project/uyuni/pull/4029/commits/8fbd8fbdaee77945f95280540fe84c4b04266c3a

moved it to the top of the file and "voilà" the error
appeared and the proxy is broken.

This commit moves it back "under the carpet" to fix the proxy and hopes
it will be fixed later.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed

- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

N/A

- [X] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [X] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #4029, opened 2021-07-23
spacewalk-proxy: Enterprise Linux modifications

sbluhm:
## What does this PR change?

Multiple changes to spacewalk-proxy and spacewalk-proxy-installer packages to build on Enterprise Linux 8.

- Python modifications to pass pylint.
- Spec file adaptions and improvements.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes

- [X] **DONE**

## Test coverage
- No tests: already covered
Part of build process. Pylint tests on RHEL based systems.

spacewalk-proxy build tested ok on
 opensuse-leap-15.3-x86_64 

spacewalk-proxy-installer build tested ok  with PR #4039 (running pylint) for
 epel-8-x86_64 
 opensuse-leap-15.3-x86_64 

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #4095, opened 2021-08-04
Revert imports order breaking proxy by PR#4029

vzhestkov:
## What does this PR change?

Fixes regression caused by https://github.com/uyuni-project/uyuni/pull/4029

CFG is used there and it's too sensitive to the order of importing

Superseeds https://github.com/uyuni-project/uyuni/pull/4092

## GUI diff

No difference.

## Documentation

## Changelogs

- [X] No changelog needed


calancha (approved):

mcalmer (approved):

sbluhm:
@vzhestkov, does the order of the imports really make a difference? (apacheHandler.py and apacheServer.py). It is importing classes with different names and they are not reused.

vzhestkov:
@sbluhm, unfortunatelly yes, there is `CFG` used in the modules and it's too sensitive for the context of usage.

PR #4098, opened 2021-08-04
Mark SSH minion actions when they're picked up (bsc#1188505)

cbbayburt:
Prior to this change, SSH minion actions stayed as "Queued" until they finish. This PR sets the actions as "Picked up" as soon as the execution starts.

See: https://bugzilla.suse.com/1188505

## Documentation

## Test coverage
- No tests: already covered

## Links

https://github.com/SUSE/spacewalk/issues/15496

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/SSHMinionActionExecutor.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/SSHMinionActionExecutor.java
@@ -16,6 +16,7 @@
 package com.redhat.rhn.taskomatic.task;
 
 import com.redhat.rhn.GlobalInstanceHolder;
+import com.redhat.rhn.common.hibernate.HibernateFactory;
 import com.redhat.rhn.domain.action.Action;
 import com.redhat.rhn.domain.action.ActionFactory;
 import com.redhat.rhn.domain.server.MinionServer;
@@ -23,6 +24,7 @@ import com.redhat.rhn.domain.server.MinionServerFactory;
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
 
+import java.util.Date;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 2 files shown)


java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
```diff
@@ -2500,11 +2500,14 @@ public void executeSSHAction(Action action, MinionServer minion) {
...
 
-            sa.setRemainingTries(sa.getRemainingTries() - 1);
-
             Map<LocalCall<?>, List<MinionSummary>> calls = callsForAction(action,
                     Arrays.asList(new MinionSummary(minion)));
 
+            sa.setRemainingTries(sa.getRemainingTries() - 1);
+            sa.setStatus(ActionFactory.STATUS_PICKED_UP);
+            sa.setPickupTime(new Date());
+            HibernateFactory.commitTransaction();
```
mcalmer, line 2509: `commit` is not so good, but I think in this case needed. I think SSH keeps the connection open until the result is there. To switch the state 2 times and make it visible to others, we have to commit.
I am right?

cbbayburt: Exactly, it waits until the whole thing is finished otherwise. I tested it manually with this, and everything works, but @rjmateus could maybe say more.

cbbayburt: And tracing back the code, there's no other write operation happening before, so in this specific case, we should be safe.

mcalmer (approved):

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
rjmateus: I get your point, and we don't have many options here. For the state to be visible to others we need to commit to the database.
One option could be open a new DB transaction to commit just this change in the action state, but I don't think that effort is needed here.

I have just one question, why are the state change and commit after the `callsForAction` method call?
In my opinion we should do this just after making the decision we will be processing this action, to avoid more code that could be interacting with the database and be affected by this commit. Could these changes be moved to line 2502, just after the last check?

rjmateus (approved):

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
cbbayburt: > In my opinion we should do this just after making the decision we will be processing this action, to avoid more code that could be interacting with the database and be affected by this commit. Could these changes be moved to line 2502, just after the last check?

Considering this, and @mcalmer's question about the unit test safety, I'll move this to an outer class, to `SSHMinionActionExecutor#execute`. This part is never called from unit tests directly, so it would be safer for the future as well.

rjmateus (approved):
lgtm

mcalmer (approved):

PR #4111, opened 2021-08-10
Uyuni tftp script multiple ip on server

rjmateus:
## What does this PR change?

Solution for: https://github.com/SUSE/spacewalk/issues/15639

When a server has IPv4 and IPv6 configured installation script only allows IPv4. With this change, if no IP is passed, the default value will contain all the visible IPs (returned by host command) and will add them to apache configuration.

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
+++ b/tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
@@ -109,11 +109,11 @@ default_or_input "TFTP Boot directory" TFTPBOOT "/srv/tftpboot"
 
 default_or_input "SUSE Manager Proxy FQDN" SUMA_PROXY_FQDN $(hostname -f)
 
-default_or_input "SUSE Manager Proxy IP Address" SUMA_PROXY_IP $(host $SUMA_PROXY_FQDN | awk '{ print $4 }' || echo "")
+default_or_input "SUSE Manager Proxy IP Address" SUMA_PROXY_IP $(getent hosts $SUMA_PROXY_FQDN | awk '{ print $1 }' || echo "")
 
 default_or_input "SUSE Manager Server FQDN" SUMA_FQDN $PARENT_FQDN
 
-default_or_input "SUSE Manager Parent IP Address" SUMA_IP $( host $SUMA_FQDN | awk '{ print $4 }' || echo "" )
+default_or_input "SUSE Manager Parent IP Address" SUMA_IP $(getent hosts $SUMA_FQDN | awk '{ print $1 }' || echo "" )
 
 
 
--- a/tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.spec
+++ b/tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.spec
@@ -83,7 +83,7 @@ if [ -f /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew ]; then
             exit 1;
     fi;
 
-    SUMA_IP=$( host $PARENT_FQDN | awk '{ print $4 }' )
+    SUMA_IP=$(getent hosts $PARENT_FQDN | awk '{ print $1 }' || echo "")
 
     sed -i "s/^[[:space:]]*Allow from[[:space:]].*$/        Allow from $SUMA_IP/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew
     sed -i "s/^[[:space:]#]*Require ip[[:space:]].*$/        Require ip $SUMA_IP/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -633,9 +633,9 @@ When(/^I configure tftp on the "([^"]*)"$/) do |host|
   when 'server'
     $server.run("configure-tftpsync.sh #{ENV['PROXY']}")
   when 'proxy'
-    cmd = "configure-tftpsync.sh --tftpbootdir=/srv/tftpboot \
---server-fqdn=#{ENV['PROXY']} --server-ip=#{$server.public_ip} \
---proxy-fqdn=#{ENV['SERVER']} --proxy-ip=#{$proxy.public_ip}"
+    cmd = "configure-tftpsync.sh --non-interactive --tftpbootdir=/srv/tftpboot \
+--server-fqdn=#{ENV['SERVER']} \
+--proxy-fqdn=#{ENV['PROXY']}"
     $proxy.run(cmd)
   end
 end
--- a/tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.changes
+++ b/tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.changes
@@ -1,3 +1,5 @@
+- adapt configure-tftpsync.sh to work on machines with multiple IP's (bsc#1189263)
+
 -------------------------------------------------------------------
 Mon Aug 09 10:45:16 CEST 2021 - jgonzalez@suse.com
 


tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
```diff
@@ -105,6 +105,10 @@ if [ -z "$PARENT_FQDN" ]; then
         exit 1;
 fi;
 
+function join_by { local IFS="$1"; shift; echo "$*"; }
+
+SUMA_IP_DEFAULT=$(join_by " " $( host $SUMA_FQDN | awk '{if($5) print $5; else print $4;}' || echo "" ))
```
aaannz, line 110: I think this is too complex. awk has `$NF` to specify number of fields and can be used to print last field as well.

`awk {print $NF}`

And then you do not even need join_by function.

mcalmer (approved):
Please check if access via IPv6 is working. I saw only examples with multiple lines of "Require ip"

aaannz:
> Please check if access via IPv6 is working. I saw only examples with multiple lines of "Require ip"

I was surprised as well, but it worked in yesterday tests. One `Require ip $ipv4 $ipv6` directive.

tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
calancha: Without dollar symbol:
```
awk {print NF}
```

aaannz: Are you sure? Without dollar it will print number of fields, but the the last field. But with `$` it will print the last field.

So this should at the end look:
```bash
SUMA_IP_DEFAULT=$( host $SUMA_FQDN | awk '{print $NF}' || echo "" )
```

tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
```diff
@@ -113,7 +117,7 @@ default_or_input "SUSE Manager Proxy IP Address" SUMA_PROXY_IP $(host $SUMA_PROX
```
aaannz, line 117: I didn't notice before because it was hidden by github, but I think the same should be done for proxy:

```bash
default_or_input "SUSE Manager Proxy IP Address" SUMA_PROXY_IP $(host $SUMA_PROXY_FQDN | awk '{ print $NF }' || echo "")
```

tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
rjmateus: New version was updated

rjmateus:
I had change the code to get the resolved IP for the machine with `getent hosts`. This will return the resolve IP address for the host.
I changed also the apache configuration. Now we validate the hostname and not the IP adrress. Anyone see a problem on this change?

mcalmer (changes requested):
`getent hosts` was a bit too much. It returns only 1 IP address in my tests and not all.

tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.spec
```diff
@@ -83,10 +83,10 @@ if [ -f /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew ]; then
             exit 1;
     fi;
 
-    SUMA_IP=$( host $PARENT_FQDN | awk '{ print $4 }' )
+    SUMA_IP=$(getent hosts $PARENT_FQDN | awk '{ print $1 }' || echo "")
```
aaannz, line 86: This is not needed anymore.

mcalmer:
@rjmateus if you want to test the hostname, you can use a different apache rule. But I think using the IP was a bit more secure as a DNS could be spoofed. I would feel better when we configure all IP addresses. The admin can always overwrite it and define there own set of IPs

tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.spec
```diff
@@ -83,10 +83,10 @@ if [ -f /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew ]; then
...
 
-    sed -i "s/^[[:space:]]*Allow from[[:space:]].*$/        Allow from $SUMA_IP/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew
-    sed -i "s/^[[:space:]#]*Require ip[[:space:]].*$/        Require ip $SUMA_IP/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew
+    sed -i "s/^[[:space:]]*Allow from[[:space:]].*$/        Allow from $PARENT_FQDN/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew
+    sed -i "s/^[[:space:]#]*Require ip[[:space:]].*$/        Require ip $PARENT_FQDN/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew
```
aaannz, line 89:
```suggestion
    sed -i "s/^[[:space:]#]*Require (ip|host)[[:space:]].*$/        Require host $PARENT_FQDN/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf.rpmnew
```

tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
```diff
@@ -163,8 +163,8 @@ else
 fi
 
 
-sed -i "s/^[[:space:]]*Allow from[[:space:]].*$/        Allow from $SUMA_IP/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf
-sed -i "s/^[[:space:]#]*Require ip[[:space:]].*$/        Require ip $SUMA_IP/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf
+sed -i "s/^[[:space:]]*Allow from[[:space:]].*$/        Allow from $SUMA_FQDN/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf
+sed -i "s/^[[:space:]#]*Require ip[[:space:]].*$/        Require ip $SUMA_FQDN/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf
```
aaannz:
```suggestion
sed -i "s/^[[:space:]#]*Require (ip|host)[[:space:]].*$/        Require host $SUMA_FQDN/" /etc/apache2/conf.d/susemanager-tftpsync-recv.conf
```

aaannz:
> `getent hosts` was a bit too much. It returns only 1 IP address in my tests and not all.

The problem is that with apache we can have multiple IP addresses, but later in the wsgi script seems to me we can have only one. So we need to choose which one we use. See https://github.com/uyuni-project/uyuni/blob/master/tftpsync/susemanager-tftpsync-recv/add.wsgi#L101

rjmateus:
> `getent hosts` was a bit too much. It returns only 1 IP address in my tests and not all.

YEs, my idea is to configure with the IP address that is being resolved for the host name. If admin wants something different, the use the ip parameter of the script, of change the rule in apache.

I see your point in the apache rule, and I will revert it to the ip check

aaannz:
> @rjmateus if you want to test the hostname, you can use a different apache rule. But I think using the IP was a bit more secure as a DNS could be spoofed. I would feel better when we configure all IP addresses. The admin can always overwrite it and define there own set of IPs

Maybe time we start to validate DNSSEC? :)

tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
rjmateus: I have reverted the host check. So this comment is not needed anymore.

tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.spec
rjmateus: I have reverted the host check. So this comment is not needed anymore.

tftpsync/susemanager-tftpsync-recv/susemanager-tftpsync-recv.spec
rjmateus: I have reverted the host check. So this comment is not needed anymore.

SchoolGuy (approved):
I don't have the time to test this fully but in my eyes the code i am currently seeing is fine. If we support split horizon DNS then maybe we need to evaluate if this is actually working? However I would not expect that this works out of the box then.

aaannz (approved):

calancha (approved):
Approving my part: the testsuite changes.
For the rest, I trust the other peers expertise.

tftpsync/susemanager-tftpsync-recv/configure-tftpsync.sh
```diff
@@ -109,11 +109,11 @@ default_or_input "TFTP Boot directory" TFTPBOOT "/srv/tftpboot"
 
 default_or_input "SUSE Manager Proxy FQDN" SUMA_PROXY_FQDN $(hostname -f)
 
-default_or_input "SUSE Manager Proxy IP Address" SUMA_PROXY_IP $(host $SUMA_PROXY_FQDN | awk '{ print $4 }' || echo "")
+default_or_input "SUSE Manager Proxy IP Address" SUMA_PROXY_IP $(getent hosts $SUMA_PROXY_FQDN | awk '{ print $1 }' || echo "")
```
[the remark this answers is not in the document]
rjmateus: If I get it right, this is should the IP that is resolved for the host, and that is what I want. 
If admins want to add more IPs or a different one, they can set the IP parameter. This will only affect the default value

mcalmer (approved):
let's try it

PR #4115, opened 2021-08-11
Link to live patching filter from a system

Etheryte:
## What does this PR change?

Adds a link to the system overview that allows the user to create a CLM filter easily.

## GUI diff

No difference.

Before:

Screenshot 2021-08-11 at 13 29 20

After:

Screenshot 2021-08-11 at 13 28 59

- [x] **DONE**

## Documentation
- No documentation needed: minor cosmetic upgrade

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15270 

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/webapp/WEB-INF/pages/systems/sdc/overview.jsp
```diff
@@ -146,6 +146,7 @@
                 </c:when>
                 <c:otherwise>
                   <c:out value="${system.runningKernel}" />
+                  <a href="/rhn/manager/contentmanagement/filters?openFilterId=-1&openTemplate=LivePatchingSystem&systemId=${system.id}&systemName=${system.name}&kernelId=0000&kernelName=${system.runningKernel}">(<bean:message key="sdc.details.overview.create_clm_filter"/>)</a>
```
Etheryte, line 149: The `kernelId=0000` bit is the part I couldn't figure out, how can I figure out what the relevant kernel id is from the system details?

cbbayburt: This was also the issue with the initial implementation of the form. The problem is that kernel string is reported by salt directly from the system and it does not match a package name.

My suggestion is to just leave the kernel input as-is. It should show the latest kernel selected. This is how the form normally works anyway. Only thing the link has to do is to preselect a system, then the kernel input will automatically select the latest available. Most of the time, systems run the latest installed kernel anyway.

Etheryte: That sounds like a good compromise, thanks for the suggestion. I've updated the relevant bits.

java/code/webapp/WEB-INF/pages/systems/sdc/overview.jsp
```diff
@@ -146,6 +146,7 @@
                 </c:when>
                 <c:otherwise>
                   <c:out value="${system.runningKernel}" />
+                  <a href="/rhn/manager/contentmanagement/filters?openFilterId=-1&openTemplate=LivePatchingSystem&systemId=${system.id}&systemName=${system.name}&kernelName=${system.runningKernel}">(<bean:message key="sdc.details.overview.create_clm_filter"/>)</a>
```
Etheryte, line 149: Does this link need to be shown conditionally or is it fine to always show this?

cbbayburt: Hmm, good question. Not every user has live patching products, but the filter template itself is available unconditionally, so I'd say let's keep it like this unless we hear any feedback against it.

java/code/webapp/WEB-INF/pages/systems/sdc/overview.jsp
ncounter: So IIUC we do not provide any `kernel` hints for the target template, we just point directly to the live patching template for the system in the context. Then my question is: is it fine to have the link right next to the `kernel` entry, or should we put it generically next to the SUSE Manager System ID instead? Or next to the Live Patching `Installed Products:` entry instead? IIRC the Live Patching is an installed product, isn't it @cbbayburt ? I am not suggesting to move it already, I am more thinking out loud looking for reasoning, you know :grinning:

web/html/src/manager/content-management/list-filters/list-filters.renderer.tsx
```diff
@@ -22,6 +25,10 @@ export const renderer = (id, { filters, projectLabel, openFilterId, flashMessage
           filters={filtersJson}
           openFilterId={openFilterId}
           projectLabel={projectLabel}
+          openTemplate={openTemplate}
+          systemId={systemId}
+          systemName={systemName}
+          kernelName={kernelName}
```
cbbayburt, line 31: One thing that bothers me a bit is this horizontal expansion of template-specific parameters. It looks like it will keep growing as we add more templates. Can we maybe wrap these in an object and call it something like `templateValues`?

java/code/src/com/suse/manager/webui/controllers/contentmanagement/ContentManagementViewsController.java
```diff
@@ -174,6 +174,20 @@ public static ModelAndView listFiltersView(Request req, Response res, User user)
         if (req.queryParams("projectLabel") != null) {
             data.put("projectLabel", req.queryParams("projectLabel"));
         }
+        String openTemplate = req.queryParams("openTemplate");
+        if (openTemplate != null) {
+            data.put("openTemplate", openTemplate);
+
+            String systemId = req.queryParams("systemId");
+            String systemName = req.queryParams("systemName");
+            String kernelName = req.queryParams("kernelName");
```
cbbayburt, line 183: After the latest discussion, I guess `kernelName` can be removed from the whole chain

cbbayburt (changes requested):

java/code/webapp/WEB-INF/pages/systems/sdc/overview.jsp
cbbayburt: > IIRC the Live Patching is an installed product, isn't it @cbbayburt?

@ncounter you're right, but when you want to use CLM for live patching, you wouldn't install Live Patching directly from vendor channels. Therefore, you won't see it here until after you set up the CLM and assign the system to it.

OTOH, putting it next to the system ID seems too general to me. Only thing that's relevant to live patching is the kernel field on this page. Also keep in mind that when a live patch is actually installed on a system, the current live patch version shows up right below the kernel value here in the same field.

ncounter: Thanks @cbbayburt for the clarification. I fully agree with the current solution now 👍

ncounter (approved):

java/code/src/com/suse/manager/webui/controllers/contentmanagement/ContentManagementViewsController.java
Etheryte: `kernelName` is used in the system display name in the dropdown, but this part became redundant regardless thanks to moving the URL params bit to the frontend.

web/html/src/manager/content-management/list-filters/list-filters.renderer.tsx
Etheryte: Fetching the URL params in the component itself should fix this issue, or is there anything else you'd like to address here @cbbayburt?

cbbayburt (approved):
Thanks for the updates, @Etheryte looks good to me now :+1: 

Have you tested any of the older occurrences of `urlBounce` manually?

Etheryte:
Yes, I tested both the url param version and the fallback manually.

PR #4116, opened 2021-08-12
Fix Build image with authenticated registry (bsc#1185951) 

mbussolotto:
## What does this PR change?

The error in CI "Build image with authenticated registry.Build an image in the authenticated image store" seems to be cause by a short timeout.
The NPE seems to be not the root cause of the issue, considering that it happens also when the test are successfull (so apparently it does not have side effect).

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- [ ] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/14927
[bsc#1185951](https://bugzilla.suse.com/show_bug.cgi?id=1185951)

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/buildhost_docker_auth_registry.feature
+++ b/testsuite/features/secondary/buildhost_docker_auth_registry.feature
@@ -34,7 +34,7 @@ Feature: Build image with authenticated registry
     And I click on "submit-btn"
     Then I wait until I see "portus_profile" text
     # Verify the status of images in the authenticated image store
-    When I wait at most 600 seconds until container "portus_profile" is built successfully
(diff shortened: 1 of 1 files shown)


srbarrios:
Are we sure is about a timeout? If we give more time it ends properly?
Saying that because in the past we had issues with portus itself.

mbussolotto:
@srbarrios, after your comment I tried to had more test and yes, also with 15 min of timeout, the problem occurs. So I will definitely focus in portus and close this PR.
Thanks!

PR #4138, opened 2021-08-19
CI: Fix ownership on package building

jordimassaguerpla:
## What does this PR change?

When building packages with yarn inside a container, the ownership of some files get
changed to root. Without this commit, the ownership is fixed **only**
for new files, not for old ones.

This commit fixes the ownership of **all** files.

Otherwise, next run in jenkins will fail because it can't clean up the
environment.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed

- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/testing/automation/push-to-obs.sh
+++ b/susemanager-utils/testing/automation/push-to-obs.sh
@@ -63,7 +63,6 @@ if [ ! -f ${CREDENTIALS} ]; then
 fi
 
 INITIAL_CMD="/manager/susemanager-utils/testing/automation/initial-objects.sh"
-CHOWN_CMD="/manager/susemanager-utils/testing/automation/chown-objects.sh $(id -u) $(id -g)"
 docker pull $REGISTRY/$PUSH2OBS_CONTAINER
 
 test -n "$PACKAGES" || {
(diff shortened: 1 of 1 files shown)


susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -73,6 +72,8 @@ echo "Starting building and submission at $(date)"
 date
 [ -d ${GITROOT}/logs ] || mkdir ${GITROOT}/logs
 for p in ${PACKAGES};do
+    pkg_dir=$(cat rel-eng/packages/${p} | tr -s " " | cut -d" " -f 2)
+    CHOWN_CMD="chown -f -R $(id -u):$(id -g) /manager/$pkg_dir"
```
juliogonzalez, line 76:
```suggestion
    CHOWN_CMD_FULL="${CHOWN_CMD}; chown -f -R $(id -u):$(id -g) /manager/$pkg_dir"
```
And then of course change the `docker_run` call below.

juliogonzalez (commented):
IIUC now you are not considering any files that can be created outside the package folder, right?

That could be a problem. I'd suggest you keep both the old and the new command, so take care of both cases.

PR #4143, opened 2021-08-19
Uyuni payg cloud

rjmateus:
## What does this PR change?

Implementation of: https://github.com/SUSE/spacewalk/issues/15671

## GUI diff

Before:

After:

List
image

Detail
image

Create
image

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/domain/scc/SCCCachingFactory.java
```diff
@@ -422,6 +422,26 @@ public static void deleteSubscriptionBySccId(Long sccSubId) {
...
 
+    /**
+     * Get cloud rmt repository auth for a given Credential and Repository
+     * @param c the Credential (null not supported)
+     * @param r the Repository (null not supported)
+     * @return a of SCCRepositoriesAuth for clou
+     */
+    public static Optional<SCCRepositoryCloudRmtAuth> lookupCloudRepositoryAuthByCredentialAndRepo(Credentials c, SCCRepository r) {
+        if (c == null || r == null){
+            return null;
```
mcalmer, line 433:
```suggestion
            return Optional.empty();
```

mcalmer: ... or throw an exception

backend/satellite_tools/repo_plugins/deb_src.py
```diff
@@ -247,8 +247,8 @@ def get_package_list(self):
 class ContentSource:
 
     def __init__(self, url, name, insecure=False, interactive=True, yumsrc_conf=None,
-                 org="1", channel_label="", no_mirrors=False, ca_cert_file=None,
-                 client_cert_file=None, client_key_file=None):
+                 org="1", channel_label="", no_mirrors=False, http_headers=None,
```
agraul, line 250: Just a note, based on a quick `grep` I don't think it causes problems here but I think it could be worth sharing anyway.

By default, Python function* arguments can be passed by position or by keyword. Once the first argument is passed by keyword, no positional arguments are accepted. Changing the order of the arguments can (silently) change behavior when callers call the function with positional arguments. For this reason, new arguments are normally added to the end of the function signature. 
However, calling functions* using positional arguments is not good practice for functions with many arguments anyway since it makes the code harder to read.

Keyword-only calling can be enforced by adding a `*` to the function arguments. Any argument after it **must** be called by keyword:
 ```
 In [7]: def foo(*, bar=None, baz=None):
   ...:     print(bar, baz)
   ...: 

In [8]: foo()
None None

In [9]: foo(1,2)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-49e36bfb7156> in <module>
----> 1 foo(1,2)

TypeError: foo() takes 0 positional arguments but 2 were given
```

For functions that enforce keyword-only calling, new keyword arguments can be added anywhere since the order is not important. Functions that don't enforce keyword-only calling should be careful.

*: strictly speaking it is any callable object, not just functions

Comment (rjmateus) [reply backend/satellite_tools/repo_plugins/deb_src.py:250]: Thanks for the explanation Alex. I will adapt this change to add the new argument to the end of the list.

**Review by agraul — DISMISSED** [review]
I only looked at the Python code, which looks good overall!

On the structure of the `payg_extract_repo_data.py` script I would like to mention that I'd prefer separating I/O from parsing. That way testing the parsing logic is easier, it does not require mocking I/O. Speaking of testing, I think it generally makes sense to unit test what we write.

The `system_exit` function should be a lot simpler in my eyes, without any explicit type checking (duck typing, yeah!).

I also noted a few nitpicks that you're, as always, free to ignore.

**Review by aaannz — COMMENTED** [review]
I tried to follow RFC with implementation and it looks good. I only noticed some nitpicks.

What I am missing are the code comments, but that is overall issue in our code base. Without RFC at hand I would not be able to figure out what is purpose of some of the parts.

I was not involved in RFC so probably this is too late now, but why was plain ssh used instead of salt-ssh?

File: java/code/src/com/redhat/rhn/taskomatic/task/payg/PaygAuthDataExtractor.java [hunk 3]
```diff
@@ -0,0 +1,168 @@
...
+                    LOG.error("Error disconnection jsch session", e);
+                }
+            }
+            if (sessionsTarget != null) {
+                try {
+                    sessionsTarget.disconnect();
+                } catch (Exception e) {
+                    LOG.error("Error disconnection jsch session", e);
+                }
+            }
```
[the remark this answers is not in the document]
rjmateus: I don't think it will be a problem, because in that case the original exception will be propagate to the calling method

java/code/src/com/suse/manager/xmlrpc/admin/AdminPaygHandler.java
```diff
@@ -0,0 +1,206 @@
...
+import com.redhat.rhn.taskomatic.TaskomaticApi;
+import com.suse.manager.admin.PaygAdminManager;
+import org.apache.log4j.Logger;
+
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Maintenance Schedule XMLRPC Handler
```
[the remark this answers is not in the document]
rjmateus: It was a wrong copy past :)
I will correct it now.

java/code/src/com/redhat/rhn/taskomatic/task/payg/script/payg_extract_repo_data.py
```diff
@@ -0,0 +1,147 @@
...
+\0
+"""
+
+CREDENTIALS_NAME = "SCCcredentials"
+
+def system_exit(code, msgs=None):
+    "Exit with a code and optional message(s). Saved a few lines of code."
+
+    if msgs:
+        if type(msgs) not in [type([]), type(())]:
```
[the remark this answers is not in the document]
rjmateus: Makes sense. Thanks for the hint

java/code/src/com/redhat/rhn/domain/cloudpayg/CloudRmtHost.java
```diff
@@ -0,0 +1,121 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package com.redhat.rhn.domain.cloudpayg;
+
+import com.redhat.rhn.domain.BaseDomainHelper;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import javax.persistence.*;
```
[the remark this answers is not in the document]
rjmateus: No we don't. My IDEA was wrongly configured. I will check all imports and re-run checkstyle

java/code/src/com/redhat/rhn/taskomatic/task/payg/script/payg_extract_repo_data.py
```diff
@@ -0,0 +1,147 @@
...
+
+def _extract_http_auth(credentials):
+    credentials_file = '/etc/zypp/credentials.d/' + credentials
+    if not Path(credentials_file).exists():
+        system_exit(5, "Credentials file not found ({})".format(credentials_file))
+    with open(credentials_file) as credFile:
+        username = ""
+        password = ""
+        for line in credFile:
+            name, var = line.partition("=")[::2]
```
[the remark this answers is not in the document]
rjmateus: It's much simpler to read. I will change the code acording

java/code/src/com/redhat/rhn/taskomatic/task/payg/script/payg_extract_repo_data.py
```diff
@@ -0,0 +1,147 @@
...
+    if not is_payg_instance():
+        system_exit(1, "instance is not pay-as-you-go")
+
+    payg_data = load_instance_info()
+    print(json.dumps(payg_data))
+
+
+if __name__ == '__main__':
+    try:
+        sys.exit(abs(main() or 0))
```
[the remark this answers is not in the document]
rjmateus: make sense. I will change the code acordingly

java/code/src/com/redhat/rhn/taskomatic/task/payg/script/payg_extract_repo_data.py
```diff
@@ -0,0 +1,147 @@
...
+    with open(credentials_file) as credFile:
+        username = ""
+        password = ""
+        for line in credFile:
+            name, var = line.partition("=")[::2]
+            if "username" == name.strip():
+                username = var.strip()
+            elif "password" == name.strip():
+                password = var.strip()
+    return {"username": username, "password": password}
```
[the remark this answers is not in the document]
rjmateus: This one I will keep, since it will map directly to the Json I need to return

java/code/src/com/redhat/rhn/taskomatic/task/payg/script/payg_extract_repo_data.py
```diff
@@ -0,0 +1,147 @@
...
+    try:
+        auth_data_output = subprocess.check_output("/usr/lib/zypp/plugins/urlresolver/susecloud", input=input, stderr=subprocess.PIPE, universal_newlines=True)
+    except subprocess.CalledProcessError as e:
+        system_exit(3, ["Got error when getting repo processed URL and headers(error {}):".format(e)])
+
+    full_output = auth_data_output.split("\n")
+    _, header_auth, _, repository_url = full_output
+    repository_url_parsed = urlparse(repository_url)
+
+    return header_auth, repository_url_parsed.netloc
```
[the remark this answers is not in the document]
rjmateus: I will change it to namedtuple

java/code/src/com/redhat/rhn/domain/credentials/CredentialsFactory.java
```diff
@@ -140,6 +140,17 @@ public static Credentials createRegistryCredentials() {
         return creds;
     }
 
+    /**
+     * Helper method for creating new Cloud Rmt {@link Credentials}
+     * @return new credential with type Cloud Rmt
+     */
+    public static Credentials createCloudRmtCredentials() {
```
moio, line 147: I wonder whether "Cloud RMT" is a good name for this concept. Given you are using "PAYG SSH" elsewhere, will it help to settle on one piece of jargon only?

java/code/src/com/suse/manager/xmlrpc/admin/AdminPaygHandler.java
```diff
@@ -0,0 +1,215 @@
...
+     * @xmlrpc.param #param_desc("string", "host", "hostname or IP address to the instance, will fail if
+     * already in use.")
+     * @xmlrpc.param #param("int", "port")
+     * @xmlrpc.param #param("string", "username")
+     * @xmlrpc.param #param("string", "password")
+     * @xmlrpc.param #param_desc("string", "key", "private key to use in authentication")
+     * @xmlrpc.param #param("string", "keyPassword")
+     * @xmlrpc.returntype #return_int_success()
+     */
+    public int create(User loggedInUser, String description, String host, Integer port, String username,
```
moio, line 75: It'd be good to qualify what `create` creates in the method name I believe.

The full call will be `admin.payg.create(...)` - I have nothing against the `admin` prefix, but it's not like the user is "creating a pay-as-you-go" here. Rather, they are creating an SSH connection to harvest data from a pay-as-you-go host.

web/html/src/manager/admin/list-payg/list-payg.tsx
```diff
@@ -0,0 +1,200 @@
...
+import { Button } from "components/buttons";
+import { showSuccessToastr } from "components/toastr/toastr";
+import withPageWrapper from "components/general/with-page-wrapper";
+import { ServerMessageType } from "components/messages";
+import { FromNow } from "components/datetime/FromNow";
+import { HelpLink } from "components/utils";
+import { SectionToolbar } from "components/section-toolbar/section-toolbar";
+import PaygStatus from "manager/admin/payg-shared/common/payg-status";
+
+const _SETUP_WIZARD_STEPS = [
```
moio, line 18: It'd be great to have a few screenshots in the PR description.

java/code/src/com/suse/manager/webui/controllers/admin/mappers/ResponseMappers.java
```diff
@@ -0,0 +1,100 @@
...
+import com.suse.manager.webui.controllers.admin.beans.PaygProperties;
+import com.suse.manager.webui.controllers.admin.beans.PaygResumeResponse;
+import com.suse.manager.webui.utils.ViewHelper;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ResponseMappers {
```
moio, line 31: Should this class be renamed/moved to clarify it is specific to PAYG pages? I find it a bit misleading that the title is very generic, and placed in a very generic package.

Of course a different thing would be if more controllers delegated some processing to mappers - but IIUC this is really the first such case, right?

schema/spacewalk/common/data/rhnTaskoBunch.sql
```diff
@@ -117,4 +117,10 @@ INSERT INTO rhnTaskoBunch (id, name, description, org_bunch)
 INSERT INTO rhnTaskoBunch (id, name, description, org_bunch)
    VALUES (sequence_nextval('rhn_tasko_bunch_id_seq'), 'mgr-forward-registration-bunch', 'Forward registrations to SUSE Customer Center', null);
 
+INSERT INTO rhnTaskoBunch (id, name, description, org_bunch)
+VALUES (sequence_nextval('rhn_tasko_bunch_id_seq'), 'update-payg-data-bunch', 'Runs update-payg-data
+Parameters:
+- integer parameter payg instance ID
+- without parameter updates data for all instances', null);
```
moio, line 124: IIRC this text is user-visible. Is it intended or should this be a comment for developers only?

moio (commented):
I admit I only had a very passing review - hopefully comments left are useful anyway.

backend/satellite_tools/spacewalk-extra-http-headers
```diff
@@ -13,9 +10,6 @@ except:
 
 from zypp_plugin import Plugin
 
-REPOSYNC_EXTRA_HTTP_HEADERS_CONF = '/etc/rhn/spacewalk-repo-sync/extra_headers.conf'
```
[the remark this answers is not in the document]
rjmateus: This is still supported in the way it is documented. But in the previous implementation, the file was read in two places: Reposync codebase[1] and zypper plugin [2].
What I have done was read it in one place (repo sync codebase) (keeping [2]), merge the information with the one we get from the database[3], and save all headers in a headers file, side-by-side with the .repo file [4].

[1] https://github.com/uyuni-project/uyuni/blob/master/backend/satellite_tools/repo_plugins/yum_src.py#L435-L449
[2] https://github.com/uyuni-project/uyuni/blob/master/backend/satellite_tools/spacewalk-extra-http-headers#L16
[3] https://github.com/uyuni-project/uyuni/pull/4143/commits/b8bf5fb58a36e5c3cba2f36e9bf92a07730b2e78#diff-a21e619e3060e6bc326e973347c4f626ef83289a02f535c2e0dd856e42c997ebR423
[4] https://github.com/uyuni-project/uyuni/pull/4143/commits/b8bf5fb58a36e5c3cba2f36e9bf92a07730b2e78#diff-a21e619e3060e6bc326e973347c4f626ef83289a02f535c2e0dd856e42c997ebR626-R629

backend/satellite_tools/reposync.py
```diff
@@ -1821,7 +1826,9 @@ def _url_with_repo_credentials(self, url_in):
             url.password = base64.decodestring(credentials['password'].encode()).decode()
             # remove query parameter from url
             url.query = ""
-        return url.getURL()
+            if 'extra_auth' in credentials and credentials['extra_auth']:
```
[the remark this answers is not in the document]
rjmateus: I hit it this problem when I was running the testsuite with some mocks. So I correct the tests and decided to be extra careful in here.

web/html/src/manager/admin/list-payg/list-payg.tsx
rjmateus: The PR description already has some screenshots. 
However, I just notice that they are not up-to-date. I will update them.
There is any specific page that you this is missing?

backend/satellite_tools/reposync.py
```diff
@@ -506,15 +507,17 @@ def sync(self, update_repodata=True):
         repo_checksum_type = 'sha1'
         start_time = datetime.now()
         for data in self.urls:
-            data['source_url'] = self.set_repo_credentials(data)
+            data['source_url_auth'] = self.set_repo_credentials(data)
+            data['source_url'] = [u["url"] for u in data['source_url_auth']]
```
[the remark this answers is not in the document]
rjmateus: It's not in use in this method. The field `source_url` is used as input for the reposync processing. So I decided to keep the same logic for this field just in case it was use in outside in any other context I would not be aware of.

java/code/src/com/redhat/rhn/domain/credentials/CredentialsFactory.java
rjmateus: I know it's a new jargon, but it is really the authentication to the cloud rmt infrastructure. 
My idea was to keep the concepts separated, and if in the future we had a new way to obtain those credentials, it will not be tied to the PAYG concept. 
For example, imagine we deploy a SUMA PAYG instance, in which we have access to cloud-rmt network. In that case we can use the same tables and authentication logic and is not tied to the instance PAYG logic.

java/code/src/com/suse/manager/xmlrpc/admin/AdminPaygHandler.java
rjmateus: The name `admin.payg.creatSshConnection` would make more sense to you?

java/code/src/com/suse/manager/webui/controllers/admin/mappers/ResponseMappers.java
rjmateus: There are other cases that also delegate the mapping to an external class [1] [2].
I agree with you that the name is too generic. I will remane it to `PaygResponseMappers.java`.

[1] https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/controllers/clusters/mappers/ResponseMappers.java
[2] https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/controllers/contentmanagement/mappers/ResponseMappers.java

schema/spacewalk/common/data/rhnTaskoBunch.sql
rjmateus: AFAIK this text is only returned in one method of tasko xml rpc API.
The text that is shown in the UI is defined in another place [1].
I follow the same logic as other job definitions [2].

[1] e5f6a68#diff-44724356f81f311b199b24e5eea35f54fcefb6e7ea7ecf413f60b9823626c36aR23718-R23723
[2] https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/common/data/rhnTaskoBunch.sql#L20-L31

java/code/src/com/redhat/rhn/domain/scc/SCCRepository.java
```diff
@@ -286,25 +286,33 @@ public boolean isAccessible() {
             else {
                 Optional<CredentialsType> currentCredType =
                         Optional.ofNullable(a.getOptionalCredentials().orElse(new Credentials()).getType());
-                if (currentCredType.isPresent() && currentCredType.get().getLabel().equals(Credentials.TYPE_SCC)) {
-                    result = Optional.of(a);
+
+                // if the credentials are of type SCC and have URL that mians that is using a mirror
+                // in this case we want to use it
```
mcalmer, line 291: This is not correct. If the URL is set, it means "this is the primary credential".
Using a mirror of SCC is not supported. Only "from disk" which has different settings and no credential at all is specified.

rjmateus: Thanks for the heads-up. I have changed the comment accordingly. 
A review from you would be much appreciated, especially in this method. I changed it and I think I kept all the previous logic. All pre-existing tests are passing without any change to them, and I added one more to test the newly introduced logic.

web/html/src/manager/admin/create-payg/create-payg.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <AsyncButton
+            id="savebutton"
+            className="btn-primary"
+            title={t("Add pay-as-you-go ssh data")}
+            text={t("Create")}
+            icon="fa-plus"
+            action={() =>
+              onAction(payg.properties, "create")
+                .then((data) => {
+                  console.log(data);
```
Etheryte, line 52: This seems obsolete?

rjmateus: it is :+1: thanks for the review

web/html/src/manager/admin/list-payg/list-payg.tsx
```diff
@@ -0,0 +1,205 @@
...
+    }
+  }, []);
+
+  const searchData = (row, criteria) => {
+    const keysToSearch = ["host", "description", "status"];
+    if (criteria) {
+      return keysToSearch
+        .map((key) => row[key])
+        .join()
+        .toLowerCase()
```
Etheryte, line 76: Here and elsewhere, please prefer `toLocaleLowerCase()`.

web/html/src/manager/admin/list-payg/list-payg.tsx
```diff
@@ -0,0 +1,205 @@
...
+      showSuccessToastr(props.flashMessage);
+    }
+  }, []);
+
+  const searchData = (row, criteria) => {
+    const keysToSearch = ["host", "description", "status"];
+    if (criteria) {
+      return keysToSearch
+        .map((key) => row[key])
+        .join()
```
Etheryte, line 75: If I'm not mistaken, this will give unexpected results when someone searches for a string containing a comma etc. Perhaps something along these lines would work instead?  

```ts
if (criteria) {
  const needle = criteria.toLocaleLowerCase();
  return keysToSearch.map(...).some(item => item.includes(needle));
}
```

web/html/src/manager/admin/list-payg/list-payg.tsx
```diff
@@ -0,0 +1,205 @@
...
+      </ul>
+    </div>
+  );
+
+  let pageContent = (
+    <Table
+      data={props.payg_instances}
+      identifier={(row) => row.host}
+      initialSortColumnKey="name"
+      initialItemsPerPage={window.userPrefPageSize}
```
Etheryte, line 113: Specifying `initialItemsPerPage` with the default value is not needed (https://github.com/uyuni-project/uyuni/pull/4617).

web/html/src/manager/admin/payg-shared/api/payg-actions-api.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Cancelable } from "utils/functions";
+import Network from "utils/network";
+
+type returnUsePaygActionsApi = {
+  onAction: (actionBodyRequest: any, action: NetworkActionKey, id?: String | null) => Promise<any>;
+  cancelAction: () => void;
+  isLoading: boolean;
+};
+
+const networkAction = {
```
Etheryte, line 12: Instead of duplicating all of this logic from `use-lifecycle-actions-api.ts`, I would recommend lifting it up instead. Since using an api this way looks like a common pattern, I would recommend making it a generic hook, akin to `usePrevious`. Feel free to ping me if you'd like some help with this.

web/html/src/manager/admin/payg-shared/info/payg-info-edit.tsx
```diff
@@ -0,0 +1,73 @@
...
+  payg: PaygFullType;
+  onChange: Function;
+  editing?: boolean;
+};
+
+const PaygInfoEdit = (props: Props) => {
+  const { onAction, cancelAction, isLoading } = useLifecyclePaygActionsApi();
+
+  const saveAction = ({ item, closeDialog, setErrors }) => {
+    console.log(item);
```
Etheryte, line 22: This seems redundant?

web/html/src/manager/admin/payg-shared/sshData/payg-ssh-data-edit.tsx
```diff
@@ -0,0 +1,79 @@
...
+        }
+      });
+  };
+  return (
+    <div id={props.labelPrefix + "-panel-wrapper"}>
+      <CreatorPanel
+        id={(props.isInstance ? "bastion_" : "") + "connection"}
+        creatingText={t("Edit " + props.labelPrefix)}
+        panelLevel="2"
+        title={props.labelPrefix + " SSH connection"}
```
Etheryte, line 46: This should probably also be translated.

web/html/src/manager/admin/payg-shared/sshData/payg-ssh-data-form.tsx
```diff
@@ -0,0 +1,104 @@
...
+  editing?: boolean;
+};
+
+export const PaygSshDataFormFields = (props: PropsFields) => {
+  let prefix = props.isInstance ? "" : "bastion_";
+  return (
+    <React.Fragment>
+      <div className="row">
+        <Text
+          required={props.isInstance ? true : false}
```
Etheryte, line 44: Since `isInstance` is already boolean, this can be `required={props.isInstance}`. If this is not the case, the prop definition needs to be updated.

web/html/src/manager/admin/payg-shared/sshData/payg-ssh-data-view.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type Props = {
+  payg: paygProperties;
+  isInstance: boolean;
+};
+
+const PaygShhDataView = (props: Props) => {
+  return (
+    <div>
+      <React.Fragment>
```
Etheryte, line 13: Both of these fragments are redundant. Fragments are required when you want to return multiple children from a single component as in a partial, but when you have wrapping divs this is not needed. Also, the wrapping divs seem to be duplicated here.

web/html/src/manager/admin/payg/payg.renderer.tsx
```diff
@@ -0,0 +1,30 @@
...
+  try {
+    paygJson = JSON.parse(payg || "");
+  } catch (error) {
+    console.log(error);
+  }
+
+  SpaRenderer.renderNavigationReact(
+    <RolesProvider>
+      <MessagesContainer />
+      <Pyag payg={paygJson} {...(wasFreshlyCreatedMessage && { wasFreshlyCreatedMessage })} />
```
Etheryte, line 26: The logic around `wasFreshlyCreatedMessage` looks a bit confusing, what's the intent here?

Etheryte (changes requested):
In general, the frontend part looks good. There's a few smaller things here and there that would be nice to tighten up a little bit. Tagging @ncounter for a review as well since this is a pretty big chunk in one go, so I would appreciate another set of eyes as well.

web/html/src/manager/admin/list-payg/list-payg.tsx
rjmateus: done

web/html/src/manager/admin/list-payg/list-payg.tsx
rjmateus: Done. Thanks for the int, it's way better now.

web/html/src/manager/admin/list-payg/list-payg.tsx
rjmateus: done. thanks

web/html/src/manager/admin/payg-shared/info/payg-info-edit.tsx
rjmateus: yes it is. Thanks for the int

web/html/src/manager/admin/payg/payg.renderer.tsx
rjmateus: I have simplified the code in this, and I have tested and looks like it is working. Thanks

web/html/src/manager/admin/payg-shared/sshData/payg-ssh-data-edit.tsx
rjmateus: Done

web/html/src/manager/admin/payg-shared/api/payg-actions-api.ts
Etheryte: Sure, I think that makes sense yeah. I think it makes sense to assign this ticket to me, the only downside is that I won't know how to test this well. So probably I will bug you for a review and input when I will work on it.

Etheryte (approved):
Thanks a lot for taking the time to go through the frontend bits, I appreciate it. I think it's very solid now. 👍

web/html/src/manager/admin/payg-shared/api/payg-actions-api.ts
rjmateus: Sure that also makes sense. If you don't mind to do the refactoring that would be great. I'm happy to help in anything or even pair program and take the opportunity to learn more.
For the test, I have added some cucumber test, which should catch any problem on the UI.
I will create the card after this PR get merged.

backend/satellite_tools/reposync.py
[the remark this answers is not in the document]
meaksh: Just wondering here if there is any particular reason to introduce this new `data['source_url_auth']` field here and not just keep using `data['source_url_auth']`?

rjmateus:
@meaksh I used a different variable `data['source_url_auth']` because I changed the type of the content, from list of strings to list of objects.
I kept the variable `data['source_url']` with the exact same logic as before, since it is assumed in the tests, and I'm not confident it's not assumed in any other place.
In this way, it will keep the same logic as before for that field.

mcalmer (commented):
first take

backend/satellite_tools/reposync.py
```diff
@@ -1829,7 +1834,9 @@ def _url_with_repo_credentials(self, url_in):
             url.password = base64.decodestring(credentials['password'].encode()).decode()
             # remove query parameter from url
             url.query = ""
-        return url.getURL()
+            if 'extra_auth' in credentials and credentials['extra_auth']:
+                headers = json.loads(credentials['extra_auth'].tobytes())
```
mcalmer, line 1838: Only 1 question left: must `extra_auth` really be a byte array/blob in DB or could it be also just "TEXT".
Might be easier to handle. But now everything is written and tested we might better stay with it.

java/code/src/com/redhat/rhn/domain/scc/SCCRepository.java
```diff
@@ -56,7 +57,19 @@
...
+                        "where r.url like :urlEndpoint"),
+        @NamedQuery(name = "SCCRepository.lookupByProductNameAndArch",
+                query = "select distinct r from SCCRepository r " +
+                        " join r.products pr " +
+                        " join pr.product p " +
+                        " join p.arch a " +
+                        " where lower(p.name) = lower(:product_name) " +
+                        " and lower(a.label) = lower(:arch_name) " +
+                        " and pr.channelLabel not like '%debuginfo%' " +
+                        " and r.installerUpdates = 'N' ")
```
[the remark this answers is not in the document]
rjmateus: I'm not sure int this one. What does this flag `pr.mandatory` are including? My goal in here is to add authentication to all channels except source and debuginfo, and then user can use the existing add product features.

schema/spacewalk/common/tables/suseCredentials.sql
```diff
@@ -28,9 +28,14 @@ CREATE TABLE suseCredentials
...
+    modified TIMESTAMPTZ DEFAULT (current_timestamp) NOT NULL,
+    extra_auth bytea,
+    payg_ssh_data_id NUMERIC
+        CONSTRAINT suse_credentials_payg_ssh_data_id_fk
+            REFERENCES susePaygSshData (id)
+);
 
-;
+ALTER TABLE susecredentials
+    ADD CONSTRAINT suse_credentials_payg_ssh_data_id_uq unique (payg_ssh_data_id);
```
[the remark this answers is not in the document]
rjmateus: I will check it, but I think it can cope with several different entries with null values. 
But I will change this and add a condition to the unique constraint.

java/code/src/com/redhat/rhn/domain/scc/SCCRepository.java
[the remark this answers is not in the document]
mcalmer: Why do you exclude "debuginfo" ? Does RMT do not mirror them?

There maybe optional channels which are not debuginfo. So maybe excluding optional channels might not be the right thing

backend/satellite_tools/reposync.py
rjmateus: It could be just text, or it can be even Json or Jsonb. I added it as byte array so it can be more flexible in the future and add anything that could be needed.
If you think is best, I can added change to text or json and adapt the code.

mcalmer: keep it as bytes. This is now tested and all places are aligned.

schema/spacewalk/common/tables/suseCredentials.sql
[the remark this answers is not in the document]
rjmateus: I just checked it and null can be used multiple times.
In docs [1] says: "Null values are not considered equal. "

However, if you think that having a partial unique index would be better, even for readability, I'm fine on changing it. 

[1] https://www.postgresql.org/docs/13/indexes-unique.html

java/code/src/com/redhat/rhn/domain/scc/SCCRepository.java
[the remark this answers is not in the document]
rjmateus: I made a test and looks like debug info repo is being mapped in cloud RMT. I will remove this filter from the query

java/code/src/com/redhat/rhn/domain/scc/SCCCachingFactory.java
```diff
@@ -461,6 +461,20 @@ public static void deleteSubscriptionBySccId(Long sccSubId) {
...
 
+    /**
+     * Returns a list of repositories for a product, independent of the version, and arch.
+     *
+     * @param productName name of the product we want to filter
+     * @param archName arch name we want to filter
+     * @return List of repositories for all version of one product and arch
+     */
+    public static List<SCCRepository> lookupRepositoriesByProductNameAndArch(String productName, String archName) {
+        return getSession().getNamedQuery("SCCRepository.lookupByProductNameAndArch")
```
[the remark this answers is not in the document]
rjmateus: I changed the name to `lookupRepositoriesByProductNameAndArchForPayg`.
Another possible solution would be to change the query to not filter by installer updates, and filter only for product name and arch, and on java side filter it to ignore the installer updates.

java/code/src/com/redhat/rhn/domain/scc/SCCRepository.java
```diff
@@ -271,10 +284,31 @@ public boolean isAccessible() {
                 }
             }
             else {
-                if (a.getOptionalCredentials().orElse(new Credentials()).getUrl() != null) {
+                Optional<CredentialsType> currentCredType =
```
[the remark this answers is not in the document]
rjmateus: thanks for the contribution @lucidd It looks much better. I will commit your suggestion

schema/spacewalk/common/tables/suseCredentials.sql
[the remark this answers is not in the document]
mcalmer: If it also works this way i am fine

java/code/src/com/redhat/rhn/domain/scc/SCCRepository.java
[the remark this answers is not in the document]
mcalmer: That sounds right

java/code/src/com/redhat/rhn/domain/scc/SCCCachingFactory.java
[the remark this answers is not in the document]
mcalmer: This is good, especially as the query does not exclude debuginfo repos anymore.

mcalmer (approved):

ktsamis (approved):

testsuite/features/secondary/srv_payg_ssh_connection.feature
```diff
@@ -0,0 +1,124 @@
+# Copyright (c) 2015-2021 SUSE LLC
```
Bischoff, line 1: 2015-2021 -> 2022

testsuite/features/secondary/srv_payg_ssh_connection.feature
```diff
@@ -0,0 +1,124 @@
...
+# Licensed under the terms of the MIT license.
+
+Feature: Listing, adding and removing ssh connection data for the payg feature
+  In order to list available ssh connections and add or remove them
+  As admin user
+  I want to use the web UI
```
Bischoff, line 7: In order to use payg
As admin user
I want to list available ssh connections and add or remove them

testsuite/features/secondary/srv_payg_ssh_connection.feature
```diff
@@ -0,0 +1,124 @@
...
+    And I click on "Save" in "Bastion SSH connection" modal
+    Then I should see a "Pay-as-you-go properties updated successfully" text
+    And I should see a "my-new-bastion.local" text in element "Bastion-panel-wrapper"
+    And I should see a "222" text in element "Bastion-panel-wrapper"
+    And I should see a "b_new_rootFull" text in element "Bastion-panel-wrapper"
+    And I should see a "b_new_passwordFull" text in element "Bastion-panel-wrapper"
+    And I should see a "b_new_keyFull" text in element "Bastion-panel-wrapper"
+    And I should see a "b_new_keyPasswordFull" text in element "Bastion-panel-wrapper"
+
+  Scenario: Delete minimal information for payg ssh connection data
```
Bischoff, line 110: Prefix with `Cleanup:` please (unless it's not part of the idempotency cleanup)

testsuite/features/secondary/srv_payg_ssh_connection.feature
```diff
@@ -0,0 +1,124 @@
...
+    When I follow the left menu "Admin > Setup Wizard > Pay-as-you-go"
+    And I follow "my-host.local"
+    And I click on "Delete"
+    And I click on "Delete" in "Delete Pay-as-you-go" modal
+    Then I should not see a "my-host.local" link
+    And I should not see a "my-host.local" text
+
+  Scenario: Delete Full information for payg ssh connection data
```
Bischoff, line 118: Prefix with `Cleanup:` please (unless it's not part of the idempotency cleanup)

testsuite/features/secondary/srv_payg_ssh_connection.feature
```diff
@@ -0,0 +1,124 @@
...
+    When I follow the left menu "Admin > Setup Wizard > Pay-as-you-go"
+    And I follow "my-host-full.local"
+    And I click on "Delete"
+    And I click on "Delete" in "Delete Pay-as-you-go" modal
+    Then I should not see a "my-bastion.local" link
+    And I should not see a "my-host.local" text
```
Bischoff, line 124: Add final linefeed

Is everything cleaned up? what happens if you run this feature two times in a row (idempotency) ?

Bischoff (changes requested):
Great work on test suite, thanks a lot.

A few required changes though.

testsuite/features/secondary/srv_payg_ssh_connection.feature
rjmateus: the delete is part of the test, it's not just for cleanup

testsuite/features/secondary/srv_payg_ssh_connection.feature
rjmateus: the delete is part of the test, it's not just for cleanup

testsuite/features/secondary/srv_payg_ssh_connection.feature
rjmateus: yes, all are cleaned at this point, and you can run it multiple time (I have tested it)

srbarrios (approved):
Test code looks good to me.

rjmateus:
> Great work on test suite, thanks a lot.
> 
> A few required changes though.

Thanks @Bischoff , all question are now addressed

Bischoff (approved):
thanks for the nitpicks

rjmateus:
@digdilem thanks for the report. I opened an issue to solve this.
https://github.com/uyuni-project/uyuni/issues/5400

PR #4159, opened 2021-08-24
fix deprecation warning in cpuinfo.py

ZsBT:
## What does this PR change?
deprecation warning appears on log.warn()

**add description**

log.warn is deprecated so use log.warning

## GUI diff

No difference.

## Documentation

## Test coverage
tested at my infrastructure

- [x] **DONE**

## Changelogs

- [x] No changelog needed


agraul (approved):
Haven't run the change, but I know that `Logger.warn` is deprecated and `Logger.warning` is the correct function to use. Thanks for the pull request!

Small side note: I don't know why there is no deprecation warning in 3.6, it's already documented as deprecated in the [3.6 documentation](https://docs.python.org/3.6/library/logging.html#logging.Logger.warning)

PR #4199, opened 2021-09-08
CI: improve error reporting when building packages

jordimassaguerpla:
## What does this PR change?

Only show logs when there has been an error and show which package
failed to build, not only the pid of the process that failed.

This should make it easier to find errors.

## GUI diff

Before: A lot of output for packages that were successfully built, so finding the one that failed was like finding a needle in a haystack

After: Only the log of the failed package is shown, with a message stating which package is it.

- [X] **DONE**

## Documentation
- No documentation needed

- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/15560

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/testing/automation/push-to-obs.sh
+++ b/susemanager-utils/testing/automation/push-to-obs.sh
@@ -117,7 +117,6 @@ if [ $PRET -ne 0 ];then
     echo "The following packages failed to build:"
     for p in ${PKG_FAILED}; do echo "- ${p}"; done
     echo "Please review the logs at ${GITROOT}/logs/"
-    echo "Review the logs."
     exit ${PRET}
 fi
 
--- a/susemanager-utils/testing/automation/push-to-obs.sh
+++ b/susemanager-utils/testing/automation/push-to-obs.sh
@@ -1,4 +1,6 @@
 #! /bin/sh -e
+# Set pipefail so the result of a pipe is different to zero if one of the commands fail.
+# This way, we can add tee but be sure the result would be non zero if the command before has failed.
 set -o pipefail
 
 HERE=`dirname $0`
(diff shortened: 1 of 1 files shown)


susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -78,8 +80,6 @@ for p in ${PACKAGES};do
     pkg_dir=$(cat rel-eng/packages/${p} | tr -s " " | cut -d" " -f 2)
     CHOWN_CMD="${CHOWN_CMD}; chown -f -R $(id -u):$(id -g) /manager/$pkg_dir"
     CMD="/manager/susemanager-utils/testing/docker/scripts/push-to-obs.sh -d '${DESTINATIONS}' -c /tmp/.oscrc -p '${p}' ${VERBOSE} ${TEST} ${OBS_TEST_PROJECT} ${EXTRA_OPTS}"
-    # Set pipefail so the result of a pipe is different to zero if one of the commands fail.
```
juliogonzalez, line 81: Not that matters, but I wonder why we had this comment here if we didn't use `set -o pipefail` :-D

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -97,19 +97,24 @@ echo "End of task at ($(date). Logs for each package at ${GITROOT}/logs/"
 set +e
 
 PRET=0
+PKG_FAILED=""
 for i in $PIDS;do
-    echo "Waiting for pid $i"
+    package_name=$(basename $(readlink ${GITROOT}/logs/${i}.log) | cut -d"." -f1 )
+    echo "Waiting for process with pid $i, building package ${package_name}"
```
juliogonzalez, line 103:
```suggestion
    echo "Waiting for process with pid ${i}, building package ${package_name}"
```

susemanager-utils/testing/automation/push-to-obs.sh
```diff
@@ -97,19 +97,24 @@ echo "End of task at ($(date). Logs for each package at ${GITROOT}/logs/"
...
+        echo "Seems there was an error with process pid ${i}"
+        echo "When building package ${package_name}"
+        cat ${GITROOT}/logs/${i}.log
+        PKG_FAILED="${PKG_FAILED} ${package_name}"
         PRET=-1
     fi
 done
 
 if [ $PRET -ne 0 ];then
+    echo "$Fail to build packages ${PKG_FAILED}"
```
juliogonzalez, line 117: Doesn't seem this `$Fail` environment exists, but anyway I suggest:

```suggestion
    echo "The following packages failed to build:"
    for p in ${PKG_FAILED}; do echo "- ${p}"; done
    echo "Please review the logs at ${GITROOT}/logs/"
```
Will be more readable if there are a few (or many) packages failing to build :-)

And since they can check the workspace it's easier to review file by file than looking at the log above.

If you like this approach, remember to remove the next line.

juliogonzalez (approved):
Some minor things.

susemanager-utils/testing/automation/push-to-obs.sh
jordimassaguerpla: I moved the set to the top and forgot to move the comment on a previous commit

PR #4202, opened 2021-09-09
Fix time zone mismatch in recurring events

Etheryte:
## What does this PR change?

Fixes time zone offset problems in recurring events views.  

Also fixes a number of missing key bugs and related issues I found when debugging the recurring events view.

## GUI diff

Minor cosmetic difference, nothing to document.

- [x] **DONE**

## Documentation
- No documentation needed: only very minor user invisible changes.

- [x] **DONE**

## Test coverage
- No tests: already covered by Cucumber which uncovered this problem.

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15841

Part of https://github.com/SUSE/spacewalk/issues/9049

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/state/highstate-summary.tsx
+++ b/web/html/src/manager/state/highstate-summary.tsx
@@ -53,6 +53,7 @@ export default function HighstateSummary({ minionId }) {
   }
 
   return (
+    // TODO: This identifier seems wrong, what's the correct identifier here?
     <>
       <Table identifier={state => state.state} data={summary} initialItemsPerPage={0}>
         <Column header={t("State Source")} comparator={Utils.sortByText} columnKey="name" cell={source => <State minionId={minionId} state={source} />} />
--- a/web/html/src/manager/state/recurring-states.tsx
+++ b/web/html/src/manager/state/recurring-states.tsx
@@ -6,6 +6,7 @@ import { RecurringStatesList } from "./recurring-states-list";
 import { RecurringStatesEdit } from "./recurring-states-edit";
 import { Utils as MessagesUtils } from "components/messages";
 import SpaRenderer from "core/spa/spa-renderer";
+import { localizedMoment } from "utils";
 
 /**
  * See:
@@ -52,9 +53,7 @@ function inferEntityParams() {
   return "";
 }
 
-type Props = {
-
-};
+type Props = {};
 
 type State = {
   messages: any[];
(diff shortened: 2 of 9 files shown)


web/html/src/manager/state/recurring-states.tsx
```diff
@@ -204,17 +206,22 @@ class RecurringStates extends React.Component<Props, State> {
...
+    const messages = this.state.messages ? <Messages key="state-messages" items={this.state.messages} /> : null;
+    const notification =
+      localizedMoment.userTimeZone !== localizedMoment.serverTimeZone ? (
+        <Messages
+          key="notification-messages"
+          items={[
+            {
+              severity: "warning",
+              text: t(
+                "The below times are displayed is the server time zone {0}. The scheduled time will be the server time.",
```
parlt91, line 218: Nit: is -> in

Etheryte: D'oh, thanks. 😄 Somehow I took "the server is in" and "times are in" and turned it into "the times is".

parlt91 (commented):
The PR looks good to me. Thanks for taking care of this and doing some cleanup on the way :)
Do you want to address the remaining to-do's in this PR?
If so I will have another look once you have them addressed.

Etheryte:
The to-dos here are problems that have been around for a bit and don't have a simple solution off the bat, except perhaps for the identifier one. Since I regularly go through misc to-dos when there's spare time I'd leave them for the future if that's fine by you.

web/html/src/manager/state/highstate-summary.tsx
```diff
@@ -53,6 +53,7 @@ export default function HighstateSummary({ minionId }) {
   }
 
   return (
+    // TODO: This identifier seems wrong, what's the correct identifier here?
```
Etheryte, line 56: Since I'm not really familiar with the logic here, I'm not sure what the identifier should be in this context. The items do not have a `state` property as far as I could see.

parlt91: Me neither. That's code that @cbbayburt recently added. He should be able to answer it.

parlt91 (approved):
No I am fine to address them later on. Approving ...

web/html/src/manager/state/highstate-summary.tsx
Etheryte: I created https://github.com/SUSE/spacewalk/issues/15856 to track this so we can go ahead with this merger.

PR #4215, opened 2021-09-15
CI: add client subrepos to cr projects

jordimassaguerpla:
## What does this PR change?

CI: This PR updates the systemsmanagement:Uyuni:Master:TEST:XX:CR repos with the client tools repos. This way, we can also freeze the client tools packages, the same way we do with the server and proxy ones.

This should fix the problem that the repos containing the client tools might have been updated and the metadata is not yet in sync.

This PR also adds an optimization. It copies the resulting RPMs into systemsmanagement:Uyuni:Master:TEST:XX:CR instead of linking the sources. This way, it is faster and we don't waste CPU time in obs.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed
- [X] **DONE**

## Test coverage
- No tests

I have tested this manually.

- [X] **DONE**

## Links

Partial fix for https://github.com/SUSE/spacewalk/issues/15846

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
+++ b/susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
@@ -13,6 +13,17 @@ usage_and_exit() {
 update_project() {
     tproject=${1}
     pproject=${2}
+    rname=${3}
+    echo "Updating ${tproject} from ${pproject} with repo ${rname}"
+    osc ls ${tproject} > /dev/null
+    if [ ${?} -ne 0 ];then
+        echo "Project ${tproject} does not exists. Creating ..."
+        echo ${new_project_config} |\
+            sed -e "s/__TEST_PROJECT_NAME__/${tproject}/g" |\
+            sed -e "s/__PARENT_PROJECT_NAME__/${pproject}/g" |\
+            sed -e "s/__REPO_NAME__/${rname}/g" > ${OUT}
+       osc meta prj --file=${OUT} -m "Created by ${0}" ${tproject}
+    fi
     for i in $(diff <( osc ls ${tproject} ) <( osc ls ${pproject} ) | grep ">" | cut -d" " -f2);do
         echo "Found new package ${i} in ${pproject}. Creating a link to ${tproject}"
         osc linkpac ${pproject} ${i} ${tproject}
@@ -32,11 +43,29 @@ if [ $# -eq 2 ];then
     test_parent_project=${2}
 fi
 
+parent_repo_name="openSUSE_Leap_15.3"
+
+new_project_config="
+    <project name=\"__TEST_PROJECT_NAME__\">
+  <title>Continuous Rebuild project for environment ${1}</title>
+  <description>This links to __PARENT_PROJECT_NAME RPMs. It will be disabled before each CI run, so there are no rebuilds during the CI build. Then, it will be enabled back to get the latest updates.</description>
+  <person userid=\"jordimassaguerpla\" role=\"maintainer\"/>
+  <person userid=\"juliogonzalezgil\" role=\"maintainer\"/>
+  <person userid=\"zypp-team\" role=\"maintainer\"/>
+  <repository name=\"__REPO_NAME__\">
+    <path project=\"__PARENT_PROJECT_NAME__\" repository=\"__REPO_NAME__\"/>
+    <arch>x86_64</arch>
+  </repository>
+</project>
+"
+
 osc ls ${parent_project} > /dev/null
 if [ ${?} -ne 0 ];then
     echo "Error. Does ${parent_project} exists?"
     exit -1
 fi
+
+OUT="$(mktemp)"
 
 update_project ${test_parent_project} ${parent_project} ${parent_repo_name}
     
--- a/susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
+++ b/susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
@@ -44,6 +44,22 @@ if [ $# -eq 2 ];then
 fi
 
 parent_repo_name="openSUSE_Leap_15.3"
+clients="Ubuntu1604-Uyuni-Client-Tools;xUbuntu_16.04 \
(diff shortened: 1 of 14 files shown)


jordimassaguerpla:
Please look at the commits one by one. I structured them for a better revision experience ;)
@juliogonzalez ^ and anyone else reviewing this

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
```diff
@@ -32,11 +43,29 @@ if [ $# -eq 2 ];then
...
 fi
 
+parent_repo_name="openSUSE_Leap_15.3"
+
+new_project_config="
+    <project name=\"__TEST_PROJECT_NAME__\">
+  <title>Continuous Rebuild project for environment ${1}</title>
+  <description>This links to __PARENT_PROJECT_NAME RPMs. It will be disabled before each CI run, so there are no rebuilds during the CI build. Then, it will be enabled back to get the latest updates.</description>
+  <person userid=\"jordimassaguerpla\" role=\"maintainer\"/>
+  <person userid=\"juliogonzalezgil\" role=\"maintainer\"/>
```
juliogonzalez, line 69: Ok, we'll need to change this later. We really need to start using a group.

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
```diff
@@ -1,15 +1,93 @@
 #!/bin/bash
 
-if [ $# -ne 1 ];then
-    echo "usage: $0 N"
+parent_project="systemsmanagement:Uyuni:Master"
+
+usage_and_exit() {
+    echo "usage: $0 N [test_project]"
```
juliogonzalez, line 6:
```suggestion
    echo "usage: ${0} N [test_project]"
```
Same for all other variables missing `{}`

jordimassaguerpla:
@srbarrios I know you are not expert on obs, but still could you review this idea? It is extending what we already had.

jordimassaguerpla:
This script is being called periodically from https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-Master-dev-sync-cr-projects/

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
```diff
@@ -1,15 +1,93 @@
...
+    echo "$0 will add new packages from ${parent_project} to test_project"
+    exit -1
+}
+
+update_project() {
+    tproject=${1}
+    pproject=${2}
+    rname=${3}
+    echo "Updating ${tproject} from ${pproject} with repo ${rname}"
+    osc ls ${tproject} > /dev/null
```
juliogonzalez, line 18: Since you are managing the error at the if, I guess you don't need `stderr`? 
```suggestion
    osc ls ${tproject} &> /dev/null
```

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
```diff
@@ -1,15 +1,93 @@
...
+    usage_and_exit
+fi
+
+test_parent_project="${parent_project}:TEST:${1}:CR"
+
+if [ $# -eq 2 ];then
+    test_parent_project=${2}
+fi
+
+parent_repo_name="openSUSE_Leap_15.3"
```
juliogonzalez, line 46: I'd place both variables right below the shebang and the comments, so if we ever need to change this, it's easy to quick and harder to break things.

juliogonzalez (approved):
I'd say it looks good. Only left some comments for some minor stuff.

However I think this is the first time I see such scripts, so maybe someone else should review as well.

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
```diff
@@ -1,15 +1,93 @@
 #!/bin/bash
```
juliogonzalez, line 1: Unrelated to the PR, but maybe we should consider exiting if there are errors.

jordimassaguerpla: I considered this, but then when I check if a project exists, I would need something like ```osc ls PROJECT || true``` and then I would need to set pipefails ... I think is clearer the way it is.

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
jordimassaguerpla: agree

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
juliogonzalez: You can disable exiting on errors when needed with `set` without any other aditional magic, and keep the "exit on error" for the general case.

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
jordimassaguerpla: both? parent_repo_name and which other one?

jordimassaguerpla:
@juliogonzalez I added error handling https://github.com/uyuni-project/uyuni/pull/4215/commits/a22c89d62980db9d7d9c9ab63568fb941f10ded6

susemanager-utils/testing/automation/update-obs-environment-cr-project.sh
juliogonzalez: Both. In the end this will change to `15.4` at some point :-)

jordimassaguerpla: done

jordimassaguerpla:
@juliogonzalez All your comments have been address. Can we merge this one or would you rather prefer to merge it when we unlock the branch?

juliogonzalez:
> @juliogonzalez All your comments have been address. Can we merge this one or would you rather prefer to merge it when we unlock the branch?

This doesn't affect the code or the package building, so we can merge it now.

PR #4221, opened 2021-09-16
CI: parallelize bootstrap clients

jordimassaguerpla:
## What does this PR change?

Instead of bootstrapping each client and waiting for it, let's first
bootstrap all of them and then wait for them to appear in the system
list.

This way, we are parallelizing the bootstrap process. Thus, the total
time should be smaller, meaning faster.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed
- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

N/A

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/init_clients/buildhost_bootstrap_end.feature
+++ b/testsuite/features/init_clients/buildhost_bootstrap_end.feature
@@ -2,21 +2,10 @@
 # Licensed under the terms of the MIT license.
 
 @buildhost
-Feature: Bootstrap a Salt build host via the GUI
+Feature: Bootstrap a Salt build host via the GUI end
 
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
-
-  Scenario: Bootstrap a SLES build host
-     When I follow the left menu "Systems > Bootstrapping"
-     Then I should see a "Bootstrap Minions" text
-     When I enter the hostname of "build_host" as "hostname"
-     And I enter "22" as "port"
-     And I enter "root" as "user"
-     And I enter "linux" as "password"
-     And I select the hostname of "proxy" from "proxies"
-     And I click on "Bootstrap"
-     And I wait until I see "Successfully bootstrapped host!" text
 
   Scenario: Check the new bootstrapped build host in System Overview page
     When I follow the left menu "Salt > Keys"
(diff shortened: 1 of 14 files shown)


jordimassaguerpla:
This is an idea that I had. I have not tested it but I wanted to know what QEng thing about it.

jordimassaguerpla:
Hi! This is nor urgent nor planned, so feel free to review it when  you have some spare time, no pressure on this one. Thanks.

srbarrios (changes requested):
I don't like that approach.
We should just run it using the cucumber:parallel command in the stage that requires it.

There is no need at all of splitting these features. Just use the rake task of cucumber:parallel where you need it.

In CI, we decided to don't parallelize at all for now... for several reasons like the reporting, but I hope that we can soon start enabling at least some parts like this.

jordimassaguerpla:
IIRC, we can't run the client tasks in parallel because the server keeps the information on what the user is in the menu, so we cannot be in different places in parallel ... do you know what I mean? Did I got it right? If so, we cannot run it in parallel, unless we had a separate user for each client.

Unless, we don't wait for the bootstrap to complete. The idea here is to basically split the bootstrap tests into two, one that does the bootstrap, and another one that does the "wait". This way , we can bootstrap all of them, and then do the wait:

"click on bootstrap sle"
"click on bootstrap centos"
"click on bootrsap ubuntu"
etc.
"wait for sle to appear in system list"
"wait for centos to appear in system list"
"wait for ...."
etc.

jordimassaguerpla:
The idea at the end is just a copy of the "fork and wait" pattern you could use when programming in parallel.

jordimassaguerpla:
I tried running this PR. I got this error for almost each client:

```Given I am authorized for the "Admin" section

Message:

    unexpected alert open: {Alert text : }```

Looks like there is an unexpected alert:

![image](https://user-images.githubusercontent.com/1148215/135980771-444b1e18-247f-4192-93ac-95d1189eefb0.png)

Running with this error took 5 minutes less of 30 minutes.

Why are we seeing this alert?

jordimassaguerpla:
image

jordimassaguerpla:
```All of your systems are actively checking into Uyuni at this time```

Might be because the different clients are bootstrapping .... which makes sense. Then, how can I make cucumber do not fail on this alert?

jordimassaguerpla:
I tried changing the default but did not worked https://github.com/uyuni-project/uyuni/pull/4221/commits/3c9b252be9b6690d019414fefde28130b38bb5aa .

In any case, this would only save 5 minutes, which is not much. Other optimizations saved 15 or 30 minutes.

Given is not a big gain, I won't push for this if QE does not like it (Oscar in this case). We can try again a different approach when tests are parallelized.

PR #4222, opened 2021-09-16
API call to accept Salt keys 

mbussolotto:
## What does this PR change?

Added some endpoint for salt key API:
- `accept` : API endpoint to accept minion keys
-  `reject` : API endpoint to reject minion keys
- `pendingList` :  API endpoint to list pending salt keys
- `acceptedList` : API endpoint to list accepted salt keys
- `rejectedList` : API endpoint to list rejected salt keys

Resolve minor issue in `delete` API
## GUI diff

No difference.

- [ ] **DONE**

## Documentation
- API documentation generated correctly

- [ ] **DONE**

## Test coverage
- No tests: Not Appilicable

- [ ] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/14189

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
@@ -50,12 +50,7 @@ public class SaltKeyHandler extends BaseHandler {
      * @xmlrpc.returntype #array_single("string", "Accepted salt key list")
      */
     public List<String> acceptedList(User loggedInUser) {
-        try {
-            ensureOrgAdmin(loggedInUser);
-        }
-        catch (PermissionException e) {
-            throw new PermissionCheckFailureException(e);
-        }
+        ensureOrgAdmin(loggedInUser);
         return saltKeyUtils.acceptedSaltKeyList(loggedInUser);
 
     }
@@ -71,12 +66,7 @@ public class SaltKeyHandler extends BaseHandler {
      * @xmlrpc.returntype #array_single("string", "Pending salt key list")
      */
     public List<String> pendingList(User loggedInUser) {
-        try {
-            ensureOrgAdmin(loggedInUser);
-        }
-        catch (PermissionException e) {
-            throw new PermissionCheckFailureException(e);
-        }
+        ensureOrgAdmin(loggedInUser);
         return saltKeyUtils.unacceptedSaltKeyList(loggedInUser);
     }
 
@@ -92,8 +82,8 @@ public class SaltKeyHandler extends BaseHandler {
      * @xmlrpc.returntype #return_int_success()
      */
     public int accept(User loggedInUser, String minionId) {
+        ensureOrgAdmin(loggedInUser);
         try {
-            ensureOrgAdmin(loggedInUser);
             saltKeyUtils.acceptSaltKey(loggedInUser, minionId);
         }
         catch (PermissionException e) {
@@ -118,8 +108,8 @@ public class SaltKeyHandler extends BaseHandler {
      * @xmlrpc.returntype #return_int_success()
      */
     public int genAccept(User loggedInUser, String minionId) {
+        ensureOrgAdmin(loggedInUser);
         try {
-            ensureOrgAdmin(loggedInUser);
             saltKeyUtils.genAcceptSaltKey(loggedInUser, minionId);
         }
         catch (PermissionException e) {
@@ -143,8 +133,8 @@ public class SaltKeyHandler extends BaseHandler {
      * @xmlrpc.returntype #return_int_success()
      */
     public int delete(User loggedInUser, String minionId) {
+        ensureOrgAdmin(loggedInUser);
         try {
-            ensureOrgAdmin(loggedInUser);
             saltKeyUtils.deleteSaltKey(loggedInUser, minionId);
         }
         catch (PermissionException e) {
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
@@ -47,7 +47,6 @@ public class SaltKeyHandler extends BaseHandler {
      *
      * @xmlrpc.doc List accepted salt keys
      * @xmlrpc.param #param("string", "loggedInUser")
-     * @xmlrpc.param #param("string", "minionId")
      * @xmlrpc.returntype #array_single("string", "Accepted salt key list")
      */
     public List<String> acceptedList(User loggedInUser) {
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
@@ -84,6 +84,20 @@ public class SaltKeyHandler extends BaseHandler {
     }
 
     /**
+     * API endpoint to list denied salt keys
+     * @param loggedInUser the user
+     * @return 1 on success
+     *
+     * @xmlrpc.doc List of denied salt keys
+     * @xmlrpc.param #param("string", "loggedInUser")
+     * @xmlrpc.returntype #array_single("string", "Denied salt key list")
+     */
+    public List<String> deniedList(User loggedInUser) {
+        ensureOrgAdmin(loggedInUser);
+        return saltKeyUtils.deniedSaltKeyList(loggedInUser);
+    }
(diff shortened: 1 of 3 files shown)


java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
```diff
@@ -37,25 +40,120 @@ public SaltKeyHandler(SaltKeyUtils saltKeyUtilsIn) {
...
     }
 
+    /**
+     * API endpoint to list accepted salt keys
+     * @param loggedInUser the user
+     * @return 1 on success
+     *
+     * @xmlrpc.doc List accepted salt keys
+     * @xmlrpc.param #param("string", "loggedInUser")
+     * @xmlrpc.param #param("string", "minionId")
```
mcalmer, line 50: This function does not have a parameter "minionId".

java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
```diff
@@ -37,25 +40,120 @@ public SaltKeyHandler(SaltKeyUtils saltKeyUtilsIn) {
...
+     * @xmlrpc.returntype #array_single("string", "Accepted salt key list")
+     */
+    public List<String> acceptedList(User loggedInUser) {
+        try {
+            ensureOrgAdmin(loggedInUser);
+        }
+        catch (PermissionException e) {
+            throw new PermissionCheckFailureException(e);
+        }
```
mcalmer, line 59: I think a try/catch is not needed here as ensureOrgAdmin() throw already a PermissionCheckFailureException

java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
```diff
@@ -37,25 +40,120 @@ public SaltKeyHandler(SaltKeyUtils saltKeyUtilsIn) {
...
+     * @xmlrpc.param #param("string", "loggedInUser")
+     * @xmlrpc.returntype #array_single("string", "Pending salt key list")
+     */
+    public List<String> pendingList(User loggedInUser) {
+        try {
+            ensureOrgAdmin(loggedInUser);
+        }
+        catch (PermissionException e) {
+            throw new PermissionCheckFailureException(e);
+        }
```
mcalmer, line 80: I think a try/catch is not needed here as ensureOrgAdmin() throw already a PermissionCheckFailureException

java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
```diff
@@ -37,25 +40,120 @@ public SaltKeyHandler(SaltKeyUtils saltKeyUtilsIn) {
...
+     * @return 1 on success
+     *
+     * @xmlrpc.doc Accept a minion key
+     * @xmlrpc.param #param("string", "loggedInUser")
+     * @xmlrpc.param #param("string", "minionId")
+     * @xmlrpc.returntype #return_int_success()
+     */
+    public int accept(User loggedInUser, String minionId) {
+        try {
+            ensureOrgAdmin(loggedInUser);
```
mcalmer, line 97: As above: This can live outside of the try/catch. Not sure if acceptSaltKey() can throw a PermissionException which needs to be converted. It might be possible when salt master cannot write some files.

java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
```diff
@@ -37,25 +40,120 @@ public SaltKeyHandler(SaltKeyUtils saltKeyUtilsIn) {
...
+            saltKeyUtils.genAcceptSaltKey(loggedInUser, minionId);
+        }
+        catch (PermissionException e) {
+            throw new PermissionCheckFailureException(e);
+        }
+        catch (IllegalArgumentException e) {
+            throw new UnsupportedOperationException(e.getMessage());
+        }
+        return 1;
+    }
```
mcalmer, line 133: I am not sure if we need an API call for "genAccept()".
This generate a pre-accepted key on the master.
We use this mechanism for bootstrapping, but there we read the key from the master and deploy the key to the minion. To make any sense about this in the API, we would need an API call which return the keys.

I would remove this function for now and add it maybe later when we have a use case.

java/code/src/com/redhat/rhn/frontend/xmlrpc/saltkey/SaltKeyHandler.java
```diff
@@ -37,25 +40,120 @@ public SaltKeyHandler(SaltKeyUtils saltKeyUtilsIn) {
...
      * @xmlrpc.doc Delete a minion key
-     * @xmlrpc.param #param("string", "sessionKey")
+     * @xmlrpc.param #param("string", "loggedInUser")
      * @xmlrpc.param #param("string", "minionId")
      * @xmlrpc.returntype #return_int_success()
      */
     public int delete(User loggedInUser, String minionId) {
-        ensureOrgAdmin(loggedInUser);
         try {
+            ensureOrgAdmin(loggedInUser);
```
mcalmer, line 148: Same as above. This can live outside of the try/catch. Not sure if acceptSaltKey() can throw a PermissionException which needs to be converted. It might be possible when salt master cannot write some files.

java/code/src/com/suse/manager/utils/SaltKeyUtils.java
```diff
@@ -81,8 +180,7 @@ public boolean deleteSaltKey(User user, String minionId) throws PermissionExcept
             });
         }
         else {
-            LOG.info(String.format("No key found for minionID [%s]", minionId));
-            return false;
+            throw new IllegalArgumentException("No key found for minionID [" + minionId + "]");
```
mcalmer, line 183: Please check, if this change has consequences at other places where this function is used.
I can see SaltServerActionService using it and I think at this place it was wanted to ignore a possible error.

mcalmer (changes requested):
There is also a "REJECT" operation for salt-keys. I think it would be good to have an endpoint for that as well.

java/code/src/com/suse/manager/utils/SaltKeyUtils.java
mbussolotto: I tried to go more in deep on the code. `deleteSaltKey` was used in `SaltServerActionService` and `MinionsAPI`: using `saltApi.deleteKey` instead of `saltKeyUtils.deleteSaltKey` , we should avoid problem with exception and remove at all `SaltKeyUtils` dependencies from `SaltServerActionService` .

java/spacewalk-java.changes
```diff
@@ -1,3 +1,5 @@
+- Add endpoints to saltkeys API
```
mcalmer, line 1: we had a discussion about enduser understandable changelogs. Maybe list the new endpoints ?

```suggestion
- Add new endpoints to saltkeys API: acceptedList, pendingList, rejectedList, deniedList, accept and reject
```

mcalmer (changes requested):

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
```diff
@@ -2715,7 +2710,7 @@ else if (action.get().getActionType().equals(ActionFactory.TYPE_KICKSTART_INITIA
                             KickstartAction ksAction = (KickstartAction) action.get();
                             if (!ksAction.getKickstartActionDetails().getUpgrade()) {
                                 // Delete salt key from master
-                                saltKeyUtils.deleteSaltKey(ksAction.getSchedulerUser(), minionId);
+                                saltApi.deleteKey(minionId);
```
lucidd, line 2713: if i understand this correctly the previous version took the users permissions into account. Whats the reason for removing it?

mbussolotto: I thought it was unintentionally used `saltKeyUtils` instead of `saltApi` , since all the other methods in `SaltServerActionService` used the last one. But yes, you are totally right :) . I also checked if user permission was checked before  calling `handleAction` and they are not, so I am going to remove this change. Thanks for  pointing that out!

java/code/src/com/suse/manager/utils/SaltKeyUtils.java
mbussolotto: revert the changes, in this way everything sholud be safe :)

mcalmer (approved):
LGTM

Issue #3406, opened 2021-03-12
CLM: implement "matcher: greater than" in "Package (NEVRA)" filter


paususe:
Currently we only have a "matcher: equals", which seems not enough if I want to deny any newer package.

In the case of SLES/openSUSE, this might not be too big of a problem because there are several options in "Patches" filters (e. g. "Patch issue date" filter supports "greater than" dates) but in the case of Debian and Ubuntu, it might be: we have no concept of patches, and even when we implement it, I'd say it will be related to security updates only, not newer packages.

wombelix:
Looks like already in Implementation in PR https://github.com/uyuni-project/uyuni/pull/4233

PR #4233, opened 2021-09-23
Add additional matchers to package (nevra) filter

parlt91:
## What does this PR change?

Add additional matchers (lower, lowereq, greater, greatereq) to CLM package (NEVRA) filter

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- No documentation needed: **TODO: Check docs if there are mentions about the package (NEVRA) matchers**

- [x] **DONE**

## Test coverage

- Unit tests were added

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/13617
Tracks https://github.com/uyuni-project/uyuni/issues/3406

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/contentmgmt/FilterCriteria.java
+++ b/java/code/src/com/redhat/rhn/domain/contentmgmt/FilterCriteria.java
@@ -71,7 +71,9 @@ public class FilterCriteria {
     static {
         validCombinations.add(Triple.of(PACKAGE, CONTAINS, "name"));
         validCombinations.add(Triple.of(PACKAGE, EQUALS, "nevr"));
+        validCombinations.add(Triple.of(PACKAGE, GREATEREQ, "nevr"));
         validCombinations.add(Triple.of(PACKAGE, EQUALS, "nevra"));
+        validCombinations.add(Triple.of(PACKAGE, GREATEREQ, "nevra"));
         validCombinations.add(Triple.of(PACKAGE, MATCHES, "name"));
         validCombinations.add(Triple.of(ERRATUM, EQUALS, "advisory_name"));
         validCombinations.add(Triple.of(ERRATUM, EQUALS, "advisory_type"));
--- a/java/code/src/com/redhat/rhn/domain/contentmgmt/FilterCriteria.java
+++ b/java/code/src/com/redhat/rhn/domain/contentmgmt/FilterCriteria.java
@@ -30,6 +30,8 @@ import static com.redhat.rhn.domain.contentmgmt.FilterCriteria.Matcher.EQUALS;
 import static com.redhat.rhn.domain.contentmgmt.FilterCriteria.Matcher.EXISTS;
 import static com.redhat.rhn.domain.contentmgmt.FilterCriteria.Matcher.GREATER;
 import static com.redhat.rhn.domain.contentmgmt.FilterCriteria.Matcher.GREATEREQ;
+import static com.redhat.rhn.domain.contentmgmt.FilterCriteria.Matcher.LOWER;
+import static com.redhat.rhn.domain.contentmgmt.FilterCriteria.Matcher.LOWEREQ;
 import static com.redhat.rhn.domain.contentmgmt.FilterCriteria.Matcher.MATCHES;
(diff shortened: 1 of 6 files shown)


java/code/src/com/redhat/rhn/domain/contentmgmt/FilterCriteria.java
```diff
@@ -69,7 +69,9 @@
     static {
         validCombinations.add(Triple.of(PACKAGE, CONTAINS, "name"));
         validCombinations.add(Triple.of(PACKAGE, EQUALS, "nevr"));
+        validCombinations.add(Triple.of(PACKAGE, GREATEREQ, "nevr"));
         validCombinations.add(Triple.of(PACKAGE, EQUALS, "nevra"));
+        validCombinations.add(Triple.of(PACKAGE, GREATEREQ, "nevra"));
```
mcalmer, line 74: Does it make sense to "only" implement "greater equal" ?
Why not just implement all variants in one shot?

parlt91: Yes that was my plan :)
I am just not satisfied with the way I do the comparison down below yet.
The way we currently save the value (nevra) in the db it's hard to exract the evr part so I might need to do some changes there (still thinking about solutions)

cbbayburt (commented):
Looks good :+1: 

I'm just wondering how it would behave if the NEVRA can't be parsed properly. With the existing "contains" filter, it's predictable. But I wonder if there would be some confusing results, especially with the comparison.

Unfortunately, we don't have any kind of validation/exception neither in the UI nor in the `PackageEvr` class.

Maybe we can start by adding some wrong parsing cases in the `PackageEvrTest` and see how it behaves?

web/html/src/manager/content-management/shared/business/filters.enum.ts
```diff
@@ -86,6 +96,11 @@ const filterMatchers: FilterMatcherEnumType = {
     text: t("later or equal"),
     longDescription: t("later or equal than"),
   },
+  GREATER: {
+    key: "greater",
+    text: t("later"),
+    longDescription: t("later than"),
```
hustodemon, line 102: Hm, looking at this, shouldn't this be "greater" and "greater than"? Same goes for the existing text above...

I think we only used this for the patch issue date so far, where "later" made sense, but now it'd be better to change it is more general "greater"/"greater than".

hustodemon (commented):
Left just one comment, I also think it would be nice to write some tests (in `ContentFilterTest.java`).

The code LGTM :+1: !

Etheryte (approved):
Frontend changes look good, thanks! 👍

java/code/src/com/redhat/rhn/domain/contentmgmt/PackageFilter.java
```diff
@@ -79,6 +83,18 @@ public boolean test(Package pack) {
         }
     }
 
+    private static String getEvr(String field, String value) {
+        if (field.equals("nevr")) {
+            return value.replaceAll("(.*)-(.*:)?(.*)-(.*)", "$2$3-$4");
+        }
+        else if (field.equals("nevra")) {
+            return value.replaceAll("(.*)-(.*:)?(.*)-(.*)\\.(.*)", "$2$3-$4");
+        }
```
hustodemon, line 114: One last-minute thought I got: does this play well with the Debian versions? I think it's worth checking it and adding it to the unit tests.

hustodemon (commented):
LGTM, but before my approving, please check and answer my only comment.

java/code/src/com/redhat/rhn/domain/contentmgmt/PackageFilter.java
parlt91: Added some test for debian versioning it seems to work fine.

hustodemon (approved):
LGTM :+1:

parlt91:
@cbbayburt Could you have another look on the PR?

cbbayburt (approved):

PR #4243, opened 2021-09-24
Use the apt-key in case of debian based machines to add the gpg keys

admd:
Port of https://github.com/SUSE/spacewalk/pull/15484


juliogonzalez (approved):
OK for now, but we will need to improve this soon, as `apt-key` is deprectated: https://www.linuxuprising.com/2021/01/apt-key-is-deprecated-how-to-add.html

PR #4258, opened 2021-09-27
return value for enabled can be boolean or string depening on the OS

mcalmer:
## What does this PR change?

On SUSE `pkg.list_repos()` return objects with enabled = boolean, while on RedHat enabled return a string which is "1" or "0".
In our states we need to handle both.

We cannot check:

```
data.get('enabled', True) in [True, '1']
```
as when "enabled" return the string '0', it would match the boolean True.
So we better do the negative test

```
data.get('enabled', True) not in [False, '0']
```

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Additional fix for https://github.com/uyuni-project/uyuni/pull/4247
Tracks https://github.com/SUSE/spacewalk/pull/15962

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/channels/disablelocalbrokenrepos.sls
+++ b/susemanager-utils/susemanager-sls/salt/channels/disablelocalbrokenrepos.sls
@@ -18,7 +18,7 @@ disable_repo_{{ repos_disabled.count }}:
 {% endif %}
 {% endfor %}
 {% else %}
-{% if data.get('enabled', '1') == '1' %}
+{% if data.get('enabled', True) in [True, '1'] %}
 {% set url = data.get('baseurl', 'file://').replace('$basearch', grains['osarch']).replace('$releasever', grains['osmajorrelease']|string) -%}
 {%- set repo_exists = (0 < salt['http.query'](url + '/repodata/repomd.xml', status=True, verify_ssl=True).get('status', 0) < 300) %}
 {% if not repo_exists %}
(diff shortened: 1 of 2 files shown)


vzhestkov (approved):
Looks good

susemanager-utils/susemanager-sls/salt/channels/disablelocalbrokenrepos.sls
```diff
@@ -18,7 +18,7 @@ disable_repo_{{ repos_disabled.count }}:
 {% endif %}
 {% endfor %}
 {% else %}
-{% if data.get('enabled', '1') == '1' %}
+{% if data.get('enabled', True) not in [False, '0'] %}
```
agraul, line 21: This does not need to be inverted:

```python
>>> data = {'enabled': '0'}
>>> data.get('enabled', True) in (True, '1')
False
```

`data.get('enabled', True)` returns either the string "0"/"1" for RHEL or `True`/`False` on SLE (or on RHEL if 'enabled' is not in `data`). Whatever is returned is not evaluated for truthiness (like what happens on `bool(foo)`), instead the check is if it is in the tuple (can be a list, a tuple is more idiomatic). This pattern is used to avoid chaining repeated conditions. Another way to write this would be `if data.get('enabled', '1') == '1' or data.get('enabled', True) is True`, the `in` condition is a bit cleaner.

mcalmer: Yes, i found out the same. I dropped the last commit and we are now back to the ond way.

PR #4264, opened 2021-09-28
Add context manager for rhnConfig

agraul:
## What does this PR change?

There is a common pattern when using rhnConfig that, when forgetting a step, can cause issues in unrelated parts later on. This pattern is

1) save current component
2) switch to another component
3) read config from this component
4) switch back to the old component

There can be unforeseen bugs when the last step is forgotten.

Luckily, we now have a one-line solution that makes it impossible to forget a step -- there is only one:

    with cfg_component("my_component") as CFG:
        print(CFG.my_value)

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: only changes implementation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

I think it's a better idea to test this in uyuni before we think about backporting.

- [ ] **DONE**

## Changelogs

No user impacting change -> no changelog entry needed

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/uyuni/common-libs/common/context_managers.py
+++ b/uyuni/common-libs/common/context_managers.py
@@ -1,0 +1,29 @@
+"""Collection of context managers for Uyuni."""
+from contextlib import contextmanager
+from spacewalk.common.rhnConfig import CFG, initCFG
+
+@contextmanager
+def cfg_component(component, root=None, filename=None):
+    """Context manager for rhnConfig.
+
+    :param comp: The configuration component to use in this context
+    :param root: Root directory location of configuration files, optional
+    :param filename: Configuration file, optional
+
+    There is a common pattern when using rhnConfig that consists of the following steps:
+    1. save current component: old = CFG.getComponent()
+    2. set CFG to another component: initCFG('my_component')
+    3. Read / Set configuration values
+    4. set CFG back to the previous component
+
+    This pattern can now be expressed using the ``with`` statement:
+
+    with cfg_component('my_component') as CFG:
+        print(CFG.my_value)
+    """
+    previous = CFG.getComponent()
+    initCFG(component=component, root=root, filename=filename)
+    try:
+        yield CFG
+    finally:
+        initCFG(previous)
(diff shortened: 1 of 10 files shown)


vzhestkov (approved):
Looks great

agraul:
BTW, I only moved `deb_src` and `yum_dnf_src` for now to limit the possible fallout. We can move other python modules whenever we want.

SchoolGuy (approved):
LGTM

SchoolGuy:
I believe that it would be more beneficial to port all things to the new context manager. With these two cases we don't really make use of the new context manager. I would vote to port all usages of the old pattern and then write additional tests to feel confident to do the change. A nice side effect of the full port from one pattern to the other would be that we don't have two different patterns in the same codebase. In the end it might just be forgotten that we have the new one when touched by someone else...

meaksh (approved):

meaksh:
Thanks for this @agraul! Definetely much better way now to deal with CFG namespaces +1:

As Enno already mentioned, IMHO it would make sense to port the rest of the old pattern to this new way. Of course, you can decide if this is done in this same PR or in different ones.

agraul:
`reposync.py` is harder to port to the context manager. I don't know how long it would take, it relies on global state a lot more than the modules I've ported so far.

SchoolGuy (approved):
I renew my faith in this PR also with the latest changes.

uyuni/common-libs/common/context_managers.py
```diff
@@ -0,0 +1,29 @@
+"""Collection of context managers for Uyuni."""
+from contextlib import contextmanager
+from spacewalk.common.rhnConfig import CFG, initCFG
+
+@contextmanager
+def cfg_component(component, root=None, filename=None):
+    """Context manager for rhnConfig.
```
mcalmer, line 7: I think you need to add this file to the "Makefile" to get it into the package and installed.

mcalmer (changes requested):
Check also the spec file if we use wildcards or name every single file in the filelist. Then you need to add it there as well.

uyuni/common-libs/common/context_managers.py
agraul: Good catch! I added it to the Makefile now and did a build locally. Now it's in the package :) The spec packages all files under `{pythonXroot}`, no need to change it.

mcalmer (approved):

PR #4268, opened 2021-09-29
Run Prometheus JMX exporter as Java agent

witekest:
## What does this PR change?

The PR disables JMX remote monitoring of Java components (Tomcat and Taskomatic) to reduce the surface of attack and not expose ports to unauthorized users.
The Prometheus JMX exporters are started as Java agents attached to Tomcat and Taskomatic JVMs accordingly.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #1184617

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


susemanager-utils/susemanager-sls/susemanager-sls.spec
```diff
@@ -141,28 +141,52 @@ base:
...
-  jmxremote_host_configured=$?
-  if [ $jmx_exporter_enabled -eq 0 ] && [ $jmxremote_host_configured -eq 1 ]; then
-    sed -ri "s/JAVA_OPTS=\"(.*)${JMXREMOTE_PORT}(.*)\"/JAVA_OPTS=\"\1${JMXREMOTE_HOST}localhost\ ${JMXREMOTE_PORT}\2\"/" ${2}
-    sed -ri "s/JAVA_OPTS=\"(.*)${RMI_SERVER_HOSTNAME}\S*(.*)\"/JAVA_OPTS=\"\1${RMI_SERVER_HOSTNAME}localhost\2\"/" ${2}
+  grep -q -- $JMXREMOTE_OPT ${2}
+  jmxremote_opt_configured=$?
+  grep -q -- $JMX_EXPORTER_JAVA_AGENT ${2}
+  jmx_exporter_java_agent_configured=$?
+  if [ $jmx_exporter_enabled -eq 0 ]; then
+    systemctl disable $service_name
```
moio, line 158: IIUC this disables the old JMX-based service if it exists.

I wonder why this was implemented as a post script instead of via Salt states, what pros and cons do you see?

java/code/src/com/redhat/rhn/taskomatic/TaskoXmlRpcHandler.java
```diff
@@ -644,7 +644,6 @@ private void isAlreadyScheduled(Integer orgId, String jobLabel)
      * @return true if JMX system props props
      */
     public boolean isJmxEnabled() {
-        return StringUtils.isNotEmpty(System.getProperty("com.sun.management.jmxremote.port")) &&
-                StringUtils.isNotEmpty(System.getProperty("java.rmi.server.hostname"));
+        return StringUtils.contains(System.getenv("JAVA_OPTS"), "jmx_prometheus_javaagent.jar");
```
moio, line 647: I do not really like the fact that `/etc/rhn/taskomatic.conf` is actually interpreted as a shell script.

I understand that is none of your fault, but I would still ask for a way not to build on top of that.

Can we check for the presence of the necessary classes via the class loader (eg. `Class.forName`) instead of using this environment variable?

susemanager-utils/susemanager-sls/susemanager-sls.spec
witekest: The post script and Salt states fulfill two different tasks.
Salt states allow interactively enabling and disabling monitoring setup and check for the current status.
The post script is meant to be run only once and migrate the setup to the new configuration in case monitoring has already been enabled in the old configuration.

PR #4283, opened 2021-10-01
fix mgr-sync refresh (bsc#1191222)

mbussolotto:
## What does this PR change?

**add description**

`mgr-sync refresh` goes in error because of an exception in  `updateChannel` . The exception happens just when a channel is removed from SCC Repositories. Changed the exception in a more clear log error.

## GUI diff

- [ ] **DONE**

## Documentation
- No documentation needed

## Test coverage

- [ ] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/16016
https://bugzilla.suse.com/show_bug.cgi?id=1191222

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
+++ b/java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
@@ -1953,7 +1953,8 @@ public class ContentSyncManager {
 
         Opt.consume(suseProductSCCRepositories.stream().findFirst(),
                 () -> {
-                    throw new ContentSyncException("No product tree entry found for label: '" + label + "'");
+                    log.warn("Cannot update channel with label " + label + ". Please run: ");
+                    log.warn("spacewalk-remove-channel -c " + label);
                 },
                 productrepo -> {
                     SUSEProduct product = productrepo.getProduct();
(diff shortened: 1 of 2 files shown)


mcalmer:
Would it be possible to remove the repo in this case? But we need to check if it makes sense in all cases to remove it.

mbussolotto:
In case we want to delete it if does not longer exist, I think we can just change here: https://github.com/uyuni-project/uyuni/blob/4b4a69d25167ddf847b825dc15bb616aabac9ea4/java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java#L1710

to something like:
```
        ChannelFactory.listVendorChannels().stream().forEach(c -> {
            if (channelsToCleanup.contains(c.getLabel())) {
                ChannelManager.disassociateChannelEntries(c);
            }
            else {
                updateChannel(c);
            }
        });
```
but yes, as already said, I am not sure if remove the channel is a good idea :)

java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
```diff
@@ -1953,7 +1953,8 @@ public static void updateChannel(Channel dbChannel) {
 
         Opt.consume(suseProductSCCRepositories.stream().findFirst(),
                 () -> {
-                    throw new ContentSyncException("No product tree entry found for label: '" + label + "'");
+                    log.warn("Cannot update channel with label " + label + ". Please run: ");
```
mcalmer, line 1956:
```suggestion
                    log.warn("Expired Vendor Channel with label '" + label + "' found. To remove it please run: ");
```

mcalmer (commented):
As a suggestion I just tried to find a better wording.

mcalmer (approved):

PR #4296, opened 2021-10-05
Log error if no base product

parlt91:
## What does this PR change?

Log an error if the server has installed products but no base product when trying to perform a Product Migration.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- No documentation needed: 

- [x] **DONE**

## Test coverage

- No tests: 

- [ ] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/15799

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/frontend/action/systems/SPMigrationAction.java
```diff
@@ -225,7 +225,13 @@ public ActionForward execute(ActionMapping actionMapping,
                 logger.debug("Installed products are 'unknown'");
                 return forward;
             }
-            installedProducts.ifPresent(pset -> logger.debug(pset.toString()));
+            installedProducts.ifPresent(pset -> {
+                logger.debug(pset.toString());
```
hustodemon, line 229: how does this solve the NPEx? looking at the exception in the Bugzilla, it seems it got caused by `pset.toString()`, but the new code still calls this method, even if the base product is `null`.

parlt91: I already created a previous PR to fix the NPE, but we tought it would be better to log a proper error in this case instead of just a debug log that displays an empty base product.

See `SUSEProductSet` line: 256
```
    public String toString() {
        StringBuilder builder = new StringBuilder();
        builder.append("[base: ");
        if (baseProduct != null) {
            builder.append(baseProduct.getFriendlyName());
        }
```

hustodemon: AHA! I was not aware of this. Makes sense. Thanks!

hustodemon (approved):
:+1:

java/code/src/com/redhat/rhn/frontend/action/systems/SPMigrationAction.java
parlt91: Yeah sorry for the confusion :smiley:

PR #4397, opened 2021-10-25
Make bootstrap script to use bash when called with different interpreter (bsc#1191656)

meaksh:
## What does this PR change?

This PR fixes an issue with the bootstrap script in Debian systems when the script is executed using `dash` instead of `bash`.

In Debian, `sh` is linked to `dash`, so it is really common (also mentioned in our docs) that the script is called like:

```console
# sh bootstrap.sh
```

After this PR, the boostrap script checks whether it was executed using `dash` or not, and will execute the script using `bash` in case it is needed.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **bugfix**

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


agraul:
I have a general question:  Why do we implement this hack to override the used shell instead of printing an error message? (When) we document that the script is supposed to be ran as `bash bootstrap.sh` and `./bootstrap.sh` also uses `bash`, isn't that already enough?

Issue #4378, opened 2021-10-20
Uyuni Bugreport - Uyuni 2021.09: SysV service '/etc/init.d/susemanager' lacks a native systemd unit file


stdevel:
# Additional Information

When running `dmesg`, a warning is shown:

```
[    5.836325] systemd-sysv-generator[523]: SysV service '/etc/init.d/susemanager' lacks a native systemd unit file. Automatically generating a unit file for compatibility. Please update package to include a native systemd unit file, in order to make it more safe and robust.
```

Anyhow, Uyuni is working fine in accordance with `spacewalk-service status`.

## Version of Uyuni Server and Proxy (if used)

```# zypper info Uyuni-Server-release
Loading repository data...
Reading installed packages...


Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : uyuni-server-stable
Name           : Uyuni-Server-release
Version        : 2021.09-141.3.uyuni1
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1,4 KiB
Installed      : Yes
Status         : up-to-date
Source package : Uyuni-Server-release-2021.09-141.3.uyuni1.src
Summary        : Uyuni Server
Description    : 
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.
```

This also applies to SUSE Manager 4.2:

```
# zypper info SUSE-Manager-*
Information for package SUSE-Manager-Server-release:
----------------------------------------------------
Repository : SLE-Product-SUSE-Manager-Server-4.2-Updates
Name : SUSE-Manager-Server-release
Version : 4.2-47.4.2
Arch : x86_64
Vendor : SUSE LLC <https://www.suse.com/>
Support Level : Level 3
Installed Size : 334.1 KiB
Installed : Yes
Status : up-to-date
Source package : SUSE-Manager-Server-release-4.2-47.4.2.src
Summary : SUSE Manager Server 4.2
Description :
SUSE Manager lets you efficiently manage physical, virtual,
and cloud-based Linux systems. It provides automated and cost-effective
configuration and software management, asset management, and system
provisioning.
```

## Details about issue

It is just a warning, but Uyuni works. Maybe the legacy SysV init script needs to be replaced as Uyuni isn't supposed to run on SLES 12 or openSUSE Leap 42.x and older?

[...]

Issue #2885, opened 2020-11-22
Create a command line wizard to create the setup_env.sh file [RHEL8]


sbluhm:
# Additional Information

For the installation and migration of Uyuni, a setup_env.sh file with configuration options is required. Currently, this file is created by a wizard using yast2. To remove the dependency to yast2, a simple standalone wizard should be created as alternative. This will allow non-SUSE distributions.

### Current installation Process:
yast2 --> /usr/lib/susemanager/bin/mgr-setup --> spacewalk-setup
### Suggested Installation Process:
< new wizard > --> /usr/lib/susemanager/bin/mgr-setup --> spacewalk-setup
### Alternative Installation Process:
/usr/lib/susemanager/bin/mgr-setup --> ( no "/root/setup_env.sh" ? run < new wizard > ) --> spacewalk-setup

## Version of Uyuni Server
Uyuni-Server-release: 2020.11 and master.

## Details about issue
Current yast2 wizard takes the wizard from the folder/files:
susemanager/yast/susemanager_*.rb
and writes the output to /root/setup_env.sh

Attached are some text copies of (some of) the screens.
[setupscript.txt](https://github.com/uyuni-project/uyuni/files/5579270/setupscript.txt)

## Additional requirements
On LEAP systems, a message is displayed on the hardware console before login informing about the installation status. This message should be updated/made compliant for this new wizard. (See #4402 )

sbluhm:
mgr-setup already asks for all the required missing variables. So technically, this issue can be closed.

Can I add the following enhancements in **mgr-setup** when setup_env.sh is not available?
- [ ] Additional wizard function to chek for disk space, hostname, etc
- [ ] More descriptive details in the existing input routine (https://github.com/uyuni-project/uyuni/blob/8774e57c4164bed9bda6cfd7f505576d0b991d7c/susemanager/bin/mgr-setup#L736).
- [ ] Should the item above maybe be added into a new function?
- [ ] Final question: Save answers as setup_env.sh and then option to continue installing or quit.

sbluhm:
Closing as not required anymore.

PR #4402, opened 2021-10-26
Remove deprecated sysv init and install issue instead

aaannz:
## What does this PR change?

For some crude status information (Uyuni configured or not) we still used old sysv init script. This caused some warnings by systemd. This PR replaces old init script by populating issue.d directories.

`issue` is similar to `motd` only it is printed before user is logged in, which our init script was doing.

Included are also small changes - replace `%{_prefix}/share` by equivalent `%{_datadir}` and remove atftp configuration from spec as we do not use atftp.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: User visible changes, but nothing actionable for docs.

- [X] **DONE**

## Test coverage
- No tests: No functionality change

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/16189, https://github.com/uyuni-project/uyuni/issues/4378

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager/susemanager.spec
+++ b/susemanager/susemanager.spec
@@ -83,6 +83,8 @@ Requires(pre):  %fillup_prereq %insserv_prereq tftp postgresql-init
 Requires(preun):%fillup_prereq %insserv_prereq tftp postgresql-init
 Requires(post): user(%{apache_user})
 Requires:       yast2-users
+BuildRequires:  issue-generator
+Requires:       issue-generator
 %endif
 Requires(pre):  salt
 Requires:       cobbler


sbluhm:
I tested an installation and got no errors. All services running. So OK from my side.

aaannz:
> I tested an installation and got no errors. All services running. So OK from my side.

Are you still able to see info about Uyuni state before login?

sbluhm:
Can you please elaborate on what you mean by the Uyuni state (maybe example text/url/location)? I didn't spot anything odd. I never noticed any state information on the login screen.

aaannz:
The init script I am removing in this PR had no other job than to display info to either call `yast susemanager_setup` or that uyuni is running at $HOSTNAME. It replaces it by using issue so the information should be printed on tty logging screen.
Now on SUSE there is issue-generator which can read issues from `/usr/lib/issue.d` but this tool is SUSE specific and I don't know it Alma or other Linux distros read that and what they do with it. AFAIK agetty does not read that directory.

But IMO even though this was fun exercise, I think best would be just to remove init script without any replacement for this info. It is all in documentation anyway.

sbluhm:
On running `/usr/lib/susemanager/bin/mgr-setup`, I get
...
```
* Restarting services.
Installation complete.  
Visit https://uyuni.example.com to create the Spacewalk administrator account.
INFO: Database configuration has been changed.
```
...
There is no change to before your change. I guess the product name needs to be adapted here at some point.

```
# ll /usr/lib/issue.d
lrwxrwxrwx. 1 root root 44  1. Nov 17:49 99-uyuni.conf -> /usr/share/rhn/config-defaults/99-uyuni.conf
```
The contents of `99-uyuni.conf` would not apply to Enterprise Linux (there is no yast2).
Some feedback: You might want to elaborate where the user should login (web server, shell,..)

`/usr/share/rhn/config-defaults/99-uyuni-installed.conf` is not linked at all in /usr/lib/issue.d.

when I log in:
```
$ ssh root@uyuni
root@uyuni's password: 
You can access Uyuni via https://uyuni.example.com
Last login: Mon Nov  1 17:40:18 2021 from 10.12.13.154

```
So I guess all works as designed?

aaannz:
> ```
> # ll /usr/lib/issue.d
> lrwxrwxrwx. 1 root root 44  1. Nov 17:49 99-uyuni.conf -> /usr/share/rhn/config-defaults/99-uyuni.conf
> ```
> 
> The contents of `99-uyuni.conf` would not apply to Enterprise Linux (there is no yast2). Some feedback: You might want to elaborate where the user should login (web server, shell,..)

That is a good point. Prior to making this PR I did not know Uyuni can run on different distro. This adds more complexity to the spec file. Definitely doable, but do we really need this info to be presented in this way?


> 
> `/usr/share/rhn/config-defaults/99-uyuni-installed.conf` is not linked at all in /usr/lib/issue.d.

This is supposed to be linked by updated mgr-setup (it links it to /etc/issue.d so it works on read only roots with etc overlay. /etc/issue.d has priority over files from /usr/lib/issue.d with same name).
rpm post upgrade also relinks /usr/lib/issue.d/ to uyuni-installed.

> 
> when I log into the shell:
> 
> ```
> $ ssh root@uyuni
> root@uyuni's password: 
> You can access Uyuni via https://uyuni.example.com
> Last login: Mon Nov  1 17:40:18 2021 from 10.12.13.154
> ```
> 
> So I guess all works as designed?

This is not from changed mgr-setup from this PR, but the original one. Original one stored this info into `motd` which is displayed after login. The service I removed was used specifically to print info before user login.


Note that similarly to the init script, `issue` texts are visible only on console ttys and not over ssh. I can switch to `motd` to be visible everywhere after login. I don't know the design decision for using init script to have this info visible before login.

sbluhm:
$ ssh root@uyuni
root@uyuni's password: 
Kickstarted on 2021-10-10
Last login: Mon Nov  1 19:25:22 2021 from 10.12.13.154

when /usr/lib/susemanager/bin/mgr-setup was not run.

sbluhm:
> Note that similarly to the init script, issue texts are visible only on console ttys and not over ssh. 

I see, then I tested wrong but start to understand the change. Will test more later.

sbluhm:
After understanding more on the change, I retested. There are no Uyuni specific messages displayed at the console before and after mgr-setup (probably as expected).

My thoughts on this one:

1. The message mentioning "Uyuni" is pretty nice as you can see the purpose of the server when you access it.

2. Message "Run yast" is not valid for Enterprise Linux. And we currently have no alternative apart from "Create a config file yourself". So I will add a similar feature to [Create a command line wizard to create the setup_env.sh file](https://github.com/uyuni-project/uyuni/issues/2885 ).

3. Message "Uyuni is running on HOSTNAME" is amazing for servers with a funny port as users/I don't remember the ports a server is accessible with. Luckily/thankfully Uyuni has a default 80/443 so the only benefit I see here is the information as mentioned in 1. above.

4. It might be a security concern if the purpose/status/content of a server are displayed before login as it gives a potential attacker a lot of information. (On the contrary an Uyuni server is heavily broadcasted across the network so a competent attacker would not be distracted by a missing message not saying "This really is the Uyuni server, you are at the right place").

As the installation still works, I would still say this PR is OK from Enterprise Linux side.

juliogonzalez:
> After understanding more on the change, I retested. There are no Uyuni specific messages displayed at the console before and after mgr-setup (probably as expected).
> 
> My thoughts on this one:

> 3. Message "Uyuni is running on HOSTNAME" is amazing for servers with a funny port as users/I don't remember the ports a server is accessible with. Luckily/thankfully Uyuni has a default 80/443 so the only benefit I see here is the information as mentioned in 1. above.
> 4. It might be a security concern if the purpose/status/content of a server are displayed before login as it gives a potential attacker a lot of information. (On the contrary an Uyuni server is heavily broadcasted across the network so a competent attacker would not be distracted by a missing message not saying "This really is the Uyuni server, you are at the right place").

That's a valid concern. That message could move from `/etc/issue` to `/etc/motd`.

aaannz:
This is unnecessary complicated. At the end decided to remove the service completely. (https://github.com/uyuni-project/uyuni/pull/5801)

PR #4404, opened 2021-10-26
NPE in SaltUtils.handleImageInspectData (bsc#1191597, bsc#1192150)

mbussolotto:
## What does this PR change?

The NPE happens when an image profile and/or image store is deleted during building and/or inspecting an image ( same root cause of https://github.com/SUSE/spacewalk/issues/16287 ). 

So these changes is:
* do not raise any NPE but print a correct log
* do not cascade when a image build or image store is deleted, but set value to null
* change the code to deal with null profile ID and store ID (hibernate does not allow null primitive type, so they were change in Java class)
* return empty instead of null json to the UI

## GUI diff

N/A

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Fix https://github.com/SUSE/spacewalk/issues/16240
Fix https://github.com/SUSE/spacewalk/issues/16287
- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


testsuite/features/secondary/buildhost_docker_build_image.feature
```diff
@@ -22,14 +22,17 @@ Feature: Build container images
...
-  Scenario: Create a simple real image profile without activation key
-    When I follow the left menu "Images > Profiles"
-    And I follow "Create"
-    And I enter "suse_real_simple" as "label"
-    And I select "galaxy-registry" from "imageStore"
-    And I enter "Docker/serverhost" relative to profiles as "path"
-    And I click on "create-btn"
+    Then I wait until I see "suse_simple" text
+     
+    #TODO: currently removed because build Docker/serverhost is failing
```
mbussolotto, line 27: `suse_real_simple` was never created because `docker build` command for `Docker/serverhost` fails. In this case any error was raised because there are no check

testsuite/features/secondary/buildhost_docker_build_image.feature
```diff
@@ -57,22 +62,34 @@ Feature: Build container images
...
+    And I wait until I see "Latest_key-activation1" text
+    And I refresh the page
+    And  I wait at most 660 seconds until table row for "Latest_simple" should contain "1"
+    And  I wait at most 660 seconds until table row for "Latest_key-activation1" should contain "1"
 
   Scenario: Delete image via XML-RPC calls
     When I delete the image "suse_key" with version "Latest_key-activation1" via XML-RPC calls
     And I delete the image "suse_simple" with version "Latest_simple" via XML-RPC calls
-    Then the image "suse_simple" with version "Latest_key-activation1" doesn't exist via XML-RPC calls
+    Then the image "suse_key" with version "Latest_key-activation1" doesn't exist via XML-RPC calls
```
mbussolotto, line 82: in this case we were checking the wrong image. It didn't fails because it does not exist

testsuite/features/secondary/buildhost_docker_build_image.feature
```diff
@@ -81,6 +98,8 @@ Feature: Build container images
     And I select the hostname of "build_host" from "buildHostId"
     And I click on "submit-btn"
     Then I wait until I see "GUI_BUILT_IMAGE" text
```
mbussolotto, line 100: this line is shown also when the image is still compiling. We need to wait "GUI_BUILT_IMAGE" should contain "1"

testsuite/features/secondary/buildhost_docker_build_image.feature
```diff
@@ -90,22 +109,34 @@ Feature: Build container images
     And I select the hostname of "build_host" from "buildHostId"
     And I click on "submit-btn"
     Then I wait until I see "GUI_DOCKERADMIN" text
```
mbussolotto, line 111: Same here, this line is shown also during the build phase.

java/spacewalk-java.changes
```diff
@@ -1,3 +1,6 @@
+  * FIX ISE and NPE when an image profile / store is deleted
```
mcalmer, line 1: I think we need to start with a dash "-" instead of a start "*".

mcalmer: I am also not sure if a customer understand "ISE" and "NPE". Maybe are more generic and say just "error".

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1157,62 +1157,61 @@ private void handleImageBuildData(ServerAction serverAction, JsonElement jsonRes
...
+            ImageInspectAction iAction = ActionManager.scheduleImageInspect(
+                    action.getSchedulerUser(),
+                    action.getServerActions()
+                            .stream()
+                            .map(ServerAction::getServerId)
+                            .collect(Collectors.toList()),
+                    Optional.of(action.getId()),
+                    details.get().getVersion(),
+                    profileOpt.map(ImageProfile::getLabel).orElse(null),
+                    profileOpt.map(ImageProfile::getTargetStore).orElse(null),
```
mackdk, line 1201: This might generate another NPE because from what I see `ActionManager.scheduleImageInspect()` [cannot handle a null `store`](https://github.com/mbussolotto/uyuni/blob/25e111d48108ea2b8c6a25cd81b4dad9a9c29143/java/code/src/com/redhat/rhn/manager/action/ActionManager.java#L2331). 

BTW can `profileOpt` actually be empty if we already checked that we have a valid `imageProfileId`?

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1157,62 +1157,61 @@ private void handleImageBuildData(ServerAction serverAction, JsonElement jsonRes
...
+                    Date.from(Instant.now())
+            );
+            try {
+                TASKOMATIC_API.scheduleActionExecution(iAction);
+            }
+            catch (TaskomaticApiException e) {
+                LOG.error("Could not schedule image inspection");
```
mackdk, line 1208: Why not simply `LOG.error("Could not schedule image inspection", e);`

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
```diff
@@ -337,21 +337,24 @@ else if (ActionFactory.TYPE_APPLY_STATES.equals(actionType)) {
         }
         else if (ActionFactory.TYPE_IMAGE_INSPECT.equals(actionType)) {
             ImageInspectAction iia = (ImageInspectAction) actionIn;
-            ImageInspectActionDetails details = iia.getDetails();
+            Optional<ImageInspectActionDetails> details = Optional.ofNullable(iia.getDetails());
+            if (details.isEmpty()) {
+                return new HashMap<>();
```
mackdk, line 342: Any reason why you are creating a new instance instead of using `Collections.emptyMap()` like you do in the  `orElseGet()`  line below? This inconsistency may lead to an issue if the caller has to change the collection since `Collections.emptyMap()` is immutable.

mcalmer (approved):

mackdk (commented):
I have only one concern with the call to `ActionManager.scheduleImageInspect()` with a possibly null value, the rest of the points are minor/open for discussion.

java/code/src/com/suse/manager/utils/SaltUtils.java
mbussolotto: agree

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1157,62 +1157,61 @@ private void handleImageBuildData(ServerAction serverAction, JsonElement jsonRes
...
-                catch (TaskomaticApiException e) {
-                    LOG.error("Could not schedule image inspection");
-                    LOG.error(e);
-                }
-
-                infoOpt.ifPresent(info -> {
-                    info.setRevisionNumber(info.getRevisionNumber() + 1);
-                    info.setInspectAction(iAction);
-                    ImageInfoFactory.save(info);
+            if (details.isEmpty()) {
```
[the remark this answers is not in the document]
mbussolotto: make sense. My first implementation had Optional in ImageBuildAction class (and probably it would be the best choice), but then due to problem with hibernate, I remove it but leave it here. I don't have a strong opinion for that, so it's fine for me to avoid wrapping

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1157,62 +1157,61 @@ private void handleImageBuildData(ServerAction serverAction, JsonElement jsonRes
...
+            if (imageProfileId.isEmpty()) { // It happens when the image profile is deleted during a build action
+                LOG.warn("Image Profile ID not found while performing: "  +
+                        action.getName() + " in handleImageBuildData");
+                return;
+            }
+            Optional<ImageProfile> profileOpt =
+                    ImageProfileFactory.lookupById(imageProfileId.get());
+            profileOpt.ifPresent(p -> p.asKiwiProfile().ifPresent(kiwiProfile -> {
+                serverAction.getServer().asMinionServer().ifPresent(minionServer -> {
+                    // Download the built Kiwi image to SUSE Manager server
```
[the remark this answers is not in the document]
mbussolotto: I agree that we should improve readability (in particular on that part) but the downside of increasing columns is using diff tool. E.g. I use a lot `git difftool -d master` with meld, and a single side is ~115 ...with more column I'm sure I ll lose something :smile: . So IMHO I would keep it at 120.

java/code/src/com/suse/manager/utils/SaltUtils.java
mbussolotto: I think `profileOpt` could not be empty.. but all this part was present before my changes so I m not confident with that. 
Anyway, I guess you are proposing to remove `Optional` wrapping from `profileOpt`, right? Or do you think it is safer a simple and dirty (but really effective) check for `store==null` at the beginning of `scheduleImageInspect` ?

mackdk: > I guess you are proposing to remove Optional wrapping from profileOpt, right? Or do you think it is safer a simple and dirty (but really effective) check for store==null at the beginning of scheduleImageInspect ?

My functional knowledge is limited here I fear :sweat_smile: Does it make sense to schedule an image inspect action if the store is null? If not, then we should just skip this part if the optional is empty (maybe with a warning). Otherwise, we should deal with the null inside the method and create the action correctly.

mcalmer: You cannot schedule an image inspect without the store to be available.
If profile or store are gone, the action details have just informational character. It is just historical data, you cannot do anything with it.

mbussolotto: now it should look better

mackdk (approved):

cbbayburt (approved):
LGTM :+1: 
The branch just needs a proper rebase to fix the commit history mess

PR #4427, opened 2021-11-03
Unify pkgset beacon to make it working with all plugins

vzhestkov:
## What does this PR change?

Unify the `pkgset` beacon to make it working with DNF notify plugin implemented with https://github.com/openSUSE/salt/pull/450

As the `dnfnotify` plugin placement is dependant on the Python version used by DNF there is no reason to check for the presence of `dnfnotify` plugin file, but it's better to check the cookie file.

Additionally `pkgset` beacon is storing the cookie in `salt.cache` to make it possible to fire the pakage list update event in case if changes was made while the minion was stopped.

## GUI diff

No difference.

## Re-run a test


susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -1,44 +1,32 @@
...
 def __virtual__():
-    global COOKIE_PATH
-
-    for plug, cookie in PKG_PLUGINS.items():
-        if os.path.exists(plug):
-            COOKIE_PATH = cookie
-            break
-
-    return COOKIE_PATH and __virtualname__ or False
+    return __virtualname__
```
agraul, line 25: I think it still makes sense to only load the plugin if it can work. What about checking for the cookie files here? We could use `COOKIE_PATH` to store the correct path and not iterate over the possible locations in `beacon`. Or, at least, do a one time check that at least one cookie exists, e.g. `any(os.path.exists(cookie) for cookie in PKGSET_COOKIES)`

vzhestkov: The problem is that the cookie file could not exists yet on loading the beacon it could appear some time after on the moment of modifying installed packages on the system. If we stop loading the beacon with such check we will loose all of the packages notifications until next restart of salt-minion

vzhestkov:
https://github.com/uyuni-project/uyuni/pull/4427/commits/556c0a9c949d9b1f86265b2ca053507aaeec0402 fixes bsc#1193045

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -1,45 +1,38 @@
...
 
-    for plug, cookie in PKG_PLUGINS.items():
-        if os.path.exists(plug):
-            COOKIE_PATH = cookie
-            break
 
-    return COOKIE_PATH and __virtualname__ or False
+def validate(config):
+    return True, "There is nothing to validate"
```
agraul, line 29: Are you sure that we need this? https://docs.saltproject.io/en/latest/topics/beacons/#writing-beacon-plugins says this function is optional.

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -49,20 +42,19 @@ def beacon(config):
...
             ck_data = ck_file.read().strip()
             if __virtualname__ not in __context__:
                 __context__[__virtualname__] = ck_data
             if __context__[__virtualname__] != ck_data:
-                ret.append({
-                    'tag': 'changed'
-                })
+                ret.append({"tag": "changed"})
                 __context__[__virtualname__] = ck_data
+                break
```
agraul, line 58: Is it correct to `break` only in one of the two branches?

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
vzhestkov: It's there to reduce the noise in DEBUG logging. Otherwise it prings the debug message that the function is missing in the module.

susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
```diff
@@ -12,10 +12,7 @@ def test_virtual():
     '''
     Test virtual function.
     '''
-    with patch('os.path.exists', MagicMock(return_value=False)):
-        assert pkgset.__virtual__() != pkgset.__virtualname__
-    with patch('os.path.exists', MagicMock(return_value=True)):
-        assert pkgset.__virtual__() == pkgset.__virtualname__
+    assert pkgset.__virtual__() == pkgset.__virtualname__
```
agraul, line 15: I think this test is meaningless, I would remove it.

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
vzhestkov: This break is preventing multiple events sending in case if salt-minion and venv-salt-minion are installed in the same time. Cookie files are in different locations for this case.

susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
vzhestkov: Oh yes, now it's

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
meaksh: I think the problem is that, even if the function is optional, if it is not present, then salt will spam logs messages every 10 seconds, of course only when `log_level: debug` even so, this is unfortunate as I will contaminate all debug logs with this repeated message.

susemanager-utils/susemanager-sls/src/tests/test_beacon_pkgset.py
vzhestkov: Removed with https://github.com/uyuni-project/uyuni/pull/4427/commits/9fe93cb39738efc3d559b031d24622b32eabfe20

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
agraul: Could you please add a comment in the source code explaining that this is the reason for including this function?

vzhestkov: Appended with https://github.com/uyuni-project/uyuni/pull/4427/commits/5d168f6c5826f72f39b639bb10166c994d42b965

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
vzhestkov: No, it's better to prevent breaking in the condition above. As the system could have two cookie files on the system at the same time, but one for some reason was not updated, so it's safer to check the rest and fire the event if at least one was changed.

agraul: I still don't get the difference between `__virtualname__ not in __context__` and `__context__[__virtualname__] != ck_data` when it comes to "stop on this cookie file or go through the rest".

If a system has 2 of these cookie files and why do we want to insert into both, but only update the first?

vzhestkov: So this part:
```
            if __virtualname__ not in __context__:
                __context__[__virtualname__] = ck_data
```
is storing the cookie data in the context, but this time we can read the other data from the other cookie file which could be different from the date we stored in this section and we will trigger the event with:
```
            if __context__[__virtualname__] != ck_data:
                ret.append({"tag": "changed"})
                __context__[__virtualname__] = ck_data
                break
```
but in the other iteration, so we actually using the cookie data from both possible files, but firing only one single event. The described case is not really common, mostly one cookie file is present on the system, but in case of using Salt Bundle with the original salt minion it's better to prevent such situation, if not we can lost at least one package change from the system.

juliogonzalez:
Not merging, as it seems there are no approvals yet. I am removing the `merge-candidate` label, as that should be added only when the PR is ready to be merged.

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -1,45 +1,49 @@
...
-        ("/usr/share/yum-plugins/venv-yumnotify.py",     "/var/cache/venv-salt-minion/rpmdb.cookie"),
-        ("/usr/bin/venv-dpkgnotify",                     "/var/cache/venv-salt-minion/dpkg.cookie"),
-        ("/usr/lib/zypp/plugins/commit/zyppnotify",      "/var/cache/salt/minion/rpmdb.cookie"),
-        ("/usr/share/yum-plugins/yumnotify.py",          "/var/cache/salt/minion/rpmdb.cookie"),
-        ("/usr/bin/dpkgnotify",                          "/var/cache/salt/minion/dpkg.cookie")
-    ])
-COOKIE_PATH = None
+__virtualname__ = "pkgset"
+
+__opts__ = salt.config.minion_config('/etc/salt/minion')
```
meaksh, line 18: We need to take care of the `venv-salt-minion` case here right?

meaksh (commented):
Thanks for the latest changed @vzhestkov!

I added few comments to clarify :-)

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
vzhestkov: Oh, yes, correct. Will adjust it for it.

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -1,45 +1,49 @@
...
+
+CACHE = salt.cache.Cache(__opts__)
```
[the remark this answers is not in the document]
vzhestkov: It's stored in the minion's cache. It's not sending it to the master.

vzhestkov:
Failing `test_module_sumautil` in `susemanager_sls_unit_tests` is not related to the change.

vzhestkov:
Failing test should be fixed with https://github.com/uyuni-project/uyuni/pull/4597

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -1,45 +1,51 @@
...
-        ("/usr/bin/dpkgnotify",                          "/var/cache/salt/minion/dpkg.cookie")
-    ])
-COOKIE_PATH = None
+__virtualname__ = "pkgset"
+
+SALT_CONFIG_DIR = os.environ.get("SALT_CONFIG_DIR")
+
+__opts__ = salt.config.minion_config(
+    os.path.join("/etc/salt" if SALT_CONFIG_DIR is None else SALT_CONFIG_DIR, "minion")
+)
```
agraul, line 22: Small suggestions to remove the conditional expression in `os.path.join`. I haven't yet found a way to run `black` in GitHub PR reviews, it could be that `black` puts everything on a single line :black_heart: 
```suggestion
SALT_CONFIG_DIR = os.environ.get("SALT_CONFIG_DIR", "/etc/salt")

__opts__ = salt.config.minion_config(
    os.path.join(SALT_CONFIG_DIR, "minion")
)
```

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
agraul: Makes sense :+1:

agraul (commented):
I left a suggestion, nothing big though. It would be nice to have more test coverage for `beacon()`, e.g. the cache logic is not covered at all. Generally it looks good though.

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
vzhestkov: Agree, it's better :+1:

vzhestkov:
@agraul thanks for the review, I just extended the test for `beacon()`. I think it should cover most of possible cases now.

SupremeGit:
I believe this thread relates to fix for salt minion spamming system logs with this:
salt-minion[1414]: [WARNING ] Unable to process beacon pkgset

So, pasting in my comment from https://github.com/uyuni-project/uyuni/issues/4581

Do people have any idea how unacceptable this sort of trash is, in logs?

Here we have a few stupid salt packages, which have the supreme arrogance to assume they are so important they can spam system logs with the same garbage warning, on and on, relentlessly, every second....

Imagine if every piece of software on a system was this stupid... it would be game over.

And even when its just these salt packages, it obscures REAL messages, which increases risks of serious impact.

It is completely unacceptable.

Whoever is actually working on fixing this, thank you, and please for F#&*sake, get the fix out there, so we can deploy it.

vzhestkov:
Hi @SupremeGit. This PR is fixing the issue you mentioned also. You can take the latest source of `pkgset.py` from the PR and place it on the Uyuni Server to `/usr/share/susemanager/salt/_beacons/pkgset.py`.

SupremeGit:
> Hi @SupremeGit. This PR is fixing the issue you mentioned also. You can take the latest source of `pkgset.py` from the PR and place it on the Uyuni Server to `/usr/share/susemanager/salt/_beacons/pkgset.py`.

Aaah, nice.
I'll grab that on monday and get it onto servers asap. Many thanks.

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -49,20 +55,25 @@ def beacon(config):
...
+        with open(cookie_path) as ck_file:
             ck_data = ck_file.read().strip()
             if __virtualname__ not in __context__:
-                __context__[__virtualname__] = ck_data
+                cache_data = CACHE.fetch("beacon/pkgset", "cookie").get("data", None)
+                if cache_data:
+                    __context__[__virtualname__] = cache_data
+                else:
+                    __context__[__virtualname__] = ck_data
+                    CACHE.store("beacon/pkgset", "cookie", {"data": ck_data})
```
meaksh, line 72: I'm adding some comments as suggestions here that maybe help clarifying a bit the logic of the code.

```suggestion
                # After a minion restart, when this is running for first time, there is nothing in context yet
                # So, if there is any data in the cache, we put it in the context, if not we put the new data.
                # and update the data in the cache.
                cache_data = CACHE.fetch("beacon/pkgset", "cookie").get("data", None)
                if cache_data:
                    __context__[__virtualname__] = cache_data
                else:
                    __context__[__virtualname__] = ck_data
                    CACHE.store("beacon/pkgset", "cookie", {"data": ck_data})
```

susemanager-utils/susemanager-sls/src/beacons/pkgset.py
```diff
@@ -49,20 +55,25 @@ def beacon(config):
...
             if __context__[__virtualname__] != ck_data:
-                ret.append({
-                    'tag': 'changed'
-                })
+                ret.append({"tag": "changed"})
+                CACHE.store("beacon/pkgset", "cookie", {"data": ck_data})
                 __context__[__virtualname__] = ck_data
+                break
```
meaksh, line 77:
```suggestion
                # Now it's time to fire beacon event only if the new data is not yet
                # inside the context (meaning not proceesed), and then stop iterating
                ret.append({"tag": "changed"})
                CACHE.store("beacon/pkgset", "cookie", {"data": ck_data})
                __context__[__virtualname__] = ck_data
                break
```

meaksh (approved):
Added some suggestions of comments. Besides of that LGTM :+1: Thanks

sbluhm:
@vzhestkov AlmaLinux 8 stopped building on OBS. By the timestamps, I assume this merge is the culprit:


395 | February 10, 2022 14:22 | source change | failed | 2m 25s | cloud112:2 |  
-- | -- | -- | -- | -- | -- | --
394 | February 10, 2022 13:33 | source change | succeeded | 2m 1s | lamb15:8 |  

395 	February 10, 2022 14:22 	[source change](https://build.opensuse.org/package/rdiff/systemsmanagement:Uyuni:Master/susemanager-sls?orev=02b7f45eaae7fcafb4ea84220407a51b&rev=dba75f1670538f809c02058ff34e2be2) 	failed 	2m 25s 	cloud112:2 	
394 	February 10, 2022 13:33 	[source change](https://build.opensuse.org/package/rdiff/systemsmanagement:Uyuni:Master/susemanager-sls?orev=7092730f68258d799eb568c6434c1026&rev=02b7f45eaae7fcafb4ea84220407a51b) 	succeeded 	2m 1s 	lamb15:8

https://build.opensuse.org/packages/susemanager-sls/job_history/systemsmanagement:Uyuni:Master/AlmaLinux_8/x86_64

But I have to admit that I don't know why. Maybe a wrong test case?
```
[   95s] + umask 022
[   95s] + cd /home/abuild/rpmbuild/BUILD
[   95s] + cd susemanager-sls-git-387.d179fbc
[   95s] + cd test
[   95s] + py.test-3 test_pillar_suma_minion.py
[   95s] ============================= test session starts ==============================
[   95s] platform linux -- Python 3.6.8, pytest-3.4.2, py-1.5.3, pluggy-0.6.0
[   95s] rootdir: /home/abuild/rpmbuild/BUILD/susemanager-sls-git-387.d179fbc/test, inifile:
[   95s] collected 3 items
[   95s] 
[   95s] test_pillar_suma_minion.py ...                                           [100%]
[   95s] 
[   95s] =========================== 3 passed in 0.14 seconds ===========================
[   95s] + cd ../src/tests
[   95s] + py.test-3
[   96s] ============================= test session starts ==============================
[   96s] platform linux -- Python 3.6.8, pytest-3.4.2, py-1.5.3, pluggy-0.6.0
[   96s] rootdir: /home/abuild/rpmbuild/BUILD/susemanager-sls-git-387.d179fbc/src/tests, inifile:
[   96s] collected 119 items / 1 errors
[   96s] 
[   96s] ==================================== ERRORS ====================================
[   96s] ____________________ ERROR collecting test_beacon_pkgset.py ____________________
[   96s] test_beacon_pkgset.py:11: in <module>
[   96s]     from ..beacons import pkgset
[   96s] ../beacons/pkgset.py:27: in <module>
[   96s]     os.path.join(__opts__["cachedir"], "rpmdb.cookie"),
[   96s] /usr/lib64/python3.6/posixpath.py:80: in join
[   96s]     a = os.fspath(a)
[   96s] E   TypeError: expected str, bytes or os.PathLike object, not MagicMock
[   96s] !!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
[   96s] =========================== 1 error in 0.68 seconds ============================
[   97s] error: Bad exit status from /var/tmp/rpm-tmp.22bvqt (%check)
[   97s] 
[   97s] 
[   97s] RPM build errors:
[   97s]     Bad exit status from /var/tmp/rpm-tmp.22bvqt (%check)
[   97s] ### VM INTERACTION START ###
[   98s] Powering off.
```

vzhestkov:
@sbluhm , interesting, for some reason the test seems working different ways for Alma and Leap. Will try to fix the test to avoid it

PR #4457, opened 2021-11-10
Make use of the Uyuni DB roster for Salt SSH calls

renner:
## What does this PR change?

This patch should makes use of @vzhestkov 's Uyuni DB roster implementation instead of generating temporary roster files in the backend for each call to a client via Salt SSH.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage

- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


vzhestkov:
@renner  After mergin https://github.com/uyuni-project/uyuni/pull/4540 this PR has conflicts, could you please fix the conflicts?
Few more modifications also required (these modifications makes sense only for bootstrapping as it will use roster files):
- We need to add `ssh_pre_flight` parameter to the roster file, it should contain the path to the pre flight script, it's better to make it configurable with `rhn.conf`
- `ssh_pre_flight_args` also need to be appended to the roster file. The value must be specified as the entry point to get the bootstrap repo from - protocol://server:port depending on the server path server need to be replaced with the hostname of the proxy closest to the managed system, In case of ussing tunnel it's better to specify `localhost` and the port used for HTTP forwarding (on time of bootstrapping the system has no /etc/hosts entry pointing the server/proxy host to the localhost)

vzhestkov:
@rjmateus could you please take a look on the SQL queries in the uyuni roster module:
https://github.com/uyuni-project/uyuni/blob/53ec61121a2a2d2c1f93b7b28f9b88b086e931a3/susemanager-utils/susemanager-sls/modules/roster/uyuni.py#L202-L221
and
https://github.com/uyuni-project/uyuni/blob/53ec61121a2a2d2c1f93b7b28f9b88b086e931a3/susemanager-utils/susemanager-sls/modules/roster/uyuni.py#L242-L257

The first one is used for cache validation only. Any suggestions to improve/optimize the queries are welcome.
One of the possible suggestions from me is to add missing `created` and `modified` fields to `suseMinionInfo` (the same way as it is for the other tables) as it also make sense for us to check for the changes in this table.

vzhestkov:
@renner @mcalmer are we allowed to grant an access to the file as it's made in https://github.com/uyuni-project/uyuni/pull/4457/commits/5fbe477eff251ce9b01985afcbc96c3172e2eb67 ?
It's the simpliest way, but I'm not really sure from the security point of view. Otherwise we have to duplicate some of the values from /etc/rhn.conf to salt config.

spacewalk/config/spacewalk-config.spec
```diff
@@ -190,6 +193,9 @@ if [ -e /etc/sudoers.d/spacewalk.rpmsave ]; then
 fi
 rm -f /etc/sudoers.d/spacewalk.{rpmnew,rpmorig,rpmsave}
 
+setfacl -m user:salt:r-x /etc/rhn
+setfacl -m user:salt:r-- /etc/rhn/rhn.conf
```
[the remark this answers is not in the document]
vzhestkov: do we support the deployments on such filesystems?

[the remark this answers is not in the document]
mcalmer: We could think about changing the group to susemanager.
susemanager goup has salt, wwwrun and tomcat as members.
group www only wwwrung and tomcat.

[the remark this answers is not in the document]
vzhestkov: I think changing group to `susemanager` should work as well in this case, right.
Thanks @mcalmer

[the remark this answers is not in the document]
vzhestkov: @mcalmer, Is `spacewalk-config` package relevant not only for Uyuni Server? Am I right?
`susemanager` group is owned by `uyuni-base-server` package, so to properly assign the group ownership we need to add `uyuni-base-server` as a required package to `spacewalk-config`, but such requirement will make the mess on the systems which are not server.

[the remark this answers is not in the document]
mcalmer: spacewalk-config is a server only package. But the problem is, that /etc/rhn is owned by uyuni-base-common which is also used by other packages.
I think you should reach out to our architect to discuss what the best approach is.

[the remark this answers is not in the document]
mcalmer: Btw: think also about the "new installation" case. When you install the pattern, the file rhn.conf does not exist.
You will get an error and when it exist, the permissions are only set when you re-install this package.

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
```diff
@@ -0,0 +1,309 @@
...
+                      COUNT(S.id),
+                      EXTRACT(EPOCH FROM MAX(SP.modified)),
+                      COUNT(SP.proxy_server_id)
+               ) AS fp
+               FROM rhnServer AS S
+               INNER JOIN suseServerContactMethod AS SSCM ON
+                     (SSCM.id=S.contact_method_id)
+               INNER JOIN suseMinionInfo AS SMI ON
+                     (SMI.server_id=S.id)
+               LEFT JOIN rhnServerPath AS SP ON
```
rjmateus, line 239: since you are using left join, the rhnServerPath may not have a matching entry. This will lead to empty data in return.
I'm assuming that is what you want, and that case represents that minion is connected directly to a server, without any proxy.

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
```diff
@@ -0,0 +1,309 @@
...
+                    (SP.server_id=S.id)
+               WHERE S.contact_method_id IN (
+                         SELECT SSCM.id
+                         FROM suseServerContactMethod AS SCCM
+                         WHERE SSCM.label IN ('ssh-push', 'ssh-push-tunnel')
+                     )
+    """
+    h = _executeQuery(query)
+    if h is not None:
+        row = h.fetchone()
```
rjmateus, line 249: It may be a dummy question, but you are getting only the first element, right? 
If I got it right the query will return a list with all the minion which has contact method ssh-push or ssh-push-tunnel. Should you get all records?

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
```diff
@@ -0,0 +1,309 @@
...
+                      COUNT(SP.proxy_server_id)
+               ) AS fp
+               FROM rhnServer AS S
+               INNER JOIN suseServerContactMethod AS SSCM ON
+                     (SSCM.id=S.contact_method_id)
+               INNER JOIN suseMinionInfo AS SMI ON
+                     (SMI.server_id=S.id)
+               LEFT JOIN rhnServerPath AS SP ON
+                    (SP.server_id=S.id)
+               WHERE S.contact_method_id IN (
```
rjmateus, line 241: if you are joining with the suseServerContactMethod in the from clause, you can just filter it by label, the same way it is being done in the query down in the code, with `WHERE SSCM.label IN ('ssh-push', 'ssh-push-tunnel')`

rjmateus (commented):
some question about the queries

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
vzhestkov: yes, you are right. `rhnServerPath` could contain no records for the minion, but I selecting all the records in one query and there is additional logic in python to build proper structure.

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
vzhestkov: The result of the query fetched here is just one record to check the validity of the cache. It returns just a fingerprint calculated based on number of records in the tables and latest time of modification of the records in these tables

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
vzhestkov: It's done different way just because I don't need to get the actual values of `SSM.label` in this case, but only validate the cache. I can do it the same way in the query as in the next one, but according to the output of `explain ....`, the cost of the query is lower in this case, not dramatically, but it was an attempt to optimize it.

vzhestkov:
@rjmateus thanks for the review

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
rjmateus: But in this case, why are you joining in the from clause with the table `suseServerContactMethod`? I think table `suseMinionInfo` is also not used.
Do you think query can be simplefied to:
```
SELECT FORMAT('%s|%s|%s|%s',
                      EXTRACT(EPOCH FROM MAX(S.modified)),
                      COUNT(S.id),
                      EXTRACT(EPOCH FROM MAX(SP.modified)),
                      COUNT(SP.proxy_server_id)
               ) AS fp
               FROM rhnServer AS S
               LEFT JOIN rhnServerPath AS SP ON
                    (SP.server_id=S.id)
               WHERE S.contact_method_id IN (
                         SELECT SCCM.id
                         FROM suseServerContactMethod AS SCCM
                         WHERE ScCM.label IN ('ssh-push', 'ssh-push-tunnel')
                     );

```

vzhestkov: Then it's better to do this way:
```
SELECT FORMAT('%s|%s|%s|%s',
                      EXTRACT(EPOCH FROM MAX(S.modified)),
                      COUNT(S.id),
                      EXTRACT(EPOCH FROM MAX(SP.modified)),
                      COUNT(SP.proxy_server_id)
               ) AS fp
               FROM rhnServer AS S
               INNER JOIN suseMinionInfo AS SMI ON
                     (SMI.server_id=S.id)
               LEFT JOIN rhnServerPath AS SP ON
                    (SP.server_id=S.id)
               WHERE S.contact_method_id IN (
                         SELECT SSCM.id
                         FROM suseServerContactMethod AS SSCM
                         WHERE SSCM.label IN ('ssh-push', 'ssh-push-tunnel')
                     )
```

We don't need to drop `suseMinionInfo` from the query as it make sense only for salt systems, but on dropping this part it will invalidate the cache on changing traditional clients also.
And additionally I think it's better to extend `suseMinionInfo` by adding `created` and `modified` fields to make fingerprint sensitive for changing this table also.

rjmateus: I Agree with your analysis. the proposed query looks good to me

vzhestkov: I just extended it slightly by appending `created` and `modified` to `suseMinionInfo`.

vzhestkov:
This PR need to have https://github.com/openSUSE/salt/pull/475 merged and promoted first.

vzhestkov (approved):
I've tested all the parts. Everything works fine, except the part from salt side I mentioned the PR fixing it before.

agraul (changes requested):
I tried the uyuni roster module some time ago and it worked well. In this review I focused mostly on the code style. I would appreciate a bit more consistency and sticking to something like [PEP 8][https://pep8.org]. We don't yet have a style guide in place, but whatever we will come up will based on PEP 8 since that's the Python standard.

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
```diff
@@ -0,0 +1,309 @@
...
+            proxies = []
+        if row is None:
+            break
+        if row[4]:
+            proxies.append(row[4])
+        prow = row
+        row = h.fetchone()
+
+    cache.store("roster/uyuni", "minions", {"fp": cache_fp, "minions": ret})
+    log.trace("Uyuni DB roster:\n%s" % dump(ret))
```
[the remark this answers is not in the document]
vzhestkov: I hope I understood correct to make it the following way:
```
    if log.isEnabledFor(logging.TRACE):
        log.trace("Uyuni DB roster:\n%s", dump(ret))
```

But for some reason it doesn't work as supposed to be. It's easy to check with this modification:
```
    if log.isEnabledFor(logging.TRACE):
        log.error("Uyuni DB roster:\n%s", dump(ret))
```
In this case it will write the error in each case no matter which logging level is set.

vzhestkov:
@agraul thanks for the review :+1:  . I've implemented all the suggested changes with https://github.com/uyuni-project/uyuni/pull/4457/commits/c0d1092278ee8a52d0a81dfcd4d305e0dc2445c8

agraul (approved):
Thanks for taking the time implementing my suggestions!

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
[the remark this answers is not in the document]
agraul: Yesterday I just meat to pass the args not yet formatted (`s/%/,/`), but today I realize this does not actually work here since `dump` will always be called before its result is passed to `log.trace`. Maybe the problem with `logging.TRACE` is that we never defined that log level. Salt has it defined, but it is not a default log level https://docs.python.org/3/howto/logging.html#logging-levels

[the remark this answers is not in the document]
vzhestkov: I've tried to change it to DEBUG, but the result is the same.

spacewalk/config/spacewalk-config.spec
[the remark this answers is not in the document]
vzhestkov: Finally the required parameters are copying with https://github.com/uyuni-project/uyuni/pull/4457/commits/a756396c58654ecd55a2f6df2544ee5fa0a22102 by `salt-secrets-config` service as @moio suggested.

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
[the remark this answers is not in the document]
vzhestkov: Finally I left it as:
```
    if log.isEnabledFor(logging.TRACE):
        log.trace("Uyuni DB roster:\n%s", dump(ret))
```
But due to the bug in salt it goes inside conditional part in any logging level. Hope it will be fixed.

agraul (approved):
I don't see how anything would break between my last review and this version, but I haven't tried running the code.

It would be great to add the reason for having a class (concurrency IIUC) to its docstring to know why that approach is used in 6 months from now.

schema/spacewalk/upgrade/susemanager-schema-4.3.6-to-susemanager-schema-4.3.7/100-add-suseMinionInfo-created-modified.sql
```diff
@@ -0,0 +1,23 @@
+ALTER TABLE suseMinionInfo ADD COLUMN IF NOT EXISTS
```
mcalmer, line 1: The directory is wrong. 4.3.7 is already released. The next version is already 4.3.8.
So the directory should be named `susemanager-schema-4.3.7-to-susemanager-schema-4.3.8`

mcalmer (changes requested):
Just checked the schema and just found the wrong dir thingy

schema/spacewalk/upgrade/susemanager-schema-4.3.6-to-susemanager-schema-4.3.7/100-add-suseMinionInfo-created-modified.sql
vzhestkov: Thanks @mcalmer. Should be fine now.

mcalmer (approved):

Issue #4077, opened 2021-08-02
Arcitecture list in Release file of generated debian repos could be incomplete


vzhestkov:
# Short description

Only channel architecture is used as an architecture for `Release` file, bul all compatible architectures should be listed (`rhnChannelPackageArchCompat` and `rhnPackageArch`).

# Details 

For `amd64` is works fine, but in case of synchronization `arm` repos (as an example - Raspbian), the repos are synchronized with no issues, but managed systems can't use it as there is only `Architectures: arm` is in Release file, but `armhf` as a compatible package arch should be also included.

The following line should add not just arch of the channel, but all of the compat package arch excluding `all` and `src`:
https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/taskomatic/task/repomd/DebReleaseWriter.java#L76

probably with modification of https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/taskomatic/task/repomd/DebReleaseWriter.java#L101
to handle `channelArch.getCompatiblePackageArches()`.

Finally for `armhf` should be included to `Architectures:` field like:
```
Architectures: arm armhf
```

PR #4459, opened 2021-11-11
Generate flat deb repos with compatible architectures support

vzhestkov:
## What does this PR change?

Fixes: https://github.com/uyuni-project/uyuni/issues/4077

## GUI diff

No difference.

## Documentation

## Test coverage
- Unit tests fixed

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/4077

## Changelogs

- Generate flat deb repos with compatible architectures support

## Re-run a test


vzhestkov:
@moio If I remember right on discussing such change you've mentioned that there is a way to trigger recreation of meta data files for the repo. It's required for this change as Release file should be regenerated otherwise the clients will try to get packages with the wrong URLs. Could you please give a hint how to do it?

moio:
Sure, you will have to add a row to `rhnRepoRegenQueue`, see for example https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/upgrade/susemanager-schema-4.0.17-to-susemanager-schema-4.0.18/005-fix-none-vendor.sql#L3-L12

vzhestkov:
:+1:  Thanks, will add to the PR

spacewalk/config/etc/httpd/conf.d/zz-spacewalk-www.conf
```diff
@@ -51,9 +51,8 @@ RewriteRule ^/index\.html$ /rhn/manager/login
 # For rhn-custom-info
 RewriteRule ^/WEBRPC /rhn/rpc/api
 
-# Rewrite rules for APT repos metadata
-RewriteRule ^/rhn/manager/download/dists/(.*)/main/(.*)/(Packages.*)$ /rhn/manager/download/$1/repodata/$3
-RewriteRule ^/rhn/manager/download/dists/(.*)/(InRelease|Release.*)$ /rhn/manager/download/$1/repodata/$2
+# Rewrite rule for APT repos metadata
+RewriteRule ^/rhn/manager/download/(.*)/(InRelease|Release.*|Packages.*)$ /rhn/manager/download/$1/repodata/$2
```
mcalmer, line 55: I see problems with the upgrade.
After you update the server, all debian like clients are broken as dist/* is not redirected anymore.
You need to regenerate all the metadata and force all clients to re-write the new style channel list .
I think this can be done via "highstate", but it is not enforced by the update.

What would happen if a client want to access a repository without the changes from `channels.sls` are applied?

vzhestkov: Release files should be regenerated. I've implemented it with https://github.com/uyuni-project/uyuni/pull/4459/commits/eefe6564136ef9b709c339eda8903fb5c82e6bce
`channels` state is applied automatically on package/patch installation with web UI, if I'm not mistaken. Package/patch installation with salt directly will fail if `channels` state not applied before. Is it possible to force apply this state only?

mcalmer: not automatically. If you use the UI and go to the channels page (or SSM Channels) and you just click through it without changing anything you get this done. But it is still a manual step.
But if it works on just updating a package it might be sufficient. Only the people using apt-get directly on the client step into it.

vzhestkov: We can left the deleted strings in the config for a while and delete them on next release or point in the documentation that applying `channels` state is required for all debian based systems after installing the update.

mcalmer: If Silvio is fine with this, I am ok as well.

vzhestkov:
Looks like `;` was missing in the schema file, will check. Thanks

spacewalk/config/etc/httpd/conf.d/zz-spacewalk-www.conf
moio: Yeah, unless it creates problems please leave old redirections intact for a while.

My only ask would be to add a `# TEMPORARY: <explanation on when following lines can be removed>` comment for posterity.

vzhestkov: @moio I think it's the best option for such change.

mcalmer (approved):

PR #4249, opened 2021-09-27
WIP: Refactor bootstapping ssl

mbussolotto:
## What does this PR change?

**add description**

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation
- No documentation needed: **add explanation. This can't be used if there is a GUI diff**
- No documentation needed: only internal and user invisible changes
- Documentation issue was created: [Link for SUSE Manager contributors](https://github.com/SUSE/spacewalk/issues/new?template=ISSUE_TEMPLATE_DOCUMENTATION.md&labels=documentation&projects=SUSE/spacewalk/31), [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/new).
- (OPTIONAL) [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pulls)

- [ ] **DONE**

## Test coverage
- No tests: **add explanation**
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #
Tracks # **add downstream PR, if any**

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #4469, opened 2021-11-15
fix changelogs with 4.2.4 changes

jordimassaguerpla:
@juliogonzalez Not sure if this is needed. I had to add these 2 entries for 4.2.4. Should they be in master as well?

### DIFF
--- a/susemanager/susemanager.changes
+++ b/susemanager/susemanager.changes
@@ -1,3 +1,6 @@
+- Reorganize bootstrap SSL state
+- Add missing packages on SSL bootstrap of Debian-10 and SLES-15
+- Update translation strings
 -------------------------------------------------------------------
 Fri Nov 05 14:04:35 CET 2021 - jgonzalez@suse.com
 
--- a/susemanager-frontend/susemanager-nodejs-sdk-devel/susemanager-nodejs-sdk-devel.changes
+++ b/susemanager-frontend/susemanager-nodejs-sdk-devel/susemanager-nodejs-sdk-devel.changes
@@ -1,3 +1,5 @@
+- Standardize frontend code style
+
 -------------------------------------------------------------------
 Fri Nov 05 14:05:13 CET 2021 - jgonzalez@suse.com
 


Etheryte:
"Standardize frontend code style" was only added so the changes could be tagged if I recall correctly. It doesn't hurt, but it also doesn't really give customers any useful information whatsoever. Perhaps Julio remembers more context?

susemanager/susemanager.changes
```diff
@@ -1,3 +1,6 @@
+- Reorganize bootstrap SSL state
+- Add missing packages on SSL bootstrap of Debian-10 and SLES-15
+- Update translation strings
```
juliogonzalez, line 3: You are missing an empty line on top of the separator, and below this line.

But anyway, why this is needed? I don't remember the reorganization of the bootstrap states was backported to 4.2. Are you sure you are adding the right changes!?

juliogonzalez (changes requested):
@jordimassaguerpla if you check master, you will see the changes are already there for `susemanager-nodejs-sdk`, so no need to port them: https://github.com/uyuni-project/uyuni/blob/master/susemanager-frontend/susemanager-nodejs-sdk-devel/susemanager-nodejs-sdk-devel.changes

It's unclear to me if the changes for `susemanager` are correct.

juliogonzalez:
Ah, now I understand, you are changing master, not 4.2.

So well, the change to `susemanager-nodejs-sdk` is not needed for sure.

But how is that all those changes are missing from master? Are you sure the relevant PRs are merged!?

jordimassaguerpla:
> Ah, now I understand, you are changing master, not 4.2.
> 
> So well, the change to `susemanager-nodejs-sdk` is not needed for sure.
> 
> But how is that all those changes are missing from master? Are you sure the relevant PRs are merged!?

The changes are in. See: https://github.com/uyuni-project/uyuni/pull/4249 . However, the "no need changelog" checkbox was marked. Given this is a refactor, I would not include it either. I had to include it because in 4.2 there was no other entry, and we can't have an empty changelog.

We could just drop this PR. My question is whether changelogs in 4.2 branch need to be in sync with the changes in master, this is why I created this PR, to port the changes I did in 4.2 to master.

juliogonzalez:
> > Ah, now I understand, you are changing master, not 4.2.
> > So well, the change to `susemanager-nodejs-sdk` is not needed for sure.
> > But how is that all those changes are missing from master? Are you sure the relevant PRs are merged!?
> 
> The changes are in. See: #4249 . However, the "no need changelog" checkbox was marked. Given this is a refactor, I would not include it either. I had to include it because in 4.2 there was no other entry, and we can't have an empty changelog.
> 
> We could just drop this PR. My question is whether changelogs in 4.2 branch need to be in sync with the changes in master, this is why I created this PR, to port the changes I did in 4.2 to master.

Ok, since I intend to submit Alpha1 before 4.2.4 is released, these changes will not show as absent from 4.3 (and even if they do, no bscs will be missing).

So in this particular case, we don't need the PR. Feel free to close :-)

PR #4474, opened 2021-11-15
Deprecate old popup component

Etheryte:
## What does this PR change?

We currently have three components taking care of showing dialog windows: `PopUp`, `LegacyDialog` and `Dialog`. This PR consolidates the first two into one. Combining all of them together turned out not to be feasible since we have miscellaneous JSP pages that might rely on some of the jQuery logic. I might return to that bit in the future, but for now this is a good chunk of work done.  

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/components/dialog/LegacyDialog.tsx
```diff
@@ -1,49 +1,72 @@
...
   id: string;
+  /** The css className for the 'modal-dialog' div */
   className?: string;
   title?: React.ReactNode;
+  /** The body of the popup */
   content?: React.ReactNode;
-  buttons?: React.ReactNode;
+  /** The footer of the popup */
+  footer?: React.ReactNode;
+  hideHeader?: boolean;
```
ncounter, line 19: (Partially out of scope, but now that we are at it....) Do we really use this? And if so, do we really want to use this? I am not sure how good it can be to hide a slice of a proper component design. It does not seem to be a _conditional content_, it affects and hide completely the `header` and the `title` of the box.
If you'd agree but still feel out of scope, feel free to create a separate issue to address this. Thanks

Etheryte: I had the same question as you actually, and turns out we use it in `web/html/src/manager/virtualization/guests/console/MessagePopUp.tsx`. I think the idea there is that the doesn't have a title so the header wants to be hidden. I think this prop can be refactored out though if we want, so that the header is hidden automatically if there's no title or header content.

ncounter (approved):
Other than the small comment I left on the `hideHeader` prop (I leave it up to you weather to refactor in this context or not), everything else looks good to me, especially centralizing, narrowing down and wrapping the mixed usage of jQuery and those kind of things.
Thanks a lot for organizing better the codebase in order to move away from legacy pieces. :+1:

web/html/src/components/dialog/LegacyDialog.tsx
Etheryte: Since deprecating this also affected a few other modals due to the way their props were, I decided to leave this be for now.

mcalmer:
Let's better run the testsuite on this: https://ci.nue.suse.com/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/1026

mcalmer:
@Etheryte The test on PR ended wih 223 Errors. Please have a look. I remove the merge-candidate label.
http://jenkins-worker-prs.mgr.prv.suse.net/workspace/suma-pr3/results/1026/

Not sure if the errors could be caused by your change or not.

Etheryte:
@mcalmer I had a look at the results and it's unclear what the problem is, a lot of issues with zypper and a number of other odd things. I reran the suite (run 1032) and it had only a few failures again. At the same time, it seems it only passed a small number of tests, not the full thing, so perhaps there's a problem with the test runner again?

Etheryte:
Will revisit and rebase this at a later date with a fresh branch.

PR #4475, opened 2021-11-16
Detect correct xmlrpc function

mcalmer:
## What does this PR change?

For the XMLRPC function `bootstrap(...)` we used as last parameter the primitive type `boolean`.
This caused problems when detecting the right function. To find the function, the code look for the name and the same number of parameters like in the function call. In our case we fin 2:

```
int bootstrap(User, String, Integer, String, String, String, Integer, boolean)
int bootstrap(User, String, Integer, String, String, String, String, boolean)
```

The method `findPerfectMethod` compare now the XML converted type with the type in the signature.
As the XMLRPC library convert all types to Objects, we compare `boolean` with `Boolean`
which does not match.

In this case the first found method is used and a type convert is tried. But the first method is randomly changing and we could get sometimes the wrong one. As an Integer can be converted to a String and a Boolean into a boolean, the function can be called successfully.

Changing the function signature to use `Boolean` fix the problem.

Port of https://github.com/SUSE/spacewalk/pull/16391
## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/BaseHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/BaseHandler.java
@@ -82,6 +82,7 @@ public class BaseHandler implements XmlRpcInvocationHandler {
      * @return the results of the method of the subclass
      * @exception XmlRpcFault if some error occurs
      */
+    @Override
     public Object invoke(String methodCalled, List params) throws XmlRpcFault {
         Class myClass = this.getClass();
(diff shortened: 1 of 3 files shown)


Bischoff (approved):
See discussion in the other PR ; approving

PR #4487, opened 2021-11-18
Add pagination support to async Select

Etheryte:
## What does this PR change?

First part of addressing https://github.com/SUSE/spacewalk/issues/16363, this PR adds pagination to the async dropdown and then switches the channel selection over from the legacy select logic so it doesn't try to render 5000 items at once. I haven't tested the new addition with all possible combinations of props that the dropdown component accepts, but for the given case it seems to work fine. In short, I expect there might be some edge case bugs that I haven't found yet, but this is the first step to address the performance problems of this UI.  

In the future, the logic for the pagination can be moved to the server, but currently I don't think it's worth putting in that effort yet until we know what other changes we might need.  

No changelog yet since this doesn't fix the whole problem.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: This definitely needs some tests, but I'll add some in a followup PR once I've figured out the big parts of fixing this L3.

- [x] **DONE**

## Links

Part of https://github.com/SUSE/spacewalk/issues/16363

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


web/html/src/manager/content-management/shared/components/panels/sources/channels/child-channels.tsx
```diff
@@ -24,47 +24,48 @@ const ChildChannels = (props: PropsType) => {
...
-              {isMandatory ? (
-                <span className="mandatory-tag-base" title={t("This channel is mandatory")}>
-                  {t("mandatory")}
-                </span>
-              ) : null}
-              <ChannelAnchorLink id={c.id} newWindow={true} />
-            </div>
-          );
-        })}
+      {props.childChannelsId.map((channelId: number) => {
```
Etheryte, line 27: This code is unchanged, I simply removed iterating over the data twice.

ncounter (approved):
So, if I get it correctly, it is more or less a *scroll loading* implementation for slicing the whole bunch of data in pieces and incrementally add them to the view, once they are loaded the list will be just long as it is.
As an initial step it is already a good step ahead from my POV. Thanks for starting this journey :wink:

Etheryte:
@ncounter Yes, the key difference here is that you can use the dropdown, type in it and search and get results, whereas previously the dropdown would also practically just freeze. Since the dropdown is searchable, I wouldn't really expect anyone to keep scrolling down for two days just to load all 5000 items. 😃 But if they did, it would still be at least somewhat manageable for the browser because it doesn't have to do all of the rendering at once.  

The virtualized scrolling bit I discussed under the ticket is planned for the rest of the view, but I'm only just starting trying to make that work so we'll see how that goes.

Etheryte:
This PR is superseded by https://github.com/uyuni-project/uyuni/pull/4527.

PR #4489, opened 2021-11-19
use a permanent config key for SSHMinionActionExecutor parallel threads

mcalmer:
## What does this PR change?

The number of parallel threads can be configured only by using the class name as config variable name.
When moving the Class to a new package or rename the class, the config variable become invalid.

To solve this, this PR introduce an annotation to specify the config key. The class name is still used as fallback to stay compatible with existing configurations. In future the class can be renamed or moved without changing the main config name.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- https://github.com/uyuni-project/uyuni-docs/pull/1331

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13612
Tracks https://github.com/SUSE/spacewalk/pull/16541

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

https://ci.nue.suse.com/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/981/

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/TaskoJob.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/TaskoJob.java
@@ -14,14 +14,11 @@
  */
 package com.redhat.rhn.taskomatic;
 
-import com.redhat.rhn.common.conf.Config;
 import com.redhat.rhn.common.hibernate.HibernateFactory;
 import com.redhat.rhn.taskomatic.domain.TaskoRun;
 import com.redhat.rhn.taskomatic.domain.TaskoSchedule;
 import com.redhat.rhn.taskomatic.domain.TaskoTask;
 import com.redhat.rhn.taskomatic.domain.TaskoTemplate;
-import com.redhat.rhn.taskomatic.task.MinionActionExecutor;
-import com.redhat.rhn.taskomatic.task.RepoSyncTask;
 import com.redhat.rhn.taskomatic.task.RhnJob;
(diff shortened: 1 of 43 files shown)


mcalmer:
@moio @lucidd I do not really like the annotation name. Do you have a better idea and is this the right way?

moio (approved):
I welcome this change. Ideally we would also have a test for the more beefy `isTaskThreadAvailable` implementation.

java/code/src/com/redhat/rhn/taskomatic/TaskoJob.java
```diff
@@ -106,8 +107,14 @@ private boolean isTaskRunning(TaskoTask task) {
     }
 
     private boolean isTaskThreadAvailable(TaskoTask task) {
-        return tasks.get(task.getName()) < Config.get().getInt("taskomatic." +
-                task.getTaskClass() + ".parallel_threads", 1);
+        int parallelThreads = Config.get().getInt("taskomatic." + task.getTaskClass() + ".parallel_threads", 1);
+        if (task.getClass().isAnnotationPresent(TaskConfigKey.class)) {
+            TaskConfigKey annotation = task.getClass().getAnnotation(TaskConfigKey.class);
```
lucidd, line 112: Is there any advantage to using an annotation rather then just a method returning the config name? With a method at least later on when every task has its config name and there is no default fallback method it would cause a compilation error not implementing it for any new task going forward.

PR #4483, opened 2021-11-18
Fix legacy timepicker passing wrong time to the backend

parlt91:
## What does this PR change?

Fix cases where the legacy `DateTimePicker` would pass the wrong time to the backend if selected date is in summer time or if server and user timezone differ.

As well as an issue where times between 12:00 and 12:59 would be parsed as 00:XX

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **Bugfix**

- [x] **DONE**

## Test coverage

- No tests: **Bugfix**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/16384 
Tracks https://github.com/SUSE/spacewalk/issues/16406

- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #4496, opened 2021-11-22
Fix timepicker not reacting to to typing input

parlt91:
## What does this PR change?

Fix an issue where the timepicker would not react to manually typed input (not using the dropdown)

This is a followup fix for https://github.com/uyuni-project/uyuni/pull/4483 and thus requires no additional changelog

Also fix issues where the wrong time would be displayed in system pending actions and action history if server timezone and user preferred timezone are differnt

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: **Bugfix**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/16384

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/components/datetime/DateTimePicker.tsx
+++ b/web/html/src/components/datetime/DateTimePicker.tsx
@@ -155,14 +155,11 @@ class TimePicker extends React.PureComponent<TimePickerProps> {
     this._input?.timepicker({
       roundingFunction: (seconds, options) => seconds,
     });
-    this._input?.on("change", () => {
-      // Do nothing
-    });
     this._input?.on("timeFormatError", () => {
       // Do nothing
     });
     this._input?.timepicker("setTime", this.toFauxBrowserDate(this.props));
-    this._input?.on("changeTime", () => {
+    this._input?.on("change", () => {
       const unsafeDate: Date | undefined = this._input?.timepicker("getTime");
 
       // Only if value has actually changed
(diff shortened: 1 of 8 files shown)


web/html/src/components/datetime/DateTimePicker.tsx
```diff
@@ -155,14 +155,11 @@ class TimePicker extends React.PureComponent<TimePickerProps> {
...
     });
-    this._input?.on("change", () => {
-      // Do nothing
-    });
     this._input?.on("timeFormatError", () => {
       // Do nothing
     });
     this._input?.timepicker("setTime", this.toFauxBrowserDate(this.props));
-    this._input?.on("changeTime", () => {
+    this._input?.on("changeTime" && "change", () => {
```
Etheryte, line 162: This looks a bit off, `"changeTime" && "change"` evaluates to `"change"` in Javascript. If only listening to the change event is correct, then it should be `"change"`, if we want to listen to both events then if my jQuery memory doesn't fail me it should be `"changeTime change"`. I'm not sure which one is correct though, if the events overlap then we probably don't want to listen to both. If you tested this code and it worked then we probably only want the change event.

parlt91: Well I know it look weird, but I tried this out and just putting `"change"` won't trigger when you type something in and hit enter. (But thinking about this that's going to be fine, because the change would trigger once you deselect the text field anyway).
Long story short I am going to change it to just `"change"`

parlt91: Updated and tested with just `"change"`

Etheryte (approved):

juliogonzalez:
@parlt91 @jordimassaguerpla was this tested with acceptance on PR? Are we sure we don't need any testsuite changes?

parlt91:
@juliogonzalez I just made some additional changes to the PR this morning. I will remove the `merge-candidate label` for now until latest changes have been reviewed and acceptance tests are green/have been adapted if needed

hustodemon (approved):
The java/jsp parts look good to me :+1:

PR #4506, opened 2021-11-25
package locking XMLRPC API

mbussolotto:
## What does this PR change?
Add new endpoints to packages API: schedulePackageLockChange, listLocks
Minor changes:
 - allow `ActionManager.schedulePackageLock` to take as argument a list of server
 - fix format mismatch between PackageListItem class and Package class NEVRA
 - add package_id `system.listInstalledPackages`
 - system.listInstalledPackages object in a more readable order 
## GUI diff

No difference.

- [x] **DONE**

## Documentation
API Documentation 
- [x] **DONE**

## Test coverage
- N/A

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15573

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/rhnpackage/PackageFactory.java
+++ b/java/code/src/com/redhat/rhn/domain/rhnpackage/PackageFactory.java
@@ -102,6 +102,17 @@ public class PackageFactory extends HibernateFactory {
     }
 
     /**
+     * Lookup Packages by IDs
+     * @param ids list of id to search for
+     * @return list of Packages found
+     */
+    private static List<Package> lookupById(List<Long> ids) {
+        Map<String, Object> params = new HashMap<String, Object>();
+        return (List<Package>)
+                singleton.listObjectsByNamedQuery("Package.findByIds", params, ids, "pids");
+    }
+
+    /**
      * Returns true if the Package with the given name and evr ids exists in the
      * Channel whose id is cid.
      * @param cid Channel id to look in
@@ -137,6 +148,18 @@ public class PackageFactory extends HibernateFactory {
     }
 
     /**
+     * Lookup Packages by the id, in the context of a given user. Does security
+     * check to verify that the user has access to the packages.
+     * @param ids List of the Package to search for
+     * @param user the user doing the search
+     * @return List of packages found
+     */
+    public static List<Package> lookupByIdAndUser(List<Long> ids, User user) {
+        return lookupByIdAndOrg(ids, user.getOrg());
+    }
+
+
+    /**
      * Lookup a Package by the id, in the context of a given org. Does security
      * check to verify that the org has access to the package.
      * @param id of the Package to search for
@@ -150,6 +173,22 @@ public class PackageFactory extends HibernateFactory {
             return null;
         }
         return lookupById(id);
+    }
+
+    /**
+     * Lookup Packages by the id, in the context of a given org. Does security
+     * check to verify that the org has access to the package.
+     * @param ids List of the Packages to search for
+     * @param org the org which much have access to the package
+     * @return List of Packages found
+     */
+    public static List<Package> lookupByIdAndOrg(List<Long> ids, Org org) {
+        if (!UserManager.verifyPackagesAccess(org, ids)) {
+            // User doesn't have access to the package... return null as if it
+            // doesn't exist.
+            return Collections.emptyList();
+        }
+        return lookupById(ids);
     }
 
     /**
--- a/java/code/src/com/redhat/rhn/frontend/action/rhnpackage/LockPackageAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/rhnpackage/LockPackageAction.java
@@ -43,12 +43,12 @@ import org.apache.struts.action.ActionMapping;
 import org.apache.struts.action.ActionMessages;
 import org.apache.struts.action.DynaActionForm;
 
-import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -104,17 +104,18 @@ public class LockPackageAction extends BaseSystemPackagesAction {
 
         DataResult<PackageListItem> lockedPackagesResult =
                 PackageManager.systemLockedPackages(sid, null);
-        for (PackageListItem pkg : lockedPackagesResult) {
-            if (!context.isSubmitted()) {
+
+        if (!context.isSubmitted()) {
+            for (PackageListItem pkg : lockedPackagesResult) {
                 // pre-select locked
                 pkgsToSelect.add(pkg.getIdCombo() + "~*~" + pkg.getNvrea());
             }
-
-            Package lockedPkg = PackageManager.lookupByIdAndUser(pkg.getPackageId(), user);
-            if (lockedPkg != null) {
-                pkgsAlreadyLocked.add(lockedPkg);
-            }
         }
+
+        List<Package> pkgsFindAlreadyLocked = PackageManager.lookupByIdAndUser(
+                lockedPackagesResult.stream().map(PackageListItem::getPackageId)
+                        .collect(Collectors.toList()), user);
+        pkgsAlreadyLocked.addAll(pkgsFindAlreadyLocked);
 
         SessionSetHelper helper = new SessionSetHelper(request);
         if (isSubmitted(form)) {
@@ -198,7 +199,7 @@ public class LockPackageAction extends BaseSystemPackagesAction {
         RequestContext context = new RequestContext(request);
         Long sid = context.getRequiredParam("sid");
         User user = context.getCurrentUser();
-        List<Package> pkgsToUnlock = new ArrayList<Package>();
+        Set<Package> pkgsToUnlock = new HashSet<Package>();
         String[] selectedPkgs = ListTagHelper.getSelected(LIST_NAME, request);
 
         if (selectedPkgs != null) {
@@ -212,7 +213,7 @@ public class LockPackageAction extends BaseSystemPackagesAction {
 
         PackageManager.setPendingStatusOnLockedPackages(pkgsToUnlock,
                                                         PackageManager.PKG_PENDING_UNLOCK);
-        ActionManager.schedulePackageLock(user, server, pkgsAlreadyLocked, scheduleDate);
+        ActionManager.schedulePackageLock(user, pkgsAlreadyLocked, scheduleDate, server);
     }
 
     /**
@@ -255,7 +256,7 @@ public class LockPackageAction extends BaseSystemPackagesAction {
 
             // Ensure pending locks and already locked items are sent to the client
             pkgsToLock.addAll(pkgsAlreadyLocked);
-            ActionManager.schedulePackageLock(user, server, pkgsToLock, scheduleDate);
+            ActionManager.schedulePackageLock(user, pkgsToLock, scheduleDate, server);
         }
         else {
             LockPackageAction.LOG.info("No packages to lock");
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
@@ -207,9 +207,11 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.SortedMap;
@@ -1536,12 +1538,15 @@ public class SystemHandler extends BaseHandler {
      * @xmlrpc.returntype
      *      #array_begin()
      *          #struct_begin("package")
+     *                 #prop_desc("int", "package_id", "PackageID, -1 if package is installed but not available in
+     *                 subscribed channels")
      *                 #prop("string", "name")
+     *                 #prop("string", "epoch")
      *                 #prop("string", "version")
      *                 #prop("string", "release")
-     *                 #prop("string", "epoch")
      *                 #prop_desc("string", "arch", "architecture label")
      *                 #prop_desc("date", "installtime", "returned only if known")
+     *                 #prop("boolean", "retracted")
      *          #struct_end()
      *      #array_end()
      */
@@ -1552,7 +1557,8 @@ public class SystemHandler extends BaseHandler {
         DataResult<PackageListItem> packageListItems = PackageManager.systemPackageList(server.getId(), null);
         packageListItems.elaborate();
         List<Map<String, Object>> maps = packageListItems.stream().map(pi -> {
-            Map<String, Object> item = new HashMap<>();
+            Map<String, Object> item = new LinkedHashMap<>();
+            item.put("package_id", Objects.isNull(pi.getPackageId()) ? -1 : pi.getPackageId().intValue());
             item.put("name", pi.getName());
             item.put("epoch", Optional.ofNullable(pi.getEpoch()).orElse(" "));
             item.put("version", pi.getVersion());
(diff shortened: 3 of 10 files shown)


java/code/src/com/redhat/rhn/frontend/xmlrpc/packages/PackagesHandler.java
```diff
@@ -597,4 +605,110 @@ public String getPackageUrl(User loggedInUser, Integer pid) {
...
+        Set<Package> pkgsAlreadyLocked = lockedPackagesResult.stream().map(
+                x -> PackageManager.lookupByIdAndUser(x.getPackageId(), loggedInUser)).collect(Collectors.toSet());
+
+        if (!pkgsAlreadyLocked.contains(pkg)) {
+            logger.warn("Package " + pkg.getNameEvra()  + " with pid " + pkg.getId() + " is not locked");
+            return 0;
+        }
+
+        PackageManager.setPendingStatusOnLockedPackages(List.of(pkg), PackageManager.PKG_PENDING_UNLOCK);
+        PackageManager.unlockPackages(sid.longValue(), Set.of(pkg));
```
mcalmer, line 709: As discussed, it needs to be scheduled

java/code/src/com/redhat/rhn/frontend/xmlrpc/packages/PackagesHandler.java
```diff
@@ -597,4 +605,110 @@ public String getPackageUrl(User loggedInUser, Integer pid) {
...
+        Set<Package> pkgsAlreadyLocked = lockedPackagesResult.stream().map(
+                x -> PackageManager.lookupByIdAndUser(x.getPackageId(), loggedInUser)).collect(Collectors.toSet());
+
+        if (pkgsAlreadyLocked.contains(pkg)) {
+            logger.warn("Package " + pkg.getNameEvra() + " with pid " + pkg.getId() + "already locked");
+            return 0;
+        }
+
+        PackageManager.setPendingStatusOnLockedPackages(List.of(pkg), PackageManager.PKG_PENDING_LOCK);
+        PackageManager.lockPackages(sid.longValue(), Set.of(pkg));
```
mcalmer, line 636: as discussed, it needs to be scheduled. And schedule functions all belong to the SystemHander.
I would suggest to move all these functions to the system handler and maybe reduce it to 1 schedule function only for changing the locks.

mcalmer (changes requested):

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -4569,6 +4614,57 @@ public int schedulePackageRemoveByNevra(User loggedInUser, final Integer sid, Li
...
+
+        Server server = SystemManager.lookupByIdAndUser(sid.longValue(), loggedInUser);
+        DataResult<PackageListItem> lockedPackagesResult =
+                PackageManager.systemLockedPackages(server.getId(), null);
+
+        Set<Package> pkgsAlreadyLocked = lockedPackagesResult.stream().map(
+                pkg -> PackageManager.lookupByIdAndUser(pkg.getPackageId(), loggedInUser)).collect(Collectors.toSet());
+        Set<Package> pkgsToLock = pkgIdsToLock.stream().map(
+                pkg -> PackageManager.lookupByIdAndUser(pkg.longValue(), loggedInUser)).
+                filter(Objects::isNull).collect(Collectors.toSet());
```
mackdk, line 4647: Not sure to follow the logic here, maybe I'm missing something. Why are you collecting the ```null``` elements? Shouldn't it be ```Objects::nonNull```? Anyway, I think in this case you can avoid traversing the stream twice by using ```peek```:

```
pkgIdsToLock.stream().map( ... ).filter( ... ).peek(pkg -> pkg.setLockPending(Boolean.TRUE)).collect(Collectors.toSet())
```

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -4569,6 +4614,57 @@ public int schedulePackageRemoveByNevra(User loggedInUser, final Integer sid, Li
...
+        PackageManager.lockPackages(server.getId(), pkgsToLock);
+
+        pkgsToLock.addAll(pkgsAlreadyLocked);
+        try {
+            Action a = ActionManager.schedulePackageLock(loggedInUser, pkgsToLock, earliestOccurrence, server);
+            return a.getId();
+
+        }
+        catch (com.redhat.rhn.taskomatic.TaskomaticApiException e) {
+            throw new TaskomaticApiException(e.getMessage());
```
mackdk, line 4665: Why is this re-throw needed? It seems to me that it would be the same if we don't catch the exception. Or should we maybe log something?

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -1536,10 +1538,11 @@ public String downloadSystemId(User loggedInUser, Integer sid) throws FaultExcep
      * @xmlrpc.returntype
      *      #array_begin()
      *          #struct_begin("package")
+     *                 #prop("int", "package_id")
```
mcalmer, line 1541: suggestion: you could explain under which conditions the package id is `-1`. I think that happenes for packages which are installed, but not available in Uyuni/SUSE Manager.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -4569,6 +4614,57 @@ public int schedulePackageRemoveByNevra(User loggedInUser, final Integer sid, Li
...
+     * Schedule package lock for a system.
+     *
+     * @param loggedInUser The current user
+     * @param sid ID of the server
+     * @param pkgIdsToLock List of package IDs to lock (as Integers)
+     * @param pkgIdsToUnlock List of package IDs to lock (as Integers)
+     * @param earliestOccurrence Earliest occurrence of the package removal
+     * @return package action id
+     *
+     * @xmlrpc.doc Schedule package installation for a system.
```
mcalmer, line 4627: wrong text

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -4569,6 +4614,57 @@ public int schedulePackageRemoveByNevra(User loggedInUser, final Integer sid, Li
...
+
+        Set<Package> pkgsToUnlock = pkgIdsToUnlock.stream().map(
+                        x -> PackageManager.lookupByIdAndUser(x.longValue(), loggedInUser)).
+                filter(Objects::isNull).collect(Collectors.toSet());
+
+        PackageManager.setPendingStatusOnLockedPackages(pkgsToUnlock, PackageManager.PKG_PENDING_UNLOCK);
+
+        PackageManager.lockPackages(server.getId(), pkgsToLock);
+
+        pkgsToLock.addAll(pkgsAlreadyLocked);
```
mcalmer, line 4660: I think you need to reduce `pkgsAlreadyLocked` by `pkgToUnlock`.

`pkgToLock = pkgsAlreadyLocked - pkgsToUnlock + pkgToLock`

or did I miss something?

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -1568,6 +1573,48 @@ public String downloadSystemId(User loggedInUser, Integer sid) throws FaultExcep
...
+     * @return Returns an array of maps representing the packages locked on a system
+
+
+     * @xmlrpc.doc List current package locks
+     * @xmlrpc.param #session_key()
+     * @xmlrpc.param #param("string", "system_id")
+     * @xmlrpc.returntype
+     *      #array_begin()
+     *          #struct_begin("package")
+     *                 #prop("int", "package_id", "PackageID, -1 if package is locked but not available in
```
mcalmer, line 1589:
```suggestion
     *                 #prop_desc("int", "package_id", "PackageID, -1 if package is locked but not available in
```

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
mbussolotto: I guess you are right, I'm still testing it. I will put PR in ready when all the changes are finished and tested (currently I put this PR in draft)

mbussolotto: Actually it seems to be right as it is: `pkgsAlreadyLocked - pkgsToUnlock + pkgToLock` add locks correctly, but packages to be unlock will remain in a `unlocking..` pending status.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
mbussolotto: Although I agree, I followed the pattern presents in the other methods of this class...my feeling is it's something related to xmlrpc implementation.. @mcalmer what do you think?

mcalmer: Oh, this is not obvious. These are 2 different Exceptions with the same name but in different Packages.
Don't ask me why we have 2.

The new one is "com.redhat.rhn.frontend.**xmlrpc**.TaskomaticApiException" which is a "Fault" Exceptions special for XMLRPC.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
mcalmer: Check what happens when the action returns. At that point the status should be updated.
SaltUtils.java handlePackageLockData()

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -4569,6 +4614,57 @@ public int schedulePackageRemoveByNevra(User loggedInUser, final Integer sid, Li
...
+    public Long schedulePackageLockChange(User loggedInUser, Long sid,
+                                          List<Long> pkgIdsToLock,
+                                          List<Long> pkgIdsToUnlock, Date earliestOccurrence) {
+
+        Server server = SystemManager.lookupByIdAndUser(sid.longValue(), loggedInUser);
+        DataResult<PackageListItem> lockedPackagesResult =
+                PackageManager.systemLockedPackages(server.getId(), null);
+
+        Set<Package> pkgsAlreadyLocked = lockedPackagesResult.stream().map(
+                pkg -> PackageManager.lookupByIdAndUser(pkg.getPackageId(), loggedInUser)).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
mbussolotto: @mcalmer what do you think about that? Should we execute just one query for all packages or should we leave it as it is?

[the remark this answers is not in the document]
mcalmer: Of cause it would be better, but we have the same code for the UI.
I doubt a bit that people will use this for massive locking of packages as it create a lot of troubles in other places.

I would be fine with leaving it, but if you want to implement a more performant solution, please do this also for the UI part.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -1568,6 +1574,52 @@ public String downloadSystemId(User loggedInUser, Integer sid) throws FaultExcep
...
+     *                 #prop("string", "name")
+     *                 #prop("string", "epoch")
+     *                 #prop("string", "version")
+     *                 #prop("string", "release")
+     *                 #prop_desc("string", "arch", "architecture label")
+     *                 #prop_desc("string", "pending status", "return only if there is a pending locking")
+     *          #struct_end()
+     *      #array_end()
+     */
+    public List<Map<String, Object>>  listLocksStatus(User loggedInUser,  Integer sid) {
```
mcalmer, line 1601: I would choose a different name for this function.

We can also "lock a system". Please could missunderstand this. 
Maybe listPackagesLockStatus() ?

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/test/SystemHandlerTest.java
```diff
@@ -124,6 +124,7 @@
 import com.redhat.rhn.frontend.xmlrpc.test.BaseHandlerTestCase;
 import com.redhat.rhn.manager.action.ActionChainManager;
 import com.redhat.rhn.manager.action.ActionManager;
+import com.redhat.rhn.manager.channel.ChannelEditor;
```
mcalmer, line 127: Seems to be the only change in this file now. Is it needed?

java/spacewalk-java.changes
```diff
@@ -1,3 +1,13 @@
+- Add new endpoints to packages API: schedulePackageLockChange, listLocks
+- fix parsing error by making SCAP Profile description attribute optional
+  (bsc#1192321)
+- fix openscap scan with tailoring-file option (bsc#1192321)
+- Pass the "allow_vendor_change" flag using the right name when installing patches
+- Fix legacy timepicker passing wrong time to the backend if server and
+  user time differ (bsc#1192699)
+- Fix legacy timepicker passing wrong time to the backend if selected
+  date is in summer time (bsc#1192776)
```
mcalmer, line 9: I think these are not all changes you made in this PR, right?

mcalmer (changes requested):

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -4569,6 +4620,76 @@ public int schedulePackageRemoveByNevra(User loggedInUser, final Integer sid, Li
...
+        List<Package> pkgsFindAlreadyLocked = PackageManager.lookupByIdAndUser(
+                lockedPackagesResult.stream().map(PackageListItem::getPackageId)
+                        .collect(Collectors.toList()), loggedInUser);
+        if (pkgsFindAlreadyLocked != null) {
+            pkgsAlreadyLocked.addAll(pkgsFindAlreadyLocked);
+        }
+
+        Set<Package> pkgsToLock = new HashSet<>();
+        List<Package> pkgsFindToLock = PackageManager.lookupByIdAndUser(pkgIdsToLock
+                .stream().mapToLong(Integer::longValue).boxed().collect(toList()), loggedInUser);
```
mackdk, line 4659: I think you can do directly 
```map(Integer::longValue)``` instead of ```mapToLong().boxed()``` since we need boxed numbers anyway.

File: java/code/src/com/redhat/rhn/domain/rhnpackage/PackageFactory.java [hunk 14]
```diff
@@ -152,6 +175,22 @@ public static Package lookupByIdAndOrg(Long id, Org org) {
...
+     * check to verify that the org has access to the package.
+     * @param ids List of the Packages to search for
+     * @param org the org which much have access to the package
+     * @return List of Packages found
+     */
+    public static List<Package> lookupByIdAndOrg(List<Long> ids, Org org) {
+        if (!UserManager.verifyPackagesAccess(org, ids)) {
+            // User doesn't have access to the package... return null as if it
+            // doesn't exist.
+            return null;
```
lucidd, line 189: we should return `Optional` here instead of `null` for any new lookup functions.

mackdk: This is returning a collection so I would say that a ```Collections.emptyList()``` is a better fit in this case.

ktsamis:
Removed the label, this is not reviewed yet. When it is then please add the label again

ktsamis:
Also rebase for the changelog

java/code/src/com/redhat/rhn/frontend/action/rhnpackage/LockPackageAction.java
```diff
@@ -104,16 +104,19 @@ public ActionForward execute(ActionMapping mapping,
...
+        }
 
-            Package lockedPkg = PackageManager.lookupByIdAndUser(pkg.getPackageId(), user);
-            if (lockedPkg != null) {
-                pkgsAlreadyLocked.add(lockedPkg);
-            }
+        List<Package> pkgsFindAlreadyLocked = PackageManager.lookupByIdAndUser(
+                lockedPackagesResult.stream().map(PackageListItem::getPackageId)
+                        .collect(Collectors.toList()), user);
+        if (pkgsFindAlreadyLocked != null) {
```
mackdk, line 118: Null check is actually no longer needed, since ```lookupByIdAndUser``` always returns a non-null value after your last change.

mackdk (approved):

mcalmer (approved):

PR #4517, opened 2021-11-29
Various fixes

Bischoff:
## What does this PR change?

Various test suite fixes:
 * 90 seconds max to pick an event ; apparently, our BV hypervisors are slower than test suite ones
 * wrong error message "Couldn't find text" when the problem is that the text was found

and cosmetical changes:
 * quotes in many other error messages, for readability
 * feature titles
 * new syntax for `raise` to make rubocop happy

## Links

Ports:
* 4.1: SUSE/spacewalk#16451
* 4.2: SUSE/spacewalk#16450

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -26,7 +26,7 @@ Then(/^I can see all system information for "([^"]*)"$/) do |host|
   node = get_target(host)
   step %(I should see a "#{node.hostname}" text)
   kernel_version, _code = node.run('uname -r')
-  puts 'i should see kernel version: ' + kernel_version
+  puts 'I should see kernel version: ' + kernel_version
   step %(I should see a "#{kernel_version.strip}" text)
   os_version, os_family = get_os_version(node)
   # skip this test for centos and ubuntu systems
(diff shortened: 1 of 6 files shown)


testsuite/features/step_definitions/common_steps.rb
```diff
@@ -64,7 +64,7 @@
   steps %(
     When I follow "Events"
     And I follow "Pending"
-    And I wait at most 60 seconds until I do not see "#{event}" text, refreshing the page
+    And I wait at most 90 seconds until I do not see "#{event}" text, refreshing the page
```
srbarrios, line 67: 😆 I told you... 👯

srbarrios (approved):

testsuite/features/step_definitions/common_steps.rb
Bischoff: Yes....

Jokes apart, I'm still wondering how it can be you have to wait for one minute and a half for an event to be picked. There must be some product issue, especially that it only happens on centos systems. I did not create a bug report as it would probably not get attention, but something is fishy there.

An investigation in the log files did not find anything. Furthermore, I have not been able to reproduce when running exactly the same features. Could it be that the server is busy doing something else when we run the full BV test suite ? Or is it linked to the package we try to patch (`timezone-data`) ?

@srbarrios

ktsamis (approved):
I like these small fixes, they make everything just a bit clearer and better, thanks Eric

PR #4519, opened 2021-11-30
Fix reboot problem

Bischoff:
## What does this PR change?

* regenerate kernel on SLE 11 SP4
* style and English fixes

## Links

See also uyuni-project/sumaform#1018

Ports:
* 4.1: SUSE/spacewalk#16458
* 4.2: SUSE/spacewalk#16457

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1717,22 +1717,30 @@ And(/^I copy vCenter configuration file on server$/) do
 end
 
 When(/^I export "([^"]*)" with ISS v2 to "([^"]*)"$/) do |channel, path|
-  node = get_target("server")
-  node.run("inter-server-sync export --channels=#{channel} --outputDir=#{path}")
+  $server.run("inter-server-sync export --channels=#{channel} --outputDir=#{path}")
 end
 
 When(/^I import data with ISS v2 from "([^"]*)"$/) do |path|
-  node = get_target("server")
-  node.run("inter-server-sync import --importDir=#{path}")
+  $server.run("inter-server-sync import --importDir=#{path}")
 end
 
 Then(/^"(.*?)" folder on server is ISS v2 export directory$/) do |folder|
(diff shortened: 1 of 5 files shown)


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -1717,22 +1717,30 @@
...
+  raise "Folder exists" if folder_exists?($server, folder)
 end
-When(/^I ensure folder "(.*?)" doesn't exists$/) do |folder|
-    folder_delete($server, folder) if folder_exists?($server, folder)
+
+When(/^I ensure folder "(.*?)" doesn't exist$/) do |folder|
+  folder_delete($server, folder) if folder_exists?($server, folder)
+end
+
+When(/^I regenerate the boot RAM disk on "([^"]*)" if necessary$/) do |host|
```
ktsamis, line 1739: I would put here instead "I regenerate the boot RAM disk on "([^"]*)" if its a SLES 11" to make sure that the check is not done in kernel files or anything like that but we check the OS

ktsamis (changes requested):
This is the only thing that is cosmetic pretty much. I don't mind the current name but mayyyybe it's better to be explicit that this only happens on 11

testsuite/features/step_definitions/command_steps.rb
Bischoff: The problem is, we don't know if it's not necessary on other systems.

It just happens that a kernel patch is the first in the list on SLES 11. That's why we know the problem only on SLES 11 so far.

 Let's imagine it's needed on SLE 12 as well. I prefer to change only on implementation than in implementation and title.
I prefer the step name to remain conceptual and leave the details to the implementation.

ktsamis: Ok then let's leave it as is and we change the implementation if needed later on. All good then

ktsamis (approved):

PR #4522, opened 2021-11-30
Integrate Retail into BV

Bischoff:
## What does this PR change?

work in progress

Retail tests in build validation test suite used a tool named `semi-xmlrpc-tester`. This had the following problems:
* in the cucumber code, it was impossible to see in clear English the details of what was being done
* we were using the internals of the formulas (json files) - this is always fragile, and bypasses potential UI or XML-RPC bugs
* it was one more tool to maintain, written in a language we don't know
* it assumed the terminals were accessible via the network, which is dangerous because of potential DHCP leakage
* the terminals could directly access the SUSE Manager server, without the test suite realizing the issue

This PR refactors the Retail tests in Build Validation to align them with what we have in CI. It also removes SLE 11 testing from Retail BV.

Missing (to be done in other PRs):

* [we should use templates for the terminals like we do for minions](https://github.com/SUSE/spacewalk/issues/17966)
* [jenkins pipelines](https://github.com/SUSE/spacewalk/issues/17973)
* now the code is a bit better in BV, it would need realignement in normal CI
  (and that we split the big Retail feature in two in CI)

## Links

Part of SUSE/spacewalk#16462
Sumaform PR uyuni-project/sumaform#1096
susemanager-ci PR SUSE/susemanager-ci#559

Ports:
* 4.2: SUSE/spacewalk#18256

## Changelogs

- [x] No changelog needed


testsuite/.rubocop_todo.yml
```diff
@@ -258,6 +259,8 @@ Metrics/AbcSize:
 # ExcludedMethods: refine
 Metrics/BlockLength:
   Max: 26
+  Exclude:
+    - 'features/step_definitions/retail_steps.rb'
```
srbarrios, line 263: I guess we can't refactor the code to don't break this rule?

testsuite/features/support/twopence_init.rb
```diff
@@ -220,6 +217,9 @@ def file_inject(node, local_file, remote_file)
 # Other global variables
 $product = product
 $pxeboot_mac = ENV['PXEBOOT_MAC']
+$sle11sp3_terminal = ENV['SLE11SP3_TERMINAL']
+$sle12sp4_terminal = ENV['SLE12SP4_TERMINAL']
+$sle15sp3_terminal = ENV['SLE15SP3_TERMINAL']
```
srbarrios, line 222: I don't get well why we moved this to gloval variables, but not buildhost ones

testsuite/.rubocop_todo.yml
Bischoff: Refactor the code, no, we already tried that and it was a mess.

We could re-introduce the individual `# rubocop:disable Metrics/BlockLength` tags, but I don't see that as much better as changing it for the whole file.

testsuite/features/support/twopence_init.rb
Bischoff: they are all global variables. The difference is that these are NOT twopence objects, because there's no SSH.

testsuite/Rakefile
```diff
@@ -61,6 +61,7 @@ namespace :utils do
   task :generate_smoke_tests do
     clients = Rake::FileList['features/build_validation/init_clients/*.feature']
     clients.reject! { |client| client.include? 'proxy' }
+    clients.reject! { |client| (client.include? 'buildhost') || (client.include? 'terminal') }
```
srbarrios, line 64: If I may suggest. I would have a template file for buildhost and another template file for the terminal. So we autogenerate buildhost/terminal features for each supported OS.

testsuite/run_sets/build_validation_retail.yml
```diff
@@ -0,0 +1,20 @@
+# This file describes the order of features in a Build Validation test suite run.
+
+## NOTE: This stage can't be run in parallel
```
srbarrios, line 3: Consider deploy and setup three branch network if that's the only way to parallelize it.
As otherwise, we will spend around more 6 hours when it can be done in 1.30h

testsuite/run_sets/build_validation_retail.yml
```diff
@@ -0,0 +1,20 @@
...
+## Build Kiwi images for all retail terminals
+## (expected time of building is around 1 hour and 20 minutes for each terminal)
+- features/build_validation/retail/sle11sp4_buildhost_build_kiwi_image.feature
+- features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature
+- features/build_validation/retail/sle15sp3_buildhost_build_kiwi_image.feature
+
+## Deployment of retail terminals
+- features/build_validation/retail/sle11sp3_terminal_deploy.feature
+- features/build_validation/retail/sle12sp5_terminal_deploy.feature
+- features/build_validation/retail/sle15sp3_terminal_deploy.feature
```
srbarrios, line 20: We should not be moving all to one YAML file, as otherwise it will be impossible to parallelize.
If we need to first to terminal_prepare, then build_kiwi_image and then terminal_deploy, let's have a YAML file for each set of sequential features, but not group all together in a unique YAML file, otherwise that can't be paralelized.

testsuite/features/step_definitions/retail_steps.rb
```diff
@@ -442,6 +476,15 @@ def compute_image_name
 When(/^I enter the MAC address of "([^"]*)" in (.*) field$/) do |host, field|
   if host == 'pxeboot_minion'
     mac = $pxeboot_mac
+  elsif host == 'sle11sp3_terminal'
+    mac = $sle11sp3_terminal_mac
+    mac = 'EE:EE:EE:00:00:05' if mac.nil?
```
srbarrios, line 481:
```suggestion
    mac = "EE:EE:EE:00:00:0${ADDRESSES[host]}" if mac.nil?
```

srbarrios: I can't see where we define variables of that kind $sle11sp3_terminal_mac, but maybe we can reduce the cyclomatic complexity of that step if we can something like:
```
elsif host.include? 'terminal'
  mac = TERMINAL_MAC.has_key?(host) ? TERMINAL_MAC[host] : "EE:EE:EE:00:00:0${ADDRESSES[host]}"
```

testsuite/Rakefile
Bischoff: Yup, we need templating. But let's do that in another PR.

testsuite/run_sets/build_validation_retail.yml
Bischoff: Yes, I need to revert this change to yaml files. I did not think about parallelization, my bad.

testsuite/run_sets/build_validation_retail.yml
Bischoff: Yes, I need to revert (at least partially) these changes to yaml files. I did not think about parallelization, my bad.

testsuite/features/step_definitions/retail_steps.rb
Bischoff: less readable :-(

srbarrios: For BV, whatever helps us introducing more client types with less effort and less maintenance is welcomed. When adding something new to BV, we need to always think if we can make it in a way that if tomorrow we have a new client, we don't need to do too many changes here and there. And even like that... we will need :D

testsuite/run_sets/build_validation_retail.yml
Bischoff: reverted.

testsuite/run_sets/build_validation_retail.yml
Bischoff: remark changed.

testsuite/features/core/proxy_branch_network.feature
```diff
@@ -31,15 +31,15 @@ Feature: Setup Uyuni for Retail branch network
...
+  Scenario: Install the Retail pattern on the SUSE Manager server
     When I refresh the metadata for "server"
     When I install pattern "suma_retail" on this "server"
     And I wait for "patterns-suma_retail" to be installed on "server"
 
 @proxy
 @private_net
 @uyuni
-  Scenario: Install the Retail pattern on the server
+  Scenario: Install the Retail pattern on the Uyuni server
```
srbarrios, line 42: Why including changes not related to BV scope?
Please refrain about adding changes not strictly related to BV into this PR.
It's already big enough.

srbarrios:
I would not delete semi-xmlrpc-tester folder. It's a nice test tool, that it can still be used when we need. We should definitely not use it in our BV or CI, that we agree. But I would keep it in our testsuite folder.

srbarrios:
We should make clear in the title that we are not only integrating retail in BV, but we are also refactoring it in CI. (Not just the note at the bottom of the description)

Bischoff:
> I would not delete semi-xmlrpc-tester folder. It's a nice test tool, that it can still be used when we need. We should definitely not use it in our BV or CI, that we agree. But I would keep it in our testsuite folder.

I agree to keep it, but not in the `testsuite/` directory, because:
- it's not related to test suite
- it's megabyte to fork and clone when you code
- it's megabytes to check out during all of our runs (!)
- it's under compiled form, we rather need to save the source code, so we can maintain it

What about https://github.com/uyuni-project/contrib/uyuni-tools, or https://github.com/uyuni-project/retail?

Bischoff:
> As agreed in Slack https://suse.slack.com/archives/C02D12TNYLS/p1653402227649199?thread_ts=1653398273.719399&cid=C02D12TNYLS we can safely discard SLES11 support. We don't need to spend time on it, and we should remove any file or code related to it.

Removed. The diff is saved to a file, just in case.

Bischoff:
> We should make clear in the title that we are not only integrating retail in BV, but we are also refactoring it in CI. (Not just the note at the bottom of the description)

That's the problem. The changes in CI are minimal, it's just the code needed to make it work both in CI and in BV (passing extra arguments, etc.).

Now I gained experience about Retail and its problems, it would be worth to reorganize the code in CI based on that gained knowledge. I would reorder features and scenarios, cut the huge "retail and mass import" feature in two, and rework idempotency.

But as this was not the scope of this PR, it's not done. I did not refactor the CI other than the minimum changes or a few easy-to-do cosmetics.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
```diff
@@ -121,7 +121,7 @@ Feature: PXE boot a Retail terminal
     And I click on "Create Group"
     Then I should see a "System group HWTYPE:Intel-Genuine created." text
 
-  Scenario: Create terminal branch group
+  Scenario: Create branch terminals group
```
nodeg, line 124: So we create a terminal group here. Why do we use the word `branch` here?

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -434,23 +434,20 @@
...
-    step %(I select "#{$build_host.full_hostname}" from "#{hostname}")
+  step %(I enter "#{system_name}" as "#{field}")
+end
+
+When(/^I select the hostname of "([^"]*)" from "([^"]*)"$/) do |host, field|
+  begin
+    node = get_target(host)
+  rescue
+    # don't try to select a non-existing system
+    return
```
nodeg, line 448: Shouldn`t we raise or log something here?

testsuite/features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature
```diff
@@ -0,0 +1,65 @@
...
+  Scenario: Clean up sumaform leftovers on a SLES 12 SP5 build host
+    When I perform a full salt minion cleanup on "sle12sp5_buildhost"
+
+  Scenario: Bootstrap the SLES 12 SP5 build host
+    When I bootstrap minion client "sle12sp5_buildhost" using bootstrap script with activation key "1-sle12sp5_buildhost_key" from the proxy
+    And I wait at most 10 seconds until Salt master sees "sle12sp5_buildhost" as "unaccepted"
+    And I accept "sle12sp5_buildhost" key in the Salt master
+    And I wait until onboarding is completed for "sle12sp5_buildhost"
+
+  Scenario: Turn the SLES 12 SP5 build host into a OS image build host
```
nodeg, line 22:
```suggestion
  Scenario: Turn the SLES 12 SP5 build host into an OS image build host
```

testsuite/features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature
```diff
@@ -0,0 +1,65 @@
...
+  Scenario: Work around issue https://github.com/SUSE/spacewalk/issues/10360
+    When I let Kiwi build from external repositories
+
+  Scenario: Build Kiwi image for SLES 12 SP5
+    When I follow the left menu "Images > Build"
+    And I select "suse_os_image_12" from "profileId"
+    And I select the hostname of "sle12sp5_buildhost" from "buildHostId"
+    And I click on "submit-btn"
+
+  Scenario: Check the OS image built
```
nodeg, line 56:
```suggestion
  Scenario: Check the OS image build
```
or
```suggestion
  Scenario: Check that the OS image built
```

testsuite/features/build_validation/retail/sle12sp5_terminal_prepare.feature
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2021-2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@sle12sp5_terminal
+Feature: Prepare prerequisities for SLE12 SP5 terminal deployment
```
nodeg, line 5:
```suggestion
Feature: Prepare prerequisites for SLE12 SP5 terminal deployment
```

testsuite/features/build_validation/retail/sle12sp5_terminal_prepare.feature
```diff
@@ -0,0 +1,56 @@
...
+
+  Scenario: Prepare activation keys for SLE12 SP5 retail systems
```
nodeg, line 7:
```suggestion
  Scenario: Prepare activation keys for SLE 12 SP5 retail systems
```
To be consistent with your other code.

testsuite/features/build_validation/retail/sle15sp3_buildhost_build_kiwi_image.feature
```diff
@@ -0,0 +1,65 @@
...
+  Scenario: Clean up sumaform leftovers on a SLES 15 SP3 build host
+    When I perform a full salt minion cleanup on "sle15sp3_buildhost"
+
+  Scenario: Bootstrap the SLES 15 SP3 build host
+    When I bootstrap minion client "sle15sp3_buildhost" using bootstrap script with activation key "1-sle15sp3_buildhost_key" from the proxy
+    And I wait at most 10 seconds until Salt master sees "sle15sp3_buildhost" as "unaccepted"
+    And I accept "sle15sp3_buildhost" key in the Salt master
+    And I wait until onboarding is completed for "sle15sp3_buildhost"
+
+  Scenario: Turn the SLES 15 SP3 build host into a OS image build host
```
nodeg, line 22:
```suggestion
  Scenario: Turn the SLES 15 SP3 build host into an OS image build host
```

testsuite/features/build_validation/retail/sle15sp3_buildhost_build_kiwi_image.feature
```diff
@@ -0,0 +1,65 @@
...
+  Scenario: Work around issue https://github.com/SUSE/spacewalk/issues/10360
+    When I let Kiwi build from external repositories
+
+  Scenario: Build Kiwi image for SLES 15 SP3
+    When I follow the left menu "Images > Build"
+    And I select "suse_os_image_15" from "profileId"
+    And I select the hostname of "sle15sp3_buildhost" from "buildHostId"
+    And I click on "submit-btn"
+
+  Scenario: Check the OS image built
```
nodeg, line 56: Same as above:

```suggestion
  Scenario: Check the OS image build
```
or
```suggestion
  Scenario: Check that the OS image has been built
```

testsuite/features/build_validation/retail/sle15sp3_terminal_prepare.feature
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2021-2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@sle15sp3_terminal
+Feature: Prepare prerequisities for SLE15 SP3 terminal deployment
```
nodeg, line 5:
```suggestion
Feature: Prepare prerequisites for SLE15 SP3 terminal deployment
```

nodeg, line 5:
```suggestion
Feature: Prepare prerequisites for SLE15 SP3 terminal deployment
```

nodeg (changes requested):
I left some suggestions and annotations.

testsuite/features/step_definitions/navigation_steps.rb
nodeg:
```bash
Offenses:

features/step_definitions/navigation_steps.rb:446:3: C: Style/RescueStandardError: Avoid rescuing without specifying an error class.
  rescue
  ^^^^^^
```
RuboCop complains, too :wink:

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: because imagine you have two branches:
- branch Seattle has terminals cash1, cash2, selfservice1, and selfservice2
- branch Miama has terminals cash1, cash2, cash3 and cash4

You would create two groups, one per branch, each with 4 terminals.

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: yes I can log something like "Ignoring" will do

as elsewhere, we will ignore this ruby cop :-P

nodeg: Ok but what about just logging this? Then we have some feedback and RuboCop is satisfied. Otherwise we have to adjust the RuboCop todo file to include this.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
nodeg: Got it, thank you!

testsuite/features/step_definitions/navigation_steps.rb
Bischoff:
```
 --- a/testsuite/.rubocop_todo.yml
+++ b/testsuite/.rubocop_todo.yml
@@ -560,6 +560,7 @@ Style/RegexpLiteral:
 Style/RescueStandardError:
   Exclude:
     - 'features/step_definitions/api_common.rb'
+    - 'features/step_definitions/navigation_steps.rb'
     - 'features/support/cobbler_test.rb'
 
 Style/SafeNavigation:
 ```
```
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -444,7 +444,7 @@ When(/^I select the hostname of "([^"]*)" from "([^"]*)"$/) do |host, field|
   begin
     node = get_target(host)
   rescue
-    # don't try to select a non-existing system
+    log "#{host} is not deployed, not trying to select its hostname from the list"
     return
   end
   step %(I select "#{node.full_hostname}" from "#{field}")
```

Comment (Bischoff) [reply testsuite/features/step_definitions/navigation_steps.rb:448]: Ideally, we should have two steps, or an option `ignoring if missing`, for the proxy case.

Comment (Bischoff) [reply testsuite/features/step_definitions/navigation_steps.rb:448]: Done, I added a `if present` option.

File: testsuite/features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature (hunk 11 above)
Comment (Bischoff) [reply testsuite/features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature:56]: Nope, it checks the image, and the image has been built.

If you really don't like it, I can use `Check the OS image just built" or `Check the image that has been built`.

File: testsuite/features/build_validation/retail/sle12sp5_terminal_prepare.feature (hunk 12 above)
Comment (Bischoff) [reply testsuite/features/build_validation/retail/sle12sp5_terminal_prepare.feature:5]: ok, it looks like we have a mess:
```
ebischoff@perso:~/SCM/github/Bischoff/uyuni> grep -R "SLES12" testsuite/ | egrep -v "profile|ENV" | wc -l
28
ebischoff@perso:~/SCM/github/Bischoff/uyuni> grep -R "SLES 12" testsuite/ | egrep -v "profile|ENV" | wc -l
66
ebischoff@perso:~/SCM/github/Bischoff/uyuni> grep -R "SLE 12" testsuite/ | egrep -v "profile|ENV" | wc -l
2
ebischoff@perso:~/SCM/github/Bischoff/uyuni> grep -R "SLE12" testsuite/ | egrep -v "profile|ENV" | wc -l
9
```

Comment (Bischoff) [reply testsuite/features/build_validation/retail/sle12sp5_terminal_prepare.feature]: Done.

File: testsuite/features/build_validation/retail/sle15sp3_buildhost_build_kiwi_image.feature (hunk 14 above)
Comment (Bischoff) [reply testsuite/features/build_validation/retail/sle15sp3_buildhost_build_kiwi_image.feature]: done

File: testsuite/features/build_validation/retail/sle15sp3_buildhost_build_kiwi_image.feature (hunk 15 above)
Comment (Bischoff) [reply testsuite/features/build_validation/retail/sle15sp3_buildhost_build_kiwi_image.feature:56]: no, see above

File: testsuite/features/build_validation/retail/sle15sp3_terminal_prepare.feature (hunk 16 above)
Comment (Bischoff) [reply testsuite/features/build_validation/retail/sle15sp3_terminal_prepare.feature]: urgh, fixed

Comment (Bischoff) [reply testsuite/features/build_validation/retail/sle15sp3_terminal_prepare.feature]: fixed

File: testsuite/features/build_validation/retail/sle12sp5_terminal_prepare.feature [hunk 17]
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2021-2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@sle12sp5_terminal
+Feature: Prepare prerequisities for SLES 12 SP5 terminal deployment
```
nodeg, line 5:
```suggestion
Feature: Prepare prerequisites for SLES 12 SP5 terminal deployment
```

nodeg (approved):

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -434,23 +434,24 @@
...
+  else
+    begin
+      node = get_target(host)
+    rescue
+      log "#{host} is not deployed, not trying to select it"
+      return
+    end
   end
+  step %(I select "#{node.full_hostname}" from "#{field}")
 end
```
srbarrios, line 455: I think this step must be re-think, the algorithm is not readable to me.
I don't have time now to think about a proposal, let me some days, and we can re-discuss it.

Bischoff: if there's a way to continue normally from the `rescue` block, then that would be more readable, yes

another way would be not to use `get_target` but i'm not that warm for that because `get_target` is what we have everywhere.

or any other solution you find.

Bischoff: ah, you can exit a `rescue` simply by doing nothing. Is it better now?

Bischoff: Is it better now?

testsuite/features/step_definitions/retail_steps.rb
Bischoff: It's definitely more readable this way.

PR #4526, opened 2021-12-01
Container-based Proxy

ncounter:
### Proxy as a containerized application

Based on the [RFC design](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00083-containerized-proxy.md#detailed-design) , this PR contains the implementation of the Proxy as a containerized application.

Containers are designed to be 4:
- `squid`
- `salt-broker`
- `httpd` and Proxy-specific Python WSGI applications
- `tftpd`

An additional container not planned in the initial RFC, but added to the picture to let bootstrapping client works:
- `sshd`

Everything related to containers can be found under `/containers` path of this project, like [the branch](https://github.com/uyuni-project/uyuni/tree/container-proxy/containers) this PR is based on

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/frontend/xmlrpc/proxy/ProxyHandler.java
```diff
@@ -146,10 +145,6 @@ public int activateProxy(String clientcert, String version)
             throw new ProxyAlreadyRegisteredException();
         }
 
-        if (!(server.hasEntitlement(EntitlementManager.MANAGEMENT) || server.hasEntitlement(EntitlementManager.SALT))) {
```
moio, line 149: This change should not be needed, as the XMLRPC API will not be used.

containers/proxy-httpd/uyuni-configure.py
```diff
@@ -0,0 +1,58 @@
...
+# read from file
+with open(config_path + "config.yaml") as source:
+    config = yaml.safe_load(source)
+    
+    # Save the systemid file
+    with open("/etc/sysconfig/rhn/systemid", "w") as file:
+        file.write(str(config['system_id']))
+    
+    # Decode and install SSL CA certificate
+    os.system(f"echo -n '{config['binary_ca_certs']}' | base64 -d > /etc/pki/trust/anchors/RHN-ORG-TRUSTED-SSL-CERT")
```
moio, line 17: I'd recommend base64 decoding and file writing to happen in Python, the intention being to shell out (eg. call `system()`) only when necessary.

containers/proxy-httpd/uyuni-configure.py
```diff
@@ -0,0 +1,58 @@
...
+    os.symlink("/etc/pki/trust/anchors/RHN-ORG-TRUSTED-SSL-CERT", "/usr/share/rhn/RHN-ORG-TRUSTED-SSL-CERT")
+    os.system("/usr/sbin/update-ca-certificates")
+
+    # Decode and install server certificate package
+    os.system(f"echo -n '{config['binary_srv_cert_rpm']}' | base64 -d > /root/rhn-org-httpd-ssl-key-pair-proxy-1.0-3.noarch.rpm")
+    os.system('rpm -Uv /root/rhn-org-httpd-ssl-key-pair-proxy-1.0-3.noarch.rpm')
```
moio, line 23: What does this RPM contain? Do we need the content to be RPM packaged at all?

ncounter: The package contains:

- `/etc/apache2/ssl.crt/server.crt`
- `/etc/apache2/ssl.csr/server.csr`
- `/etc/apache2/ssl.key/server.key`
- `/etc/pki/spacewalk/jabberd/server.pem`

The last one is useless so I dropped it. I also refactored to get rid of the rpm and we now have 3 parameters in the `config.yaml` with the encoded value of each file, to be decoded and used in the `uyuni-configure.py` script.

containers/proxy-httpd-image/uyuni-configure.py
```diff
@@ -0,0 +1,74 @@
...
+    with open("/etc/pki/trust/anchors/RHN-ORG-TRUSTED-SSL-CERT", "w") as file:
+        file.write(base64.b64decode(config['binary_ca_certs']).decode())
+    os.symlink("/etc/pki/trust/anchors/RHN-ORG-TRUSTED-SSL-CERT", "/usr/share/rhn/RHN-ORG-TRUSTED-SSL-CERT")
+    os.system("/usr/sbin/update-ca-certificates")
+
+    # Decode and install server certificate files
+    with open("/etc/apache2/ssl.crt/server.crt", "w") as file:
+        file.write(base64.b64decode(config['binary_server_crt']).decode())
+    with open("/etc/apache2/ssl.csr/server.csr", "w") as file:
+        file.write(base64.b64decode(config['binary_server_csr']).decode())
```
aaannz, line 28: This can be removed

containers/proxy-salt-broker-image/prepare_target.sh
```diff
@@ -0,0 +1,25 @@
...
+mv /etc/salt/broker /target/etc/salt
+
+mkdir -p /target/usr/share/rhn/config-defaults
+mv /usr/share/rhn/config-defaults/rhn*.conf /target/usr/share/rhn/config-defaults
+
+mkdir -p /target/usr/bin
+mv /usr/bin/salt-broker /target/usr/bin/
+mv /usr/bin/uyuni-configure.py /target/usr/bin/
+
+SITE_PACKAGES=usr/lib/$(ls /usr/lib/ | grep python)/site-packages
```
aaannz, line 19: Are you sure this will always return only one directory?

containers/proxy-tftpd-image/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+LABEL org.opencontainers.image.url="https://www.uyuni-project.org/"
+LABEL org.opencontainers.image.version="%%PKG_VERSION%%"
+LABEL org.openbuildservice.disturl="%DISTURL%"
+LABEL org.opensuse.reference="registry.opensuse.org/uyuni/proxy-tftpd:%PKG_VERSION%.%RELEASE%"
+# endlabelprefix
+
+# tftp
+EXPOSE 69/udp
+
+VOLUME "/etc/uyuni"
```
aaannz, line 26: tftpd container should have access to the same `/srv/tftpboot` volume as http proxy has.

http container will prepare it (cobbler wsgi scripts will) and tftp container then distribute it over tftp to pxe boot clients.

containers/proxy-tftpd-image/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# TODO: check installation
+
+RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses tftp
+
+# TODO: add RUN directive
```
aaannz, line 32: `/usr/sbin/in.tftpd -u $TFTP_USER -s $TFTP_DIRECTORY $TFTP_OPTIONS`

where defaults TFTP_USER is tftp, TFTP_DIRECTORY is `/srv/tftpboot` and options are generally for listening address but also timeout value (see https://bugzilla.suse.com/show_bug.cgi?id=1176177 )

containers/run-proxy-with-services.sh
```diff
@@ -0,0 +1,77 @@
...
+
+###
+# POD
+###
+podman pod create --name proxy-pod \
+  --publish 22:22 \
+  --publish 8080:8080 \
+  --publish 443:443 \
+  --publish 4505:4505 \
+  --publish 4506:4506 \
```
aaannz, line 32: missing `--publish 69:69/udp` for tftp. Also do you intentionally don't specify udp and tcp ports?

containers/run-proxy.sh
```diff
@@ -0,0 +1,58 @@
...
+do
+  podman pull --tls-verify=false $REGISTRY/$image
+done
+
+podman pod create --name proxy-pod \
+  --publish 22:22 \
+  --publish 8080:8080 \
+  --publish 443:443 \
+  --publish 4505:4505 \
+  --publish 4506:4506 \
```
aaannz, line 29: Same here `--publish 69:69/udp` for tftp.

containers/proxy-salt-broker-image/prepare_target.sh
moio: It should, at least in the limited context of the container build.

Regardless, let me clarify the context in which I wrote this (admittedly in half-hacked way): this was an attempt to try multi-staged builds as a technique to reduce image size.

It was not meant to be the one solution to be promoted in general, or in this specific image case - just an experiment to see if it was viable especially in the context of the Open Build Service. The experiment itself succeeded in that it did not highlight any OBS-specific problem, but it did not necessarily give clarity on what to do with the other images. One clear disadvantage is that the script here might not be trivial and might turn out to be brittle.

So FMPOV it is still to be determined whether this is the preferred way to trim the image size. Adjusting packages is an alternative option, and forcibly removing packages during the image build is yet another one. Even removing files directly is possible. All these possibilities (and maybe others escaping my attention) should be evaluated when we look at image trimming.

Until that time, I suggest to either leave the code here alone or remove the multi staged build altogether if it helps with review clarity - reverting later is always easily possible.

HTH

CC @rjmateus

rjmateus: In my opinion, at this stage, we should keep it as simple as possible, and tackle it when we got to the stage of reducing image size.
However, is this code is working now, and we don't expect any major problem with it, then just let it stay and it can be refactored latter.

ktsamis (approved):
Fine for me to merge, we talked and it should not cause any more turbulence in the testsuites than we have now. We will see Monday if that's true or not

PR #4529, opened 2021-12-02
Jsp date cleanup

hustodemon:
## What does this PR change?

1. unify displaying date/time in JSPs (now `rhn:formatDate` is used everywhere), this consists of 2 parts:
    1. remove the raw displaying of data (some JSPs used pre-formatted strings from backend, which they then displayed as-is using `c:out`)
    1. remove the usages of `fmt:formatDate` JSP tag
1. remove the ReactJS glue code for wrapping `.legacy-date-time` spans. I believe this is not needed (could you double-check, @Etheryte?)
1. make the `rhn:formatDate` use the so called "rhn custom dateformat" (i.e. `yyyy-MM-dd HH:mm:ss z`). The reason for this change is, that it's closer to what ReactJS frontend uses (see `localizedMoment.ts:toUserDateTimeString` and `UserPreferenceUtils.getUserDate/TimeFormat`), so now this is consistent. We should create a follow-up card for this if customization of this is desired.
1. adjust the `humanizeDates` in `spacewalk-essentials.js` to use the format that is closer to what ReactJS part does.

## GUI diff

Before:
ss-before

After:
- Date in ISO format (previously respected the language settings (JSP only))
- Localization respected (previously, the humanized date was in English, even if the language was set to something different)

ss-after

- Full date in ISO format (previously was locale-specific)
ss

- [x] **DONE**

## Documentation
- No documentation needed: this PR is more or less a bugfix (making displaying date/time consistent, removing legacy code)

- [x] **DONE**

## Test coverage
- No tests: there are no tests for jsp

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/15845

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


Etheryte (approved):
Nice, this looks very nice. I also checked and indeed with this I think removing the legacy Javascript binding there should be completely fine, which makes this a double win. Thanks a lot for digging through the jsp fiesta. 🍻

ncounter (approved):
:star_struck:

java/code/src/com/redhat/rhn/common/db/datasource/xml/Package_queries.xml
```diff
@@ -263,7 +263,7 @@ SELECT C.name,
        CS.checksum_type checksumtype,
        F.file_mode,
        F.linkto,
-       TO_CHAR(F.mtime, 'YYYY-MM-DD HH24:MI:SS') AS MTIME
```
parlt91, line 266: These seem to cause some problems with the time displayed (because we cut of the timezone information here) if server and user preferred timezone are different. So we should probably remove all of these instances. I already took care of 2 other of these in https://github.com/uyuni-project/uyuni/pull/4496. But we can create a separate issue to take care of these.

parlt91 (approved):
LGTM. Just one comment below.

java/code/src/com/redhat/rhn/common/db/datasource/xml/Package_queries.xml
hustodemon: :+1:  Yes, let's do it one after another.

cbbayburt (approved):

mcalmer:
https://ci.nue.suse.com/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/1015

PR #4562, opened 2021-12-14
Add hash to venv enabled txt

vzhestkov:
## What does this PR change?

Appends sha256 hash to `venv-enabled-ARCH.txt` files generated with `mgr-create-bootstrap-repo`.
The file is used to detect if the bootstrap repo contains Salt Bundle.

The hash is required for checking if we need to update Salt Bundle on salt-ssh managed systems.

## GUI diff

No difference.

## Documentation

## Changelogs

- Add sha256 hash to venv-enabled-ARCH.txt file generated with mgr-create-bootstrap-repo

## Re-run a test

### DIFF
--- a/susemanager/src/mgr-create-bootstrap-repo
+++ b/susemanager/src/mgr-create-bootstrap-repo
@@ -1,12 +1,14 @@
 #!/usr/bin/python3
 
-import sys
-import os
 import glob
+import hashlib
+import os
 import shutil
 import subprocess
+import sys
 import time
 import traceback
+
 from optparse import OptionParser
 from rhn import rhnLockfile
 from spacewalk.server import rhnSQL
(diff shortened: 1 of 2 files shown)


susemanager/src/mgr-create-bootstrap-repo
```diff
@@ -585,9 +587,16 @@ def create_repo(label, options, mgr_bootstrap_data, additional=[]):
             rel_path = os.path.relpath(file_path, start=destdirtmp)
             (l_path, ext) = rel_path.rsplit('.', 1)
             if ext:
+                dg = hashlib.sha256()
+                with open(file_path, "rb") as pkg_fh:
+                    while True:
+                        buff = pkg_fh.read(0x1000)
+                        if not buff:
+                            break
+                        dg.update(buff)
```
meaksh, line 596: Just a suggestion here to make this cleaner:

```suggestion
                dg = hashlib.sha256(open(file_path,'rb').read())
```

meaksh (commented):
Small suggestion. Other than that looks good :+1:

susemanager/src/mgr-create-bootstrap-repo
vzhestkov: It's cleaner, but will read whole file at once and it's ~20-30Mb, it shouldn't be a problem but is it a really good practice?

meaksh: Hmmm, I see your point. I don't think 20-30Mb will cause any problem here, as the file is not really that big, but since now I see the reason why you implemented that way, I am perfectly fine if we just keep the code as you did :)

Thanks!

meaksh (approved):

susemanager/src/mgr-create-bootstrap-repo
agraul: Calling `open()` outside of a `with` statement and without calling `close()` later on is not cleaner IMO. I prefer Victor's approach even if everything is read at once with `.read()`.

PR #4564, opened 2021-12-14
Testsuite: Test for changing proxy

nadvornik:
## What does this PR change?

This is a test for moving minion and ssh minion between direct connection and a proxy.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: only testsuite
- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Test for https://github.com/uyuni-project/uyuni/pull/4313

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/run_sets/secondary.yml
+++ b/testsuite/run_sets/secondary.yml
@@ -65,4 +65,6 @@
 - features/secondary/proxy_cobbler_pxeboot.feature
 - features/secondary/srv_restart.feature
 - features/secondary/min_timezone.feature
+- features/secondary/min_move_from_and_to_proxy.feature
+- features/secondary/minssh_move_from_and_to_proxy.feature
 ## Secondary features END ##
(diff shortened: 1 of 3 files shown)


Bischoff (approved):
LGTM ; if it's not for head only, please don't forget the ports.

testsuite/run_sets/secondary.yml
```diff
@@ -58,5 +58,8 @@
 - features/secondary/min_change_software_channel.feature
 - features/secondary/min_retracted_patches.feature
 - features/secondary/proxy_cobbler_pxeboot.feature
-
+- features/secondary/min_move_from_proxy.feature
+- features/secondary/min_move_to_proxy.feature
```
srbarrios, line 62: I see that as a unique feature, I think that for maintenance and for organization of features and their strict order, it will be better to join these two in one.

Bischoff: From a conceptual point of view both split in one or two features make sense to me.

BUT

From a practical point of view we can probably remove cleanup and preparation steps and therefore win execution time if we merge into one feature.

So yes, I think Oscar's suggestion makes sense.

nadvornik: That's what I tried first, but then the same events appeared twice and "I wait until event" on the second event incorrectly matched the first one. Is there any workaround for this?

PR #4565, opened 2021-12-14
Revert "QA: Disable IPv4 forwarding on branch formula"

ktsamis:
Reverts uyuni-project/uyuni#4556

- [x] No changelog needed


Bischoff (approved):
Approving, but would it be possible to only revert the ping test then?

srbarrios (approved):

PR #4566, opened 2021-12-14
QA: Disable IPv4 forwarding on branch formula

srbarrios:
## What does this PR change?

Related card: https://github.com/SUSE/spacewalk/issues/16486

We want to make the systems in the private network to not have access to any external server outside the private network. This way catch this kind of issues earlier.

For now, this PR prevents this communication if the systems are configured with IPv4.

Notes:
- The "ip" attribute in our twopence nodes was not including the private IP, but the domain name.
- We need to specify the use of a concrete network when we do our ping, otherwise, it tries to use the other IP.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Ports:
- Manager-4.1
- Manager-4.2

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/build_validation/core/allcli_sanity.feature
+++ b/testsuite/features/build_validation/core/allcli_sanity.feature
@@ -34,285 +34,285 @@ Feature: Sanity checks
   Scenario: The proxy is healthy
     Then "proxy" should have a FQDN
     And reverse resolution should work for "proxy"
-    And "proxy" should communicate with the server
+    And "proxy" should communicate with the server using "eth0"
     And the clock from "proxy" should be exact
 
 @sle11sp4_client
   Scenario: The SLES 11 SP4 traditional client is healthy
     Then "sle11sp4_client" should have a FQDN
     And reverse resolution should work for "sle11sp4_client"
-    And "sle11sp4_client" should communicate with the server
+    And "sle11sp4_client" should communicate with the server using "eth0"
     And the clock from "sle11sp4_client" should be exact
 
 @sle11sp4_minion
   Scenario: The SLES 11 SP4 minion is healthy
     Then "sle11sp4_minion" should have a FQDN
     And reverse resolution should work for "sle11sp4_minion"
-    And "sle11sp4_minion" should communicate with the server
+    And "sle11sp4_minion" should communicate with the server using "eth0"
     And the clock from "sle11sp4_minion" should be exact
 
 @sle11sp4_ssh_minion
   Scenario: The SLES 11 SP4 Salt SSH minion is healthy
     Then "sle11sp4_ssh_minion" should have a FQDN
     And reverse resolution should work for "sle11sp4_ssh_minion"
-    And "sle11sp4_ssh_minion" should communicate with the server
+    And "sle11sp4_ssh_minion" should communicate with the server using "eth0"
     And the clock from "sle11sp4_ssh_minion" should be exact
 
 @sle12sp4_client
   Scenario: The SLES 12 SP4 traditional client is healthy
     Then "sle12sp4_client" should have a FQDN
     And reverse resolution should work for "sle12sp4_client"
-    And "sle12sp4_client" should communicate with the server
+    And "sle12sp4_client" should communicate with the server using "eth0"
     And the clock from "sle12sp4_client" should be exact
 
 @sle12sp4_minion
   Scenario: The SLES 12 SP4 minion is healthy
     Then "sle12sp4_minion" should have a FQDN
     And reverse resolution should work for "sle12sp4_minion"
-    And "sle12sp4_minion" should communicate with the server
+    And "sle12sp4_minion" should communicate with the server using "eth0"
     And the clock from "sle12sp4_minion" should be exact
 
 @sle12sp4_ssh_minion
   Scenario: The SLES 12 SP4 Salt SSH minion is healthy
     Then "sle12sp4_minion" should have a FQDN
     And reverse resolution should work for "sle12sp4_minion"
-    And "sle12sp4_minion" should communicate with the server
+    And "sle12sp4_minion" should communicate with the server using "eth0"
     And the clock from "sle12sp4_minion" should be exact
 
 @sle12sp5_client
   Scenario: The SLES 12 SP5 traditional client is healthy
     Then "sle12sp5_client" should have a FQDN
     And reverse resolution should work for "sle12sp5_client"
-    And "sle12sp5_client" should communicate with the server
+    And "sle12sp5_client" should communicate with the server using "eth0"
     And the clock from "sle12sp5_client" should be exact
 
 @sle12sp5_minion
   Scenario: The SLES 12 SP5 minion is healthy
     Then "sle12sp5_minion" should have a FQDN
     And reverse resolution should work for "sle12sp5_minion"
-    And "sle12sp5_minion" should communicate with the server
+    And "sle12sp5_minion" should communicate with the server using "eth0"
     And the clock from "sle12sp5_minion" should be exact
 
 @sle12sp5_ssh_minion
   Scenario: The SLES 12 SP5 Salt SSH minion is healthy
     Then "sle12sp5_minion" should have a FQDN
     And reverse resolution should work for "sle12sp5_minion"
-    And "sle12sp5_minion" should communicate with the server
+    And "sle12sp5_minion" should communicate with the server using "eth0"
     And the clock from "sle12sp5_minion" should be exact
 
 @sle15_client
   Scenario: The SLES 15 traditional client is healthy
     Then "sle15_client" should have a FQDN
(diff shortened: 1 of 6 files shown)


Etheryte:
Sorry for the slight off-topic, why is @uyuni-project/frontend automatically tagged as a reviewer here? None of the files are owned by that group?

agraul:
> Sorry for the slight off-topic, why is @uyuni-project/frontend automatically tagged as a reviewer here? None of the files are owned by that group?

I have the same question for @uyuni-project/python, this is weird.

testsuite/features/core/proxy_branch_network.feature
```diff
@@ -205,5 +205,5 @@ Feature: Setup Uyuni for Retail branch network
 @proxy
 @private_net
   Scenario: The terminals should not reach the server
-    Then "sle_client" should not communicate with the server
-    And "sle_minion" should not communicate with the server
+    Then "sle_client" should not communicate with the server using "eth1"
+    And "sle_minion" should not communicate with the server using "eth1"
```
Bischoff, line 209: it should also not communicate using `eth0`...

Also, this is fragile. If you remember, we considered several times simplifying the network setup by having only one card on client and minion. That would fail with this new syntax.

testsuite/features/build_validation/core/allcli_sanity.feature
```diff
@@ -34,285 +34,285 @@ Feature: Sanity checks
   Scenario: The proxy is healthy
     Then "proxy" should have a FQDN
     And reverse resolution should work for "proxy"
-    And "proxy" should communicate with the server
+    And "proxy" should communicate with the server using "eth0"
```
Bischoff, line 37: a bit repetitive, isn't it?

testsuite/features/core/allcli_sanity.feature
```diff
@@ -34,63 +34,63 @@ Feature: Sanity checks
   Scenario: The proxy is healthy
     Then "proxy" should have a FQDN
     And reverse resolution should work for "proxy"
-    And "proxy" should communicate with the server
+    And "proxy" should communicate with the server using "eth0"
```
Bischoff, line 37: A bit repetitive, isn't it?

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -23,16 +23,16 @@
   raise "reverse resolution for #{node.ip} returned #{result}, expected to see #{node.full_hostname}" unless result.include? node.full_hostname
 end
 
-Then(/^"([^"]*)" should communicate with the server$/) do |host|
+Then(/^"([^"]*)" should communicate with the server using "([^"]*)"/) do |host, ethernet|
   node = get_target(host)
-  node.run("ping -c1 #{$server.ip}")
-  $server.run("ping -c1 #{node.ip}")
+  node.run("ping -c 1 -I #{ethernet} #{$server.ip}")
```
Bischoff, line 28: Is `-I #{ethernet}` really needed? I think we could just let the normal routing process decide which card is used.

But there is something really missing here. You should use `-4` to force IPv4.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -23,16 +23,16 @@
...
+  $server.run("ping -c 1 #{node.ip}")
```
Bischoff, line 29: same, `-4`.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -23,16 +23,16 @@
...
 end
 
-Then(/^"([^"]*)" should not communicate with the server$/) do |host|
+Then(/^"([^"]*)" should not communicate with the server using "([^"]*)"/) do |host, ethernet|
```
Bischoff, line 32: Same comment about the new parameter.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -170,7 +170,7 @@
 end
 
 When(/^I uncheck (.*) box$/) do |checkbox_name|
-  check BOX_IDS[checkbox_name]
+  uncheck BOX_IDS[checkbox_name]
```
Bischoff, line 173: oops :-)

testsuite/features/support/twopence_init.rb
```diff
@@ -335,11 +335,15 @@ def client_public_ip(host)
...
 
-  node.init_ip(node.full_hostname)
-
   host = $host_by_node[node]
   raise "Cannot resolve host for node: '#{node.hostname}'" if host.nil? || host == ''
 
+  if ADDRESSES.has_key? host
+    node.init_ip(net_prefix + ADDRESSES[host])
+  else
+    node.init_ip(node.full_hostname)
```
Bischoff, line 344: why the f*** were we doing that? ;-)

Bischoff (changes requested):
I don't like the new parameter, I would rather let ping do its best to contact the server. It can infer the interface to use from the routing table.

But one thing that is really missing is the `-4` parameter to `ping`.

Let's discuss it if you disagree or if I missed something.

srbarrios:
@Etheryte @agraul you are right guys, I experienced the same with qa group in several PRs.
I think something has changed on the GitHub configuration that handle this. Not sure who can help with this now, I did not worked on that, maybe @renner knows about it?

testsuite/features/step_definitions/command_steps.rb
srbarrios: Yes, otherwise it was using the other ethernet, the public one, and the ping was successful.

testsuite/features/core/proxy_branch_network.feature
srbarrios: Well, so then we have a problem with our product, as it communicates successfully through eth0.
Could you please take care of that network issue?

testsuite/features/build_validation/core/allcli_sanity.feature
srbarrios: Well, what I can't do is to let the "should communicate" without the parameter, and the "should not communicate" without the parameter, I prefer to have the same format in both positive and negative cases.

testsuite/features/support/twopence_init.rb
srbarrios: So, we want to have in "ip" attribute an IP not a fqdn. Before this change, we were doing `node.init_ip(node.full_hostname)` but I want "ip" attribute of node to contain the private IP if that exist. As it only exists if it's contained in the hash ADDRESSES, I first check it, and if we have private IP I set it, if not I let as it was, with fqdn.

Keep in mind we also have, public_ip attribute, which contains the public ip in all the cases.

testsuite/features/step_definitions/command_steps.rb
ktsamis: I think if we use -4 here it hides the problem if we enable ipv6 and this happens again. Not sure but I had the same impression at first that we needed to use -4. If we specify the IP and that's an IPv4 then that's taken care of automatically and we don't lose potential future problems.

testsuite/features/build_validation/core/allcli_sanity.feature
Bischoff: I would remove the parameter in both cases.

testsuite/features/step_definitions/command_steps.rb
Bischoff: We don't address the problem for IPv6 now. In fact, we would need a ping for both protocols, IPv4 **and** IPv6.

But it makes little sense to test IPv6 before Ondrej addresses the issue SUSE/spacewalk#16487.

testsuite/features/support/twopence_init.rb
Bischoff: I think I remember Dario doing that long time ago, and I was telling him "don't call `ip` a variable that holds a FQDN"...

testsuite/features/step_definitions/command_steps.rb
srbarrios: I agree having two pings for IPv4 and IPv6, what it puzzles me is that we need to specify the card otherwise it connects using the public network. As shown above.

Bischoff (approved):
Approving after direct discussion with Oscar.

The problem is that we use two pseudo-terminals, "client" and "minion", with two networking cards each, one on the private network, and one on the public network.

What really interests us is not this kind of configuration. It's real terminals behind branch server with only one card. That's the ones that created the registration bug in the first place.

We said several times we wanted to simplify the topology and have only real terminals on the private network. But not now, even though I have a branch ready with those changes.

For now let's just go with specifying the network card then.

Also, testing IPv6 will still need to be done, later.

PR #4567, opened 2021-12-14
BV terminals have no reason to be reachable by network

Bischoff:
## What does this PR change?

This PR makes the test suite expect MAC addresses instead of host names for Retail terminals.

It also fixes a small problem for Debian 11.

## Links

Sumaform PR: uyuni-project/sumaform#1021

Ports:
* 4.1: SUSE/spacewalk#16551
* 4.2: SUSE/spacewalk#16548

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/twopence_init.rb
+++ b/testsuite/features/support/twopence_init.rb
@@ -72,11 +72,8 @@ if $build_validation
   $debian11_minion = twopence_init("ssh:#{ENV['DEBIAN11_MINION']}") if ENV['DEBIAN11_MINION']
   $debian11_ssh_minion = twopence_init("ssh:#{ENV['DEBIAN11_SSHMINION']}") if ENV['DEBIAN11_SSHMINION']
   $sle11sp4_buildhost = twopence_init("ssh:#{ENV['SLE11SP4_BUILDHOST']}") if ENV['SLE11SP4_BUILDHOST']
-  $sle11sp3_terminal = twopence_init("ssh:#{ENV['SLE11SP3_TERMINAL']}") if ENV['SLE11SP3_TERMINAL']
   $sle12sp5_buildhost = twopence_init("ssh:#{ENV['SLE12SP5_BUILDHOST']}") if ENV['SLE12SP5_BUILDHOST']
-  $sle12sp5_terminal = twopence_init("ssh:#{ENV['SLE12SP5_TERMINAL']}") if ENV['SLE12SP5_TERMINAL']
   $sle15sp3_buildhost = twopence_init("ssh:#{ENV['SLE15SP3_BUILDHOST']}") if ENV['SLE15SP3_BUILDHOST']
-  $sle15sp3_terminal = twopence_init("ssh:#{ENV['SLE15SP3_TERMINAL']}") if ENV['SLE15SP3_TERMINAL']
   $opensuse153arm_minion = twopence_init("ssh:#{ENV['OPENSUSE153ARM_MINION']}") if ENV['OPENSUSE153ARM_MINION']
   $nodes += [$sle11sp4_client, $sle11sp4_minion, $sle11sp4_ssh_minion,
              $sle12sp4_client, $sle12sp4_minion, $sle12sp4_ssh_minion,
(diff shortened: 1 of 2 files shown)


testsuite/features/support/twopence_init.rb
```diff
@@ -223,6 +220,9 @@ def file_inject(node, local_file, remote_file)
 # Other global variables
 $product = product
 $pxeboot_mac = ENV['PXEBOOT_MAC']
+$sle11sp3_terminal = ENV['SLE11SP3_TERMINAL']
```
ktsamis, line 223: So you are making the env variables common between CI and BV, but not adding them in the nodes? Is this deliberate or are the nodes missing in lines 34, 104 etc?

ktsamis: Does this fall in the node_by_host in 235 maybe so it's not needed?

Bischoff: They are not twopence objects ("nodes"), they are simple strings. You can't reach them via SSH.

testsuite/features/support/twopence_init.rb
```diff
@@ -314,9 +316,9 @@ def client_public_ip(host)
   end
 
   interface = case host
-              when /^sle/, /^opensuse/, /^ssh/, /^ceos/, /^debian/, 'server', 'proxy', 'build_host'
+              when /^sle/, /^opensuse/, /^ssh/, /^ceos/, /^debian9/, /^debian10/, 'server', 'proxy', 'build_host'
                 'eth0'
-              when /^ubuntu/
+              when /^debian11/, /^ubuntu/
```
srbarrios, line 321: Mixing PRs?

testsuite/features/support/twopence_init.rb
```diff
@@ -223,6 +225,9 @@ def file_inject(node, local_file, remote_file)
 # Other global variables
 $product = product
 $pxeboot_mac = ENV['PXEBOOT_MAC']
+$sle11sp3_terminal = ENV['SLE11SP3_TERMINAL']
+$sle12sp5_terminal = ENV['SLE12SP5_TERMINAL']
+$sle15sp3_terminal = ENV['SLE15SP3_TERMINAL']
```
srbarrios, line 230: Can I ask for a rename of this variable?
I would prefer xxx_terminal_mac so we can better distinguish these ones from the variables containing twopence objects.

testsuite/features/support/twopence_init.rb
Bischoff: Yes. It's in the description though.

testsuite/features/support/twopence_init.rb
Bischoff: ok. It would make sense also if we introduce `sle11sp3_terminal_image`.

srbarrios (approved):

testsuite/features/support/twopence_init.rb
```diff
@@ -143,18 +140,25 @@ def get_target(host)
...
+  case host
+  when 'pxeboot_minion'
     # The PXE boot minion is not directly accessible on the network,
     # therefore it is not represented by a twopence node
     output, _code = $server.run('salt-key')
     system_name = output.split.find do |word|
       word =~ /example.Intel-Genuine-None-/ || word =~ /example.pxeboot-/ || word =~ /example.Intel/ || word =~ /pxeboot-/
     end
     system_name = 'pxeboot.example.org' if system_name.nil?
+  when 'sle11sp3_terminal'
```
ktsamis, line 158: Wouldn't an or statement be better here instead of multiple when? I think this won't execute anything in the first 2 when cases. You need a comma in between https://stackoverflow.com/questions/10197254/case-statement-with-multiple-values-in-each-when-block?rq=1

testsuite/features/support/twopence_init.rb
```diff
@@ -143,18 +140,25 @@ def get_target(host)
...
+  when 'sle12sp5_terminal'
```
ktsamis, line 159: Where are these defined? I only see *_terminal_mac variables

testsuite/features/support/twopence_init.rb
```diff
@@ -143,18 +140,25 @@ def get_target(host)
 # * for the usual clients, it is the full hostname, e.g. hmu-centos.tf.local
 # * for the PXE booted clients, it is derived from the branch name, the hardware type,
 #   and a fingerprint, e.g. example.Intel-Genuine-None-d6df84cca6f478cdafe824e35bbb6e3b
+# rubocop:disable Metrics/MethodLength
+# rubocop:disable Metrics/AbcSize
```
ktsamis, line 144: Why are we skipping rubocop for these? Were there warnings?

testsuite/features/support/twopence_init.rb
Bischoff: it's fixed strings

testsuite/features/support/twopence_init.rb
Bischoff: yes rubocop was barfing

ktsamis (approved):
Don't forget to create the ports again, they need adjusting also

Bischoff:
> Don't forget to create the ports again, they need adjusting also

i'm on it

PR #4577, opened 2021-12-16
Allow additional db connections with hibernate

mackdk:
## What does this PR change?

This PR allows configuring multiple database connection with hibernate.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


mackdk:
@mcalmer @mbussolotto I rebased the branch and add a bit of refactoring to ensure we do not duplicate the code. I created this PR to review and discuss the change, and to run the unit and acceptance test to make sure we are not breaking anything.

java/code/src/com/redhat/rhn/common/hibernate/ReportDbConnectionManager.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+class ReportDbConnectionManager extends AbstractConnectionManager {
+
+    ReportDbConnectionManager() {
+        super(Collections.emptySet());
+    }
+
+    @Override
+    protected Properties getConfigurationProperties() {
+        final Properties hibProperties = Config.get().getNamespaceProperties("hibernate");
```
mackdk, line 36: On second look, this might be problematic. It means we are going to use the same hibernate properties for both database except for user, password and connection. Maybe we should use a different namespace for all the reporting properties. Also on the hub we should be able to create the connection on the fly for multiple servers, so it needs to be parameterised differently for that.

java/code/src/com/redhat/rhn/common/hibernate/AbstractConnectionManager.java
```diff
@@ -0,0 +1,334 @@
...
+            try {
+                if (session != null && session.isOpen()) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("YYY Closing Hibernate Session");
+                    }
+                    session.close();
+                }
+            }
+            catch (HibernateException e) {
+                throw new HibernateRuntimeException("couldn't close session");
```
mackdk, line 327: This ```throw``` here can be tricky, because it's within a ```finally``` block, so it can happen when the code already threw an exception, and if it does, it hides to original exception. We indeed do log it above in the ```log.error()```, but it might be not that easy to figure out that the exceptions are connected. Not sure if we can change this though, because it will change the existing behaviour, even if this scenario should rarely happen.

mcalmer:
Please add all changes to https://github.com/uyuni-project/uyuni/pull/4672
This branch and PR will be closed later

java/code/src/com/redhat/rhn/common/hibernate/ReportDbHibernateFactory.java
```diff
@@ -0,0 +1,873 @@
...
+import javax.persistence.FlushModeType;
+
+/**
+ * HibernateFactory - Helper superclass that contains methods for fetching and
+ * storing Objects from the DB using Hibernate.
+ * <p>
+ * Abstract methods define what the subclass must implement to determine what is
+ * specific to that Factory's instance.
+ */
+public class ReportDbHibernateFactory {
```
lucidd, line 56: If i understand this correctly this is just a copy of HibernateFactory with a few adjustments. Is there a plan to have most of the code shared later?

mackdk: It would be indeed better, but the problem is that HibernateFactory has both static and instance methods, it is extended by multiple factories and all of them share statically the same session. Currently HibernateFactory has a static ConnectionManager, while ReportDbHibernateFactory has an instance field for that. This might be problematic if we want to extend it or have multiple instances. Not sure though if it's something we will have in our use case.

Also, I think we can get rid of most of the duplication if we plan to use ReportDbHibernateFactory only need to execute mode queries.

mcalmer:
@mackdk i think we can close this PR, right?

PR #4588, opened 2021-12-18
Fixed mirror link usage for dnf

sbluhm:
## What does this PR change?

- Removed obsolete dnf issue workaround.
- Assume mirrorlists by default.
- Fixed issue of dnf excluding packages from active modules. (https://bugzilla.redhat.com/show_bug.cgi?id=2035577)

For reference:
flags `hawkey.IGNORE_REGULAR_EXCLUDES` or `hawkey.IGNORE_EXCLUDES`  could be used to ignore config file xecludes or all excludes if required/wanted with `self.dnfbase.sack.query()` (in yum_dnf_plugin.py)

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

Manual testing of impacted file: reposync with direct URL and mirrorlist.

- [X] **DONE**

## Links

Fixes #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


SchoolGuy:
@sbluhm You have a conflict in the changefile. I would propose a rebase.

sbluhm:
@SchoolGuy done

backend/satellite_tools/repo_plugins/yum_dnf_src.py
```diff
@@ -208,18 +208,16 @@ def setup_repo(self, repo, no_mirrors, ca_cert_file, client_cert_file, client_ke
         self.digest=hashlib.sha256(self.url.encode('utf8')).hexdigest()[:16]
         self.dnfbase.repos.add(repo)
         self.repoid = repo.id
-        try:
-            logger = logging.getLogger('dnf')
-            logger.setLevel(logging.ERROR)
-            self.yumbase.repos[self.repoid].load()
-            logger.setLevel(logging.WARN)
-        except RepoError:
```
vzhestkov, line 216: I have not checked the sources yet, but are you sure we can drop exception handling here while load() is still used?

sbluhm: I believe the exception was to test whether a mirror link is working. Previously it threw an exception when it wasn't a mirror link (at least as far as I understand from the inline comment).
Now dnf automatically reverts back to base URL and only displays a warning. So there is no exception. My thought is that this removes the purpose of the original exception handling (check if a mirror link can be resolved).

Is there/can there be a different purpose based on the handled code?

sbluhm: You are right, the try/catch block is required in case the URL or repo itself is not working. I will re-add it.

backend/satellite_tools/repo_plugins/yum_dnf_src.py
```diff
@@ -209,17 +209,16 @@ def setup_repo(self, repo, no_mirrors, ca_cert_file, client_cert_file, client_ke
...
             self.dnfbase.repos[self.repoid].load()
+            # Don't use mirrors if there are none.
+            if not self.clean_urls(self.dnfbase.repos[self.repoid]._repo.getMirrors()):
+                no_mirrors = True
+                # Reload repo just in case.
+                repo.mirrorlist = ""
+                self.dnfbase.repos[self.repoid].load()
+        except RepoError as exc:
+            raise RepoMDError(exc)
+            return
```
vzhestkov, line 221: Any chances to reach this `return`?

sbluhm: Always? Unless I misunderstand the code. Raise only raises the error but the programme continuous (or `RepoMDError` always continues?). At least that is what it looks like on my tests. So I had to put the `return`. Otherwise there will be more errors further down because the repo didn't load properly.

vzhestkov: Try to run this code:
```
try:
    a = 123/0
except ZeroDivisionError:
    raise Exception("testing")
    print("test")
```

`print("test")` will never be reached

sbluhm: I just re-tried without the return and the output is indeed the same. So as confused I am now, it looks like my testing was wrong....
Will remove it.
Thank you.

backend/satellite_tools/repo_plugins/yum_dnf_src.py
```diff
@@ -17,6 +17,7 @@
 
 import dnf
 import hashlib
+import hawkey
```
SchoolGuy, line 20: Is this a new dependency or do we use that already? I wasn't able to find this in the specfile.

SchoolGuy (commented):
I found a small thing which we might need to discuss.

backend/satellite_tools/repo_plugins/yum_dnf_src.py
sbluhm: Nice catch. hawkey is provided by python3-hawkey which in turn is already a requirement of python3-dnf. Python3-dnf won't work without it so it will always be installed if python3-dnf is.

As we explicitly use it, I guess we should also put it in the spec file. What do you think?

sbluhm: I just noticed that we also don't require python3-dnf or even dnf and just assume it is installed. Now I am not sure. Either we add all or none (dnf/python3-dnf/python3-hawkey).

SchoolGuy: Thanks a lot for addressing this. I consider this resolved.

backend/spacewalk-backend.spec
```diff
@@ -227,6 +227,8 @@ Requires:       %{name}-xmlrpc = %{version}-%{release}
 Requires:       systemd
 BuildRequires:  systemd
 %if 0%{?rhel}
+Requires:       python3-dnf
+Requires:       python3-hawkey
```
vzhestkov, line 231: Actually it's redundant as it seems `python3-hawkey` is already required by `python3-dnf`:
```
# rpm -q --requires python3-dnf
/usr/libexec/platform-python
dnf-data = 4.2.17-6.el8
libmodulemd >= 1.4.0
python(abi) = 3.6
python3-gpg
python3-hawkey >= 0.39.1
python3-libcomps >= 0.1.8
python3-libdnf
python3-libdnf >= 0.39.1
python3-rpm >= 4.14.2-35
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PartialHardlinkSets) <= 4.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
```

SchoolGuy: I quickly talked with @vzhestkov: It seems that hawkeye is now a part of `libdnf` and obsolete. Thus I believe a direct requirement is not needed indeed.

See: https://github.com/rpm-software-management/hawkey

backend/satellite_tools/repo_plugins/yum_dnf_src.py
```diff
@@ -202,17 +204,15 @@ def setup_repo(self, repo, no_mirrors, ca_cert_file, client_cert_file, client_ke
         self.dnfbase.repos.add(repo)
         self.repoid = repo.id
         try:
-            logger = logging.getLogger('dnf')
-            logger.setLevel(logging.ERROR)
-            self.yumbase.repos[self.repoid].load()
-            logger.setLevel(logging.WARN)
-        except RepoError:
-            # Dnf bug workaround. Mirrorlist was provided but none worked. Fallback to baseurl and load again.
```
SchoolGuy, line 210: Which DNF bug do you refer here to? As I personally don't have the context do we need to have a minimum dnf version in the specfile?

vzhestkov: If was exact a year ago I was wery confused to see the date on the first glance:
https://github.com/uyuni-project/uyuni/commit/6cea1af1cfc4f4df82d3f5951264f6efc71c9b8a

backend/spacewalk-backend.spec
vzhestkov: @sbluhm no need to list `python3-hawkey` as required, explained in the comment above

backend/satellite_tools/repo_plugins/yum_dnf_src.py
vzhestkov: @sbluhm which bug you mentioned in the initial commit from a year ago?

vzhestkov (changes requested):

sbluhm:
@vzhestkov 

> which bug you mentioned in the initial commit from a year ago?

I don't think I understood you.

When I wrote the original Plugin, dnf failed with a terminating error if mirrorlist was filled out despite a direct URL being filled out. 
As of now, dnf throws a warning and tries the direct URL as a next step (and then hopefully succeeds.

I don't think we need a minimum DNF version. EL8 support is not public and currently always requires an updated version. Leap currently also has the same DNF version.

Did this answer the question?

vzhestkov (approved):
Looks good.
Thanks @sbluhm

sbluhm:
@SchoolGuy , can you please review and merge? I would like to have this one in master. Thank you.

SchoolGuy (approved):
LGTM

PR #4596, opened 2021-12-20
Add checking for libvirtd binary to grains.virt

vzhestkov:
## What does this PR change?

In case of using Salt Bundle, `libvirt` python module is present in the salt bundle virtual environment, but `libvirt` could be not installed on the system. In this case any call for `grains` leads to writing the following message to the log or console:
```
[salt.loaded.ext.grains.virt:47  ][ERROR   ][2215] libvirtd is not installed or is not in the PATH
```

## GUI diff

No difference.

## Documentation

## Changelogs

- Add checking for libvirtd binary to grains.virt module

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/src/grains/virt.py
+++ b/susemanager-utils/susemanager-sls/src/grains/virt.py
@@ -7,7 +7,11 @@ import re
 import subprocess
 import salt.modules.virt
 
-from salt.utils.path import which_bin as _which_bin
+try:
+    from salt.utils.path import which_bin as _which_bin
+except ImportError:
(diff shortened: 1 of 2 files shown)


cbosdo (changes requested):
This change would prevent the grains to expose the libvirt server version and thus disable some virtualization features. Would be better to revert https://github.com/uyuni-project/uyuni/commit/53ce08a67f instead... even if in some cases we may get some timing issues at reboot.

cbosdo (approved):
Forget what I said... 👍🏻

meaksh (approved):

PR #4605, opened 2021-12-21
Install product by default after a channel is subscribed

mbussolotto:
## What does this PR change?

Install product by default after a channel is subscribed

## GUI diff

No difference.

- [ ] **DONE**

## Documentation
PR: https://github.com/uyuni-project/uyuni-docs/pull/1339

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/14952

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
```diff
@@ -1400,7 +1400,7 @@ private void setFileOwner(Path path) throws IOException {
             pillar.put("_mgr_channels_items_name", "mgr_channels_new");
             pillar.put("mgr_channels_new", chanPillar);
 
-            ret.put(State.apply(Arrays.asList(ApplyStatesEventMessage.CHANNELS),
+            ret.put(State.apply(Arrays.asList(ApplyStatesEventMessage.CHANNELS, ApplyStatesEventMessage.PACKAGES),
```
mcalmer, line 1403: I think this will execute more than we want. I would suggest to move the state we want to execute into the channels/init.sls file.

mcalmer (changes requested):
I miss the change in the way when we set tokens to valid. Even if you think it works, I can only think of a race condition.
We should create the channels from the beginning as "valid" and catch a failed state.apply of the channel state to invalidate the token.

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -939,6 +939,9 @@ private void handleSubscribeChannels(ServerAction serverAction, JsonElement json
                     newTokenIds);
         }
         else {
+            //set the token as invalid
+            SubscribeChannelsAction sca = (SubscribeChannelsAction)action;
+            sca.getDetails().getAccessTokens().forEach(x -> x.setValid(false));
```
mcalmer, line 944: Have you tested that this really change the value in the DB?

mcalmer (changes requested):
One question and a small change to fix a long available bug

java/code/src/com/suse/manager/utils/SaltUtils.java
mbussolotto: no :) I'm doing it now

mbussolotto: ok, the current version saves the valid token variable correctly. Just to confirm:
- I checked the variable in susechannelaccesstoken
- I simulate a failure on subscribe channel action making file `/etc/zypp/repos.d/susemanager:channels.repo` immutable

mackdk (commented):
Code change LGTM, waiting form Michael to confirm that his concerns are all addressed.

mcalmer (changes requested):

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -764,27 +764,23 @@ private void handleSubscribeChannels(ServerAction serverAction, JsonElement json
...
-            serverAction.setResultMsg("Failed to apply state: " + ApplyStatesEventMessage.CHANNELS);
+            //set the token as invalid
+            SubscribeChannelsAction sca = (SubscribeChannelsAction)action;
+            sca.getDetails().getAccessTokens().forEach(token -> {
+                token.setValid(false);
+                AccessTokenFactory.save(token);
+            });
+
+            serverAction.setResultMsg("Failed to apply state: " + ApplyStatesEventMessage.CHANNELS + ".\n" +
+                    jsonResult);
```
mcalmer, line 783: isn't this jsonResult plain json and needs some formatting?

mbussolotto: I tried with `getJsonResultWithPrettyPrint` but the output result was exactly the same. You can see an example here:
Screenshot from 2022-01-18 17-16-21

mcalmer: Uhhh, this is looking bad.

Can you try if this produce something more human readable?

```
statesResult.setOutput(YamlHelper.INSTANCE
                .dump(Json.GSON.fromJson(jsonResult, Object.class)).getBytes());
```

mbussolotto: we have a winner :+1:
Screenshot from 2022-01-18 18-01-13. I'll substiture the content of `getJsonResultWithPrettyPrint` with this solution, and use it in the rest of the code

mcalmer: Looks much better. I think we can improve more, but not in this PR.

mbussolotto: change `getJsonResultWithPrettyPrint` seems to be safe. Just to be sure I would not squash this PR (so we have 2 commit, one for the main changes and one for  `getJsonResultWithPrettyPrint` ). What do you think?

mcalmer: Sounds good

mbussolotto: I try to subscribe channels when a minion is in blackout mode and yes, `getJsonResultWithPrettyPrint` slightly changed the String (the initial character is now `'` instead of `"` ) . I changed the `if` to match the new string

mcalmer (approved):

lucidd (approved):

PR #4626, opened 2021-12-29
Fix channels state path calculation for dnf plugin with bundle

vzhestkov:
## What does this PR change?

Fixes dnf plugin path calculation in `channels` state while using Salt Bundle. As Salt Bundle is running with python version other than used for dnf plugins.

## GUI diff

No difference.

## Documentation

## Links

Fixes https://github.com/uyuni-project/uyuni/pull/4626

## Changelogs

- Fix dnf plugin path calculation when using Salt Bundle

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/channels/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/channels/init.sls
@@ -5,7 +5,7 @@
 {%- set is_dnf = salt['pkg.version']("dnf") %}
 
 {%- if is_dnf %}
-{%- set dnf_plugins = salt['cmd.run']("find /usr/lib -type d -name dnf-plugins -printf "%T@ %p\n" | sort -nr | cut -d " " -s -f 2- | head -n 1", python_shell=True) %}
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/salt/channels/init.sls
```diff
@@ -5,9 +5,11 @@
 {%- set is_dnf = salt['pkg.version']("dnf") %}
 
 {%- if is_dnf %}
+{%- set dnf_plugins = salt['cmd.run']("find /usr/lib -type d -name dnf-plugins -printf "%T@ %p\n" | sort -nr | cut -d " " -s -f 2- | head -n 1", python_shell=True) %}
```
mcalmer, line 8: I have the feeling that the command quoting might be wrong. Please check if maybe some single quotes or escaping might be needed

mcalmer (changes requested):

susemanager-utils/susemanager-sls/salt/channels/init.sls
vzhestkov: Thanks @mcalmer for pointing. I have no idea why, but it's working. For sure it's better to fix.

https://github.com/uyuni-project/uyuni/pull/4626/commits/a7ce800b566a2e04959a5936de35fd4c56c7f9db is fixing it

mcalmer (approved):

PR #4627, opened 2021-12-29
Drop `spacewalk-login.js`

Etheryte:
## What does this PR change?

As far as I could see, `spacewalk-login.js` is not actually used anywhere anymore given the login page is now in React. This also means the `formFocus` function is not used anywhere.  

Tagging Michael for a review just in case there's some arcane knowledge background that I'm missing here.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: Unused code.

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/javascript/spacewalk-essentials.js
+++ b/web/html/javascript/spacewalk-essentials.js
@@ -253,14 +253,6 @@ function onDocumentReadyAutoBootstrapGrid() {
   });
 }
 
-// Put the focus on a given form element
-function formFocus(form, name) {
-  var focusControl = document.forms[form].elements[name];
-  if (focusControl.type != "hidden" && !focusControl.disabled) {
-     focusControl.focus();
-  }
-}
-
 // Humanizes all the time elements with the human class
 function humanizeDates() {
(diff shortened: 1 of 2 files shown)


ncounter (approved):
:+1: Good catch, I think this was just a leftover of [this commit](3c6edb785e2e781ce1749f3fa674f01d29d2bb6a) when the `susemanager-login.js` script was dropped. Thanks for the cleanup.

One remark: would it bother you if I may ask you to change the commit message from _deprecate_ to _drop_ . When I read _deprecate_ I expect something to still be there and be qualified as something legacy to avoid using, here instead we're deleting completely the `.js` file. Thank you. It LGTM in anycase :wink:

mcalmer (approved):
I have no clue about javascript. But if Dario is ok, I am ok as well :-)

Etheryte:
I think your comment about the commit message is a good idea Dario, I've made the change. Thanks for looking into it. 👍

PR #4643, opened 2022-01-09
Trust vendor gpg keys option 2

mcalmer:
## What does this PR change?

Use GPG Key Url of the channel configuration and configure the repositories on the clients.
For RPM based repos like:

```
[alias]
name = name
gpgkey=file:///etc/pki/rpm-gpg/my.key
```

for DEB use the sources.list config like:
```
deb [trusted-by=/usr/share/keyring/my.key] ....
```

Uyuni only deploy the keys to the local directories and when a channel is really used on a client the channel config say which GPG key was used to sign the packages.
This requires that the GPG keys are part of the salt filesystem (susemanager-sls) and exists in the right format (debian uses not armored keys)

Additional feature:
Customers can define own GPG keys they want to deploy to the client (only deploy). They can create a pillar list with filenames and provide the key in the gpg directory under this name:

```
$> cat /etc/pillar/mypillar.sls
custom_gpgkeys:
  - my_first_gpg.key
  - my_second_gpgkey.gpg

$> ls -la /etc/salt/gpg/
/etc/salt/gpg/my_first_gpg.key
/etc/salt/gpg/my_second_gpgkey.gpg
```

NOTE: we need to adapt spacewalk-common-channels.ini and sumatoolbox to set `gpgkeyurl` for created channels correctly to the local key.

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


lucidd (approved):

testsuite/features/core/srv_disable_local_repos_off.feature
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Turn "disable_local_repos" feature on highstate off
```
srbarrios, line 4: I'm not convinced about this tweak...
Customers will not have that tweak...


Are we sure this is the best way to setup?

mcalmer: The testsuite require it. I tried without and it failed completly.
Remember: our testsuite is not using repos comming from the server, but enabling local repos to provide the needed packages during the tests.
If we disable all local repos again and again, the testsuite cannot run.

srbarrios: Ok, let's keep this approach for now.
But we need to make sure that this feature works also fine in a production environment without this hack in the test environment, removing "manually" one of the salt states of a regular onboarding.

Bischoff: Fully agree with @srbarrios . We need to test as customers.

Furthermore, not doing real reposyncs is just the root of so many problems.

PR #4671, opened 2022-01-14
Additional checks for channel conflics

mackdk:
## What does this PR change?

This PR adds a check before adding a product to verify that the vendor channels to add do not conflict with any existing custom channel. If a conflict is detected an error is reported and syncing is preventing for any conflicting product and its extensions.

The error handling is also improved in the frontend by showing the reason for each failure and making sure it's not immediately cleared after the data refresh.

## GUI diff

Before:

https://user-images.githubusercontent.com/2292684/149491882-737da34a-1187-4ad4-9b50-a5e20905b24c.mp4

After:

https://user-images.githubusercontent.com/2292684/149491914-7d7ee968-a5f8-4600-9f09-843216f5a541.mp4

- [X] **DONE**

## Documentation
- No documentation needed: Improvement on an existing documented functionality

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Fixes SUSE/spacewalk/issues/16520

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/frontend/dto/SetupWizardProductDto.java
```diff
@@ -215,6 +215,12 @@ public boolean isFinished() {
         }
     };
 
+    /** The product id */
+    private Long id;
+
+    /** The database id */
+    private Long databaseId;
```
mcalmer, line 222: I would rename these as this could become confusing with how we named these at other placed.

The id column in the database which normally has also a sequence is always called just `id`.
The ID of the product as it comes from SCC is called `product_id`.

This seems to turn it around and could become a problem.
Please call the databaseID => id and the id => productId .

mcalmer (changes requested):

java/code/src/com/redhat/rhn/frontend/dto/SetupWizardProductDto.java
mackdk: I agree with you. I named it like this because in the class `MgrSyncProductDto` we have already a field named `id`, which is actually the `reportId`. I will refactor both classes to stay consistent.

branding/css/theme.less
```diff
@@ -475,6 +475,7 @@ h1, h2, h3, h4, h5, h6 {
 .alert {
   margin: 18px 0;
   padding-right: 34px;
+  white-space: pre-line;
```
Etheryte, line 478: @ncounter Do you think this will pose any problems in any other views?

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
```diff
@@ -98,70 +104,134 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    public Map<Long, Long> getProductTreeMap(List<Long> productsId) {
+        final SelectMode query = ModeFactory.getMode("Product_queries", "extract_product_tree");
+        final DataResult<Map<String, Long>> dataResult = query.execute(productsId);
+        if (CollectionUtils.isEmpty(dataResult)) {
+            return Collections.emptyMap();
+        }
+
+        final Map<Long, Long> resultMap = new HashMap<>();
```
lucidd, line 169: same as above

```java
dataResult.stream().collect(Collectors.toMap(
                row -> row.get("product_id"), 
                row -> row.get("parent_id")));
```

java/code/src/com/suse/manager/webui/controllers/ProductsController.java
```diff
@@ -237,25 +240,33 @@ public static String addProduct(Request request, Response response, User user) {
...
+
+        ProductSyncManager psm = new ProductSyncManager();
+        Map<String, Optional<? extends Exception>> productStatusMap = psm.addProducts(identifiers, user);
+
+        // Convert to a map specifying operation result for each product while logging the errors that have happened
+        Map<String, HashMap<Object, Object>> resultMap = productStatusMap.entrySet().stream().collect(
+            Collectors.toMap(
+            Map.Entry::getKey, entry -> {
+                // Return the outcome and the error message to give a better notification in the frontend
+                final HashMap<Object, Object> productResult = new HashMap<>();
```
lucidd, line 259: The key here seems to be always a String so we can reflect that here.

lucidd: also thinking about it the presence or absence of error should be enough to determine if it was successful so unless this format is already something we were using before i would get rid of the success flag.

mackdk: Before we were using just the boolean value in the frontend. I could indeed just return the error message and have it as `null` if there was no error. I'm not sure though if returning null for a successful is clean, from a JSON response point of view. It is anyway totally doable, I just have to change a bit more the typescript code.

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
```diff
@@ -98,70 +104,134 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    public Map<Long, List<String>> verifyChannelConflicts(List<Long> productsId) {
+        final SelectMode query = ModeFactory.getMode("Product_queries", "verify_channel_conflicts");
+        final DataResult<Map<String, Object>> dataResult = query.execute(productsId);
+        if (CollectionUtils.isEmpty(dataResult)) {
+            return Collections.emptyMap();
+        }
+
+        final Map<Long, List<String>> resultMap = new HashMap<>();
```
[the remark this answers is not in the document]
mackdk: Done :+1:

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
mackdk: Done :+1:

java/code/src/com/redhat/rhn/frontend/dto/SetupWizardProductDto.java
mackdk: I [refactored all the affected DTOs](https://github.com/uyuni-project/uyuni/pull/4671/commits/eb7253895afe780cb4760c1483d5f12fe71c683b). Now the naming is consistent with the database.

java/code/src/com/suse/manager/webui/controllers/ProductsController.java
lucidd: I see json here just as the wire format and from that perspective i personally don't care how it looks like as long as its efficient and leaves little room for error. In this case i'm just mentioning it because the 2 fields carry the same information and you can create a response where they say different things which makes little sense. I think we should ask the frontend people what they think about since they have to handle it.

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
mackdk: I had to revert this :man_facepalming: `Collectors.toMap()` [does not support null values](https://bugs.openjdk.java.net/browse/JDK-8148463) and since `parent_id` is null for the main product, it will always throw `NullPointerException`. I added a comment to explain that.

java/code/src/com/suse/manager/webui/controllers/ProductsController.java
mackdk: @Etheryte What do you think?

Etheryte: Jumping in with fairly little context here, so take my opinion with a grain of salt. Imo the current boolean value approach is not as good as it could be, as understanding it requires either the comments to be up-to-date or for the developer to jump between the frontend and backend code. From that perspective, either one of having a success flag or an error message field would work well to avoid that problem, I don't have a strong preference for either. I don't see any extra merit to having both, since you would probably end up using only one to check anyway.

branding/css/theme.less
ncounter: @mackdk While I am fine with the rule itself (the definition is `pre-line: Sequences of whitespace will collapse into a single whitespace. Text will wrap when necessary, and on line breaks`) I actually wonder what is the original reason why you added the rule.
What's the case you figured out you need it and how it would look like without it? Maybe we always missed that and it will look fine everywhere, but `alert` messages are dynamically populated and I cannot tell in advance which one will be hit by this. On top of this, IIRC we have alert messages with new lines already, it would be nice to figure out how they are handled and then align one to the other behavior.
I'd be fine adopting the rule, could you please check first how it is done in other places and see if it doesn't break anything. Thank you

mackdk: Thanks to your comments I had a second look, and even if I could not find any explicit multi-line alert, I figured out that we are using JSX for formatting the message (which, as a rusty backend developer I was not aware of). So I reverted the change in the CSS and used JSX as well to show the list of errors. Now the message looks like this:

Screenshot from 2022-01-28 11-20-16

Let me know if the code can be further improved.

java/code/src/com/suse/manager/webui/controllers/ProductsController.java
mackdk: I refactored the code and get rid of the flag returning only the error message, if present. The downside of this is that with GSON we are actually not serializing null values, so in the end the response contains only the failed products. This does not affect much the logic on the frontend side and the code is able to handle both null and missing values.

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
```diff
@@ -98,70 +104,131 @@
...
-     * @throws ProductSyncException if an error occurred
      * @return a map of added products and an error message if any
      */
-    public Map<String, Optional<? extends Throwable>> addProducts(List<String> productIdents, User user) {
+    public Map<String, Optional<? extends Exception>> addProducts(List<String> productIdents, User user) {
+        // Extract the ids from the identifiers
+        final List<Long> productsId = productIdents.stream().map(
+            // Ident are in the form 'product_id-label'. If the product id is negative, the ident starts with a minus,
+            // so we need to skip the first one when performing the indexOf
+            id -> Long.valueOf(id.substring(0, id.indexOf('-', id.startsWith("-") ? 1 : 0)))
```
mcalmer, line 114: Would this have the same result?

```suggestion
            id -> Long.valueOf(id.substring(0, id.indexOf('-', 1)))
```
 The first might be '-' or not. If we always ignore the first character it should also work for product ids between 0 and 9.

mcalmer (approved):
LGTM - one little question added but nothing which prevent merging.

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
mackdk: Yeah, indeed you are right. I will simplify this.

branding/css/theme.less
ncounter: Nice, well done. I think it looks good now :+1:

PR #4676, opened 2022-01-14
Upgrade Webpack

Etheryte:
## What does this PR change?

Upgrade Webpack from 4 to 5. Full details of the differences can be found [over here](https://webpack.js.org/migrate/5/), in our case it's mostly upgrading dependencies, new configuration schemas, and dropping obsolete configuration.  

Along with this upgrade, we're dropping manual chunk separation in favor of defaults which do a good job at automatically optimizing the bundle these days.  

Aside from resolving dependabot issues, expect faster build times (down from 58s to 40s for a cold full build on my machine).  

In addition, the syntax for running a frontend development proxy has slightly changed due to changes in how Webpack 5 handles CLI flags. I will send out a memo email about this when the branch gets merged.  

Old syntax:  
```sh
$ yarn proxy --server "https://server.tf.local"
```

New syntax:
```sh
$ yarn proxy https://server.tf.local
```

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Part of https://github.com/SUSE/spacewalk/issues/16424

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


Etheryte:
This now works for all the cases that I could think of to test, feel free to take it for a spin to see if everything works for your setup. Please remember to update the dependencies before you do via `$ yarn`.

Etheryte:
Would also appreciate a review from Michael regarding the spec files since I don't know too much about them. In short, we don't ship libraries in a separate folder anymore, but along with the rest of the manager scripts, so the extra copying operations are not needed. I _think_ I cleaned up everything, but I'm not sure if this might be defined anywhere else outside of this repository?

Etheryte:
Should [zz-spacewalk-www.conf#L152](https://github.com/uyuni-project/uyuni/blob/master/spacewalk/config/etc/httpd/conf.d/zz-spacewalk-www.conf#L152) be updated as well, since it includes a reference to vendors?

parlt91 (approved):
From my point of view these changes look good. I did some smoke testing and everything seems to work fine. I can however not comment on your questions whether the changes to the spec file could have some other consequences (maybe @mcalmer can comment on that). Also I do not know if there are any further required adaptions to `zz-spacewalk-www.conf` (I think @cbbayburt would have a better idea)

mcalmer:
Hmm, technically all changes looks correct.
But I think we cannot give up the separation of a package with the external libs and our own written code.
This does not mean, that we need the extra path (vendor/) but separate packages might be more a legal problem than a technical decission.

But I am neither a lawyer nor a javascript developer. I do not know what "javascript developers typically do" and if the lawyer is fine with it.

Etheryte:
Discussed with legal, seems all good (for future reference, the email thread is called "A question regarding packaging Javascript in Suma"). I'm going to rebase and merge this when I get back from vacation.

sbluhm:
@Etheryte , I see you removed `susemanager-web-libs` which now results in a conflict:
`- nothing provides susemanager-web-libs needed by spacewalk-html-4.3.8-8.1.uyuni1.noarch`
Was the intention to remove  `susemanager-web-libs`? If yes, can I just drop the dependencies in the spec?

Etheryte:
@sbluhm Sorry I missed this, what package is this requirement included in? As for your question, yes, `spacewalk-html` now includes the contents that were previously included in `susemanager-web-libs`.

sbluhm:
Requirement is included in spacewalk-html :)

```
$ git grep susemanager-web-libs
web/html/src/build/fill-spec-file.js:        /(?<=%package -n susemanager-web-libs[\s\S]*?)License:.*/m,
web/html/src/build/fill-spec-file.js:          `susemanager-web-libs.spec was generated successfully with the following licenses: ${mappedProcessedLicenses}`
web/spacewalk-web.spec:%package -n susemanager-web-libs
web/spacewalk-web.spec:%description -n susemanager-web-libs
web/spacewalk-web.spec:%package -n susemanager-web-libs-debug
web/spacewalk-web.spec:Requires:       susemanager-web-libs
web/spacewalk-web.spec:%description -n susemanager-web-libs-debug
web/spacewalk-web.spec:Requires:       susemanager-web-libs

```

Etheryte:
Hah, sorry for the mess, looks like I didn't look in the only place that's obvious. I'm going to see what else I might've missed and then tag you for a review when I make a pull request, thanks for the heads up.

PR #4708, opened 2022-01-20
Removed isPostgresql() check since we support only PostegreSQL

mackdk:
## What does this PR change?

This commit remove the usage of `isPostgresql()` method.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


java/code/src/com/suse/manager/webui/utils/MinionActionUtils.java
```diff
@@ -73,7 +63,7 @@
     private static final Logger LOG = Logger.getLogger(MinionActionUtils.class);
 
     /** Whether the current database is Postgres. */
-    public static final boolean POSTGRES = ConfigDefaults.get().isPostgresql();
+    public static final boolean POSTGRES = true;
```
mcalmer, line 66: Is this still somewhere used or can we remove that constant as well?

mcalmer (commented):
Looks good so far, but I think others should have a look as well.

java/code/src/com/suse/manager/webui/utils/MinionActionUtils.java
mackdk: Oh yes, it should be removed as well.

sbluhm:
Doesn't this close the door for future different DB support like MariaDB/MySQL?

mcalmer:
> Doesn't this close the door for future different DB support like MariaDB/MySQL?

Yes, but Uyuni has so many very special needs, that I do not see any chance to support other databases at all.
Even bringing Oracle support back would be challeging as we already now use features which are not supported by Oracle.
MariaDB/MySQL are are not an option at all as a lot of basics just do not work.

mackdk:
> Doesn't this close the door for future different DB support like MariaDB/MySQL?

> I do not see any chance to support other databases at all.

Moreover, from what I see, we have already far more places where the code and the queries are postgres-specific, than the ones currently covered by the `isPostgres()` method or the query suffix mechanism. So, even if we wanted to add support for another database, the code this commit removes is misleading at best, if not already outdated.

sbluhm:
ok thanks. will not open that door in future then....

lucidd (approved):

mcalmer (approved):

PR #4715, opened 2022-01-20
Optimize adding new product functions (bsc#1193707)

mbussolotto:
## What does this PR change?

The PR tries to improve the scenario where is tried to add lots of products simultaneously. Instead of running the function `getBaseProducts` that is really time consuming, we run it just once

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/16542

- [x] **DONE**

## Changelogs
- [ ] No changelog needed

## Re-run a test


mcalmer:
@mbussolotto what's up with this? It was an L3. That is closed, but the bug is still open. There is a draft PR, but not ready for review? Please have a look when you are back

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
```diff
@@ -495,10 +492,11 @@ private SetupWizardProductDto convertProduct(final MgrSyncProductDto productIn)
...
+     * @param productsWithAddons the products list with addons
      * @return the {@link SetupWizardProductDto}
      */
-    private SetupWizardProductDto findProductByIdent(String ident) {
-        for (SetupWizardProductDto p : getBaseProducts()) {
+    private SetupWizardProductDto findProductByIdent(String ident, List<SetupWizardProductDto> productsWithAddons) {
+        for (SetupWizardProductDto p : productsWithAddons) {
             if (p.getIdent().equals(ident)) {
                 return p;
             }
```
mackdk, line 502: GitHub is not letting me to put the comment in the correct line :man_facepalming: Anyway, below here on line 503, I think you can remove the loop over `p.getAddonProducts()` since now the addons are part of the main list.

mbussolotto: That's a good news, thanks! I'm wondering if maybe I can remove at all `addonProducts` https://github.com/uyuni-project/uyuni/blob/3b7d8d34dff8809a97874f35421df952d1be8150/java/code/src/com/redhat/rhn/frontend/dto/SetupWizardProductDto.java#L249  ..what do you think?

mbussolotto: same here ? https://github.com/uyuni-project/uyuni/blob/3b7d8d34dff8809a97874f35421df952d1be8150/java/code/src/com/redhat/rhn/domain/product/SUSEProductSet.java#L39

mackdk: From the DTO maybe we can since, it seems to be used only in the method you changed so I think it's safe to "flatten" the representation. In the `SUSEProductSet` business object I think it's needed.

mackdk: After 1:1 conversation, we decided it's not worth to remove the `addonProducts` field from the DTO.

java/code/src/com/redhat/rhn/manager/setup/ProductSyncManager.java
```diff
@@ -495,10 +492,11 @@ private SetupWizardProductDto convertProduct(final MgrSyncProductDto productIn)
      * Find a product for any given ident by looking through base and their addons.
      *
      * @param ident ident of a product
+     * @param productsWithAddons the products list with addons
      * @return the {@link SetupWizardProductDto}
      */
-    private SetupWizardProductDto findProductByIdent(String ident) {
-        for (SetupWizardProductDto p : getBaseProducts()) {
+    private SetupWizardProductDto findProductByIdent(String ident, List<SetupWizardProductDto> productsWithAddons) {
+        for (SetupWizardProductDto p : productsWithAddons) {
```
mcalmer, line 499: we still iterate over the whole list to find the ident. But I think we do not safe so much time if we convert it into a map using ident as key and the object as value to get a direct access.
So ok for me.

mcalmer (approved):

PR #4720, opened 2022-01-21
spacecmd: fix interactive mode for "system_applyerrata" and "errata_apply" commands (bsc#1194363)

meaksh:
## What does this PR change?

This PR fixes an issue in `spacecmd`, which is not properly detecting interactive/non-interactive mode when running `system_applyerrata` and `errata_apply` commands:

```console
suma-42-srv.tf.local:~ # spacecmd 
Welcome to spacecmd, a command-line interface to Spacewalk.

Type: 'help' for a list of commands
      'help <cmd>' for command-specific help
      'quit' to quit

INFO: Connected to https://suma-42-srv.tf.local/rpc/api as admin
spacecmd {SSM:0}> system_applyerrata suma-42-min-sles15sp3.tf.local SUSE-15-SP3-2021-2858
Errata             Systems
--------------     -------
SUSE-15-SP3-2021-2858          1

Start Time: 20220120T17:46:52
spacecmd {SSM:0}> 
```

As you see, the command exited, didn't ask for user confirmation and directly existed without scheduling any action as it was expected.

With this PR is applied, "spacecmd" is asking the expected questions when running in interactive mode:

```console
suma-42-srv.tf.local:~ # spacecmd system_applyerrata suma-42-min-sles15sp3.tf.local SUSE-15-SP3-2021-2858
INFO: Connected to https://suma-42-srv.tf.local/rpc/api as admin
Start Time [now]: now
Errata             Systems
--------------     -------
SUSE-15-SP3-2021-2858          1

Start Time: 20220121T11:55:23

Apply these patches [y/N]: y
INFO: Scheduled 1 system(s) for SUSE-15-SP3-2021-2858
```

If running in non-interactive mode, using `--yes` will directly schedule the action as it is expected.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/16628

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/src/spacecmd/errata.py
+++ b/spacecmd/src/spacecmd/errata.py
@@ -168,7 +168,10 @@ def do_errata_apply(self, args, errata=None, only_systems=None):
     print('')
     print(_('Start Time: %s') % start_time)
 
-    if (is_interactive(options) and not self.user_confirm(_('Apply these patches [y/N]:'))) or not self.options.yes:
+    if not is_interactive(options) and not self.options.yes:
+        return 1
+
+    if is_interactive(options) and not self.user_confirm(_('Apply these patches [y/N]:')):
         return 1
 
     # if the API supports it, try to schedule multiple systems for one erratum
--- a/spacecmd/src/spacecmd/errata.py
+++ b/spacecmd/src/spacecmd/errata.py
@@ -92,14 +92,17 @@ def complete_errata_apply(self, text, line, beg, end):
     return self.tab_complete_errata(text)
 
 
-def do_errata_apply(self, args, only_systems=None):
+def do_errata_apply(self, args, errata=None, only_systems=None):
     arg_parser = get_argument_parser()
+    arg_parser.add_argument('errata', nargs="*")
     arg_parser.add_argument('-s', '--start-time')
 
     args, options = parse_command_arguments(args, arg_parser)
     only_systems = only_systems or []
+    errata = options.errata or errata
+    del options.errata
 
-    if not args:
+    if not args and not errata:
         self.help_errata_apply()
         return 1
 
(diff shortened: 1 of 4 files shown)


vzhestkov (approved):
Looks good

spacecmd/src/spacecmd/errata.py
```diff
@@ -92,31 +92,34 @@ def complete_errata_apply(self, text, line, beg, end):
...
-def do_errata_apply(self, args, only_systems=None):
+def do_errata_apply(self, args, errata=None, only_systems=None):
     arg_parser = get_argument_parser()
+    arg_parser.add_argument('errata', nargs="*")
     arg_parser.add_argument('-s', '--start-time')
 
     args, options = parse_command_arguments(args, arg_parser)
     only_systems = only_systems or []
+    errata = options.errata or errata
+    del options.errata
```
agraul, line 103: I don't know much about `spacecmd`'s argument parsing, just two quick questions:

1. Can `$errata` exist in `args` *and* be passed as `errata`? If yes, why do we prefer what's in `args` over whats in `errata`?
2. I see `del` being used a bit in `spacewalk/`, but not everywhere. Can you elaborate on when we (should) use it to clear the `options` namespace?

spacecmd/src/spacecmd/errata.py
```diff
@@ -92,31 +92,34 @@ def complete_errata_apply(self, text, line, beg, end):
...
         if not options.start_time:
-            options.start_time = parse_time_input('now')
+            start_time = parse_time_input('now')
         else:
-            options.start_time = parse_time_input(options.start_time)
+            start_time = parse_time_input(options.start_time)
 
     # allow globbing and searching via arguments
-    errata_list = self.expand_errata(args)
+    errata_list = self.expand_errata(errata)
```
agraul, line 122: Isn't this redundant if `do_errata_apply` is called from `do_system_applyerrata`?

spacecmd/src/spacecmd/errata.py
meaksh: Thanks for the review Alex :+1:

Let me try to answer your questions:

  1. There shouldn't be a case when both are passed. The `errata` and `only_installed` parameters for this function are only passed when the function is called by `system_erratapply`, and in this case, we want the value coming from `errata`. When this function is directly called by the user via CLI, then errata is contained as part of `args` only. So, I can actually even use `errata = errata or options.errata` to be aligned with what we do for `only_installed`.
  
  2. Well, the thing is that the current way for `spacecmd` to check whether it is running in interactive mode or not is really bad, it basically assumes running non-interactive if there is any non-empty values passed inside `options`. This makes a bit tricky to deal with arguments passed from other spacemd functions. So, in this case, we don't want to directly assume that we run in non-interactive mode when we are passing `errata` in `args`.

I'll align what I just mentioned on 1.

spacecmd/src/spacecmd/errata.py
meaksh: Yes, it's redundant. Good catch :+1:

meaksh:
@agraul @vzhestkov I have reorganiced the code and also fixed another corner case, I think now code looks better.

PR #4739, opened 2022-01-28
Add large data set support to CLM channel selection

Etheryte:
## What does this PR change?

Update the CLM channels selection view to not lock up on large data sets.  

Main points:  
 - reduce unnecessary rerenders with memoing and update checks  
 - replace simple list rendering with a virtualized list implementation (https://github.com/petyosi/react-virtuoso)
 - reduce computational complexity of some of the data transforms

## GUI diff

Perhaps some small layout shifts here and there, but should look more or less the same.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Part of https://bugzilla.suse.com/show_bug.cgi?id=1192598

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/src/manager/content-management/shared/components/panels/sources/channels/channels-api.ts
```diff
@@ -0,0 +1,198 @@
...
+                recommended: false,
+                standardizedName: "",
+                parent: base,
+              },
+            ],
+          });
+        }
+      }
+
+      // TODO: This matches the current behavior but this should already be done on the server side, with a separate flag if necessary
```
Etheryte, line 102: @cbbayburt Is this something we can move to the server, or is there a reason why not to do it that I'm missing?

cbbayburt (commented):
Thanks, @Etheryte, looks very clean

One thing from me: let's try and address all TODOs as much as we can now, while it's fresh. Otherwise, they'll age here forever.
My suggestion is to;
1. Do all TODOs that can be done now in a reasonable time
2. Reach out if they need to be unblocked by someone else
3. If cannot be done now, but it's important; make a plan (describing it in a separate issue)
4. If none of the above is feasible, just delete the TODO

web/html/src/manager/content-management/project/project.tsx
```diff
@@ -41,8 +42,11 @@ const Project = (props: Props) => {
   const hasEditingPermissions = isOrgAdmin(roles);
 
   useInterval(() => {
-    onAction({}, "get", project.properties.label).then((project) => {
-      setProject(project);
+    onAction({}, "get", project.properties.label).then((newProject) => {
+      // Only propagate the received value if something has changed
+      if (!_isEqual(project, newProject)) {
+        setProject(newProject);
+      }
```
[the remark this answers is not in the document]
Etheryte: This is a part of the issue at hand actually, the view was pretty heavy to render and on top of that it got rerendered whenever the recurring network request replied even if the contents were the same as before.

web/html/src/core/channels/type/channels.type.ts
```diff
@@ -9,3 +11,62 @@ export type ChannelType = {
...
+  id: number;
+  archLabel: string;
+  custom: boolean;
+  isCloned: boolean;
+  label: string;
+  name: string;
+  standardizedName: string;
+  recommended: boolean;
+  subscribable: boolean;
+};
```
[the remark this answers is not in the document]
cbbayburt: Oops sorry, I was looking at the 1st commit alone

web/html/src/manager/content-management/shared/components/panels/sources/channels/channels-api.ts
Etheryte: I've made the change in[`ChannelsApiController.java`](https://github.com/uyuni-project/uyuni/pull/4739/files#diff-4d83abbd4ca969f44973a005de0adc5886dd4772d704886ab010eff4008c8280), if you have a moment to check that my copy-paste-from-google Java coding makes sense, I'd appreciate it. 😄

java/code/src/com/suse/manager/webui/controllers/channels/ChannelsApiController.java
```diff
@@ -77,6 +79,12 @@ public static String getAllChannels(Request req, Response res, User user) {
                             .filter(c -> !filterOutIds.contains(c.getBase().getId()))
                             .collect(Collectors.toList()
             );
+            Collections.sort(jsonChannelsFiltered, new Comparator<ChannelsJson>() {
+                @Override
+                public int compare(ChannelsJson c1, ChannelsJson c2) {
+                    return c1.getBase().getId().compareTo(c2.getBase().getId());
+                }
+            });
```
cbbayburt, line 87: It works just fine, but here's a simpler one-liner:
```suggestion
            jsonChannelsFiltered.sort(Comparator.comparing(cIn -> cIn.getBase().getId()));
```

Etheryte: Nice, thanks a lot for this. I don't know enough about Java so I wouldn't even have thought of something like this. 🍻

PR #4776, opened 2022-02-06
Restart squid so proxy.example.org is recognized

Bischoff:
## What does this PR change?

Cobbler tests: squid needs being restarted so that the CNAME in the SSL certificate for proxy.example.org is recognized.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1043,6 +1043,19 @@ When(/^I configure the proxy$/) do
   $proxy.run(cmd, timeout: proxy_timeout)
 end
 
+When(/^I allow all SSL protocols on the proxy's apache$/) do
+  file = '/etc/apache2/ssl-global.conf'
+  key = 'SSLProtocol'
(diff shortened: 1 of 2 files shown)


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -1043,6 +1043,19 @@
   $proxy.run(cmd, timeout: proxy_timeout)
 end
 
+When(/^I allow all SSL protocols on the proxy's apache$/) do
+  file = '/etc/apache2/ssl-global.conf'
+  key = 'SSLProtocol'
+  val = 'all -SSLv2 -SSLv3'
+  $proxy.run("grep '#{key}' #{file} && sed -i -e 's/#{key}.*$/#{key} #{val}/' #{file}")
+  $proxy.run("systemctl reload apache2.service")
```
srbarrios, line 1051: I think this reload was causing also some issue and we could need a change to a restart.

srbarrios:
LGTM.

PR #4782, opened 2022-02-06
cobbler should work without such horrible hacks

Bischoff:
## What does this PR change?

Do not do /etc/hosts hacks, especially when they can break other things

## Links

Ports:
* 4.1: SUSE/spacewalk#16882
* 4.2: SUSE/spacewalk#16883

## Changelogs

- [x] No changelog needed


srbarrios:
Cool

PR #4797, opened 2022-02-08
Longer reposync inactivity delay

Bischoff:
## What does this PR change?

When a product is selected, more time is needed than before to see the first reposync start.

This PR increases the inactivity timeout when we wait for all non-killed reposyncs to finish.

(seen only on `filemon` private test server, probably because it's slower)

## Links

Ports:
* 4.1: SUSE/spacewalk#16908
* 4.2: SUSE/spacewalk#16909

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -291,11 +291,11 @@ end
 # This function waits for all the reposyncs to complete.
 #
 # This function is written as a state machine. It bails out if no process is seen during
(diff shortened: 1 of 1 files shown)


ktsamis (approved):
If it's only on filemon all good, it should have no impact in testsuites

Bischoff:
> If it's only on filemon all good, it should have no impact in testsuites

@ktsamis well, the fix will apply to all test suites

it's the problem that could be seen only on filemon

PR #4806, opened 2022-02-09
Implement Salt Bundle support for salt-ssh

vzhestkov:
## What does this PR change?

Implements Salt Bundle support for salt-ssh

https://github.com/uyuni-project/uyuni-rfc/pull/65

## GUI diff

No difference.

## Documentation
- Documentation issue was created (TBD): [Link for SUSE Manager contributors](https://github.com/SUSE/spacewalk/issues/new?template=ISSUE_TEMPLATE_DOCUMENTATION.md&labels=documentation&projects=SUSE/spacewalk/31), [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/new).

## Test coverage
- Cucumber tests were added (TBD)

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/conf/rhn_web.conf
+++ b/web/conf/rhn_web.conf
@@ -110,6 +110,12 @@ ssh_push_sudo_user =
 # This is potentially dangerous and should not be used
 ssh_push_task_timeout = 0
 
+# preflight script to run on ssh push minions to deploy Salt Bundle
+ssh_salt_pre_flight_script = /usr/share/susemanager/salt-ssh/preflight.sh
+
+# use salt-thin as a fallback method if Salt Bundle is not available
+ssh_use_salt_thin = false
+
 # Enable the technical preview Single Page Application engine.
 # Enable this in order to test a faster UI navigation
 web.spa.enable = true
--- a/spacewalk/admin/salt-secrets-config.py
+++ b/spacewalk/admin/salt-secrets-config.py
@@ -61,6 +61,8 @@ with cfg_component(None) as CFG:
 with cfg_component("web") as CFG:
     uyuni_roster_config.update({
         "ssh_push_port_https": CFG.SSH_PUSH_PORT_HTTPS,
+        "ssh_pre_flight_script": CFG.SSH_SALT_PRE_FLIGHT_SCRIPT,
+        "ssh_use_salt_thin": CFG.SSH_USE_SALT_THIN == "true",
     })
     if CFG.SSH_PUSH_SUDO_USER:
         uyuni_roster_config.update({
@@ -91,7 +93,10 @@ with open("/etc/salt/master.d/susemanager_db.conf", "w") as f:
     os.fchmod(f.fileno(), 0o640)
 
 with open("/etc/salt/master.d/uyuni_roster.conf", "w") as f:
-    f.write(yaml.safe_dump({"uyuni_roster": uyuni_roster_config}, default_flow_style=False, allow_unicode=True))
+    uyuni_roster_cfg = {"uyuni_roster": uyuni_roster_config}
+    if "ssh_pre_flight_script" in uyuni_roster_config:
+        uyuni_roster_cfg.update({"ssh_run_pre_flight": True})
+    f.write(yaml.safe_dump(uyuni_roster_cfg, default_flow_style=False, allow_unicode=True))
     os.fchown(f.fileno(), pwd.getpwnam("salt").pw_uid, grp.getgrnam("salt").gr_gid)
     os.fchmod(f.fileno(), 0o640)
 
--- a/susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
+++ b/susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
@@ -1,0 +1,239 @@
+#!/bin/sh
+
+if [ $# -lt 2 ]; then
+    echo "Error: Wrong number of arguments!"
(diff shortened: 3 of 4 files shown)


java/code/src/com/suse/manager/webui/utils/SaltRoster.java
```diff
@@ -68,9 +73,37 @@ public void addHost(String host, String user, Optional<String> passwd,
                 Arrays.asList(option)));
         timeout.ifPresent(value -> hostData.put("timeout", value));
         minionOpts.ifPresent(options -> hostData.put("minion_opts", options));
+        sshPreflightScriptPath.ifPresent(value -> hostData.put("ssh_pre_flight", value));
+        if (sshPreflightScriptPath.isPresent()) {
+            sshPreflightScriptArgs.ifPresent(value -> hostData.put("ssh_pre_flight_args",
+                Arrays.asList(value)));
```
mackdk, line 79: > hostData.put("ssh_pre_flight_args", Arrays.asList(value))

This is putting in the map a list with only one element, which is the args list (so `ssh_pre_flight_args` is a list of lists in the end). Is this what you wanted? Or should this just have been `hostData.put("ssh_pre_flight_args", value)`?

vzhestkov: Fixed. Thanks for pointing :+1:

meaksh (commented):
Thanks for the PR @vzhestkov !

I'm still reviewing this but one thing I think we should consider here is to reuse all those functions from the bootstrap repository and not duplicate them. I really want to prevent we need to maintain this same code in all these different places.

Since the code for the bootstrap script is not available from being imported here, this would require extra effors in order to avoid the need of duplicating the logic (and maintain it in multiple places).

Since we want to have this PR merged as soon as possible, I would be fine if we simply add a new task to the bundle EPIC and do this refactor in a follow-up PR.

renner (approved):
I am still investigating if we are covering all the relevant cases with this patch, but it looks quite good already!

java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
```diff
@@ -177,6 +177,23 @@ public static String getSSHUser() {
...
+    public static String getSaltSSHPreflightScriptPath() {
+        String preflightScriptPath = Config.get().getString("ssh_salt_pre_flight_script");
+        return StringUtils.isBlank(preflightScriptPath) ? null : preflightScriptPath;
+    }
+
+    /**
+     * Returns true if salt-thin should be used to handle salt-ssh session
+     * @return the result of the call
+     */
+    public static boolean getSSHUseSaltThin() {
```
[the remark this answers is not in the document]
vzhestkov: It was discussed in the RFC. We decided to support Salt-SSH with Salt bundle only, but for development purposes it's better to left the possibility to use salt-thin as it was originally. It's mostly for us and community developers.

vzhestkov:
@meaksh I agree, it's not good to maintain two copies, but the problem that salt-ssh pre flight script must be a single file, but the bootstrap script is generated with the python script. The only option I see is to regenerate the pre flight script with the script, maybe not with the same script as used for bootstrap. The other problem is that both this opies logically shipped with different packages. Need to find out the compromise for both.

meaksh:
@vzhestkov I'm just thinking loud here, but one possible idea could be to create a new build dependency package, something like `uyuni-bootstrap-build-libs`, where we place this common python code for preflight script and bootstrap script. This would be a new build dependency for `susemanager-sls` and `spacewalk-certs-tools` (includes bootstrap script generator).

Then when building the packages for packages, we can include the functions from `uyuni-bootstrap-build-libs` and render the proper shape of the bootstrap script generator and pre flight script, based on the same code.

web/conf/rhn_web.conf
```diff
@@ -110,6 +110,12 @@ ssh_push_sudo_user =
 # This is potentially dangerous and should not be used
 ssh_push_task_timeout = 0
 
+# preflight script to run on ssh push minions to deploy Salt Bundle
+ssh_salt_pre_flight_script = /usr/share/susemanager/salt-ssh/preflight.sh
+
+# use salt-thin as a fallback method if Salt Bundle is not available
+ssh_use_salt_thin = false
```
meaksh, line 117: As we just discussed, we would need to set this as `true` for now right? As we won't have the necessary code in the Salt version shipped in SLE15SP4 public beta.

```suggestion
ssh_use_salt_thin = true
```

meaksh: Well, in fact this is not yet clear. We still need to clarify what we want to do

testsuite/features/support/commonlib.rb
```diff
@@ -66,6 +66,21 @@ def product
   raise 'Could not determine product'
 end
 
+def product_version
```
srbarrios, line 69: In theory, it should not collide, but we might need to rename this parameter here:

``` 
When(/^I (deselect|select) "([^\"]*)" as a (SUSE Manager|Uyuni) product$/) do |select, product, product_version|
  if $product == product_version
    step %(I #{select} "#{product}" as a product)
  end
end
```

srbarrios: This is a very nice to have function, thank you

srbarrios (approved):
Test code looks good to me.

PR #4815, opened 2022-02-10
QA: Add test for migrating a deleted client into a Salt minion

nodeg:
## What does this PR change?

This PR adds a new migration test. It migrates a beforehand deleted traditional client into a Salt minion. The exact steps are

- register trad. client as Minion
- unregister Minion
- register as trad. client again
- unregister trad. client
- register trad. client as Minion again

This test case does not test a valid scenario. In SUMA you normally do not delete a traditional client and then register it again as Salt minion. You would always just bootstrap the traditional client as a Salt minion again and Salt would recognize that it is a traditional client and will take the the necessary steps to migrate it.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/6781

[Manager 4.2](https://github.com/SUSE/spacewalk/pull/17030)
[Manager 4.1](https://github.com/SUSE/spacewalk/pull/17029)

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_migrate_to_trad.feature
+++ b/testsuite/features/secondary/min_migrate_to_trad.feature
@@ -1,0 +1,111 @@
+# Copyright (c) 2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+# This test case does not test a valid scenario. In SUMA you normally do not
+# delete a traditional client and then register it again as Salt minion. You
+# would always just bootstrap the traditional client as a Salt minion again
+# and Salt would recognize that it is a traditional client and will take the
+# the necessary steps to migrate it
+
+Feature: Migrate a Salt minion into a traditional client
+  As an authorized user
+  I want to migrate these Salt minions to traditional clients and have everything as before
+
+  Scenario: Log in as admin user
+    Given I am authorized for the "Admin" section
+
+  Scenario: Migrate a SLES client into a Salt minion
+    When I follow the left menu "Systems > Bootstrapping"
+    And I enter the hostname of "sle_client" as "hostname"
+    And I enter "22" as "port"
+    And I enter "root" as "user"
+    And I enter "linux" as "password"
+    And I select "1-SUSE-KEY-x86_64" from "activationKeys"
+    And I select the hostname of "proxy" from "proxies"
+    And I click on "Bootstrap"
+    And I wait until I see "Successfully bootstrapped host!" text
+
+  Scenario: Wait until the Salt minion appears
+    When I wait until onboarding is completed for "sle_client" salt minion
+
+  Scenario: Check that the migrated system is now a minion
+    Given I am on the Systems overview page of this "sle_client"
+    When I follow "Properties" in the content area
+    Then I wait until I see "Base System Type:.*Salt" regex, refreshing the page
+
+  @proxy
+  Scenario: Check connection from migrated minion to proxy
+    Given I am on the Systems overview page of this "sle_client"
+    When I follow "Details" in the content area
+    And I follow "Connection" in the content area
+    Then I should see "proxy" short hostname
+
+  @proxy
+  Scenario: Check registration on proxy of migrated minion
+    Given I am on the Systems overview page of this "proxy"
+    When I follow "Details" in the content area
+    And I follow "Proxy" in the content area
+    Then I should see "sle_client" hostname
+
+  Scenario: Unregister migrated client
+    Given I am on the Systems overview page of this "sle_client"
+    When I follow "Delete System"
+    Then I should see a "Confirm System Profile Deletion" text
+    When I click on "Delete Profile"
+    And I wait until I see "has been deleted" text
+    Then "sle_client" should not be registered
+
+  @susemanager
+  Scenario: Register minion again as traditional client
+    When I enable client tools repositories on "sle_client"
+    And I install the traditional stack utils on "sle_client"
+    And I remove package "salt-minion" from this "sle_client"
+    And I bootstrap traditional client "sle_client" using bootstrap script with activation key "1-SUSE-KEY-x86_64" from the proxy
+    Then I should see "sle_client" via spacecmd
+
+  @uyuni
+  Scenario: Register minion again as traditional client
+    When I enable client tools repositories on "sle_client"
+    And I install the traditional stack utils on "sle_client"
+    And I remove package "venv-salt-minion" from this "sle_client"
+    And I bootstrap traditional client "sle_client" using bootstrap script with activation key "1-SUSE-KEY-x86_64" from the proxy
+    Then I should see "sle_client" via spacecmd
+
+  Scenario: Wait until the traditional client appears
+    When I wait until onboarding is completed for "sle_client"
+
+  Scenario: Check that the migrated minion is again a traditional client
+    Given I am on the Systems overview page of this "sle_client"
+    When I follow "Properties" in the content area
+    Then I wait until I see "Base System Type:.*Management" regex, refreshing the page
+
+  Scenario: Unregister traditional client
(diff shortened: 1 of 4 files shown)


nodeg:
Two scenarios are failing due to the fact how Salt is handled. I do not exactly know what should be done between the transition from a traditional client into a Salt minion. I am waiting for feedback from the ion squad. @meaksh @vzhestkov @agraul 
Below are some hints on what I did and what fails.

image (4)
test_failure_trad_client_bootstrap_minion

testsuite/features/secondary/min_migrate_to_trad.feature
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) 2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+# This test case does not test a valid scenario. In SUMA you normally do not
+# delete a traditional client and then register it again as Salt minion. You
+# would always just bootstrap the traditional client as a Salt minion again
+# and Salt would recognize that it is a traditional client and will take the
+# the necessary steps to migrate it
+
+Feature: Migrate a Salt minion into a traditional client
```
nodeg, line 10: I am not sure if I should add a scope here. If so I think we should add both, `@sle_client` and `@sle_minion`.

testsuite/features/secondary/min_migrate_to_trad.feature
```diff
@@ -0,0 +1,111 @@
...
+  As an authorized user
+  I want to migrate these Salt minions to traditional clients and have everything as before
```
srbarrios, line 12: I think that the core of this feature is more about bootstrapping a salt minion in a unregistered traditional client.
Maybe we can rephrase this line and the title.

testsuite/features/secondary/min_migrate_to_trad.feature
srbarrios: We for sure need @sle_client as we make use of sle_client VM.
But this is the tag that assure that this test is not triggered when the VM is not deployed.

Another thing, are the scopes.
Scopes allow us to filter a group of features, to run only features that are on the same scope.
In that case, looking the list, https://github.com/uyuni-project/uyuni/blob/master/testsuite/config/cucumber.yml , we can either create a new scope for @scope_client_migration or if you don't want to make much complex this PR, you can use @scope_traditional_client and that's fine for now. Maybe in another PR, if you want to play with scope, you add the new one. As it's not only here that we need to add, but also into the pipeline definitions in susemanager-ci repo

testsuite/features/secondary/min_migrate_to_trad.feature
```diff
@@ -0,0 +1,111 @@
...
+    And I enter the hostname of "sle_client" as "hostname"
+    And I enter "22" as "port"
+    And I enter "root" as "user"
+    And I enter "linux" as "password"
+    And I select "1-SUSE-KEY-x86_64" from "activationKeys"
+    And I select the hostname of "proxy" from "proxies"
+    And I click on "Bootstrap"
+    And I wait until I see "Successfully bootstrapped host!" text
+
+  Scenario: Wait until the Salt minion appears
```
srbarrios, line 28: Let's combine these two?

testsuite/features/secondary/min_migrate_to_trad.feature
```diff
@@ -0,0 +1,111 @@
...
+
+  Scenario: Unregister traditional client
+    Given I am on the Systems overview page of this "sle_client"
+    When I follow "Delete System"
+    Then I should see a "Confirm System Profile Deletion" text
+    When I click on "Delete Profile"
+    And I wait until I see "has been deleted" text
+    Then "sle_client" should not be registered
+
+  Scenario: Cleanup: remove leftover package of traditional client
```
srbarrios, line 90: Please add a comment line before this step, where we describe that this is a workaround for a reported bug, that will not be fixed.

testsuite/run_sets/secondary.yml
```diff
@@ -36,6 +36,7 @@
 - features/secondary/min_bootstrap_script.feature
 - features/secondary/min_ssh_tunnel.feature
 - features/secondary/min_activationkey.feature
+- features/secondary/min_migrate_to_trad.feature
```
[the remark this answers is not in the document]
srbarrios: yes it is. but let's rename the file to be more clear?
Maybe something like that?  trad_deleted_migrate_to_minion.feature 
Not sure, just something that describes that thing we test of registering a salt on a removed client, and is I understood well, we do reverse way too.

testsuite/features/secondary/min_migrate_to_trad.feature
vandabarata: I'm biased here because that's exactly what I did but yes, I'd add `@sle_client` and `@sle_minion` if you need both. But, from looking at your tests, do you _need_ `@sle_minion`? :thinking: you start off from the sle_client right? maybe that'd be enough for the tag here :thinking: I do think you should have a scope though but I'm not sure which of the following would be better: `@scope_salt`, `@scope_traditional_client` or maybe making a new scope like `@scope_client_migration` since it seems like its own specific use case

testsuite/features/secondary/min_migrate_to_trad.feature
nodeg: I adjusted the wording.

testsuite/features/secondary/min_migrate_to_trad.feature
nodeg: Fixed.

testsuite/features/secondary/min_migrate_to_trad.feature
nodeg: I left some comment and added the Bugzilla ID.

testsuite/features/secondary/min_migrate_to_trad.feature
nodeg: Thanks for the extensive comments. I added two tags/scops now.

testsuite/features/secondary/trad_deleted_migrate_to_minion.feature
```diff
@@ -0,0 +1,114 @@
...
+    And I enter "linux" as "password"
+    And I select "1-SUSE-KEY-x86_64" from "activationKeys"
+    And I select the hostname of "proxy" from "proxies"
+    And I click on "Bootstrap"
+    And I wait until I see "Successfully bootstrapped host!" text
+
+  Scenario: Wait until the Salt minion appears
+    When I wait until onboarding is completed for "sle_client" salt minion
+
+  Scenario: Check that the migrated system is now a minion
```
Bischoff, line 31: This scenario name is already used in `trad_migrate_to_minion.feature`

Ideally, it would be better to use globally unique scenario names (that do not exist yet)

Bischoff (changes requested):
see comment about scenario names

testsuite/features/secondary/trad_deleted_migrate_to_minion.feature
nodeg: I appended the context at the end of each scenario.

Bischoff (approved):
thanks for the fixes

vandabarata (approved):

testsuite/features/secondary/min_migrate_to_trad.feature
srbarrios: I still see both scenarios separated, but it's fine, we can keep like that.

ktsamis (approved):
looks great now, thanks Dominik

PR #4828, opened 2022-02-11
Provide link for CVEs

mbussolotto:
## What does this PR change?

Provide link for CVEs 

## GUI diff

Before:
now

After:
Screenshot from 2022-02-15 17-13-04

The links are shown only after a research
- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests
- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/16651

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
@@ -154,8 +154,8 @@ public class CVEAuditController {
             }
         }
         catch (UnknownCVEIdentifierException e) {
-            return json(res, ResultJson.error("According to our records, your systems/images are not affected by " +
-                    "this CVE. Please follow the provided links for more information."));
+            return json(res, ResultJson.error("The specified CVE number was not found in our database. " +
+                    "You can get more information about this CVE from the links given below"));
         }
     }
 
--- a/java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
@@ -22,6 +22,7 @@ import static com.suse.manager.webui.utils.SparkApplicationHelper.withUserPrefer
 import static spark.Spark.get;
 import static spark.Spark.post;
 
+import com.redhat.rhn.common.localization.LocalizationService;
 import com.redhat.rhn.domain.user.User;
 import com.redhat.rhn.manager.audit.CVEAuditImage;
 import com.redhat.rhn.manager.audit.CVEAuditManager;
@@ -56,6 +57,7 @@ import spark.template.jade.JadeTemplateEngine;
  * Spark controller class the CVE Audit page.
  */
 public class CVEAuditController {
+    private static final LocalizationService LOC = LocalizationService.getInstance();
 
     private static final Gson GSON = new GsonBuilder().create();
 
@@ -154,8 +156,7 @@ public class CVEAuditController {
             }
         }
         catch (UnknownCVEIdentifierException e) {
-            return json(res, ResultJson.error("The specified CVE number was not found in our database. " +
-                    "You can get more information about this CVE from the links given below"));
+            return json(res, ResultJson.error(LOC.getMessage("cveaudit.notfound")));
         }
     }
 
--- a/java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
@@ -23,6 +23,7 @@ import static spark.Spark.get;
 import static spark.Spark.post;
 
 import com.redhat.rhn.domain.user.User;
+import com.redhat.rhn.frontend.dto.CVE;
 import com.redhat.rhn.manager.audit.CVEAuditImage;
 import com.redhat.rhn.manager.audit.CVEAuditManager;
 import com.redhat.rhn.manager.audit.CVEAuditServer;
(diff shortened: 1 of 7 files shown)


web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -377,6 +377,29 @@ class CVEAudit extends React.Component<Props, State> {
               }}
             />
           </Table>
+          <p>
+          <AsyncButton
+            id="bootstrap-btn"
+            defaultType="btn-link"
+            icon="fa-external-link"
+            text={t("SUSE Security CVE link")}
+            action={() => window.open("https://www.suse.com/security/cve/CVE-" +
```
Etheryte, line 386: Is there a reason to use this logic and an `AsyncButton` instead of using a link like a few lines further down?

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -377,6 +377,29 @@ class CVEAudit extends React.Component<Props, State> {
...
+                          this.state.cveYear  +
+                          "-" + 
+                          this.state.cveNumber +
+                          ".html", '_blank','noopener')}
```
Etheryte, line 390: It would be nice to set both `noreferrer` in addition to `noopener`.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
mbussolotto: The reason is that the `btn-link` provides the icon that show more clearly that is an external link. But if you think it is better to use a link I can change it

Etheryte: If I'm not mistaken, we can use the same classes on links too, see a-la `subscription-matching-util.tsx`. Since we want to navigate to an address, a link is semantically better here.

mbussolotto: fixed

admd:
@mbussolotto we should move these links right below the search box where the user types the CVE number, here it is way down the page. Moreover, switch the places, 1st MITRE and then SUSE, simply because there is much more probability that record exists in MITRE.

Moreover, we should also come up with a more neutral message when we don't find a CVE record in our database, now that we are providing a MITRE and SUSE link about CVE.
`The specified CVE number was not found. This can happen for very old or yet-unknown numbers, please also check it for possible typing errors`

@jcayouette @moio @mcalmer  any suggestion on how we can make this message less verbose but correct and also give users a hint towards these new links.

Etheryte (approved):
The code changes look good, thanks a lot for taking the time to address the small details.

admd:
I propose changing
`The specified CVE number was not found. This can happen for very old or yet-unknown numbers, please also check it for possible typing errors.`

to 

`According to our records, your systems/images are not affected by this CVE. Please follow the provided links for more information.`


@mcalmer @moio any objection?

mbussolotto:
@Etheryte , sorry for re-asking you a review, but I add some code to show links only after an audit is triggered, and your opinion is always welcome :)

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -251,6 +256,35 @@ class CVEAudit extends React.Component<Props, State> {
               />
             </div>
           </p>
+          {this.state.auditExecuted === true && (
```
Etheryte, line 259: This can be just `{this.state.auditExecuted && (...)}`

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -251,6 +256,35 @@ class CVEAudit extends React.Component<Props, State> {
...
+                  rel="noopener noreferrer"
+                  href={
+                    "https://cve.mitre.org/cgi-bin/cvename.cgi?name=" + this.state.cveYear + "-" + this.state.cveNumber
+                  }
+                  data-senna-off="true"
+                >
+                  <IconTag type="external-link" />
+                  {t("MITRE CVE link")}
+                </a>
+                &nbsp;&nbsp;&nbsp;&nbsp;
```
Etheryte, line 273: Instead of this, I would recommend adding a margin. Perhaps @ncounter knows if we have any utility classes for stuff like this?

Etheryte:
Thanks for the ping and absolutely no worries. Two small things caught my eye, but nothing big.

admd:
> According to our records, your systems/images are not affected by this CVE. Please follow the provided links for more information

I like the position of the links, still not sure about the message itself when there is no record of the CVE in the database.

admd:
@mbussolotto please keep the original message with small modification `The specified CVE number was not found in our database. You can get more information about this CVE from the links given below`. 

This is a bit delicate situation and my previous suggestion might give the wrong sense of being secure. Rest is good. Thank you.

java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
```diff
@@ -154,9 +154,8 @@ public static Object cveAudit(Request req, Response res, User user) {
             }
         }
         catch (UnknownCVEIdentifierException e) {
-            return json(res, ResultJson.error("The specified CVE number was not" +
-                    " found. This can happen for very old or yet-unknown numbers, please" +
-                    " also check it for possible typing errors."));
+            return json(res, ResultJson.error("According to our records, your systems/images are not affected by " +
```
admd, line 157: We should extract this to our String Resource files so it could be translated as well.

mcalmer (approved):

web/html/src/manager/audit/cveaudit/cveaudit.tsx
ncounter: Unfortunately not, not yet I'd say :smiley: 

I'd suggest to add some like the following: https://github.com/uyuni-project/uyuni/blob/master/branding/css/theme.less#L401
That would also be in line with the way of doing it in the latest version of Bootstrap [Bootstrap 4](https://getbootstrap.com/docs/4.0/utilities/spacing/) which I expect to be a bigger work, but to happen in a certain point in time in the future (I'm looking at you @Etheryte :eyes: :wink: )

mackdk (approved):

PR #4835, opened 2022-02-14
QA: In order to test the feature that builds an image we need pxe_minion deployed

srbarrios:
## What does this PR change?

In order to test the feature that builds an image we need pxe_minion deployed.
Otherwise a scenario in that feature will fail, as there is no PXEBOOT_IMAGE variable set up, where it can check the image to use.

**WARNING**: 
It needs discussion and maybe a refactor. (Here my proposal https://github.com/uyuni-project/uyuni/pull/4836)
The method `compute_image_filename` (used in `I enter the image filename relative to profiles as "path"`) shouldn't require that we have or not a pxe_minion deployed in order to provide an image to use for the building procedure.

**Ideally, we should be able to test the building image process, without being strong coupled with the process of pxebooting an image built.**

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were improved

- [x] **DONE**

## Links

Ports:
- Manager-4.1
- Manager-4.2

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/buildhost_osimage_build_image.feature
+++ b/testsuite/features/secondary/buildhost_osimage_build_image.feature
@@ -10,6 +10,7 @@
 # which means "Enable Kiwi OS Image building"
 
 @buildhost
+@pxeboot_minion
 @scope_retail
 @scope_building_container_images
 Feature: Build OS images


srbarrios:
> Remember that there are more scenarios later which depends on this scenario. The all need to be skipped

This skip the whole feature, not only a scenario.
But if there are scenarios in other features that depends on this feature... then I suggest we merge better the refactor I made https://github.com/uyuni-project/uyuni/pull/4836

Bischoff (changes requested):
no :-)

PR #4841, opened 2022-02-15
Various Retail fixes

Bischoff:
## What does this PR change?

This PR:
* works around a problem with initial events on the branch server that has always been swept under the rug (and that would need real Salt debugging someday)
* does a "Hardware Refresh" on the proxy, so `proxy.example.org` is in SUSE Manager's database, and the pxeboot minion is seen behind the proxy
* avoids using a FQDN for the proxy (see technical details below)
* adds the proxy to the SERVERS group, so command-line tools like `retail_yaml` work

## Technical details

In Retail, don't do
```
    salt         CNAME   <fqdn of proxy>
```
as this will be replaced by
```
    salt         CNAME    my-proxy.tf.local
```
if you use avahi, and we removed avahi support.

Instead, just use:
```
    salt         CNAME    proxy
```

Using the FQDN was needed because it's the one that was in the SSL certificate. But now we have an alias for `proxy.example.org` in that certificate.

## Links

Ports:
* 4.1: SUSE/spacewalk#16994 SUSE/spacewalk#17002
* 4.2: SUSE/spacewalk#16995

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/core/proxy_branch_network.feature
+++ b/testsuite/features/core/proxy_branch_network.feature
@@ -187,13 +187,33 @@ Feature: Setup Uyuni for Retail branch network
     When I follow "States" in the content area
     And I click on "Apply Highstate"
     And I wait until event "Apply highstate scheduled by admin" is completed
-    And I disable repositories after installing branch server
     Then service "dhcpd" is enabled on "proxy"
     And service "dhcpd" is active on "proxy"
     And service "named" is enabled on "proxy"
     And service "named" is active on "proxy"
     And service "firewalld" is enabled on "proxy"
     And service "firewalld" is active on "proxy"
+
+@proxy
+@private_net
+  Scenario: Disable repositories after installing branch services
+    When I disable repositories after installing branch server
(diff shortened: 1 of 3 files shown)


vandabarata (approved):
lgtm :)

testsuite/features/core/proxy_branch_network.feature
```diff
@@ -187,14 +187,33 @@ Feature: Setup Uyuni for Retail branch network
...
     And service "firewalld" is enabled on "proxy"
     And service "firewalld" is active on "proxy"
 
+@proxy
+@private_net
+  Scenario: Disable repositories after installing branch services
+    When I disable repositories after installing branch server
+    # WORKAROUND: the following event fails because the proxy needs 10 minutes to become responsive again
+    # And I wait until event "Package List Refresh scheduled by (none)" is completed
+    And I wait for "700" seconds
```
srbarrios, line 203: I'm fine with the rest of the PR, but here better we do some call to the proxy until we assure it response what we want.

srbarrios (approved):
Approving, just please refactor this wait.

PR #4846, opened 2022-02-15
Account for non deterministic ipv6 address

vandabarata:
## What does this PR change?
Account for non deterministic IPv6 address being shown

## GUI diff
No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Ports
- [4.1](https://github.com/SUSE/spacewalk/pull/17056)
- [4.2](https://github.com/SUSE/spacewalk/pull/17055)

- [ ] **DONE**

## Changelogs
- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -53,14 +53,16 @@ Then(/^the IPv6 address for "([^"]*)" should be correct$/) do |host|
   node = get_target(host)
   interface, code = node.run("ip -6 address show #{node.public_interface}")
   raise unless code.zero?
+
   lines = interface.lines
-  ipv6_line = lines.grep(/inet6 2620:[:0-9a-f]*\/64 scope global temporary dynamic/).first
-  ipv6_line = lines.grep(/inet6 2620:[:0-9a-f]*\/64 scope global dynamic mngtmpaddr/).first if ipv6_line.nil?
-  ipv6_line = lines.grep(/inet6 fe80:[:0-9a-f]*\/64 scope link/).first if ipv6_line.nil?
-  next if ipv6_line.nil?
-  ipv6_address = ipv6_line.scan(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/).first
+  # selects only lines with IPv6 addresses and proceeds to form an array with only those addresses
+  ipv6_addresses_list = lines.grep(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/)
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -53,14 +53,16 @@
...
-  ipv6_line = lines.grep(/inet6 2620:[:0-9a-f]*\/64 scope global dynamic mngtmpaddr/).first if ipv6_line.nil?
-  ipv6_line = lines.grep(/inet6 fe80:[:0-9a-f]*\/64 scope link/).first if ipv6_line.nil?
-  next if ipv6_line.nil?
-  ipv6_address = ipv6_line.scan(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/).first
+  # selects only lines with IPv6 addresses and proceeds to form an array with only those addresses
+  ipv6_addresses_list = lines.grep(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/)
+  ipv6_addresses_list.map! { |ip_line| ip_line.slice(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/) }
+
+  # confirms that the IPv6 address shown on the page is part of that list and, therefore, valid
+  ipv6_address = find(:xpath, '//*[@id="spacewalk-content"]/div[4]/div[1]/div[1]/div[2]/table/tbody/tr[3]/td[2]').text
```
srbarrios, line 63: I personally don't like too much that kind of XPATHs based on HTML structure, I rather prefer queries using IDs, Names, CSS content, whatever. BUT, let's merge this, and maybe we can improve it in the future.

nodeg (approved):
LGTM.

Bischoff (approved):
Same remark as Oscar. If there is an id available it would be much better.

testsuite/features/step_definitions/common_steps.rb
nodeg: There are definitely some IDs/classes we can use here.
image

Bischoff: or just do like this:
```
<tr>
            <td>IPv6 Address:</td>
            <td>
                fe80::a8b2:92ff:fe42:97
            </td>
</tr>

td[text()='IPv6 Address']/next-sibling()
```

Bischoff (changes requested):
simplify xpath with `//"//td[text()='IPv6 Address:']/following-sibling::td[1]/text()` (please test in cucumber first, as you might need to trim the linefeeds)

PR #4853, opened 2022-02-16
Restart salt-minion to work around bsc#1196050

Bischoff:
## What does this PR change?

Restart salt-minion after adding an IP address and a domain name to the proxy.

## Links

Ports:
* 4.1: https://github.com/SUSE/spacewalk/pull/17023
* 4.2: https://github.com/SUSE/spacewalk/pull/17024

## Changelogs

- [x] No changelog needed


testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -29,25 +29,37 @@
 When(/^I stop salt-minion on "(.*?)"$/) do |minion|
   node = get_target(minion)
   ## Uyuni uses Salt Bundle. Package name is "venv-salt-minion"
-  pkgname = $product == 'Uyuni' ? "venv-salt-minion" : "salt-minion"
-  node.run("rc#{pkgname} stop", check_errors: false) if minion == 'sle_minion'
-  node.run("systemctl stop #{pkgname}", check_errors: false) if %w[ceos_minion ubuntu_minion kvm_server xen_server].include?(minion)
+  pkgname = $product == 'Uyuni' ? 'venv-salt-minion' : 'salt-minion'
```
srbarrios, line 32: This will collide with a refactor from @vzhestkov , now we use a function use_salt_bundle , instead of check for Uyuni product only.

testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -29,25 +29,37 @@
...
 
 When(/^I start salt-minion on "(.*?)"$/) do |minion|
   node = get_target(minion)
   ## Uyuni uses Salt Bundle. Package name is "venv-salt-minion"
-  pkgname = $product == 'Uyuni' ? "venv-salt-minion" : "salt-minion"
-  node.run("rc#{pkgname} restart", check_errors: false) if minion == 'sle_minion'
-  node.run("systemctl restart #{pkgname}", check_errors: false) if %w[ceos_minion ubuntu_minion kvm_server xen_server].include?(minion)
+  pkgname = $product == 'Uyuni' ? 'venv-salt-minion' : 'salt-minion'
+  os_version, os_family = get_os_version(node)
+  if os_family =~ /^sles/ && os_version =~ /^11/
```
srbarrios, line 46: Has this considered other systems included only in build validation setup?

srbarrios (approved):

testsuite/features/step_definitions/salt_steps.rb
Bischoff: ok, i'll ping Victor.

meaksh: Yes, this is definetely colliding with our changes. We will adjust our PR to also cover this case once we get latest changes from `master` into our branch.

Thanks for the note! :+1:

testsuite/features/step_definitions/salt_steps.rb
srbarrios: @Bischoff CentOS6 is not anymore in our BV envs, neither in CI. We are good about that one.
But I'm not sure about other cases, like debians, centos-like, others

PR #4464, opened 2021-11-12
Fixing renner/spacewalk#16202

thesp0nge:
Enabling HSTS for Apache so to force traffic to be in HTTPS. For Tomcat
this, should not be necessary since all the backend and the API are
served using secure protocol.

## What does this PR change?

**add description**

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation
- No documentation needed: **add explanation. This can't be used if there is a GUI diff**
- No documentation needed: only internal and user invisible changes
- Documentation issue was created: [Link for SUSE Manager contributors](https://github.com/SUSE/spacewalk/issues/new?template=ISSUE_TEMPLATE_DOCUMENTATION.md&labels=documentation&projects=SUSE/spacewalk/31), [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/new).
- (OPTIONAL) [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pulls)

- [ ] **DONE**

## Test coverage
- No tests: **add explanation**
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #
Tracks # **add downstream PR, if any**

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #4859, opened 2022-02-18
Hardening suggestion on keyfolder

thesp0nge:
## What does this PR change?

This PR changes permissions on /srv/susemanager/salt/salt_ssh/temp_bootstrap_keys folder. Since SUMA store here private key files that are created using 0600 permission schema, also folder must be accessible only by salt user.

## GUI diff

No difference.

- [ X ] **DONE**

## Documentation

- [ X ] **DONE**

## Test coverage
- No tests: simple permission enforcement.
- 
- [ X ] **DONE**

## Links

None

- [ X ] **DONE**

## Changelogs

Fri Feb 18 09:18:51 UTC 2022 - Paolo Perego <paolo.perego@suse.com>

- Hardening suggestion: permissions for
  /srv/susemanager/salt/salt_ssh/temp_bootstrap_keys directory can be 0700
  instead of 0755. In this directory SUMA will store ssh private key files and
  it will create them with 0700 permission, so the directory itself can have
  the same.


java/spacewalk-java.spec
```diff
@@ -706,7 +706,7 @@ chown tomcat:%{apache_group} /var/log/rhn/gatherer.log
 %dir %{_localstatedir}/lib/spacewalk
 %defattr(644,tomcat,tomcat,775)
 %attr(775, %{salt_user_group}, %{salt_user_group}) %dir %{serverdir}/susemanager/salt/salt_ssh
-%attr(775, %{salt_user_group}, %{salt_user_group}) %dir %{serverdir}/susemanager/salt/salt_ssh/temp_bootstrap_keys
+%attr(700, %{salt_user_group}, %{salt_user_group}) %dir %{serverdir}/susemanager/salt/salt_ssh/temp_bootstrap_keys
```
mcalmer, line 709: I fear that tomcat needs at least to "read" the files in this directory. We have code which check for files exist and call cleanup to remove them after bootstrapping is done. That remove is called via SALT API, so I think write access is not needed.

java/spacewalk-java.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Fri Feb 18 09:18:51 UTC 2022 - Paolo Perego <paolo.perego@suse.com>
+
```
mcalmer, line 3:
```suggestion
```

mcalmer: The header is added automatically when we submit

mcalmer (changes requested):
Doing the same changes in multiple PRs does not help us much to keep the overview.
Please close all duplicate PRs like https://github.com/uyuni-project/uyuni/pull/4464 
but please also add the required changes here to this PR.

There is an apache module missing for the proxy when we want to use the new headers.
See https://github.com/uyuni-project/uyuni/pull/4464#issuecomment-983642508

thesp0nge:
Hi @mcalmer, sorry my fault. I wanted to submit a different PR without ASTS changes

PR #4872, opened 2022-02-21
Automate bsscript version

mseidl:
## What does this PR change
This get's the bootstrap script  versions for supportconfig and dynamically adds the version to the bootstrap script header.
#16805 

- [ ] **DONE**

## Documentation
- No documentation needed: 

- [ ] **DONE**

## Test coverage
- N
## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation
- No documentation needed: 

- [ ] 
## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation
- No documentation needed: 

- [ ] **DONE**

## Test coverage
- No tests: 

- [ ] **DONE**

## Links 

https://bugzilla.suse.com/show_bug.cgi?id=11863361


mcalmer (approved):

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -252,8 +253,13 @@ def getHeader(productName, options, orgCACert, isRpmYN, pubname, apachePubDirect
     # Removing paths from options.gpg_key
     org_gpg_key = ",".join([os.path.basename(gpg_key) for gpg_key in options.gpg_key.split(",")])
     exit_call = " " if options.activation_keys or options.salt else "exit 1"
+    with cfg_component('web') as CFG:
+        version = CFG.version
+        if isUyuni:
+            version = CFG.uyuni
```
meaksh, line 259: This logic is wrong here as you are not calling `isUyuni()` but just checking it exists. 

I think what you need is:

```suggestion
        version = CFG.uyuni if isUyuni() else CFG.version
```

meaksh (changes requested):
Thanks for this. I just found a minor problem with the logic

spacewalk/certs-tools/rhn_bootstrap_strings.py
agraul: The "ugly" bit here (unrelated to you suggestion, that's toally valid!) is that `CFG.uyuni` is defined as `CFG.version.uyuni` in `rhn_web.conf` but accessed as `CFG.uyuni`.

meaksh (approved):

agraul (approved):
It seems that I'm the only one who dislikes accessing `web.version.uyuni` with `CFG.uyuni`, so let's keep it like that!

PR #4893, opened 2022-02-23
Require rng-tools to generate more entropy faster for RHEL systems.

sbluhm:
## What does this PR change?
**Issue:**
The Tomcat startup/rhn context can hang "forever" on creating a pg-jdbc connection (literally days).

Background:
pg-jdbc is waiting for entropy from `/dev/random` when creating the ssl connection. On RHEL there is a chance that not enough entropy is available.
This can be checked by running `cat /proc/sys/kernel/random/entropy_avail` (value < 200 is bad, above 2000 is nice).
`rngd` or `rng-tools` (random number generator daemon) increase the entropy significantly faster by tapping additional entropy sources.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


sbluhm:
This issue showed up first for me on 2022-02-11 resulting in consistent system unavailability.

**Would this fix be relevant for Leap Systems as well?**

sbluhm:
@mcalmer , can you please flag this as a merge candidate? This one is critical for RHEL users.

mcalmer:
Wow, i wonder if we can get similar problems with SUSE as well. 
@moio FYI

sbluhm:
This is very system dependent. We tried on virtual machines from 1cores/2GB to 4cores/18GB.
The 1core/2GB worked only with full (default) debug logging enabled on Tomcat and rhn. Took 2 mins startup instead of 40secs. Also tried on 4cores/16GB and it worked but that machine was Oracle so might have had rng already installed. Will actually check.
Other setups were not successful.
Actually can you check if rng is installed by default on Leap?

mcalmer:
rng is not installed on Leap. It even does not exist on SLE15.
On SLE15 we have "haveged" installed which seems to do the same, but in an old way.
But this explains, why we do not see these problems on SLE.

mcalmer:
On SLE as well as on Leap 15, havenged is "recommeded" by the base pattern.
rng-tools exists on Leap, but not on SLE.

sbluhm:
I have put `Recommends` here and only for RHEL8. It was needed on my setup. But might not be needed on all setups. Kernel 5.6 /dev/random does not block anymore. (https://github.com/torvalds/linux/commit/30c08efec8884fb106b8e57094baa51bb4c44e32)

What do you think about a message during setup? (i.e. rng != installed: "You do not have an enhanced random number generator installed. You might want to consider installing the package rng-tools")

java/spacewalk-java.spec
```diff
@@ -664,6 +667,11 @@ echo "#### SYMLINKS END ####"
 %service_add_pre taskomatic.service
 %endif
 
+%post
+%if 0%{?rhel}
+systemctl start rngd
```
mcalmer, line 672: While rng-tools is now only "Recommended": How does this behave if rng-tools is not installed?

sbluhm: good point, forgot about that line

mcalmer: Fine with me. :+1:

mcalmer (approved):

mcalmer:
I rebased this and merged

PR #4895, opened 2022-02-24
Reposync client cert check

mcalmer:
## What does this PR change?

Checking Client Certificates in reposync exist, but was disabled when using spacewalk-repo-sync.
This change enable the check by default and improved it by checking all certificates configured.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **manual tested**

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/11861
Tracks https://github.com/SUSE/spacewalk/pull/17112

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/backend/satellite_tools/reposync.py
+++ b/backend/satellite_tools/reposync.py
@@ -35,6 +35,7 @@ import errno
 import multiprocessing
 
 from rhn.connections import idn_puny_to_unicode
+from rhn.stringutils import ustr
 
 from uyuni.common.usix import raise_with_tb
 
@@ -360,14 +361,34 @@ def clear_ssl_cache():
     shutil.rmtree(ssldir, True)
 
 
+def verify_certificates_dates(certs):
+    cert = ""
+    is_cert = False
+    has_valid_certs = False
+    for line in certs.split('\n'):
+        if not is_cert and line.startswith("-----BEGIN CERTIFICATE"):
+            is_cert = True
+        if is_cert:
+            cert += line + '\n'
+        if is_cert and line.startswith("-----END CERTIFICATE"):
+            if not verify_certificate_dates(cert):
+                return False
+            cert = ""
+            is_cert = False
+            has_valid_certs = True
+    return has_valid_certs
+
+
 def get_single_ssl_set(keys, check_dates=False):
     """Picks one of available SSL sets for given repository."""
     if check_dates:
         for ssl_set in keys:
-            if verify_certificate_dates(str(ssl_set['ca_cert'])) and \
+            if verify_certificates_dates(ustr(ssl_set['ca_cert'])) and \
                 (not ssl_set['client_cert'] or
-                 verify_certificate_dates(str(ssl_set['client_cert']))):
+                 verify_certificates_dates(ustr(ssl_set['client_cert']))):
                 return ssl_set
+            else:
+                log(0, "++ Invalid Certificate found ++")
     # Get first
     else:
         return keys[0]
@@ -380,7 +401,7 @@ class RepoSync(object):
                  filters=None, no_errata=False, sync_kickstart=False, latest=False,
                  metadata_only=False, strict=0, excluded_urls=None, no_packages=False,
                  log_dir="reposync", log_level=None, force_kickstart=False, force_all_errata=False,
-                 check_ssl_dates=False, force_null_org_content=False, show_packages_only=False,
+                 check_ssl_dates=True, force_null_org_content=False, show_packages_only=False,
                  noninteractive=False, deep_verify=False):
         self.regen = False
         self.fail = fail
--- a/backend/spacewalk-backend.changes
+++ b/backend/spacewalk-backend.changes
@@ -1,3 +1,4 @@
+- enable check for client certificates in reposync
 - remove auto inherit of host entitlements for virtual guests
   * Fix reposync update notice formatting and date parsing (bsc#1194447)
   * supportconfig spacewalk-debug: extract task schedule data from db


backend/satellite_tools/reposync.py
```diff
@@ -360,14 +360,32 @@ def clear_ssl_cache():
     shutil.rmtree(ssldir, True)
 
 
+def verify_certificates_dates(certs):
+    cert = ""
+    isCert = False
```
meaksh, line 365: Small nitpick: could we use a more pythonic `is_cert` name here instead of camelCase formatting?

backend/satellite_tools/reposync.py
```diff
@@ -360,14 +360,32 @@ def clear_ssl_cache():
...
+    valid = True
+    for line in cert_str.split('\n'):
```
meaksh, line 367: I think this should be:

```suggestion
    for line in certs.split('\n'):
```

backend/satellite_tools/reposync.py
```diff
@@ -360,14 +360,32 @@ def clear_ssl_cache():
     shutil.rmtree(ssldir, True)
 
 
+def verify_certificates_dates(certs):
+    cert = ""
+    isCert = False
+    valid = True
```
meaksh, line 366: I would suggest this change together with other one to make this validation not passing when the string does not even contain a valid certificate string at all.

```suggestion
    valid = None
```

backend/satellite_tools/reposync.py
```diff
@@ -360,14 +360,32 @@ def clear_ssl_cache():
...
+    for line in cert_str.split('\n'):
+        if not isCert and line.startswith("-----BEGIN CERTIFICATE"):
+            isCert = True
+        if isCert:
+            cert += line + '\n'
+        if isCert and line.startswith("-----END CERTIFICATE"):
+            valid &= verify_certificate_dates(cert)
+            cert = ""
+            isCert = False
+    return valid
```
meaksh, line 376:
```suggestion
    return bool(valid)
```

backend/satellite_tools/reposync.py
```diff
@@ -360,14 +360,32 @@ def clear_ssl_cache():
...
+    cert = ""
+    isCert = False
+    valid = True
+    for line in cert_str.split('\n'):
+        if not isCert and line.startswith("-----BEGIN CERTIFICATE"):
+            isCert = True
+        if isCert:
+            cert += line + '\n'
+        if isCert and line.startswith("-----END CERTIFICATE"):
+            valid &= verify_certificate_dates(cert)
```
meaksh, line 373: I was wondering if we do really want to return `True` in case passing empty string, or string with no certificate at all. 

With the current proposal these values will pass the validation, what is your expectation here?

```suggestion
            valid = True if valid is None else valid & verify_certificate_dates(cert)
```

meaksh (changes requested):
Thanks for this Michael, I have found a possible issue with variable name and also made some suggestions to make the validation function to not pass when empty strings or not even a valid certificate is detected.

backend/satellite_tools/reposync.py
mcalmer: I think we do it different and abort on the first invalid certificate found.

meaksh (approved):
LGTM, Thanks! :+1:

backend/satellite_tools/reposync.py
```diff
@@ -35,6 +35,7 @@
 import multiprocessing
 
 from rhn.connections import idn_puny_to_unicode
+from rhn.i18n import ustr
```
agraul, line 38:
```suggestion
```
`ustr()` just calls `str()` in Python 3, we can use it directly in Python3-only code

backend/satellite_tools/reposync.py
```diff
@@ -360,14 +361,34 @@ def clear_ssl_cache():
...
+            has_valid_certs = True
+    return has_valid_certs
+
+
 def get_single_ssl_set(keys, check_dates=False):
     """Picks one of available SSL sets for given repository."""
     if check_dates:
         for ssl_set in keys:
-            if verify_certificate_dates(str(ssl_set['ca_cert'])) and \
+            if verify_certificates_dates(ustr(ssl_set['ca_cert'])) and \
```
agraul, line 386:
```suggestion
            if verify_certificates_dates(str(ssl_set['ca_cert'])) and \
```

backend/satellite_tools/reposync.py
```diff
@@ -360,14 +361,34 @@ def clear_ssl_cache():
...
                 (not ssl_set['client_cert'] or
-                 verify_certificate_dates(str(ssl_set['client_cert']))):
+                 verify_certificates_dates(ustr(ssl_set['client_cert']))):
```
agraul, line 388:
```suggestion
                 verify_certificates_dates(str(ssl_set['client_cert']))):
```

agraul (changes requested):
`ustr()` is not needed for Python3-only code as it only calls `str()`. IMO it makes more sense to keep `str()` instead of changing it for `ustr()`.

backend/satellite_tools/reposync.py
mcalmer: but `str()` does not work

agraul: Oh, that's probably because in `ustr` we ignore decoding errors.
```python
str(obj, 'utf8', errors='ignore')
```
That might a bad thing because ignoring means skipping "characters". On the other hand, if the `str` is only used to verify dates, it shouldn't make a difference if we ignore errors or handle them e.g. via `errors='replace'`.

mcalmer: No, it had something to do with "newline" (\n).
When using str() I was not able to call ".split()" on the result. 
the \n was escaped and it still result in 1 line.

Doing nothing let ".split()" fail.

Only with ustr() I got the result I wanted.

agraul: How can I debug this issue?

It's very surprising that you get this behavior, calling `str()` directly or through `ustr()` should not make a difference in that regard.

agraul (approved):
LGTM!

PR #4900, opened 2022-02-24
introduce schema migration tests for reportdb

mcalmer:
## What does this PR change?

Introduce schema migration tests for the new reportdb schema.
Add a workflow to test schema migrations on pull requests

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


juliogonzalez (commented):
Do I correctly understand we are basically duplicating all the scripts for testing, using the originals for the `susemanager` DB?

That's going to lead to a lot of trouble if we ever need to do maintenance of such scripts.

susemanager-utils/testing/docker/scripts/schema_migration_reportdb_test_pgsql.sh
```diff
@@ -0,0 +1,69 @@
...
+set -e
+
+echo "============================================================"
+echo "                      Migration test ReportDB               "
+echo "============================================================"
+
+cd /manager/susemanager-utils/testing/docker/scripts/
+
+# Move Postgres database to tmpfs to speed initialization and testing up
+if [ ! -z $PG_TMPFS_DIR ]; then
```
mackdk, line 12: Nitpicking here, but you can use `-n` instead of `! -z` and avoid the negated logic.

mackdk (commented):
The only concern here is the code duplication, as Julio already pointed out.

susemanager-utils/testing/docker/scripts/schema_migration_reportdb_test_pgsql.sh
mcalmer: we have this also in the other script. When we think about rewriting it into 1 script, we can think about it.
For now I would keep it.

mcalmer:
> Do I correctly understand we are basically duplicating all the scripts for testing, using the originals for the `susemanager` DB?
> 
> That's going to lead to a lot of trouble if we ever need to do maintenance of such scripts.

I see time problems and do not want to rewrite the scripts now. We can have schema testing for reporting DB or not.
De-duplication later after release  ( https://github.com/SUSE/spacewalk/issues/17187 )

PR #4902, opened 2022-02-25
Clearify CLM environment status message

ncounter:
## What does this PR change?

On a CLM environment status the message says `Generating repositories data` which is a misleading sentence because it suggests the generation is happening right now behind the scenes. That is not really true: reality is that it happens in `async` by the `channel-repodata-default` task, and depending on the schedule value of the task it can take a while **before starting** the generation of such data. So the message is that we are rather _waiting for the data to be generated_

## GUI diff

The words are slightly different

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: no tests for this a string

- [x] **DONE**

## Links

Fixes # just a helper spin-off of https://github.com/SUSE/spacewalk/issues/15261

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


Etheryte (approved):
Linguistically the sentence feels a bit clunky, but I'm not quite sure how to make it sound "native" while also remaining reasonably accurate. I take it that `"Waiting for repositories to be generated"` would be inaccurate since the repositories themselves already exist? `"Waiting for repositories to be populated"`? `"Waiting for repository data to be generated"` would be an option too, although again, perhaps not the most accurate.

In short, I don't know, I think we need to hire a linguist. 😄

From my side you can merge as-is as well since it's more accurate than what was there before.

ncounter:
> 



> Linguistically the sentence feels a bit clunky, but I'm not quite sure how to make it sound "native" while also remaining reasonably accurate. I take it that `"Waiting for repositories to be generated"` would be inaccurate since the repositories themselves already exist? `"Waiting for repositories to be populated"`? `"Waiting for repository data to be generated"` would be an option too, although again, perhaps not the most accurate.
> 
> In short, I don't know, I think we need to hire a linguist. smile
> 
> From my side you can merge as-is as well since it's more accurate than what was there before.

Thank you for the quick review.

I agree and I know my English is not a very good one :man_facepalming: 

I will ask a review from our doc guy who is native (cc @jcayouette).

About the `populated` option, I think that would be partially wrong as well because the message is both for when the channels are freshly created and empty, but also for when they get rebuilt so repodata are re-generated but channels are already _populated_ . Anyway, I am fully open to any better wording, my only care is about the meaning of the timing that can be extracted from the words.

For the record, it'd be much better to mention about the async chain and the task that is taking care of it, but we should stay transparent in that regards as it would add a weak point to the chain if in the future the mechanism will change.

cbbayburt (approved):
Nice catch, thanks, @ncounter!

I'm not sure if we need a changelog entry though. Even though it's super trivial for the user, it's still a user-facing change and one might get curious when they see a different message.

PR #4933, opened 2022-03-03
TESTSUITE Failing Scenario: Install package in the future and check for staging

avshiliaev:
## What does this PR change?

The `Install package in the future and check for staging` scenario is failing. 

According to the comments of the colleagues: 

> The package is not anymore cached. It's happening in all our latest testsuite runs on Uyuni/Head CI.

After further investigations:

> It seems our parameters are set up a bit in a way that makes this test bit flaky. 
>
> java.salt_content_staging_window = 0.033 (~1.98 minutes)
> java.salt_content_staging_advance = 0.05 (3 minutes)
> 
> Now in our test, we schedule it 2 minutes in the future https://github.com/SUSE/spacewalk/blob/f1ba1f514b700c74a5cd34524d1615ab84b8bd9e/testsuite/features/secondary/min_salt_install_with_staging.feature#L53
> 
> So IIUC,  I see that the staging window started even before the current time, I didn't look into the code thoroughly but it seems, that is causing the issue. So I would recommend either adjusting the parameters that our salt_content_staging_advance should at least start after the current time or increasing the time in our test from 2 minutes to at least 3 minutes. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage

- Cucumber tests were adjusted

- [x] **DONE**

## Links

Fixes [#](https://github.com/SUSE/spacewalk/issues/17083)

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_salt_install_with_staging.feature
+++ b/testsuite/features/secondary/min_salt_install_with_staging.feature
@@ -50,7 +50,7 @@ Feature: Install a package on the SLES minion with staging enabled
     And I follow "Install" in the content area
     When I check row with "orion-dummy-1.1-1.1" and arch of "sle_minion"
     And I click on "Install Selected Packages"
-    And I pick 2 minutes from now as schedule time
+    And I pick 3 minutes from now as schedule time
     And I click on "Confirm"
     Then I should see a "1 package install has been scheduled for" text
     And I wait until the package "orion-dummy-1.1-1.1" has been cached on this "sle_minion"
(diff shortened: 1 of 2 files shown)


admd:
Increasing to 3 minutes is just one way to get around this, we could also adjust the `javasalt_content_staging_window`, `java.salt_content_staging_advance` parameters that would reduce our 3 minutes needless wait.

@srbarrios can you help here to set these parameters better in testsuit.

srbarrios:
> Increasing to 3 minutes is just one way to get around this, we could also adjust the `javasalt_content_staging_window`, `java.salt_content_staging_advance` parameters that would reduce our 3 minutes needless wait.
> 
> @srbarrios can you help here to set these parameters better in testsuit.

I guess we best do it from sumaform isn't it?
In any case, for the sake of CI, I'm fine to merge this until we have these parameters tweaked.

avshiliaev:
@admd thank you for pointing to it. I can change those parameters instead. 
@srbarrios I could definitely tweak it in sumaform, justt don't know what's better now in terms of design.

srbarrios:
> @admd thank you for pointing to it. I can change those parameters instead. @srbarrios I could definitely tweak it in sumaform, justt don't know what's better now in terms of design.

@avshiliaev the main use of sumaform is to run a testsuite. And in other cases, this tweak doesn't seems to affect for demos or others. So, I think we can change the default values for a sumaform deployment even if they diverge from a production environment.

srbarrios (approved):

PR #4936, opened 2022-03-03
Simplify issue templates

renner:
## Description

This patch is to further simplify the issue templates, e.g. "Bugreport" is not really needed in the title as we are using the labels for indication. The current suggestions is leaving the title completely free.

## Changelogs

- [x] No changelog needed

## Re-run a test


juliogonzalez (commented):
I think keeping the version could still be useful.

What about changing `XXXX` to `XXXX.YY`, at least for the bug reports and questions?

juliogonzalez:
F

> > I think keeping the version could still be useful.
> > What about changing `XXXX` to `XXXX.YY`, at least for the bug reports and questions?
> 
> XXX was actually meant as a placeholder for the title in general, and thus was removed. Bug reports ask for the version explicitly in the description, maybe we should add it for questions as well in the description in order to not add visual clutter to the title?

Fine by me :-)

PR #4961, opened 2022-03-07
Remove spacewalk-client-cert, as we don't need it ship it anymore

juliogonzalez:
## What does this PR change?

Remove spacewalk-client-cert, as we don't need it ship it anymore. As discovered while submitting 4.1.14.

@uyuni-project/python feel free to disagree, if you feel maintenance and releng are wrong, but:

1. This package does not have relevant changes since 3.x, other than version bumps, and fixes for build issues
2. This package is not part any Server/Proxy/Client tools definition. And it's neither at any devel or released codestream for client tools
3. I could not find this package as Buildrequirement for other packages

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Cleanup

- [x] **DONE**

## Test coverage
- No tests: Cleanup

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/17199

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


meaksh:
I have no idea about this package. This sounds as a leftover from the traditional stack, and given the comments from the PR description I guess we are safe to drop this.

@mcalmer do you have any idea about this package?

juliogonzalez:
> I have no idea about this package. This sounds as a leftover from the traditional stack, and given the comments from the PR description I guess we are safe to drop this.
> 
> @mcalmer do you have any idea about this package?

Quoting mc:

> spacewalk-client-cert is part of the traditional stack. It provide a service to update the systemid file. AFAIK it was implemented to change the signature algorythm from md5 to sha256 some years ago.
I would say it is still a valid package which provide a service. But will go away when we deprecate the traditional stack.
> [...]
> ok, fine. When you are sure that we do not ship it, than it is obsolete

So I guess python team should have a look and confirm. As a matter of fact we are not submitting it, so it something is broken, nobody reported it since the times of 4.0.

Unless this is a build requirement I am missing.

meaksh (approved):
Thanks for the quote Julio.

AFAICS this package is also not required as dependency by any other package. And we do not ship it, nobody complained so far.

I would say that we could just wait to remove this once we remove all the rest of the traditional stack, or we can just remove it now and in case any customer surprisingly complains before traditional stack EOL, then we can just push it back.

I'm approving

juliogonzalez:
> Thanks for the quote Julio.
> 
> AFAICS this package is also not required as dependency by any other package. And we do not ship it, nobody complained so far.
> 
> I would say that we could just wait to remove this once we remove all the rest of the traditional stack, or we can just remove it now and in case any customer surprisingly complains before traditional stack EOL, then we can just push it back.

Since nobody complained since 4.0, it's most likely not needed, so let's remove.

PR #4964, opened 2022-03-07
QA: Add hostname and reverse lookup time sanity check

vandabarata:
## What does this PR change?
Adds a new validation to the sanity check stage of the CI to ensure the hostname and reverse name resolution don't take more than 2 seconds - else, fails the stage as there's something wrong with the network that will fail other important features.

## GUI diff
No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- [ ] **DONE**

## Links
Fixes https://github.com/SUSE/spacewalk/issues/17203
- [4.1](https://github.com/SUSE/spacewalk/pull/17220)
- [4.2](https://github.com/SUSE/spacewalk/pull/17221)

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -9,17 +9,25 @@ require 'nokogiri'
 
 Then(/^"([^"]*)" should have a FQDN$/) do |host|
   node = get_target(host)
+  initial_time = Time.now
   result, return_code = node.run('hostname -f', check_errors: false)
+  end_time = Time.now
+  resolution_time = end_time - initial_time
   result.delete!("\n")
   raise 'cannot determine hostname' unless return_code.zero?
+  raise "FQDN resolution for #{node.full_hostname} took too long (#{resolution_time} seconds)" unless resolution_time <= 2
   raise 'hostname is not fully qualified' unless result == node.full_hostname
 end
 
 Then(/^reverse resolution should work for "([^"]*)"$/) do |host|
   node = get_target(host)
+  initial_time = Time.now
   result, return_code = node.run("getent hosts #{node.full_hostname}", check_errors: false)
+  end_time = Time.now
+  resolution_time = end_time - initial_time
   result.delete!("\n")
   raise 'cannot do reverse resolution' unless return_code.zero?
+  raise "reverse resolution for #{node.full_hostname} took too long (#{resolution_time} seconds)" unless resolution_time <= 2
   raise "reverse resolution for #{node.full_hostname} returned #{result}, expected to see #{node.full_hostname}" unless result.include? node.full_hostname
 end
 


nodeg:
In this [run](https://ci.suse.de/blue/organizations/jenkins/manager-4.1-dev-acceptance-tests-PRV/detail/manager-4.1-dev-acceptance-tests-PRV/1458/pipeline/48) e.g. we have 2 scenarios that took 3s and 4s and in this [run](https://ci.suse.de/blue/organizations/jenkins/manager-4.1-dev-acceptance-tests-PRV/detail/manager-4.1-dev-acceptance-tests-PRV/1457/pipeline/48) one with 5s.
I do not think this is related to the DNS lookup but to e.g.

```cucumber
    And "ubuntu_minion" should communicate with the server using public interface
    And the clock from "ubuntu_minion" should be exact
```
Sadly we cannot see the times of the different steps but only for the complete scenario itself.

vandabarata:
@nodeg from what I could tell, it seems these were the steps where things took longer due to the DNS due to the timestamps in them during the scenarios where it took ~4min instead of ~30s when the DNS was down, hence why I added this here. But maybe @srbarrios knows more about it? :grimacing:

nodeg:
From my point of view, too. I was just wondering if it really were those 2 steps regarding DNS or maybe this one right after
 ```ruby
Then(/^"([^"]*)" should communicate with the server using public interface/) do |host|
  node = get_target(host)
  node.run("ping -c 1 -I #{node.public_interface} #{$server.public_ip}")
  $server.run("ping -c 1 #{node.public_ip}")
end
```
I mean we have no 100% guarantee it is the DNS step but I think we can be pretty sure. I am just wondering over the whole scenario. Eric and I had saw issues regarding `ping` 2 weeks ago. See https://github.com/SUSE/spacewalk/pull/17127 for example.

**Comment by nodeg** [comment]
Have a look here, that is from run [#3003](https://ci.suse.de/job/manager-Head-dev-acceptance-tests-NUE/3003/TestSuite_20Report/) the last run where the sanity check stage took 4m according to the picture in https://github.com/SUSE/spacewalk/issues/17203
image
Here it seems the DNS resolution and reverse resolution took only ms but checking the clock took 5s.

In the next scenario also from the same run the "client should communicate with the server using public interface" took 5s and DNS was fine as well.
image

**Review by srbarrios — APPROVED** [review]

File: testsuite/features/step_definitions/command_steps.rb [hunk 1]
```diff
@@ -9,17 +9,25 @@
 
 Then(/^"([^"]*)" should have a FQDN$/) do |host|
   node = get_target(host)
+  initial_time = Time.now
   result, return_code = node.run('hostname -f', check_errors: false)
+  end_time = Time.now
+  resolution_time = end_time - initial_time
   result.delete!("\n")
   raise 'cannot determine hostname' unless return_code.zero?
+  raise "FQDN resolution for #{node.full_hostname} took too long (#{resolution_time} seconds)" unless resolution_time <= 2
```
srbarrios, line 18: I would set it as 1 second, 2 is too much.

PR #4992, opened 2022-03-10
CI: Overwrite VERSION with VERSION.local

jordimassaguerpla:
## What does this PR change?

A VERSION.local file in susemanager-utils/testing/automation will
overwrite the variables in the VERSION file in the same directory.

This way, in the CI we can create a VERSION.local with different values
for the containers to use during testing.

The VERSION.local would be stored in the CI git project.

For example, we want to test with containers that have a different
postgresql version on certain jobs.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed
- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

First step for https://github.com/SUSE/spacewalk/issues/14922

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


juliogonzalez (changes requested):
We need the two files with the exact data for Uyuni and Head here, and not at `susemanager-ci`, and then use the right one.

Otherwise developers can't run the tests locally.

And of course, we should tell developers how to use one file or the other :-)

jordimassaguerpla:
> We need the two files with the exact data for Uyuni and Head here, and not at `susemanager-ci`, and then use the right one.
> 
> Otherwise developers can't run the tests locally.
> 
> And of course, we should tell developers how to use one file or the other :-)

Head uses containers from our internal infrastructure. This is why I thought we shouldn't put this info here which is public.

The CI code is private and contains also other internal data related. This is why I thought it was a better place.

juliogonzalez:
> > We need the two files with the exact data for Uyuni and Head here, and not at `susemanager-ci`, and then use the right one.
> > Otherwise developers can't run the tests locally.
> > And of course, we should tell developers how to use one file or the other :-)
> 
> Head uses containers from our internal infrastructure. This is why I thought we shouldn't put this info here which is public.

Not a problem having the URLs for the images. We got a configuration a long time a go that this is not a real issue.

> The CI code is private and contains also other internal data related. This is why I thought it was a better place.

Not sure I am understanding what you mean. What CI code? And how do you propose our own SUSE Manager developer can run the tests? Providing all the variables as exported environment variables? I think it would be good having two files and just telling our developers which one should be used for what (we could have `VERSION.uyuni` and `VERSION.susemanager`

jordimassaguerpla:
The idea is to have a VERSION.local that overwrite whatever you have in VERSION. I was thinking this strategy so we don't have to change the code we already have. I mean that we keep the VERSION file. I don't know what would break if we remove the VERSION file and have VERSION.uyuni and VERSION.spacewalk.

So my proposal is to have

VERSION
VERSION.spacewalk

and tell developers to rename VERSION.spacewalk to VERSION.local . The question was where to have VERSION.spacewalk. If you think here is a good place, then let's do it.

If you don't like having the VERSION.local, we could also have a VERSION.uyuni, a VERSION.spacewalk, and a symlink called VERSION to VERSION.uyuni, and then tell devs to change that link.

I am doubting which solution looks simpler.

@juliogonzalez which one you like best?

thanks

jordimassaguerpla:
> The idea is to have a VERSION.local that overwrite whatever you have in VERSION. I was thinking this strategy so we don't have to change the code we already have. I mean that we keep the VERSION file. I don't know what would break if we remove the VERSION file and have VERSION.uyuni and VERSION.spacewalk.
> 
> So my proposal is to have
> 
> VERSION VERSION.spacewalk
> 
> and tell developers to rename VERSION.spacewalk to VERSION.local . The question was where to have VERSION.spacewalk. If you think here is a good place, then let's do it.
> 
> If you don't like having the VERSION.local, we could also have a VERSION.uyuni, a VERSION.spacewalk, and a symlink called VERSION to VERSION.uyuni, and then tell devs to change that link.
> 
> I am doubting which solution looks simpler.
> 
> @juliogonzalez which one you like best?
> 
> thanks

Just to emphasize it. I want to keep a file called VERSION. My concern is that things keep working as they are for uyuni, so devs nor the CI has to be changed for uyuni. This is why I am suggesting to either have a VERSION.local file or to make the VERSION file be a symlink.

juliogonzalez:
> The idea is to have a VERSION.local that overwrite whatever you have in VERSION. I was thinking this strategy so we don't have to change the code we already have. I mean that we keep the VERSION file. I don't know what would break if we remove the VERSION file and have VERSION.uyuni and VERSION.spacewalk.
> 
> So my proposal is to have
> 
> VERSION VERSION.spacewalk
> 
> and tell developers to rename VERSION.spacewalk to VERSION.local . The question was where to have VERSION.spacewalk. If you think here is a good place, then let's do it.
> 
> If you don't like having the VERSION.local, we could also have a VERSION.uyuni, a VERSION.spacewalk, and a symlink called VERSION to VERSION.uyuni, and then tell devs to change that link.

I'd do something easier for developers and for the CI.

1. Have both `VERSION.uyuni` and `VERSION.susemanager`
2. Modify the shell scripts that are running `. $HERE/VERSION` to allow parameter, something like `-p' 
3. If the parameter is absent or if the value is `uyuni`, run `. $HERE/VERSION.uyuni`
4. If the parameter is present and the value is `susemanager`, run `. $HERE/VERSION.susemanager`
5. If the parameter is present and the value is neither `susemanager` or `uyuni`, exit with an error, or show the help

That way both developers and the CI can easily select what they want to test, SUSE Manager Head or Uyuni, but just specifying (or not) a parameter.

Then no need for a `.gitignore` to ignore `VERSION.local`, and no questions about "what was the copy command I needed!?"

You can use https://www.shellscript.sh/tips/getopts/ (bash builtin, no log names for the parameters, so recommended) or https://www.shellscript.sh/tips/getopt/ (not a builtin, usually available everywhere, but maybe not installed, so less recommended)

juliogonzalez:
> > > The idea is to have a VERSION.local that overwrite whatever you have in VERSION. I was thinking this strategy so we don't have to change the code we already have. I mean that we keep the VERSION file. I don't know what would break if we remove the VERSION file and have VERSION.uyuni and VERSION.spacewalk.
> > > So my proposal is to have
> > > VERSION VERSION.spacewalk
> > > and tell developers to rename VERSION.spacewalk to VERSION.local . The question was where to have VERSION.spacewalk. If you think here is a good place, then let's do it.
> > > If you don't like having the VERSION.local, we could also have a VERSION.uyuni, a VERSION.spacewalk, and a symlink called VERSION to VERSION.uyuni, and then tell devs to change that link.
> > 
> > 
> > I'd do something easier for developers and for the CI.
> > ```
> > 1. Have both `VERSION.uyuni` and `VERSION.susemanager`
> > 
> > 2. Modify the shell scripts that are running `. $HERE/VERSION` to allow parameter, something like `-p'
> > 
> > 3. If the parameter is absent or if the value is `uyuni`, run `. $HERE/VERSION.uyuni`
> > 
> > 4. If the parameter is present and the value is `susemanager`, run `. $HERE/VERSION.susemanager`
> > 
> > 5. If the parameter is present and the value is neither `susemanager` or `uyuni`, exit with an error, or show the help
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That way both developers and the CI can easily select what they want to test, SUSE Manager Head or Uyuni, but just specifying (or not) a parameter.
> > Then no need for a `.gitignore` to ignore `VERSION.local`, and no questions about "what was the copy command I needed!?"
> > You can use https://www.shellscript.sh/tips/getopts/ (bash builtin, no log names for the parameters, so recommended) or https://www.shellscript.sh/tips/getopt/ (not a builtin, usually available everywhere, but maybe not installed, so less recommended)
> 
> Hi,
> 
> I am not fan of this idea because we have to change all the .sh files and then also the jenkins files that use them. At the end of the day, this is a configuration and should be a configuration variable and not an option on a shell script.
> 
> Our .sh scripts, do
> 
> `. $HERE/VERSION`
> 
> if they start doing
> 
> `. $HERE/VERSION -p PRODUCT`

That's not what I said :-)

> it won't work. Doing a ` . SOMETHING` does not pass parameters to SOMETHING. Instead you need to use environment variables.
> 
> Then, if we use environment variables, we can set those variables outside the .sh scripts. This way we separate code and configuration management.

No, I am against environment variables. Developers won't remember about that. In fact all the scripts should have  a `-h`, but for now that's out of the scope and we can have a simple README.

Why do I know that developers won't remember? Because for when I need to manually use the jobs from `rel-eng/` for building and submitting to OBS (not building locally), I can't never, ever, remember what variables I need to export, and I end reviewing the scripts.

> And guess what? We are already setting an environment variable PRODUCT in our CI when we run our cucumber tests. Thus, I changed the code to use that variable instead. If not set, we load the VERSION.Uyuni . If set, we loat the VERSION.${PRODUCT}.
>
> I've reviewed the pipelines to make sure this variable is exported.

That's different. Developers don't use that pipeline and usually they usually don't even run cucumber locally (which is why also you are working with cucumber as PR).

Also: that was done as a quick hack.But since developers don't use it, we don' care
 
> I think this solution is simpler than the one I proposed initially and the one @juliogonzalez proposed.
>
> Let me know what you think of it.

Do as do you prefer, but for me the proper solution is a parameter, and yes: that also means changing the jobs/pipelines for Head (4.2 and 4.1 will eventually die, no need to adapt them).

jordimassaguerpla:
I did a mixed thing.  Developers can use the "-P PRODUCT" and CI can use the environment variable. I am doing this to make is easy for developers and to make it easy for the CI. At the CI, most of these scripts are called as a parameter of tito, and passing parameters to the tests is not straightforward (not to me).

I hope you like this now and we can merge. Let me know otherwise.

jordimassaguerpla:
I hope you like this now and we can merge. Let me know otherwise.

PR #5015, opened 2022-03-16
Testsuite: Fix os image name

nadvornik:
## What does this PR change?

OS Image name is now taken from Kiwi sources.

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/retail_steps.rb
+++ b/testsuite/features/step_definitions/retail_steps.rb
@@ -45,22 +45,26 @@ def compute_image_filename
   end
 end
 
-def compute_image_name
+def compute_image_name_version
   case $pxeboot_image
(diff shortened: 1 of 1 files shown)


testsuite/features/step_definitions/retail_steps.rb
```diff
@@ -63,6 +63,24 @@ def compute_image_name
   end
 end
 
+def compute_image_name_version
```
srbarrios, line 66: I would like to refactor these two methods `compute_image_name_version` and `compute_image_name` for the sake of maintenability.
An idea could be to return `name, version` in a method, so then from wherever we call them we can build the string using only first variable `name` or both "#{name}:#{version}"

srbarrios (approved):

PR #5040, opened 2022-03-21
Hub reporting beta2 taskomatic

lucidd:
## What does this PR change?

Adding a taskomatic job for aggregating reporting data from different susemanager servers

## GUI diff

No difference.

- [ ] **DONE**

## Documentation
- No documentation needed

- [ ] **DONE**

## Test coverage
- No tests

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateDriver.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateDriver.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2021 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package com.redhat.rhn.taskomatic.task;
+
+import com.redhat.rhn.common.conf.Config;
+import com.redhat.rhn.common.conf.ConfigDefaults;
+import com.redhat.rhn.common.hibernate.HibernateFactory;
+import com.redhat.rhn.domain.server.MgrServerInfo;
+import com.redhat.rhn.taskomatic.task.threaded.QueueDriver;
+import com.redhat.rhn.taskomatic.task.threaded.QueueWorker;
+
+import org.apache.log4j.Logger;
+import org.hibernate.Hibernate;
+import org.hibernate.query.Query;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+
+/**
+ * Hub Reporting DB Task Driver
+ */
+public class HubReportDbUpdateDriver implements QueueDriver<MgrServerInfo> {
+
+    private Logger log;
+    private static AtomicBoolean hasCandidates = new AtomicBoolean(true);
+
+    @Override
+    public void setLogger(Logger loggerIn) {
+        this.log = loggerIn;
+    }
+
+    @Override
+    public Logger getLogger() {
+        return log;
+    }
+
+    @Override
+    public List<MgrServerInfo> getCandidates() {
+        CriteriaBuilder builder = HibernateFactory.getSession().getCriteriaBuilder();
+        CriteriaQuery<MgrServerInfo> criteria = builder.createQuery(MgrServerInfo.class);
+        Query<MgrServerInfo> query = HibernateFactory.getSession().createQuery(criteria);
+        List<MgrServerInfo> mgrServerInfos = (List<MgrServerInfo>)query.list();
+        return mgrServerInfos.stream().filter(info ->
+                Optional.ofNullable(info.getReportDbLastSynced())
+                        .map(time -> time.toInstant().plus(24, ChronoUnit.HOURS).isBefore(Instant.now()))
+                        .orElse(true)
+        ).map(e -> {
+            Hibernate.initialize(e.getReportDbCredentials());
+           return e;
+        }).collect(Collectors.toList());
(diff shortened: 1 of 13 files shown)


juliogonzalez:
Ah right, this is not master.

Sorry about the noise @lucidd and @mcalmer. I just reviewed because the bot added releng as reviewers.

Maybe we should review and adapt the rules and avoid adding releng for branches that are not master, Manager-4.X-milestone or Uyuni-XXXX.YY.

java/code/src/com/redhat/rhn/taskomatic/task/ReportDBHelper.java
```diff
@@ -0,0 +1,115 @@
...
+     * @return write mode query
+     */
+    public static WriteMode generateInsert(Session session, String table, long mgmId, Set<String> params) {
+        final String sqlStatement = String.format(
+                "INSERT INTO %s (mgm_id, %s) VALUES (%s, %s)",
+                table,
+                String.join(",", params),
+                mgmId,
+                params.stream().map(p -> ":" + p).collect(Collectors.joining(","))
+        );
```
mcalmer, line 90: For these and the other generated queries I wonder if we need to quote something to prevent SQL injections and other things.

mcalmer (changes requested):
A few nitpicks but also some more serious things found.

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateDriver.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public Logger getLogger() {
+        return log;
+    }
+
+    @Override
+    public List<MgrServerInfo> getCandidates() {
+        Criteria criteria = HibernateFactory.getSession().createCriteria(MgrServerInfo.class);
```
mackdk, line 53: Nitpicking here but, Hibernate `Criteria`s are deprecated.

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
```diff
@@ -0,0 +1,154 @@
...
+                TABLES.forEach(table -> {
+                    updateRemoteData(remoteDB.getSession(), localRh.getSession(), table, mgrServerInfo.getId());
+                });
+                mgrServerInfo.setReportDbLastSynced(new Date());
+                ServerFactory.save(mgrServerInfo.getServer());
+                localRcm.commitTransaction();
+                log.info("Reporting db updated for server " + mgrServerInfo.getServer().getId() + " successfully.");
+            }
+            catch (RuntimeException ex) {
+                log.warn("Unable to update reporting db", ex);
```
mackdk, line 133: I see that we are not re-throwing the exception in case of error. Is the job going to be marked as failed in this case or will it look always succeed?

java/code/src/com/redhat/rhn/taskomatic/task/ReportDBHelper.java
mackdk: I guess that doing some checks is always good, but this looks safe to me. We are just replacing the table name (which are hardcoded in the java code) and the name of the fields (that come from a result set which is already sanitized by hibernate). The actual values are managed by the usual parameter substitution made by hibernate. But indeed this is a valid point to look deeper into.

schema/spacewalk/common/data/rhnTaskoBunch.sql
```diff
@@ -126,4 +126,7 @@ Parameters:
 INSERT INTO rhnTaskoBunch (id, name, description, org_bunch)
    VALUES (sequence_nextval('rhn_tasko_bunch_id_seq'), 'mgr-update-reporting-bunch', 'Update Reporting DB with current data', null);
 
+INSERT INTO rhnTaskoBunch (id, name, description, org_bunch)
```
mackdk, line 129: Don't we need a sql upgrade script as well for this and the other queries?

mackdk (commented):
Looks mostly good to me, added some comments

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
```diff
@@ -0,0 +1,154 @@
...
+        try {
+            parentQueue.workerStarting();
+            ConnectionManager localRcm = ConnectionManagerFactory.localReportingConnectionManager();
+            ReportDbHibernateFactory localRh = new ReportDbHibernateFactory(localRcm);
+            Credentials credentials = mgrServerInfo.getReportDbCredentials();
+            ConnectionManager remoteDBCM = ConnectionManagerFactory.reportingConnectionManager(
+                    credentials.getUsername(), credentials.getPassword(),
+                    "jdbc:postgresql://" + mgrServerInfo.getReportDbHost() + ":" +
+                            mgrServerInfo.getReportDbPort() + "/" + mgrServerInfo.getReportDbName()
+                    );
```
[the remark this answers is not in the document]
lucidd: i will make the function public for now. Do you already have a better place in mind?

java/code/src/com/redhat/rhn/taskomatic/task/ReportDbUpdateTask.java
```diff
@@ -29,42 +31,14 @@
...
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+
 
 
 public class ReportDbUpdateTask extends RhnJavaJob {
 
     private static final int BATCH_SIZE = 100;
 
-    @SuppressWarnings("unchecked")
```
mcalmer, line 43: Here we have also a BATCH_SIZE where we can use the value from the config.
It is 2 lines above this, but i cannot add the comment to the correct line

mcalmer:
@lucidd where do we configure the number of workers?

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateDriver.java
```diff
@@ -0,0 +1,89 @@
...
+                        .orElse(true)
+        ).map(e -> {
+            Hibernate.initialize(e.getReportDbCredentials());
+           return e;
+        }).collect(Collectors.toList());
+    }
+
+    @Override
+    public int getMaxWorkers() {
+        return 1;
```
mcalmer, line 72: I think this tell the number of parallel jobs. We need a configuration for this

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
lucidd: It should at least that's what happened when i got exceptions during development. The way the try catch is structured here was copied from the erratacacheworker so it should hopefully be right for those type of tasks.

mcalmer (approved):
LGTM

mackdk (approved):

PR #5048, opened 2022-03-22
update last checkin only if job is successful

mbussolotto:
## What does this PR change?

update last checkin only if job is successful

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17252

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mbussolotto:
@mcalmer , what do you think about this change? I tested and it works, but IDK if it is the best approach or if it can cause any side effect...do you have any better idea to solve the issue? Thanks

mbussolotto:
`executeSSHAction` in `SaltServerActionService` it's a taskomatic task and it correctly executed. `JobReturnEventMessageAction` is a spacewalk action, triggered by PGEventListener after any salt event is executed. So in our scenario, where a minion is down `executeSSHAction` fails but `JobReturnEventMessageAction` is executed as well, since a salt event has been executed. So yes, I think changes on `JobReturnEventMessageAction`  are enough. 
Adding @vzhestkov and @renner as reviewers, any other feedback is always appreciated :)

mcalmer (approved):

java/code/src/com/suse/manager/reactor/messaging/JobReturnEventMessageAction.java
```diff
@@ -248,6 +248,11 @@ public void execute(EventMessage msg) {
                 jobReturnEvent.getMinionId());
         if (minion.isPresent()) {
             MinionServer m = minion.get();
+            if (jobResult.isEmpty()) {
+                LOG.error("Do not update server info since job=" +  jobReturnEvent.getJobId() + " in minion=" +
```
renner, line 252: I am not sure if this should actually log as `error` here? Maybe for the context of not updating the server info, a log level of `info` could be enough, assuming that the actual failure of the job had been logged as `error` elsewhere.

mcalmer: `info` is not logged by default. But maybe `warn` ?

mbussolotto: changed in warn

Issue #4850, opened 2022-02-16
Uyuni 2022.01 - http_proxy not correctly used in reposync for mirrorlist repos


ddholstad99:
Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : uyuni-server-stable
Name           : Uyuni-Server-release
Version        : 2022.01-165.2.uyuni1
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes
Status         : up-to-date
Source package : Uyuni-Server-release-2022.01-165.2.uyuni1.src
Summary        : Uyuni Server
Description    : 
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.

## Details about issue
Configured http_proxy fails to download from internet following upgrade to latest uyuni version. Url via curl from command line does work. 

Syncing Channel 'rockylinux8-x86_64-ha' failed:

RepoMDError: Cannot access repository.
Repository 'rockylinux8-x86_64-ha' is invalid.
[rockylinux8-x86_64-ha|https://mirrors.rockylinux.org/mirrorlist?repo=HighAvailability-8&arch=x86_64&proxy=http%3A//proxy.divms.uiowa.edu%3A8888] Valid metadata not found at specified URL
History:
 - [rockylinux8-x86_64-ha|https://mirrors.rockylinux.org/mirrorlist?repo=HighAvailability-8&arch=x86_64&proxy=http%3A//proxy.divms.uiowa.edu%3A8888] Repository type can't be determined.

Please check if the URIs defined for this repository are pointing to a valid repository.
Skipping repository 'rockylinux8-x86_64-ha' because of the above error.
Could not refresh the repositories because of errors.

hbrown-uiowa:
I can reproduce this error by doing the following:
install opensuse leap from openSUSE-Leap-15.3-2-DVD-x86_64-Build24.5-Media.iso
choose system role -> server
set an fqdn for hostname

```
repo=repositories/systemsmanagement:/
repo=${repo}Uyuni:/Stable/images/repo/Uyuni-Server-POOL-x86_64-Media1/
zypper ar https://download.opensuse.org/$repo uyuni-server-stable
zypper ref
zypper in patterns-uyuni_server
```
[...]

PR #4953, opened 2022-03-05
Reposync: fix to ensure configured http proxy is used

wombelix:
## What does this PR change?

After upgrade of urlgrabber to >= 4, proxy variable wasn't set correctly due to missing type conversion.
Conversion from `byte` to `string` as well as unittest added. For further details, see related issue.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: bug fix

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes #4932

- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #5051, opened 2022-03-22
Reposync doesn't use http_proxy when syncing EL7 Repositories

wombelix:
## What does this PR change?

Sort of refactoring of https://github.com/uyuni-project/uyuni/pull/4953 to ensure other plugins / methods use the configured `http_proxy`.

In case of openSUSE Repositories, the issue occurred here (and was fixed in https://github.com/uyuni-project/uyuni/pull/4953 by adding `decode`)
https://github.com/uyuni-project/uyuni/blob/0aaf348aa3cad372206cac2b3381f17264b82c43/python/spacewalk/satellite_tools/download.py#L117

For EL7 bases repositories I could track it down to:
https://github.com/uyuni-project/uyuni/blob/0aaf348aa3cad372206cac2b3381f17264b82c43/python/spacewalk/satellite_tools/repo_plugins/yum_src.py#L526

See related Issue https://github.com/uyuni-project/uyuni/issues/4850

But any of the other `urlgrabber.urlgrab` calls will also end up in ignoring the `http_proxy` settings due to the byte / string conversion issue in `urlgrabber`.

I know that Monkey Patching should be avoided but I didn't saw a better way as long the behavior is not fixed directly in `urlgrabber` (see https://github.com/rpm-software-management/urlgrabber/issues/33), happy to discuss alternative approaches.

Successfully tested with openSUSE, Oracle Linux 8, CentOS 7 and Ubunut 20.04 Repositories.

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/4850

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/spacewalk/satellite_tools/download.py
+++ b/python/spacewalk/satellite_tools/download.py
@@ -114,7 +114,7 @@ class PyCurlFileObjectThread(PyCurlFileObject):
         self.parent = parent
         (url, parts) = opts.urlparser.parse(url, opts)
         (scheme, host, path, parm, query, frag) = parts
-        opts.find_proxy(url, scheme.decode("utf-8"))
+        opts.find_proxy(url, scheme)
         super().__init__(url, filename, opts)
 
     def _do_open(self):
--- a/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
@@ -80,6 +80,17 @@ RPM_PUBKEY_VERSION_RELEASE_RE = re.compile(r'^gpg-pubkey-([0-9a-fA-F]+)-([0-9a-f
 APACHE_USER = 'wwwrun'
 APACHE_GROUP = 'www'
 
+
+# Monkey Patch 'urlgrabber.grabber' method 'find_proxy' to enforce type string for variable 'scheme'
+# Workaround due to wrong byte/string handling in 'urlgrabber' package
+# Required by reposync to connect through http_proxy as configured
+def find_proxy(self, url, scheme):
+    urlgrabber_find_proxy(self, url, scheme.decode('utf-8'))
+
+urlgrabber_find_proxy = urlgrabber.grabber.URLGrabberOptions.find_proxy
+urlgrabber.grabber.URLGrabberOptions.find_proxy = find_proxy
+
+
 class ZyppoSync:
     """
     This class prepares a environment for running Zypper inside a dedicated reposync root
--- a/python/spacewalk/spacewalk-backend.changes
+++ b/python/spacewalk/spacewalk-backend.changes
@@ -1,3 +1,5 @@
+- Fix reposync issue, http_proxy ignored when syncing EL7 repositories
+
 -------------------------------------------------------------------
 Fri Mar 11 15:44:57 CET 2022 - jgonzalez@suse.com
 


wombelix:
@agraul sorry to mention / bother you directly, this PR is at some point related to https://github.com/uyuni-project/uyuni/pull/4953, which you were working on, so I thought you might be able to take a look on this one as well. Also it's already a month since I created it, would be cool to have a fix in Uyuni 2022.05, for 2022.04 it's probably too late. Thanks in advance. Cheers Dominik

agraul:
Hi @wombelix, don't worry about that. I had this on my list for today, sorry it took so long and you didn't get a response from us.

Overall I agree that monkey patching is not nice. If I understand everything correctly, your monkey-patch works even for synchronizing Debian repositories since `yum_src` is always imported in `reposync.py` and the monkey patching happens at import time. That is a bit hidden IMO.

I need to set up a Uyuni VM as my current dev env only has a SUMA instance with the old urlgrabber to test the change a bit.

For potential alternatives, I wonder how hard it would be to get rid of urlgrabber. It's not super well maintained and we bypass its abstraction anyway. I'll take a look at that as well.

wombelix:
> Hi @wombelix, don't worry about that. I had this on my list for today, sorry it took so long and you didn't get a response from us.
> 
No worries, thanks that you take a look.

> Overall I agree that monkey patching is not nice. If I understand everything correctly, your monkey-patch works even for synchronizing Debian repositories since `yum_src` is always imported in `reposync.py` and the monkey patching happens at import time. That is a bit hidden IMO.
> 
You are right, that's why I mentioned it's sort of refactoring of https://github.com/uyuni-project/uyuni/pull/4953 as well, I removed the workaround which was implemented earlier to use always the new Monkey Patching. But I agree, from the PR title and description, without taking a look in the related one, it's not clear that this also affect for example Debian.

> I need to set up a Uyuni VM as my current dev env only has a SUMA instance with the old urlgrabber to test the change a bit.
> 
Sure take your time :) We already had some tester in https://github.com/uyuni-project/uyuni/issues/4850 and so far it's all good on my Uyuni "Production" Instance, at least a good sign so far.

> For potential alternatives, I wonder how hard it would be to get rid of urlgrabber. It's not super well maintained and we bypass its abstraction anyway. I'll take a look at that as well.
>
I had that in mind too but even though I learned a lot about the reposync code base during the last weeks, especially due to the multiple http proxy issues, I was not sure where to start and was a little overwhelmed.

wombelix:
Please let me know if there is anything else I can do to further assist, I think it would be great to have a workaround in the next Uyuni release :)

meaksh:
@wombelix I have this now in my plate, I'll provide you some feedback soon. Let's try to make it for Uyuni 2022.06 :+1:

meaksh:
Hey @wombelix, we are moving this forward!

We have fixed the issue around `find_proxy` inside "urlgrabber" package to properly handle bytes and successfully injects the proxy settings: https://github.com/rpm-software-management/urlgrabber/pull/35. This is already on the way to Leap 15.4.

With this fix to "urlgrabber" in place, then we don't need this PR anymore and we will go in favor of https://github.com/uyuni-project/uyuni/pull/5572 to finally solve this.

Are you fine closing this PR? Thanks!

wombelix:
Thank you very much @meaksh, feel free to close this PR if it's not needed anymore due to the fix in the actual urlgrabber package :) Even better that way, no need to keep my hacky workaround then ;)

meaksh:
As mentioned, closing this PR in favor of https://github.com/uyuni-project/uyuni/pull/5572. Thanks!

PR #5055, opened 2022-03-23
Improve schedule info

parlt91:
## What does this PR change?

Add help icon and info text to recurring action and maintenance window pages

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- No documentation needed: 

- [ ] **DONE**

## Test coverage

- [ ] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/17151

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/maintenance/list/maintenance-windows-list.tsx
+++ b/web/html/src/manager/maintenance/list/maintenance-windows-list.tsx
@@ -38,6 +38,12 @@ const MaintenanceWindowsList = (props: MaintenanceListProps) => {
         title={t("Maintenance") + " " + (type === "schedule" ? t("Schedules") : t("Calendars"))}
         icon="spacewalk-icon-schedule"
         buttons={buttons}
+        helpUrl="reference/schedule/maintenance-windows.html"
+        summary={
+          type === "schedule"
+            ? t("Below is a list of Maintenance Schedules available to the current user.")
+            : t("Below is a list of Maintenance Calendars available to the current user.")
+        }
       >
         <div className="panel panel-default">
           <div className="panel-heading">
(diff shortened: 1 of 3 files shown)


web/html/src/manager/maintenance/list/maintenance-windows-list.tsx
```diff
@@ -38,6 +38,11 @@ const MaintenanceWindowsList = (props: MaintenanceListProps) => {
         title={t("Maintenance") + " " + (type === "schedule" ? t("Schedules") : t("Calendars"))}
         icon="spacewalk-icon-schedule"
         buttons={buttons}
+        helpUrl="reference/schedule/maintenance-windows.html"
+        summary={t(
+          "Below is a list of Maintenance {0} available to the current user.",
+          type === "schedule" ? t("Schedules") : t("Calendars")
```
Etheryte, line 44: Instead of this, I would recommend splitting this into two long translation strings. Otherwise the translation will be tricky to do in languages where the sentence order or logic is different.

web/html/src/manager/state/recurring-states-list.tsx
```diff
@@ -60,6 +60,18 @@ class RecurringStatesList extends React.Component<Props, State> {
         title={t("Recurring States")}
         icon="spacewalk-icon-salt"
         buttons={this.props.disableCreate ? [] : buttons}
+        summary={[
```
Etheryte, line 63: Perhaps we could use a fragment here instead of an array, a-la `summary={<>...</>}`?

Etheryte (approved):
Two small questions, but looks very solid in general. Thanks for improving this. 👍

web/html/src/manager/state/recurring-states-list.tsx
parlt91: Sure Done

web/html/src/manager/maintenance/list/maintenance-windows-list.tsx
parlt91: Yeah makes sense :+1:

PR #5078, opened 2022-03-24
Add `data-testid` support to the Select component

Etheryte:
## What does this PR change?

Add `data-testid` support to the shared Select component. This will allow us to easily fetch and set the value in Cucumber tests.  

I haven't updated any of the related test code since I'm not sure which parts need to be addressed, hence the WIP in the title. Perhaps @srbarrios or @ktsamis can give me a hand with this?  

Other than that, feel free to review.  

In general, does this approach work for you?

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -172,8 +172,8 @@ When(/^I check "([^"]*)" if not checked$/) do |arg1|
 end
 
 When(/^I select "([^"]*)" from "([^"]*)"$/) do |option, field|
-  xpath_option = ".//*[contains(@class, 'class-#{field}__option') and contains(text(),'#{option}')]"
-  xpath_field = "//*[contains(@class, 'class-#{field}__control')]/../*[@name='#{field}']/.."
+  xpath_option = ".//*[contains(@class, 'data-testid-#{field}-child__option') and contains(text(),'#{option}')]"
+  xpath_field = "//*[contains(@class, 'data-testid-#{field}-child__control')]/../*[@name='#{field}']/.."
   if has_select?(field, with_options: [option], wait: 1)
     select(option, from: field)
   else
@@ -905,7 +905,7 @@ Then(/^option "([^"]*)" is selected as "([^"]*)"$/) do |option, field|
   next if has_select?(field, selected: option)
 
   # Custom React selector
-  next if has_xpath?("//*[contains(@class, 'class-#{field}__value-container')]/*[contains(text(),'#{option}')]")
+  next if has_xpath?("//*[contains(@class, 'data-testid-#{field}-child__value-container')]/*[contains(text(),'#{option}')]")
 
   raise "#{option} is not selected as #{field}"
 end
@@ -918,7 +918,7 @@ When(/^I wait until option "([^"]*)" appears in list "([^"]*)"$/) do |option, fi
     break if has_select?(field, with_options: [option])
 
     # Custom React selector
-    break if has_xpath?("//*[contains(@class, 'class-#{field}__value-container')]/*[contains(text(),'#{option}')]")
+    break if has_xpath?("//*[contains(@class, 'data-testid-#{field}-child__value-container')]/*[contains(text(),'#{option}')]")
   end
 end
 
--- a/web/html/src/components/combobox.tsx
+++ b/web/html/src/components/combobox.tsx
@@ -2,6 +2,8 @@ import * as React from "react";
 
 import Creatable from "react-select/creatable";
 
+import withTestAttributes from "./input/select-test-attributes";
+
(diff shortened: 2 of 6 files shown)


Etheryte:
On further testing, this needs some additional work to cover all use cases, will update as necessary.

Etheryte:
In short summary, this update should allow us to keep all the current `name`-based checks still functional while also adding support for a test id.

srbarrios:
For that to make it work you need to set Capybara in `env.rb` 
https://rubydoc.info/github/jnicklas/capybara/Capybara#configure-class_method
```
test_id (Symbol, String, nil = nil) - Optional attribute to match locator against with built-in selectors along with id.
```
```
Capybara.configure do |config|
  config.test_id = 'data-testid'
end
```

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -886,7 +886,7 @@
   next if has_select?(field, selected: option)
 
   # Custom React selector
-  next if has_xpath?("//*[contains(@class, 'class-#{field}__value-container')]/*[contains(text(),'#{option}')]")
+  next if has_xpath?("//*[contains(@class, 'data-testid-#{field}-child__value-container')]/*[contains(text(),'#{option}')]")
```
srbarrios, line 889: If I'm not wrong, what Capybara will expect here is something like `has_xpath?("//*[contains(@data-testid, 'xxxx')]")`

Etheryte: I don't think that's correct, we're not simply looking for `data-testid` since selects use popouts which are moved in the regular DOM tree. To explain another way, the elements we're looking for might not be direct children of the element with the test id since they're detached to allow select dropdowns to work despite scroll overflow etc.

Etheryte: Who said that that was the purpose of this improvement? This PR is a fix for the test suite error and adds `data-testid` to the frontend component while keeping backwards compatibility. That's what's also in the PR description. It is not the point of this PR to go and retroactively update all existing tests to use `data-testid`. If you want to add the required Capybara configuration, go over the affected tests, etc feel free.

srbarrios: In other words, in product code, looking this line:
```
  const classNamePrefix = testClassName ? `class-${testClassName}-child` : undefined;
```
What prevents us to have
```
  const classNamePrefix = testClassName ? `data-testid-${testClassName}__value` : undefined;
```
So we avoid changes in test side and then is transparent? And then we don't even need to setup the data-testid Capybara configuration.

Etheryte: The `class-` prefix doesn't say anything about tests and since it doesn't show up as used code, looks safe to remove if you don't magically guess it's used by tests. The whole concept of using test ids etc is to be explicit that one binding or another is for testing purposes.

Etheryte: `data-testid` is present in the solution as well, the code you're quoting is for the children of that parent, hence the classname `data-testid-foo-child`. It doesn't conflict with `data-testid` in any way, namely in that it isn't even an id.

srbarrios: Fair enough. Go with your proposal. Just let it work and make it green. Fine for now ;)

srbarrios:
@Etheryte once you confirm you tested these changes, I approve and we merge. I don't have more comments from my side. The priority is to have these tests green again, so as far this is green, I'm fine for this PR.

web/html/src/components/combobox.tsx
```diff
@@ -20,6 +22,9 @@ type ComboboxProps = {
   selectedId?: (number | null | undefined) | (string | null | undefined);
   onFocus?: () => void;
   onSelect: (value: ComboboxItem) => void;
+
+  /** Id for testing purposes */
+  "data-testid"?: string;
```
cbbayburt, line 27: Since we might wanna use this in more components as a good standard for testing, can we make this a separate global type and extend the types with it where needed?

cbbayburt (approved):
LGTM (with one optional suggestion) :+1:

web/html/src/components/combobox.tsx
Etheryte: This is a simple change to make once we actually use it in more places than one, I wouldn't recommend abstracting it out unless we have somewhere else to put it as well.

ktsamis (approved):
Looks good to me even though I don't understand all of it. Please test it in the PR tests and if it passes the CLM tests all good

PR #5081, opened 2022-03-25
HW refresh stuck

mbussolotto:
## What does this PR change?
- Due to some issue on salt-api, the result of the action `hardware refresh` is not a serializable by `HwProfileUpdateSlsResult`: in this case the action remain stuck. Although the issue is in salt-api, we should log a warning to show when this raise condition happens. 
- Not sure if it's related to the HW refresh issue, but we noticed that the bootstrapping of a salt-ssh minion generates some async calls ( `addEntitlementToServer -> updatePillarAfterGroupUpdateForServers -> refreshPillar` ). Change the call made by refreshPillar sync

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17290

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/impl/SaltService.java
+++ b/java/code/src/com/suse/manager/webui/services/impl/SaltService.java
@@ -763,13 +763,13 @@ public class SaltService implements SystemQuery, SaltApi {
         try {
             LocalCall<Boolean> call = SaltUtil.refreshPillar(Optional.empty(),
                     Optional.empty());
-            callSync(call, minionList);
+            callAsync(call, minionList);
 
             // Salt pillar refresh doesn't reload the modules with the new pillar
             LocalCall<Boolean> modulesRefreshCall = new LocalCall<>("saltutil.refresh_modules",
                     Optional.empty(), Optional.empty(), new TypeToken<>() {
             });
-            callSync(modulesRefreshCall, minionList);
+            callAsync(modulesRefreshCall, minionList);
         }
         catch (SaltException e) {
             throw new RhnRuntimeException(e);
--- a/java/code/src/com/suse/manager/webui/services/impl/SaltService.java
+++ b/java/code/src/com/suse/manager/webui/services/impl/SaltService.java
@@ -763,13 +763,13 @@ public class SaltService implements SystemQuery, SaltApi {
         try {
             LocalCall<Boolean> call = SaltUtil.refreshPillar(Optional.empty(),
                     Optional.empty());
-            callAsync(call, minionList);
+            callSync(call, minionList);
 
             // Salt pillar refresh doesn't reload the modules with the new pillar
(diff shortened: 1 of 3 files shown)


java/code/src/com/suse/manager/webui/utils/salt/custom/HwProfileUpdateSlsResult.java
```diff
@@ -78,41 +82,65 @@
...
+        }
         return cpuInfo.getChanges().getRet();
     }
 
     /**
      * @return exported contents of the udevdb
      */
     public List<Map<String, Object>> getUdevdb() {
+        if (udevdb == null) {
+            LOG.warn("udevdb value is null");
```
mcalmer, line 108: Not sure if we log a warning in all cases. Not sure if on all architectures really all values exists.
But we can add that and check the logs if something appears always.

mbussolotto: I added this check only on the field that should be mandatory ( others are `Optional` ). So yes, we can keep it and if it is shown often we can change the return type in `Optional<List<Map<String, Object>>>`  and remove the warning.

java/code/src/com/suse/manager/webui/services/impl/SaltService.java
```diff
@@ -763,13 +763,13 @@ public void refreshPillar(MinionList minionList) {
         try {
             LocalCall<Boolean> call = SaltUtil.refreshPillar(Optional.empty(),
                     Optional.empty());
-            callAsync(call, minionList);
+            callSync(call, minionList);
```
vzhestkov, line 766: Is this change affects Salt Minions also?
For Salt Minions async calls are possible, but not for Salt SSH clients. Maybe I miss the condition somewhere, but anyway it's better to ensure.

mbussolotto: yes, it affects also salt minion: having an async call for salt minion and sync call for salt ssh it's possible but I don't know if it would overengineer it...waiting a feeback from @lucidd about that :)

vzhestkov: It could drop performance dramatically on performing sync calss to the minions. For Salt SSH async calls are not implemented on the salt side, so we have to use it in sync.

mbussolotto: After talking with @lucidd , we decided to revert this part and create an ad-hoc issue for this async issue

lucidd (approved):

PR #5101, opened 2022-03-29
New c3p0 dependencies

mcalmer:
## What does this PR change?

Use new c3p0 package from Factory. The difference is, that mchange-commons is not included, but a separate package which requires a new symlink in the jar directory for tomcat, taskomatic and search server.

This PR also cleaned up some old ant checks for outdated hibernate versions which would not work anymore.

It also simplify the filelist of spacewalk-java to include all jars which are available in the directory. Same happens already for taskomatic and search server.

Ivy repos were also adapted to work with the new package.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Partly fix https://github.com/SUSE/spacewalk/issues/17321

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/buildconf/build-props.xml
+++ b/java/buildconf/build-props.xml
@@ -106,14 +106,8 @@
   <property name="jaxb" value="${jaxb-api} ${jaxb-core} ${jaxb-runtime} ${txw2} ${istack-commons}"/>
 
 
-  <property name="hibernate3.2deps" value="hibernate3"/>
-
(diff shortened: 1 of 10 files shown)


mcalmer:
> As far as I can tell, the PR looks OK.
> 
> Not sure about the cleanup, but I see you remove hibernate 3.x, at the buildconf, which should be fine.
> 
> I am not sure about the complete removal of `%{serverdir}/tomcat/webapps/rhn/WEB-INF/lib/`. Is that right?

yes, instead of including the "dir" and every single file separate, we now add the directory with all files in it excluding some special, which are in a subpackage.

cbosdo (approved):
👍🏻

cbosdo (approved):
LGTM

PR #5128, opened 2022-03-31
Fix improper error handling on extracting package by preflight script

vzhestkov:
## What does this PR change?

Fixes improper error handling on checking if the package was extracted.

## GUI diff

No difference.

## Documentation

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
+++ b/susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
@@ -247,13 +247,13 @@ if [ -n "${VENV_SOURCE}" ]; then
             mv usr usr.tmp
             mv usr.tmp/lib/venv-salt-minion/* .
             rm -rf usr.tmp
+            if [ ! -x bin/python ]; then
+                rm -f "${VENV_TMP_DIR}/${VENV_HASH_FILE}"
+                exit_with_message_code "Error: Unable to extract the bundle from ${TEMP_DIR}/${VENV_PKG_FILE}!" 6
+            fi
         else
             cp -r "${VENV_INST_DIR}" "${VENV_TMP_DIR}"
             pushd "${VENV_TMP_DIR}" > /dev/null
-        fi
-        if [ -x bin/python ]; then
-            rm -f "${VENV_TMP_DIR}/${VENV_HASH_FILE}"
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
```diff
@@ -247,14 +247,14 @@ if [ -n "${VENV_SOURCE}" ]; then
             mv usr usr.tmp
             mv usr.tmp/lib/venv-salt-minion/* .
             rm -rf usr.tmp
+            if [ ! -x bin/python ]; then
+                rm -f "${VENV_TMP_DIR}/${VENV_HASH_FILE}"
+                exit_with_message_code "Error: Unable to extract the bundle from ${TEMP_DIR}/${VENV_PKG_FILE}!" 6
+            fi
```
mcalmer, line 253: Is moving the place correct? In my case it would mean that x bit of bin/python is not checked at all.

vzhestkov: yes, as it make sense only for extracting the package

mcalmer (approved):

juliogonzalez:
@mcalmer it seems you applied this change manually to the head and master testsuites. Meaning the current builds will maybe  work, but subsequent  fails won't work as this is not merged yet.

Please DO NOT merge this during the weekend. If the testsuites are broken, they will need to stay that way. If the fix worked, confirm here, and I will merge, publish and backport this on Monday.

juliogonzalez:
Oh wait, you approved already? Now I am even more confused.

juliogonzalez:
Ok, situation is unclear, so this will need to wait until Monday.

mcalmer:
`manager-Head-dev-acceptance-tests-NUE/3104` succeeded with the bootstrapping, but failed a bit later at a different point which has nothing todo with the preflight script.
To be sure, I started another run with a changed preflight script : `manager-Head-dev-acceptance-tests-NUE/3110`

srbarrios:
Uhmmm if the job above failed, are we sure the PR is correct?

juliogonzalez:
That's the bundle 3004 with python 3.10, so I think that's a different issue, and won't affect Uyuni for now, which is still using 3002 with Python 3.9.

Looks like a problem for 4.2 beta3.

PR #5138, opened 2022-04-03
Hub reporting little UI

mcalmer:
## What does this PR change?

Little UI for peripheral server 

## GUI diff

No difference.

After:

image

- [x] **DONE**

## Documentation
- together with general Hub doc

- [x] **DONE**

## Test coverage
- No tests: **manual**

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17296

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/src/manager/systems/details/mgr-server-info.tsx
```diff
@@ -0,0 +1,124 @@
...
+  reportDbPort: number;
+  reportDbUser: string;
+  reportDbLastSynced?: string;
+  isAdmin: boolean;
+  /** Locale of the help links */
+  docsLocale: string;
+};
+
+type State = {
+  messages: any[];
```
Etheryte, line 29: The type here is _probably_ `MessageType[]` unless I misunderstand the intent?

web/html/src/manager/systems/details/mgr-server-info.tsx
```diff
@@ -0,0 +1,124 @@
...
+};
+
+const msgMap = {
+  "Not a system id": t("Not a system id"),
```
Etheryte, line 33: While using proper strings here will work, I think in most cases we use "key-like" strings that we expect not to change over time a-la `image_overview_not_found`, `base_not_found_or_not_authorized` etc to give some examples from other files. This is not a must, but would be nice for consistency. (You can grep for `const msgMap` to see what I mean.)

web/html/src/manager/systems/details/mgr-server-info.tsx
```diff
@@ -0,0 +1,124 @@
...
+class MgrServer extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      messages: [],
+    };
+  }
+
+  handlePwReset() {
+    return Network.post(`/rhn/manager/api/systems/${this.props.serverId}/mgr-server-reportdb-newpw`, { })
```
Etheryte, line 47: Unless passing this empty object along with the post request is important, I would remove it. If it is important, I would suggest adding a comment why it's needed.

web/html/src/manager/systems/details/mgr-server-info.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <div>
+        {this.state.messages ? <Messages items={this.state.messages} /> : null}
+        <TopPanel
+         title={t("Peripheral Server")}
+         icon={"spacewalk-icon-suse-manager"}
+         helpUrl={"reference/systems/systems-list.html"}
+        >
+          <div className="panel panel-default">
+            <div className="panel-body">
+              <React.Fragment>
```
Etheryte, line 82: This fragment can safely be removed.  

<hr>

A slightly longer explanation if you're interested why:  

Fragments are necessary when you want to a-la return multiple root nodes from a function without a wrapper node, or if you want to conditionally render multiple nodes without a wrapper node etc. In a regular DOM tree like this, they are no-op and are safe to remove.  

Essentially, fragments are a crutch for React, which otherwise has a hard time figuring out what elements to replace and what to keep in situations where the number of returned children changes:  

```
render() {
  return (
    <React.Fragment>
      {(Math.random() > 0.5) ? (
        <div />
      ) : (
        <div />
        <div />
      )}
    </React.Fragment>
  );
}
```

If there's no multiple roots nor any conditional rendering like this involved, fragments are not necessary.

Etheryte:
The frontend part looks good. I had a few small suggestions, but nothing important. Please check out the failing checks before merging. You can run `yarn lint` to fix most lint issues automatically.

web/html/src/manager/systems/details/mgr-server-info.tsx
mcalmer: I was not aware that you can omit the parameter.

web/html/src/manager/systems/details/mgr-server-info.tsx
mcalmer: removed - I do not need to understand why I need it :-)

Etheryte (approved):

juliogonzalez:
@mcalmer why was releng added? What am I supposed to review?

java/code/src/com/suse/manager/webui/controllers/SystemsController.java
```diff
@@ -170,6 +178,78 @@ private ModelAndView virtualListPage(Request requestIn, Response responseIn, Use
...
+         catch (NumberFormatException e) {
+             LOG.error(String.format("SystemID (%s) not a long", sidStr));
+             return json(response, HttpStatus.SC_BAD_REQUEST, ResultJson.error("invalid_systemid"));
+         }
+         Server server = null;
+         try {
+             server = SystemManager.lookupByIdAndUser(sid, user);
+         }
+         catch (Exception e) {
+             LOG.error(e.getMessage());
```
mackdk, line 235: Nitpicking, but I would change this so the stacktrace of the exception is not lost:

```suggestion
             LOG.error(String.format("System (%s) cannot be retrieved", sid), e);
```

mackdk (approved):

mcalmer:
> @mcalmer why was releng added? What am I supposed to review?

First this PR was created against master and it included tons of other commits which are in hub-reporting-beta3. 
I think something in this branch triggered the releng automation

java/code/src/com/suse/manager/webui/controllers/SystemsController.java
mcalmer: The message we get from the exception is pretty good and the full stacktrace was not really helpfull.
So I decided to log only the message.

mackdk:
> First this PR was created against master

I had the same behaviour with my pull request for the `isPostgres()` method removal. When I rebased from master onto hub-reporting-beta3, python and release-engineering got triggered.

PR #5147, opened 2022-04-05
WIP: HTTP api

Bischoff:
## What does this PR change?

**NOT** ready for proofreading

This PR lets us choose to test whether with XML-RPC API or with HTTP API. The choice is central and we don't use both APIs at same time.

This PR also reduces the need for many logins:
* you now can keep the same API login between scenarios
* you now only log in once for all namespaces

Example of previous (broken) code:
```
Given(/^I am logged in API's actionchain namespace as user "(.*?)" and password "(.*?)"$/) do |luser, password|
  # Authenticate
  @actionchain_api = APIActionChainTest.new($server.full_hostname)
  @schedule_api = APIScheduleTest.new($server.full_hostname)
  @system_api = APISystemTest.new($server.full_hostname)
  @actionchain_api.login(luser, password)
  @system_api.login(luser, password)
  @schedule_api.login(luser, password)
end
```

## Links

Ports:

## Changelogs

- [x] No changelog needed


testsuite/features/support/api_test.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+    puts ret.status
+    puts ret.headers
+    puts ret.body
+    ret
+  end
+end
+
+# base class from where others inherit
+class APIBaseTest
```
srbarrios, line 44: I would keep this as an abtract class.
Then have two classes, for instance, XMLRPCAPITest and HTTPAPITest which will implement initiallze.
With that approach, when we start the concrete Cucumber feature we instantiate a variable that will have feature scope, and this variable will be used as APIBaseTest but instantated with XMLRPC or HTTP, and the rest will be transparent.

srbarrios: I mean, there are multiple ways to play with these classes so we can use both connections instead of commenting one or the other :)

Bischoff:
> This PR contains the commits from product changes but is in your fork, not in a feature branch, that's correct?

Sorry, this PR is not ready for proofreading yet.

This PR is based on Can's branch which is not merged yet.

testsuite/features/support/api_test.rb
Bischoff: Sorry, this test is not ready for proofreading yet.

Besides, APIBaseTest describes a test. Http_client describes a connection to the server. They are of different natures.
It is considered a helper, and is accessed via a member object (`@connection`).

Commenting out is only one of the options. We could also use a global variable, for example. But at some point at run time you have to decide which implementation to use.

Bischoff: I think we would just leave it all the time in HTTP form now on, but make it possible to switch back to XML-RPC any time with a single centralized change. I don't see the point of running both XML-RPC and HTTP at same time.

Bischoff: Discussed in slack

Bischoff:
Closed in favour of #5175

PR #5163, opened 2022-04-07
QE: Cucumber methods should use log method instead of puts

srbarrios:
## What does this PR change?

Few time ago we lost in a refactor the timestamps about when a scenario starts and ends.
This PR will get them back.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Ports:
- Manager-4.1
- Manager-4.2

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -85,7 +85,7 @@ enable_assertions
 # embed a screenshot after each failed scenario
 After do |scenario|
   current_epoch = Time.new.to_i
-  STDOUT.puts "This scenario took: #{current_epoch - @scenario_start_time} seconds"
+  log "This scenario took: #{current_epoch - @scenario_start_time} seconds"
   if scenario.failed?
     begin
       Dir.mkdir("screenshots") unless File.directory?("screenshots")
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

testsuite/features/support/env.rb
```diff
@@ -85,7 +85,7 @@ def enable_assertions
 # embed a screenshot after each failed scenario
 After do |scenario|
   current_epoch = Time.new.to_i
-  STDOUT.puts "This scenario took: #{current_epoch - @scenario_start_time} seconds"
+  log "This scenario took: #{current_epoch - @scenario_start_time} seconds"
```
nodeg, line 88: So we can use `log` here because we are in a Cucumber scope, correct?

nodeg (approved):
LGTM.

testsuite/features/support/env.rb
srbarrios: Right

srbarrios: I was not aware of it, until I tested...

PR #5166, opened 2022-04-07
New container doesn't use tftp package but python server

aaannz:
## What does this PR change?

New container doesn't use tftp package but python server so we need to change _service file to reflect this

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

Fixes #

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/containers/proxy-tftpd-image/_service
+++ b/containers/proxy-tftpd-image/_service
@@ -4,7 +4,7 @@
     <param name="file">Dockerfile</param>
     <param name="regex">%PKG_VERSION%</param>
     <param name="parse-version">patch</param>
-    <param name="package">tftp</param>
+    <param name="package">python3-fbtftp</param>
   </service>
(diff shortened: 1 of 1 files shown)


ncounter (approved):
LGTM :+1: 
I also think this should be forced to be in Beta3, if possible.
Better changing base package version now than later.
@juliogonzalez what do you think?

aaannz:
The issue to discus here that this change would change version of tftp container from 5.2.x to 0.5. I am not sure if this is ok. Other solution would be to use our custom version.

juliogonzalez:
I will join the Hexagon daily, something tells me having such versions is just wrong, and I just saw the same problem for other images.

aaannz:
Closing this as it was agreed to fix versioning in push script.

PR #5168, opened 2022-04-07
Move branding assets to web, fixing hot reload

Etheryte:
## What does this PR change?

This PR moves branding source assets to web as discussed previously and updates the build pipeline to account for that.  
This allows us to get dev reloads from Webpack rather than having to do a full manual deploy to test every change.  

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: build tooling.

- [x] **DONE**

## Links

First part of https://github.com/SUSE/spacewalk/issues/17232

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/src/branding/css/uyuni-typography.less
```diff
@@ -0,0 +1,18 @@
+@font-face {
+  font-family: 'Roboto Light';
+  font-style: normal;
+  font-weight: 300;
+  //src: url('Roboto/Roboto-Light.eot');
+  // TODO: Fix
+  // src: local('Roboto Light'), local('RobotoLight'), url('../fonts/Roboto/Roboto-Light.ttf') format('truetype');
```
Etheryte, line 7: A question for @ncounter , these can't currently resolve correctly with Webpack because they come from the frontend libs package. Is there a reason to not have this font directly in the assets similar to other fonts?

juliogonzalez (approved):
Should be OK... did you try to build on an OBS project, just in case?

In any case I am approving, but please merge when we unlock master, so it gets ready for 4.3 RC1 and Uyuni 2022.05.

web/html/src/branding/css/uyuni-typography.less
ncounter: Absolutely no reason at all to not move it. And there are others they could follow this train.
The historical reason for the `frontend-libs` was decided (I guess) at the beginning of the project: keep in github only our maintained and implemented code, everything from the outside jumps in at build time in ibs/obs. I believe this is an outdated statement and we already went for different solutions in different cases so, :+1: to move at least the trivial one.

ncounter (approved):
@Etheryte not only a :+1: but also a :heart: for this job.
Thank you for taking the time and the patience and the effort to achieve it.

PR #5172, opened 2022-04-07
Update tests to be compatible with uyuni AWS

maximenoel8:
## What does this PR change?
The purpose of this PR is to make uyuni deployment on sles server compatible with cloud tests.
Some of uyuni tests are using internal repositories.
The change include : 
 - change some tests to make it compatible with cloud
 - add tag to incompatible tests to skip its

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Related PR
Suma : 
https://github.com/uyuni-project/sumaform/pull/1056
Suma-ci : 
https://github.com/SUSE/susemanager-ci/pull/490

## Re-run a test


Bischoff (approved):
looks very clean to me

testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -191,8 +191,9 @@
   fill_in('command', with: cmd, fill_options: { clear: :backspace })
 end
 
-When(/^I enter target "([^"]*)"$/) do |minion|
-  fill_in('target', with: minion, fill_options: { clear: :backspace })
+When(/^I enter target "([^"]*)"$/) do |host|
+  value = $is_cloud_provider ? get_system_name(host) : SEARCH_VALUE[host]
```
srbarrios, line 195: Personally I don't see the advantage of having the SEARCH_VALUE hash.
If what we want in the remote command scenario is to target the hostname of one of our hosts, I would say it can works fine just by doing:

```suggestion
  value = get_system_name(host)
```

In all cases :thinking: 

Otherwise, if we want to continue having the previous behavior, then just do:

```suggestion
  value = $is_cloud_provider ? get_system_name(host) : host
```

Not modifying the regex values passed by parameter in the steps.

But I would go with the first option.

testsuite/features/support/env.rb
```diff
@@ -26,13 +26,14 @@
...
 DEFAULT_TIMEOUT = ENV['DEFAULT_TIMEOUT'] ? ENV['DEFAULT_TIMEOUT'].to_i : 250
+$is_cloud_provider = ENV["PROVIDER"].include? 'aws'
 
 # QAM and Build Validation pipelines will provide a json file including all custom (MI) repositories
 custom_repos_path = File.dirname(__FILE__) + '/../upload_files/' + 'custom_repositories.json'
 if File.exist?(custom_repos_path)
   custom_repos_file = File.read(custom_repos_path)
   $custom_repositories = JSON.parse(custom_repos_file)
-  $build_validation = true
+  $build_validation = !$is_cloud_provider
```
srbarrios, line 36: Are you running the testsuite using a custom_repositories.json file on it? Are you trying to run a BV in cloud?
I really think this value should not be touched in any case. This value will be true if we have a "custom_repositories.json" file in the upload_files folder. No need to worry about being or not in cloud.

testsuite/features/step_definitions/salt_steps.rb
srbarrios: Hi, as discussed in Slack, my opinion about it, is that I think is fine if we don't cover the test of the wildcard feature in this field, if by losing this coverage we gain in transparency between VMs/Cloud and we gain in maintenability.

srbarrios (approved):

PR #5142, opened 2022-04-04
QE: Add Cobbler buildiso tests

nodeg:
## What does this PR change?

This PR will

- extend the already implemented Cobbler steps to include checking for profiles
- add tests for Cobbler `buildiso`

in preparation of including Cobbler 3.3.2 to Manager 4.3 Beta 3.

Superseeds #5045 in order to merge it into `master-fixes-for-new-cobbler` instead of `master`.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17250

- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #5175, opened 2022-04-08
HTTP API tests

Bischoff:
## What does this PR change?

This PR lets us choose to test whether with XML-RPC API or with HTTP API. 

This PR also reduces the need for many logins:
* you now can keep the same API login between scenarios
* you now only log in once for all namespaces

Example of previous extremely redundant code:
```
Given(/^I am logged in API's actionchain namespace as user "(.*?)" and password "(.*?)"$/) do |luser, password|
  # Authenticate
  @actionchain_api = APIActionChainTest.new($server.full_hostname)
  @schedule_api = APIScheduleTest.new($server.full_hostname)
  @system_api = APISystemTest.new($server.full_hostname)
  @actionchain_api.login(luser, password)
  @system_api.login(luser, password)
  @schedule_api.login(luser, password)
end
```

The choice between XML-RPC and HTTP API is central at the moment, we currently don't use both APIs during a test suite run. We can easily add a `main.tf` option to choose one API or the other at sumaform level, or add cucumber steps to choose it from cucumber features.

List of things that might go to other PRs:
* rename `And I call foobar()` steps to hide technical details of API from cucumber's natural language
* run some features with HTTP and others with XML-RPC (or leave the choice in sumaform)

## Links

Solves SUSE/spacewalk#17283.

SUSE Manager CI PR: SUSE/susemanager-ci#525

Ports are only for code alignement as HTTP API is not supported on 4.1 and 4.2:
* 4.1: SUSE/spacewalk#17529
* 4.2: SUSE/spacewalk#17528

## Changelogs

- [x] No changelog needed


testsuite/features/build_validation/add_activation_keys/add_activation_key.template
```diff
@@ -8,4 +8,6 @@ Feature: Create an activation key for <client>
   I want to use activation keys
 
   Scenario: Create an activation key with the channel and child channels for a <client>
-    When I create an activation key including custom channels for "<client>" via XML-RPC
+    When I am logged in API as user "admin" and password "admin"
+    And I create an activation key including custom channels for "<client>" via API
+    And I logout from API
```
srbarrios, line 13: I suggest we hide login logout from the API.
And let only `I create an activation key including custom channels for "<client>" via API`, which inside will handle API connection. Those are too technical details that a user don't need to read in the scenario steps.

testsuite/features/step_definitions/api_common.rb
```diff
@@ -0,0 +1,622 @@
...
+require 'json'
+require 'socket'
+
+$api_test = APITestHTTP.new($server.full_hostname)
+
+## auth namespace
+
+When(/^I am logged in API as user "([^"]*)" and password "([^"]*)"$/) do |user, password|
+  $api_test.auth_login(user, password)
+end
```
srbarrios, line 13: I would go ahead and just login directly, not in a step, but just bellow the line 
``` 
$api_test = APITestHTTP.new($server.full_hostname)
```

testsuite/features/step_definitions/api_common.rb
```diff
@@ -0,0 +1,622 @@
...
+
+When(/^I logout from API$/) do
+  $api_test.auth_logout
+end
```
srbarrios, line 17: We don't really need, unless we specially want to cover the test of this.
Let's keep a unique connection open during the whole test suite.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -287,38 +288,37 @@
   raise 'distro ' + distro + ' does not exist' unless ct.distro_exists(distro)
 end
 
-Then(/^create profile "([^"]*)" as user "([^"]*)" with password "([^"]*)"$/) do |arg1, arg2, arg3|
+Then(/^create profile "([^"]*)" as user "([^"]*)" with password "([^"]*)"$/) do |profile, user, password|
   ct = CobblerTest.new
-  ct.login(arg2, arg3)
-  raise 'profile ' + arg1 + ' already exists' if ct.profile_exists(arg1)
+  ct.login(user, password)
+  raise 'profile ' + profile + ' already exists' if ct.profile_exists(profile)
```
srbarrios, line 294: I would be caution of changing code lines not related to that API handler.
As this could conflict with other PRs.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -870,6 +870,11 @@
   else # As Ubuntu has no support, must be CentOS/SLES_ES
     node.run('yum install wget', timeout: 600)
   end
+  registration_url = if $proxy.nil?
+                       "https://#{$server.full_hostname}/XMLRPC"
+                     else
+                       "https://#{$proxy.full_hostname}/XMLRPC"
+                     end
```
srbarrios, line 877: I wonder if that's part of HTTP API tests, or that's also out of scope...

testsuite/features/step_definitions/api_common.rb
Bischoff: See answer before. The test object and the connecton can stay during the whole test suite, but I think you have to re-login every while and then. It's same problem than `Given I am connected as user ...` for UI.

If this is a wrong assumption we can do as you say.

testsuite/features/step_definitions/api_common.rb
Bischoff: Again, see concern above.

testsuite/features/support/api_test.rb
```diff
@@ -0,0 +1,585 @@
...
+  end
+
+  def is_date?(attribute)
+    attribute.class == XMLRPC::DateTime
+  end
+
+  def date_now
+    now = Time.now
+    XMLRPC::DateTime.new(now.year, now.month, now.day, now.hour, now.min, now.sec)
+  end
```
srbarrios, line 562: I'm not sure about adding that logic at that layer.
What's the special need of this class that should handle API calls to provide a Date Helper? :thinking:

testsuite/features/support/api_test.rb
```diff
@@ -0,0 +1,585 @@
...
+    rescue ArgumentError
+      ok = false
+    end
+    ok
+  end
+
+  def date_now
+    now = Time.now
+    now.strftime('%Y-%m-%dT%H:%M:%S.%L%z')
+  end
```
srbarrios, line 584: Same than above

testsuite/features/support/commonlib.rb
```diff
@@ -41,16 +41,6 @@ def compute_list_to_leave_running
...
-# get registration URL
-# the URL depends on whether we use a proxy or not
-def registration_url
-  if $proxy.nil?
-    "https://#{$server.full_hostname}/XMLRPC"
-  else
-    "https://#{$proxy.full_hostname}/XMLRPC"
-  end
-end
-
```
srbarrios, line 53: Out of scope?

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -67,8 +67,9 @@
 
 Then(/^the system ID for "([^"]*)" should be correct$/) do |host|
   node = get_target(host)
-  step %(I am logged in via XML\-RPC actionchain as user "admin" and password "admin")
-  client_id = @system_api.search_by_name(get_system_name(host)).first['id']
+  $api_test.auth_login('admin', 'admin')
```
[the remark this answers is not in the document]
Bischoff: Most of it simply disappeared, and the rest went to `features/step_definitions/api_common.rb`.

testsuite/features/support/http_client.rb
```diff
@@ -0,0 +1,78 @@
...
+    # Get session cookie from previous calls
+    if params.nil?
+      session_cookie = nil
+    else
+      session_cookie = params[:sessionKey]
+      params.delete(:sessionKey)
+    end
+
+    # Call API
+    if short_name.start_with?('list', 'get', 'is', 'find') || ['logout', 'errata.applicableToChannels'].include?(name)
```
srbarrios, line 32: Nice :) 
That might be good to go to a constant if it goes longer, but nice for now.

testsuite/features/step_definitions/common_steps.rb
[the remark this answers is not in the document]
Bischoff: In the case of this line, I changed the syntax, but kept the logic unchanged.

I agree with you it is not nice. This was already so and I tried to change as few as possible, but I agree there is room for improvement.

testsuite/features/support/http_client.rb
```diff
@@ -0,0 +1,78 @@
...
+class HttpClient
+  def initialize(host)
+    puts 'Activating HTTP API'
+    @http_client = Faraday.new('https://' + host)
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  # rubocop:disable Metrics/MethodLength
```
srbarrios, line 16: Go ahead and add the whole file as exceptions in .rubocop_todo.yml :)

testsuite/features/support/xmlrpc_client.rb
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2022 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+require 'xmlrpc/client'
+
+# Wrapper class for XML-RPC client library
+class XmlrpcClient
```
srbarrios, line 7: Matter of taste, but I prefer XMLRPCClient :P

testsuite/features/step_definitions/common_steps.rb
Bischoff: Yes, it's only style changes, i should have done them in a different commit, and even in a different PR.

I can do that if you want, but it will take time. I wanted to be fast, this is why you have a bit of piggybacking here.

testsuite/features/step_definitions/common_steps.rb
Bischoff: I asked myself the same.

It's passed to `spacecmd` and this command line tool continues to use XML-RPC independently of my PR.

testsuite/features/support/api_test.rb
Bischoff: it's because the dates are handled differently in XML-RPC and HTTP. It's due to the fact that json file format cannot handle dates, while XML-RPC can because it's based on XML. This is why HTTP sees all dates as simple strings (with all formatting issues that might arise). If you ask me, that's a regression.

testsuite/features/support/api_test.rb
Bischoff: Same answer.

testsuite/features/support/commonlib.rb
Bischoff: For commonlib, yes. It's used at only one very specific point in our code and is not reusable.

testsuite/features/support/http_client.rb
Bischoff: I started with an array, but Can told me to do it so.

We have exactly the same logic in the product source code.

If you ask me, it's utterly dirty.

testsuite/features/support/http_client.rb
Bischoff: ok, i did not want to hide the dust under the carpet :-)

testsuite/features/support/xmlrpc_client.rb
Bischoff: I prefer to camel-case with one caps per "word". Matter of taste indeed.

testsuite/features/step_definitions/centos_tradclient.rb
```diff
@@ -1,68 +0,0 @@
-# Copyright (c) 2017-2019 SUSE LLC.
```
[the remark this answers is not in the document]
Bischoff: Most of it simply disappeared, and the rest went to `features/step_definitions/api_common.rb`.

testsuite/features/step_definitions/api_common.rb
srbarrios: It would be good to have a longer keep-alive-timeout

testsuite/features/secondary/min_bootstrap_api.feature
```diff
@@ -56,16 +54,14 @@ Feature: Register a Salt minion via XML-RPC API
...
     Given I am on the Systems overview page of this "sle_minion"
     Then I check for failed events on history event page
 
-  Scenario: Bootstrap via XML-RPC a non-existing system
-    Given I am logged in via XML-RPC system as user "admin" and password "admin"
-    When I call system.bootstrap() on unknown host, I should get an XML-RPC fault with code -1
-    And I logout from XML-RPC system namespace
+  Scenario: Bootstrap via API a non-existing system
+    Given I am logged in API as user "admin" and password "admin"
+    When I call system.bootstrap() on unknown host, I should get an API fault
```
srbarrios, line 63: Nice rephrase.

testsuite/features/secondary/min_docker_api.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  Scenario: Test "image.profiles" namespace
+    When I create and delete profiles via API
+    And I create and delete profile custom values via API
+    And I list image profiles via API
+    And I set and get profile details via API
+    And I logout from API
+
+# Note: image building via API is not tested here
+#       it is tested as another feature
```
srbarrios, line 21: Can we move this comment in the top of the file and then be specific with the filename/path of the other feature?

testsuite/features/secondary/trad_cve_audit.feature
```diff
@@ -73,21 +73,21 @@ Feature: CVE Audit on traditional clients
...
     When I follow the left menu "Admin > Task Schedules"
     And I follow "cve-server-channels-default"
     And I follow "cve-server-channels-bunch"
     And I click on "Single Run Schedule"
     Then I should see a "bunch was scheduled" text
     And I wait until the table contains "FINISHED" or "SKIPPED" followed by "FINISHED" in its first rows
-    And I am logged in via XML-RPC cve audit as user "admin" and password "admin"
-    When I call audit.list_systems_by_patch_status with CVE identifier "CVE-1999-9979"
+    And I am logged in API as user "admin" and password "admin"
+    When I call audit.list_systems_by_patch_status() with CVE identifier "CVE-1999-9979"
```
srbarrios, line 84: I'm going to comment it only here, but I think it's on multiple places.
I think using parentheses in our gherkin steps could be dangerous and are very technical, not at all natural language.
I would instead have something like: `I call "audit.list_systems_by_patch_status" method`

testsuite/features/support/api_test.rb
```diff
@@ -0,0 +1,587 @@
+# Copyright (c) 2011-2022 SUSE LLC.
```
srbarrios, line 1: WOW!
I just noticed something that I totally dislike!
We have move our organized and small classes to handle XMLRPC calls by namespace.
And we now have all at same file :(

We have now another file with more than 500 lines, I think this is really not a good practice.

srbarrios (changes requested):

testsuite/features/step_definitions/common_steps.rb
srbarrios: As you can see it will collide in  @nodeg PR
https://github.com/uyuni-project/uyuni/pull/5142/files#diff-75af4fb4d32d0e2675676f25807299a41f0afdccc6184ab7c255f4a43852e605R292-R297

I would better revert these lines.

testsuite/features/support/api_test.rb
srbarrios: My suggestion is to have an APIHandler, following a composition pattern, which handle all the multiple classes per API namespace.
It will expose these attributes or we can use getters, as we prefer, so we can have something like `api.my_namespace.my_method(...)`
We can also share the connection, by sharing something like `@connection` when we initialize the different classes, so it will login when APIHandler initiate and then share the connection.

testsuite/features/secondary/trad_cve_audit.feature
Bischoff: I would go even further. I am against giving technical details like this in cucumber. This should read:
```
When I list systems with CVE identifier "CVE-1999-9979"
```

I removed a few of these `And I call`, but there are some left. I did not have the courage to remove them all.

As for the parentheses, there were some cases with parentheses and some cases without. I unified with parentheses.

If you insist for me to "fix" this, I would like to change it to full natural language.

testsuite/features/secondary/min_docker_api.feature
Bischoff: ok

srbarrios:
Please check Rubocop issues

testsuite/features/secondary/trad_cve_audit.feature
srbarrios: Well, not insist as part of this PR, but it would be indeed very good contribution.

testsuite/features/secondary/min_bootstrap_api.feature
Bischoff: Thanks

testsuite/features/support/commonlib.rb
srbarrios: I mean that this was of out scope of this PR. But ok, fine.

testsuite/features/support/api_test.rb
Bischoff: After discussion in slack, fixed.

testsuite/features/support/api_test.rb
srbarrios: I would have a class or tool, to provide date_now depending your needs (xmlrpc, http).
Otherwise we are combining responsibilities in a unique classes.
"[single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)"

But I will not made a strong opinion on that. As you feel, just sharing thoughs.

testsuite/features/secondary/trad_cve_audit.feature
Bischoff: Agree. For later.

testsuite/features/step_definitions/common_steps.rb
Bischoff: ok.

testsuite/features/support/api_test.rb
Bischoff: Well, implementing practical details in derived classes is a basic principle of object-oriented programming... I can even point you to pages in reference books about that.

Bischoff:
> Please check Rubocop issues

done

testsuite/features/build_validation/add_activation_keys/add_activation_key.template
srbarrios: Because for every feature we are starting and stopping a chrome browser process, which is clean and doesn't have cookies of other instances. That impedes us to re-use the login between features. And at same time, it's easier to follow the state of the browser page.
And API is a bit different, we just send a request and have a response, so it's somehow stateless in that sense.

srbarrios:
I would still prefer a PR previous to that one, ONLY renaming xmlrpc_<> files to api_<> files.
As otherwise we lose the git history, we lose all the commits of the people that contributed and even more, now are shown as new files with copyright at 2022, when this is not true.

testsuite/features/support/http_client.rb
```diff
@@ -0,0 +1,70 @@
...
+  def initialize(host)
+    puts 'Activating HTTP API'
+    @http_client = Faraday.new('https://' + host)
+  end
+
+  def call(name, params)
+    # Adjust method name
+    name.sub!('auth.', '') if ['auth.login', 'auth.logout'].include?(name)
+    long_name = '/rhn/manager/api/' + name.tr('.', '/')
+    short_name = name.split('.')[-1]
```
srbarrios, line 17: It looks to me like a method to be extracted from here.

testsuite/features/support/http_client.rb
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2022 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+require 'faraday'
+
+# Wrapper class for HTTP client library (Faraday)
+class HttpClient
+  def initialize(host)
+    puts 'Activating HTTP API'
```
srbarrios, line 9: Maybe it should say `Activating HTTPs API`

testsuite/features/support/http_client.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    # Call API
+    if short_name.start_with?('list', 'get', 'is', 'find') || ['logout', 'errata.applicableToChannels'].include?(name)
+      # GET
+      unless params.nil?
+        long_name += '?'
+        params.each do |key, value|
+          long_name += '&' unless long_name[-1] == '?'
+          long_name += "#{key}=#{value}"
+        end
```
srbarrios, line 35: It also looks like a method to be extracted

testsuite/features/build_validation/add_activation_keys/add_activation_key.template
Bischoff: Makes sense.

OK, let's wait for Can's response.

The fact that the old code was logging in and out all the time makes me think it's the same as the UI, tokens have a limited lifespan, and probably less than 3 hours.

Bischoff:
> I would still prefer a PR previous to that one, ONLY renaming `xmlrpc_<>` files to `api_<>` files. As otherwise we lose the git history, we lose all the commits of the people that contributed and even more, now are shown as new files with copyright at 2022, when this is not true.

Unfortunately, that would not work. git would still lose the history at some point. Renames are not stored in git's metadata.

I tried to split in two commits, and that's what happens. It's a well known "issue" (although it's by design), see for example https://stackoverflow.com/a/32816379 .

Only possible workaround, `git log --follow` would do its hardest to reconstitute the renames.

testsuite/features/support/http_client.rb
Bischoff: ? there is only one HTTP.

If you mean `HTTP's API', it's incorrect too.

testsuite/features/support/http_client.rb
Bischoff: here it would make a bit more sense, ok

nodeg (commented):
I left some comments and suggestions. Other than that, good work Eric 👍 although I would have made several commits or even several PRs 😉

testsuite/features/support/http_client.rb
srbarrios: I mean that we are establishing a https connection, not a http connection.

nodeg: Then it should be renamed to `HTTPS`.

srbarrios:
> > I would still prefer a PR previous to that one, ONLY renaming `xmlrpc_<>` files to `api_<>` files. As otherwise we lose the git history, we lose all the commits of the people that contributed and even more, now are shown as new files with copyright at 2022, when this is not true.
> 
> Unfortunately, that would not work. git would still lose the history at some point. Renames are not stored in git's metadata.
> 
> I tried to split in two commits, and that's what happens. It's a well known "issue" (although it's by design), see for example https://stackoverflow.com/a/32816379 .
> 
> Only possible workaround, `git log --follow` would do its hardest to reconstitute the renames.

I would bet that if you do a PR only with renames, and then another with changes in these files, GitHub will be able to track it properly.

testsuite/.rubocop_todo.yml
```diff
@@ -274,17 +275,17 @@ Metrics/ParameterLists:
...
+  Exclude:
+    - 'features/support/http_client.rb'
 
 # Offense count: 8
 Naming/AccessorMethodName:
   Exclude:
-    - 'features/support/xmlrpc_activationkey.rb'
-    - 'features/support/xmlrpc_api.rb'
-    - 'features/support/xmlrpc_channel.rb'
-    - 'features/support/xmlrpc_user.rb'
```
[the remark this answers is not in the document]
Bischoff: Done. Is there a way also to tell it not to suggest autocorrects ?

testsuite/documentation/cucumber-steps.md
```diff
@@ -640,23 +640,16 @@ Note that the text area variant handles the new lines characters while the other
 
 <a name="b12" />
 
-#### XML-RPC
+#### XML-RPC or HTTP API
 
-* Log in and out some XML-RPC namespace on the server
+* Log in and out API on the server
```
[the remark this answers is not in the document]
Bischoff: Hmm. You not only login to the server, you also login to the API (you get a token that's valid only for the API).

But I agree correct English should be "to the API" and "from the API".

For now, I'll change it, but only in the docs.

testsuite/documentation/pitfalls.md
```diff
@@ -321,4 +303,3 @@ Steps files should be like:
 If you have a step `Click button` in a specific feature steps, then it is wrong. Those should be in "navigation_steps" or something similar.
 
 So in this case, the fact that `And I verify the results` was so abstract, it was already hinting that something was not right.
-
```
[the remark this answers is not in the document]
Bischoff: I think that's the contrary.

testsuite/documentation/software-components.md
```diff
@@ -29,12 +29,13 @@ Below is the list of Ruby gems used by the testsuite. It may change over time as
...
 * ```date```, ```time``: date and time manipulation functions
 * ```base64```, ```json```, ```nokogiri```: support for various encodings and data formats: base64, json, XML
-* ```net```, ```openssl```, ```uri```, ```open-uri```, ```xmlrpc```: support for various network protocols
+* ```net```, ```openssl```, ```uri```: support for network access
 * ```securerandom```: UUIDs and other random generation
 * ```socket```, ```stringio```, ```tempfile```, ```tmpdir```: file manipulation
 
 ### Various
 
+* ```open-uri```, ```xmlrpc```, ```faraday```: network and RFC helpers
```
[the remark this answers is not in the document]
Bischoff: RPC helpers :-) .

testsuite/features/secondary/min_config_state_channel_api.feature
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2021-2022 SUSE LLC.
```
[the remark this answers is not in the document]
Bischoff: Yup.

testsuite/features/secondary/min_docker_api.feature
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2017-2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+# Note: image building via API is not tested here
+#       it is tested in buildhost_docker_build_image.feature and
+#                       buildhost_docker_auth_registry.feature
```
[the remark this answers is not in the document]
Bischoff: yes

Bischoff:
> I would bet that if you do a PR only with renames, and then another with changes in these files, GitHub will be able to track it properly.

Maybe github for a given PR (although I would not bet on that), but not git.

And we need the history without going to individual PRs.

We also lost quite a lot of history when we did your renamings...

Doing that huge separation of work for just one PR is too much work, sorry.

testsuite/Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'rake'
 gem 'jwt'
 gem 'mime-types'
 gem 'xmlrpc'
+gem 'faraday', '1.10.0'
```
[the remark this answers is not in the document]
Bischoff: You are right, but I'd prefer to specify the version I have tested extensively.

testsuite/features/support/http_client.rb
Bischoff: Hmmm, it's the name we chose for this API in our RFC, in Can's presentation, in the interlock... I do not want to innovate here.

Basically, it works both over HTTP **and** HTTPS. We just happen to block the HTTP version for security reason. So it **is** an HTTP API that we use over HTTPS.

Funnily enough, XML-RPC is also HTTP-based :smile_cat:  . So the name is not even discriminating.

testsuite/features/build_validation/add_activation_keys/add_activation_key.template
Bischoff: Can's answer:

Eric Bischoff  19 h 04
```
do you know how long the token remains valid ?
```

Can Bulut Bayburt  19 h 04
```
same as the web UI
it's the same token
```

so I suggest to re-login at a feature level, as we do for the UI.

testsuite/config/cucumber.yml
```diff
@@ -33,7 +33,7 @@ configuration_channels: --tags @scope_configuration_channels
 content_staging: --tags @scope_content_staging
 proxy: --tags @scope_proxy
 traditional_client: --tags @scope_traditional_client
-xmlrpc: --tags @scope_xmlrpc
+api: --tags @scope_api
```
srbarrios, line 36: At least for history of discussions.
I discussed with Eric that I would prefer continue covering all the xmlrpc_xxx features we had in all our CI pipelines.
And add a basic set of new features, specific for http api.

So, my proposal was to use a first scenario that will set which kind of api it will be instantiated, depending the kind of test to pass.

So, about scopes, we will have two scopes: scope_http_api and scope_xmlrpc_api.

As the way Eric preferred to design it, the only feasible way to use one or the other is by changing a line of code (or a variable), but it will always affect all the testsuite.

Therefore, if we want to test HTTP API we are forced to lose all the test coverage of XMLRPC API tests.

testsuite/features/build_validation/add_activation_keys/add_activation_key.template
srbarrios: Actually, most of the features I'm reviewing have login/logut by scenario, multiple times in the same feature.
It doesn't seems the exception, maybe we should take another look and see if we can remove some of them.

testsuite/features/secondary/allcli_config_channel.feature
```diff
@@ -188,63 +188,59 @@ Feature: Management of configuration of all types of clients in a single channel
     And I should see a "-COLOR=red" text
 
 @sle_client
-  Scenario: Check configuration channel and files via XML-RPC for Traditional Client
-    Given I am logged in via XML-RPC configchannel as user "admin" and password "admin"
+  Scenario: Check configuration channel and files via API for traditional client
+    Given I am logged in API as user "admin" and password "admin"
```
srbarrios, line 192: This feature is all the time login-logout per scenario.

testsuite/features/secondary/buildhost_docker_auth_registry.feature
```diff
@@ -34,7 +34,9 @@ Feature: Build image with authenticated registry
     And I click on "submit-btn"
     Then I wait until I see "auth_registry_profile" text
     # Verify the status of images in the authenticated image store
-    When I wait at most 660 seconds until container "auth_registry_profile" is built successfully
+    When I am logged in API as user "admin" and password "admin"
```
srbarrios, line 37: This feature also do a pair of login-logout.

testsuite/features/secondary/min_action_chain.feature
```diff
@@ -37,7 +37,7 @@ Feature: Action chains on Salt minions
     And I wait until the table contains "FINISHED" or "SKIPPED" followed by "FINISHED" in its first rows
 
   Scenario: Pre-requisite: remove all action chains before testing on Salt minion
-    Given I am logged in via XML-RPC actionchain as user "admin" and password "admin"
+    Given I am logged in API as user "admin" and password "admin"
```
srbarrios, line 40: This feature also do login/logout multiple times

testsuite/features/secondary/min_docker_api.feature
```diff
@@ -0,0 +1,29 @@
...
+
+@scope_building_container_images
+Feature: API "image" namespace for containers and sub-namespaces
+
+  Scenario: Test "image.store" namespace
+    When I am logged in API as user "admin" and password "admin"
```
srbarrios, line 12: I would move login to another scenario above that one.
And logout to another scenario below the one where it is, as... we are mixing what the scenario name says, having login in one scenario and logout in another.

testsuite/features/secondary/min_empty_system_profiles.feature
```diff
@@ -1,18 +1,16 @@
...
+    Given I am logged in API as user "admin" and password "admin"
     When I call system.create_system_profile() with name "empty-profile" and HW address "00:11:22:33:44:55"
-    And I logout from XML-RPC system namespace
 
-  Scenario: Create an empty minion profile with hostname via XML-RPC API
-    Given I am logged in via XML-RPC system as user "admin" and password "admin"
+  Scenario: Create an empty minion profile with hostname via API
     When I call system.create_system_profile() with name "empty-profile-hostname" and hostname "min-retail.mgr.suse.de"
-    And I logout from XML-RPC system namespace
+    And I logout from API
```
srbarrios, line 13: Same as for other cases, I think login/logout might be better in independent scenarios, as otherwise we have a login in one scenario that aimes to do something, and then logout in another kind of scenario.

testsuite/features/secondary/srv_maintenance_windows.feature
```diff
@@ -0,0 +1,142 @@
...
+    And I follow the left menu "Systems > System Set Manager > Overview"
+    And I follow "Assign" in the content area
+    And I select "None - clear schedule" from "scheduleId"
+    And I click on "Clear All"
+    Then I should see a "Maintenance schedule has been cleared" text
+
+  Scenario: Cleanup: cancel all scheduled actions
+    Given I am logged in API as user "admin" and password "admin"
+    When I cancel all scheduled actions
+    And I logout from API
```
srbarrios, line 116: Why not encapsulate the authentication inside the step?

testsuite/features/secondary/srv_power_management_redfish.feature
```diff
@@ -77,12 +77,13 @@ Feature: Redfish Power management
...
 
   Scenario: Cleanup: reset Redfish values
-    Given I am logged in via XML-RPC powermgmt as user "admin" and password "admin"
+    Given I am logged in API as user "admin" and password "admin"
     And I want to operate on this "sle_minion"
     When I set power management value "" for "powerAddress"
     And I set power management value "" for "powerUsername"
     And I set power management value "" for "powerPassword"
     And I set power management value "ipmitool" for "powerType"
+    And I logout from API
```
srbarrios, line 86: Why we add logout here?
It was not need before.

testsuite/features/secondary/trad_action_chain.feature
```diff
@@ -27,9 +27,10 @@ Feature: Action chain on traditional clients
     And I wait until the table contains "FINISHED" or "SKIPPED" followed by "FINISHED" in its first rows
 
   Scenario: Pre-requisite: remove all action chains before testing on traditional client
-    Given I am logged in via XML-RPC actionchain as user "admin" and password "admin"
+    Given I am logged in API as user "admin" and password "admin"
     When I delete all action chains
     And I cancel all scheduled actions
+    And I logout from API
```
srbarrios, line 33: Same, we are adding additional logouts that are changing the behavior the tests had before this PR.
And they might not even bring any good.

testsuite/features/secondary/trad_action_chain.feature
```diff
@@ -228,7 +232,8 @@ Feature: Action chain on traditional clients
     Then I wait until there are no more action chains
     When I run "rhn_check -vvv" on "sle_client"
     Then file "/tmp/action_chain.log" should contain "123" on "sle_client"
-    And I wait until there are no more scheduled actions
+    When I wait until there are no more scheduled actions
+    And I logout from API
```
srbarrios, line 236: More extra logouts, that were not there before :/

testsuite/features/support/http_client.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  def call(name, params)
+    name.sub!('auth.', '') if ['auth.login', 'auth.logout'].include?(name)
+
+    # Get session cookie from previous calls
+    if params.nil?
+      session_cookie = nil
+    else
+      session_cookie = params[:sessionKey]
```
srbarrios, line 41: Why sometimes we use string keys and sometimes we use Ruby symbols? :thinking:

testsuite/features/support/http_client.rb
```diff
@@ -0,0 +1,79 @@
...
+        end
+      end
+    raise "Unexpected HTTP status code #{answer.status}" unless answer.status == 200
+
+    # Return either new session cookie or HTTP body
+    if name == 'login'
+      session_cookie = ''
+      cookies = answer.headers['Set-cookie']
+      cookies.split(',').each do |cookie|
+        if cookie.include?('pxt-session-cookie=') && !cookie.include?('Max-Age=0;')
```
srbarrios, line 67: I think it will be good to clarify better this part `&& !cookie.include?('Max-Age=0;')` with a comment, as doc for people coming to these lines later.

testsuite/features/step_definitions/api_common.rb
```diff
@@ -0,0 +1,622 @@
+# Copyright (c) 2015-2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+require 'json'
+require 'socket'
+
+$api_test = ApiTestHttp.new($server.full_hostname)
```
srbarrios, line 7:
```suggestion
if $product == 'Uyuni'
  $api_test = ApiTestHttp.new($server.full_hostname)
else
  $api_test = ApiTestXMLRPC.new($server.full_hostname)
end
```

testsuite/config/cucumber.yml
Bischoff: No, we can instanciate 2 variables, one for HTTP and one for XML-RPC. Then you can create a step to choose the API to use. It works, I have tested it. We can switch to a mix of HTTP and XML-RPC whenever we want. Or to a choice for sumaform.

The fact is that I do want to test intensively the new HTTP API, at least during some time. That means running **all** the tests with this API on at least one branch during some time. That's the reason why I don't want this "mix", at least at the beginning.

The solution for now will be to test Uyuni with HTTP and Head with XML-RPC. As I said, we can change that any time, But for now, please, let's have a few full runs with HTTP, so we don't miss any bug.

Besides that, a remark: the current `@scope_api` affects only 4 features. There's a whole lot of other features, of scenarios, and even of individual steps, that use the API in some manner.

testsuite/features/secondary/allcli_config_channel.feature
Bischoff: Yes, I did not want to change (and break) too many things at once. So you will find scenarios and steps that do individual login-logouts. This is ugly and would need some cleanup work.

testsuite/features/secondary/min_empty_system_profiles.feature
Bischoff: Yup. Like a "cleanup" step.

This is an example of where i moved from per-scenario to per-feature, but obviously it still needs some polishing.

testsuite/features/secondary/srv_maintenance_windows.feature
Bischoff: As said above, let's define as a clear strategy to move all authentication to feature prerequisite and cleanup.

testsuite/features/secondary/srv_power_management.feature
```diff
@@ -81,11 +81,12 @@ Feature: IPMI Power management
     And I should see a "Reboot" button
 
   Scenario: Cleanup: reset IPMI values
-    Given I am logged in via XML-RPC powermgmt as user "admin" and password "admin"
+    Given I am logged in API as user "admin" and password "admin"
     And I want to operate on this "sle_client"
     When I set power management value "" for "powerAddress"
     And I set power management value "" for "powerUsername"
     And I set power management value "" for "powerPassword"
+    And I logout from API
```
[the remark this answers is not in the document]
Bischoff: because we wnat to move to a per-feature login and logout.

testsuite/features/support/http_client.rb
Bischoff: rubocop has preferences about that

testsuite/features/support/http_client.rb
Bischoff: ok

testsuite/features/step_definitions/api_common.rb
```diff
@@ -0,0 +1,622 @@
+# Copyright (c) 2015-2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+require 'json'
+require 'socket'
+
+$api_test = APITestHTTP.new($server.full_hostname)
```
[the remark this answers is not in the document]
Bischoff: let's do what you suggest.

testsuite/features/step_definitions/api_common.rb
Bischoff: thanks for the code :smile_cat:

testsuite/features/secondary/srv_distro_cobbler.feature
```diff
@@ -142,13 +142,14 @@ Feature: Cobbler and distribution autoinstallation
...
 
-  Scenario: Create a cobbler system record via XML-RPC
-    When I am logged in via XML-RPC system as user "admin" and password "admin"
-    And I create a System Record
-    Then I wait until file "/srv/tftpboot/pxelinux.cfg/01-00-22-22-77-ee-cc" contains "ks=.*testserver:1" on server
-    And the cobbler report should contain "testserver.example.com" for cobbler system name "testserver:1"
+  Scenario: Create a cobbler system record via API
+    When I am logged in API as user "admin" and password "admin"
+    And I create a system record
+    And I logout from API
```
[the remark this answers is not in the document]
srbarrios: Agree on not re-using. But still thinking we don't have a strong need to expose at Gherkin level the steps of login/logout and we can consider to handle it encapsulated inside the step that call the API.

Anyway, resolving the discussions, fine for now.

testsuite/features/secondary/min_docker_api.feature
srbarrios: Can we just agree on create a card or a checkbox in the current card to don't forget about it?

testsuite/features/build_validation/add_activation_keys/add_activation_key.template
Bischoff: We definitely should. But in another PR.

testsuite/features/secondary/min_docker_api.feature
Bischoff: i'll create a card

nodeg (approved):
LGTM.

Bischoff:
The commit with Can's fixes will be removed before merging. I use it to test properly.

PR #5177, opened 2022-04-08
CI: add a timeout to "osc project" command

jordimassaguerpla:
## What does this PR change?

Sometimes, the wait for build command gets stucked. This seems to be an
issue with calling "osc project". It happens seldomly, but when it
happens, it never ends. The solution is to either run a rebuild or
cancel the command and call it again.

Thus, let's add a timeout to this command. If it gets stucked, let's run
it again. If after 2 retries this does not get fixed, meaning it is
again stucked, let's exit with an error.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed
- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

N/A

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/testing/automation/wait-for-builds.sh
+++ b/susemanager-utils/testing/automation/wait-for-builds.sh
@@ -9,6 +9,7 @@ usage() {
 api=https://api.opensuse.org
 config_file=$HOME/.oscrc
 lock="yes"
+osc_timeout="2h"
 
 while getopts ":a:c:p:u" o;do
     case "${o}" in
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
@srbarrios I know you are not obs/osc expert but you don't need to be for reviewing this PR :) . This is only about adding a timeout in case it gets stucked. This morning I found a build that was stucked for 14h .... pretty annoying.

susemanager-utils/testing/automation/wait-for-builds.sh
```diff
@@ -9,6 +9,7 @@ usage() {
 api=https://api.opensuse.org
 config_file=$HOME/.oscrc
 lock="yes"
+TIMEOUT="2h"
```
srbarrios, line 12: I'm fine with it, but could we use downcase as the others? Or uppercase in all of them :P

srbarrios (approved):

PR #5201, opened 2022-04-12
HTTP container image improvements

aaannz:
## What does this PR change?

- Do saltboot path mangling so it works with single saltboot url ( https://github.com/uyuni-project/retail/pull/106 , fix for failing testsuite in regular saltboot usage)
- Instead of copying ssl.conf, generate one from uyuni configure with correct ServerName and disabled SSL stampling

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


containers/proxy-httpd-image/Dockerfile
```diff
@@ -46,6 +46,4 @@ RUN sysconf_addword /etc/sysconfig/apache2 APACHE_MODULES proxy_http
 COPY uyuni-configure.py /usr/bin/uyuni-configure.py
 RUN chmod +x /usr/bin/uyuni-configure.py
 
-COPY ssl.conf /etc/apache2/vhosts.d/ssl.conf
```
ncounter, line 49: If so, then I think we can drop completely the `ssl.conf` file in the `containers/proxy-httpd-image` folder, right?

containers/proxy-httpd-image/uyuni-configure.py
```diff
@@ -124,17 +124,58 @@ def getIPs(fqdn: str) -> Tuple[str, str]:
...
+RewriteCond %{REQUEST_URI} ^/saltboot/image.*x86_64.*x86_64
+RewriteRule ^(.*)\.x86_64(.*\.x86_64-.*) $1$2 [N]
+
+RewriteCond %{QUERY_STRING} orgid=(\d+)
+RewriteRule "^/saltboot/image(.+)$" "/os-images/%1$1"  [R,L,QSD]
+''')
+
+    with open("/etc/apache2/vhosts.d/ssl.conf", "w") as file:
+        file.write(f'''
+<IfDefine SSL>
```
ncounter, line 153: Indentation as above

containers/proxy-httpd-image/uyuni-configure.py
```diff
@@ -124,17 +124,58 @@ def getIPs(fqdn: str) -> Tuple[str, str]:
...
 WSGIScriptAlias /tftpsync/delete /srv/www/tftpsync/delete''')
 
+    with open("/etc/apache2/conf.d/saltboot.conf", "w") as file:
+        # Saltboot uses the same URL regardles containerized or normal proxy
+        # here we rewrite URL so upstream server understands it
+
+        # First condition/rule removes double arch rule for backward compatility
+        # second changes /saltboot/image/..?orgid=<org> to /os-images/<org>/
+        file.write('''
+RewriteEngine on
```
ncounter, line 143: Could you please fix indentation here?

containers/proxy-httpd-image/uyuni-configure.py
```diff
@@ -124,17 +124,58 @@ def getIPs(fqdn: str) -> Tuple[str, str]:
...
+
+	DocumentRoot "/srv/www/htdocs"
+	ServerName {config['proxy_fqdn']}
+
+	ErrorLog /proc/self/fd/2
+	TransferLog /proc/self/fd/1
+	CustomLog /proc/self/fd/1   ssl_combined
+
+	SSLEngine on
+	SSLUseStapling  off
```
ncounter, line 165: What is the reason for turning this off?

ncounter (commented):
Globally speaking it LGTM, but I left a couple of comments before approving.
Thank you for the PR btw :+1:

containers/proxy-httpd-image/uyuni-configure.py
aaannz: I might be too hasty with this here. We do not support it with our generated certificates, so I though about disabling it to get rid of warning messages. But in case customer uses their own certificates, then this can be handy.

I will reenable it.

containers/proxy-httpd-image/uyuni-configure.py
aaannz: I do not follow, what indentation problem is here?

containers/proxy-httpd-image/Dockerfile
aaannz: Indeed, I forgot to remove it. Will do

containers/proxy-httpd-image/uyuni-configure.py
ncounter: I mean I'd prefer if the string file content starts being aligned with the space indentation of the line above 142, the same like [here](https://github.com/uyuni-project/uyuni/blob/master/containers/proxy-httpd-image/uyuni-configure.py#L87). It's probably personal feeling, we don't have rules about it.

aaannz: Ah, I did it like this because before I was asked to do so ( see https://github.com/uyuni-project/uyuni/pull/5089#discussion_r836103208 )

SchoolGuy: Yeah @ncounter I would personally not prefer to have it the way you linked since this will reduce the readability of the generated file.

ncounter: Np, I'm okay with that either. Orthogonal question: do we have checkstyle or something in the python environment?

@aaannz drop my comments about it then

ncounter (approved):
:+1:

containers/proxy-httpd-image/uyuni-configure.py
agraul: > do we have checkstyle or something in the python environment?

Currently not. I'm working on setting up pylint on the side. If it works out as planned, we'll use https://github.com/mbarkhau/pylint-ignore to not overwhelm us with existing linting errors/warnings.

PR #5215, opened 2022-04-14
Handle missing tables gracefully during hub synchronisation

mbussolotto:
## What does this PR change?

Currently synchronisation mechanism beetween hub and peripheral server assume that the db are exactly the same: this scenario is not always verified (e.g. during an upgrade, when the systems have different packages version).
The changes would like to enforce this functionalities, handling the case when a table in the hub does not exist in a peripheral server.

## GUI diff

After:

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17428

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
@@ -16,6 +16,7 @@ package com.redhat.rhn.taskomatic.task;
 
 
 import static com.redhat.rhn.taskomatic.task.ReportDBHelper.generateDelete;
+import static com.redhat.rhn.taskomatic.task.ReportDBHelper.generateExistingTables;
 import static com.redhat.rhn.taskomatic.task.ReportDBHelper.generateInsert;
 import static com.redhat.rhn.taskomatic.task.ReportDBHelper.generateQuery;
 
@@ -42,6 +43,8 @@ import org.hibernate.Session;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 
 public class HubReportDbUpdateWorker implements QueueWorker {
@@ -74,6 +77,23 @@ public class HubReportDbUpdateWorker implements QueueWorker {
     @Override
     public void setParentQueue(TaskQueue queue) {
         this.parentQueue = queue;
+    }
+
+    private List<String> filterExistingTables(Session remoteSession, Long serverId) {
+        SelectMode query = generateExistingTables(remoteSession, TABLES);
+        DataResult<Map<String, Object>> result = query.execute();
+        Set<Map.Entry<String, Object>> tableEntry = result.get(0).entrySet();
+        tableEntry.removeIf(t -> {
+            if (t.getValue() == null) {
+                log.warn("Table '" + t.getKey() + "' does not exist in server " + serverId + ": " +
+                        "this table will not be updated");
+                return true;
+            }
+            return false;
+        });
+        List<String> existingTables = tableEntry.stream().map(t ->
+                String.valueOf(t.getValue())).collect(Collectors.toList());
+        return existingTables;
     }
 
     private void updateRemoteData(Session remoteSession, Session localSession, String tableName, long mgmId) {
@@ -128,8 +148,8 @@ public class HubReportDbUpdateWorker implements QueueWorker {
                     );
             ReportDbHibernateFactory remoteDB = new ReportDbHibernateFactory(remoteDBCM);
             try {
-
-                TABLES.forEach(table -> {
+                List<String> existingTables = filterExistingTables(remoteDB.getSession(), mgrServerInfo.getId());
+                existingTables.forEach(table -> {
                     updateRemoteData(remoteDB.getSession(), localRh.getSession(), table, mgrServerInfo.getId());
                 });
                 ReportDBHelper.analyzeReportDb(localRh.getSession());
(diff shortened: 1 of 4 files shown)


java/code/src/com/redhat/rhn/taskomatic/task/ReportDBHelper.java
```diff
@@ -48,6 +49,17 @@ public static <T> Stream<DataResult<T>> batchStream(SelectMode query, int batchS
...
     }
 
+    /**
+     * Generated a query for checking if a table exists
+     * @param session session the query should use
+     * @param table table name
+     * @return select mode query
+     */
+    public static SelectMode generateExistsTable(Session session, String table) {
+        final String sqlStatement = "SELECT to_regclass('" + table + "') IS NOT NULL;";
```
mackdk, line 59: I would use an alias here in the query so you can refer to it in `checkRemoteData()` instead of using `?column?`

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
```diff
@@ -76,8 +78,19 @@ public void setParentQueue(TaskQueue queue) {
         this.parentQueue = queue;
     }
 
+    private boolean checkRemoteData(Session remoteSession, String tableName) {
+        SelectMode query = generateExistsTable(remoteSession, tableName);
+        DataResult<Map<String, PGobject>> exist = query.execute();
```
mackdk, line 83: If possible, I would avoid the explicit reference on `PGobject` and use a more standard `Map<String, Object>`. I'm not very familiar with the internals, but it is also possible that, since you are using a boolean now, it gets converted automatically by the underlying framework.

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
```diff
@@ -76,8 +78,19 @@ public void setParentQueue(TaskQueue queue) {
...
+        return Boolean.parseBoolean(String.valueOf(exist.get(0).get("?column?")));
+    }
+
     private void updateRemoteData(Session remoteSession, Session localSession, String tableName, long mgmId) {
         TimeUtils.logTime(log, "Refreshing table " + tableName, () -> {
+            if (!checkRemoteData(remoteSession, tableName)) {
+                log.warn("Table " + tableName + " does not exist in server " +
+                        mgrServerInfo.getServer().getId());
```
mackdk, line 91:
```suggestion
                log.warn("Table {} does not exist in server {}", tableName, mgrServerInfo.getServer().getId());
```
We should use formatting, instead of string concatenation now that we have migrated to log4j 2

mackdk (commented):
Looks good, I've added some minor remarks. 

The only thing that comes to my mind is that with this approach we are doing one additional query per table. We could in theory write a single query to extract the information for all the tables to sync at once.

I doubt that it would normally have a significant impact, but it might be a good idea to double check the performances on a slow connection.

mbussolotto:
@mackdk Thanks for the review ;) I tried using this query 
```
        final String sqlStatement = "SELECT to_regclass('" + table + "') IS NOT NULL; " + "" +
                "SELECT * FROM " + table +
                " WHERE mgm_id = 1 ORDER BY ctid OFFSET :offset LIMIT :limit";
``
but the DataResult is not what I expected, it contains only the result of the first `select`. Do you have better ideas for running just as a single query ?

**Comment by mackdk** [comment]
> Do you have better ideas for running just as a single query ?

I was thinking of a different approach and doing initially a single query like:

```
SELECT table_name FROM information_schema.tables WHERE table_name IN (:tables);
```

This will return as result the list of the tables that actually exists in the database when given as parameter the tables to synced defined in the `TABLES` list (but we need to deal with the case sensitivity). So then instead of doing `TABLES.foreach()` we can have `tablesThatExists.foreach()`. As I said, it's probably not worth changing the approach at this stage, so my suggestion was more to keep this in mind when looking at the performance numbers.

File: java/code/src/com/redhat/rhn/taskomatic/task/ReportDBHelper.java (hunk 1 above)
Comment (mackdk) [reply java/code/src/com/redhat/rhn/taskomatic/task/ReportDBHelper.java:59]:
```suggestion
        final String sqlStatement = "SELECT to_regclass('" + table + "') IS NOT NULL AS  table_exists;";
```

What I meant was to define the alias in the query with `AS`, so that the result column has a readable name that you can use as `COLUMN_KEY`.

mbussolotto:
ok, so I think this last implementation can be ok: instead of using the query you suggested, I  used `to_regclass`

mackdk (approved):
> I think this last implementation can be ok

Indeed, looks good to me :+1:

lucidd (approved):

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
```diff
@@ -76,6 +79,23 @@ public void setParentQueue(TaskQueue queue) {
...
     }
 
+    private List<String> filterExistingTables(Session remoteSession, Long serverId) {
+        SelectMode query = generateExistingTables(remoteSession, TABLES);
+        DataResult<Map<String, Object>> result = query.execute();
+        Set<Map.Entry<String, Object>> tableEntry = result.get(0).entrySet();
+        tableEntry.removeIf(t -> {
+            if (t.getValue() == null) {
+                log.warn("Table '" + t.getKey() + "' does not exist in session " + serverId + ": " +
+                        "this table will not be updated");
```
mcalmer, line 89: ... not exist in **session** ? In or on "server", right?

mcalmer (approved):
ok, just a small nitpick in the warning message

java/code/src/com/redhat/rhn/taskomatic/task/HubReportDbUpdateWorker.java
mbussolotto: right

PR #5217, opened 2022-04-14
Flush uyuni roster cache if the config has changed

vzhestkov:
## What does this PR change?

Improves the config management for `uyuni` roster module for salt with updating the cached data on changing the config for uyuni roster module.

In some cases if the pre flight script was disabled and then enabled back the cached roster data will not use the pre flight script.
The opposite situation is also possible.

## GUI diff

No difference.

## Documentation

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/modules/roster/uyuni.py
+++ b/susemanager-utils/susemanager-sls/modules/roster/uyuni.py
@@ -263,8 +263,8 @@ class UyuniRoster:
         h = self._execute_query(query)
         if h is not None:
             row = h.fetchone()
-            log.trace("db cache fingerprint: %s", row)
             if row and row.fp:
+                log.trace("db cache fingerprint: %s", row.fp)
                 new_fp = row.fp
                 log.trace("cache check: old:%s new:%s", cache_fp, new_fp)
                 if (
--- a/susemanager-utils/susemanager-sls/modules/roster/uyuni.py
+++ b/susemanager-utils/susemanager-sls/modules/roster/uyuni.py
@@ -62,6 +62,9 @@ class UyuniRoster:
     """
 
     def __init__(self, db_config, uyuni_roster_config):
+        self.config_hash = hashlib.sha256(
+            str(uyuni_roster_config).encode(errors="backslashreplace")
+        ).hexdigest()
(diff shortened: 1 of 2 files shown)


SchoolGuy (approved):
Change looks good to me!

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
```diff
@@ -62,6 +62,9 @@ class UyuniRoster:
     """
 
     def __init__(self, db_config, uyuni_roster_config):
+        self.config_hash = hashlib.sha256(
+            str(uyuni_roster_config).encode(errors="backslashreplace")
```
agraul, line 66: What's the type of `uyuni_roster_config`?

vzhestkov: It's a dictionary.

agraul: Ah, ok. The order of the keys is stable, right?

SchoolGuy: Order of keys is only stable from Python 3.6 onwards: https://docs.python.org/3/library/stdtypes.html#dict

SchoolGuy: But I guess we only have Python 3.6+ anyway, so we should be safe.

vzhestkov: Right, 3.6 is the only version we are using for Uyuni/4.2/4.1 on the master side where this module could be invoked.

agraul: I was actually thinking about whatever code path adds the stuff to the dict (let's image something parses a YAML file and randomly sorts it before adding the contents to `uyuni_roster_config`), but I agree - the worst case is not that bad anyway.

vzhestkov: The source of this dict actually a `__opts__` from the master (the same for salt-api): `__opts__.get("uyuni_roster")`
It's defined in the master's config.

PR #5226, opened 2022-04-19
QE: Increase timeout to apply highstate after parametrize a formula

srbarrios:
## What does this PR change?

Increase timeout to apply highstate after parametrize a formula.
it actually only happen in Head CI, in Uyuni CI we are good, it could be the hardware below it.

image

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.1
- Manager-4.2

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_salt_formulas.feature
+++ b/testsuite/features/secondary/min_salt_formulas.feature
@@ -59,7 +59,7 @@ Feature: Use salt formulas
      When I click on the "disabled" toggler
      And I click on "Apply Highstate"
      Then I should see a "Applying the highstate has been scheduled." text
-     And I wait until event "Apply highstate in test-mode scheduled by admin" is completed
+     And I wait at most 300 seconds until event "Apply highstate in test-mode scheduled" is completed
 
   Scenario: Apply the parametrized formula via the highstate
      And I follow "States" in the content area


testsuite/features/secondary/min_salt_formulas.feature
```diff
@@ -59,7 +59,7 @@ Feature: Use salt formulas
      When I click on the "disabled" toggler
      And I click on "Apply Highstate"
      Then I should see a "Applying the highstate has been scheduled." text
-     And I wait until event "Apply highstate in test-mode scheduled by admin" is completed
+     And I wait at most 300 seconds until event "Apply highstate in test-mode scheduled" is completed
```
vandabarata, line 62: is the `by admin` part no longer necessary or was it removed accidentally?

srbarrios: Actually I did it in purpose, we don't need to always check that the even is triggered by the only user we have :)
I think we can just short steps without it, in cases like this one.

PR #5254, opened 2022-04-22
Remove old workaround

Bischoff:
## What does this PR change?

Remove old workaround for bsc#1196050 - Hardware list refresh was not sufficient to detect a change in interface's IP address. Now the product problem is fixed, and we don't need to restart salt-minion on the proxy anymore.

## Links

Ports:
* 4.1: SUSE/spacewalk#17596
* 4.2: SUSE/spacewalk#17595

## Changelogs

- [x] No changelog needed


srbarrios (approved):
Approved for master. Let's wait for 4.2 until it's unlocked.

PR #5292, opened 2022-04-26
Ensure correct order of precedence with GSON serializers

cbbayburt:
Serializers that serialize classes in the same class hierarchy override each other in the order they are added. To ensure that subclass serializers take precedence, they must be added later than serializers of their parent classes.

This PR reorders serializers according to the hierarchy relationship between their supported classes to obey the mentioned rule.

## Test coverage
- Unit tests were added

## Changelogs

- [x] No changelog needed

## Re-run a test


parlt91 (approved):
LGTM

PR #5322, opened 2022-04-29
HTTP API polishing

Bischoff:
## What does this PR change?

* Can fixed missing `auth` namespace, this PR removes the corresponding workarounds
* Can moved some more methods from POST to GET, this PR accomodates for it
* I forgot to use `channel.software` namespace in a call used for Build Validation, this PR fixes that

**TO BE MERGED ONLY ON 2022-04-30 AS CAN'S MODIFICATIONS ARE NOT YET ON REBUILT PACKAGES**

## Links

Ports:
* 4.1: SUSE/spacewalk#17676
* 4.2: SUSE/spacewalk#17678

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -177,9 +177,9 @@ def get_client_type(name)
 end
 
 def repository_exist?(repo)
-  $api_test.login('admin', 'admin')
-  repo_list = $api_test.channel_software_repository_list_user_repos
-  $api_test.logout
+  $api_test.auth.login('admin', 'admin')
+  repo_list = $api_test.channel.software.list_user_repos
+  $api_test.auth.logout
   repo_list.include? repo
 end
 
--- a/testsuite/features/support/http_client.rb
+++ b/testsuite/features/support/http_client.rb
@@ -13,7 +13,9 @@ class HttpClient
   def prepare_call(name, params)
     short_name = name.split('.')[-1]
     call_type =
-      if short_name.start_with?('list', 'get', 'is', 'find') || ['logout', 'errata.applicableToChannels'].include?(name)
+      if short_name.start_with?('list', 'get', 'is', 'find') ||
+         name.start_with?('system.search.', 'packages.search.') ||
+         ['auth.logout', 'errata.applicableToChannels'].include?(name)
         'GET'
       else
         'POST'
@@ -32,8 +34,6 @@ class HttpClient
   end
 
   def call(name, params)
-    name.sub!('auth.', '') if ['auth.login', 'auth.logout'].include?(name)
-
     # Get session cookie from previous calls
     if params.nil?
       session_cookie = nil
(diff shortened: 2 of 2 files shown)


nodeg:
If this is needed for Uyuni 2022.04 you have to merge this into the `2022.04` branch and not into `master`.

Bischoff:
> If this is needed for Uyuni 2022.04 you have to merge this into the `2022.04` branch and not into `master` but I guess Can made his changed to master, too so yeah just a heads up.

exactly

the other bug affects both branches, but i don't think oscar tests in 2022.04 branch.

testsuite/features/support/http_client.rb
```diff
@@ -13,7 +13,7 @@ def initialize(host)
   def prepare_call(name, params)
     short_name = name.split('.')[-1]
     call_type =
-      if short_name.start_with?('list', 'get', 'is', 'find') || ['logout', 'errata.applicableToChannels'].include?(name)
+      if short_name.start_with?('list', 'get', 'is', 'find') || ['auth.logout', 'errata.applicableToChannels'].include?(name)
```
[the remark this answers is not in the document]
Bischoff: `start_with?` is too permissive, let's keep the logic the same as in java code

testsuite/features/support/commonlib.rb
```diff
@@ -178,7 +178,7 @@ def get_client_type(name)
 
 def repository_exist?(repo)
   $api_test.login('admin', 'admin')
-  repo_list = $api_test.channel_software_repository_list_user_repos
+  repo_list = $api_test.channel.software.list_user_repos
   $api_test.logout
```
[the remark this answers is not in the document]
Bischoff: oops! fixed, thanks

testsuite/features/support/http_client.rb
```diff
@@ -13,7 +13,9 @@ def initialize(host)
   def prepare_call(name, params)
     short_name = name.split('.')[-1]
     call_type =
-      if short_name.start_with?('list', 'get', 'is', 'find') || ['logout', 'errata.applicableToChannels'].include?(name)
+      if short_name.start_with?('list', 'get', 'is', 'find') ||
+         name.start_with?('system.search.', 'packages.search.') ||
+         ['auth.logout', 'errata.applicableToChannels'].include?(name)
```
srbarrios, line 18: I would prefer a unique list in include to treat all those:


```suggestion
         api_methods = ['auth.logout', 'errata.applicableToChannels', 'system.search.', 'packages.search.']
         api_methods.include?(name)
```

srbarrios (approved):

testsuite/features/support/http_client.rb
Bischoff: won't work, we need `system.search.hostname`, `system.search.ip`, etc. `include?` does an exact match.

cbbayburt: Good idea. We already have an endpoint to list all the namespaces and methods. We should add the GET/POST info to it.

Then if you want to invest, you can request the list at the beginning of each run. That way you won't have to update this list when something changes.

cbbayburt: Check out the methods in the 'api' namespace

PR #5335, opened 2022-05-01
Minor Spec updates

sbluhm:
## What does this PR change?

Minor changes and cleanup to spec files to fix dependencies or simplify the file.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/spacewalk-java.changes
+++ b/java/spacewalk-java.changes
@@ -1,3 +1,4 @@
+- Minor Spec file cleanup.
 - Add script examples for HTTP API
 - Refactor API docs for HTTP API
 - Branding updates
--- a/java/spacewalk-java.spec
+++ b/java/spacewalk-java.spec
@@ -106,13 +106,12 @@ BuildRequires:  httpcomponents-client
 BuildRequires:  ical4j
 BuildRequires:  jade4j
 BuildRequires:  (gnu-jaf or jakarta-activation)
-%if 0%{?rhel} >= 8 || 0%{?fedora}
+%if 0%{?rhel} || 0%{?fedora}
 BuildRequires:  glassfish-jaxb-core
 BuildRequires:  glassfish-jaxb-runtime
 BuildRequires:  glassfish-jaxb-txw2
 BuildRequires:  istack-commons-runtime
 BuildRequires:  java-11-openjdk-devel
-BuildRequires:  maven-javadoc-plugin
 BuildRequires:  (glassfish-jaxb-api or jaxb-api)
 %else
 BuildRequires:  java-devel >= %{java_version}
@@ -190,7 +189,7 @@ Requires:       glassfish-jaxb-runtime
 Requires:       glassfish-jaxb-txw2
 Requires:       istack-commons-runtime
 Requires:       (glassfish-jaxb-api or jaxb-api)
-%if 0%{?rhel} == 8
+%if 0%{?rhel}
 Recommends:       rng-tools
 %endif
 %endif
@@ -202,7 +201,7 @@ Requires:       hibernate5
 Requires:       httpcomponents-client
 Requires:       ical4j
 Requires:       jade4j
-%if 0%{?rhel} >= 8
+%if 0%{?rhel}
 Requires:       java-11-openjdk
 %else
 Requires:       java >= %{java_version}
@@ -372,7 +371,7 @@ Requires:       hibernate-commons-annotations
 Requires:       hibernate5
 Requires:       httpcomponents-client
 Requires:       httpcomponents-core
-%if 0%{?rhel} >= 8
+%if 0%{?rhel}
 Requires:       java-11-openjdk
 %else
 Requires:       java >= %{java_version}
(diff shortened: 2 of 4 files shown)


web/spacewalk-web.changes
```diff
@@ -1,3 +1,4 @@
+- Detailled build requirements.
```
juliogonzalez, line 1: Nor relevant for customers. And TBH I wonder if the previous entry is relevant either.

java/spacewalk-java.changes
```diff
@@ -1,3 +1,4 @@
+- Minor Spec file cleanup.
```
juliogonzalez, line 1: Not relevant for customers. It can go away.

java/spacewalk-java.spec
```diff
@@ -106,13 +106,12 @@ BuildRequires:  httpcomponents-client
 BuildRequires:  ical4j
 BuildRequires:  jade4j
 BuildRequires:  (gnu-jaf or jakarta-activation)
-%if 0%{?rhel} >= 8 || 0%{?fedora}
+%if 0%{?rhel} || 0%{?fedora}
```
juliogonzalez, line 109: Why are we removing `>= 8`?

This means someone could think this can be built for CentOS7.

java/spacewalk-java.spec
```diff
@@ -190,7 +189,7 @@ Requires:       glassfish-jaxb-runtime
 Requires:       glassfish-jaxb-txw2
 Requires:       istack-commons-runtime
 Requires:       (glassfish-jaxb-api or jaxb-api)
-%if 0%{?rhel} == 8
```
juliogonzalez, line 193: Same as before. Truth to be told I wonder why this is `== 8` and not `>= 8`. Maybe someone (@sbluhm) noticed that this is already broken for EL9?

juliogonzalez (changes requested):

java/spacewalk-java.spec
sbluhm: @juliogonzalez EL7 is not supported so I removed it. This is server side only. I would see following sensible combinations (unless there are specific edge cases)
- `0%{?rhel}` - all supported versions.
- `0%{?rhel} == 8` - anything 8 specific.
- `0%{?rhel} >= 9` - anything 9 specific.

If we want to keep 7 in there then I happily revert. My thoughts are that the logic might get too clustered in future.

java/spacewalk-java.spec
sbluhm: :) I don't know why I limited it to 8 in the first place. Am wondering whether I thought that there is a kernel fix. Either way, the package is available on 9 and optional. So in best case it solves problems before they show up.

juliogonzalez: Clearly I didn't notice this came from you @sbluhm :-D

In that case, I assume you already kept anything in mind, so removing `>= 8` should be ok, but I guess this one is not, unless you can check that it works with `9`  as well.

Changelog entries should go away, in this case. They are not relevant for customers, and we already have others that will allow me to tag :-)

sbluhm: :) It works on 9

web/spacewalk-web.spec
```diff
@@ -46,6 +46,7 @@ Source1:        node-modules.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildArch:      noarch
 Requires(pre):  uyuni-base-common
+BuildRequires:  gettext
```
juliogonzalez, line 49: Did you verify that this works with openSUSE?

web/spacewalk-web.spec
```diff
@@ -55,6 +56,8 @@ BuildRequires:  apache2
 BuildRequires:  nodejs-default
 %else
 BuildRequires:  nodejs
+BuildRequires:  perl-srpm-macros
```
juliogonzalez, line 59: Same for this and the next package

juliogonzalez (commented):
Just a couple of questions about the new buildrequires.

web/spacewalk-web.spec
sbluhm: I did via copr: https://copr.fedorainfracloud.org/coprs/sbluhm/uyuni-server/build/4351686/
`Gettext` should be required by both as the binary gets called.. It seems that OBS is installing more additional packages than COPR.

juliogonzalez (approved):
LGTM, but I want an approval from @mcalmer as well... just in case.

web/spacewalk-web.changes
```diff
@@ -1,4 +1,3 @@
-- Removed invalid '\r' character from Roboto*/Icon.
```
mcalmer, line 1: This was maybe a bit too much? Not sure if we should remove changelog entries which are already in?

mcalmer: On the other hand it is also irrelevant for the end user. @juliogonzalez ?

mcalmer (approved):
Only this one question about the removed changelog entry. If this is ok, we can merge.

web/spacewalk-web.changes
juliogonzalez: It was not tagged/released, so I am OK with removing it :-)

PR #5363, opened 2022-05-05
Slim down the images

avshiliaev:
## What does this PR change?

**add description**

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes # https://github.com/SUSE/spacewalk/issues/17519

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/containers/proxy-httpd-image/Dockerfile
+++ b/containers/proxy-httpd-image/Dockerfile
@@ -1,18 +1,40 @@
 #!BuildTag: uyuni/proxy-httpd:latest uyuni/proxy-httpd:%PKG_VERSION% uyuni/proxy-httpd:%PKG_VERSION%.%RELEASE% uyuni/proxy-httpd:beta
 
 ARG BASE=registry.suse.com/bci/bci-base:15.3
-FROM $BASE
-
+FROM $BASE AS base
 
 ARG PRODUCT_REPO
+
+# Add distro and product repos
+COPY add_repos.sh /usr/bin
+RUN sh add_repos.sh ${PRODUCT_REPO}
+
+# Main packages
+RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
+ python3-rhnlib spacewalk-proxy-broker \
+ spacewalk-proxy-common spacewalk-proxy-package-manager \
+ spacewalk-proxy-redirect spacewalk-ssl-cert-check spacewalk-proxy-html \
+ susemanager-tftpsync-recv
+
+# Additional material
+COPY uyuni-configure.py /usr/bin/uyuni-configure.py
+RUN chmod +x /usr/bin/uyuni-configure.py
+
+# Prepare for slimming down
+COPY prepare_target.sh .
+RUN sh prepare_target.sh
+
+# Define slim image
+ARG BASE=registry.suse.com/bci/bci-base:15.3
+FROM $BASE AS slim
+
 ARG PRODUCT=Uyuni
 ARG VENDOR="Uyuni project"
 ARG URL="https://www.uyuni-project.org/"
 ARG REFERENCE_PREFIX="registry.opensuse.org/uyuni"
 
-# Add distro and product repos
-COPY add_repos.sh /usr/bin
-RUN sh add_repos.sh ${PRODUCT_REPO}
+RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses python3-PyYAML
+COPY --from=base /target /
 
 # Build Service required labels
 # labelprefix=org.opensuse.uyuni.proxy-httpd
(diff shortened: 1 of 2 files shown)


containers/proxy-httpd-image/Dockerfile
```diff
@@ -1,18 +1,41 @@
...
+FROM $BASE AS base
 
 ARG PRODUCT_REPO
+
+# Add distro and product repos
+COPY add_repos.sh /usr/bin
+RUN sh add_repos.sh ${PRODUCT_REPO}
+
+# Main packages
+RUN echo "rpm.install.excludedocs = yes" >>/etc/zypp/zypp.conf
```
juliogonzalez, line 13: It's interesting you need to add this. The BCI image should already configure this by default. Otherwise it means a user of the bci-image would install the doc if he adds packages.

Maybe you can report this to the BCI maintainers?

containers/proxy-httpd-image/prepare_target.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# Removes any unnecessary files and packages before moving to the next build stage
+
+set -xe
+
+# remove rpm-build and its dependencies
+rpm -e rpm-build --nodeps
+rpm -e dwz make gcc patch diffutils python-rpm-macros systemd-rpm-macros glibc-locale
```
juliogonzalez, line 8: When are such packages getting installed? During the zypper call at https://github.com/uyuni-project/uyuni/pull/5363/files#diff-c81a9cc5087e00f7ef306e86a8aaaca03bcfbf348a9f5d1b0aefe11dd58fcc6fR14?

If so, you could consider calling this script at that same call `with && sh prepare_target.sh`

juliogonzalez (commented):
Some comments for the first image, that maybe are relevant for the other images as well.

containers/proxy-httpd-image/Dockerfile
avshiliaev: Thank you for your feedback :+1: Let me double check if it's something we really need to do. We've just brainstormed different strategies and I don't remember if it's really a problem

containers/proxy-httpd-image/prepare_target.sh
avshiliaev: The `spacewalk-certs-tools` is responsible for installing it. Again I believe it's a necessary dependency of the package, or is it @cbosdo ?

regarding merging the calls, me personally, I would rather keep them separate since we use multistage build anyway and merging calls wouldn't bring us any improvement in size. At least for me it's more readable, but if we have some other conventions for such things, I'm totally open for it :smile:

juliogonzalez: > regarding merging the calls, me personally, I would rather keep them separate since we use multistage build anyway and merging calls wouldn't bring us any improvement in size. At least for me it's more readable, but if we have some other conventions for such things, I'm totally open for it smile

Well, the benefit is that you won't ever had such packages on any of the stages, if you don't need them for the final image.

If right now there are no size benefits that's probably because a later stage is bigger than the stage before prepare_target.sh is called.

containers/proxy-httpd-image/Dockerfile
avshiliaev: Well yes 7.5M in `/usr/share/doc/packages` if called without `RUN echo "rpm.install.excludedocs = yes" >>/etc/zypp/zypp.conf`

juliogonzalez: Yup, I am not saying you should remove this. Keep it in the PR.

But I think the BCI image should have this configured by default? I guess nobody wants doc by default when you install packages on containers? Can you check with them?

JeOS has the installation of doc disable, so I guess it should be the same for the BCI containers.

containers/proxy-httpd-image/prepare_target.sh
avshiliaev: You're totally right. 
If I understand it correctly, at least starting from docker >17.05 we got two mutually excluding options. 
1. we can either merge together the _add_ and _remove_ operations so that we don't create additional layers/commits in between and thus saving us space. 
2. or as an alternative we can use multistage builds where we copy over only the files we need from the head of a previous stage, not the layers/commits as such. 
Both options lead to the same image size, I've just tested :smiley:

avshiliaev: But surely I don't insist on any particular approach. I just decided to go for the best practices as they're declared out there.

avshiliaev: Ok I see the real benefit. We don't overuse the cache while building the image if we merge together the calls. So I'll merge it.

containers/proxy-httpd-image/Dockerfile
avshiliaev: I'll check with them, thank you!

avshiliaev:
Merged into helm-charts

PR #5362, opened 2022-05-05
Testsuite: fixes for retail test

nadvornik:
## What does this PR change?

This fixes the setup of retail environment.

Before, the image-sync was called before the retail groups were created. As a result, the bug https://bugzilla.suse.com/show_bug.cgi?id=1199157 was not detected. This PR moves  image-sync to the right place.

Also, it adds the required restart after installation of saltboot formula and assigns the branch server to branch group.

The second commit increases the timeout for saltboot state to 600s. It took 400s at 
https://ci.suse.de/view/Manager/view/Manager-Head/job/manager-Head-dev-acceptance-tests-NUE/3237/

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation
- No documentation needed: **add explanation. This can't be used if there is a GUI diff**
- No documentation needed: only internal and user invisible changes
- Documentation issue was created: [Link for SUSE Manager contributors](https://github.com/SUSE/spacewalk/issues/new?template=ISSUE_TEMPLATE_DOCUMENTATION.md&labels=documentation&projects=SUSE/spacewalk/31), [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/new).
- (OPTIONAL) [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pulls)

- [ ] **DONE**

## Test coverage
- No tests: **add explanation**
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #
Tracks # **add downstream PR, if any**

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #5381, opened 2022-05-06
Testsuite: Test reboot of retail terminal

nadvornik:
## What does this PR change?

Add a test of retail terminal reboot.
This verifies that the terminal configuration (PXE files, salt key, additionally installed packages) is stored permanently.
The test takes 104s on my machine.

It requires the fixes from https://github.com/uyuni-project/uyuni/pull/5362
I created separate PR as this might be unstable at the beginning.

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
+++ b/testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
@@ -220,6 +220,14 @@ Feature: PXE boot a Retail terminal
     Then I should see a "1 package install has been scheduled" text
     When I wait until event "Package Install/Upgrade scheduled by admin" is completed
 
+  Scenario: Check that the PXE boot minion stays configured after reboot
+    # salt key, channels and installed packages must survive reboot
+    When I reboot the Retail terminal "pxeboot_minion"
+    And I follow the left menu "Systems > System List > All"
+    And I follow this "pxeboot_minion" link
+    And I wait at most 350 seconds until event "Apply states [util.syncstates, saltboot] scheduled by (none)" is completed
+    Then the Salt master can reach "pxeboot_minion"
+
   Scenario: Cleanup: remove a package on the new Retail terminal
     Given I navigate to the Systems overview page of this "pxeboot_minion"
     When I follow "Software" in the content area
--- a/testsuite/features/step_definitions/salt_steps.rb
+++ b/testsuite/features/step_definitions/salt_steps.rb
@@ -11,7 +11,7 @@ Given(/^the Salt master can reach "(.*?)"$/) do |minion|
   # 300 is the default 1st keepalive interval for the minion
   # where it realizes the connection is stuck
   repeat_until_timeout(timeout: 300, retries: 3, message: "Master can not communicate with #{minion}", report_result: true) do
-    out, _code = $server.run("salt #{system_name} test.ping")
+    out, _code = $server.run("salt #{system_name} test.ping", check_errors: false, timeout: 90)
     if out.include?(system_name) && out.include?('True')
       finished = Time.now
       log "Took #{finished.to_i - start.to_i} seconds to contact the minion"


testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
```diff
@@ -215,6 +215,19 @@ Feature: PXE boot a Retail terminal
     Then I should see a "1 package install has been scheduled" text
     When I wait until event "Package Install/Upgrade scheduled by admin" is completed
 
+  Scenario: Chect that the PXE boot minion stays configured after reboot
+    # salt key, channels and installed packages must survive reboot
+    When I reboot the PXE boot minion
+    And I wait for "60" seconds
```
srbarrios, line 221: I don't like this, I would prefer to wait until some event happens, for instance waiting until the salt-minion service inside that VM is active, or something similar.

Example: 
`I wait until "salt-minion" service is active on "pxeboot_minion"`

srbarrios: I guess you can refactor the step "completed during last minute", in order that in does not only check it's completed during last minute, but also checks it's the first row of the table. 🤔

srbarrios: What about...
```
    And I wait until I see "Apply states [util.syncstates, saltboot]" text, refreshing the page
    And I follow first "Apply states [util.syncstates, saltboot]"
    And I wait at most 60 seconds until the event is completed, refreshing the page
```

Bischoff: I am not a 100 % sure we are safe against race conditions.

For example, maybe the event is not in "Pending" yet, and you would go directly to "History". And then the first event is the old one.

Just saying. Maybe it's safe enough for our needs. We would need to try.

Bischoff (commented):
Changes from Oscar would be worth trying.

Other than that, we really need to split this feature again into two separate features and work out correctly the idempotency of each half. It's just getting too fat. But not now, after 4.3.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
srbarrios: If it fails, I come back to my first suggestion:
```
When I reboot the PXE boot minion
And I wait until "salt-minion" service is active on "pxeboot_minion"
Given I am on the Systems overview page of this "pxeboot_minion"
And I follow "Software Channels" in the content area
Then radio button "Test-Channel-x86_64" is checked
```

nodeg:
> Other than that, we really need to split this feature again into two separate features and work out correctly the idempotency of each half. It's just getting too fat. But not now, after 4.3.

Can we document those things somewhere to have them ready for discussion in our tech-dept meeting? Since we have more of those candidates it would be nice to have an overview to work on them after the 4.3 release.

nadvornik:
I have added code for filtering-out old events based on action ID.

Also I fixed use of test.ping.  Before, whole test failed immediately after the first failed test.ping. Now it fails after 3 failed attempts.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
```diff
@@ -215,6 +215,14 @@ Feature: PXE boot a Retail terminal
     Then I should see a "1 package install has been scheduled" text
     When I wait until event "Package Install/Upgrade scheduled by admin" is completed
 
+  Scenario: Chect that the PXE boot minion stays configured after reboot
```
Bischoff, line 218: `Chect that the PXE boot minion stays configured after reboot`
 =>
`The PXE boot minion stays configured after reboot`

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -85,23 +85,33 @@
   end
 end
 
+$max_completed_aid = 0
```
Bischoff, line 88: I don't like this state machine approach, with a global variable. It's complicated and I have a feeling it could be fragile.

Why can't we simply follow @srbarrios 's suggestion? If a race really occurs, then we can reconsider.

testsuite/features/step_definitions/salt_steps.rb
```diff
@@ -11,7 +11,7 @@
   # 300 is the default 1st keepalive interval for the minion
   # where it realizes the connection is stuck
   repeat_until_timeout(timeout: 300, retries: 3, message: "Master can not communicate with #{minion}", report_result: true) do
-    out, _code = $server.run("salt #{system_name} test.ping")
+    out, _code = $server.run("salt #{system_name} test.ping", check_errors: false, timeout: 90)
```
Bischoff, line 14: when one increases a timeout, one should always ask himself/herself the question:

"would a customer find this timeout acceptable?"

Is 90 seconds something acceptable, or rather the sign of a problem of our product?

Bischoff (changes requested):
see change requests

testsuite/features/step_definitions/salt_steps.rb
srbarrios: A ping can't definitely not take 90 seconds.
Instead we should just have a 2-3 seconds timeout, and re-try the ping X seconds until it replies.

testsuite/features/step_definitions/navigation_steps.rb
srbarrios: I think that handling the events using a cache of processed Ids could be a good approach, I would not use a integer, but a list of integers for all processed events, as otherwise that might conflict if we try to parallelize our tests.

**BUT**, we can't afford this refactor now, now we only can fix stuff in the simple way we can or keep it failing until GM.
I don't want to risk breaking other scenarios as this refactor don't only affects the test to reboot a terminal, but will affect any other scenario checking for events.

Said that, if we find an easier approach now, even if it could have some corner case, I would prefer it, and then we can keep this PR open to apply this refactor after well discuss it other squad members. But in overall, I like the idea of handling our events by ID, as any other ways we always had issues.

testsuite/features/step_definitions/salt_steps.rb
nadvornik: When the minion is down, then the ping can take up to 300s as noted in the comment above.
Timeout 90 ensures that it is tried 3 times within this 300s.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
```diff
@@ -215,6 +215,14 @@ Feature: PXE boot a Retail terminal
...
+    # salt key, channels and installed packages must survive reboot
+    When I reboot the PXE boot minion
+    And I follow the left menu "Systems > Overview"
+    And I follow this "pxeboot_minion" link
+    And I wait at most 350 seconds until event "Apply states [util.syncstates, saltboot] scheduled by (none)" is completed
```
srbarrios, line 223:
```suggestion
When I reboot the PXE boot minion
And I wait until "salt-minion" service is active on "pxeboot_minion"
```

srbarrios: If you need, refactor this step to provide a bigger timeout, and that should be good enough. For now.

nadvornik:
With https://github.com/uyuni-project/uyuni/pull/6191 the event for saltboot reboot differs from initial boot. This means that handling of repeated events is not needed and this PR could be simplified.

aaannz (approved):

testsuite/features/step_definitions/salt_steps.rb
srbarrios: Let's try it!

srbarrios (approved):

Bischoff (approved):

PR #5383, opened 2022-05-06
WIP: Investigate better test id integration

Etheryte:
## What does this PR change?

With this change, React dropdowns should be openable with whatever is XPath's equivalent of `document.querySelector('[data-testid="selectedBaseChannel"]').focus()`. If you want to test this out in your local browser, wrap it in a timeout and restore focus to the main browser area before the timeout fires, the browser won't give focus away if the focus is in the devtools instead.  

This change most definitely breaks whatever logic we had bound to the class names since I removed it, so I'm expecting this to break other tests too. I'm not sure if it's completely fine to remove the class name logic. Do we still use it to a-la select the correct item from the dropdown or something similar? I'm not sure if there's edge cases I haven't thought of or other things I've missed, so would appreciate your input Can.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: trying to fix existing tests

- [x] **DONE**

## Links

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/components/combobox.tsx
+++ b/web/html/src/components/combobox.tsx
@@ -80,7 +80,7 @@ export class Combobox extends React.Component<ComboboxProps, ComboboxState> {
       label: item.text,
     }));
 
-    const testAttributes = withTestAttributes(this.props["data-testid"], this.props.name);
+    const testAttributes = withTestAttributes(this.props["data-testid"]);
     return (
       <Creatable
         id={this.props.id}
--- a/web/html/src/components/input/Select.tsx
+++ b/web/html/src/components/input/Select.tsx
@@ -192,10 +192,12 @@ export function Select(props: Props) {
             isClearable: isClearable,
             styles: bootstrapStyles,
             isMulti: props.isMulti,
+            // NB! This is mandatory for Cucumber integration to work
+            openMenuOnFocus: true,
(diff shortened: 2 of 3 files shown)


cbbayburt (commented):
> If you want to test this out in your local browser, wrap it in a timeout and restore focus to the main browser area before the timeout fires, the browser won't give focus away if the focus is in the devtools instead.

My favorite hack :)

> This change most definitely breaks whatever logic we had bound to the class names since I removed it, so I'm expecting this to break other tests too. I'm not sure if it's completely fine to remove the class name logic. Do we still use it to a-la select the correct item from the dropdown or something similar? I'm not sure if there's edge cases I haven't thought of or other things I've missed, so would appreciate your input Can.

Looking at the whole testsuite code, we use the class selector "data-testid-*" for the following elements that belong to React select:

- `data-testid-<fieldname>-child__control`
- `data-testid-<fieldname>-child__option`
- `data-testid-<fieldname>-child__value-container`

With your proposed change, does the element with the `data-testid` attribute enclose all these children? If so, it should be quite easy to select them all with XPath, starting from the parent's `data-testid` attribute.

Etheryte:
This is obsolete with the other updates we made for both master and 4.2.

Issue #5370, opened 2022-05-05
Uyuni 2022.04 -- ssh-push bootstrap no longer working through Uyuni Proxy


benled:
# Problem description
All clients are managed via ssh-push behind a Uyuni-Proxy.
Since the Upgrade to 2022.03 (and today to 2022.04)  it is no longer possible to bootstrap new ssh-push manged clients.
When bootstrapping via WI the SSH-Keys, etc. are pushed to the new client, but the system never appear.
In rhn_web_ui.log the following messages are recorded:

```
2022-05-05 15:03:36,674 [ajp-nio-127.0.0.1-8009-exec-2] ERROR com.suse.manager.reactor.messaging.RegisterMinionEventMessageAction - Aborting: needed grains are not found for minion: newhost.local.intra
2022-05-05 15:08:13,400 [ajp-nio-127.0.0.1-8009-exec-3] WARN  com.suse.manager.webui.services.impl.SaltService - SaltSSHError(255, ssh: connect to host newhost.local.intra port 22: Connection timed out)
```

When watching the processes it is obvious that at one point the Uyuni attempts a direct ssh connection instead of going through the UyuniProxy:
```

salt       2113  0.0  0.1 278000 58988 ?        Ss   11:32   0:01 /usr/bin/python3 /usr/bin/salt-api
salt       2225  2.9  0.5 4281532 241004 ?      Sl   11:32   9:43  \_ /usr/bin/python3 /usr/bin/salt-api
salt      70978  6.7  0.5 4281532 234504 ?      S    16:57   0:00      \_ /usr/bin/python3 /usr/bin/salt-api
[...]

PR #5384, opened 2022-05-07
create a roster for pending minions not yet in DB

mcalmer:
## What does this PR change?

When bootstrapping a ssh minion via a proxy and the server cannot reach the minion directly, we must use
the proxy as jump host also when getting the grains via salt-ssh after the first bootstrap call but before the client is
created in the database.
In this case we still must create a roster file as the autogenerated roster from DB will not yet find the minion.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Cucumber test added

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/5370
Fixes https://github.com/SUSE/spacewalk/issues/17762

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
+++ b/java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
@@ -212,8 +212,13 @@ public class SaltSSHService {
      */
     public <R> Map<String, Result<R>> callSyncSSH(LocalCall<R> call, MinionList target, Optional<String> extraFileRefs)
             throws SaltException {
+        // Using custom LocalCall timeout if included in the payload
+        Optional<Integer> sshTimeout = call.getPayload().containsKey("timeout") ?
+                Optional.ofNullable((Integer) call.getPayload().get("timeout")) :
+                    getSshPushTimeout();
+        Optional<SaltRoster> roster = prepareSaltRoster(target, sshTimeout);
         return unwrapSSHReturn(
-                callSyncSSHInternal(call, target, Optional.empty(), false, isSudoUser(getSSHUser()), extraFileRefs));
+                callSyncSSHInternal(call, target, roster, false, isSudoUser(getSSHUser()), extraFileRefs));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


testsuite/features/secondary/minssh_bootstrap_api_proxy.feature
```diff
@@ -0,0 +1,62 @@
...
+    Then I should see a "Push via SSH" text
+
+  Scenario: Check registration on proxy of SSH minion bootstrapped via API
+    Given I am on the Systems overview page of this "proxy"
+    When I follow "Details" in the content area
+    And I follow "Proxy" in the content area
+    Then I should see "ssh_minion" hostname
+
+  Scenario: Check spacecmd system ID of SSH minion bootstrapped via API
+    Given I am on the Systems overview page of this "ssh_minion"
```
srbarrios, line 43: Unnecessary Given step.

testsuite/features/secondary/minssh_bootstrap_api_proxy.feature
```diff
@@ -0,0 +1,62 @@
...
+    When I follow "Delete System"
+    Then I should see a "Confirm System Profile Deletion" text
+    When I click on "Delete Profile"
+    And I wait until I see "has been deleted" text
+    Then "ssh_minion" should not be registered
+
+  Scenario: Bootstrap a SLES SSH minion via API
+    Given I am logged in API as user "admin" and password "admin"
+    And I block connections from "server" on "ssh_minion"
```
srbarrios, line 23: Personally, the block and flush of connections I would put it in a different scenario, to highlight it, as it's kind of important to explain that we don't have connection to the server, as part of the test scenario.

_Also, it might be interesting a description of the acceptance criteria just below the Feature title_

mcalmer: As discussed: I have integrated this into the existing feature test with extra scenarios which are executed only if a proxy is part of the test

testsuite/features/secondary/minssh_bootstrap_api_proxy.feature
mcalmer: feature dropped and integrated into an existing one

mbussolotto (approved):

srbarrios (approved):

renner (approved):
It makes sense, sorry, I might have removed a bit too much code with my initial PR.

PR #5386, opened 2022-05-07
Fix CLM sources modal

Bischoff:
## What does this PR change?

When we select a channel in the CLM "add sources" dialog box, a summary of the channel and its child channels appears, where the channel is selected via a checkbox.

If this summary has not the time to appear and we press the "Save" button too early, then the channel will not be selected in the sources.

This PR waits for the summary to appear, by waiting for the selected channel name to appear.

There are also cosmetical changes (in a separate commit).

Note: an unresolved problem in 4.2 is that the drop-down field is callled `undefined`. This should probably be fixed at product level.

## Links

Ports:
* 4.1: SUSE/spacewalk#17767
* 4.2: SUSE/spacewalk#17766

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/secondary/srv_content_lifecycle.feature
+++ b/testsuite/features/secondary/srv_content_lifecycle.feature
@@ -19,7 +19,7 @@ Feature: Content lifecycle
     And I enter "clp_name" as "name"
     And I enter "clp_desc" as "description"
     And I click on "Create"
-    Then I wait until I see "Content Lifecycle Project - clp_name" text
+    And I wait until I see "Content Lifecycle Project - clp_name" text
 
   Scenario: Verify the content lifecycle project page
     When I follow the left menu "Content Lifecycle > Projects"
@@ -38,21 +38,21 @@ Feature: Content lifecycle
     And I click on "Attach/Detach Sources"
     And I select "SLES12-SP5-Pool for x86_64" from "selectedBaseChannel"
     And I click on "Save"
-    Then I wait until I see "SLES12-SP5-Pool for x86_64" text
-    And I should see a "Version 1: (draft - not built) - Check the changes below" text
+    And I wait until I see "SLES12-SP5-Pool for x86_64" text
+    Then I should see a "Version 1: (draft - not built) - Check the changes below" text
 
 @uyuni
   Scenario: Verify added sources for Uyuni
     When I follow the left menu "Content Lifecycle > Projects"
     And I follow "clp_name"
-    And I should see a "SLES12-SP5-Updates for x86_64" text
+    Then I should see a "SLES12-SP5-Updates for x86_64" text
     And I should see a "Build (2)" text
 
 @susemanager
   Scenario: Verify added sources for SUSE Manager
     When I follow the left menu "Content Lifecycle > Projects"
     And I follow "clp_name"
-    And I should see a "SLE-Manager-Tools12-Updates for x86_64 SP5" text
+    Then I should see a "SLE-Manager-Tools12-Updates for x86_64 SP5" text
     And I should see a "SLES12-SP5-Updates for x86_64" text
     And I should see a "SLE-Manager-Tools12-Pool for x86_64 SP5" text
     And I should see a "Build (4)" text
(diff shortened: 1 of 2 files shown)


testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -172,12 +172,12 @@
...
-  xpath_field = "//*[contains(@class, 'data-testid-#{field}-child__control')]"
   if has_select?(field, with_options: [option], wait: 1)
     select(option, from: field)
   else
     # Custom React selector
+    xpath_field = "//*[contains(@class, 'data-testid-#{field}-child__control')]"
+    xpath_option = ".//*[contains(@class, 'data-testid-#{field}-child__option') and contains(text(), '#{option}')]"
     find(:xpath, xpath_field).click
     find(:xpath, xpath_option, match: :first).click
   end
```
nodeg, line 183: I wonder why we did not do it this way in the first place.

nodeg (approved):
LGTM.

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: good qustion :smile_cat:

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -172,12 +172,12 @@
...
-  xpath_option = ".//*[contains(@class, 'data-testid-#{field}-child__option') and contains(text(),'#{option}')]"
-  xpath_field = "//*[contains(@class, 'data-testid-#{field}-child__control')]"
   if has_select?(field, with_options: [option], wait: 1)
     select(option, from: field)
   else
     # Custom React selector
+    xpath_field = "//*[contains(@class, 'data-testid-#{field}-child__control')]"
+    xpath_option = ".//*[contains(@class, 'data-testid-#{field}-child__option') and contains(text(), '#{option}')]"
     find(:xpath, xpath_field).click
     find(:xpath, xpath_option, match: :first).click
```
srbarrios, line 182:
```suggestion
    xpath_field = "//*[contains(@class, 'data-testid-#{field}-child__control')]"
    xpath_option = ".//*[contains(text(), '#{option}')]"
    find(:xpath, xpath_field).click
    first(:xpath, xpath_option).click
```

testsuite/features/secondary/srv_content_lifecycle.feature
```diff
@@ -37,22 +37,23 @@ Feature: Content lifecycle
     And I follow "clp_name"
     And I click on "Attach/Detach Sources"
     And I select "SLES12-SP5-Pool for x86_64" from "selectedBaseChannel"
+    And I wait until I see "SLES12-SP5-Pool for x86_64" text
```
srbarrios, line 40: If we do:
`first(:xpath, ".//*[contains(text(), 'SLES12-SP5-Pool for x86_64')]").click`

In theory, the Capybara `first` method it will wait the default timeout until it found a first web element containing this text.
So, in theory, it will not be a reason to wait again until we see the text. Unless... selecting the option, the text disappear and appear again after few milliseconds.

cbbayburt:
> Note: an unresolved problem in 4.2 is that the drop-down field is callled undefined. This should probably be fixed at product level.

I'll look into this today.

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: That's dangerous, this will not be searched within the field, but in the whole HTML page. Checking for `'data-testid-#{field}-child__option'` guarantees we will match only this field option.

I tried to keep the changes minimal in order to not introduce further problems.

testsuite/features/secondary/srv_content_lifecycle.feature
srbarrios: Yes, but if you get my point, depending the speed of the web, the select before is already looking for that exact text, so if we select it and we wait for the same exact text, depending the speed of the web, it might just go ahead and pass it.

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: It's even worse then. Same answer applies.

srbarrios: Anyway, if it worked, merge it, all right.

Bischoff: Yes, I would have preferred using a `within`, it's good prectice.

However it would not work here: ReactJS does not store the options of the combobox within the combobox, but separately, at the end of the HTML page!

It nearly drove me mad until I understood it. I spent hours trying to understand why my `within` was not working :smiling_imp: .

PR #5395, opened 2022-05-09
Improved handling of bootstrap errors

mackdk:
## What does this PR change?

This PR fixes the error messages displayed when the bootstrap of a new system fails. 

## GUI diff

Before:

https://user-images.githubusercontent.com/2292684/167445978-c2d807aa-d05d-4e6f-981c-db8cae2e6de6.mp4

After:

https://user-images.githubusercontent.com/2292684/167445492-20c5b51e-b454-457d-a041-544a06ad0a64.mp4

- [X] **DONE**

## Documentation
- No documentation needed: error handling is not described in the documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17490

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/src/components/messages.tsx
```diff
@@ -123,29 +123,56 @@ export const fromServerMessage = (
   return msg;
 };
 
-function msg(severityIn: Severity, ...textIn: Array<React.ReactNode>) {
-  return textIn.map(function (txt) {
-    return { severity: severityIn, text: textIn };
-  });
+function msg(severityIn: Severity, textIn: React.ReactNode, listMultiple: boolean, header?: string) {
```
Etheryte, line 126: I'm not sure if the addition of these flags makes sense to me. At the end of the day, the server is providing the error message, so the header etc should come from the server as well. With this setup you tightly couple the error handling logic between the backend and fronted.  

Instead, I think it would make more sense to simply allow the backend to send an error message with DOM if necessary. If we want to avoid multiple separate message boxes, the backend should not send multiple separate messages. What do you think @cbbayburt and @ncounter?

web/html/src/manager/systems/bootstrap/bootstrap-minions.tsx
```diff
@@ -17,6 +19,81 @@ declare global {
...
+
+      buttons = (
+        <div>
+          <Button
+            className="btn-default"
+            text={t("Close")}
+            title={t("Close")}
+            icon="fa-close"
+            handler={() => {
+              jQuery("#show-error-details").modal("hide");
```
Etheryte, line 77: It would be preferable not to use jQuery in any new code. Is there a case the React-based modal logic doesn't cover?

web/html/src/manager/systems/bootstrap/bootstrap-minions.tsx
```diff
@@ -196,28 +287,34 @@ class BootstrapMinions extends React.Component<Props, State> {
...
           });
         } catch (err) {
-          var errMessages = DEPRECATED_unsafeEquals(xhr.status, 0)
-            ? [
-                t(
-                  "Request interrupted or invalid response received from the server. Please check if your minion was bootstrapped correctly."
-                ),
-              ]
-            : [Network.errorMessageByStatus(xhr.status)];
+          var errMessage = DEPRECATED_unsafeEquals(xhr.status, 0)
```
Etheryte, line 306: If you're updating this code, please try and remove the use of `DEPRECATED_unsafeEquals`. It should most likely be `xhr.status === 0 ? ...`, but please check it.

web/html/src/manager/systems/bootstrap/bootstrap-minions.tsx
mackdk: No, no specific reason. This is just coming from the other page I copy-pasted for reference :sweat_smile: . I will fix it.

web/html/src/components/messages.tsx
mackdk: > With this setup you tightly couple the error handling logic between the backend and frontend

I'm not sure to have fully understood what you meant, so please correct me if I got your point wrong.

The backend here has no knowledge of messages, it is just reporting the errors that have happened during the bootstrapping process in the response. 

In my opinion, it should be up to the frontend logic to decide how to show those errors, depending on the context. That's the reason why I added the additional parameters, to have an API to specify how to present multiple errors from the server. In my idea, the header is only for presentation purpose and has no meaning for the backend.

> At the end of the day, the server is providing the error message

Indeed, but ideally that should not be the case, in my opinion. But I don't think we have any kind of common error coding in place.

> I think it would make more sense to simply allow the backend to send an error message with DOM if necessary. If we want to avoid multiple separate message boxes, the backend should not send multiple separate messages

It seems to me that sending DOM or deciding the grouping on the backend side will lead to far more tight coupling, because implementing that would mean knowing in the backend how the UI works and looks like.

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -2032,44 +2036,58 @@ public static String uuidToLittleEndian(String uuidIn) {
...
-                }
+            json = ((JsonParsingError)message).getJson();
+        }
+        else if (message instanceof SaltSSHError) {
+            try {
+                // Try parsing the error message as JSON
+                json = JsonParser.GSON.fromJson("{" + ((SaltSSHError) message).getMessage() + "}", JsonElement.class);
+            }
+            catch (JsonParseException ex) {
+                LOG.debug("Unable to parse SaltSSHError message as json");
```
mcalmer, line 2050: maybe better a warning? Otherwise we have no idea that an error happened when looking at the logs.
Is there a way to raw dump the message? Would it make sense ?

java/code/src/com/suse/manager/webui/controllers/bootstrap/AbstractMinionBootstrapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 SUSE LLC
+ * Copyright (c) 2022 SUSE LLC
```
mcalmer, line 2:
```suggestion
 * Copyright (c) 2016--2022 SUSE LLC
```

java/code/src/com/suse/manager/webui/controllers/bootstrap/RegularMinionBootstrapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 SUSE LLC
+ * Copyright (c) 2022 SUSE LLC
```
mcalmer, line 2:
```suggestion
 * Copyright (c) 2016--2022 SUSE LLC
```

java/code/src/com/suse/manager/webui/controllers/bootstrap/SSHMinionBootstrapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 SUSE LLC
+ * Copyright (c) 2022 SUSE LLC
```
mcalmer, line 2:
```suggestion
 * Copyright (c) 2016--2022 SUSE LLC
```

mcalmer (changes requested):
a few smaller things found

java/code/src/com/suse/manager/utils/SaltUtils.java
mackdk: Indeed. I refactored the code to propagate the actual error message to the frontend even when it is not possible to parse it as json. Since now the error is correctly reported I don't think we need to have this log as a warning, but I upgraded it to info.

web/html/src/manager/systems/bootstrap/bootstrap-minions.tsx
mackdk: I replaced the jQuery based `PopUp` component with the `Dialog`

web/html/src/components/messages.tsx
cbbayburt: Server-side DOM is not an option in this case, because in most places, the same errors are sent to media other than the web UI like API responses, logs, etc.

I think the separation of concerns is good here. It's basically: the server has multiple errors to display, and the frontend decides how to present them, with some formatting and additional explanatory text.

> In my idea, the header is only for presentation purpose and has no meaning for the backend.

I agree. It's not a part of the backend errors. It's mostly cosmetic and should be considered part of the page design.

I specifically wanted to have these flags (or some similar functionality) because the outcome of this PR is very neat, and we might consider using the same logic in more pages where multiple messages are possible (`products.tsx` in this PR, for example). The general problem we have with error reporting is that every page has its own implementation for such cases. With the introduction of reusable stuff like this, we'll move towards a more unified appearance.

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -2029,28 +2028,31 @@ public static String uuidToLittleEndian(String uuidIn) {
...
+            moduleNotSupported -> null,
+            jsonParsingError -> jsonParsingError.getJson(),
+            genericError -> null,
+            saltSshError -> {
+                try {
+                    // Try parsing the error message as JSON
+                    return JsonParser.GSON.fromJson("{" + saltSshError.getMessage() + "}", JsonElement.class);
+                }
+                catch (JsonParseException ex) {
+                    LOG.info("Unable to parse SaltSSHError message \"{}\"as json: {}",
```
cbbayburt, line 2043: I think I agree with @mcalmer here to make this a `WARN` entry.
We should be able to parse all the expected messages. If we cannot parse something, it's something unexpected and it needs to be addressed by improving our error handling.

cbbayburt (approved):
Thanks for the great effort, Thomas! Nice solution.

My only suggestion is about some opportunity to translate some of the messages.

mcalmer (approved):

srbarrios (approved):
Thank you very much for this improvement!

testsuite/features/secondary/min_bootstrap_negative.feature
```diff
@@ -61,9 +70,12 @@ Feature: Negative tests for bootstrapping normal minions
...
      And I click on "Bootstrap"
-     And I wait until I see "ssh: connect to host" text
-     Then I should not see a "GenericSaltError" text
-     And I should see a "port 11: Connection refused" text or "port 11: Invalid argument" text
+     And I wait until I see "An error has occurred during salt execution: unable to establish ssh connection." text
+     Then I should see a "Details" link
+     When I follow "Details" in the content area
+     Then I should see "Error Details" text
+     And I should see "Standard Error" text
+     And I should see "ssh: connect to host uyuni-minion-ubuntu port 11: Connection refused" in the textarea
```
srbarrios, line 78: I would say the name is wrong : `uyuni-minion-ubuntu` should probably be have the hostname of `sle_minion`. I suggest you just check for the text "Connection refused" and that should be enough.

Issue #5241, opened 2022-04-21
Uyuni 2022.02: Debian packages not cached by Squid


phibid:
We have observed that the *.deb packages not seems to be cached at all by the Squid embedded in the Uyuni proxy:

A simple test downloading several times in a raw from a register Ubuntu server a 91KB package size ends up with a TCP_MISS:

```
10.264    242 127.0.0.1 TCP_MISS/200 91356 GET http://proxy_uyuni /rhn/manager/download/ubuntu-1804-amd64-main-security-uyuni/getPackage/gzip_1.6-5ubuntu1.2.amd64-deb.deb - HIER_DIRECT/10.206.29.108 application/octet-stream
1650464072.536    310 127.0.0.1 TCP_MISS/200 91356 GET http://proxy_uyuni /rhn/manager/download/ubuntu-1804-amd64-main-security-uyuni/getPackage/gzip_1.6-5ubuntu1.2.amd64-deb.deb - HIER_DIRECT/10.206.29.108 application/octet-stream
1650464272.667    171 127.0.0.1 TCP_MISS/200 91356 GET http://proxy_uyuni /rhn/manager/download/ubuntu-1804-amd64-main-security-uyuni/getPackage/gzip_1.6-5ubuntu1.2.amd64-deb.deb - HIER_DIRECT/10.206.29.108 application/octet-stream
```

To reproduce this test, just execute this command on a Ubuntu server registered in Uyuni to download only the package:

```
# rm -f /var/cache/apt/archives/gzip_1.6-5ubuntu1.2_amd64.deb && apt-get install --download-only gzip
```

Regarding RPM like OS, packages seems to be correctly cached:

```
access.log:1650178125.754      0 127.0.0.1 TCP_MEM_HIT/200 154067 GET http://proxy_uyuni/rhn/manager/download/rockylinux8-x86_64/getPackage/1/a69d8873b2131a2322457e26850d9b9a33daf98f2143f648cd418e7fe1aec9ce/psmisc-23.1-5.el8.x86_64.rpm - HIER_NONE/- application/octet-stream
```

mcalmer:
With a bit debug logging in squid I got this:
```
http.cc(1054) haveParsedReplyHeaders: decided: do not cache and do not share because authenticated and server reply missing Cache-Control; HTTP status 200 e:=p2XIV/0x5614498fe520*3
```
At least I know where to look.

phibid:
[...]

PR #5406, opened 2022-05-10
convert Basic Auth into token auth to make caching possible

mcalmer:
## What does this PR change?

Debians "apt" does not work with token auth. As a workaround we configured basic auth and use the token as "login" (username). The password is empty.
But squid does not cache files recieved via an authenticated transaction which make the use of the proxy useless.

To solve this problem, this PR convert the basic auth into token auth. With token auth we have a solution to get
the files cached.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- TBD

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17776
Fixes https://github.com/uyuni-project/uyuni/issues/5241
Tracks https://github.com/SUSE/spacewalk/pull/17831

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/proxy/proxy/broker/rhnBroker.py
+++ b/proxy/proxy/broker/rhnBroker.py
@@ -169,9 +169,12 @@ class BrokerHandler(SharedHandler):
             # we need to remove Basic Auth, otherwise squid does not cache the package
             # so we convert it into token auth
             # The token is the login. So it is not secret
-            lpw = ustr(base64.b64decode(self.req.headers_in['Authorization'][6:])) # "Basic " == 6 characters
-            self.authToken = lpw[:lpw.find(':')]
-            del self.req.headers_in['Authorization']
+            try:
+                lpw = ustr(base64.b64decode(self.req.headers_in['Authorization'][6:])) # "Basic " == 6 characters
+                self.authToken = lpw[:lpw.find(':')]
+                del self.req.headers_in['Authorization']
+            except Exception as e:
+                log_error("Unable to decode Authorization header.", e)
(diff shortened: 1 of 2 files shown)


proxy/proxy/broker/rhnBroker.py
```diff
@@ -156,12 +158,20 @@ def _initConnectionVariables(self, req):
...
         # Traditional SLE and RHEL clients uses 'X-RHN-Auth' header, but
         # no auth token is used in order to authenticate.
         if 'X-Mgr-Auth' in self.req.headers_in:
             self.authToken = self.req.headers_in['X-Mgr-Auth']
             del self.req.headers_in['X-Mgr-Auth']
+        elif 'Authorization' in self.req.headers_in and effectiveURI_parts.path.startswith('/rhn/manager/download/'):
+            # we need to remove Basic Auth, otherwise squid does not cache the package
+            # so we convert it into token auth
+            # The token is the login. So it is not secret
+            lpw = ustr(base64.b64decode(self.req.headers_in['Authorization'][6:])) # "Basic " == 6 characters
```
vzhestkov, line 172: @mcalmer I think it's quite possible to get exception here on calling `base64.b64decode`. Is it safe not to catch the exception here?

mcalmer: I catch now exceptions, log an error and keep the Authorization header. It will fail later as it does now.

vzhestkov (approved):
Looks good to me.
Thanks for addressing the exception handling.

PR #5411, opened 2022-05-10
Fix numerous branding bugs

Etheryte:
## What does this PR change?

Fixes a large number of branding bugs, adds some structure to the branding directories.  
Long-term it's still a fair bit of messy yarn, but perhaps at least a bit less tangled now.  

## GUI diff

Numerous bugfixes.

- [ ] **DONE**

## Documentation
- No documentation needed: Docs ticket already exists.

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Part of https://github.com/SUSE/spacewalk/issues/17232

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/javascript/spacewalk-essentials.js
```diff
@@ -221,7 +236,13 @@ function makeAjaxHandler(callbackFunction, errorHandlerFunction) {
 function showFatalError(message, exception) {
   console.log("DWR AJAX call failed with message: " + message);
   console.log(exception);
-  alert("Unexpected error, please reload the page and check server logs.");
+  var message = "Unexpected error, please reload the page and check server logs.";
+  // TODO: Fix
```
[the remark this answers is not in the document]
Etheryte: Oh, originally `window.showErrorToastr` didn't work, but I forgot to remove the todo after fixing it. Thanks for catching this.

web/html/src/branding/css/susemanager-light.less
```diff
@@ -1,32 +1,5 @@
...
-@import url(susemanager/panels.less);
-@import url(susemanager/nav.less);
-@import url(susemanager/titles.less);
-@import url(susemanager/toolbar.less);
-@import url(susemanager/inputs.less);
-@import url(susemanager/tables.less);
-
-// LEAVE THIS LINE AT THE END ALWAYS TO WARRANTY THE RESPONSIVE RULES REWRITE THE EXISTING STYLES
-@import url(responsive-rules.less);
+@import url(susemanager/variables.less);
```
ncounter, line 3: Maybe this could be moved into `susemanager/susemanager-theme.less` as well, and drop it from here and from `susemanager-dark.less`?

Btw, well done with this refactoring and centralizing what is common and shared. I remember when I started a while ago with this long journey, there is still a lot to do, but it already looks much much better now with this changes you are making they are very inline with the target, TYVM :heart_decoration:

Etheryte: I originally wanted to do that, but I left it as is because this way we can have shared variables first (`susemanager/variables.less`), the any overrides if we need them for specifically for light/dark theme (`susemanager/susemanager-light-variables.less`) and then the theme. Does this sound ok for you?

ncounter: A-ha! I was thinking exactly like you said, but I didn't realized that `susemanager/susemanager-theme.less` is imported *after* `susemanager/susemanager-light-variables.less` so my suggestion wouldn't work because `susemanager/variables.less` would overrides the specific light/dark one. It is tricky and sad we cannot simplify it further, but that's how it is. Forget about my comment then. :+1:

Etheryte:
Merging as we discussed on Slack.

PR #5421, opened 2022-05-11
Wait for inspect action before checking image packages

nadvornik:
## What does this PR change?

The image packages list is valid after the successful end of inspect action.
The GUI test for finished build checked only build action. API test checks both build and inspect actions.

This PR makes sure that the API test is called before the test for image packages.

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/buildhost_docker_build_image.feature
+++ b/testsuite/features/secondary/buildhost_docker_build_image.feature
@@ -59,6 +59,11 @@ Feature: Build container images
     And I schedule the build of image "suse_real_key" via API calls
     And I wait at most 660 seconds until event "Image Build suse_real_key scheduled by admin" is completed
     And I wait at most 60 seconds until all "3" container images are built correctly in the GUI
+    # We should see the same result via API.
+    # Also, check that all inspect actions are finished:
+    And I wait at most 660 seconds until container "suse_key" with version "latest" is built successfully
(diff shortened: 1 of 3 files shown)


srbarrios (approved):

Bischoff (changes requested):
I am a bit concerned by the phrasing...

First we check that the 3 images are "built correctly", then we check that each image is "built successfully"... That sounds a bit silly and redundant.

What about changing the "all 3 images are built correctly" implementation instead, to wait for the inspection as well? Or rename the "built successfully" step to "inspected successfully" ?

testsuite/features/secondary/buildhost_docker_build_image.feature
```diff
@@ -59,6 +59,11 @@ Feature: Build container images
     And I schedule the build of image "suse_real_key" via API calls
     And I wait at most 660 seconds until event "Image Build suse_real_key scheduled by admin" is completed
     And I wait at most 60 seconds until all "3" container images are built correctly in the GUI
+    # We should see the same result via API.
```
Bischoff, line 62: add TODO: keyboard?

Bischoff (approved):
Thanks for the changes.

Just add `TODO:` keyword to first comment?

Other than that, LGTM. Don't forget the ports :-)

PR #5422, opened 2022-05-12
Added a property to avoid automatic cleaning before each ant compilation

mackdk:
## What does this PR change?

This PR makes a small change in the ant `manager-build.xml` file. It allows to skip the `clean` step performed currently before each `compile`. In order to enable it, a new property `build.incremental.compile` needs to be defined inside the custom properties file `manager-developer-build.properties`. Without it, the build process maintains the old behaviour. 

When the property is enabled, the `clean` step will work as expected if invoked directly. So the old behaviour can still be obtained by chaining the tasks:

```
ant -f manager-build.xml clean webapp
```

This change is meant only as a developer convenience, to save time during build and deploy. 

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: change only affect the build process

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/java/buildconf/manager-developer-build.properties.example
+++ b/java/buildconf/manager-developer-build.properties.example
@@ -4,6 +4,9 @@ deploy.host = d52.suse.de
 # Uncomment to use .class files from the build/classes directory directly to avoid calling javac
 #precompiled = true
 
+# Uncomment to avoid performing a clean before each compilation
+#build.incremental.compile = true
+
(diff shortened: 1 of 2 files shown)


renner (approved):
Sorry, I overlooked this PR for a while. To me it makes sense and the patch looks good, thanks!

PR #5429, opened 2022-05-12
Add Basesystem repo

Bischoff:
## What does this PR change?

This PR adds the missing basesystem repo to the Docker profile now we switched Docker builds to sle 15 sp2.

PLEASE CHECK THE REPO URL. IT IS PROBABLY WRONG.

(and yeah, it's using sle 15 sp4 repos on sle 15 sp2 images... but do we really want to fix that?)

## Links

No ports.

## Changelogs

- [x] No changelog needed


Bischoff:
No reviewers available, merging, cannot be worse than current situation

PR #5430, opened 2022-05-12
Fix the confirm message on SW and HW refresh action

wweellddeerr:
## What does this PR change?

It fixes the confirm message on SW and HW refresh action by adding a link to the Pending Actions page on it.

## GUI diff

No difference. Bug fix.

- [x] **DONE**

## Documentation
- No documentation needed: just a bug fix. 

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17240

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
```diff
@@ -5067,7 +5067,7 @@ organization. You may grant or remove the organization administrator role in the
         </context-group>
         </trans-unit>
         <trans-unit id="ssm.hw.systems.confirmmessage">
-          <source>{0} hardware {1} will be refreshed.</source>
+          <source>{0} hardware {1} &lt;a href="/rhn/schedule/PendingActions.do"&gt;{2} scheduled&lt;/a&gt; to be refreshed.</source>
```
cbbayburt, line 5070: All the strings in this file are translated to various languages, particularly to Chinese by 3rd party translators. When we hide some words in the parameters (like "profile", "is/are") translators cannot figure it out and the translated text is most likely going to be broken.

So to fix this in a way that both supports translation and singular/plural distinction, I'd suggest we duplicate this entry to have a singular and a plural version. You can use `ssm.hw.systems.confirmmesage` as the singular version, and add another `<trans-unit>` entry with ID `ssm.hw.systems.confirmmessage.multiple` and write the plural text there. In the end, the text itself will be intact, and only parameter will be the number of profiles.

java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
```diff
@@ -5098,7 +5098,7 @@ organization. You may grant or remove the organization administrator role in the
         </context-group>
         </trans-unit>
         <trans-unit id="ssm.sw.systems.confirmmessage">
-          <source>{0} package {1} will be refreshed.</source>
+            <source>{0} package {1} &lt;a href="/rhn/schedule/PendingActions.do"&gt;{2} scheduled&lt;/a&gt; to be refreshed.</source>
```
cbbayburt, line 5101: Same as above

cbbayburt (changes requested):
Thanks for the quick fix, @wweellddeerr 

I've commented some suggested changes, mainly for translation support.

parlt91 (approved):
Other than what Can already mentioned it looks good to me
Thank you :tada:

cbbayburt (approved):
All looks good now :+1: 

Next: please rebase your work on top again fixing the conflicts, wait for the tests to run, and then you can merge directly. Also please consider squashing both commits into one when merging, since they both basically achieve the same goal.

PR #5434, opened 2022-05-13
HTTP API: Handle cases where there is an error code, but no error message

Bischoff:
## What does this PR change?

Test suite: better error management for HTTP API.

## Links

No ports, Uyuni only.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/http_client.rb
+++ b/testsuite/features/support/http_client.rb
@@ -58,6 +58,8 @@ class HttpClient
         end
       end
     unless answer.status == 200
+      raise "Unexpected HTTP status code #{answer.status}" if answer.body.empty?
+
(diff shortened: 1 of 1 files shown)


vandabarata (approved):

nodeg (approved):
LGTM.

testsuite/features/support/http_client.rb
```diff
@@ -58,6 +58,8 @@ def call(name, params)
         end
       end
     unless answer.status == 200
+      raise "Unexpected HTTP status code #{answer.status}" if answer.body.empty?
```
srbarrios, line 61: I think we should also expose the answer.headers, to debug it better.

PR #5438, opened 2022-05-13
Fix issue removing duplicated packages names in the same action (bsc#1198686)

meaksh:
## What does this PR change?

This PR fixes and issue happening when scheduling a package remove action targeting multiple versions of the same package name. For example: `kernel-devel` or `kernel-source` packages, which we can have multiple versions installed at the same time on a system. 

The problem is that Salt is not able to process duplicates packages names as part of the same `pkg.removed` state execution. So, in case we are under this situation, and we schedule the removal of these packages from the Uyuni UI, the generated `packages.pkgremove` state would be something like:

```yaml
pkg_removed:
  pkg.removed:
    -   pkgs:
        - kernel-source.noarch: 4.12.14-122.80.1
        - kernel-source.noarch: 4.12.14-122.60.1
    -   require:
        - file: mgrchannels*
```

Notice the duplicated packages names as part of `pkgs`.

Now, when this state is proceesed by Salt, only one of the duplicaed items in `pkgs` will remain and used for the execution, producing unconsistent results (as it is not taking all the specified packages into consideration). Fixing this behavior in the Salt side to allow duplicates packages in this context is not really straight forward. This has to do with Salt internals structures and data are organized and it would require time and some considerable refactoring.

For now, I've pushed a fix to Salt (here: https://github.com/saltstack/salt/pull/62019) in order to produce an error message and failed state execution and not cause inconsistent results when the user is passing duplicated packages names as above.

So in order to fix this situation, this PR is making Uyuni to generate the `packages.pkgremove` state in a way that is able to be consumed by Salt in case of additional packages.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/17591

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
+++ b/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
@@ -1181,12 +1181,28 @@ public class SaltServerActionService {
     private Map<LocalCall<?>, List<MinionSummary>> packagesRemoveAction(
             List<MinionSummary> minionSummaries, PackageRemoveAction action) {
         Map<LocalCall<?>, List<MinionSummary>> ret = new HashMap<>();
-        List<List<String>> pkgs = action
+        List<List<String>> pkgsAll = action
                 .getDetails().stream().map(d -> Arrays.asList(d.getPackageName().getName(),
                         d.getArch().toUniversalArchString(), d.getEvr().toUniversalEvrString()))
                 .collect(Collectors.toList());
+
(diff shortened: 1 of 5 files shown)


mcalmer (approved):
LGTM

mcalmer (approved):
LGTM

PR #5442, opened 2022-05-13
Fix notification message on system properties update to ensure style consistency

wweellddeerr:
## What does this PR change?

When activating Ansible Control Node add-on, its notification had a different style from others.
A line break and the "Note" word were added to make its style consistent.

## GUI diff

No difference. Bug fix.

- [x] **DONE**

## Documentation
- No documentation needed: just a bug fix.

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17242

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/strings/java/StringResource_en_US.xml
+++ b/java/code/src/com/redhat/rhn/frontend/strings/java/StringResource_en_US.xml
@@ -9094,25 +9094,25 @@ Alternatively, you will want to download &lt;strong&gt;Incremental Channel Conte
         </context-group>
       </trans-unit>
       <trans-unit id="system.entitle.added.monitoring_entitled" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


cbbayburt (changes requested):
Virtualization and monitoring cases have different messages because of the different tasks they do. But I think we can still unify those messages a bit too. Maybe something like:

___
**Virtualization** type has been applied.
**Note:** Since you added this system type, we also performed some extra steps to...
___
**Monitoring** type has been applied.
**Note:** Since you added this system type, we automatically assigned the 'Prometheus Exporters' formula to the system. Please apply...
___

With this change, all the messages will follow the same structure.

wweellddeerr:
@cbbayburt I hadn't noticed the difference in the monitoring case, and I agree with you that this message can also follow the same structure, so I changed it.

cbbayburt (approved):
All looks good now. Thanks :+1:

PR #5444, opened 2022-05-16
Fix docker tests

Bischoff:
## What does this PR change?

This PR attempts to fix the building of containers in the test suite:
 - the first commit reverts a previous PR of mine which added the basesystem updates and was not needed
 - the second commit upgrades SLES 15 SP2 to SLES 15 SP4 in top-level directory, to align with the other 2 directories
 - the third commit stops installing  `python3-xml` which does not exist in SLES 15

## Links

No ports, Uyuni only.

## Changelogs

- [x] No changelog needed


nodeg (approved):
LGTM.

PR #5453, opened 2022-05-17
Do not allow creating path if nonexistent user or group in fileutils

### LINKED ISSUE #5106: uyuni.common.uyuni.common.makedirs creates folder despite error
# Problem description

When using Python function `uyuni.common.uyuni.common.makedirs` with a non-existing user/group name, the function fails with an error but still creates the next level folder.

```
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/repo_plugins/deb_src.py", line 171, in __init__
    fileutils.makedirs(self.basecachedir, user='wwwrun', group='www')
  File "/usr/lib/python3.6/site-packages/uyuni/common/fileutils.py", line 271, in makedirs
    os.chown(dirname, uid, gid)
TypeError: uid should be integer, not NoneType
```

## Version of Uyuni Server and Proxy (if used)

Master 2022-03-30 and earlier

## Details about the issue

Easily reproducible:
```
# python3
Python 3.6.8 (default, Nov 17 2021, 16:10:06) 
[GCC 8.5.0 20210514 (Red Hat 8.5.0-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from uyuni.common import fileutils
>>> fileutils.makedirs("/tmp/a/b/c/d/e/f/g" , user='sdfesaerfc', group='vsdrgtrmt')
XXX: User sdfesaerfc does not exist
XXX: Group vsdrgtrmt does not exist
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/uyuni/common/fileutils.py", line 271, in makedirs
    os.chown(dirname, uid, gid)
TypeError: uid should be integer, not NoneType
```
Folder `/tmp/a` has been created. Rerun and folder `/tmp/a/b/` will be created.

Expectation options:
1. Folder won't be created it it will run in an error.
2. Action will be reverted if there is an error.
3. Information that there was an error and a folder has been created.
4. Continue on error and create full folder and inform that ownership was not set as expected.

ycedres:
## What does this PR change?

Fixes the function `makedirs` in `uyuni.common.fileutils` that was creating the required path even if the specified user/group did not exist, and then failing to chown. Now it does not create the path and raises an exception.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed:  user invisible changes

- [X] **DONE**

## Test coverage
- No tests:  bugfix

- [X] **DONE**

## Links

Fixes #5106 
Tracks https://github.com/SUSE/spacewalk/issues/17730

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/uyuni/common/fileutils.py
+++ b/python/uyuni/common/fileutils.py
@@ -239,12 +239,29 @@ def rhn_popen(cmd, progressCallback=None, bufferSize=16384, outputLog=None):
 
 
 def makedirs(path,  mode=int('0755', 8), user=None, group=None):
-    "makedirs function that also changes the owners"
+    """Creates all required directories on a path and changes its owner and group
+
+    :param path: path to create
+    :type path: str
+    :param mode: mode for the created directories
+    :type mode: int
+    :param user: desired owner
+    :type user: str
+    :param group: desired group
+    :type group:str
+    :returns: None
+    """
 
     dirs_to_create = []
     dirname = path
 
     uid, gid = getUidGid(user, group)
+
+    if uid is None:
+        raise OSError("*** ERROR: user %s doesn't exist. Cannot create path." % user)
+
+    if gid is None:
+        raise OSError("*** ERROR: group %s doesn't exist. Cannot create path." % group)
 
     while 1:
         if os.path.isdir(dirname):
@@ -356,8 +373,6 @@ class GecosCache:
         try:
             uid = pwd.getpwnam(name)[2]
         except KeyError:
-            # XXX misa: gripe? taw: I think we need to do something!
-            sys.stderr.write("XXX: User %s does not exist\n" % name)
             return None
         self._users[name] = uid
         return uid
@@ -369,8 +384,6 @@ class GecosCache:
         try:
             gid = grp.getgrnam(name)[2]
         except KeyError:
-            # XXX misa: gripe?
-            sys.stderr.write("XXX: Group %s does not exist\n" % name)
             return None
         self._groups[name] = gid
         return gid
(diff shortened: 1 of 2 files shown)


vzhestkov (approved):
Looks good :+1: 
Just one really nitpicky comment: maybe additionally `blacken` the modified python source file?

python/uyuni/common/fileutils.py
```diff
@@ -239,13 +239,19 @@ def rhn_popen(cmd, progressCallback=None, bufferSize=16384, outputLog=None):
...
-    "makedirs function that also changes the owners"
+    """makedirs function that also changes the owners"""
 
     dirs_to_create = []
     dirname = path
 
     uid, gid = getUidGid(user, group)
 
+    if uid is None:
+        raise OSError(f"*** ERROR: user {user} doesn't exist. Cannot create path.")
```
agraul, line 250: Code in `uyuni.common` is used on both Uyuni server and the clients, not all of which use Python 3.6+. We need to check that only code that's exclusive to Uyuni server can hit this code path or wait until we're free to be incompatible with the Python interpreters on client systems.

SUMA 4.3 will be the last release that supports the traditional stack, after its release we can start introducing incompatibilities (on purpose).

python/uyuni/common/fileutils.py
```diff
@@ -381,24 +383,20 @@ def reset(self):
 
 
 def getUidGid(user=None, group=None):
-    "return uid, gid given user and group"
+    """return uid, gid given user and group"""
```
agraul, line 386: Same as the other docstring.

agraul (changes requested):
Logic is fine, but I'm not sure if we can ship f-strings just yet. Maybe @meaksh can share his opinion on how to progress.

agraul:
> Looks good +1 Just one really nitpicky comment: maybe additionally `blacken` the modified python source file?

IMO we shouldn't run `black` on individual files just yet. I think it's better to do it once for the whole project, in its own commit, at the same time we add the tooling into our CI and git hooks. I share Black's creator's [reasoning](https://lukasz.langa.pl/36380f86-6d28-4a55-962e-91c2c959db7a/) on this topic.

ycedres:
Successfully built in the CI (build 1464) but without running all scopes. Another build including the secondary tests has been run (1468).

juliogonzalez:
Since the branch freeze was on Friday, this a P3, and we have some failures on secondary, let's postpone merging this for the next version.

PR #5456, opened 2022-05-17
Back to SLE 15 SP2 for containers

Bischoff:
## What does this PR change?

In previous PR, I forgot we don't have SLES 15 SP4 toaster images. It was correct to align between `Docker/` and `Docker/serverhost` and `Docker/authprofile`, but should have been in the reverse direction. My bad.

The alternative could be to get help from @agraul to move away from toaster images and use BCI (Base Container Images).

## Links

No ports, Uyuni only.

## Changelogs

- [x] No changelog needed


vandabarata (approved):
lgtm

srbarrios:
That card https://github.com/SUSE/spacewalk/issues/17778 aimes to add a new image that could help us to move to SLES15SP4, even if ideally we should just use BCI, we might still just using the one produced here. Also, to generate the one for Salt, they need a BCI :)

PR #5462, opened 2022-05-17
Fix Internal Server Error when URI contains invalid sysid

wweellddeerr:
## What does this PR change?

When the URL is a valid path but contains an invalid sysid, the system was trying to lookup the server based on that wrong sysid
generating a LookupException.

The LookupException was captured and logged as a warning leading to a permission error to the user instead of an Internal Server Error.

## GUI diff

No difference. Bug fix.

- [x] **DONE**

## Documentation
- No documentation needed: bug fix.

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17216

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/common/security/acl/Access.java
+++ b/java/code/src/com/redhat/rhn/common/security/acl/Access.java
@@ -18,6 +18,7 @@ import com.redhat.rhn.common.conf.Config;
 import com.redhat.rhn.common.db.datasource.DataResult;
 import com.redhat.rhn.common.db.datasource.ModeFactory;
 import com.redhat.rhn.common.db.datasource.SelectMode;
+import com.redhat.rhn.common.hibernate.LookupException;
 import com.redhat.rhn.domain.channel.Channel;
 import com.redhat.rhn.domain.channel.ChannelFactory;
 import com.redhat.rhn.domain.errata.Errata;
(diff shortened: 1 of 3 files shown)


java/code/src/com/redhat/rhn/common/security/acl/Access.java
```diff
@@ -214,11 +215,17 @@ public boolean aclSystemHasManagementEntitlement(Object ctx, String[] params) {
...
-        if (server == null) {
+        try {
+            Server server = SystemManager.lookupByIdAndUser(sid, user);
+            if (server == null) {
+                return false;
+            }
+            return server.hasEntitlement(EntitlementManager.MANAGEMENT);
+        }
+        catch (LookupException e) {
+            LOG.warn(e);
```
cbbayburt, line 226: I think we don't need to log this, because this is the expected result with nonexisting systems, and there's no further action needed from the user.

java/code/src/com/redhat/rhn/common/security/acl/Access.java
```diff
@@ -234,9 +241,14 @@ public boolean aclSystemHasSaltEntitlement(Object ctx, String[] params) {
...
-            if (server != null) {
-                ret = server.hasEntitlement(EntitlementManager.SALT);
+            try {
+                Server server = SystemManager.lookupByIdAndUser(sid, user);
+                if (server != null) {
+                    ret = server.hasEntitlement(EntitlementManager.SALT);
+                }
+            }
+            catch (LookupException e) {
+                LOG.warn(e);
```
cbbayburt, line 251: same as above

cbbayburt (approved):
LGTM :+1: 

Just added one comment below

java/code/src/com/redhat/rhn/common/security/acl/Access.java
wweellddeerr: Thank you for the review. I agree with you and I removed the logs.

PR #5463, opened 2022-05-18
Use basesystem repo

Bischoff:
## What does this PR change?

Use the basesystem module for sle15sp2

Another PR to move back to SP4 will follow once we get the registry issue fixed

## Links

No ports, Uyuni only.

## Changelogs

- [x] No changelog needed


vandabarata (approved):
lgtm

PR #5492, opened 2022-05-24
Do not mingle with generic step

Bischoff:
## What does this PR change?

This PR:
* removes search-specific code from the generic `I click on "..."` step
* makes use of `repeat_until_timeout` construct

## Links

Ports:
* 4.2: SUSE/spacewalk#18264

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -255,21 +255,6 @@ end
 #
 When(/^I click on "([^"]*)"$/) do |text|
   click_button_and_wait(text, match: :first)
-
-  # In case of a search reindex not having finished yet, keeps retrying until successful search or timeout
-  if text == 'Search' && has_text?('Could not connect to search server.', wait: 0)
-    click_button(text, match: :first, wait: false)
-    start = Time.new
-    timeout = 10
-    while (has_text?('Could not connect to search server.', wait: 0) || has_text?('No matches found', wait: 0)) &&
-          (Time.new - start <= timeout)
-
-      click_button(text, match: :first, wait: false)
-    end
-    if Time.new - start > timeout
-      raise 'Could not perform a successful search after reindexation'
-    end
-  end
 end
 
 #
@@ -1104,6 +1089,19 @@ Then(/^I should land on system's overview page$/) do
     And I should see a "System Properties" text
     And I should see a "Subscribed Channels" text
         )
+end
+
+# In case of a search reindex not having finished yet, keep retrying until successful search or timeout
+When(/^I click on the search button$/) do
+  click_button_and_wait('Search', match: :first)
+  # after a search reindex, the UI will show a "Could not connect to search server" followed by a false "No matches found" for a while
+  if has_text?('Could not connect to search server.', wait: 0)
+    repeat_until_timeout(message: 'Could not perform a successful search after reindexation', timeout: 10) do
+      break unless has_text?('Could not connect to search server.', wait: 0) || has_text?('No matches found', wait: 0)
+      sleep 1
+      click_button('Search', match: :first, wait: false)
+    end
+  end
 end
 
 When(/^I enter "([^"]*)" hostname on the search field$/) do |host|
--- a/testsuite/features/secondary/srv_advanced_search.feature
+++ b/testsuite/features/secondary/srv_advanced_search.feature
@@ -17,7 +17,7 @@ Feature: Advanced Search
     And I enter "sle_minion" hostname on the search field
     And I select "Hostname" from "Field to Search"
     And I check "invertlabel"
-    And I click on "Search"
+    And I click on the search button
     Then I should see a "No results found." text
 
   Scenario: One search result for City
@@ -26,7 +26,7 @@ Feature: Advanced Search
     And I enter "Little Whinging" on the search field
     And I select "City" from "Field to Search"
     And I check "fineGrainedlabel"
-    And I click on "Search"
+    And I click on the search button
     Then I should land on system's overview page
 
   Scenario: One search result for State/Province
@@ -35,7 +35,7 @@ Feature: Advanced Search
     And I enter "Surrey" on the search field
     And I select "State/Province" from "Field to Search"
     And I check "fineGrainedlabel"
-    And I click on "Search"
+    And I click on the search button
     Then I should land on system's overview page
 
   Scenario: One search result for Country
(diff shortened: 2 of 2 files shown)


testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -1107,6 +1092,17 @@
         )
 end
 
+# In case of a search reindex not having finished yet, keep retrying until successful search or timeout
+When(/^I click on the search button$/) do
+  click_button_and_wait('Search', match: :first)
+
+  repeat_until_timeout(message: 'Could not perform a successful search after reindexation', timeout: 10) do
+    break unless has_text?('Could not connect to search server.', wait: 0) || has_text?('No matches found', wait: 0)
+    sleep 1
```
srbarrios, line 1101: Why we wait 1 second?

Bischoff: to avoid active wait ; all other calls return immediately

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -1107,6 +1092,17 @@
         )
 end
 
+# In case of a search reindex not having finished yet, keep retrying until successful search or timeout
+When(/^I click on the search button$/) do
+  click_button_and_wait('Search', match: :first)
+
+  repeat_until_timeout(message: 'Could not perform a successful search after reindexation', timeout: 10) do
```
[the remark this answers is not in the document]
Bischoff: OK, after discussion, the meaning of "No results found" changes according to the iteration. I resurrected your initial test.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -1107,6 +1092,19 @@
         )
 end
 
+# In case of a search reindex not having finished yet, keep retrying until successful search or timeout
```
vandabarata, line 1095:
```suggestion
# Specifc test for search due to having to deal with search reindex conditionals
```

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -1107,6 +1092,19 @@
...
+When(/^I click on the search button$/) do
+  click_button_and_wait('Search', match: :first)
+  if has_text?('Could not connect to search server.', wait: 0)
+    # the further "No matches found" after an initial "Could not connect" are not real, it's just that it's reindexing
```
vandabarata, line 1099:
```suggestion
    # After a search reindex, the UI will show a "Could not connect to search server" followed by a false "No matches found" for a while
```

testsuite/features/step_definitions/navigation_steps.rb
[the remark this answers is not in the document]
vandabarata: Apparently, everything seems to be working out great - not sure what the issue was months ago, but it doesn't seem to still be happening so let's go with this approach, it's cleaner. We can always revert it if it causes any issues

vandabarata:
Left this running on CI just to double check it's working properly - https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/1334/

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: Yup.

Bischoff:
> Left this running on CI just to double check it's working properly - https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/1334/

Good idea.

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: Done.

testsuite/features/step_definitions/navigation_steps.rb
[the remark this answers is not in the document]
Bischoff: Nice!

testsuite/features/step_definitions/navigation_steps.rb
srbarrios: In such case, we should move that wait after clicking the button, in that position it doesn't make sense.

vandabarata (approved):

testsuite/features/step_definitions/navigation_steps.rb
Bischoff: You need to test the result immediately after pressing the button.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -1107,6 +1092,19 @@
...
+When(/^I click on the search button$/) do
+  click_button_and_wait('Search', match: :first)
+  # after a search reindex, the UI will show a "Could not connect to search server" followed by a false "No matches found" for a while
+  if has_text?('Could not connect to search server.', wait: 0)
+    repeat_until_timeout(message: 'Could not perform a successful search after reindexation', timeout: 10) do
+      break unless has_text?('Could not connect to search server.', wait: 0) || has_text?('No matches found', wait: 0)
+      sleep 1
+      click_button('Search', match: :first, wait: false)
+    end
+  end
```
srbarrios, line 1105:
```suggestion
# after a search reindex, the UI will show a "Could not connect to search server" followed by a false "No matches found" for a while
  repeat_until_timeout(message: 'Could not perform a successful search after reindexation', timeout: 10) do
    click_button('Search', match: :first)
    break unless has_text?('Could not connect to search server.', wait: 1) || has_text?('No matches found', wait: 1)
  end
```

Bischoff: That would indeed be much simpler, and that was also my first idea.

That will not work though because of the weird behaviour explained in the comment:
```
# after a search reindex, the UI will show a "Could not connect to search server" followed by a false "No matches found" for a while
```
For more information, ask Vanda.

vandabarata: Also, to add, the sleep 1 actually helps us not overload the system and keep retrying every single possible moment - gives the testsuite some time to "breathe" before retrying, without taking up too much time - it takes about as long as my previous version, I tested it, but without constantly overloading the system. Kind of like awaitility works for java, where you can set a retry interval - this sleep immitates that behaviour

vandabarata: I just realised you added the wait to waiting for the text but that is _not_ the same thing - it's not about the text showing up, it's about us waiting _before_ retrying the search again

srbarrios: Ok :)

PR #5494, opened 2022-05-24
Set __low__  on mgrcompat.module_run to pass to module.run

vzhestkov:
## What does this PR change?

Prevents traceback with undefined `__low__` as `mgrcompat` doesn't respect to LazyLoader and not passing `__low__` to `module.run`

## GUI diff

No difference.

## Documentation

## Test coverage
- Unit tests were added

## Changelogs

- [ ] No changelog needed

## Re-run a test


agraul:
@vzhestkov could we instead import `salt.states.modules` the way it's supposed to be done ? I haven't looked into the details, I imagine it might be a cleaner approach.

vzhestkov:
Yes, you are correct, but this change is more simple and in general with the latest plans of upstream for `module.run`, I think we can just remove `mgrcompat` at all.

vzhestkov:
Closing in favor of https://github.com/uyuni-project/uyuni/pull/5526

@agraul thanks for the idea for better solution :+1:

PR #5506, opened 2022-05-26
Block parallel_tests Gem version

srbarrios:
## What does this PR change?

Some recent versions of parallel_tests are failing.
Let's keep an older stable release.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.1
- Manager-4.2

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/Gemfile
+++ b/testsuite/Gemfile
@@ -17,7 +17,7 @@ gem 'minitest'
 gem 'websocket'
 gem 'websocket-driver'
 gem 'syntax'
-gem 'parallel_tests'
+gem 'parallel_tests', '3.8.1'
 gem 'simplecov'
 gem 'rubocop'
 # Pre-requisite:
--- a/testsuite/Gemfile
+++ b/testsuite/Gemfile
@@ -17,7 +17,7 @@ gem 'minitest'
 gem 'websocket'
 gem 'websocket-driver'
 gem 'syntax'
-gem 'parallel_tests', '3.8.1'
+gem 'parallel_tests', '3.1.0'
 gem 'simplecov'
 gem 'rubocop'
 # Pre-requisite:


nodeg:
Here is the [changelog](https://github.com/grosser/parallel_tests/blob/master/CHANGELOG.md) for reference:

```bash
3.10.1 - 2022-05-23
Fixed
Running rake tasks with number of processes or extra args
3.10.0 - 2022-05-23
Added
Changed Rake subtasks to always use the same Rake executable as the parent process.
3.9.1 - 2022-05-23
Fixed
Fixed NoMethodError exception when running Rake task parallel:setup.
3.9.0 - 2022-05-22
Added
Subprocesses execute without a shell.
```
Not that much changed with newer versions. I guess 3.10.0 could have introduced the issue.

srbarrios:
> Here is the [changelog](https://github.com/grosser/parallel_tests/blob/master/CHANGELOG.md) for reference:
> 
> ```shell
> 3.10.1 - 2022-05-23
> Fixed
> Running rake tasks with number of processes or extra args
> 3.10.0 - 2022-05-23
> Added
> Changed Rake subtasks to always use the same Rake executable as the parent process.
> 3.9.1 - 2022-05-23
> Fixed
> Fixed NoMethodError exception when running Rake task parallel:setup.
> 3.9.0 - 2022-05-22
> Added
> Subprocesses execute without a shell.
> ```
> 
> Not that much changed with newer versions. I guess 3.10.0 could have introduced the issue.

As a matter of fact the Gem installed when the pipeline ran was 3.9.1

testsuite/Gemfile
```diff
@@ -17,7 +17,7 @@ gem 'minitest'
 gem 'websocket'
 gem 'websocket-driver'
 gem 'syntax'
-gem 'parallel_tests'
+gem 'parallel_tests', '3.1.0'
```
nodeg, line 20:
```suggestion
gem 'parallel_tests', '3.8.1'
```
According to https://github.com/grosser/parallel_tests/issues/872 we should use version 3.8.1 as a quick fix for now.

nodeg (changes requested):

testsuite/Gemfile
srbarrios: Yes. But according to our env, 3.8.1 will cause an issue in cascade now, with another dependency requiring ruby 2.7

I prefer to just be in the safe side for once 😁

nodeg: Oh, I did not know that. So then `3.1.0` it is.

nodeg (approved):

grosser:
fixed in 3.11.1

testsuite/Gemfile
nodeg, line 20:
```suggestion
gem 'parallel_tests', '3.11.1'
```
I guess we can now use the most recent version.

nodeg (changes requested):

srbarrios:
We don't have issues with this anymore, let's give a second chance and keep using the latest. Closing the PR for now.

PR #5538, opened 2022-06-03
Add non-executable-script rpmlint filter for salt shell files

vzhestkov:
## What does this PR change?

Prevent the following output on building the package:
```
[   53s] susemanager-sls.noarch: W: non-executable-script /usr/share/susemanager/salt-ssh/preflight.sh 644 /bin/sh 
[   53s] susemanager-sls.noarch: W: non-executable-script /usr/share/susemanager/salt/actionchains/force_restart_minion.sh 644 /bin/bash 
[   53s] This text file contains a shebang or is located in a path dedicated for
[   53s] executables, but lacks the executable bits and cannot thus be executed.  If
[   53s] the file is meant to be an executable script, add the executable bits,
[   53s] otherwise remove the shebang or move the file elsewhere.
```

It's safe to have the shell scripts with 0644 mod under `/usr/share/susemanager/salt` as these scipts are intended to be running on the minions with the states or salt-ssh preflight.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: not required

## Changelogs

- [x] No changelog needed

## Re-run a test


susemanager-utils/susemanager-sls/susemanager-sls.changes
```diff
@@ -1,3 +1,5 @@
+- Ignore non-executable-script rpmlint ceheck for shell scripts
```
juliogonzalez, line 1: No changelog needed in this case, as this is not relevant for users, and we already have another entry that will allow us to tag.

susemanager-utils/susemanager-sls/susemanager-sls.spec
```diff
@@ -37,6 +37,7 @@ Summary:        Static Salt state files for SUSE Manager
 License:        Apache-2.0 AND LGPL-2.1-only
 Group:          Applications/Internet
 Source:         %{name}-%{version}.tar.gz
+Source1:        https://raw.githubusercontent.com/uyuni-project/uyuni/%{name}-%{version}-1/susemanager-utils/susemanager-sls/%{name}-rpmlintrc
```
juliogonzalez, line 40: Are you sure you need this as a source? This is not used for building the package, but after building it, IIRC?

juliogonzalez (changes requested):

susemanager-utils/susemanager-sls/susemanager-sls.spec
vzhestkov: I see the same for the other packages having `rpmlintrc`, not really sure what was the reason for adding it.

susemanager-utils/susemanager-sls/susemanager-sls.changes
vzhestkov: Just dropped the commit with the changelog entry

susemanager-utils/susemanager-sls/susemanager-sls.spec
juliogonzalez: Interesting... Can you try to build the package without it?

Just branch susemanager-sls from Uyuni master, add the `rpmlintrc` to the branched package, and see what happens.

vzhestkov: It will not bring this file as a source inside the build system then, so `SourceX: ... ` should be specified anyway. If I understood OBS internals correctly it basically brings the local file in case if it has long URL and not trying to fetch this file.

But the same URLs are specified for the following specs:
./client/rhel/mgr-daemon/mgr-daemon.spec
./client/rhel/spacewalk-client-tools/spacewalk-client-tools.spec
./client/tools/mgr-cfg/mgr-cfg.spec
./client/tools/mgr-osad/mgr-osad.spec
./client/tools/mgr-push/mgr-push.spec
./client/tools/mgr-virtualization/mgr-virtualization.spec
./client/tools/spacewalk-koan/spacewalk-koan.spec
./client/tools/spacewalk-oscap/spacewalk-oscap.spec
./java/spacewalk-java.spec
./projects/ssl-cert-check/spacewalk-ssl-cert-check.spec
./schema/reportdb/uyuni-reportdb-schema.spec
./schema/spacewalk/susemanager-schema.spec
./containers/proxy-systemd-services/uyuni-proxy-systemd-services.spec

vzhestkov: You can check with `find . -name '*.spec' | xargs -I {} grep -H '\-rpmlintrc' {}` from the root of the project

juliogonzalez: > It will not bring this file as a source inside the build system then, so `SourceX: ... ` should be specified anyway. If I understood OBS internals correctly it basically brings the local file in case if it has long URL and not trying to fetch this file.
> 
> But the same URLs are specified for the following specs: ./client/rhel/mgr-daemon/mgr-daemon.spec ./client/rhel/spacewalk-client-tools/spacewalk-client-tools.spec ./client/tools/mgr-cfg/mgr-cfg.spec ./client/tools/mgr-osad/mgr-osad.spec ./client/tools/mgr-push/mgr-push.spec ./client/tools/mgr-virtualization/mgr-virtualization.spec ./client/tools/spacewalk-koan/spacewalk-koan.spec ./client/tools/spacewalk-oscap/spacewalk-oscap.spec ./java/spacewalk-java.spec ./projects/ssl-cert-check/spacewalk-ssl-cert-check.spec ./schema/reportdb/uyuni-reportdb-schema.spec ./schema/spacewalk/susemanager-schema.spec ./containers/proxy-systemd-services/uyuni-proxy-systemd-services.spec

I don't really understand what you mean: why do we exactly need `SourceX`?  Because if it doesn't exist, then the tooling we use to submit to OBS will not submit the rpmlintrc file?

vzhestkov: OBS is bringing rpmlintrc file to the sources directory inside the build instance only if it's specified with spec as I know. And the check scripts is reading the file inside this instance after building the package.

juliogonzalez: https://build.opensuse.org/package/live_build_log/home:juliogonzalezgil:branches:systemsmanagement:Uyuni:Master/susemanager-sls/openSUSE_Leap_15.4/x86_64
https://build.opensuse.org/package/show/home:juliogonzalezgil:branches:systemsmanagement:Uyuni:Master/susemanager-sls

I added only the rpmlintrc to the OBS package, but not to the spec, and it seems it's working just fine?

vzhestkov: Hmm, strange. I remember the different behavior with one of the previous package I've modified with adding rpmlintrc. Maybe we can drop the source line.

BTW, just noticed the different issue:
```
[   54s] susemanager-sls.noarch: W: script-without-shebang /usr/share/susemanager/salt/certs/suse.sls
[   54s] This text file has executable bits set or is located in a path dedicated for
[   54s] executables, but lacks a shebang and cannot thus be executed.  If the file is
[   54s] meant to be an executable script, add the shebang, otherwise remove the
[   54s] executable bits or move the file elsewhere.
```

juliogonzalez: But the problem is not OBS.

If you don't add the file to the SPEC sources, then `build-packages-for-obs.sh` does not consider it.

So in the end having the source is a hack, so that script considers the file. You can leave it.

juliogonzalez: One day we should rewrite those scripts, but today is not that day :-D

As for the other issues, feel free to fix it as well. I think you just need to remove the executable bit, as it's a SLS?

vzhestkov: done :+1:

juliogonzalez (approved):

PR #5539, opened 2022-06-03
Make Select.jsx menuPortalTarget target specific div instead of document body

wweellddeerr:
## What does this PR change?

This is a refactoring to make the property menuPortalTarget of the Select.jsx component target a specific div (also introduced here) instead of document.body, and avoid the layout to be invalidated.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: refactoring.

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17076

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/spacewalk-web.changes
```diff
@@ -1,3 +1,5 @@
+- Target specific div in Select component instead of document.body
```
Etheryte, line 1: I think this changelog entry could be a bit improved. The changelog is intended for customers, not technical people, and for them things like `div` and `document.body` won't necessarily say anything useful. I'm not even sure if I would add a changelog entry here, since from the customer's point of view, nothing visibly tangible has changed. We've improved the technical situation, yes, but they don't see it verbatim. Alternatively, could write something along the lines of a-la `Improved dropdown layout handling` or something similar which is understandable without a technical background.

Etheryte:
I think this looks good, thanks a lot for taking the time to fix this up. One question, don't we need to also update `layout_equals.jsp`? I'm not sure what that layout is used for though. Ccing @cbbayburt for a second opinion on that bit.

Etheryte (approved):

web/spacewalk-web.changes
juliogonzalez: Fully agree with @Etheryte

juliogonzalez (changes requested):

web/spacewalk-web.changes
wweellddeerr: Yes, you are right. I hadn't really stopped to think about it properly and I think this entry should be removed as you suggested.

juliogonzalez (approved):

PR #5560, opened 2022-06-10
Remove `nodejs-packaging` as a build requirement

Etheryte:
## What does this PR change?

This PR removes the build-time dependency on `nodejs-packaging`. The version we're using relies on Python 2 which is EOL.

If I understood correctly, this package was used to add provides etc information into packages in the past for security audits (perhaps @SchoolGuy can chime in with a more correct description), however it seems we currently don't use it as far as I could see and find when testing builds.  
The frontend build itself does not rely on this package, as we don't use `nodejs-less` etc anymore (Webpack handles those things instead now).

Using `rel-eng/build-packages-local-obs.sh`, `spacewalk-web` builds fine with `BuildRequires:  nodejs-packaging` removed.  

From the security side, frontend dependencies are covered by both Github's Dependabot, as well as a Github action (`frontend-dependency-audit.yml`). If we want to add information about frontend dependencies into packages, we could update or augment the way we wrap those dependencies up in the future (currently we make a tarball via `yarn zip`). The information is readily available, we just need to agree on if and how we want to approach this.  

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: build tooling.

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/13521

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/spacewalk-web.changes
+++ b/web/spacewalk-web.changes
@@ -1,3 +1,4 @@
+- Remove nodejs-packaging as a build requirement
 - Update translation strings
 
 -------------------------------------------------------------------
--- a/web/spacewalk-web.spec
+++ b/web/spacewalk-web.spec
@@ -47,7 +47,6 @@ BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildArch:      noarch
 Requires(pre):  uyuni-base-common
 BuildRequires:  gettext
-BuildRequires:  nodejs-packaging
 BuildRequires:  uyuni-base-common
 BuildRequires:  perl(ExtUtils::MakeMaker)
 


mcalmer (approved):

SchoolGuy (approved):
We should definitly readd this once we have the scripts fixed. However according to https://build.opensuse.org/package/view_file/devel:languages:nodejs/nodejs-packaging/nodejs-packaging.spec?expand=1 this doesn't even seem to provide anything to us that is important atm. Thus lgtm.

juliogonzalez (approved):

PR #5561, opened 2022-06-10
Add workspace configuration for VS Code

renner:
## What does this PR change?

This patch adds workspace configuration for developing Uyuni with VS Code, to be used together with [provided instructions](https://github.com/uyuni-project/uyuni/wiki/VS-Code-specific-development-instructions). The goal of these changes is to make it as easy as possible to get started in VSCode for:

- Install recommended extensions
- Build the java code in the IDE
- Run tests from within the IDE
- Attach the debugger to tomcat, taskomatic or the search server

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- Documentation is here: https://github.com/uyuni-project/uyuni/wiki/VS-Code-specific-development-instructions

- [X] **DONE**

## Test coverage

- No tests are needed, only configuration is added.

- [X] **DONE**

## Links

No further links.

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


avshiliaev:
Everything looks cool and it certainly works. Thank you, cause I'm also considering to switch over to the VSCode. :smiley:

java/.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "java",
+            "name": "Debug Uyuni",
+            "request": "attach",
+            "hostName": "server.tf.local",
```
aaannz, line 8: Thing is that this line will make my config file permanently modified. I do not use local sumaform, but on remote server. I am not sure here, but is there any support for variables in vscode?

java/.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
...
+            "port": 8000
+        },
+        {
+            "type": "java",
+            "name": "Debug Taskomatic",
+            "request": "attach",
+            "hostName": "server.tf.local",
```
aaannz, line 15: Same here

java/code/src/com/redhat/rhn/common/conf/Config.java
```diff
@@ -483,7 +483,7 @@ public void parseFiles() {
                 props.load(new StringReader(configString.replace("\\", "\\\\")));
             }
             catch (IOException e) {
-                logger.error("Could not parse file {}", curr, e);
+                logger.error("Could not parse file {}", curr);
```
aaannz, line 486: Wouldn't it be better to show the error?
```suggestion
                logger.error("Could not parse file {}: {}", curr, e);
```

java/.vscode/launch.json
renner: Thanks, good point, I am trying to find a solution using variables so users can override the default.

java/code/src/com/redhat/rhn/common/conf/Config.java
renner: Right, I removed the error from the log, because in a local testing context this error is expected. So I wanted to hide the long stacktrace from the console for each local test run from the IDE. Maybe we could use `logger.error("Could not parse file {}: {}", curr, e.getMessage());` at least, will try it out.

renner:
Reopening this to eventually merge it.

java/.vscode/launch.json
renner: I couldn't find a good solution to extract only the `hostNames` to another config file (this is already a config file), so the best idea that I have here is to maybe just provide a template file that needs to be copied before usage? `java/.vscode/launch.json` could be added to `.gitignore` then.

nodeg:
This PR should be rebased properly so it only shows the necessary changes.

java/.vscode/launch.json
renner: Changed `launch.json` to a template file (`launch.json.template`), which needs to be copied to `launch.json`.

java/code/src/com/redhat/rhn/common/conf/Config.java
renner: Removed this part of the patch for now, I also have not seen the respective stacktrace anymore recently.

mackdk (approved):
Looks good to me. The code is compiled correctly by VSCode and the tests run fine.  A couple of remarks:

1. The template approach is ok-ish. The alternative would be to [use input variables ](https://code.visualstudio.com/docs/editor/variables-reference#_input-variables). You can find [an example in my gist](https://gist.github.com/mackdk/e8c5e96ff46e9b7c7cb05a8315bb799f#file-launch-json),  but it's not optimal anyway: with my set up, the user will have to type everytime the local server name if different from `server.tf.local` because VSCode does not remember the last used value out of the box. I used in the past `pickString` to build selector with the servers I have locally. Anyway, both options would still require tweaking the `launch.json` file, so the template is still the best option. The only way to improve would be to experiment with extensions like [Command Variable](https://marketplace.visualstudio.com/items?itemName=rioj7.command-variable) and [Task Shell Input](https://marketplace.visualstudio.com/items?itemName=augustocdias.tasks-shell-input) which extends the capabilities of input variables, but given the low usage of VSCode for java I doubt it's worth the effort (I myself don't use it as my main IDE).
2. the extension "Ant target runner" never works for me. It just loads forever never showing the targets list and always throws a file not found error (even though that's not the real issue, it's more the parsing of the target dependencies).
3. Since Ant target runner doesn't work for me, I set up a [`task.json` file with the common actions](https://gist.github.com/mackdk/e8c5e96ff46e9b7c7cb05a8315bb799f#file-tasks-json) like clean, refresh dependencies, running test database, deploy and so on. Deploy of course requires the target server, thus a similar to the `launch.json`.

PR #5570, opened 2022-06-16
bootstrap repo: set optional package

mbussolotto:
## What does this PR change?

Currently the bootrap repo contains some packages from Uyuni Client tools channel: since these packages are just used for the traditional stack and in some scenario the Uyuni Client tools channel is not synced (e.g. when just a uyuni proxy is connected to the uyuni server), they should be marked as optional.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: no documentation present for describing packages on bootstap repo.

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


juliogonzalez:
Also adding mc to the review list.

mcalmer:
I think the packages should not be optional for SUMA as SUMA Proxy should have the traditional packages available.
I think the problem is only on "Uyuni Proxy", right?

So I would mark only the packages as optional which are used for a "Uyuni Proxy". 
I think this is the "PKGLIST15_TRAD" list.

juliogonzalez:
> I think the packages should not be optional for SUMA as SUMA Proxy should have the traditional packages available. I think the problem is only on "Uyuni Proxy", right?
> 
> So I would mark only the packages as optional which are used for a "Uyuni Proxy". I think this is the "PKGLIST15_TRAD" list.

True, but do we still support traditional proxies on 4.3?

mcalmer:
> > I think the packages should not be optional for SUMA as SUMA Proxy should have the traditional packages available. I think the problem is only on "Uyuni Proxy", right?
> > So I would mark only the packages as optional which are used for a "Uyuni Proxy". I think this is the "PKGLIST15_TRAD" list.
> 
> True, but do we still support traditional proxies on 4.3?

Yes, I am not aware there we forbid that.

mcalmer (approved):
I hope this is enough to fix the problem.

mcalmer:
> ok, now the changes are just in `PKGLIST15_TRAD`, but it is also used by SLE/SUMA ..should I create another category just for uyuni?

I think this is not easy possible. We have not really the possibility to differ between SUMA and Uyuni at this point.
I think you need the fix "now". Maybe we rollback this after we branched 4.3 and remove the optional again?

And short after this, we can remove all the traditional packages from the bootstrap repos as Uyuni / Master and SUMA 4.4 will not support it anymore.

PR #5577, opened 2022-06-20
Fix bootstrapping for Ubuntu 18.04 with classic Salt package (bsc#1200707)

meaksh:
## What does this PR change?

This PR fixes an issue on bootstrapping with classic Salt 3004 package for Ubuntu 18.04 caused by missing `python3-contextvars` and `python3-immutables`not being installed during bootstrapping.

These packages are set as "Recommends" in the Salt DEB package, because we have only a single DEB package we build for all OSes based distros. These two dependencies are supposed to get installed automatically during bootstrapping (as recommends are installed by default), but this is not happening because bootstrapping explicitely disabling adding recommended packages.

In the end, since we cannot add these two packages as hard dependencies for Salt in DEB based distros as those packages are not existing in, i.a., Ubuntu 20.04, what we are doing with this PR is to make boostrapping to also explicitely install the two dependencies if we are installing `salt-minion` (classic package) in a DEB based distro with Python version < 3.7 (where these deps are expected).

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/18177

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
@@ -144,6 +144,9 @@ salt-minion-package:
     - require:
       - file: bootstrap_repo
 
+{# These two dependencies are only needed on DEB based distros with Python version < 3.7 #}
+{# We cannot make these packages as hard depedendencies for Salt package because this is only needed for Ubuntu 18.04 #}
+{# and we only maintain a single DEB package for all DEB based distros #}
 {% if salt_minion_name == 'salt-minion' and grains['os_family'] == 'Debian' and grains['pythonversion'][0] >= 3 and grains['pythonversion'][1] < 7 %}
 salt-install-contextvars:
   pkg.installed:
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -144,6 +144,19 @@ salt-minion-package:
     - require:
       - file: bootstrap_repo
 
+{# These two dependencies are only needed on DEB based distros with Python version < 3.7 #}
+{# We cannot make these packages as hard depedendencies for Salt package because this is only needed for Ubuntu 18.04 #}
+{# and we only maintain a single DEB package for all DEB based distros #}
+{% if salt_minion_name == 'salt-minion' and grains['os_family'] == 'Debian' and grains['pythonversion'][0] >= 3 and grains['pythonversion'][1] < 7 %}
+salt-install-contextvars:
```
agraul, line 151: nitpick: state ids should be in passive

agraul (approved):
Haven't tried this on Ubuntu, but looks like it should work.

PR #5584, opened 2022-06-21
QE: Fixup for checking open/closed ports

nodeg:
## What does this PR change?

`ss` gives the better output than `lsof`, especially when we have a small port number in a higher one:

```bash
head-srv:~ # lsof -i -P -n | grep 3333
postgres  13333   postgres   10u  IPv6 174132      0t0  UDP [::1]:50162->[::1]:50162 
postgres  13333   postgres   11u  IPv4 593821      0t0  TCP 127.0.0.1:5432->127.0.0.1:34556 (ESTABLISHED)


head-srv:~ # ss --listening --numeric | grep 3333

```

- CI PR run: https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-prs-ci-tests/1460/
- Issue: https://github.com/SUSE/spacewalk/issues/18314

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Cucumber tests were edited

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/18314
Needs: https://github.com/SUSE/spacewalk/issues/18459
Manager 4.1
Manager 4.2

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -1332,11 +1332,11 @@ When(/^I schedule a task to update ReportDB$/) do
 end
 
 Then(/^port "([^"]*)" should be (open|closed)$/) do |port, selection|
-  output, _code = $server.run("lsof -i -P -n  | grep #{port}  || echo 'closed'", check_errors: false)
-  case selection
-  when 'open'
-    raise "Port '#{port}' not open although it should be!" if output == "closed\n"
-  when 'closed'
-    raise "Port '#{port}' open although it should not be!" if output != "closed\n"
+  _output, code = $server.run("ss --listening --numeric | grep :#{port}", check_errors: false, verbose: true)
+  port_opened = code.zero?
+  if selection == 'closed'
+    raise "Port '#{port}' open although it should not be!" if port_opened
+  else
+    raise "Port '#{port}' not open although it should be!" unless port_opened
   end
 end


nodeg:
The fix did not work:
image

nodeg:
Using the exit code did not work either. I will do some more debugging and testing.
image

nodeg:
Some debug output from/for me:
```bash
    And I should see a "Disable" button                                      # features/step_definitions/navigation_steps.rb:662
lsof -i -P -n | grep 3333 returned status code = 1.
Output:
Code: 1
port_opened: false
    And port "3333" should be closed                                         # features/step_definitions/common_steps.rb:1346
Port '3333' open although it should not be! (RuntimeError)
      ./features/step_definitions/common_steps.rb:1364:in `/^port "([^"]*)" should be (open|closed)$/'
features/secondary/srv_monitoring.feature:31:in `port "3333" should be closed'
    And port "3334" should be closed                                         # features/step_definitions/common_steps.rb:1346    
```

nodeg:
The CI somehow looks better now and only fails once:
image
image

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1332,7 +1332,7 @@ def server_secret
 end
 
 Then(/^port "([^"]*)" should be (open|closed)$/) do |port, selection|
-  output, _code = $server.run("lsof -i -P -n  | grep #{port}  || echo 'closed'", check_errors: false)
+  output, _code = $server.run("ss --listening --numeric | grep #{port}  || echo 'closed'", check_errors: false, verbose: true)
```
witekest, line 1335: That should work fine. Filtering the entry in options is probably even more reliable.

```suggestion
  output, _code = $server.run("ss --listening --numeric -H src :#{port}  || echo 'closed'", check_errors: false, verbose: true)
```

nodeg: As you mentioned in Slack, we need some command after `ss` since it always returns `0` as [return value](https://unix.stackexchange.com/questions/384920/ss-exit-code-always-0).
I will add a grep after the command again.

nodeg:
This should work as intended after https://github.com/SUSE/spacewalk/issues/18459 gets addressed according to @witekest.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1332,11 +1332,11 @@ def server_secret
 end
 
 Then(/^port "([^"]*)" should be (open|closed)$/) do |port, selection|
-  output, _code = $server.run("lsof -i -P -n  | grep #{port}  || echo 'closed'", check_errors: false)
-  case selection
-  when 'open'
-    raise "Port '#{port}' not open although it should be!" if output == "closed\n"
-  when 'closed'
-    raise "Port '#{port}' open although it should not be!" if output != "closed\n"
+  _output, code = $server.run("ss --listening --numeric | grep :#{port}", check_errors: false, verbose: true)
```
vandabarata, line 1335: is there any reason why we need the `:` before the port being grepped?

witekest: The intention is to avoid catching false positive port 13333 when grepping 3333.
Even more bulletproof would be:
`ss --listening --numeric srv :#{port} | grep #{port}`

nodeg: `ss --listening --numeric srv :#{port} | grep #{port}` does not work but `ss --listening --numeric -H src :#{port} | grep #{port}` I guess?

```bash
head-srv:~ # ss --listening --numeric srv :5556 | grep 5556
Error: an inet prefix is expected rather than "srv".
Cannot parse dst/src address.

head-srv:~ # ss --listening --numeric -H src :5556 | grep 5556
tcp LISTEN 0      3      *:5556 *:*
```
But yes, the semicolon is intentional.

PR #5589, opened 2022-06-22
Modernize tested channels

Bischoff:
## What does this PR change?

Fix CI reposync tests after we removed outdated repositories from our mirrors.

```
SLES 11 SP4 -> SLES 12 SP3 
SLES 11 SP4 SDK -> SLES 12 SP3 live patching
SLES 12 SP4 -> SLES 12 SP5
Proxy 4.0 -> Proxy 4.1
```
This is intentionally **unneeded stuff** for the other tests, where we will kill the reposync (with the exception of proxy 4.1 on 4.1 branch).

## Changelogs

- [x] No changelog needed


testsuite/features/reposync/srv_sync_channels.feature
```diff
@@ -1,75 +1,69 @@
...
+    And I should get "[ ] SUSE Manager Proxy 4.1 x86_64"
 
-@scc_credentials
 @susemanager
   Scenario: List all products for SUSE Manager
     When I execute mgr-sync "list products --expand"
-    Then I should get "[ ] SUSE Linux Enterprise Server 12 SP4 x86_64"
-    And I should get "[ ] SUSE Manager Proxy 4.0 x86_64"
+    Then I should get "[ ] SUSE Linux Enterprise Server 12 SP5 x86_64"
+    And I should get "[ ] SUSE Manager Proxy 4.1 x86_64"
```
srbarrios, line 36:
```suggestion
    And I should get "[ ] SUSE Manager Proxy 4.3 x86_64"
```

testsuite/features/reposync/srv_sync_channels.feature
```diff
@@ -1,75 +1,69 @@
...
+    And I shouldn't get "sles12-sp3-installer"
 
-@scc_credentials
 @susemanager
   Scenario: List products
     When I execute mgr-sync "list products"
-    Then I should get "[ ] SUSE Linux Enterprise Server 12 SP4 x86_64"
-    And I should get "[ ] SUSE Manager Proxy 4.0 x86_64"
+    Then I should get "[ ] SUSE Linux Enterprise Server 12 SP5 x86_64"
+    And I should get "[ ] SUSE Manager Proxy 4.1 x86_64"
```
srbarrios, line 30:
```suggestion
    And I should get "[ ] SUSE Manager Proxy 4.3 x86_64"
```

srbarrios:
I understand the problem of proxy is independent, but why don't just use the right value here instead of adding again a old value?
Why 4.1? We even plan to stop 4.1 CI very soon, and most of the content will also go away.
We should "at least" use 4.2

testsuite/features/reposync/srv_sync_channels.feature
Bischoff: No, intentionally using old stuff that are meant to be killed, see my explanations

Let's reposync 4.3 separately.

Bischoff:
> I understand the problem of proxy is independent, but why don't just use the right value here instead of adding again a old value? Why 4.1? We even plan to stop 4.1 CI very soon, and most of the content will also go away. We should "at least" use 4.2

Because 4.2 is used "for real" in 4.2 branch.

I hope we can use real reposyncs for good before I have to bump this once again. For now let's keep problems separate. The puprose of this PR is mostly to fix a breakage in emergency.

testsuite/features/reposync/srv_sync_channels.feature
Bischoff: Same

Bischoff:
Green on 4.2 branch.

ktsamis (approved):

nodeg (approved):

PR #5598, opened 2022-06-23
Fixed unexpected unauthorized when using HTTP API

wweellddeerr:
## What does this PR change?

It adds a `commitTransaction` call just after saving the WebSession during user authentication.

Without this, in some random situations, [this call](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/session/WebSessionFactory.java#L63) to get WebSession by its id returned null, even this WebSession being saved just some fractions of second before.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/18176

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/session/SessionManager.java
+++ b/java/code/src/com/redhat/rhn/manager/session/SessionManager.java
@@ -85,6 +85,7 @@ public class SessionManager extends BaseManager {
 
         s.setExpires(TimeUtils.currentTimeSeconds() + duration);
         WebSessionFactory.save(s);
+        WebSessionFactory.getSession().flush();
         return s;
     }
 


rjmateus (approved):
lgtm
good catch

cbbayburt:
@wweellddeerr the failing unit tests might be relevant. can you please check those?

cbbayburt (approved):

rjmateus:
With the side effects a commit can have in the code block I'm not sure if we should do it this way.
I know you tried a flush. Did you try a flush and relead (`HibernateFactory.reload(s);`)?
If that not works, maybe start a transaction so save the object and then commit just that transaction

jordimassaguerpla:
Hi! Any progress on this one? Thanks

wweellddeerr:
> @wweellddeerr the failing unit tests might be relevant. can you please check those?

@cbbayburt thanks for pointing out the unit tests were failing on CI, it was a side effect of doing a `WebSessionFactory.commitTransaction()` (my first proposed solution). The tests were OK running isolated on my local environment, but, when running all of them, the commit causes some data to persist from one test to another causing the failures.

wweellddeerr:
> With the side effects a commit can have in the code block I'm not sure if we should do it this way. I know you tried a flush. Did you try a flush and relead (`HibernateFactory.reload(s);`)? If that not works, maybe start a transaction so save the object and then commit just that transaction

@rjmateus thank you for your review and suggestions. Calling `HibernateFactory.reload(s);` after a `flush()` causes a NonUniqueObjectException and I wasn't able to use a new transaction as approach.

My idea now is to merge the changes with the call to `WebSessionFactory.getSession().flush();` since it seems be a good improvement, considering some tests I'm doing: the behavior before the patch is to have from 1 to 5 unexpected `401` responses for each block of **500** requests; after the patch I got from 0 to 2 unexpected `401` responses for each block of **10.000** requests.

wweellddeerr:
> Hi! Any progress on this one? Thanks

@jordimassaguerpla I plan to merge this pull request ASAP and I think we need to observe how it affects the testsuite considering that, according with some experiments I did, it doesn't solve completely the problem but reduces significantly the chances of an unexpected 401 status as response.

PR #5600, opened 2022-06-23
Turned API information endpoints public

wweellddeerr:
## What does this PR change?

It turns the following API information endpoints public:
- /api/getVersion
- /api/systemVersion

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17951

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/frontend/security/PxtAuthenticationService.java
```diff
@@ -64,6 +64,8 @@ public class PxtAuthenticationService extends BaseAuthenticationService {
         set.add("/img/favicon.ico");
         set.add("/rhn/common/DownloadFile");
         set.add("/rhn/manager/sso");
+        set.add("/rhn/manager/api/api/getVersion");
+        set.add("/rhn/manager/api/api/systemVersion");
         // password-reset-link destination
         set.add("/rhn/ResetLink");
         set.add("/rhn/ResetPasswordSubmit");
```
cbbayburt, line 71: For better encapsulation, I think we should get this list from the `HttpApiRegistry` via a static method call. Therefore we can make further changes in the API without having to change this class.

Moreover, I was thinking that if we need to extend this list in the future, we can implement a logic to automatically determine which endpoints are public by looking at each API method either for an annotation or checking if a `User` parameter exists or not (for now, it seems like overkill though).

A third benefit of my suggestion is we can call this `getUnauthenticatedRoutes` from the API doc generator code and add a note to those methods' documentation.
 
```suggestion
        // password-reset-link destination
        set.add("/rhn/ResetLink");
        set.add("/rhn/ResetPasswordSubmit");
        
        // HTTP API public endpoints
        set.addAll(HttpApiRegistry.getUnautenticatedRoutes());
```

cbbayburt (changes requested):

java/code/src/com/redhat/rhn/frontend/security/PxtAuthenticationService.java
wweellddeerr: I fully agree with your considerations and have changed the code accordingly.

cbbayburt (approved):

PR #5614, opened 2022-06-29
Fixed missing "options" in args, for spacecmd/src/spacecmd/system.py

slsnow:
With the "options" value missing, `spacecmd -- system_listmigrationtargets` fails with "ERROR: too many values to unpack (expected 1)".
https://bugzilla.suse.com/show_bug.cgi?id=1201003

## What does this PR change?
- Fixes that command output.

- That command is necessary for this documented workflow to work properly:
https://github.com/uyuni-project/uyuni-docs/blob/master/modules/client-configuration/pages/client-upgrades-product-migration.adoc

## GUI diff
- No difference.

## Documentation
- No documentation needed: 

## Test
- Unsure about tests

## Links
- https://bugzilla.suse.com/show_bug.cgi?id=1201003

## Changelogs
- Not sure if changelog is needed


mbussolotto:
Fixes: https://github.com/SUSE/spacewalk/issues/18272

mbussolotto:
> Not sure if changelog is needed

Yes, they are always required. You should add a line on top of `spacecmd/spacecmd.changes` like:
```
- Fix missing argument on system_listmigrationtargets (bsc#1201003)
```
Then I think the PR can be approved

slsnow (commented):
Added the requested changelog

spacecmd/spacecmd.changes
```diff
@@ -1,3 +1,7 @@
+Tue July 19 09:20 AM CEST 2022 - ssnow@suse.com
```
mbussolotto, line 1: Line 1 and 2 should be removed (they're added when the package is released). Line 3 and 4 are enough.

mbussolotto (changes requested):

spacecmd/spacecmd.changes
mbussolotto: ok, I was able to make the change by myself :). Before merging, please wait python team or @meaksh review

agraul (approved):

PR #5635, opened 2022-07-06
Revert "QE: Enable temporarily disabled features again"

nodeg:
Reverts uyuni-project/uyuni#5625. This did cause too much trouble in the testsuite. We have to enable and investigate this later on.


vandabarata (approved):

Bischoff:
> Reverts #5625. This did cause too much trouble in the testsuite. We have to enable and investigate this later on.

... or simply kill every traditional client code :smile:  - as soon as it's not supported anymore.

Sorry, I missed that those two features were about traditional clients.

PR #5643, opened 2022-07-07
Add option to use a remote PostgreSQL DB

KeithMnemonic:
## What does this PR change?

This PR adds the ability to create the susemanager and reportdb databases on a remote postgres server or a postgres instance running in AWS RDS.  Additional configuration variables have been added to provide the "remote postgres" credentials as there is not a "local postgres user passthough authentication" mechanism when using a remote instance.  In addition there is also an EXTERNAL_PROVIDER option that allows special handling for SUPERUSER role as AWS does not provide that. Instead there is an rds_superuser role.  

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- No documentation needed: only internal and user invisible changes.  There will need to be documentation but not sure who/how to handle that until this PR is reviewed

- [ ] **DONE**

## Test coverage
- No tests:  Tests can only be added once sumaform nd the are able to deploy the proper type instance. At that time, no new functional tests are needed but rather the same test suite just using a remote DB instead of the local DB.

- [ ] **DONE**

## Links

In order for AWS to work correctly this PR first needs merged, https://github.com/uyuni-project/uyuni/pull/5633

Fixes #

https://github.com/SUSE/spacewalk/issues/15271

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


spacewalk/setup/bin/spacewalk-setup
```diff
@@ -854,6 +854,10 @@ sub populate_initial_configs {
      db_host => $answers->{'db-host'},
      db_port => $answers->{'db-port'},
      db_ssl_enabled => $answers->{'db-ssl-enabled'},
+     external_db => $answers->{'external-db'},
+     external_db_user => $answers->{'external-db-user'},
+     external_db_password => $answers->{'external-db-password'},
```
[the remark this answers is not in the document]
mcalmer: After reading a bit: Should this be the RDS "Master DB User/Password" ?
This cannot be used as the DB user? Or should not be used?
Need background infos to understand this.

spacewalk/setup/lib/Spacewalk/Setup.pm
```diff
@@ -880,11 +880,41 @@ sub postgresql_reportdb_setup {
...
+                      push @cmd, "--external-provider", $answers->{'external-db-provider'};
+              }
+
+              print_progress(-init_message => "*** Progress: #",
+                      -log_file_name => DB_INSTALL_LOG_FILE,
+                      -log_file_size => DB_INSTALL_LOG_SIZE,
+                      -err_message => "Could not install report database.\n",
+                      -err_code => 15,
+                      -system_opts => \@cmd);
+            }
```
mcalmer, line 917: All this duplicate a lot of code. Here in the setup and also in the tool itself.
Can't this be joined? 
- keep just `create` and `remove` command.
- if you mean external DB, provide the extra options. 
- In the tool provide for every command the local and the external variant and select it depending on what options were provided.

spacewalk/setup/lib/Spacewalk/Setup.pm
```diff
@@ -1119,10 +1149,19 @@ sub postgresql_clear_db {
...
+                    "--external-host", $answers->{'report-db-host'},
+                    "--external-user", $answers->{'external-db-user'},
+                    "--external-password", $answers->{'external-db-password'});
+          return 1;         
+        }
+        else {  
+          system_debug('/usr/bin/uyuni-setup-reportdb', 'remove',
+              '--db='.$answers->{'report-db-name'}, '--user='.$answers->{'report-db-user'});
+          return 1;
+        }
```
mcalmer, line 1164: Example to combine this:
```
@cmd = ('/usr/bin/uyuni-setup-reportdb', 'remove', '--db='.$answers->{'report-db-name'},
   '--user='.$answers->{'report-db-user'});
if ($answers->{'external-db'}) {
    push @cmd, "--external-host", $answers->{'report-db-host'},
                    "--external-user", $answers->{'external-db-user'},
                    "--external-password", $answers->{'external-db-password'};
}
system_debug(@cmd);
return 1;
```
(not tested)

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
```diff
@@ -13,6 +13,10 @@ help() {
     echo "    --password <password>        Password for the database user"
     echo "    --autogenpw                  Auto generate a password for the database user"
     echo "    --standalone                 Configure the database server independent of Uyuni or SUSE Manager Server Database"
+    echo "    --external-host              The database host when configuring the report database using an external database server"
```
mcalmer, line 16: I think you can compute this. If external-user is set, you should assume that things are external.

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
```diff
@@ -13,6 +13,10 @@ help() {
...
+    echo "    --external-user              The user when configuring the report database using an external database server"
+    echo "    --external-password          The password when configuring the report database using an external database server"
+    echo "    --external-provider          The Cloud provider that is hosting the external report database"
```
mcalmer, line 19: Do you expect differences between aws, google and azure ?

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
```diff
@@ -294,16 +298,81 @@ EOF
 
 }
 
+create_remote() {
```
mcalmer, line 301: Can we reuse some parts? 
I think you copied  things from `create()`. So maybe it is possible to call this from "create" to prevent duplicate code.

Otherwise we forget about it and change only 1 place and forget the other.

susemanager/bin/mgr-setup
```diff
@@ -199,75 +199,108 @@ cleanup_hostname() {
...
-            rm -Rf ${DATADIR}
-            POSTGRES_LANG=en_US.UTF-8
-
-            # Define LC_CTYPE in postgres user profile. (Used during any DB creation)
-            PGHOME=$(getent passwd postgres | awk -F: '{print $6}')
-            . $PGHOME/.i18n 2>/dev/null
-            if [ -z $LC_CTYPE ]; then
-                grep "^LC_CTYPE" $PGHOME/.i18n > /dev/null 2>&1
+    if [ $LOCAL_DB = 1 ]; then
+        DATADIR="/var/lib/pgsql/data"
```
mcalmer, line 229: My feeling is, that we should re-think the DB setup implementation in `mgr-setup`.
We have now an external tool for the reportdb and also for the main db there exists an external tool (spacewalk-setup-postgresql) which we do not use.

Maybe we should combine this all in 1 setup tool.
This would make the code much cleaner.
CC @admd

mcalmer (commented):
A first quick review with a lot of questions to understand the limitations of RDS.

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: well my implementation works with a remote postgres running as well as an RDS instance. So in this case, the password is for the "postgres" used to connect to the remote instance. If we do not need the option of a remote postgres and only RDS I can clean some of this up. @moio What is you idea in terms of the offering?

spacewalk/setup/lib/Spacewalk/Setup.pm
KeithMnemonic: it is a lot of code. The other option is add a huge if block to the create in uyuni-setuo-reportdb.  You can see here https://github.com/uyuni-project/uyuni/blob/master/spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb#L119  this function is 99% seting up the local postgres. I am fine either way

spacewalk/setup/bin/spacewalk-setup
```diff
@@ -854,6 +854,10 @@ sub populate_initial_configs {
...
+     external_db_ca_cert => $answers->{'external-db-ca-cert'},
```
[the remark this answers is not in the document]
KeithMnemonic: let me fix this up

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
KeithMnemonic: well ATM, no since google/azure do not allow a superuser equivalent like AWS.  if that would ever change then yes i would expect the role name to not be rds_superuser so each provider would need special handling

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: To follow up, I though we were considering also supporting a non managed remote postgres in the cloud.  I.E. suma running on one instance and an installed postgres on another cloud instance.

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: so part of the problem is this code

`report-db-ca-cert=$CA_PATH/LOCAL-RHN-ORG-TRUSTED-SSL-CERT`

The reason being that if as you said "once this becomes an external database, we should use ssl"for both susemanager and reportdb", then this value here can not be hardcoded.

susemanager/bin/mgr-setup
jeremy-moffitt: clarification question... are you suggesting this as a larger scope change for this PR or something that should be on the roadmap in the longer term?

spacewalk/setup/bin/spacewalk-setup
```diff
@@ -854,6 +854,10 @@ sub populate_initial_configs {
      db_host => $answers->{'db-host'},
      db_port => $answers->{'db-port'},
      db_ssl_enabled => $answers->{'db-ssl-enabled'},
+     external_db => $answers->{'external-db'},
```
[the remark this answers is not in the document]
KeithMnemonic: I think migration https://github.com/uyuni-project/uyuni/blob/master/susemanager/bin/mgr-setup#L657 will not work in this first iteration

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: @moio @rjmateus It would be good to clarify if we are targeting only RDS or also a non RDS remote postgres. Comments please?

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: also checking db_host != "localhost" or "hostname -f"  is not sufficient if they do not have a full dns name or what if they us the ip address.  Since this gets passed in via the setup-env.sh file, we would need more code to test all of this. seems setting a variable is much easier given how tangled all of this code already is.

susemanager/bin/mgr-setup
```diff
@@ -199,75 +199,108 @@ cleanup_hostname() {
 
 exists_db() {
     PGNAME=$1
-    EXISTS=$(runuser - postgres -c 'psql -t -c "select datname from pg_database where datname='"'$PGNAME'"';"')
+    if [ $LOCAL_DB = 1 ]; then
+        EXISTS=$(runuser - postgres -c 'psql -t -c "select datname from pg_database where datname='"'$PGNAME'"';"')
+    else
+        EXISTS=$(echo "select datname from pg_database where datname='$PGNAME';" | (export PGPASSWORD=$EXTERNAL_DB_PASS; exec psql -t -U postgres -h $MANAGER_DB_HOST))
```
maximenoel8, line 205: I think postgres user should be replace by $EXTERNAL_DB_USER other it's force the RDS to be created with postgres user and $EXTERNAL_DB_USER = postgres .

susemanager/bin/mgr-setup
```diff
@@ -199,75 +199,108 @@ cleanup_hostname() {
...
     else
         return 1
     fi
 }
 
+exists_user() {
+    if [  $LOCAL_DB = 1 ] ; then
+        EXISTS=$($RUNUSER - postgres -c 'psql -t -c "select usename from pg_user where usename='"'$MANAGER_USER'"';"')
+    else
+        EXISTS=$(echo "select usename from pg_user where usename='$MANAGER_USER';" | (export PGPASSWORD=$EXTERNAL_DB_PASS; exec psql -t -U postgres -h $MANAGER_DB_HOST))
```
maximenoel8, line 218: Same than before : 
I think postgres user should be replace by $EXTERNAL_DB_USER other it's force the RDS to be created with postgres user and $EXTERNAL_DB_USER = postgres .

maximenoel8 (changes requested):

susemanager/bin/mgr-setup
```diff
@@ -199,75 +199,108 @@ cleanup_hostname() {
...
+            echo "Database exists. Preparing for resetup. All data will be removed."
         fi
     else
-        # Define POSTGRES_LANG in system wide config. (Used during any DB creation)
-        . /etc/sysconfig/postgresql
-        if [ -z $POSTGRES_LANG ]; then
-            grep "^POSTGRES_LANG" /etc/sysconfig/postgresql > /dev/null 2>&1
-            if [ $? = 0 ]; then
-                sed -i -e "s/^POSTGRES_LANG.*$/POSTGRES_LANG=\"en_US.UTF-8\"/" /etc/sysconfig/postgresql
+        if ! exists_db $MANAGER_DB_NAME; then
```
maximenoel8, line 290: Maybe we should have the user creation before the database creation and replace $EXTERNAL_DB_USER by $MANAGER_USER.

maximenoel8 (changes requested):

susemanager/bin/mgr-setup
```diff
@@ -199,75 +199,108 @@ cleanup_hostname() {
...
+        fi
+        if ! exists_user $MANAGER_USER; then
+            echo  $EXTERNAL_DB_PROVIDER
+            if [[ -n $EXTERNAL_DB_PROVIDER && "${EXTERNAL_DB_PROVIDER,,}" == "aws" ]] ; then
+                echo "CREATE ROLE $MANAGER_USER PASSWORD '$MANAGER_PASS'  NOCREATEDB NOCREATEROLE INHERIT LOGIN;" | ( export PGPASSWORD=$EXTERNAL_DB_PASS; exec psql -h $MANAGER_DB_HOST -U $EXTERNAL_DB_USER)
+                echo "GRANT rds_superuser to $MANAGER_USER;" | ( export PGPASSWORD=$EXTERNAL_DB_PASS; exec psql -h $MANAGER_DB_HOST -U $EXTERNAL_DB_USER)
+                echo "GRANT create on database $MANAGER_DB_NAME to $MANAGER_USER;" | ( export PGPASSWORD=$EXTERNAL_DB_PASS; exec psql -h $MANAGER_DB_HOST -U $EXTERNAL_DB_USER)
             else
-                echo "POSTGRES_LANG=\"en_US.UTF-8\"" >> /etc/sysconfig/postgresql
+                echo "CREATE ROLE $MANAGER_USER PASSWORD '$MANAGER_PASS' SUPERUSER NOCREATEDB NOCREATEROLE INHERIT LOGIN;" | ( export PGPASSWORD=$EXTERNAL_DB_PASS; exec psql -h $MANAGER_DB_HOST -U $EXTERNAL_DB_USER)
```
maximenoel8, line 300: Maybe we should have the user creation before the database creation and replace $EXTERNAL_DB_USER by $MANAGER_USER.

maximenoel8 (changes requested):

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
admd: @moio is off this week so I will jump in here. For now, the focus is RDS but there is no reason not to expand supportability to other remote Postgres. In fact, recently there was a request to support Azure Postgres as well.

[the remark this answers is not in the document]
KeithMnemonic: @admd so this is the tricky issue. For Google and Azure, the managed databases do not work as the do not offer an alternative "SUPERUSER" as AWS does. If  a customer wanted to run in zure of GCP they need to deploy postgres in a VM/Container.  This is why my PR also offers a "remote postgres" option but it seems there is some debate still needed to see if we want to keep it this way.

susemanager/bin/mgr-setup
KeithMnemonic: good catch @maximenoel8  I thought in some places I did this but let me make sure and push an update. Thank you.

susemanager/bin/mgr-setup
KeithMnemonic: I think we still need two users here. one (typically postgres) that is able to connect to the deployed remote postgres instance with the correct  role permissions that allows the creation of the susemanager database and the spacewalk role  (with SUPERUSER or the equivalent) .  perhaps I am not fully understanding your comment. Could you please provide more details?

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
rjmateus: If it's working with any remote Database (with power user access) then I think is fine for me. Another question is if we should give official support for another external database besides AWS RDS.
At this point I think we should only officially support RDS, and then we can see if we should open the spectrum, and support any remote database.
Just for completeness, we have one (special) customer that is using a remote database.

susemanager/bin/mgr-setup
KeithMnemonic: fixed with my latest push

susemanager/bin/mgr-setup
KeithMnemonic: fixed

spacewalk/setup/lib/Spacewalk/Setup.pm
```diff
@@ -1119,10 +1149,19 @@ sub postgresql_clear_db {
...
         my $answers = shift;
-
-        system_debug('/usr/bin/uyuni-setup-reportdb', 'remove',
-            '--db='.$answers->{'report-db-name'}, '--user='.$answers->{'report-db-user'});
-        return 1;
+        if ($answers->{'externaldb'}) {
+          system_debug('/usr/bin/uyuni-setup-reportdb', 'remove',
+              '--db='.$answers->{'report-db-name'}, '--user='.$answers->{'report-db-user'},
+                    "--externaldb-host", $answers->{'report-db-host'},
+                    "--externaldb-user", $answers->{'externaldb-admin-user'},
```
maximenoel8, line 1156: externaldb-user should be externaldb-admin-user

spacewalk/setup/lib/Spacewalk/Setup.pm
```diff
@@ -1119,10 +1149,19 @@ sub postgresql_clear_db {
...
+                    "--externaldb-password", $answers->{'externaldb-admin-password'});
```
maximenoel8, line 1157: externaldb-password should be externaldb-admin-password

maximenoel8 (changes requested):

spacewalk/setup/lib/Spacewalk/Setup.pm
KeithMnemonic: already fixed

spacewalk/setup/lib/Spacewalk/Setup.pm
KeithMnemonic: already fixed

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: I fixed this temporarily now, but someone may want to revisit this to also add an ip address check.

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: this is fixed now simply using three new variables.  the PROVIDER is needed as there is some special handling for AWS and this allows further expansion if we can ever get the this working for AZURE or GCE. 

EXTERNALDB_ADMIN_USER="postgres"
EXTERNALDB_ADMIN_PASS="susemanager"
EXTERNALDB_PROVIDER="AWS"

spacewalk/setup/lib/Spacewalk/Setup.pm
```diff
@@ -880,11 +880,41 @@ sub postgresql_reportdb_setup {
 	postgresql_drop_reportdb($answers);
     }
 
-    $ENV{PGSSLROOTCERT} = $answers->{'report-db-ca-cert'};
+    if ($answers->{'external-db'}) {
+        $ENV{PGSSLROOTCERT} = $answers->{'external-db-ca-cert'};
+    }
+    else {
+        $ENV{PGSSLROOTCERT} = $answers->{'report-db-ca-cert'};
+    }
```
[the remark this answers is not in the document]
KeithMnemonic: i fixed this to either use the provided value for report-db-ca-cert or if not set use 

`report-db-ca-cert=$CA_PATH/LOCAL-RHN-ORG-TRUSTED-SSL-CERT`
there is no need for external-db-ca-cert any longer

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
KeithMnemonic: yes this should be left in as long as we are considering any other option besides AWS. It seems from comments from 
@admd @rjmateus and @moio  they want to leave it open. I will mark this as resolved for now.

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
KeithMnemonic: yes but unless I am missing something the host is not passed in as it assumes localhost.  Somehow I need to get the "remote host" passed in.  Do you have a better suggestion @mcalmer

spacewalk/setup/lib/Spacewalk/Setup.pm
KeithMnemonic: Thank you for the suggestion. I will clean this up.

KeithMnemonic: fixed

spacewalk/setup/bin/spacewalk-setup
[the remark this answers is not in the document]
KeithMnemonic: fixed

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
KeithMnemonic: or is the point rather just rename it to --host?

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
KeithMnemonic: i put this all in create now and removed create_remote

spacewalk/setup/lib/Spacewalk/Setup.pm
KeithMnemonic: I cleaned this up in Setup.pm as well now.

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
KeithMnemonic: i have this working but using the host

maximenoel8 (approved):
LGTM after testing

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
```diff
@@ -516,9 +604,18 @@ while true ; do
...
+        EXTERNALDB=0 ;;
+    *)
+        EXTERNALDB=1 ;;
+esac
+
 case $1 in
         create) create
             ;;
+        create_remote) create_remote
+            ;;
```
mcalmer, line 618: Seems this is not implemented. The code was integrated into create.

susemanager/bin/mgr-setup
```diff
@@ -387,13 +422,18 @@ db-name=$MANAGER_DB_NAME
 db-host=$MANAGER_DB_HOST
 db-port=$MANAGER_DB_PORT
 db-protocol=$MANAGER_DB_PROTOCOL
+externaldb=$EXTERNALDB
+externaldb-admin-user=$EXTERNALDB_ADMIN_USER
+externaldb-admin-password=$EXTERNALDB_ADMIN_PASS
+db-ca-cert=$MANAGER_DB_CA_CERT
+report-db-ca-cert=$REPORT_DB_CA_CERT
```
mcalmer, line 429: Is it possible / should we support that we have 2 different CAs here?
What does AWS provide? Can you have 2 DBs and say you want to use the same CA (different server certs, but signed from the same CA)?
This would make some things a lot of easier.

susemanager/bin/mgr-setup
mcalmer: I think we need a refactoring, but not on this PR. Maybe we start this after this is in if time permits

mcalmer:
@KeithMnemonic I think you checked this already, but `uyuni-setup-reportdb-user` does not need to be ported as it work already plain over psql, right?

mcalmer:
Running our testsuite on this PR : https://ci.nue.suse.com/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/1500/

spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
KeithMnemonic: thanks for catching. i will remove this.

spacewalk/setup/lib/Spacewalk/Setup.pm
```diff
@@ -17,17 +17,7 @@ use Pod::Usage qw(pod2usage);
...
-    require Net::LibIDN2;
-    Net::LibIDN2->import();
-    sub idn_to_ascii { return Net::LibIDN2::idn2_lookup_u8( shift ) }
-    1;
-} or do {
-    my $error = $@;
-    require Net::LibIDN;
-    Net::LibIDN->import( qw(idn_to_ascii) );
-};
+use Net::LibIDN ();
```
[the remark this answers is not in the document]
KeithMnemonic: Hmm, I did not remove that intentionally , let me put it back. thanks for catching it.

susemanager/bin/mgr-setup
KeithMnemonic: If you are using RDS, it is the same root cert.  Looking at the existing code  you will see originally there were different certs used in Setup.pm

[1)](https://github.com/uyuni-project/uyuni/blob/master/spacewalk/setup/lib/Spacewalk/Setup.pm)  

`$ENV{PGSSLROOTCERT} = $answers->{'report-db-ca-cert'};`

[2)](https://github.com/uyuni-project/uyuni/blob/master/spacewalk/setup/lib/Spacewalk/Setup.pm#L849)

`$ENV{PGSSLROOTCERT} = $answers->{'db-ca-cert'};`

It seems in Setup.pm there exists some code for doing ssl on the manager DB but I do not think that is fully implemented. as the mgr-setup does not have code for SSL on the manager DB.

You can see in mgr-setup [here](https://github.com/uyuni-project/uyuni/blob/master/susemanager/bin/mgr-setup#L396) only one value is read in

`report-db-ca-cert=$CA_PATH/LOCAL-RHN-ORG-TRUSTED-SSL-CERT`

I was just keeping this in place and adding in the missing variable handling for 

`db-ca-cert`. 

Perhaps this gets addressed more adequately during a refactoring of the whole DB setup handling. I was just trying to add RDS with the least amount of impact/changes.  

Can we resolve this or is further discussion needed?

KeithMnemonic:
> @KeithMnemonic I think you checked this already, but `uyuni-setup-reportdb-user` does not need to be ported as it work already plain over psql, right?

@mcalmer I did not look at uyuni-setup-reportdb-user as it was never called during my mgr-setup runs. What scenario uses this script?  The "pythia" psql  user"  gets added in uyuni-setup-reportdb. What users are these and when would this script be used?

susemanager/bin/mgr-setup
mcalmer: I will have a look tomorrow what happens in the setup code if we get 2 different CAs.

KeithMnemonic:
> > > @KeithMnemonic I think you checked this already, but `uyuni-setup-reportdb-user` does not need to be ported as it work already plain over psql, right?
> > 
> > 
> > @mcalmer I did not look at uyuni-setup-reportdb-user as it was never called during my mgr-setup runs. What scenario uses this script? The "pythia" psql user" gets added in uyuni-setup-reportdb. What users are these and when would this script be used?
> 
> It is used later by the customer to create read-only DB users for reporting tools. We do not want that the users use the RW superuser for just generating reports from the Database.

Ok I need toadd to the script. How is it called, via the shell only?  Is this is in the docs somewhere?  It looks like it reads from rhn.conf. so I will need to add the variables in there as well.  Any other scripts I missed?

mcalmer:
> > > > @KeithMnemonic I think you checked this already, but `uyuni-setup-reportdb-user` does not need to be ported as it work already plain over psql, right?
> > > 
> > > 
> > > @mcalmer I did not look at uyuni-setup-reportdb-user as it was never called during my mgr-setup runs. What scenario uses this script? The "pythia" psql user" gets added in uyuni-setup-reportdb. What users are these and when would this script be used?
> > 
> > 
> > It is used later by the customer to create read-only DB users for reporting tools. We do not want that the users use the RW superuser for just generating reports from the Database.
> 
> Ok I need toadd to the script. How is it called, via the shell only? Is this is in the docs somewhere? It looks like it reads from rhn.conf. so I will need to add the variables in there as well. Any other scripts I missed?

We have other scripts talking to the db directly, but i think they just do normal things like adding or removing data. Nothing else with usermanagement afaik.
But maybe even this script does not need changes, just have a look and i think you will find out.
It is really only called from the terminal directly by the user on the server.

KeithMnemonic:
> > > > > @KeithMnemonic I think you checked this already, but `uyuni-setup-reportdb-user` does not need to be ported as it work already plain over psql, right?
> > > > 
> > > > 
> > > > @mcalmer I did not look at uyuni-setup-reportdb-user as it was never called during my mgr-setup runs. What scenario uses this script? The "pythia" psql user" gets added in uyuni-setup-reportdb. What users are these and when would this script be used?
> > > 
> > > 
> > > It is used later by the customer to create read-only DB users for reporting tools. We do not want that the users use the RW superuser for just generating reports from the Database.
> > 
> > 
> > Ok I need toadd to the script. How is it called, via the shell only? Is this is in the docs somewhere? It looks like it reads from rhn.conf. so I will need to add the variables in there as well. Any other scripts I missed?
> 
> We have other scripts talking to the db directly, but i think they just do normal things like adding or removing data. Nothing else with usermanagement afaik. But maybe even this script does not need changes, just have a look and i think you will find out. It is really only called from the terminal directly by the user on the server.

it needs fixed, let me work on it.

```
server:~ # uyuni-setup-reportdb-user
Report DB Name is: reportdb
[a]dd/[m]odify/[d]elete user (case insensitive). Default is  [m]: a
User: []: reporter1
Password:
Password (again):
Confirm? [y/n] [y]: y
psql:<stdin>:1: ERROR:  permission denied to create role
```

mcalmer:
@KeithMnemonic Seems something break the normal setup of SUMA with this PR. I got the following error when setting up the report DB:

```
** Database: Setting up report database.
** Database: --clear-db option used.  Clearing report database.
*** Progress: #
Could not install report database.

psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  Peer authentication failed for user "-h"
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  Peer authentication failed for user "-h"
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  Peer authentication failed for user "-h"
psql: error: could not translate host name "-U" to address: Name or service not known
```

susemanager/bin/mgr-setup
```diff
@@ -812,6 +857,17 @@ do_setup() {
         REPORT_DB_USER="pythia_susemanager"
     fi
 
+    if [ -z "$MANAGER_DB_HOST" ]; then
+        MANAGER_DB_HOST="localhost"
+    fi
+
+    case $MANAGER_DB_HOST in
+        "localhost"|$(hostname -s)|$(hostname -f))
```
mcalmer, line 865:
```suggestion
        "localhost"|$(hostname -s)|$(hostname -f)|"")
```
Empty host is also "localhost". Or we need to force a value already in Setup.pm. This would also work.

mcalmer (changes requested):

susemanager/bin/mgr-setup
KeithMnemonic: Thanks for catching this.  I pushed up a fix if you can retest.

KeithMnemonic:
> > > > > > @KeithMnemonic I think you checked this already, but `uyuni-setup-reportdb-user` does not need to be ported as it work already plain over psql, right?
> > > > > 
> > > > > 
> > > > > @mcalmer I did not look at uyuni-setup-reportdb-user as it was never called during my mgr-setup runs. What scenario uses this script? The "pythia" psql user" gets added in uyuni-setup-reportdb. What users are these and when would this script be used?
> > > > 
> > > > 
> > > > It is used later by the customer to create read-only DB users for reporting tools. We do not want that the users use the RW superuser for just generating reports from the Database.
> > > 
> > > 
> > > Ok I need toadd to the script. How is it called, via the shell only? Is this is in the docs somewhere? It looks like it reads from rhn.conf. so I will need to add the variables in there as well. Any other scripts I missed?
> > 
> > 
> > We have other scripts talking to the db directly, but i think they just do normal things like adding or removing data. Nothing else with usermanagement afaik. But maybe even this script does not need changes, just have a look and i think you will find out. It is really only called from the terminal directly by the user on the server.
> 
> it needs fixed, let me work on it.
> 
> ```
> server:~ # uyuni-setup-reportdb-user
> Report DB Name is: reportdb
> [a]dd/[m]odify/[d]elete user (case insensitive). Default is  [m]: a
> User: []: reporter1
> Password:
> Password (again):
> Confirm? [y/n] [y]: y
> psql:<stdin>:1: ERROR:  permission denied to create role
> ```



> > > > > @KeithMnemonic I think you checked this already, but `uyuni-setup-reportdb-user` does not need to be ported as it work already plain over psql, right?
> > > > 
> > > > 
> > > > @mcalmer I did not look at uyuni-setup-reportdb-user as it was never called during my mgr-setup runs. What scenario uses this script? The "pythia" psql user" gets added in uyuni-setup-reportdb. What users are these and when would this script be used?
> > > 
> > > 
> > > It is used later by the customer to create read-only DB users for reporting tools. We do not want that the users use the RW superuser for just generating reports from the Database.
> > 
> > 
> > Ok I need toadd to the script. How is it called, via the shell only? Is this is in the docs somewhere? It looks like it reads from rhn.conf. so I will need to add the variables in there as well. Any other scripts I missed?
> 
> We have other scripts talking to the db directly, but i think they just do normal things like adding or removing data. Nothing else with usermanagement afaik. But maybe even this script does not need changes, just have a look and i think you will find out. It is really only called from the terminal directly by the user on the server.

@mcalmer Can I get a list of all other scripts that might be used. anything calling psql will not work as I am guessing it does not pass in host information?

mcalmer:
> @mcalmer Can I get a list of all other scripts that might be used. anything calling psql will not work as I am guessing it does not pass in host information?

These tools should all be available in the main git repository. Some of them are already tested:

* spacewalk-clone-by-date
* spacewalk-data-fsck
* spacewalk-debug (it collect DB informations - we need to exclude it when external DB is used)
* spacewalk-hostname-rename
* spacewalk-remove-channel
* spacewalk-report
* spacewalk-repo-sync
* spacewalk-schema-upgrade
* spacewalk-sql
* spacewalk-update-signatures
* rhn-charsets
* rhn-config-schema.pl
* rhn-db-stats
* rhn-schema-stats
* rhn-schema-version
* rhn-ssl-dbstore
* satellite-sync
* satpasswd
* satwho

From the package "inter-server-sync" (must be explicitly installed):
* inter-server-sync

This is in GO. It export data from DB and import it on a different DB.

KeithMnemonic:
> > @mcalmer Can I get a list of all other scripts that might be used. anything calling psql will not work as I am guessing it does not pass in host information?
> 
> These tools should all be available in the main git repository. Some of them are already tested:
> 
> * spacewalk-clone-by-date
> * spacewalk-data-fsck
> * spacewalk-debug (it collect DB informations - we need to exclude it when external DB is used)
> * spacewalk-hostname-rename
> * spacewalk-remove-channel
> * spacewalk-report
> * spacewalk-repo-sync
> * spacewalk-schema-upgrade
> * spacewalk-sql
> * spacewalk-update-signatures
> * rhn-charsets
> * rhn-config-schema.pl
> * rhn-db-stats
> * rhn-schema-stats
> * rhn-schema-version
> * rhn-ssl-dbstore
> * satellite-sync
> * satpasswd
> * satwho
> 
> From the package "inter-server-sync" (must be explicitly installed):
> 
> * inter-server-sync
> 
> This is in GO. It export data from DB and import it on a different DB.

@mcalmer @maximenoel8  Do we know which of these are already tested?

mcalmer:
spacewalk-clone-by-date
spacewalk-data-fsck
spacewalk-debug (it collect DB informations - we need to exclude it when external DB is used)
spacewalk-hostname-rename
spacewalk-remove-channel
spacewalk-report

Start with these. I think they are not yet tested.

jeremy-moffitt:
> spacewalk-clone-by-date spacewalk-data-fsck spacewalk-debug (it collect DB informations - we need to exclude it when external DB is used) spacewalk-hostname-rename spacewalk-remove-channel spacewalk-report
> 
> Start with these. I think they are not yet tested.

How documented and/or straightforward are these? I'm a little cautious about scope creep for Keith on this. From what was explained to me the CI tests are passing, which was our baseline criteria. Obviously we can be a little flexible on this, but it would be good to have a concrete understanding of if this is the full set of additional manual validation needed, and if there are folks more familiar with these scripts to help Keith out.

KeithMnemonic:
> > spacewalk-clone-by-date spacewalk-data-fsck spacewalk-debug (it collect DB informations - we need to exclude it when external DB is used) spacewalk-hostname-rename spacewalk-remove-channel spacewalk-report
> > Start with these. I think they are not yet tested.
> 
> How documented and/or straightforward are these? I'm a little cautious about scope creep for Keith on this. From what was explained to me the CI tests are passing, which was our baseline criteria. Obviously we can be a little flexible on this, but it would be good to have a concrete understanding of if this is the full set of additional manual validation needed, and if there are folks more familiar with these scripts to help Keith out.

I have to agree with Jeremy here. Maybe someone like @maximenoel8  or yourself @mcalmer  can run these and if you get errors related to `psql`  I can look at them and provide guidance on resolving them.  From a quick glance, most of these  seem to reference [rhnSQL](https://github.com/uyuni-project/uyuni/tree/master/python/spacewalk/server/rhnSQL) and that should work as long as those tools do not need real SUPERUSER permissions. 

I did fix uyuni-setup-reportdb-server so it works in either case now.  I would be good if someone can test that change as well

mcalmer:
I think it would make sense if @maximenoel8 would test these tools on an installation first.
The following are tools required for debugging when something goes wrong:

- spacewalk-data-fsck
- spacewalk-debug (it collect DB informations - we need to exclude it when external DB is used)
- rhn-charsets
- rhn-db-stats
- rhn-schema-stats
- rhn-schema-version


These tools are used by customers for cleanup or re-configuration
- spacewalk-hostname-rename (used to give a SUMA Server a different name)
- spacewalk-remove-channel (used to remove channels and packages which are not used anymore)


These are tools we should test:

- spacewalk-report (generate CSV reports from the Database)
- satpasswd (change the password of a user)
- satwho (display info about a user)
- satellite-sync (ISSv1)
- inter-server-sync (ISSv2)


These should be already tested as part of the testsuite:
- spacewalk-repo-sync
- spacewalk-schema-upgrade
- spacewalk-sql
- rhn-config-schema.pl (part of setup)
- rhn-ssl-dbstore (part of setup)

These are not so important tools. Maybe nobody use them:
- spacewalk-update-signatures

KeithMnemonic:
> I think it would make sense if @maximenoel8 would test these tools on an installation first. The following are tools required for debugging when something goes wrong:
> 
> * spacewalk-data-fsck
> * spacewalk-debug (it collect DB informations - we need to exclude it when external DB is used)
> * rhn-charsets
> * rhn-db-stats
> * rhn-schema-stats
> * rhn-schema-version
> 
> These tools are used by customers for cleanup or re-configuration
> 
> * spacewalk-hostname-rename (used to give a SUMA Server a different name)
> * spacewalk-remove-channel (used to remove channels and packages which are not used anymore)
> 
> These are tools we should test:
> 
> * spacewalk-report (generate CSV reports from the Database)
> * satpasswd (change the password of a user)
> * satwho (display info about a user)
> * satellite-sync (ISSv1)
> * inter-server-sync (ISSv2)
> 
> These should be already tested as part of the testsuite:
> 
> * spacewalk-repo-sync
> * spacewalk-schema-upgrade
> * spacewalk-sql
> * rhn-config-schema.pl (part of setup)
> * rhn-ssl-dbstore (part of setup)
> 
> These are not so important tools. Maybe nobody use them:
> 
> * spacewalk-update-signatures

@maximenoel8 are you alll set this this task now?  @mcalmer Did  you test the updated uyuni-setup-reportdb-user or is that also for @maximenoel8

rjmateus:
@mcalmer ISSv2 are tested in the testsuite with cumcumber tests. ISSv2 uses a golang driver to connect the database.

maximenoel8:
I created a new task for this specific testing : https://github.com/SUSE/spacewalk/issues/18709

maximenoel8:
QA report about the different command : 

Failing commands : 
 -  spacewalk-debug
 -  spacewalk-hostname-rename 
 -  spacewalk-report

Not tested commands : 
 -  spacewalk-update-signatures
 -  inter-server-sync
 -  inter-server-sync

For further information : 

https://github.com/SUSE/spacewalk/issues/18709

mcalmer:
@KeithMnemonic Can you rebase this one more time please? We had a lot of changes in the java dependencies and it seems your "diff" cause building an old package.

KeithMnemonic:
> @KeithMnemonic Can you rebase this one more time please? We had a lot of changes in the java dependencies and it seems your "diff" cause building an old package.

I just updated, see if that resolves the issue

jeremy-moffitt:
> * spacewalk-debug
> * spacewalk-hostname-rename
> * spacewalk-report

How big a concern are these? I was hoping to have Keith pivot to a different project, but also don't want to leave this hanging 98% of the way done...

mcalmer:
> > * spacewalk-debug
> > * spacewalk-hostname-rename
> > * spacewalk-report
> 
> How big a concern are these? I was hoping to have Keith pivot to a different project, but also don't want to leave this hanging 98% of the way done...

Not too big.

spacewalk-debug is calling smdba which is a tool for configuring a local postgresql DB. We just need to prevent that this is called when RDS is used.

spacewalk-hostname-rename: not sure if this is needed in the cloud. I would risk it to merge without looking into it.

spacewalk-report: I think this was an error in the test. The report DB is "filled with data" only 1 time per day. If you directly call it, the DB is empty and you get no result. I just think nobody run the job to fill the DB with data manually before trying this tool.

I currently run the testsuite on this PR. The last run came bck with 90 errors. A bit too much, but could be related to infrastructure problems. There is currently another run in progress. Let's see how this look like.

KeithMnemonic:
> > > * spacewalk-debug
> > > * spacewalk-hostname-rename
> > > * spacewalk-report
> > 
> > 
> > How big a concern are these? I was hoping to have Keith pivot to a different project, but also don't want to leave this hanging 98% of the way done...
> 
> Not too big.
> 
> spacewalk-debug is calling smdba which is a tool for configuring a local postgresql DB. We just need to prevent that this is called when RDS is used.
> 
> spacewalk-hostname-rename: not sure if this is needed in the cloud. I would risk it to merge without looking into it.
> 
> spacewalk-report: I think this was an error in the test. The report DB is "filled with data" only 1 time per day. If you directly call it, the DB is empty and you get no result. I just think nobody run the job to fill the DB with data manually before trying this tool.
> 
> I currently run the testsuite on this PR. The last run came bck with 90 errors. A bit too much, but could be related to infrastructure problems. There is currently another run in progress. Let's see how this look like.

@mcalmer I can add this check around the call to smdba in spacewalk-debug

```
if [ "$(spacewalk-cfg-get externaldb)" = "0" ] ; then
    if [ -f /usr/bin/smdba ] ; then
        /usr/bin/smdba space-overview > $DIR/db-control-report
    elif [ -f /usr/bin/db-control ] && [ "$(spacewalk-cfg-get db_backend)" = "postgresql" ] ; then
        /usr/bin/db-control report > $DIR/db-control-report
    fi
fi
```

mcalmer:
@KeithMnemonic please do. Thanks. And please rebase also. In best case i can merge tomorrow

KeithMnemonic:
> @KeithMnemonic please do. Thanks. And please rebase also. In best case i can merge tomorrow

updated, rebased and pushed.

mcalmer (approved):
Looks good. Testsuite reported only 18 unrelated errors

PR #5654, opened 2022-07-12
Properly handling of Debian packages in rhnpush

### LINKED ISSUE #5212: Enhance rhnpush to handle Debian packages. Was: 3 Problems with creating custom channels for Ubuntu.
# Problem description

We've been trying to set up a custom channel for some 3rd party packages that we want to deploy across the organization. the idea is to load 3rd party rpm and deb packages with rhnpush so that they can then be installed with yum /dnf and apt-get. This is failing in a number of ways, some of which we were able to patch/work around, and another not so much.

## Version of Uyuni Server and Proxy (if used)

Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : uyuni-server-stable
Name           : Uyuni-Server-release
Version        : 2022.03-172.4.uyuni1
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes (automatically)
Status         : out-of-date (version 2022.02-171.1.uyuni1 installed)
Source package : Uyuni-Server-release-2022.03-172.4.uyuni1.src

## Details about the issue

The first problem is that rhnpush doesn't support uploading .deb packages, and the staging directory on the server back-end has the wrong permissions preventing the creation of the temp file.  The following minor change allowed rhnpush to process .deb files as well

```
#diff -u /usr/lib/python3.6/site-packages/rhnpush/uploadLib.py.bak /usr/lib/python3.6/site-packages/rhnpush/uploadLib.py
--- /usr/lib/python3.6/site-packages/rhnpush/uploadLib.py.bak   2022-04-13 11:02:58.598147587 -0700 
+++ /usr/lib/python3.6/site-packages/rhnpush/uploadLib.py       2022-04-13 10:07:14.590320048 -0700
@@ -177,7 +177,7 @@

         for filename in listdir(self.options.dir):
             # only add packages
-            if filename[-3:] in ("rpm", "mpm"):
+            if filename[-3:] in ("rpm", "mpm", "deb"):
                 self.files.append(filename)

     def filter_excludes(self):
```

Next, we ran into some permission issues on the back end when running rhnpush. 

Exception Handler Information
File "/usr/lib/python3.6/site-packages/spacewalk/server/apacheUploadServer.py", line 102, in _wrapper ret = function(req) File "/usr/share/rhn/upload_server/handlers/package_push/package_push.py", line 117, in handler self.file_checksum_type, self.file_checksum) File "/usr/lib/python3.6/site-packages/spacewalk/server/rhnPackageUpload.py", line 281, in save_uploaded_package prefix='-'.join((nevra[0], nevra[2], nevra[3], nevra[4]))) File "/usr/lib64/python3.6/tempfile.py", line 551, in NamedTemporaryFile (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type) File "/usr/lib64/python3.6/tempfile.py", line 262, in _mkstemp_inner fd = _os.open(file, flags, 0o600)
PermissionError: [Errno 13] Permission denied: '/var/spacewalk/packages/1/stage/Newpackaget-8.3.0-X-amd64-debmdg_d6ri'

Turns out permissions on /var/spacewalk/packages/1/stage were root.root. Running
`chown wwwrun.www  /var/spacewalk/packages/1/stage `
to match the other dirs in  /var/spacewalk/packages/1 allowed the rhnpush to succeed.

The second problem has to do with package matching. One of the (3rd-party built) packages has a mixed case package name in the metadata. As you can see, using the mixed name of the package fails to match it in the package search on Ubuntu. Only using a lower case conversion of the name works. 

root@CARMD-EX-DWSBX1:/home/t1_ppanon@sierrawireless.local# apt-cache show nessusagent
Package: NessusAgent
Version: 8.3.0
Architecture: amd64
Maintainer: Debian
Conflicts: nessusd (<< 3.0)
Replaces: nessusd (<< 3.0)
Filename: ubuntu20-it_baseline-ubuntu-2004-amd64-itsec/getPackage/NessusAgent_8.3.0-X.amd64-deb.deb
Size: 16904902
SHA256: 4534e66bd3d85e31dba1c2db0d789395e9d453edd204819d82acae2f5166820e
Section: NoGroup
Priority: extra
Description: The Nessus Client Agent
Description-md5: 1c986ede11904b59f30b8d23d2104a8e

root@CARMD-EX-DWSBX1:/home/t1_ppanon@sierrawireless.local# apt-cache show NessusAgent
N: Unable to locate package NessusAgent
E: No packages found

This is a problem in that this search failure also applies to apt install commands.  Trying to install the package from the Uyuni system software tab will try to use the mixed case metadata name, and therefore also fails. While I haven't tried to add a newer version of the package to the channel to trigger an update, I expect that doing so also would result in a failed cache match. I'm not sure if this is an upstream apt bug or in the uyuni shim libraries.

ycedres:
## What does this PR change?

Fixes `rhnpush` to be able to upload a deb package so it can be pushed into a channel.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Fixes #5212 
Tracks https://github.com/SUSE/spacewalk/issues/17726

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


client/tools/mgr-push/mgr-push.changes
```diff
@@ -1,3 +1,4 @@
+- Allow rhnpush to handle deb packages
```
meaksh, line 1: We always put a line break after the line of the previous header. So it should be like:

```suggestion
- Allow rhnpush to handle deb packages

```

meaksh (commented):
Thanks for the PR! This is looking good but I still see some things we should tackle here I think. Looking for "rpm" in the `uploadLib.py` file I see there are cases, particularly around `packageCompare` function that we should cover for the "deb" case.

We could test this by using `mgrpush --newest` AFAICS.

client/tools/mgr-push/Makefile.python
```diff
@@ -2,18 +2,46 @@ THIS_MAKEFILE := $(realpath $(lastword $(MAKEFILE_LIST)))
 CURRENT_DIR := $(dir $(THIS_MAKEFILE))
 include $(CURRENT_DIR)../../../rel-eng/Makefile.python
 
+MGR-PUSH_ROOT_DIR = /manager/client/tools/mgr-push
+RHNPUSH_PACKET_DIR = $(MGR-PUSH_ROOT_DIR)/rhnpush
```
meaksh, line 6: Nitpick: I would better use this name:

```suggestion
RHNPUSH_PACKAGE_DIR = $(MGR-PUSH_ROOT_DIR)/rhnpush
```

client/tools/mgr-push/Makefile.python
```diff
@@ -2,18 +2,46 @@ THIS_MAKEFILE := $(realpath $(lastword $(MAKEFILE_LIST)))
...
 DOCKER_REGISTRY       = registry.opensuse.org
-DOCKER_RUN_EXPORT     = "PYTHONPATH=$PYTHONPATH"
+DOCKER_RUN_EXPORT     = "PYTHONPATH=/manager/client/tools/mgr-push/rhnpush:/manager/python/:/manager/client/tools/mgr-push:/manager/client/rhel/spacewalk-client-tools/src/"
 DOCKER_VOLUMES        = -v "$(CURDIR)/../../../:/manager"
 
+define setup_structure
+    mv $(MGR-PUSH_ROOT_DIR)/rhnpush $(MGR-PUSH_ROOT_DIR)/rhnpush_wrapper
+    mkdir $(RHNPUSH_PACKET_DIR)
+    cd $(MGR-PUSH_ROOT_DIR) && mv archive.py connection.py rhnpush_cache.py rhnpush_config.py rhnpush_confmanager.py rhnpush_main.py rhnpush_v2.py rpm2mpm.py uploadLib.py utils.py $(RHNPUSH_PACKET_DIR)
+    touch $(RHNPUSH_PACKET_DIR)/__init__.py
```
meaksh, line 19: Here, instead of `mv` I would use `cp` to some testing path, to avoid messing the enviroment in the host of the container while the tests are running.

client/tools/mgr-push/test/test_uploadclass.py
```diff
@@ -0,0 +1,315 @@
...
+
+                self._upload.packages()
+                package.assert_called_with(
+                    "newPackage_2.0-1_amd64.deb",
+                    "sha256",
+                    "ece4eedf7a5c65396d136b5765226e2c8b10f268c744b0ab1fa2625e35384a00",
+                )
+
+                # test call to uploadLib.call(self.server.packages.channelPackageSubscriptionBySession, self.session.getSessionString(), info)
+                # info [{'name': 'avra', 'version': '1.3.0', 'release': '3', 'epoch': '', 'arch': 'amd64-deb', 'checksum_type': 'sha256', 'checksum': 'e08f76afa7bd472a3a7db84874a208bda3ca59b4f6784136a31df35dcf231ec6'}]
```
meaksh, line 181: Is something missing here?

meaksh (commented):
Thanks for this @ycedres! I just added some small comments :+1:

client/tools/mgr-push/Makefile.python
ycedres: Done as you suggested.

client/tools/mgr-push/Makefile.python
```diff
@@ -2,18 +2,46 @@ THIS_MAKEFILE := $(realpath $(lastword $(MAKEFILE_LIST)))
 CURRENT_DIR := $(dir $(THIS_MAKEFILE))
 include $(CURRENT_DIR)../../../rel-eng/Makefile.python
 
+MGR-PUSH_ROOT_DIR = /manager/client/tools/mgr-push
```
vzhestkov, line 5: Are you sure MGR-PUSH_ROOT_DIR is valid for variable name?

client/tools/mgr-push/Makefile.python
ycedres: You're right, that's better.

client/tools/mgr-push/Makefile.python
ycedres: At least it seems to work, do you think it shouldn't for any reason?

client/tools/mgr-push/test/test_uploadclass.py
ycedres: That's a leftover that I wanted to remove, thank you for noticing.

client/tools/mgr-push/Makefile.python
meaksh: The `-` is sometimes a tricky character :) I would also have preferred using `MGRPUSH_ROOT_DIR` or `MGR_PUSH_ROOT_DIR`.

client/tools/mgr-push/Makefile.python
```diff
@@ -2,18 +2,46 @@ THIS_MAKEFILE := $(realpath $(lastword $(MAKEFILE_LIST)))
...
 DOCKER_REGISTRY       = registry.opensuse.org
-DOCKER_RUN_EXPORT     = "PYTHONPATH=$PYTHONPATH"
+DOCKER_RUN_EXPORT     = "PYTHONPATH=/manager/client/tools/mgr-push/rhnpush:/manager/python/:/manager/client/tools/mgr-push:/manager/client/rhel/spacewalk-client-tools/src/"
 DOCKER_VOLUMES        = -v "$(CURDIR)/../../../:/manager"
 
+define setup_structure
+    mv $(MGR-PUSH_ROOT_DIR)/rhnpush $(MGR-PUSH_ROOT_DIR)/rhnpush_wrapper
+    mkdir $(RHNPUSH_PACKAGE_DIR)
+    cd $(MGR-PUSH_ROOT_DIR) && cp archive.py connection.py rhnpush_cache.py rhnpush_config.py rhnpush_confmanager.py rhnpush_main.py rhnpush_v2.py rpm2mpm.py uploadLib.py utils.py $(RHNPUSH_PACKAGE_DIR)
+    touch $(RHNPUSH_PACKAGE_DIR)/__init__.py
```
meaksh, line 19: It seems you are still missing to tackle the comment I did at https://github.com/uyuni-project/uyuni/pull/5654#discussion_r974163415 here (using `cp` instead of `mv`)

client/tools/mgr-push/Makefile.python
ycedres: Variable renamed :)

client/tools/mgr-push/Makefile.python
```diff
@@ -2,18 +2,46 @@ THIS_MAKEFILE := $(realpath $(lastword $(MAKEFILE_LIST)))
...
+
 # Docker tests variables
+DOCKER_REF            = $(DOCKER_REGISTRY)/$(DOCKER_IMAGE)
 DOCKER_CONTAINER_BASE = systemsmanagement/uyuni/master/docker/containers/uyuni-master
 DOCKER_REGISTRY       = registry.opensuse.org
-DOCKER_RUN_EXPORT     = "PYTHONPATH=$PYTHONPATH"
+DOCKER_RUN_EXPORT     = "PYTHONPATH=/manager/client/tools/mgr-push/rhnpush:/manager/python/:/manager/client/tools/mgr-push:/manager/client/rhel/spacewalk-client-tools/src/"
 DOCKER_VOLUMES        = -v "$(CURDIR)/../../../:/manager"
 
+define setup_structure
```
agraul, line 15: I feel like this is a step in the wrong direction. We have a problem with our Python packages: we can't install them into a virtual environment to run tests against (like any well maintained Python project does), and while this workaround let's us run the tests, it does not fix the underlying issue.

Have you considered fixing this in-tree with a [`src` layout](https://hynek.me/articles/testing-packaging/)? That's what many projects use successfully.

ppanon2022:
In the PR line that updates line 565 of uploadLib.py
        if filename.endswith("rpm") or filename.endswith("npm"):
shouldn't that be
        if filename.endswith("rpm") or filename.endswith("mpm"):
All the other extension references appear to be mpm so I'm guessing that might be a typo.

client/tools/mgr-push/Makefile.python
ycedres: After clarifying this with you (the problem is that I need to create a transitory rhnpush package so it can be imported in the unit tests) the approach is now to create it in a temporary directory inside the container instead of making any change in the code structure when testing.

client/tools/mgr-push/Makefile.python
ycedres: Would you consider reasonable to create a follow-up card about this issue? I believe this would take a longer time to be properly solved.

client/tools/mgr-push/uploadLib.py
```diff
@@ -560,7 +558,14 @@ def _processFile(filename, relativeDir=None, source=None, nosig=None):
             lh.append(sstr(a_pkg.header['arch']))
 
         # Build the header hash to be sent
-        info = {'header': Binary(a_pkg.header.unload()),
+        if filename.endswith("deb"):
+            # The 'header' data in the returned dictionary is used with the --headers option for mgr-push
+            # that have to be deprecated as is no longer needed.
```
agraul, line 563: I don't understand this comment. What needs to be deprecated?

client/tools/mgr-push/Makefile.python
agraul: Yes, it's reasonable to change the way we run the tests in a follow-up pull request. In the meantime, please make sure we run the new unit tests in Jenkins or via GitHub Actions.

client/tools/mgr-push/uploadLib.py
ycedres: It seems that the upload-headers-only feature (--headers) is not useful any more as it this was needed by the up2date tool.

client/tools/mgr-push/test/test_uploadclass.py
```diff
@@ -0,0 +1,315 @@
+from unittest.mock import Mock, patch, MagicMock
+import sys
+sys.modules['platform'] = MagicMock()
+sys.modules['rpm'] = MagicMock()
```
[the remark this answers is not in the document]
ycedres: In fact I'm mocking the calls to the `debian` package as it is obviously used to read `.deb` files and in the tests any file is actually read. So I wonder if it would make any difference to actually import or not those packages (`debian` and `rpm`) in the unit tests.

meaksh (approved):
Besides of what have been already mentioned, this looks good to me.

IMPORTANT NOTE: The changes in this PR are making `mgrpush` to have a new dependency: `python-debian` or `python3-debian`.

Since `mgrpush` is part of the client tools channels for the different OS, we need to make sure we have this new package building and included in the different client tools channels at OBS/IBS and also added to the channels definition.

You need to coordinate with @juliogonzalez on this at the time of merging this PR.

.github/workflows/mgrpush-unit-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
+
+      - name: Install Python dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          pip install pytest mock pyyaml pyOpenSSL python-debian
```
agraul, line 22: In the future we should think about how we want to manage Python dependencies. That's not in the scope of this PR, but we need to define something that we then follow everywhere. Right now we maintain them in too many places.

client/tools/mgr-push/test/test_archive.py
```diff
@@ -38,6 +38,7 @@
 # test case --------------------------------------------------------------
 
 
+@unittest.skip("skipping ArchiveTest")
```
agraul, line 41: Why is this skipped?

agraul (changes requested):
I haven't tested the changes to the directory layout.

I have a few questions around the unit tests, it would be nice if we could clarify that before we merge the PR.

client/tools/mgr-push/test/test_uploadclass.py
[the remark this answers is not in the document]
ycedres: It was just a way to avoid the requirement of having the `python3-rpm` package installed on the testing system, as this was a problem for running the tests as a Github action (no rpm-based supported runners). I don't know is there's a better way to achieve this.

Same for the `platform` module, but this `import` is going to be removed as this is part of the traditional stack (that `platform` import occurs inside the `up2date_client` package).

client/tools/mgr-push/test/test_archive.py
ycedres: There're some old tests that I believe don't make sense or don't match the current implementation. I didn't think that it was in the scope of this PR so I skipped them for the moment. Surely they have to be rewritten.

.github/workflows/mgrpush-unit-tests.yml
ycedres: Do you think it should be done differently at the moment?

agraul: No, we don't need to change it now.

client/tools/mgr-push/uploadLib.py
agraul: I still don't understand the comment as it is written in the code, could you please reword it?


If `hdr` is not needed anymore because it's only for the traditional stack, does is still make sense to have it set to different values depending on the package type?

client/tools/mgr-push/test/test_uploadclass.py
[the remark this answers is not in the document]
agraul: I understand why you're doing it like that, but I'm not sure that I will remember that when I open this file in 3 months from now. I think this deserves a code comment.

client/tools/mgr-push/test/test_archive.py
agraul: I guess the hint in the commit message is enough to document this.

client/tools/mgr-push/test/test_uploadclass.py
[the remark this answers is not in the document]
ycedres: All right, I misunderstood :)

client/tools/mgr-push/uploadLib.py
ycedres: I'm not sure if with the `--header` option `mgr-push` runs parts of the traditional stack or not. It does call functionality in the `server.importlib.packageUpload` module inside the XML-RPC server, where the header extraction for an rpm file takes place. 

When the `--newest` option is used, the `unload()` method for the package header is called. This only works for rpm packages so I had to avoid this to happen for a deb packages. 

I believe most of the information in that `info` dictionary is used precisely when the `--header` option is used. With the  `--newest` option, just `info['nvrea']` is later used.

In order to completely remove this header treatment I reckon we should remove the `--header` option.

client/tools/mgr-push/test/test_uploadclass.py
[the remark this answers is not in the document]
ycedres: I added a comment clarifying the mocking of the `sys.modules` keys.

client/tools/mgr-push/src/rhnpush/uploadLib.py
```diff
@@ -793,11 +798,14 @@ def packageCompare(pkg1, pkg2, is_mpm=None):
             e = str(e)
         evr = (e, str(pkg[1]), str(pkg[2]))
         packages.append(evr)
-    if is_mpm:
+    if pkgtype == "mpm":
         func = rhn_mpm.labelCompare
-    else:
-        from uyuni.common import rhn_rpm
+    elif pkgtype == "deb":
+        func = rhn_deb.label_compare
```
agraul, line 804: IMO this should also be named `labelCompare`. Yes, that's not the style that's commonly used on Python, and it's also not what I would do, but being consistent is more important than sticking to snake_case.

After the traditional stack is removed, we will implement linting again and then we can define our style and update the code accordingly.

client/tools/mgr-push/test/test_RhnpushConfig.py
```diff
@@ -34,9 +34,10 @@ def testReadConfigFiles(self):
         assert self.userconfig.settings != None and self.defaultconfig.settings != None
 
     def testGetOption(self):
-        a = self.userconfig.get_option('usage')
-        b = self.defaultconfig.get_option('usage')
-        assert a != None and b != None and a == '0' and b == '0'
+        with patch('configparser.ConfigParser.get', return_value='0'):
+            a = self.userconfig.get_option('usage')
+            b = self.defaultconfig.get_option('usage')
+            assert a != None and b != None and a == '0' and b == '0'
```
agraul, line 40: nitpick: `a != None and b != None` are redundant here

client/tools/mgr-push/src/rhnpush/uploadLib.py
ycedres: Agreed. I changed it upon SonarCloud check failure but apparently it wasn't even retriggered.

.github/workflows/mgrpush-unit-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+  mgrpush_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
```
agraul, line 17: How did you come to the decision to use Python 3.8?

ycedres: The GitHub Action was failing running `pytest` with Python 3.11 because `python/uyuni/common/checkum.py` uses `inspect.getargspec()` and this function has been [removed](https://docs.python.org/3.11/library/inspect.html) in Python 3.11. However, it it still there for Python 3.10 (although marked as deprecated).

agraul: I think it would be good to test with the versions that will be used in production. With GH we don't pay for additional Python version, I would test with all of the supported ones. I created a [wiki page](https://github.com/uyuni-project/uyuni/wiki/Python-Version-Overview) with an overview

ycedres: I'm having trouble testing with Python 2.7 as `httplib` doesn't seem to be available in the runner, neither available via `pip`.

ycedres: Regarding the 3.x versions, 3.4 is available for Ubuntu [18.04](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json) and this runner image has been marked as [deprecated](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/).

ycedres:
> I haven't tested the changes to the directory layout.
> 
> I have a few questions around the unit tests, it would be nice if we could clarify that before we merge the PR.

@agraul Is there anything else you notice worth taking care of before merging?

agraul:
> > I haven't tested the changes to the directory layout.
> > I have a few questions around the unit tests, it would be nice if we could clarify that before we merge the PR.
> 
> @agraul Is there anything else you notice worth taking care of before merging?

Nothing we haven't already discussed!

agraul (approved):

Issue #5573, opened 2022-06-17
Script configure-proxy.sh fails to get certificate from Uyuni server


phibid:
While trying to configure a new Uyuni proxy, I get the following error:

```
# configure-proxy.sh --answer-file=/tmp/proxyanswers.txt
Requesting certificate from server. [1/20]
Requesting certificate from server. [2/20]
Requesting certificate from server. [3/20]
Requesting certificate from server. [4/20]
Requesting certificate from server. [5/20]
Requesting certificate from server. [6/20]
Requesting certificate from server. [7/20]
Requesting certificate from server. [8/20]
Requesting certificate from server. [9/20]
Requesting certificate from server. [10/20]
Requesting certificate from server. [11/20]
Requesting certificate from server. [12/20]
Requesting certificate from server. [13/20]
Requesting certificate from server. [14/20]
Requesting certificate from server. [15/20]
Requesting certificate from server. [16/20]
Requesting certificate from server. [17/20]
Requesting certificate from server. [18/20]
Requesting certificate from server. [19/20]
Requesting certificate from server. [20/20]
Certificate not received from server. Exit.
/etc/sysconfig/rhn/systemid:1: parser error : Document is empty
 
^
unable to parse /etc/sysconfig/rhn/systemid
SUSE Manager Parent [uyuni]: uyuni.xx.test
Using CA Chain (from /etc/sysconfig/rhn/up2date): /usr/share/rhn/RHN-ORG-TRUSTED-SSL-CERT
HTTP Proxy []:
Traceback email [[philippe.bidault@getronics.com](mailto:philippe.bidault@getronics.com)]: [philippe.bidault@getronics.com](mailto:philippe.bidault@getronics.com)
You will now need to either generate or import an SSL certificate.
This SSL certificate will allow client systems to connect to this Uyuni Proxy
securely. Refer to the Uyuni Proxy Installation Guide for more information.
Do you want to import existing certificates? [N]: N
Organization [XX]: XX
Organization Unit [XX]: XX
Common Name [uyuni_proxy]: uyuni_proxy
City [XX]: XX
State [XX]: XX
Country code [XX]: XX
Email [philippe.bidault@getronics.com]: philippe.bidault@getronics.com
Cname aliases (separated by space) [uyuni_proxy]: uyuni_proxy
Using CA key at /root/ssl-build/RHN-ORG-PRIVATE-SSL-KEY.
Generating SSL key and public certificate.
Rotated out: 'server.key.6'
[...]

PR #5681, opened 2022-07-18
Fix proxy minion -  Uyuni 5573

mbussolotto:
## What does this PR change?

When salt bundle is used, we need to use `/usr/lib/venv-salt-minion/bin/python` python version.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/5573
Fixes https://github.com/SUSE/spacewalk/issues/18206

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/proxy/installer/configure-proxy.sh
+++ b/proxy/installer/configure-proxy.sh
@@ -329,8 +329,8 @@ PYTHON_ENV=""
 systemctl is-active --quiet salt-minion && PYTHON_CMD="/usr/bin/python3"
 systemctl is-active --quiet venv-salt-minion && PYTHON_CMD="/usr/lib/venv-salt-minion/bin/python"
 
-if [[ -n $PYTHON_ENV ]]; then
-    $PYTHON_ENV /usr/share/rhn/proxy-installer/fetch-certificate.py $SYSTEMID_PATH
+if [[ -n $PYTHON_CMD ]]; then
+    $PYTHON_CMD /usr/share/rhn/proxy-installer/fetch-certificate.py $SYSTEMID_PATH
     MASTER_CONF=/etc/salt/minion.d/susemanager.conf
     if [ -f /etc/venv-salt-minion/minion.d/susemanager.conf ]; then
         MASTER_CONF=/etc/venv-salt-minion/minion.d/susemanager.conf
--- a/proxy/installer/configure-proxy.sh
+++ b/proxy/installer/configure-proxy.sh
@@ -326,7 +326,7 @@ UP2DATE_FILE=$SYSCONFIG_DIR/up2date
 SYSTEMID_PATH=$(awk -F '=[[:space:]]*' '/^[[:space:]]*systemIdPath[[:space:]]*=/ {print $2}' $UP2DATE_FILE)
 
 PYTHON_ENV=""
-systemctl is-active --quiet salt-minion && PYTHON_ENV="/usr/bin/python3 -u "
+systemctl is-active --quiet salt-minion && PYTHON_ENV="/usr/bin/python3 "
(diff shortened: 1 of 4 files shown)


vzhestkov:
@mbussolotto are there any chances for `fetch-certificate` to be executed by the user?

mbussolotto:
I would say no. Currently it's just used by `configure-proxy.sh`

proxy/installer/configure-proxy.sh
```diff
@@ -325,9 +325,12 @@ SQUID_DIR=/etc/squid
 UP2DATE_FILE=$SYSCONFIG_DIR/up2date
 SYSTEMID_PATH=$(awk -F '=[[:space:]]*' '/^[[:space:]]*systemIdPath[[:space:]]*=/ {print $2}' $UP2DATE_FILE)
 
-systemctl status salt-minion > /dev/null 2>&1 || systemctl status venv-salt-minion > /dev/null 2>&1
-if [ $? -eq 0 ]; then
-    /usr/sbin/fetch-certificate $SYSTEMID_PATH
+PYTHON_ENV=""
+systemctl is-active --quiet salt-minion && PYTHON_ENV="/usr/bin/python -u "
```
agraul, line 329: `/usr/bin/python` is Python2, is that on purpose?

vzhestkov: It must be running with `python3` for sure

mbussolotto: is it possible it's the cause of the issue https://github.com/uyuni-project/uyuni/issues/5573#issuecomment-1187347996 ? Maybe `fetch-certificate.py` would work just using `#!/usr/bin/python3 -u` as shebang and adding `/usr/lib/venv-salt-minion/bin/python` is not required

vzhestkov: I think no, before fetch-certificate was executed with the python specified in the shebang and it will just fail with python2 as we are not shipping salt module for it.

mbussolotto:
Tested and it works properly

proxy/installer/fetch-certificate.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/python -u
```
vzhestkov, line 1: I'm not sure if it really make sense to enforce unbuffered STDOUT/STDERR streams with `-u`, which is not used for the bundle on calling `/usr/lib/venv-salt-minion/bin/python` with `configure-proxy.sh`. And please check the mod of the file, it looks like you've removed shebang here, but the file is still set as 0755 with the spec. And it's not so good to put it to `bin` with no exec mod.

mbussolotto: it should looks better now, thanks for your suggestion!

vzhestkov: Thanks. It's better, but there is still the difference with `-u` for non-bundled python. Does it make any sense there to force unbuffered streams? If does, `-u` should be added to the bundled pythin call also.

mbussolotto: correct, I forgot to check it. It seems that `-u` does not provide any further information (I also tried to force an error condition), so I remove it from `python3`

vzhestkov: Sorry for one more nitpick. Does the space at the end of `PYTHON_ENV` make any sense? I would also suggest to rename the variable to `PYTHON_CMD` as ENV is not really related to environment.

vzhestkov: Thanks for addressing all the suggestions :+1:

vzhestkov (approved):
Looks good :+1:

PR #5690, opened 2022-07-20
Adding scenario to check on taskomatic logs.

elariekerboull:
## What does this PR change?

Adding tests to check taskomatic logs for errors.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Cucumber tests were added : Taskomatic log check based on Tomcat log check previously implemented.

- [x] **DONE**

## Links

Fixes #https://github.com/SUSE/spacewalk/issues/15565
Tracks # 
4.2
4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -487,6 +487,14 @@
   end
 end
 
+Then(/^the taskomatic logs should not contain errors$/) do
+  output, _code = $server.run('cat /var/log/rhn/rhn_taskomatic_daemon.log && cat /var/log/rhn/reposync/*')
```
vandabarata, line 491: honest curiosity, did you test this exact command? trying to understand what we're `cat`ing in the 2nd part of it :thinking:

elariekerboull: Good question, it is `cat`ing the reposync logs but since it is a PR I started some time ago, I am not sure why I also check these.
I did test it without no issue otherwise, but I'll check again why I have this

vandabarata: interesting, didn't know it could also work like that, without a specific file :thinking: more importantly though, do we need to check that in this step? are the reposync logs related to taskomatic? or should we maybe have a different check/step for them?

elariekerboull: Oh and according to Eric reposync is managed through taskomatic

vandabarata (approved):

testsuite/features/step_definitions/command_steps.rb
elariekerboull: And we already check reposync logs with this step: https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/step_definitions/command_steps.rb#L410
So it was indeed completely not necessary

PR #5692, opened 2022-07-20
Using session id parameter as String when communicating with the search server

wweellddeerr:
## What does this PR change?

In some cases, when the session id reaches a high value, the endpoints of the xml-rpc api using it as a parameter stop working, since it is serialized as an Integer by default. This commit adds the possibility to pass the session id as a String when communicating with the search server.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/15122

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


moio:
@wweellddeerr: thanks for this contribution. I appreciate it addresses the problem in a straightforward, pragmatic way.

I was wondering together with @rjmateus, though, if we could take the chance and clean up this part of the codebase a bit further. One of the ideas that floated was to not use the XMLRPC API at all for Search going forward, since we know this limitation is inherent to the protocol and can’t be removed. Instead, we could use the newer http+json API, which we ourselves introduced for the main API because of this and other problems (and we should lead by giving the good example!)

@rjmateus, @wweellddeerr: could you please have a look and determine if bringing the http api to the Search server is something that can be achieved without a lot of pain? It is understood this is bigger than just addressing the point issue we are facing here, and we definitely do not want to spend months on this. But if that is doable with a mid-sized effort (1 week or so) I guess it might be worth it, even just from the learning perspective.

Thoughts welcome.

moio:
cc @admd

cbbayburt:
> @rjmateus, @wweellddeerr: could you please have a look and determine if bringing the http api to the Search server is something that can be achieved without a lot of pain? It is understood this is bigger than just addressing the point issue we are facing here, and we definitely do not want to spend months on this. But if that is doable with a mid-sized effort (1 week or so) I guess it might be worth it, even just from the learning perspective.

@moio IIUC, the search server communication happens the other way around. The search server is the server and the main Java app is the client. So unless we export the HTTP API framework as a jar and use it in both processes (not sure how doable at this point), it's not an option in this case.

Also, please keep in mind that this one is a fix for an L3. It shouldn't be a problem if this is going to be overridden with a more comprehensive refactoring, but if you approve the solution in itself, I suggest we go ahead with this and make it to this MU.

moio:
> @moio IIUC, the search server communication happens the other way around. The search server is the server and the main Java app is the client. So unless we export the HTTP API framework as a jar and use it in both processes (not sure how doable at this point), it's not an option in this case.

My memory could be failing - doesn't the search server also import rhn.jar, thereby inheriting most of the code base? If that is not the case I agree, it's a bigger change. I'll leave it up to you guys to decide whether it's worth it or not in that case.

> Also, please keep in mind that this one is a fix for an L3. It shouldn't be a problem if this is going to be overridden with a more comprehensive refactoring, but if you approve the solution in itself, I suggest we go ahead with this and make it to this MU.

I was under the impression the L3 got already closed with a workaround, so this effort was about having the definitive fix. I can agree with your plan in case there is any pressure from users or customers in the meantime.

Thanks for the insights

cbbayburt:
> My memory could be failing - doesn't the search server also import rhn.jar, thereby inheriting most of the code base? If that is not the case I agree, it's a bigger change. I'll leave it up to you guys to decide whether it's worth it or not in that case.

Even if it does, I'm not sure the HTTP API is not too tightly coupled with the app so that it can be used with the search server as well. I need to take a look into it from that perspective. @rjmateus if you wish, I can help out with this part.

> I was under the impression the L3 got already closed with a workaround, so this effort was about having the definitive fix. I can agree with your plan in case there is any pressure from users or customers in the meantime.

Also true, but they sounded a bit dissatisfied with our last response, and it's only a matter of time before they hit the integer limit again. Since we already have this patch in hand, they'd be happy to receive it.

moio:
No problem from my side to have this in order to buy time.

Looking forward to see research results as well.

cbbayburt (approved):

rjmateus (commented):
LGTM

rjmateus:
@cbbayburt @moio I think this patch is the best solution to solve the current problem, and should be delivered to the customer and merged into our codebase.

Regarding the API modification to be JSON over HTTP in the search server. IIUC the source code from the search server is based on old libraries and is one of the black holes in the project.
I'm afraid we would just give new paint to a really old house, without changing much (besides the integer problem).

Suma already had a JSON API mechanism with the SPARK framework.
The search application is not a "traditional" web application with web.xml file. It uses a framework/library (which I don't know) named picocontainer [1] [2] and then starst RpcServer [3]. I don't know how easy would be to serve one more handler type.

In my opinion, after a quick look, I think we should do this fix and then if needed rebuild the search module, instead of trying to squeeze one more layer into a really old piece of code.
What are your thoughts on this?

I don't mind going dip on researching this regard topic, if someone this it worth our time. 

[1] http://picocontainer.com/introduction.html (I think this is the correct link)
[2] https://github.com/uyuni-project/uyuni/blob/master/search-server/spacewalk-search/src/java/com/redhat/satellite/search/Main.java
[3] https://github.com/uyuni-project/uyuni/blob/master/search-server/spacewalk-search/src/java/com/redhat/satellite/search/rpc/RpcServer.java#L88-L113

moio:
Yuck. I guess we could tackle this the moment we re-do the Search server from scratch - which was more or less in the plans already.

Thanks for the excavation work!

PR #5701, opened 2022-07-22
Fix bootstrap issue in Debian 9 with classic minion because missing python3-contextvars (bsc#1201782)

meaksh:
## What does this PR change?

This PR fixes an issue happening when bootstrapping a Debian 9 system using the "classic" Salt minion package:

```console
Apply States failed during bootstrap (retcode=2). Details:

pkg_|-salt-install-contextvars_|-python3-contextvars_|-installed: Problem encountered installing package(s). Additional info follows:

errors:
    - Running scope as unit: run-rd65abbb383994fdb8d268b384a1ca046.scope
      E: Unable to locate package python3-contextvars
```

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: manually tested

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/18452

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
@@ -145,10 +145,19 @@ salt-minion-package:
     - require:
       - file: bootstrap_repo
 
-{# These dependencies are only needed on DEB based distros, running Salt 3004, with Python version < 3.7 #}
-{# We cannot make these packages as hard depedendencies for Salt package because this is only needed for Ubuntu 18.04 #}
-{# and we only maintain a single DEB package for all DEB based distros #}
-{% if salt_minion_name == 'salt-minion' and grains['os_family'] == 'Debian' and grains['saltversioninfo'][0] >= 3004 and grains['pythonversion'][0] >= 3 and grains['pythonversion'][1] < 7 %}
+{# We must install "python3-contextvars" on DEB based distros, running Salt 3004, with Python version < 3.7, like Ubuntu 18.04 #}
+{# We cannot make this package a hard depedendency for Salt DEB package because this is only needed in Ubuntu 18.04 #}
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -145,10 +145,10 @@ salt-minion-package:
     - require:
       - file: bootstrap_repo
 
-{# These two dependencies are only needed on DEB based distros with Python version < 3.7 #}
-{# We cannot make these packages as hard depedendencies for Salt package because this is only needed for Ubuntu 18.04 #}
+{# These dependencies are only needed on DEB based distros, running Salt 3004, with Python version < 3.7 #}
+{# We cannot make these packages as hard depedendencies for Salt package because this is only needed for Ubuntu 18.04 and Debian 9 #}
```
juliogonzalez, line 149:
```suggestion
{# We cannot make these packages as hard depedendencies for Salt package because this is only needed for Ubuntu 16.04 and Debian 9 #}
```

juliogonzalez (changes requested):

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
juliogonzalez: My comment is not right either. Ubuntu 18.04 has python 3.6

juliogonzalez (approved):

vzhestkov (approved):
I would prefer to put the dependency into the package, but with dsc control it's not that simple to make it conditional for different OS versions. So it's fine to solve it this way.

meaksh:
Agree. We should definetely move to "debbuild" for building the Debian packages, so we could move this logic to OBS.

PR #5714, opened 2022-07-26
CI: fix push-packages-to-obs for the proxy-helm case

jordimassaguerpla:
fix push-packages-to-obs for the proxy-helm case.


cbosdo (commented):
👍🏻  without the test commits

PR #5733, opened 2022-08-03
Add mgrnet salt module

vzhestkov:
## What does this PR change?

Implements `mgrnet` salt module with the only function `mgrnet.dns_fqdns`.

`mgrnet.dns_fqdns` returns the list `dns_fqdns` with all found FQDNs with native DNS tools: `host` or `nslookup`.
If none of these tools available on the system, the function returns the list `fqdns` identical to the result of `network.fqdns` salt function.

## GUI diff

No difference.

## Documentation

## Test coverage
TBD
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/17929

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/src/modules/mgrnet.py
+++ b/susemanager-utils/susemanager-sls/src/modules/mgrnet.py
@@ -1,0 +1,90 @@
+"""
+Module for gathering DNS FQDNs
+"""
+
+import logging
+import re
+import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+import salt.utils.network
+
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
+    """
+    Only works on POSIX-like systems having 'host' or 'nslookup' available
+    """
+    if not (__utils__["path.which"]("host") or __utils__["path.which"]("nslookup")):
+        log.debug("Neither 'host' nor 'nslookup' is available on the system")
+    return True
+
+
+def dns_fqdns():
+    """
+    Return all known DNS FQDNs for the system by enumerating all interfaces and
+    then trying to reverse resolve them with native DNS tools
+    """
+    # Provides:
+    # dns_fqdns
+
+    grains = {}
+    fqdns = set()
+    cmd_run_all_func = __salt__["cmd.run_all"]
+    if __utils__["path.which"]("host"):
+        cmd = "host"
+        cmd_ret_regex = re.compile(r".* domain name pointer (.*)\.$")
+    elif __utils__["path.which"]("nslookup"):
+        cmd = "nslookup"
+        cmd_ret_regex = re.compile(r".*\tname = (.*)\.$")
+    else:
+        log.error("Neither 'host' nor 'nslookup' is available on the system")
+        return {"dns_fqdns": []}
+
+    def _lookup_dns_fqdn(ip):
+        try:
+            ret = cmd_run_all_func([cmd, ip], ignore_retcode=True)
+        except Exception as e:
+            log.error("Error while trying to use '%s' to resolve '%s': %s", cmd, ip, e)
+        if ret["retcode"] != 0:
+            log.debug("Unable to resolve '%s' using '%s': %s", ip, cmd, ret)
+            return []
+        fqdns = []
+        for line in ret["stdout"].split("\n"):
+            match = cmd_ret_regex.match(line)
+            if match:
+                fqdns.append(match.group(1))
+        return fqdns
+
+    start = time.time()
+
+    addresses = salt.utils.network.ip_addrs(
+        include_loopback=False, interface_data=salt.utils.network._get_interfaces()
+    )
+    addresses.extend(
+        salt.utils.network.ip_addrs6(
+            include_loopback=False, interface_data=salt.utils.network._get_interfaces()
+        )
+    )
+
+    results = []
+    try:
+        # Create a ThreadPoolExecutor to process the underlying calls
+        # to resolve DNS FQDNs in parallel.
+        with ThreadPoolExecutor(max_workers=8) as executor:
+            results = {executor.submit(_lookup_dns_fqdn, ip): ip for ip in addresses}
+            for item in as_completed(results):
+                item = item.result()
+                if item:
+                    fqdns.update(item)
+    except Exception as exc:  # pylint: disable=broad-except
+        log.error(
+            "Exception while running ThreadPoolExecutor for FQDNs resolution: %s",
+            exc,
+        )
+
+    elapsed = time.time() - start
+    log.debug("Elapsed time getting DNS FQDNs: %s seconds", elapsed)
+
+    return {"dns_fqdns": sorted(list(fqdns))}
--- a/susemanager-utils/susemanager-sls/src/modules/mgrnet.py
+++ b/susemanager-utils/susemanager-sls/src/modules/mgrnet.py
@@ -17,7 +17,7 @@ def __virtual__():
     Only works on POSIX-like systems having 'host' or 'nslookup' available
     """
     if not (__utils__["path.which"]("host") or __utils__["path.which"]("nslookup")):
-        log.debug("Neither 'host' nor 'nslookup' is available on the system")
+        return (False, "Neither 'host' nor 'nslookup' is available on the system")
     return True
 
 
--- a/susemanager-utils/susemanager-sls/src/tests/test_module_mgrnet.py
+++ b/susemanager-utils/susemanager-sls/src/tests/test_module_mgrnet.py
@@ -1,0 +1,94 @@
+import sys
+
+from unittest.mock import MagicMock, patch
+from . import mockery
+
+mockery.setup_environment()
+
+from ..modules import mgrnet
+
+
+mgrnet.__salt__ = {}
+mgrnet.__utils__ = {}
+
+
+def test_mgrnet_dns_fqdns():
+    """
+    Test getting possible FQDNs with DNS tools
+    """
+
+    check_calls = {"host": [], "nslookup": []}
+
+    ipv4_addresses = ["10.0.0.1", "172.16.0.1", "192.168.0.1", "10.10.1.1"]
+    ipv6_addresses = ["fd12:3456:789a:1::1", "fd12:abcd:1234:1::1"]
+
+    names = {
+        "10.0.0.1": "host10.example.org",
+        "172.16.0.1": "host172.example.org",
+        "192.168.0.1": "host10.example.org",
+        "fd12:3456:789a:1::1": "ipv6host3456.example.org",
+        "fd12:abcd:1234:1::1": "ipv6hostabcd.example.org",
+    }
+
+    def _cmd_run_host_nslookup(cmd, ignore_retcode=False):
+        """
+        This function is emulating the output of 'host' or 'nslookup'
+        """
+        ip = cmd[1]
+        cmd = cmd[0]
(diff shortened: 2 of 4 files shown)


susemanager-utils/susemanager-sls/src/modules/mgrnet.py
```diff
@@ -0,0 +1,92 @@
+"""
+Module for gathering DNS FQDNs
+"""
+
+import logging
+import re
+import time
+from multiprocessing.pool import ThreadPool
```
agraul, line 8: Could you share why you opted for `multiprocessing.pool.ThreadPool`? What's the advantage over `multiprocessing.Pool` and the `threading` module?

agraul (commented):
Before starting a real review, I have a question about the select concurrency mechanism used.

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
vzhestkov: I think you are asking good question. I have no idea why `multiprocessing` was used originally for `network` module. Actually this part is more or less a copy of the logic from `network` module.
I'll check with `threading` and update the PR.

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
```diff
@@ -0,0 +1,137 @@
...
+
+import logging
+import re
+import time
+from threading import Thread
+from queue import Queue
+
+import salt.utils.network
+import salt.utils.path
+import salt.utils.platform
```
aaannz, line 13: I don't know the details enough, but why is this imported when not used?

aaannz (approved):
LGTM.

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
vzhestkov: I'll check, thanks for pointing :+1: . Maybe I just forgot to remove something what was used before one of the modifications.

vzhestkov:
@aaannz please note that `bind-utils` package must be installed on the client to make `host` or `nslookup` available for the module.

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
vzhestkov: Thanks @aaannz . There was a code using this import before in `__virtual__`, but I've removed it as redundant.

nadvornik:
The fallback to network.fqdns is not needed, returning empty list is enough. We have to run both modules anyway and merge results, to cover avahi and /etc/hosts, see #5770

vzhestkov:
@nadvornik do you want it to fail or add some specific result in case if `host` and `nslookup` us missing on the client?

nadvornik:
I think that returning empty list and error message in the log is OK.

vzhestkov:
Let's do it this way: Put the debug message on loading, but let the module to be loaded to prevent the errors on calling `mgrnet.dns_fqdns`, but return the empty list and emit error log message with the details.

nadvornik:
Now it works fine, thanks.

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
```diff
@@ -0,0 +1,137 @@
...
+def __virtual__():
+    """
+    Only works on POSIX-like systems having 'host' or 'nslookup' available
+    """
+    if not (__utils__["path.which"]("host") or __utils__["path.which"]("nslookup")):
+        log.debug("Neither 'host' nor 'nslookup' is available on the system")
+    return True
+
+
+class ThreadPool:
```
agraul, line 25: Could you share your reasoning behind having this `ThreadPool`? What's the advantage over using something simple like:
```python
ips = ["1.1.1.1", "2.2.2.2", "3.3.3.3"]
threads = []
for ip in ips: # I didn't account for an upper limit to keep this brief, adding it is simple
    t = threading.Thread(target=lookup_dns_fqdn, args=(ip,))
    threads.append(t)
    t.start()

for t in threads:
    t.join()
```

Alternatively, have you considered using https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor instead of our own implementation?

agraul (changes requested):
I'm unsure about `__virtual__` always returning `True`. Why is that better than force the caller to check that it's available (`if "mrgnet" in salt:`)?

I like that we moved to `threading`, that makes it clearer that we're using threads and not processes. I think we can further simplify the implementation, see my comment below.

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
```diff
@@ -0,0 +1,137 @@
...
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
+    """
+    Only works on POSIX-like systems having 'host' or 'nslookup' available
+    """
+    if not (__utils__["path.which"]("host") or __utils__["path.which"]("nslookup")):
+        log.debug("Neither 'host' nor 'nslookup' is available on the system")
+    return True
```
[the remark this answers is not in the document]
vzhestkov: It was made to prevent fails for calling `mgrnet.dns_fqdns` if neither `host` nor `nslookup` is available on the system.

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
vzhestkov: There could be significant amount of attached interfaces on the client. Not really usual, but some of the customers have more than 10 network interfaces, so it was made just to limit the possible large list of IPs. I'll check with the docs of ThreadPoolExecutor. Thanks for pointing to it :+1:

agraul: As I wrote in the code comment, instead of using `for ip in ips` we can use a way that has an upper limit. That was not the point if the suggestion, the point was to create the X number of threads in a simple loop instead of having a class that abstracts it away.
```python
for i in range(min(8, len(ips))):
    ip = ips[i]
...
```

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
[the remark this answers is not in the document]
agraul: I don't see the use of loading this module if it then doesn't do anything. What's wrong with returning `False`, as one would expect from Salt modules?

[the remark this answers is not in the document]
vzhestkov: Here we have 2 possible options:
1. return `False`  with `virtual` in case if `host` and `nslookup` are absent on the minion. The possible drawbacks to do it this way:
  - need to add error handling in case if `mgrnet.dns_fqdns` is not available in the `__salt__` with the state
  - in case if `host` and `nslookup` become available it will take some time to reload the modules list, probably need to restart the `salt-minion` service to make the module available after installing `bind-utils` package
2. keep as it is now and just merge the values returned by `mgrnet.dns_fqdns` with the values from `network.fqdns`.

I would prefer to keep as it is now to prevent extra complexity with the state using this function.
@nadvornik what do you think? @agraul is correct here that returning `False` is the native solution for salt modules, but not sure if you are ok to modify the module this way.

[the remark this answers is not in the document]
nadvornik: The check for `mgrnet.dns_fqdns` not available is already there:
https://github.com/uyuni-project/uyuni/pull/5770/files#diff-1a0526ffc994048d3015ad1157ef8a0ee2e7b901d4cd6ba6fc950e2e593041d1R134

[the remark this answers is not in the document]
vzhestkov: @nadvornik I'm not really sure about the case when the `host` or `nslookup` is available on the system and in case if we return False with virtual function how long it could take for the module to become available after installing the tools, I'll try to check on the test system. Do you have any changes to the way how to make `bind-utils` installed on the system? State or system installation profile for example.

vzhestkov:
I've tested with `__virtual__` checking for `host` and `nslookup`, seems it works fine with installing/removing the `bind-utils` package between the attempts to call `mgrnet.dns_fqdns` function.

@agraul thanks for all the suggestions. I think I've addressed all of them, could you please check one more time with the latest changes?

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
```diff
@@ -0,0 +1,90 @@
...
+    cmd_run_all_func = __salt__["cmd.run_all"]
+    if __utils__["path.which"]("host"):
+        cmd = "host"
+        cmd_ret_regex = re.compile(r".* domain name pointer (.*)\.$")
+    elif __utils__["path.which"]("nslookup"):
+        cmd = "nslookup"
+        cmd_ret_regex = re.compile(r".*\tname = (.*)\.$")
+    else:
+        log.error("Neither 'host' nor 'nslookup' is available on the system")
+        return {"dns_fqdns": []}
```
agraul, line 43: nitpick: Given the check in `__virtual__`, I guess this is dead code now?

susemanager-utils/susemanager-sls/src/tests/test_module_mgrnet.py
```diff
@@ -0,0 +1,114 @@
...
+    def _cmd_run_host_nslookup(cmd, ignore_retcode=False):
+        """
+        This function is emulating the output of 'host' or 'nslookup'
+        """
+        ip = cmd[1]
+        cmd = cmd[0]
+        check_calls[cmd].append(ip)
+        rc = 0
+        if ":" in ip:
+            # the conversation is not very accurate here, but it's enough for testing
```
agraul, line 62: nitpick: typo in comment
```suggestion
            # the conversion is not very accurate here, but it's enough for testing
```

agraul (approved):
Thanks for all the work you put into improving this PR! It looks good overall. I commented two nitpicks, as always it's up to you to decide what to do with them

susemanager-utils/susemanager-sls/src/tests/test_module_mgrnet.py
vzhestkov: ouch, thanks :+1:

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
vzhestkov: Yes, looks redundant, but there are still some chances that `host` and `nslookup` could be removed between the moments of invoking `__virtual__` and `dns_fqdns`, so I feel safer if leave it as it is if you don't mind.

vzhestkov:
@nadvornik today is the last day to merge changes for `4.2.9`, do you want it to be backported to `4.2` and be available with this MU?

susemanager-utils/susemanager-sls/src/modules/mgrnet.py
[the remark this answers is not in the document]
nadvornik: For retail use case we need it for detecting FQDNs of a branch server. `bind-utils` is always installed there.
For other cases it might be enough to say in the documentation that the FQDN detection can be improved by installing `bind-utils` and running HW detection again.

nadvornik (approved):
LGTM

No need to hurry, IMO. It can wait till the next 4.2 MU.

PR #5735, opened 2022-08-03
Disable gpgckeck for container build

mcalmer:
## What does this PR change?

When building Docker containers we would need to deploy the GPG signing keys into the container to be able to install packages. As this requires additional work for customers to prepare the Dockerfile correctly to use another buildarg we decided that it is better to disable the GPG ckeck on the repositories we use during the building of the image.
The repository is available only during the buildtime and provide trusted content. We use SSL to access the repository to secure the connection.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- TBD

- [ ] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
+++ b/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
@@ -1574,10 +1574,8 @@ public class SaltServerActionService {
                                     "autorefresh=1\n\n" +
                                     "baseurl=" + getChannelUrl(minion, s.getLabel()) + "\n\n" +
                                     "type=rpm-md\n\n" +
-                                    "gpgcheck=1\n\n" +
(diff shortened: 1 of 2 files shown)


renner (approved):
This seems okay to me. I understand the content is "trusted", because we verify the repos with gpg when they are synced to the SUSE Manager Server. Then, because SSL is used, we can be sure to talk to the real server. I added Paolo as an additional reviewer just to keep him in the loop.

mcalmer:
@renner we should talk about this next week
@thesp0nge when I get no reaction, I think this change is ok from security standpoint.

mcalmer:
No answer so i merged for 5.0

PR #5738, opened 2022-08-03
Fix cobbler conflicts

Bischoff:
## What does this PR change?

To avoid cascading failures, during the KVM tests:

1) make sure to close the VNC window even when the Salt key was not accepted

2) choose for the auto-installable distribution a name that is not the same as the one used in Cobbler tests

## Links

Fixes SUSE/spacewalk#18532

Ports:
* 4.2:

## Changelogs

- [x] No changelog needed


nodeg (approved):

vandabarata (approved):

SchoolGuy (approved):
Wonderful! That is a very annoying one that gets fixed here!

nodeg:
@Bischoff Can we add the `merge-candidate` label? @cbosdo should be on vacation AFAIK.

PR #5747, opened 2022-08-05
Reduce hibernate deprecation

mackdk:
## What does this PR change?

This PR removes the deprecation warning generated by Hibernate by updating the code and the configuration up the the 5.3 best practices when possible. It also filter out the entries from the logs only for the warnings that could not be fixed.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/18088

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/Action_legacyUser.hbm.xml
+++ b/java/code/src/com/redhat/rhn/domain/action/Action_legacyUser.hbm.xml
@@ -652,7 +652,7 @@ select {ra.*}
     SELECT sa.server.id
         FROM ServerAction sa
         JOIN sa.server s
-        WHERE sa.server.class = com.redhat.rhn.domain.server.MinionServer AND action_id = :id
+        WHERE type(s) = com.redhat.rhn.domain.server.MinionServer AND action_id = :id
     ]]>
     </query>
     
@@ -661,7 +661,7 @@ select {ra.*}
         SELECT sa
             FROM ServerAction AS sa
                 JOIN sa.server AS s
-            WHERE s.class != com.redhat.rhn.domain.server.MinionServer
+            WHERE type(s) != com.redhat.rhn.domain.server.MinionServer
                 AND action_id = :id
     ]]>
     </query>
@@ -672,7 +672,7 @@ select {ra.*}
             FROM ServerAction AS sa
                 JOIN sa.server AS s
                 JOIN s.contactMethod AS c
-            WHERE sa.server.class = com.redhat.rhn.domain.server.MinionServer
+            WHERE type(s) = com.redhat.rhn.domain.server.MinionServer
                 AND action_id = :id
     ]]>
     </query>
--- a/java/code/src/com/redhat/rhn/domain/action/Action_legacyUser.hbm.xml
+++ b/java/code/src/com/redhat/rhn/domain/action/Action_legacyUser.hbm.xml
@@ -8,8 +8,9 @@ PUBLIC "-//Hibernate/Hibernate Mapping DTD 3.0//EN"
         discriminator-value="-1" >
         <id name="id" type="long" column="id">
             <meta attribute="scope-set">protected</meta>
-            <generator class="sequence">
-                <param name="sequence">rhn_event_id_seq</param>
+            <generator class="org.hibernate.id.enhanced.SequenceStyleGenerator">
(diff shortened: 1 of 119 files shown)


mackdk:
Acceptance test run: https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-prs-ci-tests/1519/

java/code/src/com/redhat/rhn/domain/channel/PublicChannelFamily.java
```diff
@@ -53,6 +72,7 @@ public boolean equals(final Object other) {
         }
         PublicChannelFamily otherFamily = (PublicChannelFamily) other;
         return new EqualsBuilder()
+            .append(getId(), otherFamily.getId())
```
mcalmer, line 75: Not sure if the "id" should be in the equals and hashcode method. I remember that this is a bad idea.
@moio knows more.

java/code/src/com/redhat/rhn/domain/channel/PublicChannelFamily.java
```diff
@@ -63,6 +83,7 @@ public boolean equals(final Object other) {
     @Override
     public int hashCode() {
         return new HashCodeBuilder()
+            .append(getId())
```
mcalmer, line 86: Not sure if the "id" should be in the equals and hashcode method. I remember that this is a bad idea.

java/code/src/com/redhat/rhn/domain/action/Action_legacyUser.hbm.xml
```diff
@@ -49,8 +49,7 @@ PUBLIC "-//Hibernate/Hibernate Mapping DTD 3.0//EN"
                         lazy="true" discriminator-value="5">
                         <set name="errata" outer-join="false" table="rhnActionErrataUpdate">
                                 <key column="action_id" />
-                                <many-to-many column="errata_id" outer-join="false"
-                                        class="com.redhat.rhn.domain.errata.Errata" />
+                                <many-to-many column="errata_id" class="com.redhat.rhn.domain.errata.Errata" />
```
mcalmer, line 52: Are we sure this can just be dropped? I read something about using "lazy=extra" instead.
But I could not really find a good doc about it.

java/code/src/com/redhat/rhn/domain/channel/PublicChannelFamily.java
mackdk: In this specific use case it should not matter as the `id` is the `channel_family_id`. It's basically the only column of the table and it must already exists as it is a foreign key. But anyway you have a point, and there is strictly no reason to change the behaviour of the equals/hashcode in the first place. I will remove it.

java/code/src/com/redhat/rhn/domain/channel/PublicChannelFamily.java
mackdk: See my other comment on the equals. I will fix this.

java/code/src/com/redhat/rhn/domain/action/Action_legacyUser.hbm.xml
mackdk: I thought it could be dropped because the collections containing the definition are already defined as lazy. Moreover in all other places where we have a `many-to-many`, the definition is done like this. So I don't think we are doing anything particular that could require extra laziness. To be extra sure however, I can have a deeper look and write some specific unit test.

java/code/src/com/redhat/rhn/domain/channel/PublicChannelFamily.java
mackdk: I removed the change to `equals()` and `hashcode()`.

java/code/src/com/redhat/rhn/domain/action/Action_legacyUser.hbm.xml
mackdk: I removed this commit from the PR. It will be part of a different PR with additional testing.

mcalmer (approved):

PR #5752, opened 2022-08-08
make zypper use proxy with mirrorlist

tiltingpenguin:
## What does this PR change?

Adds the configured proxy to the url when using mirrorlist in reposync

## GUI diff

No difference.

## Documentation

## Test coverage

- [x] Added test for adding proxy to url
- [x] Added test for building the url correctly

## Links

Fixes https://github.com/SUSE/spacewalk/issues/17493
Part of https://github.com/SUSE/spacewalk/issues/18523

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
@@ -662,6 +662,7 @@ type=rpm-md
 
         :returns: str
         """
+        ret_url = None
         query_params = {}
         if self.proxy_hostname:
             query_params['proxy'] = quote(self.proxy_hostname)
@@ -689,7 +690,7 @@ type=rpm-md
             return "{0}&{1}".format(url, new_query)
         parsed_url = urlparse(url)
         existing_query = parsed_url.query
-        combined_query = "&".join(filter(None, [existing_query, new_query]))
+        combined_query = "&".join([q for q in [existing_query, new_query] if q])
         return urlunparse((
             parsed_url.scheme,
             parsed_url.netloc,
--- a/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
@@ -582,6 +582,8 @@ class ContentSource:
 
         returnlist = _replace_and_check_url(returnlist)
 
+        returnlist = [self._prep_zypp_repo_url(url, False) for url in returnlist]
+
         try:
            # Write the final mirrorlist that is going to be pass to Zypper
            with open(mirrorlist_path, 'w') as mirrorlist_file:
@@ -660,7 +662,6 @@ type=rpm-md
 
         :returns: str
         """
-        ret_url = None
         query_params = {}
         if self.proxy_hostname:
             query_params['proxy'] = quote(self.proxy_hostname)
@@ -683,11 +684,20 @@ type=rpm-md
         if self.sslclientkey:
             query_params['ssl_clientkey'] = self.sslclientkey
         new_query = unquote(urlencode(query_params, doseq=True))
-        if self.authtoken or uln_repo:
-            ret_url = "{0}&{1}".format(url, new_query)
-        else:
-            ret_url = "{0}?{1}".format(url, new_query) if new_query else url
-        return ret_url
+        # urlparse cannot handle uln urls, so we need to keep this check
+        if uln_repo:
+            return "{0}&{1}".format(url, new_query)
+        parsed_url = urlparse(url)
+        existing_query = parsed_url.query
+        combined_query = "&".join(filter(None, [existing_query, new_query]))
+        return urlunparse((
+            parsed_url.scheme,
+            parsed_url.netloc,
+            parsed_url.path,
+            parsed_url.params,
+            combined_query,
+            parsed_url.fragment
+        ))
 
     def _md_exists(self, tag):
         """
--- a/python/spacewalk/spacewalk-backend.changes
+++ b/python/spacewalk/spacewalk-backend.changes
@@ -1,3 +1,4 @@
+- Make reposync use the configured http proxy with mirrorlist (bsc#1198168)
 - Prevent tracebacks on running spacewalk-repo-sync
   on loading update notice with no version specified in the meta data
 
--- a/python/test/unit/spacewalk/satellite_tools/test_yum_src.py
+++ b/python/test/unit/spacewalk/satellite_tools/test_yum_src.py
@@ -19,6 +19,7 @@ import shutil
 import os
 import solv
 import unittest
+import pytest
 from urllib.parse import quote
 import xml.etree.ElementTree as etree
 try:
@@ -66,6 +67,10 @@ class YumSrcTest(unittest.TestCase):
         yum_src.ContentSource.setup_repo = real_setup_repo
 
         return cs
+
+    @pytest.fixture(autouse=True)
+    def set_temp_path(self, tmpdir):
+        self.tmpdir = tmpdir.strpath
 
     def setUp(self):
         patch('spacewalk.satellite_tools.repo_plugins.yum_src.fileutils.makedirs').start()
@@ -197,11 +202,13 @@ class YumSrcTest(unittest.TestCase):
     @patch("spacewalk.satellite_tools.repo_plugins.yum_src.etree.parse", MagicMock(side_effect=Exception))
     def test_mirror_list_arch(self):
         cs = self._make_dummy_cs()
+        fake_mirrorlist_file = self.tmpdir + "/mirrorlist.txt"
+        self.repo.root = self.tmpdir
         cs.channel_arch = "arch1"
         grabber_mock = Mock()
 
         with patch("spacewalk.satellite_tools.repo_plugins.yum_src.urlgrabber.urlgrab", grabber_mock):
-            with open(os.path.join(self.repo.root, "mirrorlist.txt"), "w") as fake_list:
+            with open(fake_mirrorlist_file, "w") as fake_list:
                 fake_list.writelines([
                     "http://host1/base/$basearch/os/\n",
                     "http://host2/base/$BASEARCH/os/\n",
@@ -222,6 +229,8 @@ class YumSrcTest(unittest.TestCase):
     @patch("spacewalk.satellite_tools.repo_plugins.yum_src.etree.parse", MagicMock(side_effect=Exception))
     def test_proxy_usage_with_mirrorlist(self):
         cs = self._make_dummy_cs()
+        fake_mirrorlist_file = self.tmpdir + "/mirrorlist.txt"
+        self.repo.root = self.tmpdir
         proxy_url = "http://proxy.example.com:8080"
         proxy_user = "user"
         proxy_pass = "pass"
@@ -237,11 +246,11 @@ class YumSrcTest(unittest.TestCase):
             ]
         for url in url_list:
             if "?" in url:
-                query = "&"
+                separator = "&"
             else:
-                query = "?"
+                separator = "?"
             expected_url_list.append("{}{}proxy={}&proxyuser={}&proxypass={}".format(url,
-                                                                                     query,
+                                                                                     separator,
                                                                                      quote(proxy_url),
                                                                                      proxy_user,
                                                                                      proxy_pass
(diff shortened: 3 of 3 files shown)


python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -681,7 +683,8 @@ def _prep_zypp_repo_url(self, url, uln_repo):
         if self.sslclientkey:
             query_params['ssl_clientkey'] = self.sslclientkey
         new_query = unquote(urlencode(query_params, doseq=True))
-        if self.authtoken or uln_repo:
+        query_check = urlparse(url)
+        if (query_check.query != "") or uln_repo:
```
vzhestkov, line 687: I'm not really sure about this part. Why not just let `urlunparse` do it's job to compose the URL back?

tiltingpenguin: You're right, rewriting this part to use unparse might be the better way to go than keeping these checks with weird logic.  I'll change it

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -681,7 +683,8 @@ def _prep_zypp_repo_url(self, url, uln_repo):
...
             ret_url = "{0}&{1}".format(url, new_query)
         else:
             ret_url = "{0}?{1}".format(url, new_query) if new_query else url
```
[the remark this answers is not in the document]
tiltingpenguin: I obsoleted this logic with urlunparse now

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -681,13 +683,19 @@ def _prep_zypp_repo_url(self, url, uln_repo):
...
+        return urlunparse((
+            parsed_url.scheme,
+            parsed_url.netloc,
+            parsed_url.path,
+            parsed_url.params,
+            final_query,
+            parsed_url.fragment
+        ))
+
+def _md_exists(self, tag):
```
vzhestkov, line 698: ouch, are you sure it should move here?

tiltingpenguin: oops, fixed it

tiltingpenguin:
I just noticed I still have some errors in my tests. Going to fix them quickly

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
```diff
@@ -213,6 +214,66 @@ def test_mirror_list_arch(self):
...
+        proxy_url = "http://proxy.example.com:8080"
+        proxy_user = "user"
+        proxy_pass = "pass"
+        cs.proxy_hostname = proxy_url
+        cs.proxy_user = proxy_user
+        cs.proxy_pass = proxy_pass
+
+        for i in range(0, len(url_list)):
+            if not uln_list[i]:
+                expected_url = "{}?proxy={}&proxyuser={}&proxypass={}".format(url_list[i], quote(proxy_url), proxy_user, proxy_pass)
```
vzhestkov, line 269: Not sure here about the way of checking ULN. I supposed ULN just must have some query parameters, that's why it was set with `&`, but not `?`. Am I wrong?

tiltingpenguin: No, you are right, but this is the negative branch, so it should be right

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
```diff
@@ -213,6 +214,66 @@ def test_mirror_list_arch(self):
...
+        proxy_pass = "pass"
+        cs.proxy_hostname = proxy_url
+        cs.proxy_user = proxy_user
+        cs.proxy_pass = proxy_pass
+        expected_url_list = []
+        url_list = [
+            "http://host1/base/arch1/os/",
+            "http://host2/",
+            "http://host3.com/",
+            ]
```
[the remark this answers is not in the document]
tiltingpenguin: Ok, I'll add it

python/spacewalk/spacewalk-backend.changes
```diff
@@ -1,3 +1,4 @@
+- Make reposync always use the configured http proxy (bsc#1198168)
```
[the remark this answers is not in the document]
tiltingpenguin: No problem, I will update it

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
tiltingpenguin: Or do you mean I shouldn't decide if a url is uln manually?

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -683,11 +684,20 @@ def _prep_zypp_repo_url(self, url, uln_repo):
...
-            ret_url = "{0}&{1}".format(url, new_query)
-        else:
-            ret_url = "{0}?{1}".format(url, new_query) if new_query else url
-        return ret_url
+        # urlparse cannot handle uln urls, so we need to keep this check
+        if uln_repo:
+            return "{0}&{1}".format(url, new_query)
+        parsed_url = urlparse(url)
+        query = parsed_url.query
+        final_query = "&".join(filter(None, [query, new_query]))
```
agraul, line 692: nitpick: I'd prefer a more verbose way than `filter(None, iterable)`, because it is a special case for `filter(function, iterable)` that I had to look up to understand. With list comprehensions, `filter` is not used that much anymore and not as intuitive.

agraul, line 692: The names `query`, `new_query` and `final_query` are a bit confusing, it's unclear what the purpose of each of them is. Naming is hard, but I think we can come up with better fitting names here.

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
```diff
@@ -213,6 +214,85 @@ def test_mirror_list_arch(self):
...
+            expected_url_list.append("{}{}proxy={}&proxyuser={}&proxypass={}".format(url,
+                                                                                     query,
+                                                                                     quote(proxy_url),
+                                                                                     proxy_user,
+                                                                                     proxy_pass
+                                                                                     ))
+        grabber_mock = Mock()
+
+        with patch("spacewalk.satellite_tools.repo_plugins.yum_src.urlgrabber.urlgrab", grabber_mock):
+            with open(os.path.join(self.repo.root, "mirrorlist.txt"), "w") as fake_list:
```
agraul, line 252: I would put the `mirrorlist.txt` into a [temporary directory ](https://docs.pytest.org/en/6.2.x/tmpdir.html#temporary-directories-and-files)that get's cleaned up by pytest afterwards

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
```diff
@@ -213,6 +214,85 @@ def test_mirror_list_arch(self):
...
+        uln_list = [False, False, False, True, True]
+        proxy_url = "http://proxy.example.com:8080"
+        proxy_user = "user"
+        proxy_pass = "pass"
+        cs.proxy_hostname = proxy_url
+        cs.proxy_user = proxy_user
+        cs.proxy_pass = proxy_pass
+
+        for i in range(0, len(url_list)):
+            if uln_list[i] or "?" in url_list[i]:
```
agraul, line 280: nitpick: I would put the url and the "is uln" boolean into a tuple instead of spreading them into distinct lists, e.g.
```python
urls = [("https://example.com/", False)]
...
for url, is_uln in urls:
    if is_uln or "?" in url":
        separator = "&"
    else:
        separator = "?"
```

agraul (changes requested):
I haven't tried this myself, but it looks like it should work.

I don't like adding even more complexity to `_prep_zypp_repo_url`, it does a lot of things which makes it hard to understand. At the very least, I think we should improve naming the variables, but IMO it has reached a point where we should refactor it.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
tiltingpenguin: Well, query is the query that already is in the url. The new query is the one we add in this method and is composed of the proxy and ssl certificates. I can see that we should change final_query, but the other two are just that. A query and a new query we add to the url.

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
tiltingpenguin: Good point, I'll change that

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
vzhestkov: Sounds like you can just reuse `new_query` here.

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
tiltingpenguin: I modeled this part after the existing test `test_mirror_list_arch`. So I'll change it in both tests to be consistent

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
tiltingpenguin: Would something like `[q in [query, new_query] if q != None]` be better?

tiltingpenguin: I tried it out and it seems the behavior is different with list comprehensions. If there is no query already present in the url it still puts a `&` at the start. Filter doesn't do that and that is also the desired behavior

tiltingpenguin: I think if I rename them into `existing_query`, `new_query` and `combined_query` then it should be clear

agraul: That's because `filter(None, iterable)` is **special** and exactly the reason I don't like it.
The equivalent list comprehension is `[ q for q in [query, new_query] if q]`:
```python
>>> list(filter(None, ['', "foo"]))
['foo']
>>> [x for x in ['', "foo"] if x]
['foo']
```

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
tiltingpenguin: It seems like this doesn't work as easily in uyuni. Probably because we use an older version of pytest. At least my testing showed that all the fixtures return tuples instead of path or pathlike objects like the documentation says.

SchoolGuy (approved):
So in some areas the code is formatted a littlebit odd but other then that I see no issue with the bugfix at the moment.

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
```diff
@@ -66,6 +68,10 @@ def _make_dummy_cs(self):
 
         return cs
 
+    @pytest.fixture(autouse=True)
+    def set_temp_path(self, tmpdir):
+        self.tmpdir = tmpdir.strpath
```
agraul, line 73: I don't like this "additional `setUp` function", why is it needed? Isn't it possible to use the `tmpdir` fixture as lined out in the [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures)

agraul (changes requested):
There is just a small thing about the test fixture, other than that I'm okay with the changes

python/test/unit/spacewalk/satellite_tools/test_yum_src.py
agraul: I just read more about this topic, it's true that `unittest.TestCase` type tests request fixtures ([docs](https://docs.pytest.org/en/7.1.x/how-to/unittest.html?highlight=unittest#pytest-features-in-unittest-testcase-subclasses)) or access them directly. Without migrating away from `unittest.TestCase` I currently see no way to improve this, but that's outside of the scope of this PR.

agraul (approved):

PR #5754, opened 2022-08-09
The PXE boot minion is now SLE 15 SP4

Bischoff:
## What does this PR change?

This PR fixes the Cobbler tests. The PXE boot minion could not choose a SLE 15 SP2 entry in the boot menu that did not exist anymore.

It also removes quotes around something that is not a "string variable" but part of the step's name.

## Links

No ports, Uyuni and 4.3 only.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -616,7 +616,7 @@ When(/^I synchronize the tftp configuration on the proxy with the server$/) do
   raise 'cobbler sync failed' if out.include? 'Push failed'
 end
 
-When(/^I set the default PXE menu entry to the "([^"]*)" on the "([^"]*)"$/) do |entry, host|
+When(/^I set the default PXE menu entry to the (target profile|local boot) on the "([^"]*)"$/) do |entry, host|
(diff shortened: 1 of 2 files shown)


srbarrios (approved):

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -616,7 +616,7 @@
   raise 'cobbler sync failed' if out.include? 'Push failed'
 end
 
-When(/^I set the default PXE menu entry to the "([^"]*)" on the "([^"]*)"$/) do |entry, host|
+When(/^I set the default PXE menu entry to the ([^ ]*) on the "([^"]*)"$/) do |entry, host|
```
srbarrios, line 619: Just a suggestion, but approving anyway.

```suggestion
When(/^I set the default PXE menu entry to the (target profile|local boot) on the "([^"]*)"$/) do |entry, host|
```

Bischoff: right, much more legible

nodeg (approved):

PR #5758, opened 2022-08-09
Remove obsolete XML tags from cobbler profile

Bischoff:
## What does this PR change?

This PR removes obsolete XML tags from SLE 15 SP4 cobbler profile.

See https://bugzilla.suse.com/show_bug.cgi?id=1201600.

## Links

No ports, Uyuni/4.3 only.

## Changelogs

- [x] No changelog needed


nodeg (approved):
Nice catch with the `<image>` that is actually not used.

aaannz:
See also https://github.com/SUSE/manager-build-profiles/pull/32 for the rest of changes, particularly switch to salt bundle.

PR #5772, opened 2022-08-12
Action Chain: reboot on transactional update minions

mbussolotto:
## What does this PR change?
Currently the reboot of a transactional update minions does not work inside in a action chains. This is because all the actions are converted in a sls file, and this file is executed as a single transaction: since reboot is not allowed inside a transaction, the action chains stop to work.

In order to have the reboot action working:
-  more than one sls file are created, each one of them contains all the action until a reboot (if present in the action chain)
- the reboot action is not converted in an sls file, but it managed directly by `mgractionchains.py` (using the argument `activate_transaction=true`)

**KNOWN ISSUE:**

- If action does not generate any new snapshot, the reboot action would not work (since there's no reboot action but an `activate_transaction=true` ) . A possible solution is to perform a dummy action to force the creation of a new snapshot, but not sure about the dummy action to execute. 

- the approach seems it's not working if with `REBOOT_METHOD=auto` (in `/etc/transactional-update.conf` ) , but the reason is on transaction-update module itself. It work as expected with `REBOOT_METHOD=systemd`.  **(UPDATE: after some debug, I opened this issue: https://github.com/openSUSE/transactional-update/issues/92)** **(UPDATE: the issue was caused by a misleading log, it works as expected)**

- although reboot is performed, WebUI `events -> pending` still show that the reboot action was not executed. No ideaabout the reason. **(UPDATE: FIXED)**

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- Doc changes are required. DOC Issue here: https://github.com/uyuni-project/uyuni-docs/issues/1843

- [x] **DONE**

## Test coverage
- Added unit test

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/18381

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
+++ b/java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
@@ -148,7 +148,8 @@ public class SaltActionChainGeneratorService {
             if (isSaltUpgrade(state)) {
                 fileStates.add(
                         endChunk(actionChain, chunk, nextActionId,
-                                prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.empty()));
+                                prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.empty(),
+                                Optional.empty()));
                 fileStates.add(state);
                 fileStates.add(stopIfPreviousFailed(prevRequisiteRef(fileStates)));
                 fileStates.add(forceRestartServiceIfNeeded("force_restart_if_needed",
@@ -163,10 +164,11 @@ public class SaltActionChainGeneratorService {
                     if (i < states.size() - 1) {
                         fileStates.add(
                                 endChunk(actionChain, chunk, nextActionId,
-                                        prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.of(true)));
+                                        prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.of(true),
+                                        currentActionId(states, i)));
                     }
                     else {
-                        fileStates.add(endLastChunk(Optional.of(true)));
+                        fileStates.add(endLastChunk(actionChain, currentActionId(states, i), Optional.of(true)));
                     }
                 }
                 else { // it's not transactional update
@@ -174,7 +176,8 @@ public class SaltActionChainGeneratorService {
                     if (i < states.size() - 1) {
                         fileStates.add(
                                 endChunk(actionChain, chunk, nextActionId,
-                                        prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.empty()));
+                                        prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.empty(),
+                                        Optional.empty()));
                     }
                 }
 
@@ -209,14 +212,26 @@ public class SaltActionChainGeneratorService {
         return Optional.empty();
     }
 
+    private Optional<Long> currentActionId(List<SaltState> states, int currentPos) {
+        SaltState state = states.get(currentPos);
+        if (state instanceof ActionSaltState) {
+            ActionSaltState actionState = (ActionSaltState)state;
+            return Optional.of(actionState.getActionId());
+        }
+        return Optional.empty();
+    }
+
+
     private SaltState endChunk(ActionChain actionChain, int chunk, Optional<Long> nextActionId,
                                Optional<Pair<String, String>> lastRef, Optional<String> sshExtraFileRefs,
-                               Optional<Boolean> rebootRequired) {
+                               Optional<Boolean> rebootRequired, Optional<Long> currentActionId) {
         Map<String, Object> args = new LinkedHashMap<>(2);
         args.put("actionchain_id", actionChain.getId());
         args.put("chunk", chunk + 1);
         nextActionId.ifPresent(actionId ->
                 args.put("next_action_id", actionId));
+        currentActionId.ifPresent(actionId ->
+                args.put("current_action_id", actionId));
         sshExtraFileRefs.ifPresent(refs ->
                 args.put("ssh_extra_filerefs", refs));
         rebootRequired.ifPresent(refs ->
@@ -226,8 +241,12 @@ public class SaltActionChainGeneratorService {
         return modRun;
     }
 
-    private SaltState endLastChunk(Optional<Boolean> rebootRequired) {
+    private SaltState endLastChunk(ActionChain actionChain, Optional<Long> currentActionId,
+            Optional<Boolean> rebootRequired) {
         Map<String, Object> args = new LinkedHashMap<>(2);
+        args.put("actionchain_id", actionChain.getId());
+        currentActionId.ifPresent(actionId ->
+                args.put("current_action_id", actionId));
         rebootRequired.ifPresent(refs ->
                 args.put("reboot_required", refs));
         SaltModuleRun modRun = new SaltModuleRun("schedule_next_chunk", "mgractionchains.clean", args);
--- a/java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
+++ b/java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
@@ -145,32 +145,43 @@ public class SaltActionChainGeneratorService {
             }
             Optional<Long> nextActionId = nextActionId(states, i);
 
-            if (mustSplit(state, minion)) {
-                if (isSaltUpgrade(state)) {
-                    fileStates.add(
-                            endChunk(actionChain, chunk, nextActionId,
-                                    prevRequisiteRef(fileStates), sshExtraFileRefs));
-                    fileStates.add(state);
-                    fileStates.add(stopIfPreviousFailed(prevRequisiteRef(fileStates)));
-                    fileStates.add(forceRestartServiceIfNeeded("force_restart_if_needed",
-                            prevRequisiteRef(Collections.singletonList(state))));
-                    saveChunkSLS(fileStates, minion, actionChain.getId(), chunk);
-                    fileStates.clear();
-                    chunk++;
-                    fileStates.add(checkSaltUpgradeChunk(state));
+            if (isSaltUpgrade(state)) {
+                fileStates.add(
+                        endChunk(actionChain, chunk, nextActionId,
+                                prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.empty()));
+                fileStates.add(state);
+                fileStates.add(stopIfPreviousFailed(prevRequisiteRef(fileStates)));
+                fileStates.add(forceRestartServiceIfNeeded("force_restart_if_needed",
+                        prevRequisiteRef(Collections.singletonList(state))));
+                saveChunkSLS(fileStates, minion, actionChain.getId(), chunk);
+                fileStates.clear();
+                chunk++;
+                fileStates.add(checkSaltUpgradeChunk(state));
+            }
+            else if (isRebootAction(state)) {
+                if (minion.isTransactionalUpdate()) {
+                    if (i < states.size() - 1) {
+                        fileStates.add(
+                                endChunk(actionChain, chunk, nextActionId,
+                                        prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.of(true)));
+                    }
+                    else {
+                        fileStates.add(endLastChunk(Optional.of(true)));
+                    }
                 }
-                else {
+                else { // it's not transactional update
                     fileStates.add(state);
                     if (i < states.size() - 1) {
                         fileStates.add(
                                 endChunk(actionChain, chunk, nextActionId,
-                                        prevRequisiteRef(fileStates), sshExtraFileRefs));
+                                        prevRequisiteRef(fileStates), sshExtraFileRefs, Optional.empty()));
                     }
-
-                    saveChunkSLS(fileStates, minion, actionChain.getId(), chunk);
-                    chunk++;
-                    fileStates.clear();
                 }
+
+                saveChunkSLS(fileStates, minion, actionChain.getId(), chunk);
+
+                chunk++;
+                fileStates.clear();
             }
             else {
                 fileStates.add(state);
@@ -199,7 +210,8 @@ public class SaltActionChainGeneratorService {
     }
 
     private SaltState endChunk(ActionChain actionChain, int chunk, Optional<Long> nextActionId,
-                               Optional<Pair<String, String>> lastRef, Optional<String> sshExtraFileRefs) {
+                               Optional<Pair<String, String>> lastRef, Optional<String> sshExtraFileRefs,
+                               Optional<Boolean> rebootRequired) {
         Map<String, Object> args = new LinkedHashMap<>(2);
         args.put("actionchain_id", actionChain.getId());
         args.put("chunk", chunk + 1);
@@ -207,8 +219,18 @@ public class SaltActionChainGeneratorService {
                 args.put("next_action_id", actionId));
         sshExtraFileRefs.ifPresent(refs ->
                 args.put("ssh_extra_filerefs", refs));
+        rebootRequired.ifPresent(refs ->
+                args.put("reboot_required", refs));
         SaltModuleRun modRun = new SaltModuleRun("schedule_next_chunk", "mgractionchains.next", args);
         lastRef.ifPresent(ref -> modRun.addRequire(ref.getKey(), ref.getValue()));
+        return modRun;
+    }
+
+    private SaltState endLastChunk(Optional<Boolean> rebootRequired) {
+        Map<String, Object> args = new LinkedHashMap<>(2);
+        rebootRequired.ifPresent(refs ->
+                args.put("reboot_required", refs));
+        SaltModuleRun modRun = new SaltModuleRun("schedule_next_chunk", "mgractionchains.clean", args);
         return modRun;
     }
 
@@ -317,31 +339,42 @@ public class SaltActionChainGeneratorService {
         return false;
     }
 
-    private boolean mustSplit(SaltState state, MinionSummary minion) {
-        boolean split = false;
+    private boolean isRebootAction(SaltState state) {
         if (state instanceof SaltModuleRun) {
-            SaltModuleRun moduleRun = (SaltModuleRun)state;
-
-            Optional<String> mods = getModsString(moduleRun);
-
-            if (mods.isPresent() &&
-                    mods.get().contains(PACKAGES_PKGINSTALL) && isSaltUpgrade(state) &&
-                    !minion.isSshPush()) {
-                // split only for regular minions, salt-ssh minions don't have a salt-minion process
-                split = true;
-            }
-            else if (mods.isPresent() &&
-                    mods.get().contains(PACKAGES_PATCHINSTALL) && isSaltUpgrade(state)) {
-                split = true;
-            }
-            else if ("system.reboot".equalsIgnoreCase(moduleRun.getName())) {
-                split = true;
+            SaltModuleRun moduleRun = (SaltModuleRun) state;
+            if ("system.reboot".equalsIgnoreCase(moduleRun.getName())) {
+                return true;
             }
         }
         else if (state instanceof SaltSystemReboot) {
-            split = true;
+            return true;
         }
-        return split;
+        return false;
+    }
+
+    private boolean isUpgradeWithRebootRequired(SaltState state, MinionSummary minion) {
+        if (state instanceof SaltModuleRun) {
+            SaltModuleRun moduleRun = (SaltModuleRun) state;
+
+            Optional<String> mods = getModsString(moduleRun);
+
+            if (mods.isPresent() && mods.get().contains(PACKAGES_PKGINSTALL) &&
+                    isSaltUpgrade(state) && !minion.isSshPush()) {
+                // split only for regular minions, salt-ssh minions don't have a
+                // salt-minion process
+                return true;
+            }
+            else if (mods.isPresent() && mods.get().contains(PACKAGES_PATCHINSTALL) &&
+                    isSaltUpgrade(state)) {
+                return true;
+            }
+
+        }
+        return false;
+    }
+
+    private boolean mustSplit(SaltState state, MinionSummary minion) {
+        return isUpgradeWithRebootRequired(state, minion) || isRebootAction(state);
     }
 
     private Optional<String> getModsString(SaltModuleRun moduleRun) {
--- a/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
+++ b/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
@@ -161,6 +161,7 @@ import com.suse.utils.Opt;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.reflect.TypeToken;
 
@@ -757,8 +758,7 @@ public class SaltServerActionService {
                                 rebootServerAction.ifPresentOrElse(
                                         ract -> {
                                             if (ract.getStatus().equals(ActionFactory.STATUS_QUEUED)) {
-                                                rebootServerAction.get().setStatus(ActionFactory.STATUS_PICKED_UP);
-                                                rebootServerAction.get().setPickupTime(new Date());
+                                                setActionAsPickedUp(ract);
                                             }
                                         },
                                         () -> LOG.error("Action of type {} found in action chain result but not " +
@@ -2537,9 +2537,9 @@ public class SaltServerActionService {
                             saltUtils.updateServerAction(sa, 0L, true, "n/a", jsonResult,
                                     Optional.of(Xor.right(function)));
                         }
+
                         else if (sa.getStatus().equals(ActionFactory.STATUS_QUEUED)) {
-                            sa.setStatus(ActionFactory.STATUS_PICKED_UP);
-                            sa.setPickupTime(new Date());
+                            setActionAsPickedUp(sa);
                         }
 
                         // Perform a "check-in" after every executed action
@@ -2635,6 +2635,18 @@ public class SaltServerActionService {
     }
 
     /**
+     * In action chains at this point the action is still queued so we have
+     * to set it to picked up.
+     * This could still lead to the race condition on when event processing is slow.
+     *
+     */
+    private void setActionAsPickedUp(ServerAction sa) {
+        sa.setStatus(ActionFactory.STATUS_PICKED_UP);
+        sa.setPickupTime(new Date());
+        return;
+    }
+
+    /**
      * Update the action properly based on the Job results from Salt.
      *
      * @param actionId the ID of the Action to handle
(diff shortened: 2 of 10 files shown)


mbussolotto:
@meaksh just as a reminder, thanks!

mbussolotto:
@meaksh I tried several approaches and this one seems fine (from the first test, more tests are required). Shortly:

1. I removed the entire `transactional_update` logic from mgractionchains.py
2. when the sls files is created from from the actions ( `convertToState` ) if it's a `transactional_update` minion, it used `transactional_update.apply` instead of `state.apply`.

What do you think about this approach? do you see any side effect?

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -160,6 +156,11 @@ def next(actionchain_id, chunk, next_action_id=None, ssh_extra_filerefs=None):
         yaml_dict['ssh_extra_filerefs'] = ssh_extra_filerefs
     _persist_next_ac_chunk(yaml_dict)
 
+def reboot(actionchain_id, chunk, next_action_id=None, ssh_extra_filerefs=None):
+    ret = __salt__['rebootmgr.reboot']()
```
agraul, line 160: We can't assume that `rebootmgr` is used, in fact we'll probably be setting SLE Micro clients to not use `rebootmgr` and instead use `systemd` as the 'reboot manager' as discussed in https://github.com/uyuni-project/uyuni-rfc/pull/74.

To make this generic, this should just be 
```suggestion
    ret = __salt__['transactional_update.reboot']()
```

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -160,6 +156,11 @@ def next(actionchain_id, chunk, next_action_id=None, ssh_extra_filerefs=None):
...
+    next(actionchain_id, chunk, next_action_id, ssh_extra_filerefs)
```
agraul, line 161: A bit off-topic: when reading this, I thought "that does not look right, `next` requires an iterator and is not called like that". Then I realized this is not the [built-in `next()`](https://docs.python.org/3/library/functions.html#next), and we shadow its name. We should address that, but that can be in a different PR.

agraul (changes requested):
@mbussolotto could you elaborate on the call-chain? Forgive me that I haven't looked deep into the code yet. How is a reboot triggered on SLE Micro when there are two actions, a package installation followed by a reboo? And how are the two reboots triggered when there are 4 actions, package installation, reboot, another package installation, reboot?

It would be great if you elaborate on that a bit. Other than this fundamental question, I have requested a change in `mgractionchains.reboot`, please see the comment in-line.

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
mbussolotto: oh actually with the new implementation this function in not call at all. I'll remove it.

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
```diff
@@ -987,7 +986,7 @@ private List<SaltState> convertToState(long actionChainId, ServerAction serverAc
                         }
                         else {
                             return new SaltModuleRun(stateId,
-                                    "state.apply",
+                                    minion.isTransactionalUpdate() ? "transactional_update.apply" : "state.apply",
```
vzhestkov, line 989: Why do we need to do it explicitly? Transactional executor should translate state.apply to transactional_update.apply automatically.

mbussolotto:
@agraul @vzhestkov thanks for your feedback. Currently I found some issues in the current implementation, so I'm working on them. As soon I'll solve the problem, I'll provide some of the sls file generated to explain better the workflow (comparing them to the curent generated one)

mbussolotto:
This version works almost fine (there is still something it's not working, but I need some suggestions to fix them :) ). 
Shortly:

- for a transactional update minion, `spacewalk-java` generates a new chunk after a reboot action (the reboot action is not converted into the chunk). If at the end of a chunk there's a reboot action, this information is stored in the sls file `schedule_next_chunk -> mgrcompat.module_run -> reboot_required` 
- for a transactional update minion, `mgractionchains.start` retrieves sls file from the master and check for `schedule_next_chunk -> mgrcompat.module_run -> reboot_required` : if it's `true` the chunk is run using `activate_transaction=true`.
-  `mgractionchains.next` persists  `reboot_required` value on `_mgractionchains.conf` (already used for other info) 
- `mgractionchains.resume` use information in `_mgractionchains.conf` to start the new chunk in the correct wait ( if  `reboot_required=true` -> `activate_transaction=true` )

I put some known issues in the description, I'll paste here. Any suggestion is welcome :D

**KNOWN ISSUE:**
- If action does not generate any new snapshot, the reboot action would not work (since there's no reboot action but an `activate_transaction=true `) . A possible solution is to perform a dummy action to force the creation of a new snapshot, but not sure about the dummy action to execute.

- the approach seems it's not working if  `REBOOT_METHOD=auto` in `/etc/transactional-update.conf` , but the reason is on transaction-update module itself. It work as expected with `REBOOT_METHOD=systemd`.

- although reboot is performed, WebUI `events -> pending` still show that the reboot action was not executed. No idea about the reason.

@agraul some example that should help to understand the workflow 

**EXAMPLE 1 : package installation and reboot**
_Non transaction update system_
```
mgr_actionchain_2_action_52_chunk_1:
    mgrcompat.module_run:
    -   name: state.apply
    -   mods:
        - packages.pkginstall
    -   kwargs:
            pillar:
                param_pkgs:
                -   - adcli
                    - x86_64
                    - 0.8.2-150400.15.8
            queue: true
mgr_actionchain_2_action_53_chunk_1:
    mgrcompat.module_run:
    -   name: system.reboot
    -   at_time: 3
    -   require:
        -   mgrcompat: mgr_actionchain_2_action_52_chunk_1
```

_Transaction update system_
```
mgr_actionchain_6_action_27_chunk_1:
    mgrcompat.module_run:
    -   name: state.apply
    -   mods:
        - packages.pkginstall
    -   kwargs:
            pillar:
                param_pkgs:
                -   - aws-cli
                    - noarch
                    - 1.20.7-30.3.1
            queue: true
schedule_next_chunk:
    mgrcompat.module_run:
    -   name: mgractionchains.clean
    -   reboot_required: true
```
So the package installation is performed in one snapshot: at the end, the snapshot is activated (reboot).

**EXAMPLE 2 : package installation, reboot, package installation, reboot**
_Non transaction update system_
1st chunk:
```
mgr_actionchain_3_action_54_chunk_1:
    mgrcompat.module_run:
    -   name: state.apply
    -   mods:
        - packages.pkginstall
    -   kwargs:
            pillar:
                param_pkgs:
                -   - accel-config
                    - x86_64
                    - 3.4.2-150400.1.8
            queue: true
mgr_actionchain_3_action_55_chunk_1:
    mgrcompat.module_run:
    -   name: system.reboot
    -   at_time: 3
    -   require:
        -   mgrcompat: mgr_actionchain_3_action_54_chunk_1
schedule_next_chunk:
    mgrcompat.module_run:
    -   name: mgractionchains.next
    -   actionchain_id: 3
    -   chunk: 2
    -   next_action_id: 56
    -   require:
        -   mgrcompat: mgr_actionchain_3_action_55_chunk_1
```
2nd chunk
```
mgr_actionchain_3_action_56_chunk_2:
    mgrcompat.module_run:
    -   name: state.apply
    -   mods:
        - packages.pkginstall
    -   kwargs:
            pillar:
                param_pkgs:
                -   - acpica
                    - x86_64
                    - 20180105-1.23
            queue: true
mgr_actionchain_3_action_57_chunk_2:
    mgrcompat.module_run:
    -   name: system.reboot
    -   at_time: 3
    -   require:
        -   mgrcompat: mgr_actionchain_3_action_56_chunk_2
```
_Transaction update system_
1st chunk
```
mgr_actionchain_8_action_37_chunk_1:
    mgrcompat.module_run:
    -   name: state.apply
    -   mods:
        - packages.pkginstall
    -   kwargs:
            pillar:
                param_pkgs:
                -   - bluez-firmware
                    - x86_64
                    - 1.2-1.23
            queue: true
schedule_next_chunk:
    mgrcompat.module_run:
    -   name: mgractionchains.next
    -   actionchain_id: 8
    -   chunk: 2
    -   next_action_id: 39
    -   reboot_required: true
    -   require:
        -   mgrcompat: mgr_actionchain_8_action_37_chunk_1
```
2nd chunk
```
mgr_actionchain_8_action_39_chunk_2:
    mgrcompat.module_run:
    -   name: state.apply
    -   mods:
        - packages.pkginstall
    -   kwargs:
            pillar:
                param_pkgs:
                -   - apparmor-parser
                    - x86_64
                    - 2.13.6-150300.3.15.1
            queue: true
schedule_next_chunk:
    mgrcompat.module_run:
    -   name: mgractionchains.clean
    -   reboot_required: true
 ```
Since the information of the first chunk are persisted in `_mgractionchains.conf`, also the second chunk is run using `activate_transaction=true` .

File: java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java [hunk 4]
```diff
@@ -199,19 +210,30 @@ private Optional<Long> nextActionId(List<SaltState> states, int currentPos) {
...
+                               Optional<Boolean> rebootRequired) {
         Map<String, Object> args = new LinkedHashMap<>(2);
         args.put("actionchain_id", actionChain.getId());
         args.put("chunk", chunk + 1);
         nextActionId.ifPresent(actionId ->
                 args.put("next_action_id", actionId));
         sshExtraFileRefs.ifPresent(refs ->
                 args.put("ssh_extra_filerefs", refs));
+        rebootRequired.ifPresent(refs ->
+                args.put("reboot_required", refs));
```
mcalmer, line 223: if we do not check for "true" or "false", why do we need an Optional?
Could we just use "boolean" and set `True` and `False` ?
In this case it is impossible to not set this value.

java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
```diff
@@ -199,19 +210,30 @@ private Optional<Long> nextActionId(List<SaltState> states, int currentPos) {
...
         SaltModuleRun modRun = new SaltModuleRun("schedule_next_chunk", "mgractionchains.next", args);
         lastRef.ifPresent(ref -> modRun.addRequire(ref.getKey(), ref.getValue()));
         return modRun;
     }
 
+    private SaltState endLastChunk(Optional<Boolean> rebootRequired) {
+        Map<String, Object> args = new LinkedHashMap<>(2);
+        rebootRequired.ifPresent(refs ->
+                args.put("reboot_required", refs));
```
mcalmer, line 232: Same as above

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -138,7 +162,7 @@ def start(actionchain_id):
         raise CommandExecutionError(ret)
     return ret
 
-def next(actionchain_id, chunk, next_action_id=None, ssh_extra_filerefs=None):
+def next(actionchain_id, chunk, next_action_id=None, ssh_extra_filerefs=None, reboot_required=None):
```
mcalmer, line 165: Maybe set the default value for reboot_required to "False" ?

mbussolotto: yes, changed :)

java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
mbussolotto: Since this parameter is  used just for transactional update minion, I thought it was misleading to have `reboot_required: False` for non-transactional update system (since it's never evaluated). But I don't have a strong opinion about that.
Probably the best would be always set `reboot_required` just transactional update system, and use `Optional.empty()` for traditional system , what do you think?

mbussolotto:
Some update:

> **KNOWN ISSUE:**
> 
> * If action does not generate any new snapshot, the reboot action would not work (since there's no reboot action but an `activate_transaction=true `) . A possible solution is to perform a dummy action to force the creation of a new snapshot, but not sure about the dummy action to execute.

Although it's a bug, it should be a real use case. Need to understand how to manage it, also considering the new changes on SLE Micro reboot on uyuni (cc @wweellddeerr )

> * the approach seems it's not working if  `REBOOT_METHOD=auto` in `/etc/transactional-update.conf` , but the reason is on transaction-update module itself. It work as expected with `REBOOT_METHOD=systemd`.

I'll investigate on it to understand the root cause of the issue (salt, transactional update or rebootmgr)

> * although reboot is performed, WebUI `events -> pending` still show that the reboot action was not executed. No idea about the reason.

The error is probably in `SaltServerActionService` : the reboot action needs to be picked up. I'm currently work on it

mbussolotto:
> > * although reboot is performed, WebUI `events -> pending` still show that the reboot action was not executed. No idea about the reason.
> 
> The error is probably in `SaltServerActionService` : the reboot action needs to be picked up. I'm currently work on it

fixed with last commit

mbussolotto:
> > * the approach seems it's not working if  `REBOOT_METHOD=auto` in `/etc/transactional-update.conf` , but the reason is on transaction-update module itself. It work as expected with `REBOOT_METHOD=systemd`.
> 
> I'll investigate on it to understand the root cause of the issue (salt, transactional update or rebootmgr)
> 

After some investigation, I opened this issue: https://github.com/openSUSE/transactional-update/issues/92

mbussolotto:
> > > * the approach seems it's not working if  `REBOOT_METHOD=auto` in `/etc/transactional-update.conf` , but the reason is on transaction-update module itself. It work as expected with `REBOOT_METHOD=systemd`.
> > 
> > 
> > I'll investigate on it to understand the root cause of the issue (salt, transactional update or rebootmgr)
> 
> After some investigation, I opened this issue: [openSUSE/transactional-update#92](https://github.com/openSUSE/transactional-update/issues/92)

Work as expected (there was just a misleading log with a wrong configuration)

java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
mcalmer: In Line 179 you set reboot to `Optional.of(false)`. But this has here now the same effect as `Optional.of(true)`.
I think this is a bug.

mcalmer: Ahh, sorry. Forget it. you set the value to the arg and not hard true.
This looks correct.

java/spacewalk-java.changes
```diff
@@ -1,3 +1,5 @@
+- Manage reboot in transactional update action chain (bsc#1201476)
+- Upgrade Bootstrap to 3.4.1
```
mcalmer, line 2: seems a wrong changelog entry sneaked in

java/code/src/com/suse/manager/webui/services/SaltActionChainGeneratorService.java
mbussolotto: I'm wondering if there's a better variable (and arg) name than `reboot_required`. I think it can be misleading in this context, since in other context it means that the user should run a reboot... but I didn't find anything better :)

mcalmer (changes requested):
This changelog thingy needs to be fixed. The rest looks good.

java/spacewalk-java.changes
mbussolotto: fixed

mcalmer (approved):

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -178,27 +208,46 @@ def resume():
...
     log.debug("Resuming execution of SUSE Manager Action Chain -> Target SLS: "
               "{0}".format(next_chunk))
-
-    inside_transaction = os.environ.get("TRANSACTIONAL_UPDATE")
-    if __grains__.get("transactional") and not inside_transaction:
-        return __salt__['transactional_update.sls'](next_chunk, queue=True)
+    
+    reboot_required = False
+    inside_transaction = False
+    if __grains__.get("transactional"):
```
vzhestkov, line 225: On my taste `__grains__.get("transactional")` is too repetative after, maybe just put it to the distinct variable and check for it?

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -128,8 +145,15 @@ def start(actionchain_id):
     except Exception as exc:
         log.error("There was an error while syncing custom states and execution modules")
 
-    inside_transaction = os.environ.get("TRANSACTIONAL_UPDATE")
-    if __grains__.get("transactional") and not inside_transaction:
+    reboot_required = False
+    inside_transaction = False
+    if __grains__.get("transactional"):
```
vzhestkov, line 150: `__grains__.get("transactional")` is repetative, I would prefer to put it to the local variable firtst and check for it after.

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -57,22 +57,39 @@ def _get_ac_storage_filenamepath():
...
+    if not os.path.isfile(filename):
+        log.error("File {0} not found".format(filename))
+        return None
     ret = None
     try:
-        with fopen(f_storage_filename, "r") as f_storage:
-            ret = yaml.load(f_storage.read())
-        if clear:
-            os.remove(f_storage_filename)
+        with fopen(filename, "r") as f:
```
vzhestkov, line 88: Try to use `salt.utils.yamlloader.load` instead. Not sure if it's really safe here.

vzhestkov (changes requested):
Put some suggestions in the comments.
But main question regarding `mgractionchains`: what are `sls` under `/actionchains/` are generated with? Is there always `schedule_next_chunk` inside?

mbussolotto:
> Put some suggestions in the comments. But main question regarding `mgractionchains`: what are `sls` under `/actionchains/` are generated with? Is there always `schedule_next_chunk` inside?

sls files are generated by `SaltActionChainGeneratorService` , and it generates a chunk until there's a reboot action. The end of a chunk is `schedule_next_chunk` and it contains information that will be stored in the minion (`_mgractionchains.conf` ). So no, there's not always a `schedule_next_chunk`, just when the action chains contains a reboot.
Then, after a reboot is performed, `mgractionchains.resume` is called by the Salt Reactor as a response to the 'minion/start/event' and reads information in `_mgractionchains.conf` to resume the action chain.

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
mbussolotto: Right, fixed :)

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -57,22 +57,39 @@ def _get_ac_storage_filenamepath():
...
     return os.path.join(minion_d_dir, '_mgractionchains.conf')
 
+def check_reboot_required(target_sls):
+    sls_file_on_minion = __salt__['cp.cache_file']('{0}{1}.sls'.format('salt://actionchains/', target_sls.replace('actionchains.','')))
+    yaml_file = _read_sls_file(sls_file_on_minion)
+    list_param = yaml_file['schedule_next_chunk']['mgrcompat.module_run']
+
+    for dic in list_param:
+        if 'reboot_required' in dic.keys() :
+            return dic.values()
```
agraul, line 67: I think this might not be exactly what we need here. `dict.values()` returns a memory-view of values, but the only caller just evaluates the truthiness of `check_reboot_required()`. This check will always be `True` as long as `reboot_required` is in the `dict`, no matter what the value is:

```python
>>> def check():
...     return {"reboot_required": False}.values()
... 
>>> required = check()
>>> bool(required)
True
```

So what this is doing is basically the same as 
```python
return 'reboot_required' in dic
```
There are two options now: change the calling or the called function:

```python
reboot_required = next(iter(check_reboot_required(target_sls)))
```
This approach has a big downside: `check_reboot_required` does not always return something that can be turned into an `iterator`. If `reboot_required` is not a parameter, the return will be `False` and
```python
>>> iter(False)
  File "<stdin>", line 1, in <module>
TypeError: 'bool' object is not iterable
```
There would be another problem, and that is that `next()` is shadowed in this module. 

That brings us to changing the return value of this function. Based on the function name, I'd only expect a boolean anyway, so it's a good idea to change this loop.

```python
    for dic in list_param:
        if 'reboot_required' in dic: # or dict.keys(), they are equivalent here
            return dic["reboot_required"]
```

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -57,22 +57,39 @@ def _get_ac_storage_filenamepath():
...
     if not os.path.isfile(f_storage_filename):
         return None
+    ret = _read_sls_file(f_storage_filename)
+    if clear:
+        os.remove(f_storage_filename)
+    return ret
+
+def _read_sls_file(filename):
+    if not os.path.isfile(filename):
+        log.error("File {0} not found".format(filename))
```
agraul, line 84: This code path won't be reached if `_read_next_ac_chunk` sees that the file does not exist, but in that case there is no error message. Wouldn't it make sense to also print an error message in that case, i.e. in  `_read_next_ac_chunk`?

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -128,8 +145,16 @@ def start(actionchain_id):
...
+    transactional_update = __grains__.get("transactional")
+    reboot_required = False
+    inside_transaction = False
+    if transactional_update:
+        reboot_required = check_reboot_required(target_sls)
+        inside_transaction = os.environ.get("TRANSACTIONAL_UPDATE")
+
+    if transactional_update and reboot_required:
+        ret = __salt__['transactional_update.sls'](target_sls, queue=True, activate_transaction=True)
+    elif transactional_update and not inside_transaction:
```
agraul, line 157: Why is one of them checking `inside_transaction` but not the other? If both branches would do `if transactional_update and not inside_transaction [and reboot_required]`, they could be collapsed to:

```python
if transactional_update and not inside_transaction:
    ret = __salt__['transactional_update.sls'](target_sls, queue=True, activate_transaction=reboot_required)

File: susemanager-utils/susemanager-sls/src/modules/mgractionchains.py [hunk 14]
```diff
@@ -178,27 +209,47 @@ def resume():
 
     This method is called by the Salt Reactor as a response to the 'minion/start/event'.
     '''
-    next_chunk = _read_next_ac_chunk()
-    if not next_chunk:
+    yaml_file = _read_next_ac_chunk()
+    if not yaml_file:
         return {}
-    if type(next_chunk) != dict:
+    if type(yaml_file) != dict:
```
agraul, line 215: I think `yaml_file` is a bit misleading as a name when it has to be a `dict`. I would expect a file-like object is bound to a variable that's name like that.

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -178,27 +209,47 @@ def resume():
...
+    reboot_required = False
+    inside_transaction = False
+    if transactional_update:
+        reboot_required = yaml_file.get('reboot_required')
+        inside_transaction = os.environ.get("TRANSACTIONAL_UPDATE")
+
+    if transactional_update and reboot_required:
+        ret =  __salt__['transactional_update.sls'](next_chunk, queue=True, activate_transaction=True)
+    elif transactional_update and not inside_transaction:
+        ret =  __salt__['transactional_update.sls'](next_chunk, queue=True)
```
agraul, line 234: Same as above, do they need to be seperate? Is it ok to be inside a transaction and call `__salt__['transactional_update.sls'](next_chunk, queue=True, activate_transaction=True)`?

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -178,27 +209,47 @@ def resume():
...
-    return {"success": True}
+    yaml_dict = {}
+    yaml_dict['success'] = True
+    if actionchain_id:
+        yaml_dict['actionchain_id'] = actionchain_id
+    if current_action_id:
+        yaml_dict['current_action_id'] = current_action_id
+    if reboot_required:
+        yaml_dict['reboot_required'] = reboot_required
+    return yaml_dict
```
agraul, line 255: It surprised me that a function called `clean` returns so much data, how is it expected to be used? It looks like it's only returning its input arguments.

agraul (changes requested):
I only looked at the Python code right now and I have a few questions. It would be great if we could clarify them.

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
mbussolotto: thanks for the explanation! I'll fix it

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
mbussolotto: right, actually my idea was that `_read_next_ac_chunk` should not check if file exist, it come back in the code somehow. I'll fix it.

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
mbussolotto: actually there's a reason for it. These parameters were introduced to fix
> although reboot is performed, WebUI events -> pending still show that the reboot action was not executed.

In minion without transactional update, the job result for the reboot action was collected by `SaltServerActionService::handleActionChainResult` : if reboot was scheduled successfully, `handleActionChainResult` put the action in `picked up` status. When the reboot completed, the `picked up` status become `successful`.

In minion with transactional update, there's no job result for the reboot action (since there's no action, the reboot is applied by `activate_transaction=true`), so I changed `mgractionchains.next` and  `mgractionchains.clean`  in order to provide as job result all the information required by `SaltServerActionService::handleActionChainResult` to set a "fake" reboot action as `picked up` .  You can find the code here: https://github.com/uyuni-project/uyuni/blob/c0588c3fe46f62b0a6af4bd7aad6538694119fb8/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java#L2805

(actually `actionchain_id` is not used, just `current_action_id` and `reboot_required`, but I prefer to add this one because I think it could help if error happens)

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -57,22 +58,39 @@ def _get_ac_storage_filenamepath():
...
+    return False
+
 def _read_next_ac_chunk(clear=True):
     '''
     Read and remove the content of '_mgractionchains.conf' file. Return the parsed YAML.
     '''
     f_storage_filename = _get_ac_storage_filenamepath()
-    if not os.path.isfile(f_storage_filename):
+    ret = _read_sls_file(f_storage_filename)
+    if ret == None:
```
agraul, line 77: Small suggestion, `None` is a singleton (like e.g. `True`/`False`) and singletons should be compared by identity (not by value) according to [PEP8](https://peps.python.org/pep-0008/#programming-recommendations)

```suggestion
    if ret is None:
```

susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
```diff
@@ -57,22 +58,39 @@ def _get_ac_storage_filenamepath():
...
+        log.error("File {0} not found".format(filename))
         return None
     ret = None
     try:
-        with fopen(f_storage_filename, "r") as f_storage:
-            ret = yaml.load(f_storage.read())
-        if clear:
-            os.remove(f_storage_filename)
+        with fopen(filename, "r") as f:
+            ret = yaml.load(f.read(), Loader=SaltYamlSafeLoader)
```
agraul, line 90: [nitpick] `f.read()` is not needed here, `PyYAML` can operate on a stream.

agraul (approved):
The Python code looks good now, I left two small comments that are nit-picks. Thanks for the work on getting AC working on SLE Micro!

mbussolotto:
@vzhestkov can you please review the changes? Merge is blocked cause you request some changes, I addressed them so now it should be fine :) Thanks!!!

mbussolotto:
@agraul I'm waiting the result of cucumber test (in particular the scope_action_chains). If everything goes fine, it can be merged. I'll ping you when tests will be completed

mbussolotto:
CI tests: https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/1906/

PR #5780, opened 2022-08-16
QE: Fix maintenance crawler script header

nodeg:
## What does this PR change?

This will fix the shebang of the maintenance crawler script and some
whitespace left inside.
Without the proper position of the shebang the script can only be
called with `ruby maintenance_crawler.rb` instead of just
`maintenance_crawler.rb`.

```bash
$ crawler.rb 25405                                                                                                                                                            
/home/dgedon/bin/crawler.rb: line 7: require: command not found                                                                                                               
/home/dgedon/bin/crawler.rb: line 8: require: command not found                                                                                                               
/home/dgedon/bin/crawler.rb: line 9: require: command not found                                                                                                               
/home/dgedon/bin/crawler.rb: line 10: require: command not found                                                                                                              
/home/dgedon/bin/crawler.rb: line 12: class: command not found                                                                                                                
/home/dgedon/bin/crawler.rb: line 14: syntax error near unexpected token `('                                                                                                  
/home/dgedon/bin/crawler.rb: line 14: `  def initialize(root_url, options = {})'

$ ruby crawler.rb 25405                                                                                                                                                       
http://download.suse.de/download/ibs/SUSE:/Maintenance:/25405/SUSE_Updates_SLE-Manager-Tools_15_x86_64/                                                                       
(...)
```

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Manager 4.3
Manager 4.2

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/ext-tools/maintenance_crawler.rb
+++ b/testsuite/ext-tools/maintenance_crawler.rb
@@ -1,8 +1,7 @@
+#!/usr/bin/env ruby
 # source 'https://rubygems.org'
 #
 # gem 'nokogiri'
-
-#!/usr/bin/env ruby
 
 require 'open-uri'
 require 'optparse'
@@ -135,7 +134,7 @@ class MaintenanceCrawler
         return URI.join(@root_url, url).to_s
     end
   end
-  
+
   # Output the current completions/total_queued to the console
   # Defaults to single-line-update but verbose (-v) mode triggers full output
   def print_status(url)


testsuite/ext-tools/maintenance_crawler.rb
```diff
@@ -1,9 +1,8 @@
+#!/usr/bin/env ruby
```
Bischoff, line 1: just change the shebang to `#! /usr/bin/ruby` and it will be possible to call it with simply `./crawler.rb`

Bischoff (changes requested):

testsuite/ext-tools/maintenance_crawler.rb
nodeg: According to [Wikipedia](https://en.wikipedia.org/wiki/Shebang_(Unix)) the way Oscar did it is correct no need to introduce a white space there. It further states
> a shebang is the character sequence [...] at the beginning of a script.

FMPOV putting this at the beginning of the script (where it belong) is enough :wink:

Bischoff: ok for the space

but i suggest to change `#!/usr/bin/env ruby` to `#! /usr/bin/ruby`

Bischoff (changes requested):

testsuite/ext-tools/maintenance_crawler.rb
nodeg: With this we would loose portability. According to my research that's what the `env` command is somehow for in this context.

nodeg: I for example use different Ruby vesions:

```bash
❯ /usr/bin/ruby --version
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]

❯ which ruby
/home/dgedon/.rvm/rubies/ruby-2.5.9/bin/ruby

❯ ruby --version
ruby 2.5.9p229 (2021-04-05 revision 67939) [x86_64-linux]
```

Bischoff: you basically have no portability - inside a given version of ruby all gems tend to be incompatible... :trollface: 

also normally `usr/bin/ruby` should be a symlink to the default installed version.

but i checked, it works to run `./crawler.rb` with the `env` construct, so approving (although i don't see why it should be better)

Bischoff (approved):
approving, even though i'm not sure of what using `env` brings

nodeg:
We can also let Oscar decide if he wants to change that since he is the author. I just wanted to fix the fact that the script did not work on my machine which was due to the fact that the shebang was not at the very beginning of the script.

Issue #4769, opened 2022-02-04
Uyuni Bugreport - Uyuni 2022.01 - Content Lifecycle AppStream Filter not working anymore


lumarel:
# Additional Information

@cbbayburt As we were talking in Gitter, I noticed that since January (most likely after the upgrade to 2022.01) my Content Lifecycles of Rocky Linux suddenly have all packages in it without the AppStream data (comps files), which basically means, the set filters didn't filter out the disabled packages from the AppStream filters.
Mostly the filter builds got initiated via the api, but some few were initiated via the UI.
The filters themselves were created before the template was in place, but they were working until now, so they should be correct.

I will add all the information about the channel, repos and the CL in the details below!

Here is the real result shown with the `perl` packages in the `exp` environment, if you look at the filters you will notice, that it should only be the packages of the `5.26` Stream should be in the repo:
![image](https://user-images.githubusercontent.com/42647570/152550035-3678c7b5-b720-40cd-b9ce-45954878b767.png)

Thank you in regards for looking at it!

## Version of Uyuni Server and Proxy (if used)
```
# zypper info Uyuni-Server-release
Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : Uyuni Server Stable
Name           : Uyuni-Server-release
Version        : 2022.01-165.2.uyuni1
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes (automatically)
Status         : up-to-date
Source package : Uyuni-Server-release-2022.01-165.2.uyuni1.src
Summary        : Uyuni Server
Description    :
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.
```

## Details about issue

Here comes the wall of text:

- The originating channels:

```
spacecmd {SSM:0}> softwarechannel_listchildchannels rockylinux8-x86_64
epel8-everything-rockylinux8-x86_64
epel8-modular-rockylinux8-x86_64
rockylinux8-docker-ce-x86_64
rockylinux8-elasticsearch-7x-x86_64
rockylinux8-grafana-x86_64
rockylinux8-influxdb-x86_64
rockylinux8-microsoft-prod-x86_64
rockylinux8-prometheus-rpm-x86_64
rockylinux8-uyuni-client-devel-x86_64
[...]

PR #5786, opened 2022-08-16
Fix filtering of modular packages on CentOS flavors (bsc#1201893)

cbbayburt:
Most CentOS flavors serve only the latest packages in their repositories. When a new release comes out, old packages are wiped out, new packages are served, and the modules.yaml file is replaced with a new one.

SUSE Manager on the other hand never removes synced packages. When a new release is synced, both the old and new packages are still available in channels. We read modules.yaml to look up the AppStream packages to filter out. Since the modules.yaml file is replaced with a new one and it does not contain any information about the older packages, those older packages leak into target repositories.

In SUSE Manager, we already keep the older modules.yaml files in the file system. This PR changes the behavior to also keep them in the database, and use them when retrieving the list of packages to filter during a CLM build.

Followup: migration script to fix the existing channels without the need for a re-sync: https://gist.github.com/cbbayburt/0f83039a21d021e8e1415618cff414ec

See also:
 - https://bugzilla.suse.com/show_bug.cgi?id=1201893
 - https://github.com/uyuni-project/uyuni/issues/4769

## Documentation

## Test coverage
- Unit tests were added

## Links
Fixes https://github.com/SUSE/spacewalk/issues/18483
Fixes https://github.com/SUSE/spacewalk/issues/17961
Fixes https://github.com/uyuni-project/uyuni/issues/4769

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/channel/Channel.java
+++ b/java/code/src/com/redhat/rhn/domain/channel/Channel.java
@@ -26,6 +26,7 @@ import com.redhat.rhn.manager.channel.ChannelManager;
 import com.redhat.rhn.manager.system.IncompatibleArchException;
 import com.redhat.rhn.manager.system.SystemManager;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
@@ -34,6 +35,7 @@ import org.apache.logging.log4j.Logger;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
@@ -78,7 +80,7 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
     private ChannelProduct product;
     private ProductName productName;
     private Comps comps;
-    private Modules modules;
+    private Set<Modules> modules;
     private MediaProducts mediaProducts;
     private String summary;
     private Set<Errata> erratas = new HashSet<>();
@@ -221,8 +223,29 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
     /**
      * @param modulesIn The Modules to set.
      */
-    public void setModules(Modules modulesIn) {
+    public void setModules(Set<Modules> modulesIn) {
         this.modules = modulesIn;
+    }
+
+    /**
+     * Add a module metadata file to the channel
+     * @param modulesIn the module metadata entity to add
+     */
+    public void addModules(Modules modulesIn) {
+        if (this.modules == null) {
+            this.modules = new HashSet<>();
+        }
+        modulesIn.setChannel(this);
+        this.modules.add(modulesIn);
+    }
+
+    /**
+     * Remove an existing module metadata file from the channel
+     * @param modulesIn the module metadata entity to remove
+     */
+    public void removeModules(Modules modulesIn) {
+        this.modules.remove(modulesIn);
+        modulesIn.setChannel(null);
     }
 
     /**
@@ -235,14 +258,27 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
     }
 
     /**
+     * Gets the synced module metadata files belonging to the channel
+     * <p>
+     * See {@link Channel#getLatestModules()} to get the latest metadata file currently in use.
+     *
      * @return Returns the Modules.
      */
-    public Modules getModules() {
+    public Set<Modules> getModules() {
         return modules;
     }
 
+    /**
+     * Gets the latest module metadata file in use
+     *
+     * @return the module metadata (modules.yaml) file
+     */
+    public Modules getLatestModules() {
+        return modules.stream().max(Comparator.comparing(Modules::getLastModified)).orElse(null);
+    }
+
     public boolean isModular() {
-        return modules != null;
+        return CollectionUtils.isNotEmpty(modules);
     }
 
     /**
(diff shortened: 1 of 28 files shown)


schema/spacewalk/common/tables/rhnChannelComps.sql
```diff
@@ -31,12 +31,14 @@ CREATE TABLE rhnChannelComps
...
-                               REFERENCES rhnCompsType(id)
+                               REFERENCES rhnCompsType(id),
+    CONSTRAINT rhn_channelcomps_cid_ctype_filename_uq
+        UNIQUE(channel_id, comps_type_id, relative_filename)
 )
 
 ;
 
-CREATE UNIQUE INDEX rhn_channelcomps_cid_ctype_uq
+CREATE INDEX rhn_channelcomps_cid_ctype
```
mcalmer, line 41: normal index names end with `_idx`. This makes looking for indexes easier.

schema/spacewalk/postgres/triggers/rhnChannelCloned.sql
```diff
@@ -26,7 +26,7 @@ begin
 			( id, channel_id, comps_type_id, relative_filename,
 				last_modified, created, modified )
 		select nextval('rhn_channelcomps_id_seq'), new.id, comps_type_id, relative_filename,
-				current_timestamp, current_timestamp, current_timestamp
+				last_modified, current_timestamp, current_timestamp
```
rjmateus, line 29: I get the goal, but It will be strange that last_modified can be older than the current_timestamp. For that reason I'm not sure if you should copy the last_modify from the original channel comps.

java/code/src/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -234,14 +257,27 @@ public void cloneModulesFrom(Channel from) {
...
         return modules;
     }
 
+    /**
+     * Gets the latest module metadata file in use
+     *
+     * @return the module metadata (modules.yaml) file
+     */
+    public Modules getLatestModules() {
+        return modules.stream().max(Comparator.comparing(Modules::getLastModified)).orElse(null);
```
rjmateus, line 276: I let a comment in the database transformation. But now I understand why you need to keep the old last modified date when cloning channels. You need it here. (I left the other comment anyway, so you can check it).

java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java
```diff
@@ -1450,19 +1450,8 @@ public static void analyzeServerNeededCache() {
...
-        else {
-            if (!to.isModular()) {
-                Modules modules = new Modules();
-                modules.setChannel(to);
-                to.setModules(modules);
-            }
-            to.getModules().setRelativeFilename(from.getModules().getRelativeFilename());
-        }
+        to.getModules().clear();
+        from.getModules().forEach(fromModule ->
```
rjmateus, line 1454: I think the from.getModules() can return null. Should it be protected with a isModular?
```suggestion
        if (fromisModular()) {
            from.getModules().forEach(fromModule ->
            ....
        }
```

java/code/src/com/redhat/rhn/domain/channel/test/ChannelFactoryTest.java
```diff
@@ -556,4 +557,27 @@ public void testFindAllByUserOrderByChild() throws Exception {
         assertEquals("a_child1", channels.get(2).getLabel());
         assertEquals("b_parent3", channels.get(3).getLabel());
     }
+
+    @Test
+    public void testCloneModulesMetadata() throws Exception {
+        User user = UserTestUtils.findNewUser("testUser", "testOrg" + this.getClass().getSimpleName());
+        Channel orig = ChannelTestUtils.createTestChannel(user);
+        Modules modules = new Modules("modules1.yaml", new Date());
```
rjmateus, line 565: I think you can create with a date in the past, to make it more explicit the date should be kept.
I'm also not sure about default precision in date, or we can beeing use the same date in both create modules and hiding a possible error in case someone changes the logic.

java/code/src/com/redhat/rhn/manager/channel/CloneChannelCommand.java
```diff
@@ -117,14 +116,9 @@ public Channel create()
...
+        // Comps files are cloned by the DB trigger 'rhn_channel_cloned_comps_trig'
         if (stripModularMetadata) {
-            if (c.getModules() != null) {
-                HibernateFactory.getSession().delete(c.getModules());
-            }
-            c.setModules(null);
-        }
-        else {
-            c.cloneModulesFrom(original);
+            c.getModules().clear();
```
rjmateus, line 121: I think getModules can return null. It can be tested with:
```suggestion
            if (c..isModular()) {
                c.getModules().clear();
             }
```

rjmateus (commented):
overall looks good to me, just a few comments on it.

schema/spacewalk/postgres/triggers/rhnChannelCloned.sql
cbbayburt: In this table, `last_modified` refers to the last modification date of a comps/modules file in the filesystem. When cloning the row in the DB, the referenced file is not actually touched, so the `last_modified` should stay the same as the original.
Note that there are also generic `created` and `modified` columns that store timestamp information about the DB entities themselves.

java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java
cbbayburt: This set is managed by Hibernate and there's no path that sets these null currently, but you're right, we can foolproof it with a check :+1:

java/code/src/com/redhat/rhn/domain/channel/test/ChannelFactoryTest.java
cbbayburt: Agreed :+1:

mcalmer (approved):

vzhestkov (approved):
Looks good :+1:

PR #5826, opened 2022-08-22
Add copy grains to mgr_switch_to_venv_minion

vzhestkov:
## What does this PR change?

The file `grains` was missing on switching salt-minion to the Salt Bundle.
This PR is implementing the copying of grains file if it exists.

## GUI diff

No difference.

## Documentation

## Test coverage

No tests: Switching to the Salt Bundle is not covered with the testsuite

## Links

Tracks https://github.com/SUSE/spacewalk/issues/18694

## Changelogs

- [ ] No changelog needed

## Re-run a test


agraul (approved):
LGTM!

susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
```diff
@@ -34,6 +34,15 @@ mgr_copy_salt_minion_configs:
...
       - ([ {{ venv_managed_minion }} = "False" ])
 
+mgr_copy_salt_minion_grains:
+  file.copy:
+    - name: /etc/venv-salt-minion/grains
+    - source: /etc/salt/grains
+    - require:
+      - pkg: mgr_venv_salt_minion_pkg
+    - onlyif:
+      - test -f /etc/salt/grains
```
SchoolGuy, line 44: Is there maybe a Python native mechanism that we can use instead of this shell mechanism? With a quick google I found that the second option in [this](https://serverfault.com/a/884194/408066) Stackoverflow suggest that it is possible but I am unsure if it is working in our case.

vzhestkov: With the link there last message points to test also, which is more relevant here. There is a check with calling the salt module function, but this feature was implemented with salt `3000` or maybe higher, don't remember for sure, but not a huge difference here, the main idea than the check for something other than just to call external command could be inconsistent between different salt versions with chan be used on outdated minions. So I would prefer to keep it as it is now.

SchoolGuy: Okay. I accept your rationale here. Thanks for the clarification.

PR #5827, opened 2022-08-22
Check reformatting in Cobbler

Bischoff:
## What does this PR change?

Like in Retail, we would like to know in test suite's Cobbler feature tests if reformatting really happened. For that we drop an `/intact` file in the PXE boot minion before trying to PXE boot it. After the PXE boot, the disk should have been reformatted and that file should be gone.

Also, when debugging a failed PXE boot, we would like to know which one of Cobbler or Retail was last to happen, by inspecting whether `/intact` contains `Cobbler` or `Retail`.

## Links

Ports:
* 4.2: SUSE/spacewalk#18713
* 4.3: SUSE/spacewalk#18712

## Changelogs

- [x] No changelog needed


nodeg (approved):
Nice approach and good idea!

Bischoff:
> Don't forget the ports :p

Done. I stumbled on a missing backport on 4.2.

Issue #5830, opened 2022-08-23
Replace ssl certificate at proxy -> ERROR: 'authorityKeyIdentifier'


mrstigmata:
# Problem description

_The following information is very important in order to help us to understand your problem. Omission of the following details can cause delays or can receive no attention at all._

## Version of Uyuni Server and Proxy (if used)

`zypper info Uyuni-Server-release` _and_ `zypper info Uyuni-Proxy-release` _(if used)_

`Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : Uyuni Server Stable
Name           : Uyuni-Server-release
Version        : 2022.08-185.2.uyuni2
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes (automatically)
Status         : up-to-date
Source package : Uyuni-Server-release-2022.08-185.2.uyuni2.src
Summary        : Uyuni Server`

`Information for package Uyuni-Proxy-release:
--------------------------------------------
Repository     : Uyuni Proxy Stable for openSUSE Leap 15.4 (x86_64)
Name           : Uyuni-Proxy-release
Version        : 2022.08-185.2.uyuni2
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes
Status         : up-to-date
Source package : Uyuni-Proxy-release-2022.08-185.2.uyuni2.src
Summary        : Uyuni Proxy`

## Details about the issue

we want to import a new proxy and uyuni server ssl certificate. the ssl certificate is created by our company ca.
we run that command:

` mgr-ssl-cert-setup -c -v -r root_ca.crt -i subca01.crt -s proxy01.pem -k proxy01-serverkey.pem`

the Output is the same at the Uyuni Proxy and the Uyuni Server:

`ERROR: 'authorityKeyIdentifier'`

Log file: /var/log/rhn/mgr-ssl-cert-setup.log

`2022/08/23 10:43:12 +02:00: /usr/lib/python3.6/site-packages/certs/mgr_ssl_cert_setup.py.processCommandline(105) - ['/usr/bin/mgr-ssl-cert-setup', '-v', '-c', '-r', 'root_ca.crt', '-i', 'subca01.crt', '-s', 'proxy01.pem', '-k', 'proxy01-serverkey.pem']
[...]

PR #5834, opened 2022-08-23
fix root CAs which do not have authorityKeyIdentifier set to themself

mcalmer:
## What does this PR change?

Letsencrypt Root CA does not have an authorityKeyIdentifier set. This might be correct as it would point
anyway only to itself. But the script is using it to identify the self signed root ca.
This change detect "subject" == "issuer" and "isCA" and set the value for the internal evaluation.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/5830
Tracks https://github.com/SUSE/spacewalk/pull/18734

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacewalk/certs-tools/mgr_ssl_cert_setup.py
+++ b/spacewalk/certs-tools/mgr_ssl_cert_setup.py
@@ -278,6 +278,9 @@ def getCertData(cert):
     data["content"] = cert
     if data["subject"] == data["issuer"]:
         data["root"] = True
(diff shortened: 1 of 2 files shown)


vzhestkov (approved):
Looks good :+1:

PR #1388, opened 2019-09-18
Enable "Azure", "Amazon EC2" and "Google Compute Engine" as Virtual Host Managers

meaksh:
## What does this PR change?

This PR enables "Azure", "Amazon EC2" and "Google Compute Engine" as available "Virtual Host Managers".

- Allows systems that reports "uuid" based on virtual instance id instead of SMBIOS "uuid".
- Use "instance_id" grain as the "uuid" for minions that contains this grain.
- Enable UI to select the new types of "Virtual Host Managers"
- Add new types of "CPUArch", "ServerArch" and "VirtualInstanceTypes" on the DB.
- Add schema migration files
- Provide custom `public_cloud` grain to report `instance_id` when running on a Public Cloud instance

The support for the new Public Cloud providers on `virtual-host-gatherer` is coming from https://github.com/openSUSE/virtual-host-gatherer/pull/3

## Documentation
- Issue created: https://github.com/SUSE/spacewalk/issues/9501

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

https://github.com/SUSE/spacewalk/issues/6576
https://github.com/SUSE/spacewalk/issues/9467
https://github.com/SUSE/spacewalk/issues/4574
https://github.com/SUSE/spacewalk/issues/9468
https://github.com/SUSE/spacewalk/issues/9488

- [x] **DONE**

## Changelogs

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_lint_checkstyle"		 
- [ ] Re-run test "java_pgsql_tests"		 
- [ ] Re-run test "ruby_rubocop"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"		 
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"		 
- [ ] Re-run test "spacecmd_unittests"

PR #5866, opened 2022-08-30
Detect Amazon EC2 instances

vzhestkov:
## What does this PR change?

Show `Amazon EC2` in `Virtualization` field in the system profile for Amazon EC2 instances.

Requires https://github.com/openSUSE/salt/pull/555 to be present on the salt minion side.

## GUI diff

Before:
Virtualization field in system profile was shown as `KVM/QEMU` or `Para-Virtualized` if running as `xen` instance.

After:
Virtualization field in system profile is labeled as `Amazon EC2`

## Documentation

## Test coverage
- No tests: there is no tests for this part of functionality

## Links

Fixes https://github.com/SUSE/spacewalk/issues/16940

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/reactor/hardware/HardwareMapper.java
+++ b/java/code/src/com/suse/manager/reactor/hardware/HardwareMapper.java
@@ -608,6 +608,19 @@ public class HardwareMapper {
                         LOG.info(String.format("Detected virtual instance type '%s' for minion '%s'",
                                 virtTypeLabel, server.getMinionId()));
                 }
+                if (virtSubtype.startsWith("Amazon EC2")) {
+                    switch (virtTypeLowerCase) {
+                        case "xen":
+                            virtTypeLabel = "aws_xen";
+                            break;
+                        case "qemu":
+                        case "kvm":
+                            virtTypeLabel = "aws_kvm";
+                            break;
+                        default:
+                            virtTypeLabel = "aws";
+                    }
+                }
                 type = VirtualInstanceFactory.getInstance()
                         .getVirtualInstanceType(virtTypeLabel);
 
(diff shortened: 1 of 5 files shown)


vzhestkov:
@meaksh I found that initially `Amazon EC2` was added with https://github.com/uyuni-project/uyuni/pull/1388, but I didn't find the way of detecting it from salt side. The other point here that `Amazon EC2` instance can be `kvm` or `xen` based. Do we need to distinguish them somehow? There is a single entry in `rhnVirtualInstanceType` with `aws` label. In the current state of this PR, both of these instance types will look the same.

meaksh:
@vzhestkov let me try to recall here...

When we implemented support for Amazon EC2 instances, we did it also for the context of "Virtual Host Managers", so we use `virtual-host-gatherer` to collect the instances on your AWS region, and you get a list of virtual instances (uuids) running there. 

Those virtual instances, if not existing in the DB as registered systems, are created as "aws" type, as we don't really have access to those virtual instances to perform any operation and gather more information at this point to those instances.

NOTE: Maybe `virtual-host-gatherer` can collect extra info for the virtual instance (about "xen" or "kvm") so we can have more accurate type from the begining.

In any case, once we do actually register that virtual instance in SUMA (via bootstrapping), then SUMA should notice the "uuid" is the same of that virtual instance and should match them. Maybe then is when we see this wrong "Para-Virtualized" or "QEMU/KVM" in the virtualization field.

AFAICS this is a valid bug and the changes you are doing look correct.

meaksh (approved):

meaksh:
@vzhestkov hmmm, yeah, maybe we can create those new types and then during HW refresh we assign the right type now that we have more information about the instance. Small new feature to be definetely more accurate :smile: 

IIUC this shouldn't harm, we simply create a new type in the DB (+ schema migration file) and assign proper type during HW refresh action.

rjmateus (approved):
LGTM

PR #5867, opened 2022-08-30
Synchronize code between branches (4)

Bischoff:
## What does this PR change?

Synchronize branches: traditional clients.

Bug caught: `@susemanager` tag is not the contrary of `@salt_bundle`. `@skip_if_salt_bundle` is.

## Links

Sibling PRs:
* 4.2:
* 4.3:

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/secondary/trad_migrate_to_minion.feature
+++ b/testsuite/features/secondary/trad_migrate_to_minion.feature
@@ -92,7 +92,7 @@ Feature: Migrate a traditional client into a Salt minion
     And I wait until I see "has been deleted" text
     Then "sle_client" should not be registered
 
-  @susemanager
+@skip_if_salt_bundle
   Scenario: Cleanup: register minion again as traditional client
     When I enable client tools repositories on "sle_client"
     And I install the traditional stack utils on "sle_client"
@@ -100,7 +100,7 @@ Feature: Migrate a traditional client into a Salt minion
     And I bootstrap traditional client "sle_client" using bootstrap script with activation key "1-SUSE-KEY-x86_64" from the proxy
     Then I should see "sle_client" via spacecmd
 
-  @salt_bundle
+@salt_bundle
   Scenario: Cleanup: register minion again as traditional client
     When I enable client tools repositories on "sle_client"
     And I install the traditional stack utils on "sle_client"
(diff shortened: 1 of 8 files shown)


testsuite/features/secondary/trad_migrate_to_minion.feature
```diff
@@ -92,15 +92,15 @@ Feature: Migrate a traditional client into a Salt minion
     And I wait until I see "has been deleted" text
     Then "sle_client" should not be registered
 
-  @susemanager
+@skip_if_salt_bundle
```
vandabarata, line 95: is this the right change here? just making sure

Bischoff: Yes, we have @ on left margin on all our files, excepted a few exceptions like this one.

After this PR, there will be 4 "exceptions" left.

nodeg:
```
@susemanager
```
are executed only if the server has SUSE Manager installed and will
not run if Uyuni is detected.

Inside the testsuite, the scenarios that are tagged with

```
@uyuni
```
are executed only if the server has Uyuni installed and will
not run if SUSE Manager is detected.

So would @uyuni be not the right tag here?

vandabarata: the issue is that this and the scenario below are the same but with different filters. but instead of the filters being susemanager and uyuni, they're salt bundle or not salt bundle
but this is different in 4.2

Bischoff: for completeness, here is the definition used by the bundle tag:
```
def use_salt_bundle
# Use venv-salt-minion in Uyuni, or SUMA Head and 4.3
  $product == 'Uyuni' || ['head', '4.3'].include?($product_version)
end
```

Bischoff: And a screenshot of the bug in action:

image

vandabarata (approved):

testsuite/features/secondary/trad_migrate_to_minion.feature
nodeg: Interesting. Then I guess the changes make sense.

nodeg (approved):

PR #5895, opened 2022-09-05
Do not use same SP for SLES and SLED

Bischoff:
## What does this PR change?

When we test the product wizard, we enable SLED 15 SP4, while we intend to bootstrap SLES 15 SP4.

The problem is that both products share same packages. This creates a problem when we kill the reposyncs, the bootstrap repo for SLES 15 SP4 does not get created because SUSE Manager waits for the killed repos to be fully synced.

Strictly speaking, this is a product issue, but it needs a serious refactoring, for a use case that not many customers will meet. So let's work around it by reposyncing different SPs.

## Links

Ports:
* 4.2: https://github.com/SUSE/spacewalk/pull/18893
* 4.3: https://github.com/SUSE/spacewalk/pull/18892

Fixes #18707

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/reposync/srv_enable_sync_products.feature
+++ b/testsuite/features/reposync/srv_enable_sync_products.feature
@@ -17,17 +17,19 @@ Feature: Be able to list available products and enable them
     And I should get "  [ ] (R) Basesystem Module 15 SP3 x86_64"
     And I should get "  [ ] Desktop Applications Module 15 SP3 x86_64"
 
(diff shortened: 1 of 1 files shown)


ktsamis (approved):
Since it's a problem to kill the reposync and then expect the bootstrap repo, yes let's do this. But that means more space and time, so we need to be aware of that.

Bischoff:
> Since it's a problem to kill the reposync and then expect the bootstrap repo, yes let's do this. But that means more space and time, so we need to be aware of that.

Hopefully, killing the reposync also spares space and time (that's the whole point of doing it).

PR #5899, opened 2022-09-07
Tools for SLED are shared with SLES, we must not kill their reposync

Bischoff:
## What does this PR change?

Tools for SLED are shared with SLES, we must not kill their reposync.

Note: version for 4.2 has a piggyback commit, meant to use sle 15 sp3 as the default for auto-installation, instead of sp2.

## Links

Ports:
* 4.2: https://github.com/SUSE/spacewalk/pull/18915
* 4.3: https://github.com/SUSE/spacewalk/pull/18914

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/constants.rb
+++ b/testsuite/features/support/constants.rb
@@ -275,12 +275,14 @@ CHANNEL_TO_SYNCH_BY_OS_VERSION = {
   %w[
     sle-product-sles15-sp4-pool-x86_64
     sle-manager-tools15-pool-x86_64-sp4
+    sle-manager-tools15-pool-x86_64-sled-sp4
     sle-manager-tools15-beta-pool-x86_64-sp4
     sle-module-containers15-sp4-pool-x86_64
     sle-module-basesystem15-sp4-pool-x86_64
     sle-module-server-applications15-sp4-pool-x86_64
     sle-product-sles15-sp4-updates-x86_64
     sle-manager-tools15-updates-x86_64-sp4
+    sle-manager-tools15-updates-x86_64-sled-sp4
     sle-manager-tools15-beta-updates-x86_64-sp4
     sle-module-containers15-sp4-updates-x86_64
     sle-module-basesystem15-sp4-updates-x86_64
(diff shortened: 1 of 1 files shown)


testsuite/features/support/constants.rb
```diff
@@ -275,12 +275,14 @@
   %w[
     sle-product-sles15-sp4-pool-x86_64
     sle-manager-tools15-pool-x86_64-sp4
+    sle-manager-tools15-pool-x86_64-sled-sp4
```
elariekerboull, line 278: We currently sync sle-manager-tools15-pool-x86_64-sled-sp3 for some reason in the CI, maybe we should update to  sled-sp4 as well?

elariekerboull, line 278: Currently for Uyuni at least we partially sync SLED 15 SP3, if I understand properly the reposync kill condition, it will still be killed since we won't have any SLE15-SP3 product
We could either change the default dependency to SLED15-SP3 or update the synced channel to be SLED15-SP4

ktsamis:
Since they have different labels are we sure they are shared? It seems strange to me to have 2 labels for the same product

mcalmer (approved):

Bischoff:
> Since they have different labels are we sure they are shared? It seems strange to me to have 2 labels for the same product

it's same product ID, but different channel labels, if I understood well.

testsuite/features/support/constants.rb
Bischoff: > Currently for Uyuni at least we partially sync SLED 15 SP3, if I understand properly the reposync kill condition, it will still be killed since we won't have any SLE15-SP3 product We could either change the default dependency to SLED15-SP3 or update the synced channel to be SLED15-SP4

We do start a reposync of SLED 15 SP3 product in order to test the products wizard. As we don't really use this product, we can kill its reposyncs. However, two of them, the ones I add here, are shared with SLES 15 SP4 and should not be killed. Reason is that the tools are shared between all service packs.

I know, it's very convoluted.

PR #5900, opened 2022-09-07
QE: Use SLE base product

elariekerboull:
## What does this PR change?

This PR introduces the use of synchronized base-products to use in the different SLES minions.
An activation key using these synchronized base-product and having the test channel that we previously used as child channel is used to bootstrap the minions.
This allows us to replace the use of some test repositories enabled and disabled in each minion.

Some additional activation keys had to be added since the original one does not fit to all minions such as the Redhat-like minion.

- [ ] Create a test-channel as a child of each base product - _only SLE for now_
- [x] Create an activation key with the base channel and the test-channel
- [x] Rename repositories and test channels
- [ ] Remove completely the use of `Test-Channel x86_64` - _KVM feature still needs it_
- [x] Replace SLE basic key
- [x] Replace Kiwi Key
- [ ] Replace KVM Key - _is still under leap 15.3 for now_
- [x] Replace Docker Key
- [x] **Remove extra waiting steps before merging**

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Links

Fixes # https://github.com/SUSE/spacewalk/issues/17645
Tracks # 
4.2 https://github.com/SUSE/spacewalk/pull/19581
4.3 https://github.com/SUSE/spacewalk/pull/19209

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


testsuite/features/init_clients/srv_create_activationkey.feature
```diff
@@ -9,13 +9,28 @@ Feature: Create activation keys
...
     When I follow the left menu "Systems > Activation Keys"
     And I follow "Create Key"
     And I enter "SUSE Test Key x86_64" as "description"
     And I enter "SUSE-KEY-x86_64" as "key"
     And I enter "20" as "usageLimit"
-    And I select "Test-Channel-x86_64" from "selectedBaseChannel"
+    And I select "SLE-Product-SLES15-SP4-Pool for x86_64" from "selectedBaseChannel"
+    And I include the recommended child channels
+    And I check "SLE-Module-DevTools15-SP4-Pool for x86_64"
+    And I check "SLE-Custom-Channel-x86_64"
```
srbarrios, line 33: Ok, but we might also consider using constants.rb to collect and provide this information, keeping a unique source of information.

testsuite/features/secondary/allcli_action_chain.feature
```diff
@@ -9,38 +9,23 @@ Feature: Action chains on several systems at once
...
-    And I remove package "andromeda-dummy" from this "rhlike_minion" without error control
+    And I enable repository "test_repo_rpm_pool" on this "ssh_minion"
+    And I remove package "andromeda-dummy" from this "ssh_minion" without error control
     And I remove package "andromeda-dummy" from this "sle_minion" without error control
     And I install package "andromeda-dummy-1.0" on this "sle_minion"
-    And I install package "andromeda-dummy-1.0" on this "rhlike_minion"
+    And I install package "andromeda-dummy-1.0" on this "ssh_minion"
     And I refresh the metadata for "sle_minion"
-    And I refresh the metadata for "rhlike_minion"
+    And I refresh the metadata for "ssh_minion"
```
srbarrios, line 20: Why are you changing here a RH-like client by a SLES SSH client?
This seems out of the scope of this PR.

If you want this change, let's do it in a different PR and we can discuss it there.

elariekerboull: That is because they need to have the same base-channel in this feature
rhlike_minion does not have the SLE base channel so this test cannot pass with it.

testsuite/features/init_clients/srv_create_activationkey.feature
```diff
@@ -9,13 +9,28 @@ Feature: Create activation keys
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
 
+  Scenario: Clone the child custom channel including test repositories
+    When I follow the left menu "Software > Manage > Channels"
+    And I follow "Clone Channel"
+    And I select the custom architecture channel for "sle_minion" as the origin channel
+    And I choose "current"
+    And I click on "Clone Channel"
+    And I enter "SLE-Custom-Channel-x86_64" as "Channel Name"
```
srbarrios, line 18: I think we still should include the word "test" in this custom channel.
Custom channel is the kind of channel, test channel describes the purpose of this channel. Will "SLE-Test-Custom-Channel-x86_64" work for you?

testsuite/features/init_clients/srv_create_activationkey.feature
```diff
@@ -9,13 +9,28 @@ Feature: Create activation keys
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
 
+  Scenario: Clone the child custom channel including test repositories
```
srbarrios, line 12: Could you explain why we need to clone this channel?

srbarrios, line 12: We need Cucumber tags in SLE scenarios, as we do for rhlike_minion scenarios.

testsuite/features/secondary/min_check_patches_install.feature
```diff
@@ -26,19 +26,17 @@ Feature: Display patches
 
   Scenario: Check SLES release 6789 patches
     When I follow the left menu "Patches > Patch List > Relevant"
-    And I follow "andromeda-dummy-6789"
+    And I follow "CL-andromeda-dummy-6789"
```
srbarrios, line 29: uhmm the clone it might be complicating this stuff. I would consider avoid the clone and create that child channel directly as a primary channel.

testsuite/features/init_clients/srv_create_activationkey.feature
elariekerboull: The initial idea was to avoid creating a new channel and syncing it again, and clone directly the initial test channel with everything that we need in it, it saves some steps and some time.
However the result would be the about the same by not cloning (apart the names of patches) so we can do that instead.

testsuite/features/secondary/min_salt_software_states.feature
```diff
@@ -40,13 +40,14 @@ Feature: Salt package states
...
   Scenario: Remove a package through the UI
-    Then I follow "States" in the content area
+    When I follow "States" in the content area
     And I follow "Packages"
     And I follow "Search"
     And I should see a "Package States" text
     And I list packages with "dummy"
-    Then I should see a "milkyway-dummy" text
-    And "milkyway-dummy" should be installed on "sle_minion"
+    And I wait for "60" seconds
```
srbarrios, line 48: Uf, avoid explicit waits as much as possible.
Use a "wait until I see" and play with the timeout. But don't strictly wait always 60 seconds, because that depends very mucho on the environment we run it.

testsuite/features/init_clients/srv_create_activationkey.feature
elariekerboull: I'll have to check if we have corresponding tags everywhere it is referenced
It is also used for ssh_minion

testsuite/features/secondary/min_check_patches_install.feature
elariekerboull: True, it did give me a bit of a headache with these prefixes

testsuite/features/secondary/min_salt_software_states.feature
elariekerboull: This was the Net timeout issue, there was nothing I could do outside of that since waiting for any element resulted in the same failure
However it might be only a PR pipeline issue and it could not be necessary in other CIs

testsuite/features/init_clients/srv_create_activationkey.feature
elariekerboull: Didn't see the second part, sorry
It seems synchronized base products cannot be assigned as child-channels so the test channel needs to be a child

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -199,8 +199,10 @@
 
 When(/^I include the recommended child channels$/) do
   toggle = "//span[@class='pointer']"
+  xpath = "//i[contains(@class, 'fa-toggle-off')]"
+
   if page.has_xpath?(toggle, wait: 5)
-    find(:xpath, toggle).click
+    find(:xpath, toggle).click if page.has_xpath?(xpath, wait: 5)
```
srbarrios, line 205:
```suggestion
    find(:xpath, wait: 5, toggle).click
```

testsuite/run_sets/init_clients.yml
```diff
@@ -6,6 +6,7 @@
 
 ## Core - Initialize clients BEGIN ##
 
+- features/init_clients/srv_create_activationkey.feature
```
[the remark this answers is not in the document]
elariekerboull: The activation key needs to have the reposync completed, which happens after the core stage
We can simply change the filename otherwise

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -199,8 +199,10 @@
...
   end
 end
```
srbarrios, line 207: That would be more elegant I guess, as otherwise if that timeout pass, you will just skip the command but it will not raise any kind of error and you will not notice it.

In any case, we have a default wait time, not sure if it's lower than 5.

```suggestion
  When(/^I include the recommended child channels$/) do
    toggle_off = "//i[contains(@class, 'fa-toggle-off')]"
    raise 'The toggle is not expanded' unless page.has_xpath?(toggle_off, wait: 5)
    
    toggle = "//span[@class='pointer']"
    if page.has_xpath?(toggle, wait: 5)
      find(:xpath, toggle).click
    end
  end
```

elariekerboull: We do want to skip it if it does not find it, otherwise it raises an error

elariekerboull: Basically, if the toggle is already on, we do not want to do anything

elariekerboull: Ah but if the toggle is missing completely it might be an issue, so it can still be adapted

elariekerboull: Maybe
```
toggle = "//span[@class='pointer']"
raise 'The toggle is not present' unless page.has_xpath?(toggle, wait: 5)
toggle_off = "//i[contains(@class, 'fa-toggle-off')]"
find(:xpath, toggle).click if page.has_xpath?(toggle_off, wait: 5)
```

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -418,7 +426,8 @@
   node = get_target(host)
   arch, _code = node.run('uname -m')
   arch.chomp!
-  cmd = "zgrep '#{text}' /var/cache/zypp/raw/susemanager:test-channel-#{arch}/repodata/*updateinfo.xml.gz"
+  # TODO: adapt for architectures
+  cmd = "zgrep '#{text}' /var/cache/zypp/raw/susemanager:test-channel-for-sle/repodata/*updateinfo.xml.gz"
```
srbarrios, line 430: Yeah, I think this TODO must be done before we merge this PR.
This step might probably be used also in BV.

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -299,7 +304,7 @@
   unless out.include? 'ssh-push-tunnel'
     steps %(
       Given I am authorized as "testing" with password "testing"
-      And I follow this "#{host}" link
+      And I am on the Systems overview page of this "sle_client"
```
srbarrios, line 307: No no, don't hardcore this step with "sle_client" as this might be used with other values in "host", keep using #{host} but ok to use this different step.

elariekerboull: Good thing you pointed that out, I had this same change in another PR that was actually already merged in some branches
I'll fix that asap

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -251,6 +251,11 @@
   step %(I select "#{label}" from "original_id")
 end
 
+When(/^I select the custom architecture channel for "(.*?)" as the origin channel$/) do |client|
+  label = deb_host?(client) ? "Test-Channel-Deb-AMD64 Child Channel" : "Test-Channel-x86_64 Child Channel"
```
srbarrios, line 255: I would suggest to have a dict/map into constants.rb, so you can have a unique source of this info, and then you select it here with something like TEST_CHANNEL_BY_ARCH[node.architecture] after including an attribute architecture in our Labanda module.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -1032,7 +1032,7 @@
 Then(/^The metadata buildtime from package "(.*?)" match the one in the rpm on "(.*?)"$/) do |pkg, host|
   # for testing buildtime of generated metadata - See bsc#1078056
   node = get_target(host)
-  cmd = "dumpsolv /var/cache/zypp/solv/spacewalk\:test-channel-x86_64/solv | grep -E 'solvable:name|solvable:buildtime'| grep -A1 '#{pkg}$'| perl -ne 'if($_ =~ /^solvable:buildtime:\\s*(\\d+)/) { print $1; }'"
+  cmd = "dumpsolv /var/cache/zypp/solv/spacewalk\:test-channel-for-sle/solv | grep -E 'solvable:name|solvable:buildtime'| grep -A1 '#{pkg}$'| perl -ne 'if($_ =~ /^solvable:buildtime:\\s*(\\d+)/) { print $1; }'"
```
srbarrios, line 1035: That might need to be adapted to the architecture as the other TODOS.

testsuite/features/step_definitions/common_steps.rb
elariekerboull: It is apparently currently not used anymore as it was only for the client, I'll still adapt it before merging so it will be ready

testsuite/features/step_definitions/api_common.rb
```diff
@@ -479,7 +479,7 @@ def wait_action_complete(actionid, timeout: DEFAULT_TIMEOUT)
   channel = if arch != 'x86_64'
               'test-channel-i586'
             else
-              'test-channel-x86_64'
+              'test-channel-for-sle'
```
srbarrios, line 482: Not really fun with this label name, as I said above, I think it will be better to have a label name, matching the channel name, and including then x86_64 word on it.

testsuite/features/secondary/srv_distro_cobbler.feature
```diff
@@ -140,7 +140,8 @@ Feature: Cobbler and distribution autoinstallation
     And I wait until file "/srv/tftpboot/pxelinux.0" exists on server
 
   Scenario: Trigger the creation of a cobbler system record
-    When I trigger cobbler system record on the "sle_minion"
+    When I wait for "300" seconds
```
srbarrios, line 143: No pleaseeee! This wait no!!! :D 
Try to use a step that use a timeout, checking it periodically, but avoid waiting idle for 300 seconds

testsuite/features/secondary/srv_content_lifecycle.feature
```diff
@@ -156,6 +156,6 @@ Feature: Content lifecycle
   Scenario: Clean up the Content Lifecycle Management feature
     When I follow the left menu "Content Lifecycle > Projects"
     And I follow "clp_name"
-    And I click on "Delete"
-    And I click on "Delete" in "Delete Project" modal
-    Then I should see a "There are no entries to show." text
+    #And I click on "Delete"
+    #And I click on "Delete" in "Delete Project" modal
+    #Then I should see a "There are no entries to show." text
```
srbarrios, line 161: Remove it, if we don't need them anymore.

testsuite/features/step_definitions/common_steps.rb
elariekerboull: This step was created to anticipate the use of synced products for other architecture so right now it is single-use
However if we might use it for other architectures in the future it is a good idea to do that

testsuite/features/secondary/minkvm_guests.feature
```diff
@@ -14,14 +14,24 @@ Feature: Be able to manage KVM virtual machines via the GUI
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
 
+  Scenario: Create KVM activation key
```
srbarrios, line 17: Do we need this key for something special?
I would say we can reuse "Test-Channel-x86_64" right?

testsuite/features/secondary/min_salt_software_states.feature
```diff
@@ -85,13 +88,14 @@ Feature: Salt package states
...
   Scenario: Upgrade a package through the UI
-    Then I follow "States" in the content area
+    When I follow "States" in the content area
     And I follow "Packages"
     And I follow "Search"
     And I should see a "Package States" text
     And I list packages with "dummy"
-    Then I should see a "andromeda-dummy" text
-    And "andromeda-dummy-1.0" should be installed on "sle_minion"
+    And I wait for "60" seconds
```
srbarrios, line 96: Same hehe

testsuite/features/secondary/min_salt_software_states.feature
```diff
@@ -55,13 +56,14 @@ Feature: Salt package states
...
   Scenario: Install a package through the UI
-    Then I follow "States" in the content area
+    When I follow "States" in the content area
     And I follow "Packages"
     And I follow "Search"
     And I should see a "Package States" text
     And I list packages with "dummy"
-    Then I should see a "milkyway-dummy" text
-    And "milkyway-dummy" should not be installed on "sle_minion"
+    And I wait for "60" seconds
```
srbarrios, line 64: Same same! Avoid using explicit waits.

testsuite/features/step_definitions/command_steps.rb
elariekerboull: That was also exclusively used by the trad client, should we still keep this step?

testsuite/features/secondary/min_salt_software_states.feature
srbarrios: If it's a problem on the PR pipeline, we need to find a way to fix it at that level, but waiting 60 seconds idle for sure is not an option.

testsuite/features/secondary/srv_distro_cobbler.feature
elariekerboull: Another Net Timeout, in the test runs it didn't have anything to check on, I could not find any other solution
But once again it might not be needed in the main pipelines

testsuite/features/secondary/srv_content_lifecycle.feature
elariekerboull: Might be something I forgot to uncomment and is still needed

testsuite/features/secondary/minkvm_guests.feature
elariekerboull: There were issues with using the same activation key with the base product if I recall correctly and it needed its own key (or one of the others keys whicj would have to be renamed), I'll check again what the issue was though

testsuite/features/secondary/min_salt_software_states.feature
elariekerboull: I'll reproduce it to see how we can fix it, but there was absolutely no logs to identify the issue, and the ajax transition did not work on it, it was really odd

testsuite/features/init_clients/srv_create_activationkey.feature
elariekerboull: Not cloning would imply adding scenarios from https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/core/srv_create_repository.feature#L13 to https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/core/srv_create_repository.feature#L51 in order to recreate the channel
To be discussed to see if it is worth it

testsuite/features/step_definitions/common_steps.rb
elariekerboull: On master branch the corresponding line  seems to not make any sense since it refers to `zypp` directory with various architectures https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/step_definitions/common_steps.rb#L421

I would rather see this fixed in a different PR, I can drop changes related to it in this PR, mostly since it is only relevant to 4.3 and 4.2 now

testsuite/features/secondary/min_salt_software_states.feature
elariekerboull: The timeout occurs when waiting for the text to appear and is completely different from the timeout expected when waiting the text: 
```
When(/^I wait until I see "([^"]*)" text$/) do |text|
  raise "Text '#{text}' not found" unless has_text?(text, wait: DEFAULT_TIMEOUT)
end
```
image

Searching for more information online provided this page: https://www.exceptionalcreatures.com/bestiary/Net/ReadTimeout

It seems to be an issue that cannot be solved by waiting for an element of the page.
Keeping this failing results in the scenario failing after 400 seconds, when keeping this explicit wait makes it succeed after 120 seconds.

It is beyond my understanding of what is happening and would require time and help to identify what the whole problem is.
I will keep these explicit waits until the PR is ready to merge to avoid these scenarios failing at every test run, hopefully this issue will not appear on the main pipelines.

testsuite/features/secondary/allcli_action_chain.feature
srbarrios: OK, I see what you mean. What a limitation! OK, let's change it then.
@mcalmer I think this could be an interesting scenario for your awareness.

testsuite/features/init_clients/srv_create_activationkey.feature
srbarrios: From my point of view yup, it's worth to do it that way.

elariekerboull: That also means repeating it for every activation key we want to associate to a base product in the future
I'll apply it for now, but it might need to be optimized in the future

testsuite/features/secondary/srv_distro_cobbler.feature
elariekerboull: Following the discussion [on that topic](https://suse.slack.com/archives/C02D12TNYLS/p1665740743091379) I will remove this wait once everything else is ready along with the other extra explicit waits

testsuite/run_sets/init_clients.yml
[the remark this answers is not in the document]
srbarrios: As far as I know, you don't need to have the reposync completed to create an AK key.
You can create it even if the sync is in progress.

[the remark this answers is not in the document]
elariekerboull: We still need the reposync to have started at least
We could create the activation key in core without any base channel and modify it in one of the init_client features, that would make it more modulable

testsuite/features/secondary/min_monitoring.feature
```diff
@@ -76,6 +76,3 @@ Feature: Monitor SUMA environment with Prometheus on a SLE Salt minion
     And I click on "Apply Highstate"
     Then I should see a "Applying the highstate has been scheduled." text
     And I wait until event "Apply highstate scheduled by admin" is completed
-
```
Bischoff, line 79: looks like a mistake to remove this

testsuite/features/secondary/min_salt_pkgset_beacon.feature
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2016-2021 SUSE LLC
+# Copyright (c) 2016-2022 SUSE LLC
```
Bischoff, line 1: correct ! good catch

testsuite/features/secondary/min_salt_software_states.feature
```diff
@@ -100,22 +104,22 @@ Feature: Salt package states
     And I wait for "andromeda-dummy-2.0-1.1" to be installed on "sle_minion"
 
   Scenario: Verify the package states
-    Then I follow "States" in the content area
+    When I follow "States" in the content area
```
Bischoff, line 107: thanks

testsuite/features/step_definitions/api_common.rb
```diff
@@ -479,7 +479,7 @@ def wait_action_complete(actionid, timeout: DEFAULT_TIMEOUT)
   channel = if arch != 'x86_64'
               'test-channel-i586'
```
Bischoff, line 480: missing `fake-` probably

testsuite/features/support/constants.rb
```diff
@@ -256,6 +256,11 @@
     sle-module-containers15-sp4-updates-x86_64
     sle-module-basesystem15-sp4-updates-x86_64
     sle-module-server-applications15-sp4-updates-x86_64
+    sle15-sp4-installer-updates-x86_64
+    sle-module-desktop-applications15-sp4-pool-x86_64
+    sle-module-desktop-applications15-sp4-updates-x86_64
+    sle-module-devtools15-sp4-pool-x86_64
+    sle-module-devtools15-sp4-updates-x86_64
```
Bischoff, line 263: I am wondering if we can spare some of them, like the development tools

Bischoff (changes requested):
i think i found a few issues - please fix

the rest is probably a matter of taste, feel free to ignore, or not

testsuite/features/init_clients/srv_create_activationkey.feature
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) 2010-2022 SUSE LLC
```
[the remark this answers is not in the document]
elariekerboull: It was moved from another folder, hence the 2010 despite it appearing to be created
Related to this comment: https://github.com/uyuni-project/uyuni/pull/5900#discussion_r994743175

testsuite/features/step_definitions/api_common.rb
elariekerboull: Apparently did not update this name yet, good catch

testsuite/features/support/constants.rb
elariekerboull: I believe I added all those that were necessary for the features in the testsuite, so they would be the minimum required
I'll double check on a test branch just in case and document it in the PR description

testsuite/features/core/srv_create_repository.feature
```diff
@@ -13,15 +13,15 @@ Feature: Add a repository to a channel
   Scenario: Add a test repository for x86_64
     When I follow the left menu "Software > Manage > Repositories"
     And I follow "Create Repository"
-    And I enter "Test-Repository-x86_64" as "label"
+    And I enter "fake-rpm-repo" as "label"
```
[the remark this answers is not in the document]
srbarrios: :D I suggested to @elariekerboull to use "fake", because using the word "test" was for me too ambiguous.

testsuite/features/secondary/allcli_action_chain.feature
```diff
@@ -73,16 +58,16 @@ Feature: Action chains on several systems at once
   Scenario: Add an action chain using system set manager for Red Hat-like minion and SLE minion
     When I follow the left menu "Systems > Overview"
     And I check the "sle_minion" client
-    And I check the "rhlike_minion" client
+    And I check the "ssh_minion" client
     And I follow the left menu "Systems > System Set Manager > Overview"
     And I follow "Install" in the content area
-    And I follow "Test-Channel-x86_64" in the content area
+    And I follow "Fake-RPM-SLES15SP4-Channel" in the content area
```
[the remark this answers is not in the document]
srbarrios: I'm not even sure if what you suggested is feasible, but I think that for the scope of this PR, let's avoid refactoring test cases.
And focusing only in using SCC products for our registered clients, plus the custom channel for the test/fake repo.

testsuite/features/support/constants.rb
srbarrios: DevTools are sadly a parent of one of the channels we need, so it's a requirement we can't remove.

testsuite/features/secondary/minkvm_guests.feature
elariekerboull: Keeping it that way for now since proxy product might be involved.

testsuite/features/core/srv_create_repository.feature
```diff
@@ -50,40 +50,61 @@ Feature: Add a repository to a channel
     And I wait until the channel "test-channel-x86_64" has been synced
     And I disable source package syncing
 
+  Scenario: Add the repository to the x86_64 child channel
+    When I follow the left menu "Software > Manage > Channels"
+    And I follow "Test-Channel-x86_64 Child Channel"
```
srbarrios, line 55: Double check if "Test-Channel-x86_64 Child Channel" is the name we have now, I would say that is changed now by "Fake-RPM-SLES15SP4-Channel" right? It's hard to follow it, maybe we need a diagram of the current implementation, similar to what I did as first diagram draft but with the current values for everything. (We could even add the diagram as part of our documentation inside the testsuite folder)

testsuite/features/core/srv_create_repository.feature
```diff
@@ -50,40 +50,61 @@ Feature: Add a repository to a channel
...
+    And I enter "file:///etc/pki/rpm-gpg/uyuni-tools-gpg-pubkey-0d20833e.key" as "GPG key URL"
+    And I click on "Update Channel"
+    Then I should see a "Channel Test-Channel-x86_64 Child Channel updated" text
+    When I follow "Repositories" in the content area
+    And I select the "fake-rpm-repo" repo
+    And I click on "Save Repositories"
+    Then I should see a "Test-Channel-x86_64 Child Channel repository information was successfully updated" text
+
+  Scenario: Synchronize the repository in the x86_64 child channel
+    When I enable source package syncing
```
mcalmer, line 65: Here you enable source package syncing, but you do not disable it again.
From here on, all repos try to mirror source packages

mcalmer (dismissed):

testsuite/features/core/srv_create_repository.feature
elariekerboull: In that case this specific child channel is not needed anymore as the `Fake-RPM-SLES15SP4-Channel` one is itself a child channel.
Following Michael's findings I reviewed the need of these steps and removed them in commits to come.
The last references to this child channel are in https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/core/srv_channels_add.feature#L24, wondering if we should plainly remove this, refactor the feature so it deletes the child channel or just leave it be.

testsuite/run_sets/reposync.yml
```diff
@@ -13,7 +13,6 @@
 - features/reposync/srv_sync_products.feature
 - features/reposync/srv_enable_sync_products.feature
 - features/reposync/srv_wait_for_reposync.feature
-- features/reposync/srv_check_reposync.feature
```
elariekerboull, line 16: Realised this one is the correct one and the one in init_clients should not be there, currently fixing.

testsuite/features/core/srv_create_activationkey.feature
```diff
@@ -24,28 +23,32 @@ Feature: Create activation keys
     And I should see a "Groups" link
     And I should see a "Activated Systems" link
 
+@rhlike_minion
+  Scenario: Create an activation key for RedHat-like minion
+    When I follow the left menu "Systems > Activation Keys"
+    And I follow "Create Key"
+    And I enter "RedHat like Test Key" as "description"
+    And I enter "RH-LIKE-KEY" as "key"
+    And I select "Test-Channel-x86_64" from "selectedBaseChannel"
```
Bischoff, line 32: Why not `Fake-` here ? Or do I miss something?

testsuite/features/init_clients/buildhost_bootstrap.feature
```diff
@@ -14,6 +14,7 @@ Feature: Bootstrap a Salt build host via the GUI
      And I enter "22" as "port"
      And I enter "root" as "user"
      And I enter "linux" as "password"
+    And I select "1-SUSE-KEY-x86_64" from "activationKeys"
```
Bischoff, line 17: is indentation correct ?

testsuite/features/support/constants.rb
Bischoff: ok

Bischoff (changes requested):
some last changes

testsuite/features/core/srv_create_activationkey.feature
elariekerboull: Ah right I should create a new channel for RedHat-like minions instead of reusing this one, it does not make sense to keep this channel here, nice catch.

testsuite/features/init_clients/buildhost_bootstrap.feature
elariekerboull: Something is off yes, looking into it

Bischoff (approved):
thanks for all the changes

PR #5906, opened 2022-09-12
Sanity check for resource for custom endpoint test

Bischoff:
## What does this PR change?

This PR checks resource FTP port 445 on minima-mirror.mgr.{suse.de,prv.suse.net}

## Links

Ports:
* 4.2:
* 4.3:

## Changelogs

- [x] No changelog needed


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -71,6 +71,12 @@
   $server.run("curl --insecure --proxy '#{proxy}' --proxy-anyauth --location '#{url}' --output /dev/null")
 end
 
+Then(/^it should be possible to use the FTP server$/) do
+  # Use an URL in Provo after the release
```
vandabarata, line 75:
```suggestion
  # TODO: Use a URL in Provo after the release
```
It's easier to track our temporary "hacks" if we have them as a TODO :wink:

Bischoff: right

vandabarata (approved):

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -71,6 +71,12 @@
   $server.run("curl --insecure --proxy '#{proxy}' --proxy-anyauth --location '#{url}' --output /dev/null")
 end
 
+Then(/^it should be possible to use the FTP server$/) do
+  # TODO: use a URL in Provo after the release
+  url = 'ftp://minima-mirror.mgr.suse.de:445/rhn/manager/download/test-channel-x86_64/repodata/repomd.xml'
```
srbarrios, line 76: That domain should be parameterizable somehow... depending if you are in NUE or PRV.

Bischoff: yes, but then I have to do the same for the SLS pillar file, and that complicates another step

I'd like to do something simple today, we can make it more intelligent some other day

Bischoff: I'll create a card to make it better

PR #5922, opened 2022-09-14
Use package download endpoint in NUE for HEAD and Uyuni

Bischoff:
## What does this PR change?

Follow-up of 5904

Use minima-mirror.mgr.suse.de when testing in NUE.

## Links

No ports, Uyuni branch only

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/secondary/min_custom_pkg_download_endpoint.feature
+++ b/testsuite/features/secondary/min_custom_pkg_download_endpoint.feature
@@ -54,7 +54,8 @@ Feature: Repos file generation based on custom pillar data
   Scenario: Check the channel.repo file to see the custom RPM download point
     Given I am on the Systems overview page of this "sle_minion"
     Then the susemanager repo file should exist on the "sle_minion"
-    And I should see "ftp", "minima-mirror.mgr.prv.suse.net" and "445" in the repo file on the "sle_minion"
+    # TODO: move the endpoint to Provo after the release
+    And I should see "ftp", "minima-mirror.mgr.suse.de" and "445" in the repo file on the "sle_minion"
 
   Scenario: Cleanup: remove the custom RPM download point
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

testsuite/features/secondary/min_custom_pkg_download_endpoint.feature
```diff
@@ -54,7 +54,8 @@ Feature: Repos file generation based on custom pillar data
   Scenario: Check the channel.repo file to see the custom RPM download point
     Given I am on the Systems overview page of this "sle_minion"
     Then the susemanager repo file should exist on the "sle_minion"
-    And I should see "ftp", "minima-mirror.mgr.prv.suse.net" and "445" in the repo file on the "sle_minion"
+    # TODO: move the endpoint to Provo after the release
+    And I should see "ftp", "minima-mirror.mgr.suse.de" and "445" in the repo file on the "sle_minion"
```
srbarrios, line 58: Approving, but as I suggested in the previous PR we should have it dynamically depending the environment where we are.

```
domain, _code = node.run("grep '^search' /etc/resolv.conf | sed 's/^search//'")
mirror_fqdn = "minima-mirror.#{domain}"
```

And not expose the fqdn as part of the Gherkin file. Keep it only inside the step definition code or if you prefer using a initialized Twopence object with its attributes.

srbarrios (approved):

testsuite/features/secondary/min_custom_pkg_download_endpoint.feature
Bischoff: I updated the card that discusses this: #18953

And the best solution is to have an environment variable & have this fully configurable in the main.tf.

PR #5927, opened 2022-09-14
Revert "Sanity check for resource for custom endpoint test"

jordimassaguerpla:
Reverts uyuni-project/uyuni#5906

PR testing is done in PRV and we can't access minima-mirror in NUE. Let's not do this test until https://github.com/SUSE/spacewalk/issues/18953 is fixed.


Bischoff (changes requested):
I'd prefer we just comment in the sanity check feature, it is supposed to be immadiately reverted again

jordimassaguerpla:
> I'd prefer we just comment in the sanity check feature, it is supposed to be immadiately reverted again

https://github.com/uyuni-project/uyuni/pull/5928

Bischoff:
thx

PR #5931, opened 2022-09-15
QE: Printing node OS in twopence_init.rb

elariekerboull:
## What does this PR change?

Prints the OS version of each node in the testsuite at runtime.

## Documentation

- [x] **DONE**

## Links

Fixes # https://github.com/SUSE/spacewalk/issues/18722
Tracks # 
4.2
4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/twopence_init.rb
+++ b/testsuite/features/support/twopence_init.rb
@@ -132,7 +132,7 @@ $nodes.each do |node|
 
   STDOUT.puts "Host '#{$named_nodes[node.hash]}' is alive with determined hostname #{hostname.strip} and FQDN #{fqdn.strip}" unless $build_validation
   result, _code = node.run('grep PRETTY /etc/os-release')
-  puts "'#{$named_nodes[node.hash]}' is running OS #{result}"
+  STDOUT.puts "'#{$named_nodes[node.hash]}' is running OS #{result}"
 end
 
 # This function is used to get one of the nodes based on its type
--- a/testsuite/features/support/twopence_init.rb
+++ b/testsuite/features/support/twopence_init.rb
@@ -131,6 +131,8 @@ $nodes.each do |node|
   node.init_full_hostname(fqdn)
 
   STDOUT.puts "Host '#{$named_nodes[node.hash]}' is alive with determined hostname #{hostname.strip} and FQDN #{fqdn.strip}" unless $build_validation
+  result, _code = node.run('grep PRETTY /etc/os-release')
+  puts "'#{$named_nodes[node.hash]}' is running OS #{result}"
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

testsuite/features/support/twopence_init.rb
```diff
@@ -131,6 +131,8 @@ def twopence_init(target)
   node.init_full_hostname(fqdn)
 
   STDOUT.puts "Host '#{$named_nodes[node.hash]}' is alive with determined hostname #{hostname.strip} and FQDN #{fqdn.strip}" unless $build_validation
+  result, _code = node.run('grep PRETTY /etc/os-release')
+  puts "'#{$named_nodes[node.hash]}' is running OS #{result}"
```
srbarrios, line 135: Fine for now.
I think we can have in the future an improvement, and have a new attribute in Lavanda class, that contains interesting info from that file.

And then do something like:
node.os_release(result)

INitializing for instance attributes like:
node.os_release_id
node.os_release_version_id


Getting rid of this common called method:
```
def get_os_version(node)
  os_family_raw, code = node.run('grep "^ID=" /etc/os-release', check_errors: false)
  return nil, nil unless code.zero?
  os_family = os_family_raw.strip.split('=')[1]
  return nil, nil if os_family.nil?
  os_family.delete! '"'
  os_version_raw, code = node.run('grep "^VERSION_ID=" /etc/os-release', check_errors: false)
  return nil, nil unless code.zero?
  os_version = os_version_raw.strip.split('=')[1]
  return nil, nil if os_version.nil?
  os_version.delete! '"'
  # on SLES, we need to replace the dot with '-SP'
  os_version.gsub!(/\./, '-SP') if os_family =~ /^sles/
  STDOUT.puts "Node: #{node.hostname}, OS Version: #{os_version}, Family: #{os_family}"
  [os_version, os_family]
end
```

Bischoff, line 135: You could also have used `node.full_hostname` or `hostname.strip`, but that's probably ok that way too

In the past, `STDOUT.puts` was needed to work around jenkin's buffering, but maybe it's not needed anymore

Bischoff (approved):
just left remarks

testsuite/features/support/twopence_init.rb
elariekerboull: Switched `puts` with `STDOUT.puts` just in case and to be consistent with its use two lines above.

elariekerboull: I'll create a card with this suggestion

PR #5936, opened 2022-09-15
Fix tftpsync plugin problems

SchoolGuy:
## What does this PR change?

- Cleans up temporary TFTP directory on the Proxy to prevent an out-of-diskspace problem
- More granularly selects the files that need to be replaced during transfer.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Fixes [#13909](https://github.com/SUSE/spacewalk/issues/13909)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


tftpsync/susemanager-tftpsync-recv/add.wsgi
```diff
@@ -45,17 +43,19 @@ ch.setFormatter(formatter)
...
 logger.addHandler(ch)
 
+
 class TftpFieldStorage(cgi.FieldStorage):
 
     def make_file(self, binary=None):
         tmpdir = os.path.join(CFG.TFTPBOOT, "tmp")
         if not os.path.exists(tmpdir):
-                os.makedirs(tmpdir)
+            os.makedirs(tmpdir)
```
vzhestkov, line 52: For me it looks like the race condition could appear here also.
Maybe avoid it this way:
```suggestion
            os.makedirs(tmpdir, exist_ok=True)
```

tftpsync/susemanager-tftpsync-recv/add.wsgi
```diff
@@ -140,3 +144,10 @@ def application(environ, start_response):
     return [content.encode()]
 
 
+def application(environ, start_response):
+    try:
+        _application(environ, start_response)
+    finally:
+        # Cleanup tmpdir to cleanup temporary files in case server times out
+        tempdir = os.path.join(CFG.TFTPBOOT, "tmp")
+        shutil.rmtree(tempdir)
```
meaksh, line 153: Could it happen that this cleanup runs while another execution is being performed? In the case of several files being pushed at the same time?

SchoolGuy: Yes, that indeed is a problem. But we cannot prevent that from happening easily as we require the cleanup due to SUMA server timeouts that might occur. Since this is a temporary directory, this should only require a second `cobbler sync` and all the necessary files should be present.

My assumption is that during the next `cobbler sync` not the exact same file is running into the problem you just described. If that assumption would be wrong, we would run into a loop.

SchoolGuy: @meaksh Would you be okay if we clean up files that are older than 12h? A `cobbler sync` should never take longer than an hour in ideal conditions so we should have plenty of time to finish the sync up with an 11h buffer.

meaksh: IIUC this cleanup would run on every file that is pushed to the server, and not once per each cobbler sync execution. If that is the case, I wouldn't run this cleanup (even if only for 12h files) on every single file pushed.

What if we create another wsgi script, that is triggered at the end of `cobbler sync` execution, just in order to perform this cleanup of older files after `cobbler sync` execution finished?

SchoolGuy: That sounds very reasonable imho as it is backwards compatible and would solve the problem from the date the submission is integrated onwards. However, due to the containerization efforts, I would say that a complete refactor of all endpoints is then a more reasonable thing to do as `wsgi` scripts in containers are not so "nice".

SchoolGuy: I have found a better solution for the future:

We can introduce signal the sync with a UUID that we send over the wire. That would identify clearly the `cobbler sync` that is currently writing to disk. That also gives us the advantage that we can log those uuids on the proxy side to gain an increased insight in to the data flow.

PR #5962, opened 2022-09-21
import gpg keys on SUSE systems directly

mcalmer:
## What does this PR change?

When using `gpgkeys` parameter in the repository configuration, the installer import the key only when needed.
As the repositories are not signed, a refresh operation does not import the key.
Only a package install/update operation would import this key, but we do not want to use "gpg-auto-import-gpg-keys" option in every state and teach also the users to set this option always.

So we better directly import the GPG keys which are set in **assigned** channels.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **nothing relevant changed**

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Port of https://github.com/SUSE/spacewalk/pull/19074

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/channels/gpg-keys.sls
+++ b/susemanager-utils/susemanager-sls/salt/channels/gpg-keys.sls
@@ -12,6 +12,16 @@ mgr_deploy_customer_gpg_key:
     - source: salt://gpg/mgr-gpg-pub.key
     - makedirs: True
     - mode: 644
+
+{%- if grains['os_family'] == 'Suse' %}
+mgr_trust_customer_gpg_key:
+  cmd.run:
+    - name: rpm --import /etc/pki/rpm-gpg/mgr-gpg-pub.key
+    - runas: root
+    - require:
+      - file: mgr_deploy_customer_gpg_key
+{%- endif %}
+
 {%- endif %}
 {%- endif %}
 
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/salt/channels/gpg-keys.sls
```diff
@@ -12,6 +12,16 @@ mgr_deploy_customer_gpg_key:
...
     - makedirs: True
     - mode: 644
+
+{%- if grains['os_family'] == 'Suse' %}
+mgr_trust_customer_gpg_key:
+  cmd.run:
+    - name: rpm --import /etc/pki/rpm-gpg/mgr-gpg-pub.key
+    - runas: root
+    - require:
+      - file: mgr_deploy_customer_gpg_key
```
agraul, line 22: This will run `rpm --import` every time this state is applied, is that needed? If not, I'd add a [`onchanges`](https://docs.saltproject.io/en/latest/ref/states/requisites.html#onchanges) requisite.

mcalmer: I changed this in the PR of 4.3.
As we changed a lot more in 4.3, I will close this PR and create a new one with all the changes we did there

mcalmer:
Needs to be re-done as the source PR has changed a lot

PR #5966, opened 2022-09-21
Add check translations

jordimassaguerpla:
## What does this PR change?

This adds back the need translation check but instead of failing it adds the "needs_translation" label.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation
- No documentation needed
- 
- [ ] **DONE**

## Test coverage
- No tests

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/18640

- [ ] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/check_translations.yml
+++ b/.github/workflows/check_translations.yml
@@ -54,8 +54,6 @@ jobs:
 
     - name: Add label
       if: ${{ env.translations_needed }} == 'true'
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       uses: andymckay/labeler@1.0.4
       with:
         add-labels: "needs-translation"
--- a/.github/workflows/check_translations.yml
+++ b/.github/workflows/check_translations.yml
@@ -1,0 +1,64 @@
+name: Check if translation is required
+
+on: 
+  pull_request_target:
+    paths:
+      - 'java/code/src/com/redhat/rhn/frontend/strings/**'
+      - 'backend/**'
+      - 'client/rhel/yum-rhn-plugin/**'
+      - 'client/rhel/mgr-daemon/**'
+      - 'client/rhel/spacewalk-client-tools/**'
+      - 'web/**'
+      - 'susemanager/**'
+      - 'spacecmd/**'
+
+jobs:
+  run:
+    name: Check by trying to locally update translation files 
+    if: github.repository == 'uyuni-project/uyuni'
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
I  have tested it with a branch and with a fork:

https://github.com/uyuni-project/uyuni/pull/5965
https://github.com/uyuni-project/uyuni/pull/5967

.github/workflows/check_translations.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo "{translations_needed}={true}" >> $GITHUB_ENV
+        else
+          echo "{translations_needed}={false}" >> $GITHUB_ENV
+        fi
+
+    - name: Add label
+      if: ${{ env.translations_needed }} == 'true'
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      uses: andymckay/labeler@master
```
renner, line 59: Generally it is not recommended to use `master` versions of actions, probably better to go for a stable version like `@v1` or `@1.0.2`. Not a big issue though, I know he is using `master` also in the upstream examples.

By the way, [labeler.yml](https://github.com/uyuni-project/uyuni/blob/master/.github/workflows/labeler.yml) from QE is using `actions/labeler@v4`.

.github/workflows/check_translations.yml
```diff
@@ -0,0 +1,64 @@
...
+          git diff origin_check_translations check_translations
+          echo "{translations_needed}={true}" >> $GITHUB_ENV
+        else
+          echo "{translations_needed}={false}" >> $GITHUB_ENV
+        fi
+
+    - name: Add label
+      if: ${{ env.translations_needed }} == 'true'
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
renner, line 58: Is the token needed at all (here and below), and why twice - as env variable as well as a parameter? Basic usage examples from [upstream](https://github.com/andymckay/labeler) are not referencing the `GITHUB_TOKEN`.

renner (approved):
Okay, it looks fine to me, thanks Jordi! I am assuming that all steps previous to the labeling are working as expected.

.github/workflows/check_translations.yml
jordimassaguerpla: I have updated it to use latest release version 1.0.4.
actions/labeler uses a yaml file to decide which labels to add depending on the directory structure. It is another use case.

.github/workflows/check_translations.yml
jordimassaguerpla: Thanks. The token is needed , otherwise you do not have permissions to add a label.
I have removed the duplication .

PR #5996, opened 2022-09-29
CI: fix needs translation

jordimassaguerpla:
## What does this PR change?

We need to compare if the PR needs more translations than the master branch. Without this fix, we were looking whether the master branch needed translations, because we use the event "pull_request_target", and with this event, the checkout is from the base branch (base branch=target branch), instead for the branch from the PR.

## GUI diff

This is a fix for a github action (CI), not the product.

- [X] **DONE**

## Documentation
- No documentation needed

This is a fix for a github action (CI), not the product.

- [X] **DONE**

## Test coverage
- No tests

This is a fix for a github action (CI), not the product.

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/19144

- [ ] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/check_translations.yml
+++ b/.github/workflows/check_translations.yml
@@ -18,45 +18,70 @@ jobs:
     if: github.repository == 'uyuni-project/uyuni'
     runs-on: ubuntu-latest
     steps:
+    - name: Setup tooling
+      run: sudo apt-get install -y make git gettext intltool python3
 
     - name: Cancel Previous Runs
       uses: styfle/cancel-workflow-action@0.8.0
       with:
         access_token: ${{ github.token }}
 
-    - name: Checkout repo
-      uses: actions/checkout@v2
-
-    - name: Setup tooling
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
@renner : hi! This is WIP because I am setting up some tests, but still I'd like to hear your opinion on the strategy used, which is to check if translations are needed for master, or better said, how many, and then compare that with the code from your PR.

Otherwise, we were getting false positives when translations were requested from master. Actually, the previous code was **only** looking into master because of using the event "pull_request_target", which makes the checkout pull from the target/base branch, instead of from the PR.

jordimassaguerpla:
@renner This is ready for you to review.

renner (approved):
It's a bit more complex now, but it looks good. I understand we check for the translations needed status in both branches (`master` and `feature`-branch) and notify (label) only if it differs. Otherwise if there is translations needed for `master` at the time of branching, we would always notify.

PR #6023, opened 2022-10-12
Perform refresh with packages.pkgupdate state (bsc#1203884)

vzhestkov:
## What does this PR change?

Add `refresh` to `packages.pkgupdate` the same way as set for `packages.pkginstall` as it could fail on updating the packages on the client if the repo was not refreshed before and has no this update in the cached repo metadata yet.

## GUI diff

No difference.

## Documentation
- No documentation needed: only internal and user invisible changes

## Links

Tracks https://github.com/SUSE/spacewalk/issues/19151

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #6026, opened 2022-10-13
Exclude invalid mirror urls for reposync

zzaimeche:
## What does this PR change?

This commit fixes bsc#1203826. Previously, reposync would treat broken urls in the list of mirrors as valid, causing it to fail. The first commit checks whether urls contain any forbidden characters, and if they do, it discards them. The second commit checks that any page used as a candidate for a mirrorlist is a plaintext or xml page.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [x] **DONE**

## Re-run a test

### DIFF
--- a/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
@@ -31,6 +31,7 @@ import bz2
 import lzma
 import os
 import re
+import requests
 import solv
 import subprocess
 import sys
@@ -525,6 +526,21 @@ class ContentSource:
         mirrorlist_path = os.path.join(repo.root, 'mirrorlist.txt')
         returnlist = []
         content = []
+        # If page not plaintext or xml, is not a valid mirrorlist or metalink,
+        # so continue without it.
+        proxies = get_proxies(self.proxy_url, self.proxy_user, self.proxy_pass)
+        cert = (self.sslclientcert, self.sslclientkey)
+        verify = self.sslcacert
+        try:
+            webpage = requests.get(url, proxies=proxies, cert=cert, verify=verify)
+            content_type = webpage.headers["Content-Type"]
+            if "text/plain" not in content_type and "xml" not in content_type:
+                # Not a valid mirrorlist or metalink; continue without it
+                return returnlist
+        except requests.exceptions.RequestException as exc:
+            self.error_msg("ERROR: Failed to reach repo url: {} - {}".format(url, exc))
+            return returnlist
+
         try:
             urlgrabber_opts = {}
             self.set_download_parameters(urlgrabber_opts, url, mirrorlist_path)
@@ -548,7 +564,7 @@ class ContentSource:
                 )
                 msg = msg.replace(url, repl_url)
                 log(0, msg)
-            # no mirror list found continue without
+            # no mirror list or metalink found continue without
             return returnlist
 
         def _replace_and_check_url(url_list):
@@ -562,6 +578,7 @@ class ContentSource:
                 forbidden_characters = "<>^`{|}"
                 url_is_invalid = [x for x in forbidden_characters if(x in url)]
                 if url_is_invalid:
+                    self.error_msg("Discarding invalid url: {}".format(url))
                     continue
                 try:
                     # This started throwing ValueErrors, BZ 666826
--- a/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
@@ -555,8 +555,13 @@ class ContentSource:
             goodurls = []
             skipped = None
             for url in url_list:
-                # obvious bogons get ignored b/c, we could get more interesting checks but <shrug>
+                # obvious bogons get ignored
                 if url in ['', None]:
+                    continue
+                # Discard any urls containing some invalid characters
+                forbidden_characters = "<>^`{|}"
+                url_is_invalid = [x for x in forbidden_characters if(x in url)]
+                if url_is_invalid:
                     continue
                 try:
                     # This started throwing ValueErrors, BZ 666826
--- a/python/spacewalk/spacewalk-backend.changes
+++ b/python/spacewalk/spacewalk-backend.changes
@@ -1,15 +1,10 @@
+- Exclude invalid mirror urls for reposync (bsc#1203826)
 - require python3-debian version which support new compression
   methods to sync ubuntu22-04 repositories (bsc#1205212)
 - Update system overview table in reposync
 - Keep older module metadata files in database (bsc#1201893)
 - Used the legacy reporting system in spacewalk-debug to obtain
   up-to-date information
-
--------------------------------------------------------------------
-Thu Oct 13 09:28:27 UTC 2022 - Zara Zaimeche <zara.zaimeche@suse.com>
-
-- Exclude invalid mirror urls for reposync (bsc#1203826)
-  * Ignore any mirror urls that contain forbidden characters
 
 -------------------------------------------------------------------
 Wed Sep 28 11:14:14 CEST 2022 - jgonzalez@suse.com
--- a/python/spacewalk/spacewalk-backend.changes
+++ b/python/spacewalk/spacewalk-backend.changes
@@ -6,6 +6,12 @@
   up-to-date information
 
 -------------------------------------------------------------------
+Thu Oct 13 09:28:27 UTC 2022 - Zara Zaimeche <zara.zaimeche@suse.com>
+
+- Exclude invalid mirror urls for reposync (bsc#1203826)
+  * Ignore any mirror urls that contain forbidden characters
+
+-------------------------------------------------------------------
 Wed Sep 28 11:14:14 CEST 2022 - jgonzalez@suse.com
 
 - version 4.4.1-1
--- a/python/test/unit/spacewalk/satellite_tools/test_yum_src.py
+++ b/python/test/unit/spacewalk/satellite_tools/test_yum_src.py
@@ -206,20 +206,58 @@ class YumSrcTest(unittest.TestCase):
         self.repo.root = self.tmpdir
         cs.channel_arch = "arch1"
         grabber_mock = Mock()
+        mirror_request_mock = Mock()
+        mirror_request_mock.return_value.headers = {"Content-Type":"text/plain"}
+        nonmirror_request_mock = Mock()
+        nonmirror_request_mock.return_value.headers = {"Content-Type":"text/html"}
 
-        with patch("spacewalk.satellite_tools.repo_plugins.yum_src.urlgrabber.urlgrab", grabber_mock):
-            with open(fake_mirrorlist_file, "w") as fake_list:
-                fake_list.writelines([
-                    "http://host1/base/$basearch/os/\n",
-                    "http://host2/base/$BASEARCH/os/\n",
-                    "http://host3/base/$ARCH/os/\n",
-                ])
-            mirrors = cs._get_mirror_list(self.repo, "https://fake/repo/url")
-            self.assertEqual(mirrors, [
-                "http://host1/base/arch1/os/",
-                "http://host2/base/arch1/os/",
-                "http://host3/base/arch1/os/",
-            ])
+        # If webpage is plaintext format, check list of mirrors is returned
+        with patch("spacewalk.satellite_tools.repo_plugins.yum_src.requests.get", mirror_request_mock):
+            with patch("spacewalk.satellite_tools.repo_plugins.yum_src.urlgrabber.urlgrab", grabber_mock):
+                with open(fake_mirrorlist_file, "w") as fake_list:
+                    fake_list.writelines([
+                        "http://host1/base/$basearch/os/\n",
+                        "http://host2/base/$BASEARCH/os/\n",
+                        "http://host3/base/$ARCH/os/\n",
+                        ])
+                mirrors = cs._get_mirror_list(self.repo, "https://fake/repo/url")
+                self.assertEqual(mirrors, [
+                    "http://host1/base/arch1/os/",
+                    "http://host2/base/arch1/os/",
+                    "http://host3/base/arch1/os/",
+                    ])
+
+        # If webpage is html format, and so not mirrorlist, check list of 'mirrors' is blank
+        with patch("spacewalk.satellite_tools.repo_plugins.yum_src.requests.get", nonmirror_request_mock):
+            with patch("spacewalk.satellite_tools.repo_plugins.yum_src.urlgrabber.urlgrab", grabber_mock):
+                with open(fake_mirrorlist_file, "w") as fake_list:
+                    fake_list.writelines([
+                        "http://host1/base/$basearch/os/\n",
+                        "http://host2/base/$BASEARCH/os/\n",
+                        "http://host3/base/$ARCH/os/\n",
+                        ])
+                mirrors = cs._get_mirror_list(self.repo, "https://fake/repo/url")
+                self.assertEqual(mirrors, [])
+
+        # If mirrorlist contains invalid repos, check they are discarded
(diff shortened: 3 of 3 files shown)


vzhestkov:
@zzaimeche thanks for taking care of it.
Just additional suggestion. I've tried before to analyze some of the cases and found that it's possible that the code is trying to get the URLs from the HTML document returned on request for the root of the repo, but in this case it's useless to read the HTML and try to get URLs from it as mirrorlist supposed to be the plain text output if I'm not wrong. Maybe we need to check for the document format returned on request and avoid trying to get list of mirrors in case if it's HTML document?

zzaimeche:
This works fine on my system, but I'm a bit worried that someone might have a strange setup somewhere where they use urls that are forbidden in the RFC (and I left out " as a character, because that's such an edge case and I thought I was more likely to introduce a bug trying to escape it). I still need to write tests, but if users are using weird urls, that won't be caught by tests anyway...

zzaimeche:
> @zzaimeche thanks for taking care of it. Just additional suggestion. I've tried before to analyze some of the cases and found that it's possible that the code is trying to get the URLs from the HTML document returned on request for the root of the repo, but in this case it's useless to read the HTML and try to get URLs from it as mirrorlist supposed to be the plain text output if I'm not wrong. Maybe we need to check for the document format returned on request and avoid trying to get list of mirrors in case if it's HTML document?

From what I understand of the reposync code so far, it goes to the url given by the user, then uses urlgrabber to look through the page at that url for anything that starts with http, ftp, https (and so on), and then when it finds something that matches, it saves *that* to the local textfile (mirrorlist.txt). If there is nothing in the mirrorlist.txt at the end, reposync uses the repourl directly, otherwise it uses urls from that local mirrorlist.txt file. I don't think there's a structural difference between pages with valid lists and the ones without but I might have missed something (though even if people are meant to upload their lists as plain text, I wouldn't be surprised if they're all using different formats anyway).

zzaimeche:
> I think the proper mirror list should be something like this `http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=updates` 

Ah okay, I thought they might be also expecting html pages more like: https://www.debian.org/mirror/list  and making it stricter would break it for users pointing to something like that. (Though I haven't checked many urls yet)

vzhestkov:
@zzaimeche actually I'm not sure about it. And `apt` repos are handled with `deb_src` plugin. I don't remeber if there was any way to process the mirrors.
I think it's better to ask those who have more experience with different kinds of mirrors. Or make this mirror list parsing optional to prevent accidental selection of improper URLs from the page.

zzaimeche:
> @zzaimeche actually I'm not sure about it. And `apt` repos are handled with `deb_src` plugin. I don't remeber if there was any way to process the mirrors. I think it's better to ask those who have more experience with different kinds of mirrors. Or make this mirror list parsing optional to prevent accidental selection of improper URLs from the page.

Okay. I'll try to find out if yum mirrorlists are certain to be plaintext; at the moment I'm a bit worried that users might have their lists of urls up on all sorts of weird pages that just happen to work with the current code, even if they're not supposed to, so fixing it might break things for them in ways that are hard to check in advance. But the current code is really fragile so I'd rather check the document format if I can. Thanks. :)

zzaimeche:
Bit messy and am not thrilled at adding another import, but I couldn't see a way to use urlgrabber for it. I haven't been able to find out yet if only plaintext pages are mirrorlists, so I've kept the commits separate in case that does cause problems.

zzaimeche:
Ugh: https://fedoraproject.org/wiki/Infrastructure/MirrorManager#Metalink_format_mirrorlist

Apparently this is an example: http://ftp5.gwdg.de/pub/opensuse/discontinued/distribution/11.3/iso/ (Content-Type | text/html)

Those pages look really old, so I wondered if it was no longer used, but I found: https://serverfault.com/questions/1021273/all-yum-mirrors-failing-with-errno-1-repomd-xml-does-not-match-metalink which led to an example here: https://mirror.vcu.edu/pub/gnu_linux/epel/7/x86_64/

vzhestkov:
Not sure what you mean. Metalink is the different format which is not supported by reposync, if I'm not wrong.
But mirrorlist from the example `https://mirrors.fedoraproject.org/mirrorlist?repo=fedora-10&arch=x86_64` returns exact plain text list of URLs.

zzaimeche:
> Not sure what you mean. Metalink is the different format which is not supported by reposync, if I'm not wrong. But mirrorlist from the example `https://mirrors.fedoraproject.org/mirrorlist?repo=fedora-10&arch=x86_64` returns exact plain text list of URLs.

Oh, I thought if yum supported it then reposync was expected to support it because it's the yum_src plugin. If we don't have to then that's easier.

vzhestkov:
It could be confusing, but `yum_src` is using `zypper` internally. Reposync plugin does all of these checks for the mirror lists etc. before calling `zypper`.

zzaimeche:
Looks like the metalink support was explictly added: https://github.com/SUSE/spacewalk/commit/4e6f367f1f7b03956c7124e8a72f177a4a9d4ae3 . I think the code's currently structured so that it assumes a list is a mirrorlist, then tries to treat it as a metalink and if that fails it goes back to treating it as a mirrorlist. So I think I could just change my second commit so it checks 'if page plaintext or xml', and that way at least html pages would still be excluded.

vzhestkov:
Oh, sorry I haven't checked deeply to the case when it tries to parse XML. In general it seems that we need to cover only clear text case with mirrorlist. Thanks for pointing :+1:

zzaimeche:
> Oh, sorry I haven't checked deeply to the case when it tries to parse XML. In general it seems that we need to cover only clear text case with mirrorlist. Thanks for pointing +1

Yeah, I think the code is a bit confusingly named/commented because it talks about mirrorlists, but at the start I think it's also getting xml if that's what it finds. So it's called _get_mirror_list but it's also getting the xml stuff if it sees that (unless I'm very confused). I tried to test against a repo with a metalink set as the repourl (this one: https://mirrors.fedoraproject.org/metalink?repo=fedora-18&arch=x86_64 ) to make sure I hadn't broken the xml bit... and it didn't work anyway. =D Not sure if there's a setting I should change somewhere or if that code just never worked or if I just misunderstood it anyway. If it just never worked, I'll change it back to checking for plaintext only. :)

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -525,6 +526,13 @@ def _get_mirror_list(self, repo, url):
         mirrorlist_path = os.path.join(repo.root, 'mirrorlist.txt')
         returnlist = []
         content = []
+        # If page not plaintext or xml, is not a valid mirrorlist or metalink,
+        # so continue without it.
+        with request.urlopen(url) as webpage:
```
vzhestkov, line 531: It seems in this case it will ignore HTTP_PROXY settings set with web UI (rhn.conf).

python/spacewalk/spacewalk-backend.changes
```diff
@@ -1,3 +1,15 @@
+-------------------------------------------------------------------
```
vzhestkov, line 1: No need to have 2 distinct entries in the changelog. And no need to set the header for the entry, check with the other PRs for the formatting. For example: https://github.com/uyuni-project/uyuni/pull/6023/files

zzaimeche: ah okay, I followed some linked opensuse docs that said to use the output of `osc vc` so I used that, I'll try to update those too if I remember.

agraul: The "authoritative" source for the changelogs would be https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs (openSUSE changelog guidelines still apply)

zzaimeche:
I've marked it as a wip because I haven't had a chance to check if the proxy code does what I think it does, but I wanted to put it up before I forgot about it at the end of the week!

python/spacewalk/spacewalk-backend.changes
```diff
@@ -1,3 +1,9 @@
+- Exclude invalid mirror urls for reposync (bsc#1203826)
```
vzhestkov, line 1: The format of the entry is changing on releasing the package. There is no need to format it this way, just keep it as a one level entry.
Something like:
```
- Use only plain text as mirrorlist (bsc#1203826)
```
as XML is relevant for metalinks and this change is not affecting this part if I'm not wrong.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -548,16 +564,21 @@ def _get_mirror_list(self, repo, url):
...
         def _replace_and_check_url(url_list):
             goodurls = []
             skipped = None
             for url in url_list:
-                # obvious bogons get ignored b/c, we could get more interesting checks but <shrug>
+                # obvious bogons get ignored
                 if url in ['', None]:
                     continue
+                # Discard any urls containing some invalid characters
+                forbidden_characters = "<>^`{|}"
```
vzhestkov, line 578: Is it still required as we are not reading HTML as a list of URLs anymore?

vzhestkov (commented):
In general looks like it tries to parse the mirrorlist as a plaintext even if it was detected that it's xml. Am I wrong?

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
zzaimeche: I think they could still upload a plaintext list of urls with invalid entries containing forbidden characters, so now this checks for that. Probably not a common problem but I figured it couldn't hurt to keep the check.

vzhestkov: Do we have examples of repo URLs containing such characters?

zzaimeche:
> In general looks like it tries to parse the mirrorlist as a plaintext even if it was detected that it's xml. Am I wrong?

I don't think it does, because of when it runs _replace_and_check_url() ; this is after it has tried to treat the list as xml and got a list of mirror urls from that (but it looks confusing because that method is defined between the try-excepts, but called later, and I might have said the wrong thing earlier...) I think the bigger problem is the way the surrounding code is nested/structured; the control flow isn't obvious and it makes things hard to follow.

(Earlier, I tried to double-check behaviour with an url pointing to a metalink and that failed, but at a later point in the process; the part I'd changed seemed to work. So I'm not sure if the xml parsing currently works on master...)

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
zzaimeche: I've not seen any. (I based the list of forbidden characters on the response here: https://stackoverflow.com/a/36667242 . If we think there are people using urls that actually contain those characters somehow, I'll take it out again.)

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -525,6 +526,21 @@ def _get_mirror_list(self, repo, url):
...
         content = []
+        # If page not plaintext or xml, is not a valid mirrorlist or metalink,
+        # so continue without it.
+        proxies = get_proxies(self.proxy_url, self.proxy_user, self.proxy_pass)
+        cert = (self.sslclientcert, self.sslclientkey)
+        verify = self.sslcacert
+        try:
+            webpage = requests.get(url, proxies=proxies, cert=cert, verify=verify)
+            content_type = webpage.headers["Content-Type"]
+            if "text/plain" not in content_type and "xml" not in content_type:
```
vzhestkov, line 537: maybe change it to somethinkg like:
```
if webpage.headers["Content-Type"] not in ("text/plain", "xml"...)
```
Just a tuple with the exact possible values for `Content-Type`.

vzhestkov (changes requested):
In general it looks like this part of the code https://github.com/uyuni-project/uyuni/pull/6026/files#diff-a4f551c989af6141e94f0e9fcdfb392087152bed7218be989b28377c924812aaR597-R601 is used actually for both of `xml` and `text/plain` cases. Not sure if we really need to do it.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
zzaimeche: Oh, sorry, just checked the behaviour; I did it this way because the content_type is output in the form:
"text/plain; charset=UTF-8"
so it did need to be part of the output rather than the whole thing! I suppose it could split on the semicolon, but I think that would overcomplicate it, so unless there are times when it's going to include 'xml' or 'text/plain' but be the wrong type of document, I think it's better to leave it.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -525,6 +526,21 @@ def _get_mirror_list(self, repo, url):
...
+                # Not a valid mirrorlist or metalink; continue without it
+                return returnlist
+        except Exception as exc:
```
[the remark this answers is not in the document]
zzaimeche: I'm not sure if the preferred course is to remove it entirely so it fails, or to handle each exception specifically; if it's the latter then I'd rather do that in a followup patch because it may take some time to work out what needs to be there.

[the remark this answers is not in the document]
meaksh: It seems all exceptions raised by `requests` are inherited from `requests.exceptions.RequestException`, so we could use that one.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
meaksh: Hmmm, if we include this extra check, I would make some log/warning message to be displayed in case we are excluding an URL as invalid.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -31,6 +31,7 @@
 import lzma
 import os
 import re
+import requests
```
agraul, line 34: Is it needed to include `requests`? I'm asking because we already use `urlgrabber` in reposync and, while I'd like to get rid of it, I think it's better to stick to one http library if possible.

zzaimeche: Yeah, I was hoping to avoid importing anything but I couldn't find a tidy way to get the content-type headers with urlgrabber. It seemed like it was possible to _supply_ headers but not to retrieve them in a consistent format; I think you'd have to parse the page itself looking for something starting <!DOCTYPE> or <?xml> and then map it to types by hand, and that seemed fragile... So unless I missed something with urlgrabber I think requests is the less messy option.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
zzaimeche: Good point, I'll add something for that.

python/spacewalk/spacewalk-backend.changes
```diff
@@ -1,3 +1,7 @@
...
+
+------------------------------------------------------------------
+
```
meaksh, line 4: This is not needed, you just need to put your new entry on top of the existing ones:

```suggestion
```

meaksh (approved):
LGTM - thanks!

meaksh:
@zzaimeche since there is a conflict here in the `spacewalk-backend.changes` file, you would need to rebase latest changes on "master" branch here to your branch before we can merge this.

zzaimeche:
I'm going to dismiss Victor's review unless there are objections, since it wouldn't be fair to rerequest a review and I can't merge while changes are requested. He had suggested a change but the link to the relevant lines doesn't work for me now, so I can't find what it refers to, but it doesn't seem to be something I added. If there are no other objections I'll merge this.

PR #6034, opened 2022-10-14
Use more vague IPv6 prefix

Bischoff:
## What does this PR change?

Fix the test suite for new network prefixes.

I did not change lines like:
```
testsuite/features/upload_files/virtualization/vCenter.json:        "10.162.186.111": {
```
as such addresses are fake anyway.

## Links

Ports:
* 4.3: https://github.com/SUSE/spacewalk/pull/19253
* 4.2: https://github.com/SUSE/spacewalk/pull/19254

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -57,8 +57,8 @@ Then(/^the IPv6 address for "([^"]*)" should be correct$/) do |host|
 
   lines = interface.lines
   # selects only lines with IPv6 addresses and proceeds to form an array with only those addresses
-  ipv6_addresses_list = lines.grep(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/)
-  ipv6_addresses_list.map! { |ip_line| ip_line.slice(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/) }
+  ipv6_addresses_list = lines.grep(/2[:0-9a-f]*|fe80:[:0-9a-f]*/)
+  ipv6_addresses_list.map! { |ip_line| ip_line.slice(/2[:0-9a-f]*|fe80:[:0-9a-f]*/) }
 
   # confirms that the IPv6 address shown on the page is part of that list and, therefore, valid
   ipv6_address = find(:xpath, "//td[text()='IPv6 Address:']/following-sibling::td[1]").text


srbarrios (approved):

elariekerboull (approved):

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -57,8 +57,8 @@
 
   lines = interface.lines
   # selects only lines with IPv6 addresses and proceeds to form an array with only those addresses
-  ipv6_addresses_list = lines.grep(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/)
-  ipv6_addresses_list.map! { |ip_line| ip_line.slice(/2620:[:0-9a-f]*|fe80:[:0-9a-f]*/) }
+  ipv6_addresses_list = lines.grep(/2a07:[:0-9a-f]*|fe80:[:0-9a-f]*/)
```
srbarrios, line 60: Maybe... we can think about using a regex that works for any MAC Address?
`^([0-9A-F]{2}[:-]){5}([0-9A-F]{2})$`

Or we want to keep this filtered by concrete prepend?

ktsamis (approved):

nodeg (approved):

testsuite/features/step_definitions/common_steps.rb
Bischoff: We could indeed reduce the size of the prefix to digit `2` . This is the prefix for "global unicast" addresses. More exactly:
```
 ... (global unicast addresses) have a prefix of 2000::/3, (all the addresses that begin with binary 001).
```

Apart from that, it already works with any MAC address.

I'll do that change if prefix changes again (for example for the cloud).

srbarrios: But... why we need to use one that has a prefix?
I miss something I guess. We are just checking that the MAC address is correct on that step right?

Bischoff: I need to make sure I am controlling a real IPv6 address and not some random text.

We don't check a MAC address in that step, we check an IP address. It can include the MAC, or not, for example it can be loopback `::1` or a privacy address as in https://www.rfc-editor.org/rfc/rfc7217 .

Let's huddle.

Bischoff: After discussion, let's check only the `2` (knowing that, strictly speaking, it could also be a `3`).

srbarrios: Just to safe somewhere what discussed, we also discussed the possibility to use https://github.com/ruby/ipaddr which is part of the standard library, to validate the IPs by instantiating them.

Issue #6037, opened 2022-10-14
Unable to delete systems following upgrade to 22.10


ddholstad99:
### Problem description

attempt to delete system using web interface returns:
Server error, please check log files.

### Steps to reproduce

1.upgrade to uyuni 22.10
2.tail rhn_web_ui.log 
3.try to delete a system in the web interface. 
...

### Uyuni version

```bash
nformation for package Uyuni-Server-release:
---------------------------------------------
Repository     : Uyuni Server Stable
Name           : Uyuni-Server-release
Version        : 2022.10-220400.191.2.uyuni2
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes
Status         : up-to-date
Source package : Uyuni-Server-release-2022.10-220400.191.2.uyuni2.src
Summary        : Uyuni Server
Description    : 
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.
```

### Uyuni proxy version (if used)

```bash
no proxy in use
```

### Useful logs

```bash
rhn_web_ui.log gets the following error:
2022-10-14 16:06:26,609 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-9] ERROR com.redhat.rhn.manager.kickstart.cobbler.CobblerLoginCommand - XmlRpcFault while logging in.  most likely user doesn't have permissions. 
redstone.xmlrpc.XmlRpcFault: <class 'socket.gaierror'>:[Errno -2] Name or service not known
	at redstone.xmlrpc.XmlRpcClient.handleResponse(XmlRpcClient.java:444) ~[redstone-xmlrpc.jar:?]
	at redstone.xmlrpc.XmlRpcClient.endCall(XmlRpcClient.java:376) ~[redstone-xmlrpc.jar:?]
	at redstone.xmlrpc.XmlRpcClient.invoke(XmlRpcClient.java:165) ~[redstone-xmlrpc.jar:?]
	at com.redhat.rhn.manager.kickstart.cobbler.CobblerXMLRPCHelper.invokeMethod(CobblerXMLRPCHelper.java:70) ~[rhn.jar:?]
[...]

Issue #6045, opened 2022-10-18
Cobblerd stopped working on 2022.10


ZsBT:
### Problem description

systemd unit refuses to start, saying:

`ValueError: os_version for breed "suse" must be one of suse, opensuse11.2, opensuse11.3, opensuse11.4, opensuse12.1, opensuse12.2, opensuse12.3, opensuse13.1, opensuse13.2, opensuse15.0, opensuse15.1, opensuse15.2, opensuse15.3, opensuse15.4, sles10generic, sles11, sles11sp1, sles11sp2, sles11sp3, sles11sp4, sles11generic, sles12, sles12generic, sles12sp1, sles12sp2, sles12sp3, sles12sp4, sles12sp5, sles15generic, given was "virtio26"
`

### Steps to reproduce

1. upgrade Uyuni from 2022.08 to 2022.10
2. reboot
3. journalctl -u cobblerd
...

### Uyuni version

```bash
Loading repository data...
Reading installed packages...


Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : Uyuni Server Stable
Name           : Uyuni-Server-release
Version        : 2022.10-220400.191.2.uyuni2
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes
Status         : up-to-date
Source package : Uyuni-Server-release-2022.10-220400.191.2.uyuni2.src
Summary        : Uyuni Server
Description    : 
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.
```

### Useful logs

```bash
Oct 18 18:25:58 uyuni cobblerd[5570]: INFO | running: ['/usr/sbin/selinuxenabled']
Oct 18 18:25:58 uyuni cobblerd[5570]: INFO | received on stdout:
Oct 18 18:25:58 uyuni cobblerd[5570]: INFO | Automigration NOT executed
Oct 18 18:25:58 uyuni cobblerd[5570]: INFO | python3-hivex not found. If you need Automatic Windows Installation support, please install.
Oct 18 18:25:58 uyuni cobblerd[5570]: INFO | 11 breeds and 131 OS versions read from the signature file
Oct 18 18:25:58 uyuni cobblerd[5570]: Traceback (most recent call last):
[...]

Issue #6041, opened 2022-10-17
http://localhost/rpc/api stopped working with uyuni 2022.10


hbrown-uiowa:
We have processes that were connecting to http://localhost/rpc/api that started failing after upgrading from 2022.08 to 2022.10.

We are using this code to log in:
```
#!/usr/bin/ruby
require "xmlrpc/client"
@MANAGER_URL = "http://localhost/rpc/api"
@MANAGER_LOGIN = "<username>"
@MANAGER_PASSWORD = "<password>"
@client = XMLRPC::Client.new2(@MANAGER_URL)
@key = @client.call('auth.login', @MANAGER_LOGIN, @MANAGER_PASSWORD)
users = @client.call('user.listUsers', @key)

p @key
```

On a 2022.08 system I run it and get:
```
"7230321xcd589904a4aa05bbdc3894afd0673c3ad566d4f8506e4de68be9ee7376f55d64"
```

When I run the same code against our test instance on 2022.10, I get:
```
Traceback (most recent call last):
        3: from ./foo.rb:7:in `<main>'
        2: from /usr/lib64/ruby/gems/2.5.0/gems/xmlrpc-0.3.0/lib/xmlrpc/client.rb:268:in `call'
        1: from /usr/lib64/ruby/gems/2.5.0/gems/xmlrpc-0.3.0/lib/xmlrpc/client.rb:287:in `call2'
/usr/lib64/ruby/gems/2.5.0/gems/xmlrpc-0.3.0/lib/xmlrpc/client.rb:509:in `do_rpc': HTTP-Error: 404 Not Found (RuntimeError)

```

apache is listening on port 80 on 2022.10.

Fredouye:
Hi

same error when using spacewalk-common-channels : 

```
# spacewalk-common-channels -u admin -a x86_64 rockylinux9 rockylinux9-extras rockylinux9-uyuni-client rockylinux9-appstream -s uyuni
.home.lab
SUSE Manager password: 
Traceback (most recent call last):
  File "/usr/bin/spacewalk-common-channels", line 278, in <module>
    client, key = connect(options.user, options.password, options.server)
  File "/usr/bin/spacewalk-common-channels", line 120, in connect
    key = xmlrpc_login(client, options.user, password)
  File "/usr/lib/python3.6/site-packages/uyuni/common/cli.py", line 74, in xmlrpc_login
    sessionkey = client.auth.login(username, password)
[...]

PR #6050, opened 2022-10-19
Uyuni 2022.10: Fix issues triggering Cobbler migrations

meaksh:
## What does this PR change?

This PR fixes some issues detected in Uyuni 2022.10 around Cobbler, that makes Cobbler service to break causing different symthoms in the Uyuni server:

- https://github.com/uyuni-project/uyuni/issues/6037
- https://github.com/uyuni-project/uyuni/issues/6041
- https://github.com/uyuni-project/uyuni/issues/6045

These different issues were caused because the migrations of Cobbler settings and collections to newer Cobbler 3.3.3, triggered by `spacewalk-setup` package during the upgrade, didn't happen as it was expected. This makes the Cobbler service not able to start properly as it still has non-migrated settings and collections.

With the changes in this PR, the new update of `spacewalk-setup` will trigger the execution of the migration as expected, making the Cobbler service able to start successfully.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **tested manually**

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/6037 https://github.com/uyuni-project/uyuni/issues/6041 https://github.com/uyuni-project/uyuni/issues/6045
Tracks https://github.com/SUSE/spacewalk/issues/19276

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacewalk/setup/bin/spacewalk-setup-cobbler
+++ b/spacewalk/setup/bin/spacewalk-setup-cobbler
@@ -67,6 +67,10 @@ def manipulate_cobbler_settings(config_dir: str, settings_yaml: str):
     except OSError:
         filecontent["next_server_v6"] = ""
 
+    if not filecontent["next_server_v4"] and not filecontent["next_server_v6"]:
+        print("ERROR: Neither IPv4 or IPv6 addresses can be resolved for configured FQDN: {}. Please check your DNS and hostname configuration.".format(filecontent["server"]))
+        exit(1)
+
     filecontent["pxe_just_once"] = True
     filecontent["redhat_management_server"] = socket.getfqdn()
     yaml_dump = yaml.safe_dump(filecontent)
--- a/spacewalk/setup/bin/spacewalk-setup-cobbler
+++ b/spacewalk/setup/bin/spacewalk-setup-cobbler
@@ -68,7 +68,7 @@ def manipulate_cobbler_settings(config_dir: str, settings_yaml: str):
         filecontent["next_server_v6"] = ""
 
     if not filecontent["next_server_v4"] and not filecontent["next_server_v6"]:
-        print("ERROR: Neither IPv4 or IPv6 addresses can be resolved for configured FQDN: {}. Please check your DNS and hostname configuration.".format(filecontent["server"]))
+        print("ERROR: Neither IPv4 nor IPv6 addresses can be resolved for configured FQDN: {}. Please check your DNS and hostname configuration.".format(filecontent["server"]))
         exit(1)
 
     filecontent["pxe_just_once"] = True
--- a/spacewalk/setup/bin/spacewalk-setup-cobbler
+++ b/spacewalk/setup/bin/spacewalk-setup-cobbler
@@ -53,10 +53,15 @@ def manipulate_cobbler_settings(config_dir: str, settings_yaml: str):
     backup_file(full_path)
     with open(full_path) as settings_file:
         filecontent = yaml.safe_load(settings_file.read())
-    filecontent["server"] = socket.getfqdn()
-    filecontent["next_server_v4"] = socket.getaddrinfo(filecontent["server"], None, socket.AF_INET)[1][4][0]
 
-    # In case there is no IPv6 address, we get a OSError (socket.gaierror)
+    filecontent["server"] = socket.getfqdn()
+
+    # In case of failing DNS resolution, we get a OSError (socket.gaierror)
+    try:
+        filecontent["next_server_v4"] = socket.getaddrinfo(filecontent["server"], None, socket.AF_INET)[1][4][0]
+    except OSError:
+        filecontent["next_server_v4"] = ""
+
     try:
         filecontent["next_server_v6"] = socket.getaddrinfo(filecontent["server"], None, socket.AF_INET6)[1][4][0]
     except OSError:
--- a/spacewalk/setup/spacewalk-setup.spec
+++ b/spacewalk/setup/spacewalk-setup.spec
@@ -259,6 +259,8 @@ if [ ! -f /etc/cobbler/settings -a -f /etc/cobbler/settings.rpmsave ]; then
     cobbler-settings automigrate -d
     echo "* Readjust settings needed for spacewalk"
     spacewalk-setup-cobbler
+    echo "* Change group to Apache for /etc/cobbler/settings.yaml file"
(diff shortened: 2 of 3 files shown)


spacewalk/setup/spacewalk-setup.spec
```diff
@@ -244,9 +244,13 @@ if grep 'authn_spacewalk' /etc/cobbler/modules.conf > /dev/null 2>&1; then
...
 fi
 
-# if old /etc/cobbler/settings exists, we need to perform migration
-if [ -e /etc/cobbler/settings ]; then
-    echo "* Creating a backup of /etc/cobbler/settings file to /etc/cobbler/settings.before-migration-backup before migrating settings"
+# When upgrading to Cobbler 3.3.3, the old /etc/cobbler/settings config file from previous Cobbler version
+# is removed as it not existing anymore in the new version, but a copy is kept with the local changes done
+# at /etc/cobbler/settings.rpmsave. If this file exists, it means we need to perform the migration of these
+# settings and also trigger the migration of stored Cobbler collections.
+if [ ! -f /etc/cobbler/settings -a -f /etc/cobbler/settings.rpmsave ]; then
```
juliogonzalez, line 251: Since we are going to release this as a patch on top of 2022.10, are you sure `/etc/cobbler/settings` will never exist?

spacewalk/setup/bin/spacewalk-setup-cobbler
```diff
@@ -53,10 +53,15 @@ def manipulate_cobbler_settings(config_dir: str, settings_yaml: str):
...
+
     filecontent["server"] = socket.getfqdn()
-    filecontent["next_server_v4"] = socket.getaddrinfo(filecontent["server"], None, socket.AF_INET)[1][4][0]
 
-    # In case there is no IPv6 address, we get a OSError (socket.gaierror)
+    # In case of failing DNS resolution, we get a OSError (socket.gaierror)
+    try:
+        filecontent["next_server_v4"] = socket.getaddrinfo(filecontent["server"], None, socket.AF_INET)[1][4][0]
+    except OSError:
+        filecontent["next_server_v4"] = ""
```
agraul, line 63: small question, what happens when "next_server_v4" and "next_server_v6" are both empty? Is any functionality impacted?

spacewalk/setup/spacewalk-setup.spec
meaksh: As soon as Cobbler 3.3.3 is installed, the file should be gone.

If `/etc/cobbler/settings` exists prior to Cobbler 3.3.3 upgrade, then it will be renamed to `settings.rpmsave` during the Cobbler upgrade before reaching this point, as we are making this package to `Requires(post)`.

If Cobbler 3.3.3 is already installed, so current users with Uyuni 2022.10 already installed, the `/etc/cobbler/settings` file should not be existing and already renamed to `settings.rpmsave` during the upgrade to Uyuni 2022.10.

spacewalk/setup/bin/spacewalk-setup-cobbler
SchoolGuy: Without the next server Cobbler is not functional. No auto-installations can work. An IP that the clients can resolve must be present.

meaksh: These are new settings in Cobbler 3.3.3, after looking at Cobbler code I see both of them are accepted as empty values.

In principle, as having DNS resolution working is a premise, you should get at least IPv4 address resolution. I added here the "try/except" because while debugging, using wrong hostname in "server", I noticed this would make the execution of `spacewalk-setup-cobbler` to crash ending up in a more inconsistent state.

agraul: If I understand @SchoolGuy correctly, it doesn't make sense to continue with the script if the "server" hostname is not resolvable.

IMO we should check if we have at least one "next server" IP before this function overrides the old settings. Silently creating a broken (as in "not working") config is more inconsistent than bailing out.

SchoolGuy: Well if we could raise the problem to the UI that we have a broken config, all daemons would start and we could use the Cobbler API to fix that during runtime. However that is sadly not possible afaik and thus I agree with @agraul that we should bail out and print a clear message with the reasoning for bailing out.

spacewalk/setup/spacewalk-setup.spec
juliogonzalez: ACK, looks good then.

juliogonzalez (approved):

spacewalk/setup/bin/spacewalk-setup-cobbler
meaksh: Ok, I have made the script to crash in case we cannot resolve any IP addresses for the fqdn.

Also, in case any exception during the migration, the error is shown and execution will stop. Zypper displays a warning meesage and will not exit cleanly. Making the user hopefully to notice that something when wrong.

Triggering reinstallation of `spacewalk-setup` package should make the Cobbler migration to run again.

spacewalk/setup/bin/spacewalk-setup-cobbler
```diff
@@ -53,15 +53,24 @@ def manipulate_cobbler_settings(config_dir: str, settings_yaml: str):
...
+
     try:
         filecontent["next_server_v6"] = socket.getaddrinfo(filecontent["server"], None, socket.AF_INET6)[1][4][0]
     except OSError:
         filecontent["next_server_v6"] = ""
 
+    if not filecontent["next_server_v4"] and not filecontent["next_server_v6"]:
+        print("ERROR: Neither IPv4 or IPv6 addresses can be resolved for configured FQDN: {}. Please check your DNS and hostname configuration.".format(filecontent["server"]))
```
agraul, line 71: nitpick
```suggestion
        print("ERROR: Neither IPv4 nor IPv6 addresses can be resolved for configured FQDN: {}. Please check your DNS and hostname configuration.".format(filecontent["server"]))
```

agraul (approved):

SchoolGuy (approved):
LGTM

spacewalk/setup/bin/spacewalk-setup-cobbler
meaksh: thanks!

PR #6053, opened 2022-10-19
Restoring scenario initiating Rocky8 sync

elariekerboull:
## What does this PR change?

Restore Rocky 8 channel sync scenario


srbarrios (commented):
Could you explain why we need to revert this scenario in the description please?

nodeg:
Hm, if we still need this I need to find a way to sync this custom channel via the API. `mgr-sync list channels` does not show the custom channel.

elariekerboull:
Not necessary anymore.

PR #6076, opened 2022-10-26
QE: Add tests for distribution channel mapping

ByteOtter:
## What does this PR change?

Add automated tests for the Distribution Channel Mapping page

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Cucumber tests were added

- [x] **DONE**

## Links

Partially Fixes https://github.com/SUSE/spacewalk/issues/10287

Manager-4.2

Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -528,15 +528,6 @@ And(/^I select "(.*?)" in the dropdown list of the architecture filter$/) do |ar
   raise "Architecture #{architecture} not found" unless find(:xpath, "//div[@id='select2-drop']/ul/li/div[contains(text(), '#{architecture}')]").click
 end
 
-# select an item from any dropdown
-When(/^I select "(.*?)" from "([^"]*)" dropdown/) do |selection, label|
-  # let the the select2js box filter open the hidden options
-  xpath_query = "//select[@name='#{label}']"
-  raise "xpath: #{xpath_query} not found" unless find(:xpath, xpath_query).click
-  # select the desired option
-  raise "#{label} #{selection} not found" unless find(:xpath, "//select[@name='#{label}']/option[contains(text(), '#{selection}')]").click
-end
-
 When(/^I (deselect|select) "([^\"]*)" as a product$/) do |select, product|
   # click on the checkbox to select the product
   xpath = "//span[contains(text(), '#{product}')]/ancestor::div[contains(@class, 'product-details-wrapper')]/div/input[@type='checkbox']"
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -528,6 +528,15 @@ And(/^I select "(.*?)" in the dropdown list of the architecture filter$/) do |ar
   raise "Architecture #{architecture} not found" unless find(:xpath, "//div[@id='select2-drop']/ul/li/div[contains(text(), '#{architecture}')]").click
 end
 
+# select an item from any dropdown
+When(/^I select "(.*?)" from "([^"]*)" dropdown/) do |selection, label|
+  # let the the select2js box filter open the hidden options
+  xpath_query = "//select[@name='#{label}']"
+  raise "xpath: #{xpath_query} not found" unless find(:xpath, xpath_query).click
+  # select the desired option
+  raise "#{label} #{selection} not found" unless find(:xpath, "//select[@name='#{label}']/option[contains(text(), '#{selection}')]").click
+end
+
 When(/^I (deselect|select) "([^\"]*)" as a product$/) do |select, product|
   # click on the checkbox to select the product
   xpath = "//span[contains(text(), '#{product}')]/ancestor::div[contains(@class, 'product-details-wrapper')]/div/input[@type='checkbox']"
@@ -1452,4 +1461,8 @@ end
 When(/^I clean up the server's hosts file$/) do
   command = "sed -i '$d' /etc/hosts && sed -i '$d' /etc/hosts"
   $server.run(command)
+end
+
+Then(/^I should see the text "(.*?)" in the (Operating System|Architecture|Channel Label) field/) do |text, field|
+  page.has_field?("#{text}", with: field)
 end
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -1453,7 +1453,3 @@ When(/^I clean up the server's hosts file$/) do
   command = "sed -i '$d' /etc/hosts && sed -i '$d' /etc/hosts"
   $server.run(command)
 end
-
-Then(/^I should see the text "(.*?)" in the (Operating System|Architecture|Channel Label) field/) do |text, field|
-  page.has_field?("#{text}", with: field)
-end
--- a/testsuite/run_sets/secondary_parallelizable.yml
+++ b/testsuite/run_sets/secondary_parallelizable.yml
@@ -37,6 +37,7 @@
 - features/secondary/srv_payg_ssh_connection.feature
 - features/secondary/srv_push_package.feature
 - features/secondary/srv_reportdb.feature
+- features/secondary/srv_dist_channel_mapping.feature
 
 - features/secondary/proxy_retail_pxeboot_and_mass_import.feature
 - features/secondary/allcli_overview_systems_details.feature
--- a/testsuite/features/secondary/srv_dist_channel_mapping.feature
+++ b/testsuite/features/secondary/srv_dist_channel_mapping.feature
@@ -1,0 +1,128 @@
+# Copyright (c) 2022 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Distribution Channel Mapping
+
+  Scenario: Log in as admin user
+    Given I am authorized for the "Admin" section
+
+  Scenario: Check if Distribution Channel Mapping page exists
+    When I follow the left menu "Software > Distribution Channel Mapping"
+    Then I should see a "Distribution Channel Mapping" text
+    And I should see a "Channel List" link in the left menu
+    And I should see a "Package Search" link in the left menu
+    And I should see a "Manage" link in the left menu
+    And I should see a "Distribution Channel Mapping" link in the left menu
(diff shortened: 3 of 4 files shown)


testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1448,3 +1448,26 @@ def server_secret
   command = "sed -i '$d' /etc/hosts && sed -i '$d' /etc/hosts"
   $server.run(command)
 end
+
+# select an item from any dropdown
+And(/^I select "(.*?)" from "([^"]*)" dropdown/) do |selection, label|
```
elariekerboull, line 1453: Wondering, can't https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/step_definitions/navigation_steps.rb#L181 be used instead of this step?

ByteOtter: I was wondering aswell and tried to use this definition. Without success sadly. *If* I remember correcty, it gave me a Capybara Element not Found.

testsuite/features/secondary/srv_check_dist_channel_mapping_page.feature
```diff
@@ -0,0 +1,133 @@
...
+        Then I should see a "Distribution Channel Mapping" text
+        And I should see a "Channel List" link in the left menu
+        And I should see a "Package Search" link in the left menu
+        And I should see a "Manage" link in the left menu
+        And I should see a "Distribution Channel Mapping" link in the left menu
+        And I should see a "Create Distribution Channel Mapping" link
+        And I should see a "No distribution channel mappings currently exist." text in the content area
+
+    Scenario: Create new map for x86_64 suse clients
+        When I am on the "Distribution Channel Mapping" page
```
elariekerboull, line 20: I think it probably is better to repeat `When I follow the left menu "Software > Distribution Channel Mapping"` every time 
If the previous scenario fails and the test did not redirect to this page, the scenario will automatically fail.

elariekerboull: Which would make the definition of this associated test not necessary

ByteOtter: You mean have `When I follow the left menu "Software > Distribution Channel Mapping"` in every scenario?

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1448,3 +1448,26 @@ def server_secret
...
+  # let the the select2js box filter open the hidden options
+  xpath_query = "//select[@name='#{label}']"
+  raise "xpath: #{xpath_query} not found" unless find(:xpath, xpath_query).click
+  # select the desired option
+  raise "#{label} #{selection} not found" unless find(:xpath, "//select[@name='#{label}']/option[contains(text(), '#{selection}')]").click
+end
+
+# check if a certain link is present in the channel mappings table
+Given(/^I see a "(.*?)" link in the table/) do |link_name|
```
elariekerboull, line 1462: Also wondering for that one and https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/step_definitions/navigation_steps.rb#L616
Maybe there is a reason?

elariekerboull (commented):
Few questions/suggestions to discuss
Also you might have used tabs instead of spaces in the feature file

testsuite/features/step_definitions/common_steps.rb
ByteOtter: If I just look for a link it can't find it. I have to look for a specific link within a table field apparently.

testsuite/features/step_definitions/common_steps.rb
elariekerboull: Maybe the element is not visible without scrolling, that could cause the step to fail, not sure
If that's the case, that would be a good fix.

testsuite/features/step_definitions/common_steps.rb
elariekerboull: There also is https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/step_definitions/navigation_steps.rb#L655
However I guess we can keep it that way, we might have a number of steps that we reinvent and we would have to do some cleanup at some point

srbarrios:
Hi! Nice that we cover this feature! Thank you!

I'm currently reviewing it from my phone, it looks good, only one important note!
The way we work with our Cucumber features is by parsing and looping yaml files that contain the list of features (we do that through our Rakefile tasks), then in order to run this feature, you need to include it in the list of the yaml file secondary_parallelizable.yml file (if that test has not any step that could collide with other tests, like adding/removing a client).

I promise a better look on Monday :)
I'm very happy to have you to fight together in that jungle 💪

testsuite/features/secondary/srv_check_dist_channel_mapping_page.feature
```diff
@@ -0,0 +1,133 @@
...
+    Then I should not see a "SUSE Linux Enterprise Server 15 SP 4 iSeries modified" link
+
+  Scenario: Sanity check whether the page is in its default state
+    When I follow the left menu "Software > Distribution Channel Mapping"
+    Then I should see a "Distribution Channel Mapping" text
+    And I should see a "Channel List" link in the left menu
+    And I should see a "Package Search" link in the left menu
+    And I should see a "Manage" link in the left menu
+    And I should see a "Distribution Channel Mapping" link in the left menu
+    And I should see a "Create Distribution Channel Mapping" link
```
nodeg, line 132:
```suggestion
```
FMPOV this can be skipped since you already checked that in the very first scenario. Here you only want to see that no distribution channel mapping exists after you deleted all.

nodeg (commented):
For you first feature not bad :wink: As Oscar mentioned you have to add your feature to the following file: https://github.com/uyuni-project/uyuni/blob/master/testsuite/run_sets/secondary_parallelizable.yml

Bischoff (changes requested):
see nitpicks

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1448,3 +1448,26 @@ def server_secret
...
+  page.has_field?("#{link_name}", with: 'Operating System')
+end
+
+Then(/^I should see a "(.*?)" link in the table/) do |link_name|
```
vandabarata, line 1466: Do we need 2 of the exact same step? Honest question, even if they use different starters

vandabarata (approved):

nodeg (approved):

testsuite/run_sets/secondary_parallelizable.yml
```diff
@@ -37,6 +37,7 @@
 - features/secondary/srv_payg_ssh_connection.feature
 - features/secondary/srv_push_package.feature
 - features/secondary/srv_reportdb.feature
+- features/secondary/srv_check_dist_channel_mapping_page.feature
```
srbarrios, line 40: We can remove some words from the feature file :)

```suggestion
- features/secondary/srv_dist_channel_mapping.feature
```

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1448,3 +1448,22 @@ def server_secret
...
 end
+
+# select an item from any dropdown
+When(/^I select "(.*?)" from "([^"]*)" dropdown/) do |selection, label|
+  # let the the select2js box filter open the hidden options
+  xpath_query = "//select[@name='#{label}']"
+  raise "xpath: #{xpath_query} not found" unless find(:xpath, xpath_query).click
+  # select the desired option
+  raise "#{label} #{selection} not found" unless find(:xpath, "//select[@name='#{label}']/option[contains(text(), '#{selection}')]").click
+end
```
srbarrios, line 1459: I suggest to move this step to navigation_steps or either close to other similar steps in this file (around line 520ish)

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1448,3 +1448,22 @@ def server_secret
...
+
+# check if a certain link is present in the channel mappings table
+Then(/^I should see a "(.*?)" link in the table/) do |link_name|
+  page.has_field?("#{link_name}", with: 'Operating System')
+end
+
+# check if a certain architecture text is present in the mappings table
+Then(/^I should see a "(.*?)" architecture description in the table/) do |text|
+  page.has_field?("#{text}", with: 'Architecture')
+end
```
srbarrios, line 1469:
```suggestion
Then(/^I should see the text "(.*?)" in (Operating System|Architecture) field/) do |text, field|
  page.has_field?("#{text}", with: '#{field}')
end
```

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -1448,3 +1457,7 @@ def server_secret
   command = "sed -i '$d' /etc/hosts && sed -i '$d' /etc/hosts"
   $server.run(command)
 end
+
+Then(/^I should see the text "(.*?)" in the (Operating System|Architecture|Channel Label) field/) do |text, field|
+  page.has_field?("#{text}", with: field)
+end
```
srbarrios, line 1463: New line at the end :)

testsuite/features/step_definitions/common_steps.rb
```diff
@@ -523,6 +523,15 @@
   raise "Architecture #{architecture} not found" unless find(:xpath, "//div[@id='select2-drop']/ul/li/div[contains(text(), '#{architecture}')]").click
 end
 
+# select an item from any dropdown
+When(/^I select "(.*?)" from "([^"]*)" dropdown/) do |selection, label|
```
Bischoff, line 527: question:
```
When(/^I select "([^"]*)" from "([^"]*)"$/) do |option, field|
```
(in same file) was not appropriate?

testsuite/features/step_definitions/common_steps.rb
Bischoff: Was about to say it :smile_cat:

Bischoff (commented):
one question + same remark as Oscar

Bischoff (approved):
lgtm

PR #6082, opened 2022-10-29
Remove Python 2 compatibilty code for spacecmd

### LINKED ISSUE #6005: spacecmd: Remove Python 2 compatibilty code
`spacecmd` is only built for Python 3 systems, that means we don't need to keep code that specifically provides backwards-compatibility with Python 2.

A few examples of Python 2 compatibility code:
```python
### Imports
try:
    from xmlrpc import client as xmlrpclib
except ImportError:
    import xmlrpclib
try:
    import json
except ImportError:
    import simplejson as json  # python < 2.6

### Not using super()
class UnknownCallException(Exception):

    def __init__(self):
        Exception.__init__(self)
```
`spacecmd` does not contain a lof of Python 2 compatibility code, all of it can be done in a single PR.

iremugurlu:
## What does this PR change?

Removes Python 2 compatibility code for spacecmd

## Links

Fixes #6005 

## Changelogs

- [x] No changelog needed

### DIFF
--- a/spacecmd/src/bin/spacecmd
+++ b/spacecmd/src/bin/spacecmd
@@ -35,10 +35,7 @@ except ImportError:
 import codecs
 import locale
 import argparse
-try: # python 3
-    from configparser import SafeConfigParser
-except ImportError: # python 2
-    from ConfigParser import SafeConfigParser
+from configparser import SafeConfigParser
 import socket
 
 translation = gettext.translation('spacecmd', fallback=True)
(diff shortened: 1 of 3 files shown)


agraul:
Thank you a lot for the PR @iremugurlu! The changes you have done look good!

As I wrote in https://github.com/uyuni-project/uyuni/issues/6005 it would be great to have all these changes in a single PR, could you take a look at inheritance that does not use `super()` as well? The example in the issue should help you understand what I mean, but if anything is unclear please leave a question as a comment!

PR #6083, opened 2022-10-31
Improves schedule product migration in spacecmd

wweellddeerr:
## What does this PR change?

It changes `spacecmd` to use the correct API method when scheduling product migrations. Instead of using `scheduleProductMigration`, it was using `scheduleSPMigration` that is a deprecated method.

This PR also adds `--allow-vendor-change` and `--remove-products-without-successor` parameters to schedule product migration in `spacecmd`.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Documentation issue was created: [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/1859).

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/19208

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/src/spacecmd/system.py
+++ b/spacecmd/src/spacecmd/system.py
@@ -4607,6 +4607,8 @@ def help_system_scheduleproductmigration(self):
 \n    Options: \
 \n        -s START_TIME \
 \n        -d pass this flag, if you want to do a dry run \
+\n        --allow-vendor-change pass this flag if you want to allow vendor change \
+\n        -r pass this flag if you want to remove remove products which have no successors \
 \n        -c CHILD_CHANNELS (comma-separated child channels labels (with no spaces))'))
     print('')
     print(self.HELP_SYSTEM_OPTS)
@@ -4617,6 +4619,8 @@ def do_system_scheduleproductmigration(self, args):
     arg_parser.add_argument('-s', '--start-time')
     arg_parser.add_argument('-d', '--dry-run', action='store_true', default=False)
     arg_parser.add_argument('-c', '--child-channels')
+    arg_parser.add_argument('--allow-vendor-change', action='store_true', default=False)
+    arg_parser.add_argument('-r', '--remove-products-without-successor', action='store_true', default=False)
 
     (args, options) = parse_command_arguments(args, arg_parser)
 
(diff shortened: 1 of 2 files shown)


spacecmd/src/spacecmd/system.py
```diff
@@ -4606,6 +4606,8 @@ def help_system_scheduleproductmigration(self):
 \n    Options: \
 \n        -s START_TIME \
 \n        -d pass this flag, if you want to do a dry run \
+\n        -v pass this flag if you want to allow vendor change \
```
agraul, line 4609: `-v` is not always used consistently in `spacecmd` as a short-hand for `--verbose` but quite often that's now it is used. Many other tools also use `-v` for the same purpose. I would prefer if we could name the option for allowing vendor change differently.

wweellddeerr: Thank you for the review, do you think something like `-avc` would be better than `-v`? For this parameter I have a hard time thinking of a single letter that makes sense other than `v`.

When I added this shortcut, I didn't care too much about these possible overlaps because we've already used `-d` as a shortcut to `--dry-run` for example and it's typically used to enable debug mode.

agraul: I'm not sure if `-avc` is possible, it might be interpreted as `-a -v -c`. For me, there just is no single letter that makes sense. In `zypper` there is no short flag either. Do we need to have a single letter flag?

> When I added this shortcut, I didn't care too much about these possible overlaps because we've already used `-d` as a shortcut to `--dry-run` for example and it's typically used to enable debug mode.

`-d` is also used inconsistently in `spacecmd`, that's true. We can't easily (or quickly) change that, but we can pay more attention when we add new `-d`/`-v`flags.

spacecmd/src/spacecmd/system.py
```diff
@@ -4616,6 +4618,8 @@ def do_system_scheduleproductmigration(self, args):
     arg_parser.add_argument('-s', '--start-time')
     arg_parser.add_argument('-d', '--dry-run', action='store_true', default=False)
     arg_parser.add_argument('-c', '--child-channels')
+    arg_parser.add_argument('--allow-vendor-change', action='store_true', default=False)
```
parlt91, line 4621: I am not 100% sure how the interaction between spacecmd and our java code works exactly. But could you try if using all the options except `allow-vendor-change` also schedules the SP migration successful?

wweellddeerr: Thank you @parlt91 for the review. I checked it and it works as the default for the parameter is false.

PR #6144, opened 2022-11-14
Do not execute actions on minion that are not in pending status

mackdk:
## What does this PR change?

This PR make sure that an action is executed only on minions having a pending entry in the `ServerAction` table. Currently, if the user tries to reschedule the action on all the failed servers (through the web ui or the api), the taskomatic job actually executes the action on all of them. This is because the target minions were extracted without checking their action status.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/19445
Tracks https://github.com/SUSE/spacewalk/pull/19529 https://github.com/SUSE/spacewalk/pull/19528

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/ActionFactory.java
+++ b/java/code/src/com/redhat/rhn/domain/action/ActionFactory.java
@@ -1043,6 +1043,12 @@ public class ActionFactory extends HibernateFactory {
             lookupActionStatusByName("Failed");
 
     /**
+     * All the possible action statuses
+     */
+    public static final List<ActionStatus> ALL_STATUSES = List.of(STATUS_QUEUED, STATUS_PICKED_UP, STATUS_COMPLETED,
+        STATUS_FAILED);
+
+    /**
      * The constant representing Package Refresh List action.  [ID:1]
      */
     public static final ActionType TYPE_PACKAGES_REFRESH_LIST =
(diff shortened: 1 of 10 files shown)


java/code/src/com/redhat/rhn/domain/action/Action.java
```diff
@@ -262,13 +264,28 @@ public void setServerActions(Set serverActionsIn) {
         this.serverActions = serverActionsIn;
     }
 
+    /**
+     * Retrieve the server action for the specified server, if exists
+     * @param serverId the id of the server
+     * @return the {@link ServerAction} for the specified server
+     */
+    public Optional<ServerAction> getServerActionForServer(Long serverId) {
```
mcalmer, line 272: Seems it can have multiple actions for the requested server. You use "findFirst()". Wouldn't it be better to rename the function and call it "getFirstServerActionForServer" or return a List of all server actions.

mackdk: > Seems it can have multiple actions for the requested server

How is it possible? There is a unique constraint in `rhnserveraction` on (`server_id`, `action_id`)

```
rhn_server_action_sid_aid_uq" UNIQUE CONSTRAINT, btree (server_id, action_id)
```

Anyway, this method is a leftover of some testing I did, and it is not needed. I will remove it.

mcalmer: Was this really limited to 1 specific action id ? If yes, than it is unique.
But anyway, not it is gone :-)

mcalmer (approved):

lucidd (approved):

PR #6147, opened 2022-11-14
Add vendor_advisory to errata.getDetails

parlt91:
## What does this PR change?

Add the `Vendor Advisory` url to the errata.getDetails API call

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- No documentation needed: **Apidoc adjusted**

- [x] **DONE**

## Test coverage

- No tests: **bugfix**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/19487

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


parlt91:
Seems I need to adapt some tests

java/code/src/com/redhat/rhn/frontend/xmlrpc/errata/ErrataHandler.java
```diff
@@ -152,6 +156,15 @@ public Map<String, Object> getDetails(User loggedInUser, String advisoryName) th
...
             errataMap.put("advisory_status", errata.getAdvisoryStatus().getMetadataValue());
         }
+
+        try {
+            final VendorSpecificErrataParser parser = ErrataParserFactory.getParser(errata);
+            errataMap.put("vendor_advisory", parser.getAdvisoryUri(errata).toString());
+        }
+        catch (ErrataParsingException ex) {
+            throw new InvalidErrataException("Unable to parse errata vendor");
+        }
```
cbbayburt, line 166: I don't know how this is possible, but maybe we could still continue to output without this data? Is this case expected occasionally?

spacecmd/spacecmd.changes
```diff
@@ -1,3 +1,4 @@
+- Add vendor_advisory to errata.getDetails (bsc#1205207)
```
agraul, line 1: `spacecmd` users might not know what `errata.getDetails` means. I'd use the `spacecmd` command(s?) in the changelog that now show this information.

agraul (changes requested):
I only reviewed the Python parts. The code is fine, but I think we could improve the changelog a bit. Right now it's not really clear for users.

java/code/src/com/redhat/rhn/frontend/xmlrpc/errata/ErrataHandler.java
parlt91: Alternatively we could just `errataMap.put("vendor_advisory", "");`
Spacecmd would then print
```
Vendor Advisory
---------------
N/A
```
in this case

parlt91: I just double checked how we handle it in the webui: We catch the exception and log a info message. So it seems this could potentially happen and isn't really handled as an error.

cbbayburt: Or I'd say even `debug` level suits better. I don't think any logging is necessary unless we're debugging something about it.

spacecmd/spacecmd.changes
parlt91: Thanks. I made some adjustments and hope it's clearer now

agraul (approved):

PR #6226, opened 2022-11-28
Fix CVE Audit incorrectly displaying predecessor product

parlt91:
## What does this PR change?

Fix a case where CVE Audit would incorrectly advertise a predecessor product as successor product / migration target

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: **TODO**

- [ ] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/19632

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


parlt91:
@mcalmer I need your expertise on this one.
Let me sum up what the problem is:
Customer ran into a case where a previous (predecessor) product was advertised as a migration target with a fix for a CVE, but the system isn't even affected by said CVE (only the predecessor was).
See linked bug description for details

Now I found out that how we check if a system is affected by a CVE is to look at the `suseCveServerChannel` table and check if one of the associated channels has an errata which addresses the CVE.
But that means that we can't distinguish between:
  1) Systems that are not affected (but have a predecessor which was)
  2) Systems that are affected (but don't (yet) have a patch available but a predecessor already has)

That brings me to the question of why we even look at predecessor products when checking CVEs.
The system only seems to be used in this place in
`CVEAuditManager.java`
```
Optional<CVEPatchStatus> latestInstalled = packageResults.stream()
         .filter(r -> r.isPackageInstalled() &&
                r.getChannelRank().orElse(null) < PREDECESSOR_PRODUCT_RANK_BOUNDARY)
        .max(evrComparator);
```
which IIUC only seems to address one corner case that we encountered back in 2015

This PR covers this corner case and should solve the customer issue, but essentially removes predecessor products being looked at.
So I am worried that this might break some cases that I am not aware of (`java_pgsql_tests` are passing though)

I already asked around but you seem to be the only one who can (hopefully) answer this.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManager.java
```diff
@@ -954,6 +954,7 @@ private static List<CVEAuditSystemBuilder> listSystemsByPatchStatus(List<CVEPatc
             // Group results for the system further by package names, filtering out 'not-affected' entries
             Map<String, List<CVEPatchStatus>> resultsByPackage =
                     systemResults.stream().filter(r -> r.getErrataId().isPresent())
+                            .filter(r -> r.getChannelRank().get() < PREDECESSOR_PRODUCT_RANK_BOUNDARY)
```
mcalmer, line 957:
```suggestion
                            .filter(r -> r.getChannelRank().orElse(0L) < PREDECESSOR_PRODUCT_RANK_BOUNDARY)
```

If for whatever "rank" is empty, we assume 0 and include it into the result check. I think this is better
then risk an exception.

mcalmer (changes requested):
added a small suggestion to improve the code

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManager.java
parlt91: Yes that makes sense. I'll make the change

mcalmer (approved):

PR #6250, opened 2022-12-02
Fix password generation in uyuni-setup-reportdb

### LINKED ISSUE #6247: Broken password generation
When using a non-UTF-8 locale, `uyuni-setup-reportdb` produces a non-UTF-8 password string, which is not accepted by spacewalk.

# Error message
Spacewalk expects an UTF-8 password:
```
File "/usr/lib/python3.6/site-packages/spacewalk/common/rhnTB.py", line 30, in <module>
    from spacewalk.common.rhnConfig import CFG, PRODUCT_NAME
  File "/usr/lib/python3.6/site-packages/spacewalk/common/rhnConfig.py", line 539, in <module>
    ALL_CFG.parse()
  File "/usr/lib/python3.6/site-packages/spacewalk/common/rhnConfig.py", line 145, in parse
    self.__parsedConfig = parse_file(self.filename)
  File "/usr/lib/python3.6/site-packages/spacewalk/common/rhnConfig.py", line 422, in parse_file
    lines = read_file(filename)
  File "/usr/lib/python3.6/site-packages/spacewalk/common/rhnConfig.py", line 463, in read_file
    return configfile.readlines()
  File "/usr/lib64/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfa in position 970: invalid start byte
```

# Source of the problem
Random data is filtered to generate a password, but the filtering is done by using a character class, and these depend on the locale.

Here's the problematic line:
https://github.com/uyuni-project/uyuni/blob/33e8e0bc04ce3c1c6f4d0babd1f703c5fd79fc64/spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb#L555

`[:alnum:]` only generates an universally valid string when using an UTF-8 locale because of a (well-known) bug/limitation in GNU `tr`, which doesn't support multi-byte locales properly, so it fails to match anything outside ASCII in the UTF-8 case, but will happily produce non-UTF-8 symbols in single-byte locales.

The script needs to avoid character classes, perhaps just use the equivalent `[0-9A-Za-z]`.

But this approach is also wasteful with randomness... it might be a better idea to use:
```
PGPASSWORD=$(dd status=none bs=19 count=1 if=/dev/random | base64 | sed 's/...$//')
```

Or perhaps an even more dense encoding.

iluceno:
## What does this PR change?

* Fixes password generation on non-UTF-8 locales.
* Uses an approach less wasteful of random data.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: trivial?

- [ ] **DONE**

## Links

Fixes #6247

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
+++ b/spacewalk/uyuni-setup-reportdb/bin/uyuni-setup-reportdb
@@ -552,7 +552,7 @@ while true ; do
                 echo "Invalid options: do not use --password together with --autgenpw!" >&2
                 exit 1
             fi
-            PGPASSWORD=$(head -10 /dev/urandom | tr -dc '[:alnum:]' | fold -w 25 | head -n 1)
+            PGPASSWORD=$(dd status=none bs=18 count=1 < /dev/random | base64)
             ;;
         -s|--standalone)
             STANDALONE=true
--- a/spacewalk/uyuni-setup-reportdb/uyuni-setup-reportdb.changes
+++ b/spacewalk/uyuni-setup-reportdb/uyuni-setup-reportdb.changes
@@ -1,3 +1,5 @@
+- Fix password generation in uyuni-setup-reportdb (bsc#1205919)
+
 -------------------------------------------------------------------
 Wed Sep 28 11:01:10 CEST 2022 - jgonzalez@suse.com
 


mcalmer:
@iluceno 
Would you please also write a changelog entry to uyuni-setup-reportdb.changes
https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs

Without any header, please.
Just a line describing the change and has the bugnumber on top of the file.
I saw you in the bug, so you know how it goes.

Thanks.

mcalmer (approved):
LGTM - thanks

mcalmer:
L3 closed now

PR #6246, opened 2022-12-02
Add 'octet-stream' to accepted mirrorlist content-types

zzaimeche:
## What does this PR change?

This commit adds 'octet-stream' to the list of acceptable content-types for urls pointing to mirrorlists. This fixes:
https://github.com/uyuni-project/uyuni/issues/6244 and ensures that amazonlinux channels will continue to work.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes
- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #
Tracks # **add downstream PR, if any**

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #6252, opened 2022-12-02
Revert "Add 'octet-stream' to accepted mirrorlist content-types"

zzaimeche:
Reverts uyuni-project/uyuni#6246


meaksh (approved):
We are good to revert this as we are going to release it via patch for Uyuni 2022.11.

As discussed, we need these changes at `Uyuni-2022.11` branch and then Julio will push them forward to `master` again.

PR #6261, opened 2022-12-05
QE: Login/Logout from API in order to enable the use of API calls during the steps

srbarrios:
## What does this PR change?

Fix this issue: https://github.com/SUSE/spacewalk/issues/19755

In a previous PR I decoupled the responsibility of login/logout from our API handler in a step that had these actions in an atomic way inside the step. I made that way, to be consistent to the way we were handling API calls in the rest of step definitions.

Applying this PR, will fix missing login/logout actions from outside the step, at the beginning and at the end of each feature using that step. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.3 (It was only ported in 4.3)

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_activationkey.feature
+++ b/testsuite/features/secondary/min_activationkey.feature
@@ -6,6 +6,7 @@ Feature: Bootstrap a Salt minion via the GUI with an activation key
 
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
+    And I am logged in API as user "admin" and password "admin"
 
   Scenario: Delete SLES minion system profile
     Given I am on the Systems overview page of this "sle_minion"
@@ -121,3 +122,6 @@ Feature: Bootstrap a Salt minion via the GUI with an activation key
   Scenario: Check events history for failures on SLES minion with activation key
     Given I am on the Systems overview page of this "sle_minion"
     Then I check for failed events on history event page
+
+  Scenario: Cleanup: Logout from API
+    When I logout from API
--- a/testsuite/features/secondary/min_baremetal_discovery.feature
+++ b/testsuite/features/secondary/min_baremetal_discovery.feature
@@ -6,6 +6,7 @@ Feature: Bare metal discovery
 
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
+    And I am logged in API as user "admin" and password "admin"
 
   Scenario: Delete the Salt Minion for bare metal feature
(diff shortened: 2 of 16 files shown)


testsuite/features/secondary/min_activationkey.feature
```diff
@@ -121,3 +122,6 @@ Feature: Bootstrap a Salt minion via the GUI with an activation key
   Scenario: Check events history for failures on SLES minion with activation key
     Given I am on the Systems overview page of this "sle_minion"
     Then I check for failed events on history event page
+
+  Scenario: Cleanup: Logout from API
```
srbarrios, line 126: Before someone mention it, yes all the Cleanup scenarios to logout have the same name, because when we do the exact same actions in a scenario, I consider we don't need to bother to use a different name, and it's even better when we need to refactor that scenario and apply it to all the equivalent ones.

orestis87 (approved):

testsuite/features/secondary/min_activationkey.feature
```diff
@@ -6,6 +6,7 @@ Feature: Bootstrap a Salt minion via the GUI with an activation key
 
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
+    And I am logged in API as user "admin" and password "admin"
```
ktsamis, line 9: If I can add a nitpick, "logged from API as" to match the "Logout from API" which I think is the grammatically correct. That would require the renaming of the step as well though.

ktsamis (changes requested):

testsuite/features/secondary/min_activationkey.feature
srbarrios: Sorry but that's out of scope, those step definitions were created looong time ago by Eric. Let's keep it like they are for this PR.

nodeg: I would agree. E.g.
```suggestion
    And I am logged in via API as user "admin" and password "admin"
```
sound better.

srbarrios: Sorry but no, I will not change this step definition in this PR. This is out of the scope.

ktsamis (approved):

PR #6273, opened 2022-12-06
Add the config parameter for tls protocols that are supported

admd:
## What does this PR change?

Some security tools complain that TLSv1 and TLSv1.1 are enabled even if they are not being used and some users would like to disable them. This PR gives this possibility so the user can override those protocols with the help of the configuration parameter.

One possible downside is that it also opens the door for users to add protocols like SSLv3 which is not supported by us. So open to discussion if we could come up with a better solution.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: although visible to the user decided against documenting it.

- [ ] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #

- [] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
```diff
@@ -209,6 +209,12 @@ public class ConfigDefaults {
      */
     private static final String TASKOMATIC_CHANNEL_REPODATA_WORKERS = "java.taskomatic_channel_repodata_workers";
 
+    /**
+     * TLS protocols which should be allowed by Uyuni server
+     */
+
+    private static final String TLS_PROTOCOLS_SUPPORTED_BY_SERVER = "java.supported.tls.protocols";
```
mcalmer, line 216: dots are separators. I would suggest to use underscore as for the other values:

`java.supported_tls_protocols`

java/conf/rhn_java.conf
```diff
@@ -252,3 +252,6 @@ java.min_schema_version = 4.3.11
 
 # minimal required DB reporting schema version
 java.min_report_schema_version = 4.3.5
+
+# TLS protocols that are supported by the server
+java.supported.tls.protocols = TLSv1, TLSv1.1, TLSv1.2, TLSv1.3
```
mcalmer, line 257: Should be changed to `java.supported_tls_protocols`

mcalmer (changes requested):
We also use a second HTTP lib which seems not to support such settings.
I am unsure in providing such a config value which work only for a part of the product.

mcalmer:
@admd what about this ?

PR #6277, opened 2022-12-06
Fix products controller to keep loading mandatory channels

wweellddeerr:
## What does this PR change?

When loading mandatory channels, if there is any broken channel in the database a `NoSuchElementException` is thrown and the remaining channel are not processed. This makes CLM stop working even when the user don't want to use these broken channels.

This PR fixes this problem by allowing the mandatory channels map to be loaded for channels that are not broken.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/19250

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


parlt91 (approved):
LGTM

PR #6285, opened 2022-12-07
QE: Refactor to speed up navigation to overview page plus login to API

srbarrios:
## What does this PR change?

This is a draft where I refactored the navigation step that goes to the overview page of a host name passed by parameter. For that purpose, it uses our API Handler, but this API handler must first have logged in. 

Then, with our current design, we were relying on log-in using another step definition before each API call, and log-out after it using another step. The API Handler is instantiated in a global variable, therefore if we log-out in a feature and we run several features in parallel it will collide and we will have an API handler logged out. 

To solve that issue and also enable the use of API calls from our step definitions in a transparent manner, I decided to include into this draft a proposal that will login/logout inside every API call, delegating the responsibility to authenticate directly to the API Handler, making it transparent from the user point of view at the Cucumber steps, and making reliable when multiple calls are made at same time, as we integrated now a semaphore in the call method.

This refactor in the API Handler will make our development of tests using the API much easier to code and will assure that we don't have collisions with logouts when we run multiple features in parallel.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

- Manager-4.3 https://github.com/SUSE/spacewalk/pull/19940

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


srbarrios:
https://ci.suse.de/blue/organizations/jenkins/uyuni-prs-ci-tests/detail/uyuni-prs-ci-tests/1997/pipeline

testsuite/features/support/env.rb
```diff
@@ -133,6 +133,14 @@ def enable_assertions
   end
 end
 
+BeforeAll do
+  begin
+    $api_test.auth.login('admin', 'admin')
```
mcalmer, line 138: AFAIK we have special API tests which use different users. Does this still work? I think you need logout first at these tests, right?

srbarrios: None of our calls to login to the API were using other user than "admin", as you can see in the deleted lines.
We might be testing with a different user only for WebUI authentication.

srbarrios:
Another approach (I would need more time to explore it), it could be to instantiate an API Handler per Feature, instead of having it as global variable during all the life of the test framework.
In such case, we should use BeforeAll and AfterAll, to instantiate, login, and finally logout and delete the instance of each API Handler.

Bischoff (changes requested):
1) I would separate this PR in two features:
* the one for the UI
* the one for the API
as these are two different problems with two different technical approaches to solve them.


2) API part: I suspect that it will make our manual testing of individual scenarios harder, as we might have to introduce login steps before the scenarios we want to test. We already have that problem when testing UI scenarios.

If the performance benefit is huge, we might find the drawback acceptable.
Also, please ignore this remark if this will not be necessary.


3) The UI has a timeout for being logged in ; does the API have one too? if so, you might not be able to keep being logged in all the time.


4) I don't like very much to hide things in `BeforeAll` clauses.

First, they will be applied at the beginning of every feature, even if we don't need the API, resulting in a loss of performance. Most features don't use the API.

Second, it's way more legible to place a "Given" clause at the beginning of a feature, as we already do for both UI and API.

srbarrios:
> 1) I would separate this PR in two features:
> * the one for the UI
> * the one for the API
> as these are two different problems with two different technical approaches to solve them.
> 
> 

What I tried to explain in the description is that in order to retrieve the system ID that we will use to navigate directly to the overview page of a host passed by parameter (instead of navigating as we do now with clicks), we must use an API method, therefore we need to have access to the API, therefore we need to be logged in.
And as I tried to remark in the description we navigate to the overview page of a host in almost all our features, requiring then API access in every feature.

> 2) API part: I suspect that it will make our manual testing of individual scenarios harder, as we might have to introduce login steps before the scenarios we want to test. We already have that problem when testing UI scenarios.
> 

No you don't need it, BeforeAll is automatically called at the beginning of every feature indepently the scenarios you run, it will be then be even better that  our current state, where we must always assure to keep the scenario that logs in (WebUI+API)

> If the performance benefit is huge, we might find the drawback acceptable.
> Also, please ignore this remark if this will not be necessary.
> 
> 
> 3) The UI has a timeout for being logged in ; does the API have one too? if so, you might not be able to keep being logged in all the time.
> 

I also explained it in the description. Yes we have one, that's the reason why we log in on every feature to avoid the expiration.

> 
> 4) I don't like very much to hide things in `BeforeAll` clauses.
> 
> First, they will be applied at the beginning of every feature, even if we don't need the API, resulting in a loss of performance. Most features don't use the API.
> 

Not anymore true. If we navigate to the overview page using the direct URI through the system_id, as explained in my comment above.

> Second, it's way more legible to place a "Given" clause at the beginning of a feature, as we already do for both UI and API.

I consider we dont need to follow the same approach we have for UI, in our API case, personally I prefer to have an API Handler that automatically handle the authentication for us, consider the API as a tool for our test framework, not as a feature to test as a user will do.

srbarrios:
> Another approach (I would need more time to explore it), it could be to instantiate an API Handler per Feature, instead of having it as global variable during all the life of the test framework.
> In such case, we should use BeforeAll and AfterAll, to instantiate, login, and finally logout and delete the instance of each API Handler.

After research about it, no, it will be not possible to instantiate a variable at feature scope.
Only globally using World, or by scenario.

So, another option it could be to handle API Handler at global scope, but api login/logout at a API Call method inside the API Handler class using a semaphore. That way we also preserve no colisions between scenarios running in parallel.

Bischoff:
(being edited, might change)

OK, so this is not anymore about keeping logged in all the time and not logging out.
This is now about logging in before every call and logging out after every API call.

Why not. Open questions and remarks:

1. This this is not how people use the API, they log in once, do several operations, and then log out.
    But I guess that is okay, our main goal is not to test the API like customers.
2. Isn't that slower?
    You already answered in chat that it isn't slower.
3. Are you sure there still won't be problems with parallelism?
     Even if you have a login / logout sequence per API call, you could still logout
     while the other thread is issuing an API call. Unless you use thread-safe semaphores, which you apparently
    do, I have to check.
4. About the other approach of using one login per feature:
    Yes we do not have variables per feature, but we could simulate them by using a map of handlers
    indexed by the name of the feature.
5. About the other approach of letting the software use host names instead of SIDs (@cbosdo),
    finally, that was not a bad idea, it would spare us a great load of gymnastic.

(being edited, might change)

testsuite/documentation/cucumber-steps.md
```diff
@@ -590,15 +590,6 @@ Note that the text area variant handles the new lines characters while the other
   When I apply highstate on "sle_minion"
 ```
 
-### XML-RPC or HTTP API
```
Bischoff, line 593: don't remove the title :smile_cat:

Bischoff (approved):
Approving, but please:
 - fix the small gotcha in the doc file
 - read my existential questioning :smiley_cat:  in the comments

srbarrios:
The testsuite passed, we can merge it.
https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/2026/TestSuite_20Report_20for_20Pull_20Request_20systemsmanagement_3aUyuni_3aMaster_3aPR_3a6285/

PR #6316, opened 2022-12-13
Testsuite: Enable error detection in SMDBA feature

elariekerboull:
## What does this PR change?

Add a step checking the database status output to detect potential errors.

## GUI diff

No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Cucumber tests were edited

- [x] **DONE**

## Links

Tracks # 
4.2 https://github.com/SUSE/spacewalk/pull/19900
4.3 https://github.com/SUSE/spacewalk/pull/19901

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/smdba_steps.rb
+++ b/testsuite/features/step_definitions/smdba_steps.rb
@@ -65,7 +65,7 @@ Then(/^the configuration should not be set to "(.*?)"$/) do |value|
 end
 
 Then(/^I issue command "(.*?)"$/) do |command|
-  $output, _code = $server.run(command, check_errors: false)
+  $output, _code = $server.run(command, check_errors: true)
 end
 
 Then(/^tablespace "([^"]*)" should be listed$/) do |ts|
--- a/testsuite/features/step_definitions/smdba_steps.rb
+++ b/testsuite/features/step_definitions/smdba_steps.rb
@@ -1,4 +1,4 @@
-# Copyright 2011-2020 SUSE LLC.
+# Copyright 2011-2022 SUSE LLC.
 # Licensed under the terms of the MIT license.
 
 Given(/^a postgresql database is running$/) do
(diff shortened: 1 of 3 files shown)


testsuite/features/step_definitions/smdba_steps.rb
```diff
@@ -65,7 +65,7 @@
 end
 
 Then(/^I issue command "(.*?)"$/) do |command|
-  $output, _code = $server.run(command, check_errors: false)
+  $output, _code = $server.run(command, check_errors: true)
```
vandabarata, line 68: looks good to me, just don't forget to update the year at the top of the file

srbarrios: This step only collect the logs, so the next step have an assert that it suppose to throw the error.

elariekerboull: Good point, some have output checking afterwards, others don't, it might make more sense to add some steps to check the output in those that don't have any checking (e.g.  https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/finishing/srv_smdba.feature#L88)

Bischoff (commented):
see suggestions

testsuite/features/step_definitions/smdba_steps.rb
srbarrios: Personally, I would refactor all these steps that use this global variable storing the output.
This is impossible to handle it as it is when we start parallelizing features.
Instead, we should have a global variable (as feature scope is not supported) and have a hashmap/dictionary on it, and we have a key_value pair, where in the key we store the feature filename, and in the value we store another map where we can store stuff for the feature, like for instance this "output" variable.

elariekerboull: That specific feature might be an exception to parallelization since it is being run in the finishing stage since it `is destructive for other features` ( [quote](https://github.com/uyuni-project/uyuni/blob/master/testsuite/run_sets/finishing.yml#L12) ) and starting to implement this hashmap dictionary might require its own card.

srbarrios: Sure, sure :)

vandabarata (approved):

srbarrios (approved):

Bischoff (approved):

PR #6322, opened 2022-12-13
QE: Refactor get_system_name

srbarrios:
## What does this PR change?

In a previous changes I refactored the method get_system_name so it raised and error if the node did not exist, I did it so we could catch easily some issues with not defined host.
But I missed a case where we are actually relying on not having it defined to return the text of the host parameter. I recovered this with some extra adjustments.

image

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Ports:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


elariekerboull (approved):
Rubocop does not seem happy, besides that it looks good, thanks for fixing that!

PR #6353, opened 2022-12-15
Fix addition of custom channels to activation key

Bischoff:
## What does this PR change?

Fix addition of custom channels to activation key

- for the filtered CLM result
- for the monitoring server
- for the build host and terminal

Tested on 4.3.3 BV deployment.

## Links

Ports:
* 4.2: SUSE/spacewalk#19914
* 4.3: SUSE/spacewalk#19913

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/api_common.rb
+++ b/testsuite/features/step_definitions/api_common.rb
@@ -245,21 +245,27 @@ When(/^I create an activation key including custom channels for "([^"]*)" via AP
   # Get the list of child channels for this base channel
   child_channels = $api_test.channel.software.list_child_channels(base_channel)
 
-  # Select all the child channels for this client
-  client.sub! 'ssh_minion', 'minion'
-  if client.include? 'buildhost'
-    selected_child_channels = ["custom_channel_#{client.sub('buildhost', 'minion')}", "custom_channel_#{client.sub('buildhost', 'client')}"]
-  elsif client.include? 'terminal'
-    selected_child_channels = ["custom_channel_#{client.sub('terminal', 'minion')}", "custom_channel_#{client.sub('terminal', 'client')}"]
-  else
-    custom_channel = "custom_channel_#{client}"
-    selected_child_channels = [custom_channel]
-  end
-  child_channels.each do |child_channel|
-    selected_child_channels.push(child_channel) unless child_channel.include? 'custom_channel'
-  end
+  # Filter out the custom channels
+  # This is needed because we might have both a traditional custom channel and a Salt custom channel
+  child_channels.reject! { |channel| channel.include? 'custom_channel' }
 
-  $api_test.activationkey.add_child_channels(key, selected_child_channels)
+  # Re-add the desired custom channel
+  # This too can go away when we get rid of traditional clients for good
+  client.sub! 'ssh_minion', 'minion'
+  client.sub! 'buildhost', 'minion'
+  client.sub! 'terminal', 'minion'
+  client.sub! 'monitoring_server', 'sle15sp4_minion'
+  custom_channel = if client.include? 'rocky8'
+                     'no-appstream-8-result-custom_channel_rocky8_minion'
+                   elsif client.include? 'rocky9'
+                     'no-appstream-9-result-custom_channel_rocky9_minion'
+                   else
+                     "custom_channel_#{client}"
+                   end
+  child_channels.push(custom_channel)
+
+  # Add child channels to the key
+  $api_test.activationkey.add_child_channels(key, child_channels)
 end
 
 ## actionchain namespace


testsuite/features/step_definitions/api_common.rb
```diff
@@ -245,21 +245,26 @@
...
+  client.sub! 'terminal', 'minion'
+  client.sub! 'monitoring_server', 'sle15sp4_minion'
+  custom_channel = if client.include? 'rocky8'
+                     "no-appstream-8-result-custom_channel_#{client}"
+                   elsif client.include? 'rocky9'
+                     "no-appstream-9-result-custom_channel_#{client}"
+                   else
+                     "custom_channel_#{client}"
+                   end
+  child_channels.push(custom_channel)
```
srbarrios, line 264: What happen if the client has multiple custom channels?

testsuite/features/step_definitions/api_common.rb
```diff
@@ -245,21 +245,26 @@
...
+  child_channels.reject! { |channel| channel.include? 'custom_channel' }
 
-  $api_test.activationkey.add_child_channels(key, selected_child_channels)
+  # Re-add the desired custom channel
+  client.sub! 'ssh_minion', 'minion'
+  client.sub! 'buildhost', 'minion'
+  client.sub! 'terminal', 'minion'
+  client.sub! 'monitoring_server', 'sle15sp4_minion'
+  custom_channel = if client.include? 'rocky8'
+                     "no-appstream-8-result-custom_channel_#{client}"
```
srbarrios, line 258: Why not just "no-appstream-8-result-custom_channel_rocky8_minion" ?

testsuite/features/step_definitions/api_common.rb
```diff
@@ -245,21 +245,26 @@
...
+                   elsif client.include? 'rocky9'
+                     "no-appstream-9-result-custom_channel_#{client}"
```
srbarrios, line 260: I don't like to have defined here those channels.
Can't we use any map in the constants where these channels are already defined?

testsuite/features/step_definitions/api_common.rb
Bischoff: It does not happen, but if you really want to ignore YAGNI, I can make it an array again.

testsuite/features/step_definitions/api_common.rb
Bischoff: By symmetry with the rest, but yeah, if you prefer like that, I have no strong feeling.

Please confirm I should do it.

testsuite/features/step_definitions/api_common.rb
srbarrios: I see multiple custom channels defined in our JSON file for each client.
https://github.com/SUSE/spacewalk/issues/19518#issuecomment-1329379893

testsuite/features/step_definitions/api_common.rb
Bischoff: We can, but that's yet another array for just one very specific case.

Should I do it nonetheless?

testsuite/features/step_definitions/api_common.rb
srbarrios: I said nothing Eric. We only have one custom channel per client, and we have inside that one all our MU repos.

testsuite/features/step_definitions/api_common.rb
srbarrios: It will depend if we can use the map or not, if not, well, as you prefer, it's also fine like that.

testsuite/features/step_definitions/api_common.rb
Bischoff: i had a look, and no

the existing map is for the *base channels*, here we are dealing with the *custom chanels*

(`no-appstream-8-result-rhel8-pool-x86_64` versus `no-appstream-8-result-custom_channel_rocky8_minion`)

testsuite/features/step_definitions/api_common.rb
Bischoff: Perfect then

testsuite/features/step_definitions/api_common.rb
Bischoff: I used the "fixed strings" version, it's indeed a bit ridiculous to interpolate in this case.

srbarrios (approved):
Go ahead, if we consider it, we can iterate over it in the future

testsuite/features/step_definitions/api_common.rb
srbarrios: OK, let's keep it for now, I don't want to slow down this task.
But if this grow, I think we need to move these definitions to a map in our constants, having definitions spread over the steps is bad in terms of maintenance.

PR #6416, opened 2023-01-03
Fix ise activation key create

mcalmer:
## What does this PR change?

When creating an activation key it might happen that Loading the channels takes a while.
The testsuite does not always wait for the page being fully loaded.
This PR detect this case and throw a better message than an ISE and change the testsuite to wait for the 
"Loading..." text to disappear.
Additionally it align all occurrence of "Loading..." to use the same number of dots (3 dots)

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Cucumber tests were adapted

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/pull/20026 https://github.com/SUSE/spacewalk/pull/20027

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/token/ActivationKeyDetailsAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/token/ActivationKeyDetailsAction.java
@@ -14,6 +14,7 @@
  */
 package com.redhat.rhn.frontend.action.token;
 
+import com.redhat.rhn.common.localization.LocalizationService;
 import com.redhat.rhn.common.validator.ValidatorException;
 import com.redhat.rhn.domain.channel.Channel;
 import com.redhat.rhn.domain.channel.ChannelFactory;
@@ -331,7 +332,10 @@ public class ActivationKeyDetailsAction extends RhnAction {
 
     private Channel lookupChannel(DynaActionForm daForm, User user) {
         Long selectedChannel = (Long)daForm.get(SELECTED_BASE_CHANNEL);
-
+        if (selectedChannel == null) {
+            throw new ValidatorException(
+                    LocalizationService.getInstance().getMessage("activation-key.java.nochannel"));
+        }
         if (!DEFAULT_CHANNEL_ID.equals(selectedChannel)) {
             return ChannelManager.lookupByIdAndUser(
                                 selectedChannel, user);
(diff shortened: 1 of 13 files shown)


Etheryte (approved):
👍

java/code/src/com/redhat/rhn/frontend/action/token/ActivationKeyDetailsAction.java
```diff
@@ -331,7 +331,9 @@ private ActivationKey create(DynaActionForm daForm, RequestContext context) {
 
     private Channel lookupChannel(DynaActionForm daForm, User user) {
         Long selectedChannel = (Long)daForm.get(SELECTED_BASE_CHANNEL);
-
+        if (selectedChannel == null) {
+            throw new ValidatorException("No Base Channel selected");
```
mackdk, line 335: Is this message shown in the page and should we localized it? According to the `ValidatorException` constructor usages, I see that currently  sometimes we do and sometimes we don't...

mackdk (approved):

testsuite/features/core/srv_create_activationkey.feature
```diff
@@ -12,6 +12,7 @@ Feature: Create activation keys
   Scenario: Create an activation key with a channel
     When I follow the left menu "Systems > Activation Keys"
     And I follow "Create Key"
+    And I wait until I do not see "Loading..." text
```
srbarrios, line 15: We really need to check including dots?
I think we can be more flexible in our checks if we don't add the final dots, three dots, etc.
In any case... approving the PR.

java/code/src/com/redhat/rhn/frontend/action/token/ActivationKeyDetailsAction.java
mcalmer: I added the translation

testsuite/features/core/srv_create_activationkey.feature
mcalmer: I think it is better to add them. What happens if somebody add "Loading" as part of the description?
Or we add a text on this page which contain the test "Loading".
I just feel better when we have it.

mcalmer:
PR testsuite https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/2060/

PR #6424, opened 2023-01-04
Fix transaction commit behavior for Spark routes

wweellddeerr:
## What does this PR change?

It adds a Spark after filter to deal with hibernate sessions for spark routes compatible with `SessionFilter`. Without this patch, responses are send back without having the transaction commited because Spark closes response stream for the requests it handles before `SessionFilter` finishes running. It was even possible to have an HTTP 200 response and a failure when committing the transaction.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
+++ b/java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
@@ -72,6 +72,8 @@ public abstract class HibernateFactory {
     private static final Logger LOG = LogManager.getLogger(HibernateFactory.class);
     private static final int LIST_BATCH_MAX_SIZE = 1000;
 
+    public static final String ROLLBACK_MSG = "Error during transaction. Rolling back";
+
     protected HibernateFactory() {
     }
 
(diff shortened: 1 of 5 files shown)


java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
```diff
@@ -379,6 +381,31 @@ public static void commitTransaction() throws HibernateException {
         connectionManager.commitTransaction();
     }
 
+    /**
+     * @param committed - if it was possible to commit the transaction.
+     * Roll back transaction in case it is not committed and close the Hibernate session.
+     */
+    public static void finishTransaction(boolean committed) {
```
rjmateus, line 388: I'm not sure about this method here. It's not also clear what the method does. The name is finish transaction, receives a parameter named committed, but it doesn't commit anything.
Maybe it could change this method and not receive any parameter, and we perform the commit and rollback logic here (or another helper class) and remove it from the existing caller.
The exception handling can be done outside, since each caller have is logic.

rjmateus (changes requested):
just a comment, but other then that looks good.

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
mcalmer: I understood that "commited" is the info if the commit was successful or not. I think the behavior differ if it was successful or not.

rjmateus: sure, but why not do the commit here, if the logic is the same in both places, in this way we treat the errors in a central place as the commit.
Also, the method name looks like it will deal with all, but it doesn't, and if you don't read what the parameter description says, you can be fooled. 
One of the biggest problem in computer science: naming stuff :)

rjmateus (approved):
LGTM

cbosdo (approved):
LGTM

PR #6441, opened 2023-01-09
Do not include pillar_only formulas in highstate

aaannz:
## What does this PR change?

Do not include pillar_only formulas in highstate

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py
+++ b/susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py
@@ -288,7 +288,8 @@ def formula_pillars(system_formulas, group_formulas, all_pillar):
         formula_metadata = load_formula_metadata(formula_name)
         if formula_name in out_formulas:
             continue # already processed
-        out_formulas.append(formula_name)
+        if not formula_metadata.get('pillar_only', False):
+            out_formulas.append(formula_name)
         pillar = salt.utils.dictupdate.merge(pillar,
                        load_formula_pillar(system_formulas.get(formula_name, {}),
                            group_formulas[formula_name],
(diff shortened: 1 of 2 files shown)


agraul:
Before reviewing this PR, it would be great to have a bit more context on the change. The two main questions that I am curious about are:

- What is "formula_only pillar" ? 
- Why shouldn't it be in highstate?

aaannz:
> Before reviewing this PR, it would be great to have a bit more context on the change. The two main questions that I am curious about are:
> 
>     * What is "formula_only pillar" ?

You mean `pillar_only`. It's not pillar, it is metadata info from metadata.yml of respective formula. It marks that the formula does not have any state, just provide pillar values.

> 
>     * Why shouldn't it be in highstate?

If formula does not have any state, when salt tries to apply formula it will fail it did not find any state named after the formula. So far workaround was to have empty state.

Example: [saltboot-group](https://github.com/uyuni-project/retail/blob/master/saltboot-formula/metadata/saltboot-group-metadata.yml) formula which only provides pillar values for saltboot, but there is no state called `saltboot-group`. Similarly [saltboot](https://github.com/uyuni-project/retail/blob/master/saltboot-formula/metadata/metadata.yml) formula is marked as pillar_only even though there is saltboot state, but that is ignored when applied outside of initrd.

agraul:
> > Before reviewing this PR, it would be great to have a bit more context on the change. The two main questions that I am curious about are:
> > ```
> > * What is "formula_only pillar" ?
> > ```
> 
> Do you mean `pillar_only`?. 

Probably. I copied "formula_only pillar" from the PR title and description verbatim.

> It's not pillar, it is metadata info from metadata.yml of respective formula. It marks that the formula does not have any state, just provide pillar values.
>

Thanks for the explanation. I have to admit it surprises me to have a Salt formula without a state. After all, formulas are collection of states, with optional configuration via pillar data.

> > ```
> > * Why shouldn't it be in highstate?
> > ```
> 
> If formula does not have any state, then when Salt tries to apply formula, it will fail because it couldn't find any state named after the formula. Workaround was to have empty state so far.
> 

So we've been "abusing" formulas to just provide pillar data for some time then? I don't really like it, it feels wrong to do that. But I guess it's too late to change that now.

aaannz:
> 
> Probably. I copied "formula_only pillar" from the PR title and description verbatim.

🤦🏾 sorry! Completely missed that, my mistake. Fixed.

> 
> > It's not pillar, it is metadata info from metadata.yml of respective formula. It marks that the formula does not have any state, just provide pillar values.
> 
> Thanks for the explanation. I have to admit it surprises me to have a Salt formula without a state. After all, formulas are collection of states, with optional configuration via pillar data.

Well we only call it formulas, but it is the Forms which we are (ab)using there.

> > If formula does not have any state, then when Salt tries to apply formula, it will fail because it couldn't find any state named after the formula. Workaround was to have empty state so far.
> 
> So we've been "abusing" formulas to just provide pillar data for some time then? I don't really like it, it feels wrong to do that. But I guess it's too late to change that now.

Yes. Since the retail started.

aaannz:
@agraul Any further questions here? I would like to merge this.

agraul (approved):

PR #6444, opened 2023-01-10
Remove python3-simplejson dependency

zzaimeche:
## What does this PR change?

The spacecmd specfile lists python3-simplejson as a dependency, but this makes no sense, as simplejson is only used when python is < 2.6. The requirement is causing install errors. This patch removes the dependency. There should be no code that expects it.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/spacecmd.changes
+++ b/spacecmd/spacecmd.changes
@@ -1,3 +1,5 @@
+- Remove python3-simplejson dependency
+
 -------------------------------------------------------------------
 Wed Dec 14 14:07:34 CET 2022 - jgonzalez@suse.com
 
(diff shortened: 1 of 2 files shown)


spacecmd/spacecmd.changes
```diff
@@ -1,3 +1,4 @@
+- Remove python3-simplejson dependency
```
meaksh, line 1: Nitpick: let's add some marging here:
```suggestion
- Remove python3-simplejson dependency

```

meaksh (approved):

PR #6464, opened 2023-01-13
Ensure custom Salt modules are still compatible with old Python 2.6

meaksh:
## What does this PR change?

This PR is introducing a new GitHub action to ensure we are not introducing new code which is not Python 2.6 compatible.

Even if Python 2.6 is EOL, we want to prevent breaking existing functionality for those old systems that might be running out there. This is why, we want to make the code for these custom Salt modules to be compatible with Python 2.6.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Added new tests via GitHub action.

- [x] **DONE**

## Links

Tracks 

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
+++ b/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
@@ -10,6 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     container: centos:6
     steps:
+      # We need to use "checkout@v1" here as is the only version compatible with Python 2.6
       - name: Check out repository code
         uses: actions/checkout@v1
 
--- a/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
+++ b/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
@@ -2,6 +2,7 @@ name: Check for non-compatible Python 2.6 in susemanager-sls Salt modules
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened]
     paths: 'susemanager-utils/susemanager-sls/src/**/*'
 
 jobs:
--- a/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
+++ b/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
@@ -35,5 +35,7 @@ jobs:
       - name: Execute tests to check Python 2.6 compatiblity of custom Salt modules
         run: |
           cd $GITHUB_WORKSPACE/susemanager-utils/susemanager-sls/src/
-          for i in $(find .|grep py$ | grep -v tests); do echo "Testing $i .... "; (python $i && echo "OK") || IS_FAILED=1; echo; done
+          # Exclude tests and uyuni_config.py modules (which are for server and not included in susemanager-sls)
+          TESTS_LIST=$(find .|grep py$ | grep -v "^./tests" | grep -v "uyuni_config.py")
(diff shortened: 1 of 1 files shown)


.github/workflows/susemanager-sls-compatible-python26-syntax.yml
```diff
@@ -0,0 +1,40 @@
...
+    types: [opened, synchronize, reopened]
+    paths: 'susemanager-utils/susemanager-sls/src/**/*'
+
+jobs:
+  susemanager_sls_check_non_python26_syntax:
+    runs-on: ubuntu-latest
+    container: centos:6
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v1
```
agraul, line 14: Is there a reason we don't use the current version?

.github/workflows/susemanager-sls-compatible-python26-syntax.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Install old Salt from products:old
+        run: |
+          yum install -y salt
+
+      - name: Execute tests to check Python 2.6 compatiblity of custom Salt modules
+        run: |
+          cd $GITHUB_WORKSPACE/susemanager-utils/susemanager-sls/src/
+          # Exclude tests and uyuni_config.py modules (which are for server and not included in susemanager-sls)
+          TESTS_LIST=$(find .|grep py$ | grep -v "^./tests" | grep -v "uyuni_config.py")
```
agraul, line 38: In the future we can change this to only run on changed files. I guess that's what we'll do when we add a linter etc.

agraul (changes requested):
If the action works, then the code changes are good as well!

The actions looks fine, but I don't like that we mix adding a github actions with a bug fix. Can we split that into two pull requests?

.github/workflows/susemanager-sls-compatible-python26-syntax.yml
meaksh: I had to use `v1` as it was the only version compatible with Python 2.6.

.github/workflows/susemanager-sls-compatible-python26-syntax.yml
meaksh: Yes, fully agree.

meaksh:
@agraul I've splited this in two PRs as suggested. This one just introduce the new GH action, and the fixes for the current issues are coming in this other PR: https://github.com/uyuni-project/uyuni/pull/6475

.github/workflows/susemanager-sls-compatible-python26-syntax.yml
agraul: Then we should add a comment, otherwise we'll forget why we use the old version the next time we look at the workflow yaml.

meaksh: Added a comment to clarify this :+1:

meaksh:
BTW just in case you are curious about how the output from this new GH action looks when there are failures, you can see it here on a fake commit a previously did on my branch (don't pay much attention on the `uyuni_config.py` failures as those files are now excluded from validation - not shipped at `susemanager-sls` package):

https://github.com/uyuni-project/uyuni/actions/runs/3911741372/jobs/6685475836

The new GH action is not being executed on the present PR as we are not doing any change to files under `susemanager-utils/susemanager-sls/src/`.

agraul (approved):

PR #6475, opened 2023-01-16
Fix non-compatible Python 2.6 syntax in custom Salt modules (bsc#1206979) (bsc#1206981)

meaksh:
## What does this PR change?

This is a spin-off PR from https://github.com/uyuni-project/uyuni/pull/6464 that fixes the issues raised by new introduced GH action to check we are not introducing non-compatible Python 2.6 code in our custom Salt modules provided by `susemanager-sls` package.

**UPDATE**: this PR additionally adapts `mgrnet` code to be compatible with old Salt 2016.11.10.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #
Tracks https://github.com/SUSE/spacewalk/issues/20270

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
+++ b/.github/workflows/susemanager-sls-compatible-python26-syntax.yml
@@ -37,5 +37,5 @@ jobs:
           cd $GITHUB_WORKSPACE/susemanager-utils/susemanager-sls/src/
           # Exclude tests and uyuni_config.py modules (which are for server and not included in susemanager-sls)
           TESTS_LIST=$(find .|grep py$ | grep -v "^./tests" | grep -v "uyuni_config.py")
-          for i in $TESTS_LIST; do echo "Testing $i .... "; (python $i && echo "OK") || IS_FAILED=1; echo; done
+          for i in $TESTS_LIST; do echo "Testing $i .... "; (python -m py_compile $i && echo "OK") || IS_FAILED=1; echo; done
           exit $IS_FAILED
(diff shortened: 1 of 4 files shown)


susemanager-utils/susemanager-sls/src/modules/sumautil.py
```diff
@@ -11,6 +11,7 @@
 import re
 import time
 import salt.utils
+import salt.modules.cmdmod
```
ycedres, line 14: @meaksh Just out of curiosity, does this have anything to do with that Python 2.6 compatibility or does this fixes something else?

meaksh: This is actually something else that was spotted by the new GH action to detect non-compatible syntax, as it was causing an `ImportError` when calling the file with `python` directly.

Apparentely this missing `import` was not really causing any issue when executed by Salt and loaded via LazyLoader.

agraul: Is the `import` also needed if we just compile the code? We don't need to run the code to see syntax errors:

```
➤ podman run -it --rm centos:centos6 /bin/bash
[root@9b505b1e64c4 /]# cat foo.py
results = {k:v for k,v in [(1,1), (2,2)]}
[root@9b505b1e64c4 /]# python -m py_compile foo.py
SyntaxError: ('invalid syntax', ('foo.py', 1, 18, 'results = {k:v for k,v in [(1,1), (2,2)]}\n'))

ycedres (approved):
LGTM

meaksh:
@ycedres I just re-requested review here since I had to push an extra commit to this PR to adapt the `mgrnet` module also to work with Salt 2016.11.10.

ycedres (approved):

PR #6481, opened 2023-01-17
Prevent api parameters parsing as dates if not in ISO-8601 (bsc#1205759)

ycedres:
## What does this PR change?

The dateutil `parse` function parses as date/time stamps some strings that are not in ISO-8601 format which causes trouble in `api` calls in spacecmd when the user passes strings as parameters that are similar to dates (for example, to create a label).

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage

- No tests: already covered

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/19690

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/src/spacecmd/utils.py
+++ b/spacecmd/src/spacecmd/utils.py
@@ -51,7 +51,7 @@ from difflib import unified_diff
 from tempfile import mkstemp
 from textwrap import wrap
 from subprocess import Popen, PIPE
-from dateutil.parser import parse as parse_datetime, ParserError
+from dateutil.parser import isoparse
 
 try:
     import json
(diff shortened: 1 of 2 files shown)


spacecmd/src/spacecmd/utils.py
```diff
@@ -51,7 +51,7 @@
 from tempfile import mkstemp
 from textwrap import wrap
 from subprocess import Popen, PIPE
-from dateutil.parser import parse as parse_datetime, ParserError
+from dateutil.parser import isoparse as parse_datetime, ParserError
```
agraul, line 54: I can't see it from this diff in GitHub, is `ParseError` still used?

agraul (approved):
LGTM, I personally would not rename `isoparse` to `parse_datetime` because `isoparse` is a nice name IMO, but that's just my preference.

I left one comment, it would be nice if we could clarify that before we merge the PR.

spacecmd/src/spacecmd/utils.py
ycedres: It is not, I'll remove that from the import :+1:

PR #6489, opened 2023-01-18
Fix custom "mgrcompat.module_run" state function to work with Salt 3005.1

meaksh:
## What does this PR change?

On Salt 3005, the `module.run` state function was refactored and, it was finally decided upstream, that there is no deprecation of legacy syntax for `module.run` happening at all. Both syntax will be supported in future Salt versions.

(See NOTE at the beggining of https://docs.saltproject.io/en/latest/ref/states/all/salt.states.module.html)

In fact, after the refactor done by upstream on 3005 codebase for `module.run` (in order to support both syntaxes), if we tailor our parameters from the old syntax to the new syntax, this would cause issues in the execution of some functions.

Examples:

```
'docker.login' failed: login() got an unexpected keyword argument 'registries'
'cp.push' failed: push() got multiple values for argument 'path'
```

So, this PR simply fixes our custom `mgrcompat.module_run` module to not tailor any parameters (at it is not really needed) in case we are running Salt versions higher than 3004.

(This should be fixing some of the issues we currentely see at HEAD acceptance tests)

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/19924

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/src/states/mgrcompat.py
+++ b/susemanager-utils/susemanager-sls/src/states/mgrcompat.py
@@ -50,21 +50,12 @@ def module_run(**kwargs):
 
     '''
 
-    # The new syntax will be used as the default
-    use_new_syntax = True
+    # We use classic "module.run" syntax by default.
+    use_new_syntax = False
 
-    if __grains__['saltversioninfo'][0] > 3004:
-        # Only new syntax - default behavior for Phosphorus and future releases
-        pass
-    elif __grains__['saltversioninfo'][0] > 2016 and 'module.run' in __opts__.get('use_superseded', []):
+    if 2016 < __grains__['saltversioninfo'][0] < 3005 and 'module.run' in __opts__.get('use_superseded', []):
         # New syntax - explicitely enabled via 'use_superseded' configuration on 2018.3, 2019.2, 3000.x, 3002.x, 3003.x and 3004.x
-        pass
-    elif __grains__['saltversioninfo'][0] > 2016 and not 'module.run' in __opts__.get('use_superseded', []):
-        # Old syntax - default behavior for 2018.3, 2019.2, 3000.x, 3002.x, 3003.x and 3004.x
-        use_new_syntax = False
-    elif __grains__['saltversioninfo'][0] <= 2016:
-        # Only old syntax - the new syntax is not available for 2016.11 and 2015.8
-        use_new_syntax = False
+        use_new_syntax = True
 
     if use_new_syntax:
         log.debug("Minion is using the new syntax for 'module.run' state. Tailoring parameters.")
--- a/susemanager-utils/susemanager-sls/src/tests/test_state_mgrcompat.py
+++ b/susemanager-utils/susemanager-sls/src/tests/test_state_mgrcompat.py
@@ -27,7 +27,7 @@ def test_module_run_on_phosphorous():
         mgrcompat.__grains__, {'saltversioninfo': [3005, None, None, None]}
     ), patch.dict(mgrcompat.__states__, {'module.run': mock}):
         mgrcompat.module_run(**MGRCOMPAT_MODULE_RUN_KWARGS)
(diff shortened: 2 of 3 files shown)


agraul (changes requested):
Given that `module.run` accepts both syntaxes now, I expected that we don't change anything for Salt 3005+ in `mgrcompat.module_run`. Can you clarify why we still run `_tailer_kwargs_to_new_syntax(old_name, **kwargs)`?

meaksh:
Thanks for the quick review! 

Maybe the code is a bit messy, but by default, we are not changing anything for 3005+ - we don't call `_tailor_kwargs_to_new_syntax(old_name, **kwargs)`. But it is true that, in case you place the "use_superseded" config in your minion, then the code will still try to force to the new syntax (which should not really happen).

Let me cover this possible corner case better in the code.

meaksh:
@agraul I think this looks better now

agraul:
> Maybe the code is a bit messy, but by default, we are not changing anything for 3005+ - we don't call `_tailor_kwargs_to_new_syntax(old_name, **kwargs)`.

Are you sure about that? The way I read the code, `use_new_syntax = True` stays `True` for Salt 3005+. The `if use_new_syntax` branch is taken, which calls `_tailor_kwargs_to_new_syntax()`

> But it is true that, in case you explicitely place the "use_superseded" config in your minion, then the code will still try to force to the new syntax (which should not really happen).
>
> Let me cover this possible corner case better in the code.
 

I was only thinking about the default, but if that's also the case we should fix it as well.

agraul:
> @agraul I think this looks better now

Yes, it's better now :+1: . I'd swap the default and get rid of all but one if statement (2016 -- 3004 + use_superseded) to set `use_new_syntax = True` to further simplify the code, but otherwise it's good!

meaksh:
> > @agraul I think this looks better now
> 
> Yes, it's better now +1 . I'd swap the default and get rid of all but one if statement (2016 -- 3004 + use_superseded) to set `use_new_syntax = True` to further simplify the code, but otherwise it's good!

Yeah, this sounds reasonable at this point in time :+1:  - thanks for the suggestion.

I will push some extra changes here.

meaksh:
@agraul suggested changed done :+1:

susemanager-utils/susemanager-sls/src/states/mgrcompat.py
```diff
@@ -50,21 +50,15 @@ def module_run(**kwargs):
...
-    # The new syntax will be used as the default
-    use_new_syntax = True
+    # We use classic "module.run" syntax by default.
+    use_new_syntax = False
 
     if __grains__['saltversioninfo'][0] > 3004:
-        # Only new syntax - default behavior for Phosphorus and future releases
+        # Keep using old syntax - since 3005+ both syntax are supported. Not need of tailoring parameters anymore.
         pass
     elif __grains__['saltversioninfo'][0] > 2016 and 'module.run' in __opts__.get('use_superseded', []):
```
agraul, line 59:
```suggestion
   if 2016 < __grains__['saltversioninfo'][0] < 3005 and 'module.run' in __opts__.get('use_superseded', []):
```

agraul (changes requested):
I was proposing to just have one if statement. Without testing, I suggested what I was thinking  of. The first if statement can then be deleted (GitHub didn't let me do a suggestion for both if statements at once).

susemanager-utils/susemanager-sls/src/states/mgrcompat.py
meaksh: Done

agraul (approved):
Again, haven't tested the code, but it looks good now. Before merging I'd squash the commits since `git log` does not really benefit from the all but the first commit.

meaksh:
I tested this locally and it works fine as expected :+1:

PR #6492, opened 2023-01-18
Include 'gpg' states and files as "ssh_extra_filerefs" to avoid issues on action chains for SSH minions

meaksh:
## What does this PR change?

This PR fixes a problem with channels / packages related actions triggered via actions chains on SSH minions that we have seen happening in the HEAD acceptance tests. We need to explicitely add `gpg` as `ssh_extra_filerefs` on SSH executions made by Uyuni, in order to ensure those extra files are available on the context of SSH action chain execution.

As mentioned, this should fix some issues seen at Head acceptance tests:

```
----------
          ID: mgr_deploy_tools_uyuni_key
    Function: file.managed
        Name: /etc/pki/rpm-gpg/uyuni-tools-gpg-pubkey-0d20833e.key
      Result: false
     Comment: Source file salt://gpg/uyuni-tools-gpg-pubkey-0d20833e.key not found in saltenv 'base'
     Started: 13:57:27.009611
    Duration: 2.696
         SLS: channels.gpg-keys
     Changed: {}
----------
```

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
+++ b/java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
@@ -126,7 +126,7 @@ public class SaltSSHService {
             "distupgrade", "hardware", "images", "packages/init.sls",
             "packages/patchdownload.sls", "packages/patchinstall.sls", "packages/pkgdownload.sls",
             "packages/pkginstall.sls", "packages/pkgupdate.sls", "packages/pkgremove.sls", "packages/profileupdate.sls",
-            "packages/redhatproductinfo.sls", "remotecommands", "scap", "services",
+            "packages/redhatproductinfo.sls", "remotecommands", "scap", "services", "gpg",
             "custom", "custom_groups", "custom_org", "util", "bootstrap", "formulas.sls");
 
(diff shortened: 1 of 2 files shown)


mcalmer (approved):
I was not aware that we need to list every directory we add.

mcalmer:
@meaksh does it mean customers cannot create there own directories in /srv/salt/ and use salt-ssh?

meaksh:
> @meaksh does it mean customers cannot create there own directories in /srv/salt/ and use salt-ssh?

IIUC this would be only problematic when running Action Chains for SSH minions. Because of the way Action Chains are build and created (SLS chunks that calls "module.run" to execute "state.apply", then Salt is not really able to collect all possible necessary SLS and files referenced via `salt://`. This is why we explicitetly add these `ssh_extra_filerefs`.

For normal actions for SSH minion, this shouldn't be an issue.

PR #6499, opened 2023-01-19
Do not pass server related grains to minions

witekest:
## What does this PR change?

Do not pass server related grains to minions.

The grains `is_uyuni` and `has_report_db` are evaluated by reading server configuration files entries. These configuration files are not available on common clients and the grains were always set to default values and of no use.

In case of missing server configuration files the grains `is_uyuni` and `has_report_db` are not being set now.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Existing tests updated

- [x] **DONE**

## Links

Fixes #SUSE/spacewalk#20145

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/src/grains/mgr_server.py
+++ b/susemanager-utils/susemanager-sls/src/grains/mgr_server.py
@@ -13,6 +13,7 @@ RHNCONF = "/etc/rhn/rhn.conf"
 RHNCONFDEF = "/usr/share/rhn/config-defaults/rhn.conf"
 RHNWEBCONF = "/usr/share/rhn/config-defaults/rhn_web.conf"
 
+
 def _simple_parse_rhn_conf(cfile):
     result = {}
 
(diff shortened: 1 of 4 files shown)


agraul:
Before reviewing this PR it would be good if you could provide more context around the change: Why is it needed? What problem does it fix? Why is it safe to do so?

susemanager-utils/susemanager-sls/src/grains/mgr_server.py
```diff
@@ -33,7 +34,7 @@ def server_grains():
     """ returns if this minion is a Uyuni/SUSE Manager Server and
         if it has a reporting database configured.
     """
```
vandabarata, line 36: please also update the method documentation :)

witekest:
> Before reviewing this PR it would be good if you could provide more context around the change: Why is it needed? What problem does it fix? Why is it safe to do so?

I've updated the PR description.

susemanager-utils/susemanager-sls/src/grains/mgr_server.py
witekest: Thanks, done.

mcalmer (approved):
Looks good. Please check the failing test. Maybe it is not related to this change

witekest:
> Looks good. Please check the failing test. Maybe it is not related to this change

I think it's unrelated:

```
Testing ./grains/mgr_server.py .... 
OK
```

PR #6503, opened 2023-01-19
add /etc/passwd to supportutils-plugin-susemanager

mbussolotto:
## What does this PR change?

add /etc/passwd to supportutils-plugin-susemanager. It would help to check user information.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/19477

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/supportutils-plugin-susemanager/susemanager
+++ b/susemanager-utils/supportutils-plugin-susemanager/susemanager
@@ -78,4 +78,5 @@ plugin_command "echo \"select count(*) from susesaltevent;\" | spacewalk-sql --s
 plugin_command "/sbin/supportconfig-sumalog $LOG"
 plugin_command "cp /var/log/zypper.log $LOG"
 plugin_command "grep \"Client bootstrap script\" /srv/www/htdocs/pub/bootstrap/*.sh"
+plugin_command "cat /etc/passwd"
 
--- a/susemanager-utils/supportutils-plugin-susemanager/supportutils-plugin-susemanager.changes
+++ b/susemanager-utils/supportutils-plugin-susemanager/supportutils-plugin-susemanager.changes
@@ -1,3 +1,4 @@
+- add /etc/passwd content
 - update susemanager plugin to export the number of pending salt events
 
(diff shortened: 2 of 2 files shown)


susemanager-utils/supportutils-plugin-susemanager/susemanager
```diff
@@ -78,4 +78,5 @@ plugin_command "echo \"select count(*) from susesaltevent;\" | spacewalk-sql --s
 plugin_command "/sbin/supportconfig-sumalog $LOG"
 plugin_command "cp /var/log/zypper.log $LOG"
 plugin_command "grep \"Client bootstrap script\" /srv/www/htdocs/pub/bootstrap/*.sh"
+plugin_command "cat /etc/passwd"
```
mcalmer, line 81: I think people will not like it. passwords should be in shadow, but paranoid people will see this as a problem.
Especially as we print all entries.

mcalmer (changes requested):
I need to think a bit about it to come up with a acceptable solution

susemanager-utils/supportutils-plugin-susemanager/susemanager
mbussolotto: if we had something like this?
```
    POSTGRES_USER_ENABLED=$(su - postgres -c /bin/true)
    if [ $POSTGRES_USER_ENABLED != "0" ] ; then
        echo "ERROR: Postgres user disabled" 
    else
        echo "Postgres user enabled" 
    fi
``` 
Would it be better?

mcalmer: I just found out that spacewalk-debug get this allready.
Check spacewalk-debug/conf/passwd

It only copy the entries for special users not the whole file.
So maybe you can close this. We have this already

PR #6504, opened 2023-01-19
Fix uyuni-proxy shell script not packaged

avshiliaev:
## What does this PR change?

Fixes for the PR

* https://github.com/uyuni-project/uyuni/pull/6240

Adds the uyuni-proxy script to the package. 

## GUI diff

No difference.

## Documentation

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

* https://github.com/uyuni-project/uyuni/pull/6240

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


avshiliaev:
@cbosdo FYI about the script

# Using a custom container image for a service

By default, the proxy suite is set to use the same image version and registry path for each of its services. However, there’s a possiblity to change this behavior by overriding the default values for a specific service. The `proxy` CLI bundled with the package, has an `update image` command taking the following arguments:

- `-s` for the service name
- `-t` for the version tag
- `-r` for the registry path

An example usage would look like this:

```bash
proxy update image -s httpd -t 0.1.0 -r registry.opensuse.org/uyuni
```

It would adjust a configuration file for the httpd service, taking `registry.opensuse.org/uyuni` as the registry and `0.1.0` as the version tag and then restart it.

To reset the values to defaults, run the `proxy reset` command, specifying the service with the `-s` flag. An example usage would look like the following:

```bash
proxy reset -s httpd
```

This command would reset the configuration of the `httpd` service to the global defaults and restarts it.

See the `proxy —help` for more detail.

PR #6505, opened 2023-01-19
Prevent saving duplicate comps files on repo sync (bsc#1206907)

cbbayburt:
A recent change in reposync allowed multiple `modules.yaml` files to be saved in the DB. This follow-up explicitly prevents  duplicates of other types of comps files to be saved.

See: https://bugzilla.suse.com/show_bug.cgi?id=1206907

## GUI diff

Before (renamed to `comps.x86_64.xml`):
```
19:48:27   Renaming non-standard filename dae7e104812099a2f632ea4c5ef2769aca18ca1205abdd2c3ba6d171e319df3d-comps-BaseOS.x86_64.xml to comps.x86_64.xml.
```

After (renamed to `comps.xml`):
```
19:48:27   Renaming non-standard filename dae7e104812099a2f632ea4c5ef2769aca18ca1205abdd2c3ba6d171e319df3d-comps-BaseOS.x86_64.xml to comps.xml.
```

## Documentation

## Test coverage
- No tests: reposync :man_shrugging: 

## Links

Fixes https://github.com/SUSE/spacewalk/issues/20072

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/spacewalk/satellite_tools/reposync.py
+++ b/python/spacewalk/satellite_tools/reposync.py
@@ -72,6 +72,7 @@ relative_modules_dir = 'rhn/modules'
 relative_mediaproducts_dir = 'suse/media.1'
 checksum_cache_filename = 'reposync/checksum_cache'
 default_import_batch_size = 20
+compressed_suffixes = ['.gz', '.bz', '.xz']
 
 errata_typemap = {
     'security': 'Security Advisory',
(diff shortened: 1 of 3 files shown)


agraul (changes requested):
It would be really useful to add tests for this change. I know the reposync code is ugly and the test coverage is low, but this change is a good candidate to improve that situation. I think it makes sense to extract the "filename manipulations" into its own (pure) function, which can then be tested easily.

cbbayburt:
> It would be really useful to add tests for this change. I know the reposync code is ugly and the test coverage is low, but this change is a good candidate to improve that situation. I think it makes sense to extract the "filename manipulations" into its own (pure) function, which can then be tested easily.

@agraul I couldn't agree more. Is there any guide or pointers on how to write tests for it? (it's alien territory for me :see_no_evil: )

agraul:
> Is there any guide or pointers on how to write tests for it? (it's alien territory for me see_no_evil )

All I'm aware of is https://github.com/uyuni-project/uyuni/blob/master/python/README.md, but that's not really complete, e.g. there are more dependencies that are not listed. Sadly, due to us using `python3-rpm`, it's also not possible to run the tests in a virtual environment (as of now).

The easiest way is to run a container as explained at the end of that README.md:
```sh
➤ UYUNI_ROOT=/home/agraul/src/uyuni/uyuni-main
➤ podman run -t -i --rm=true -e "PYTHONPATH=/manager/python/:/manager/client/rhel/spacewalk-client-tools/src" -v $UYUNI_ROOT:/manager registry.opensuse.org/systemsmanagement/uyuni/master/docker/containers/uyuni-master-pgsql /bin/bash
$ pytest /manager/python/test/unit/spacewalk/satellite_tools/test_reposync.py
```

Now onto the tests themselves. As you probably already know, they are not in a nice shape. Currently they are based on the [unittest ](https://docs.python.org/3/library/unittest.html) framework, but we're executing them with `pytest`. [Pytest](https://docs.pytest.org/en/7.2.x/) is a different framework, but it can also run unittest-based tests. We have two choices when we write new tests: follow what's already there and add a new `test_foo()` function to the `RepoSyncTest` class, or add a new `test_foo()` function to the bottom of the module (outside of any class). The latter is *not* a unittest-based test and can therefore use all of pytests features (such as parametrized-input or fixtures).

Pytest is very commonly used these days in many Python projects, including in some of the projects we contribute to: Salt and Cobbler. I'd like us to move towards pure-pytest based tests, but if you feel more confident in a unittest-based test, that would be great to have as well!

Further resources:
- https://realpython.com/pytest-python-testing/: shows the advantages that pytest has over unittest and how to use it
- https://realpython.com/python-mock-library/: unittest makes heavy use of mocks, this article explains how to use them

cbbayburt:
Thanks, @agraul for the useful pointers. The PR is ready for another review.

cbbayburt:
I have another PR coming up that patches the same code. This one's on hold for now.

PR #6512, opened 2023-01-21
on download errors try a 2nd time

mcalmer:
## What does this PR change?

Flaky download server produce `[Errno 14] curl#92 - "HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)"`. Give a second try using the same URL.

```
2023/01/21 16:14:39 +02:00 WARNING: Download failed: https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/EL8-Uyuni-Client-Tools/EL_8/x86_64/libsodium-debugsource-1.0.18-2.34.uyuni.x86_6
4.rpm - [Errno 14] curl#92 - "HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)". Trying next mirror...
2023/01/21 16:14:39 +02:00     19/97 : libsodium-devel-1.0.18-2.34.uyuni.x86_64.rpm
2023/01/21 16:14:39 +02:00     20/97 : libsodium-debuginfo-1.0.18-2.34.uyuni.x86_64.rpm
2023/01/21 16:14:40 +02:00     21/97 : golang-github-wrouesnel-postgres_exporter-0.4.7-2.17.uyuni.x86_64.rpm
2023/01/21 16:14:40 +02:00     22/97 : libsodium-static-1.0.18-2.34.uyuni.x86_64.rpm
2023/01/21 16:14:40 +02:00     23/97 : libsodium-debugsource-1.0.18-2.34.uyuni.x86_64.rpm
```

This try to do a re-try via URLGrabber feature

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Partialy workaround https://github.com/SUSE/spacewalk/issues/20009

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/python/spacewalk/satellite_tools/download.py
+++ b/python/spacewalk/satellite_tools/download.py
@@ -201,6 +201,7 @@ class DownloadThread(Thread):
                                    checksum=params['checksum']):
                 return True
 
+        # 14 => HTTPError (https://github.com/rpm-software-management/urlgrabber/blob/1e6d2debe79efdd1ba2f39913dc808723e51a7f7/urlgrabber/grabber.py#L757)
         retrycodes = URLGrabberOptions().retrycodes
         if 14 not in retrycodes:
             retrycodes.append(14)
--- a/python/spacewalk/satellite_tools/download.py
+++ b/python/spacewalk/satellite_tools/download.py
@@ -201,6 +201,10 @@ class DownloadThread(Thread):
                                    checksum=params['checksum']):
                 return True
 
+        retrycodes = URLGrabberOptions().retrycodes
+        if 14 not in retrycodes:
+            retrycodes.append(14)
+
         opts = URLGrabberOptions(
             ssl_ca_cert=params["ssl_ca_cert"],
             ssl_cert=params["ssl_client_cert"],
@@ -211,6 +215,8 @@ class DownloadThread(Thread):
             timeout=params["timeout"],
             minrate=params["minrate"],
             keepalive=True,
+            retry=3,
+            retrycodes=retrycodes,
         )
 
         mirrors = len(params['urls'])
--- a/python/spacewalk/satellite_tools/download.py
+++ b/python/spacewalk/satellite_tools/download.py
@@ -174,13 +174,13 @@ class DownloadThread(Thread):
             # No codes at all or some specified codes
             # 58, 77 - Couple of curl error codes observed in multithreading on RHEL 7 - probably a bug
             if (retrycode is None and code is None) or (retrycode in opts.retrycodes or code in [58, 77]):
-                log2(0, 2, "ERROR: Download failed: %s - %s. Retrying..." % (url, sys.exc_info()[1]),
+                log2(0, 2, "WARNING: Download failed: %s - %s. Retrying..." % (url, sys.exc_info()[1]),
                      stream=sys.stderr)
                 return True
 
         # 14 - HTTP Error
         if retry < (mirrors - 1) and retrycode == 14:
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -1270,6 +1270,10 @@ def set_download_parameters(self, params, relative_path, target_file=None, check
         params["minrate"] = self.minrate
         params['proxies'] = get_proxies(self.proxy_url, self.proxy_user, self.proxy_pass)
 
+        if len(self.repo.urls) == 1:
+            # when we have only 1 url copy it to have a 2nd try
```
meaksh, line 1274: Small suggestion:

```suggestion
            # When the repository has only 1 possible URL, we duplicate it and add it as possible next mirror to try.
            # This will cause reposync does a 2nd try in case of possible spurious errors while downloading files from this repository.
```

meaksh (commented):
Hmmm, I think this is an easy fix that would make "reposync" already more robust in case of these temporary failures that occurs spuriously (without having to make big refactor on the reposync backend).

The current reposync retrying mechanism is based on retrying URLs using a different mirror. This happens automatically, but as mentioned, we only use a single URLs for most of our repos, so in such cases with a single URL, duplicating this URL as a possible next mirror to try sounds reasonable to me.

OTOH I would maybe change the comment to clarify better the motivation of that change.

python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -1270,6 +1270,10 @@ def set_download_parameters(self, params, relative_path, target_file=None, check
...
+            params['urls'].append(self.repo.urls[0])
```
agraul, line 1276: Wouldn't it be nicer to set the "retry" option to 2 instead of duplicating the URL?

agraul: I think we're talking about different "retry" options then. [urlgrabber's default is `None`](https://github.com/rpm-software-management/urlgrabber/blob/master/urlgrabber/grabber.py#L1055), which disabled re-trying.

mcalmer: I think we would need to change also the "retrycodes" as it would not retry on HTTP errors.

mcalmer: please check the new commit if this is what you had in mind

python/spacewalk/satellite_tools/download.py
```diff
@@ -201,6 +201,10 @@ def __fetch_url(self, params):
                                    checksum=params['checksum']):
                 return True
 
+        retrycodes = URLGrabberOptions().retrycodes
```
agraul, line 204: It would be a god idea to document what 14 stands for, e.g.
```suggestion
        # 14 => HTTPError (https://github.com/rpm-software-management/urlgrabber/blob/1e6d2debe79efdd1ba2f39913dc808723e51a7f7/urlgrabber/grabber.py#L757)
        retrycodes = URLGrabberOptions().retrycodes
```

agraul (commented):
Yes, that's what I have in mind. I'm a bit conflicted about having this and the "other" retry mechanism both. It's not much better than the duplicated url. We should schedule refactoring the retry mechanisms after merging this, no matter which of the two we go with. 

Feel free to decide which of the two methods we use for now.

PR #6519, opened 2023-01-24
Include a static reboot needed warning message for transactional systems

wweellddeerr:
## What does this PR change?

It puts an static warning message indicating that reboot is needed after any change in transactional systems (SLE micro).

The message was dynamic according to notifications about reboot needed sent by the `reboot_info` beacon. However the beacon is not being refreshed when registration finishes, being necessary to perform an additional reboot in order to have it working properly. For this reason, PO decided that is better to have a static message.

## GUI diff

Before:

Only when reboot needed is detected in the beacon (no warning is shown when it is not detected):

image

After:

When reboot needed is detected in the beacon:

image

When reboot needed isn't detected in the beacon:

image

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
+++ b/java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
@@ -24890,6 +24890,9 @@ given channel.</source>
       <trans-unit id="overview.jsp.transactionalupdate.reboot" xml:space="preserve">
         <source>There is a pending transaction for this system, please reboot it to activate the changes.</source>
       </trans-unit>
+      <trans-unit id="overview.jsp.transactionalupdate.reboot-static" xml:space="preserve">
+        <source>OS with transactional updates: please reboot after any packages and/or channels changes.</source>
+      </trans-unit>
       <trans-unit id="overview.jsp.traditionalstack.deprecated" xml:space="preserve">
         <source>The traditional stack is unsupported and scheduled for removal. Consider migrating to salt minions.</source>
       </trans-unit>
--- a/java/code/src/com/redhat/rhn/frontend/servlets/SystemDetailsMessageFilter.java
+++ b/java/code/src/com/redhat/rhn/frontend/servlets/SystemDetailsMessageFilter.java
@@ -45,7 +45,7 @@ import javax.servlet.http.HttpServletRequest;
 @WebFilter("/systems/details/*")
 public class SystemDetailsMessageFilter implements Filter {
 
-    private static final String REBOOT_MESSAGE_KEY = "overview.jsp.transactionalupdate.reboot";
+    private static final String REBOOT_MESSAGE_STATIC_KEY = "overview.jsp.transactionalupdate.reboot-static";
     public static final String TRADITIONAL_STACK_MESSAGE_KEY = "overview.jsp.traditionalstack.deprecated";
 
     @Override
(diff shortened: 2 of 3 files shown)


java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
```diff
@@ -24890,6 +24890,9 @@ given channel.</source>
       <trans-unit id="overview.jsp.transactionalupdate.reboot" xml:space="preserve">
         <source>There is a pending transaction for this system, please reboot it to activate the changes.</source>
       </trans-unit>
+      <trans-unit id="overview.jsp.transactionalupdate.reboot-static" xml:space="preserve">
+        <source>OS with transactional updates: please reboot after any packages and/or channels changes.</source>
```
aaannz, line 24894: Channel change? Isn't channel changing just repositories in `/etc/zypp`? That should be fine as it is in overlay and is not creating new snapshot.

Etheryte (approved):

java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
aaannz: This is good question. Technically correct is to suggest reboot when new snapshot is created as without reboot it can lead to data loss.
New snapshot is created by running `transactional-update` or `tukit`. But salt transactional-update executor covers many states and runs it always through `transactional-update` which may or may not create snapshot depending if there was a change in watched subvolumes (e.g. someone uses `file.managed` on file in `/usr/`). We can't tell when new transaction is created if beacon is not running yet.

So either we go with `reboot after any package change` or be paranoid and go with `reboot after any salt state application` or `reboot after doing any change on the system`.

juliogonzalez: Does changing the channels generate a snapshot? I'd expect that's not the case as the changes should happen at /etc/zypp?

About the other alternatives, I don't have a strong opinion. Either we cover too little, wither we cover too much (even with `after any salt state application`).

admd: I get the point Ondrej is making. I would suggest the following to update the message

`OS with transactional updates:Please reboot after any change in order to apply them.`

I would also suggest to not merge this PR till tomorrow at least and see if we could get the fix for the beacon issue.

aaannz: > Does changing the channels generate a snapshot? I'd expect that's not the case as the changes should happen at /etc/zypp?

It should not keep it. When we add channels, we are doing `state.apply channels`. `state.apply` is one of the delegated functions to `transactional_update.apply` ([see](https://github.com/saltstack/salt/blob/master/salt/executors/transactional_update.py#L17)) and `transactional_update.apply` calls `salt-call` through `transactional-update run -d`. `-d` means it will drop the snapshot if no change in snapshot is detected.

So it creates new snapshot, but because modifying repositories modifies only `/etc` in overlay it is not detected as change in snapshot and changes are dropped, however changes in overlay are moved back to real `/etc` (here is the bug in transactional_update and potential data loss if you do not reboot after snapshot is created. It should move data to original snapshot, not currently running one and all gets complicated if users ignore reboot requests).

admd:
@wweellddeerr can you please also add the screenshot of how UI will look like overall with this static message +  with the pending transaction message from beacon, like it is shown here

admd:
Ok I am not sure about this PR anymore. This static message doesn't add much value and I would rather add the note in release notes about requiring reboot twice and see the current message when needed `There is a pending transaction for this system, please reboot to activate the changes` .  Yes, it doesn't cover the initial use case right after bootstrap but that should be fine for now with the release note as known issue.

juliogonzalez:
Closing, as agreed with @admd

PR #6549, opened 2023-01-30
Install the reboot info beacon using a conf file instead of using pillars

wweellddeerr:
## What does this PR change?

The reboot info beacon, used in transactional systems to notify when a reboot is required in a system, was being installed via pillar data. But, it turns out that salt is currently requiring an additional restart in order to have pillar defined beacons running according with defined in pillars (even after running `saltutil.refresh_pillar`).

To avoid this undesired behavior of having to perform a second reboot in the system to have the beacon running, this PR changes the beacon installation to be via a conf file included in the bootstrap state. With this the beacon should be running after the first reboot.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/20278

Related issue created in salt project https://github.com/saltstack/salt/issues/63583

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java
+++ b/java/code/src/com/suse/manager/webui/services/pillar/MinionGeneralPillarGenerator.java
@@ -46,13 +46,10 @@ public class MinionGeneralPillarGenerator implements MinionPillarGenerator {
     public static final String CATEGORY = "general";
 
     private static final int PKGSET_INTERVAL = 5;
-    private static final int REBOOT_INFO_INTERVAL = 10;
 
     private static final Map<String, Object> PKGSET_BEACON_PROPS = new HashMap<>();
-    private static final Map<String, Object> REBOOT_INFO_BEACON_PROPS = new HashMap<>();
     static {
         PKGSET_BEACON_PROPS.put("interval", PKGSET_INTERVAL);
-        REBOOT_INFO_BEACON_PROPS.put("interval", REBOOT_INFO_INTERVAL);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -181,6 +181,14 @@ salt-minion-package:
...
     - require:
       - file: {{ salt_config_dir }}/minion.d/susemanager.conf
+
+{{ salt_config_dir }}/minion.d/beacons.conf:
+  file.managed:
+    - source:
+      - salt://transactional_reboot/reboot_info_beacon.conf
+    - mode: 644
+    - makedirs: True
+
```
meaksh, line 191: You probably want to include a `require` here, to ensure this state needs to run after the salt minion package is installed. The execution order of states is not necessary sequential.

I would suggest to add something this here:
```suggestion
    - require:
      - file: {{ salt_config_dir }}/minion.d/susemanager.conf

```

meaksh (changes requested):
Thanks for the PR. I made a small suggestion

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -181,6 +181,16 @@ salt-minion-package:
     - makedirs: True
     - require:
       - file: {{ salt_config_dir }}/minion.d/susemanager.conf
+
+{{ salt_config_dir }}/minion.d/beacons.conf:
```
meaksh, line 185: I wonder if we should use a custom conf file here, instead of `beacons.conf`. Like:

```suggestion
{{ salt_config_dir }}/minion.d/susemanager_beacons.conf:
```

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -189,6 +189,7 @@ salt-minion-package:
     - makedirs: True
     - require:
       - file: {{ salt_config_dir }}/minion.d/susemanager.conf
+
```
meaksh, line 192: Are we adding this line break here because of any reason?

meaksh (approved):
LGTM - I've only added an small comment

PR #6573, opened 2023-02-02
Bash completion for spacewalk tools

cbbayburt:
Provide bash completion scripts for the following CLI tools:
 - spacewalk-remove-channel
 - spacewalk-common-channels
 - mgr-sync
 - mgr-create-bootstrap-repo
 - spacewalk-repo-sync

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager/susemanager.spec
+++ b/susemanager/susemanager.spec
@@ -162,6 +162,13 @@ BuildRequires:  docbook-utils
 %description tools
 This package contains SUSE Manager tools
 
+%package bash-completion
+Summary:        Bash completion for SUSE Manager CLI tools
+Group:          Productivity/Other
+
+%description bash-completion
+Bash completion for SUSE Manager CLI tools
+
 %prep
 %setup -q
 
@@ -174,6 +181,9 @@ do
     sed -i '1s=^#!/usr/bin/\(python\|env python\)[0-9.]*=#!/usr/bin/python3=' $i;
 done
 %endif
+
+# Bash completion
+%make_build -C bash-completion
 
 %install
 mkdir -p %{buildroot}/%{_prefix}/lib/susemanager/bin/
@@ -229,6 +239,9 @@ rm -f %{buildroot}/%{_prefix}/lib/susemanager/bin/server-migrator.sh
 make -C po install PREFIX=$RPM_BUILD_ROOT
 
 %find_lang susemanager
+
+# Bash completion
+%make_install -C bash-completion
 
 %check
 # we need to build a fake python dir. python did not work with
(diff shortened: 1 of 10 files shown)


mcalmer (approved):
Seems to work, but the completions we have installed by default are written in a different way. I hope that there is no styleguide which will prevent the acceptance in the OBS.

Mabe check the shebang. This "env" construct is often forbidden. I think it is here just a comment and not used. So might be ok.

cbbayburt:
thanks @mcalmer. I'll check out some examples on OBS

susemanager/susemanager.spec
```diff
@@ -161,6 +162,13 @@ BuildRequires:  docbook-utils
 %description tools
 This package contains SUSE Manager tools
 
+%package bash-completion
+Summary:        Bash completion for SUSE Manager CLI tools
```
deneb-alpha, line 166: Are we going to call it in this way also on Uyuni?

cbbayburt: Good point. Do we have a trick to set the product name dynamically in the description/summary? Or do we need to differentiate it between branches?

cbbayburt: Looks like we use "Spacewalk" invariably in most spec files. But the other packages in `susemanager.spec` uses "SUSE Manager" for Uyuni as well.

mcalmer: We do not have different package names for SUMA and Uyuni. Only for the GPG keys we have it different. But there only because we trust different keys.
And as this is a sub-package, we should not make it more complicated. Just use susemanager-bash-completion is totally fine for me

mcalmer (approved):
Looks good to me

python/spacewalk/spacewalk-backend.spec
```diff
@@ -252,6 +252,7 @@ Requires:       %{name}-xml-export-libs
 Requires:       cobbler
 Requires:       python3-requests
 Requires:       python3-rhnlib  >= 2.5.57
+Recommends:     susemanager-bash-completion
```
deneb-alpha, line 255: do we really want to use recommends? 
what about suggest? at the end, spacewalk-backend can work fine without it. personally I  would  prefer to  a less strong requirement where  the user can decide to  install  this or not. 

see for example https://github.com/uyuni-project/uyuni-tools/pull/94

deneb-alpha (changes requested):
not sure if the bash completion should really be a recommended subpackage. I  think we should relax  a bit the  requirement and use suggest.

Last but not least,  are we going to support only bash or  do we want to  cover more shells?

see for example https://github.com/uyuni-project/uyuni-tools/pull/94

python/spacewalk/spacewalk-backend.spec
cbbayburt: AFAIK, it's pretty standard to use `recommends`/`supplements` with accompanying bash-completion packages. Your example `uyuni-tools` also uses `supplements`, not `suggests`. If we go with `suggests`, I'm worried that some users won't be even aware of this functionality.

cbbayburt:
> ... are we going to support only bash or do we want to cover more shells?

Yes, currently only `bash` is supported. Keep in mind that these tools are in Uyuni server. I haven't seen any customer that has `zsh` in their SUMA/uyuni server yet.

python/spacewalk/spacewalk-backend.spec
deneb-alpha: ...but if we go with recommends the bash completion will be mandatory for all the users

cbbayburt: Do you mean `requires` instead of `recommends`? :)

deneb-alpha: ...sorry.I have no  idea what I was thinking about.

deneb-alpha (approved):
LGTM,thanks

PR #6576, opened 2023-02-03
PoC: Push code coverage from each feature into a Redis DB

srbarrios:
## What does this PR change?

This PR add a new functionality to our test framework.
If a flag in the environment variable is enabled, it will produce code coverage in the Java server using Jacoco.

1. It will generate a report once each test feature finish, and flush the cove coverage triggered to have new metrics on next test feature.

2. Also, once a feature finish and we have the code coverage report, we parse the XML report using Nokogiri and we push the results to a Redis database instance that is configurable by environment variables in our controller.

1. Finally, this PR includes a new GitHub Action that will suggest as a comment in all new PRs opened, which are the Cucumber tests that will cover the product changes.

**Note**:
Before merging this PR, I will revert a commit including a change in `java/code/src/com/suse/manager/api/HttpApiResponse.java` that was only commited to test the GitHub action and the comment that you can see bellow this PR.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Documentation might come later

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Internal Code Coverage Jenkins Pipeline:
https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-master-dev-acceptance-tests-code-coverage/

Ports:
- GitHub actions are only supported in Uyuni, so we will only generate code coverage for Uyuni

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/.github/scripts/test_covering_pr.js
+++ b/.github/scripts/test_covering_pr.js
@@ -1,0 +1,30 @@
+const Redis = require('ioredis');
+const fs = require('fs').promises;
+
+const main = async () => {
+    const changedFiles = process.argv.slice(2)
+    const redis = new Redis({
+        host: process.env.REDIS_HOST,
+        port: process.env.REDIS_PORT,
+        username: process.env.REDIS_USER,
+        password: process.env.REDIS_PASS,
+        enableReadyCheck: false
+    });
+    let tests = new Set();
+
+    /**
+     * Loop over all the files changed in the PR and check if it's covered by a test
+     **/
+    for (const filepath of changedFiles) {
+        var classpath = filepath.replace('java/code/src/', '');
+        await redis.smembers(classpath, function(err, test_names) {
+            test_names.forEach(function(test) {
+                tests.add(test);
+            });
+        });
+    }
+
+    console.log('<html><ul><li>%s</ul></html>', Array.from(tests).join('<li>'));
+    process.exit(0);
+}
+main();
--- a/.github/workflows/test-covering-pr.yml
+++ b/.github/workflows/test-covering-pr.yml
@@ -1,0 +1,41 @@
+name: 'Suggested tests to cover this Pull Request'
+
+on:
+  pull_request:
+    paths:
+      - '**.java'
+  pull_request_target:
+    paths:
+      - '**.java'
+
+jobs:
+  tests-covering-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - name: Use Node.js 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: npm i ioredis
+      - id: files
+        uses: Ana06/get-changed-files@v2.2.0
+        with:
+          filter: '*.java'
+      - id: tests_covering_pr
+        name: Tests covering PR
+        env:
+          REDIS_USER: ${{ secrets.REDIS_USER }}
+          REDIS_PASS: ${{ secrets.REDIS_PASS }}
+          REDIS_HOST: ${{ secrets.REDIS_HOST }}
+          REDIS_PORT: ${{ secrets.REDIS_PORT }}
+        run: echo "result=$(node .github/scripts/test_covering_pr.js ${{ steps.files.outputs.added_modified }})" >> $GITHUB_OUTPUT
+      - name: Write a comment into the Pull Request
+        uses: actions-cool/maintain-one-comment@v3
+        with:
+          body: |
+            #### :mag_right: Suggested tests to cover this Pull Request
+            ${{ steps.tests_covering_pr.outputs.result }}
+          body-include: ':mag_right:'
--- a/testsuite/features/support/code_coverage.rb
+++ b/testsuite/features/support/code_coverage.rb
@@ -1,0 +1,76 @@
+# Copyright (c) 2016-2023 SUSE LLC.
+# Licensed under the terms of the MIT license.
+require 'redis'
+require 'nokogiri'
+require 'open-uri'
+
+# CodeCoverage handler to produce, parse and report Code Coverage from the Jave Server to our GitHub PRs
+class CodeCoverage
+  include(Nokogiri::XML)
+  ##
+  # Initialize a connection with a Redis database
+  def initialize(redis_host, redis_port, redis_username, redis_password)
+    @database = Redis.new(host: redis_host, port: redis_port, username: redis_username, password: redis_password)
+  end
+
+  ##
+  # Close the connection with the Redis database
+  def close
+    @database.close
+  end
+
+  ##
+  # Parse a JaCoCo XML report, extracting information that will be included in a Set on a Redis database
+  #
+  # Args:
+  #   feature_name: The feature name of the code coverage report we want to parse and push to Redis
+  def push_feature_coverage(feature_name)
+    filename = "/tmp/jacoco-#{feature_name}.xml"
+    begin
+      xml = File.read(filename, mode: 'r')
+      options = Nokogiri::XML::ParseOptions.new(ParseOptions::HUGE | ParseOptions::RECOVER | ParseOptions::NONET)
+      tree = Nokogiri::XML::Document.parse(xml, nil, nil, options)
+      tree.xpath('.//package').each do |package|
+        package_name = package.attr('name')
+        package.xpath('.//sourcefile').each do |sourcefile|
+          sourcefile_name = sourcefile.attr('name')
+          counter_class = sourcefile.xpath(".//counter[@type='CLASS']")
+          next if counter_class.nil? || counter_class.attr('covered').to_s == ''
+
+          next unless Integer(counter_class.attr('covered').to_s).positive?
+
+          begin
+            @database.sadd("#{package_name}/#{sourcefile_name}", feature_name)
+          rescue StandardError => e
+            warn("#{e.backtrace} > #{package_name}/#{sourcefile_name} : #{feature_name}")
+          end
+        end
+      end
+    rescue StandardError => e
+      warn(e.backtrace)
+    ensure
+      File.delete(filename)
+    end
+  end
+
+  ##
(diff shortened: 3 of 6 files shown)


.github/workflows/test-covering-pr.yml
```diff
@@ -0,0 +1,34 @@
...
+          filter: '*.java'
+      - id: tests_covering_pr
+        name: Tests covering PR
+        run: echo "result=$(node .github/scripts/test_covering_pr.js ${{ steps.files.outputs.added_modified }})" >> $GITHUB_OUTPUT
+      - name: Write a comment into the Pull Request
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            **Suggested tests to cover this Pull Request**:
+            ${{ steps.tests_covering_pr.outputs.result }}
```
nodeg, line 34: A newline is missing at the end.
```suggestion
            ${{ steps.tests_covering_pr.outputs.result }}

```

.github/workflows/test-covering-pr.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  tests-covering-pr:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [ 14.x ]
```
nodeg, line 12: Why using this old version of NodeJS?

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -516,7 +516,7 @@
   raise 'File injection failed' unless return_code.zero?
 end
 
-When(/^I copy "([^"]*)" file from "([^"]*)" to "([^"]*)"$/) do |file_path, from_host, to_host|
+When(/^$/) do |file_path, from_host, to_host|
```
nodeg, line 519: Did you do this by mistake?

testsuite/features/support/code_coverage.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  ##
+  # Close the connection with the Redis database
+  def close
+    @database.close
+  end
+
+  ##
+  # Parse a Jacoco XML report, extracting information that will be included in a Set on a Redis database
```
nodeg, line 23:
```suggestion
  # Parse a JaCoCo XML report, extracting information that will be included in a set on a Redis database
```

testsuite/features/support/code_coverage.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+    rescue StandardError => e
+      warn(e.backtrace)
+    ensure
+      File.delete(filename)
+    end
+  end
+
+  ##
+  # Parse a Jacoco XML report, extracting information that will be included in a Set on a Redis database
```
nodeg, line 57: This is a duplicate. You already have this description above in line 23. Furthermore it does not describe `jacoco_dump()`

testsuite/features/support/code_coverage.rb
```diff
@@ -0,0 +1,75 @@
...
+  #
+  # Args:
+  #   feature_name: The feature name of the code coverage report we want to generate
+  #   html: A boolean to export it as HTML
+  #   xml: A boolean to export it as XML
+  #   source: The path of the class files of the product (Jacoco will provide extra information if we provide it)
```
nodeg, line 63:
```suggestion
  #   source: The path of the class files of the product (JaCoCo will provide extra information if we provide it)
```

testsuite/features/support/env.rb
```diff
@@ -14,12 +14,20 @@
 require 'selenium-webdriver'
 require 'multi_test'
 require 'set'
+require_relative 'code_coverage'
 
 ## code coverage analysis
 # SimpleCov.start
```
nodeg, line 20: I know this is not covered by your PR, but do we still need this? Same above in line 11.

.github/scripts/test_covering_pr.js
```diff
@@ -0,0 +1,36 @@
+const Redis = require('ioredis');
+const fs = require('fs').promises;
+
+const main = async () => {
+    const changedFiles = process.argv.slice(2)
+
+    /**
+     * Connects to a Redis Database
+     * TODO: Move these variables to GitHub secrets of our project, only admins have rights for that.
```
nodeg, line 9: Ist this TODO still for this PR or will you take care of this in the future?

java/code/src/com/suse/manager/api/HttpApiResponse.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 SUSE LLC
+ * Copyright (c) 2023 SUSE LLC
```
nodeg, line 2: Do not forget to revert this before merging.

nodeg (commented):
I left some suggestions.

.github/workflows/test-covering-pr.yml
srbarrios: My only reason is to be consistent with the version used in the other existing GitHub actions.

testsuite/features/step_definitions/command_steps.rb
srbarrios: Wow, yes, it might be commited by mistake, on my latest push force.

testsuite/features/support/code_coverage.rb
srbarrios: Set is a concrete kind of Redis type.
https://redis.io/docs/data-types/sets/

But ok :)

testsuite/features/support/code_coverage.rb
srbarrios: Right. I rushed too much. Good catch.

java/code/src/com/suse/manager/api/HttpApiResponse.java
srbarrios: Yup.

.github/scripts/test_covering_pr.js
ktsamis: Also when these get moved to secrets they need to be changed, the history would still have them.

testsuite/features/support/code_coverage.rb
nodeg: I am ware that you do not mean the verb `set` here but rather the data structure. Nevertheless it is written without capital letter 😉

.github/workflows/test-covering-pr.yml
nodeg: Correct. Thank you for providing the background information. The end of maintenance for Node.js 14 is 2023-04-30. Are there any plans on dropping support for it and update to a more recent version?

.github/scripts/test_covering_pr.js
srbarrios: I would prefer to keep some time with this working, and then later, we move it to GitHub secrets, in any case, it's read-only.
Also, we might also decide to move this Redis Instance somewhere else, for now I created it with my own account in a free tier in AWS.

.github/workflows/test-covering-pr.yml
srbarrios: Keep in mind there is a limit we can reach of minutes in GitHub actions. :)
For now, I will keep that as it is. I prefer we all use the same version.

testsuite/features/support/env.rb
srbarrios: That's different, this is code coverage for the test code itself.
For now, let's keep it there to don't mix concepts.

.github/scripts/test_covering_pr.js
srbarrios: Would it be a big deal to keep this temporary Redis instance, and it's read-only user, in that first attempt?
@renner @juliogonzalez 

If we feel happy with that outcome, we can then proceed to use GitHub secrets and use a more production environment for the Redis DB.

.github/workflows/test-covering-pr.yml
nodeg: I created one in the naica board as a reminder: https://github.com/SUSE/spacewalk/issues/20578

.github/scripts/test_covering_pr.js
juliogonzalez: I am OK with this as something temporary, as it's readonly, and as long as it does not contain confidential data.

But **only** if @renner approves it (as he's our security champion :-))

As for the git history, true, but when we change to something else, I'd not say `Uyuni2023!` is a password I'd consider secure, so we'll need to change it anyway :-D

srbarrios: Some more context:

- I created a completely ephemeral Redis instance in a personal account on [app.redislabs.com](http://app.redislabs.com/)
- It's a free DB, it doesn't cost me anything, it's located in an AWS Free tier, managed by RedisLabs
- We only add content there from one of our internal jenkins pipelines, and the user that have write access it's located in susemanager-ci (there it will be good to have a GitHub Secrets too, in the future)
- In the GitHub actions, we only have a read-only user for Redis, that query content based on the filename of the changes of a PR, the file is the key in that Redis Set, and as Values it returns all the tests that triggers at some point that class
- The only content stored in that Redis DB, is Java class names as keys, and cucumber feature names as values of these keys.
- Every time the pipeline https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-master-dev-acceptance-tests-code-coverage/ runs, it will override the content on that Redis DB, so we don't care if for some reason it dies.

renner: This might be okay for a prototype or during the course of an experiment, but from my perspective is not great to merge it like this for a temporary solution.  At least hostname, port, username and password should be configurable instead of hard-coded and stored as proper secrets, if possible. I would rather do that right away before merging, temporary solutions have a tendency to stay around longer than you intially might think.

srbarrios: OK, changed :)

nodeg (approved):
LGTM. Do not forget to revert the changes to the `.java` file.

renner (approved):

PR #6579, opened 2023-02-06
Specify the required cobbler version at a single place: patterns

juliogonzalez:
## What does this PR change?

Specify the required cobbler version at a single place: patterns, so we have a single place to change when we need a new cobbler version, and so we lower the risk of conflicts with PackageHub, as discussed at https://github.com/SUSE/spacewalk/issues/19497 with Marina, and at Slack with Pablo.

**WARNING: To be merged with the SR https://build.opensuse.org/request/show/1064014!**

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Internal change

- [x] **DONE**

## Test coverage
- No tests: Installation and acceptance tests will test it.

- [x] **DONE**

## Links

Partial fix for https://github.com/SUSE/spacewalk/issues/19497 (will require backporting to SUSE Manager 4.3)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/spacewalk-java.spec
+++ b/java/spacewalk-java.spec
@@ -177,7 +177,7 @@ Requires:       bcel
 Requires:       byte-buddy
 Requires:       c3p0 >= 0.9.1
 Requires:       cglib
-Requires:       cobbler >= 3.3.3
+Requires:       cobbler
 Requires:       concurrent
 Requires:       dwr >= 3
 Requires:       glassfish-jaxb-runtime
@@ -355,7 +355,7 @@ Requires:       bcel
 Requires:       byte-buddy
 Requires:       c3p0 >= 0.9.1
 Requires:       cglib
-Requires:       cobbler >= 3.3.3
+Requires:       cobbler
 Requires:       concurrent
 Requires:       hibernate-commons-annotations
 Requires:       httpcomponents-client
--- a/python/spacewalk/spacewalk-backend.spec
+++ b/python/spacewalk/spacewalk-backend.spec
@@ -253,7 +253,7 @@ Requires:       mod_ssl
 %endif
 Requires:       %{m2crypto}
 Requires:       %{name}-xml-export-libs
-Requires:       cobbler >= 3.3.3
+Requires:       cobbler
 Requires:       python3-requests
 Requires:       python3-rhnlib  >= 2.5.57
 
--- a/spacewalk/package/spacewalk.spec
+++ b/spacewalk/package/spacewalk.spec
@@ -84,7 +84,7 @@ Requires:       susemanager-sls
 
 Obsoletes:      spacewalk-monitoring < 2.3
 
-Requires:       cobbler >= 3
+Requires:       cobbler
 Requires:       susemanager-jsp_en
 
 # weakremover used on SUSE to get rid of orphan packages which are
--- a/spacewalk/setup/spacewalk-setup.spec
+++ b/spacewalk/setup/spacewalk-setup.spec
@@ -89,7 +89,7 @@ BuildRequires:  perl-libwww-perl
 %else
 Requires:       %{sbinpath}/restorecon
 %endif
-Requires(post): cobbler >= 3.3.3
+Requires(post): cobbler
 Requires:       perl-Satcon
 Requires:       spacewalk-admin
 Requires:       spacewalk-backend-tools
--- a/java/spacewalk-java.changes
+++ b/java/spacewalk-java.changes
@@ -1,6 +1,8 @@
 - Improve automatic dependency selection for vendor clones
 - Fix taskomatic logging (bsc#1207867)
 - Added APIs to allow frontend to install and remove ptf
+- Do not specify a cobbler version, as that is now centralized at the
+  patterns
 - Fix not being able to delete CLM environment if there are custom child
   channels that where not built by the environment (bsc#1206932)
 - Makes systems column sortable on relevant patch page, to list by most affected systems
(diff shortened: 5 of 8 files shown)


java/spacewalk-java.spec
```diff
@@ -355,7 +355,7 @@ Requires:       bcel
 Requires:       byte-buddy
 Requires:       c3p0 >= 0.9.1
 Requires:       cglib
-Requires:       cobbler >= 3.3.3
+Requires:       cobbler
```
juliogonzalez, line 358: Server only (`spacewalk-taskomatic`)

java/spacewalk-java.spec
```diff
@@ -177,7 +177,7 @@ Requires:       bcel
...
```
juliogonzalez, line 180: Server only.

python/spacewalk/spacewalk-backend.spec
```diff
@@ -252,7 +252,7 @@ Requires:       mod_ssl
 %endif
 Requires:       %{m2crypto}
 Requires:       %{name}-xml-export-libs
-Requires:       cobbler >= 3.3.3
+Requires:       cobbler
```
juliogonzalez, line 255: Server only (`spacewalk-backend-tools`)

spacewalk/package/spacewalk.spec
```diff
@@ -84,7 +84,7 @@ Requires:       susemanager-sls
 
 Obsoletes:      spacewalk-monitoring < 2.3
 
-Requires:       cobbler >= 3
+Requires:       cobbler
```
juliogonzalez, line 87: Server only (`spacewalk-common`)

spacewalk/setup/spacewalk-setup.spec
```diff
@@ -89,7 +89,7 @@ BuildRequires:  perl-libwww-perl
 %else
 Requires:       %{sbinpath}/restorecon
 %endif
-Requires(post): cobbler >= 3.3.3
+Requires(post): cobbler
```
juliogonzalez, line 92: Server only.

mcalmer (commented):
I wonder if the pattern is the right place for the version.
AFAIK when we containerize we do not have the pattern.
On the other hand in containers we do not really need this dependency as ew require the service and not the package.
Does the RH port also uses pattern?

SchoolGuy (commented):
Thanks for taking the time to tackle this problem. However, I would like to change this PR a bit.

First of all, I would like to check if we even need the `Requires: cobbler` in all these spec files. As SUMA doesn't touch the Cobbler filesystem and vice-versa, we could already move Cobbler to another server to my understanding. Thus for example the Java package has no business requiring it IMHO. The only package really actually needing the CLI is `spacewalk-setup` as it configures the service and settings.

Secondly and lastly: I think the pattern is indeed the right approach. My approach would be to specify the version with `= 3.3.3` again to prevent repository priority mishappenings in Leap etc.

mcalmer:
@SchoolGuy spacewalk-java create and write snippets and it lists existing snippets and show them in the UI?
Did you change all of that requesting it via API instead of looking directly on the filesystem?

SchoolGuy:
@mcalmer It seems you are right that we are using the file system to do that stupidly. The endpoints are present for ages. Then yes we do need Cobbler also for spacewalk-java.

juliogonzalez:
> I wonder if the pattern is the right place for the version. 

Could be. Would then `spacewalk-setup` be a good alternative?

> AFAIK when we containerize we do not have the pattern. On the other hand in containers we do not really need this dependency we require the service and not the package.

What do you mean with this text about the dependency not needed? AFAIK we still require the cobbler RPM :-?

> Does the RH port also uses pattern?

I have no idea, so poing @sbluhm :-)

sbluhm:
> > Does the RH port also uses pattern?
> 
> I have no idea, so poing @sbluhm :-)

I am still being ignorant on the pattern functionality (to me it is just a simple package with dependencies...). My installation instructions of Uyuni is just `dnf install patterns-uyuni_server`. So adding the version requirement there would work for Enterprise Linux.

juliogonzalez:
> > > Does the RH port also uses pattern?
> > 
> > 
> > I have no idea, so poing @sbluhm :-)
> 
> I am still being ignorant on the pattern functionality (to me it is just a simple package with dependencies...). My installation instructions of Uyuni is just `dnf install patterns-uyuni_server`. So adding the version requirement there would work for Enterprise Linux.

That's basically what it is :-)

Thanks for the confirmation, @sbluhm.

So all we need to clarify is about the containers. I will have a look.

juliogonzalez:
> So all we need to clarify is about the containers. I will have a look.

Well, no, I cannot, because the containers we have are for the proxy, which does not require cobbler.

So again my question is for @mcalmer and probably @rjmateus.

If our plan for now is a big container, then I'd assume we'd still install the pattern for the server?

Otherwise, what's your suggestion?

mcalmer:
Right, I think we should look deeper into the container usecase when we are there.
Alternative would be the "spacewalk" package. (spacewalk/package/spacewalk.spec). This is from old spacewalk times and act also as a kind of pattern. 
But I am fine to go with our pattern.

rjmateus:
In the big container we are installing the pattern, so all good. 
When we decided to remove cobbler from big container to its own, then we will need to revisit it.

juliogonzalez:
Very well, I will fix conflicts and merge.

@SchoolGuy @mcalmer @ricardoasmarques are you OK with a backport to `Manager-4.3` but *not* to `Manager-4.2`?

rjmateus:
I'm fine with it Julio, but you ping the wrong Ricardo.

juliogonzalez:
> Thanks for taking the time to tackle this problem. However, I would like to change this PR a bit.
> [...]
> Secondly and lastly: I think the pattern is indeed the right approach. My approach would be to specify the version with `= 3.3.3` again to prevent repository priority mishappenings in Leap etc.

Done: https://build.opensuse.org/request/show/1064014

PR #6587, opened 2023-02-07
Use allocationSize = 1 on all sequences (bsc#1207799)

nadvornik:
## What does this PR change?

Sequence Generator on table `suseImageFile` was created with  default `allocationSize`, which is 50.
The corresponding sequence in SQL is however created with `INCREMENT BY 1`, but hibernate expects
`INCREMENT BY 50`. 
This results in ids created off by -49 and in duplicate key errors after java restart.

The tables `suseImageRepoDigest` and `suseVirtualHostManagerNodeInfo` have the same problem.

This PR sets `allocationSize = 1` everywhere.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: testing this in cucumber would require building multiple images with java restart in between.

## Links

Fixes #
https://bugzilla.suse.com/show_bug.cgi?id=1207799
https://github.com/SUSE/spacewalk/issues/20352

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


aaannz:
I understand what you are writing and reasoning is sound also the error is clearly there, but one thing is not clear to me: why would hibernate by default come up with something prone to these errors? Default postgres increment for sequence is 1. Also how we get to the negative values? I would assume hibernate would take next 50 on start.

Googling a little, setting hibernate allocation size to 1 have some performance impact. It is true that it is no issue for imageFile, but I see now we use this on plenty other places where extra db query may have noticeable impact.

So the question is, is this really how to solve this?

The more I google about it the more it seemed like hibernate folks messed this up for postgresql. 🤷🏾

nadvornik:
As far as I understand it, it is an optimization for creating multiple entries in one transaction. This is not common in SUMA, so we should be OK with `allocationSize = 1`.

nadvornik:
For negative values - hibernate is supposed to get 50 from the sequence and then create ids 50-49, 50-48, etc.
When it gets 1 from the sequence, it creates 1-49, 1-48, etc.

nadvornik:
This might be related to the change of `hibernate.id.new_generator_mappings` in PR https://github.com/uyuni-project/uyuni/pull/5747

aaannz (approved):
Thanks for explaining it to me. 👍🏾 

Just changelog is missing.

cbosdo (approved):
LGTM

PR #6589, opened 2023-02-07
Use transactional_update.reboot to reboot SLE Micro in an Action Chain

mbussolotto:
## What does this PR change?

Now that reboot in SLE Micro works as expected, we can remove the `activate_transaction=true` workaround and run `transactional_update.reboot` outside the main transaction when the action chain contains a reboot command

LIMITATION:
`transactional_update.reboot` does not have the option `at_time` as `system.reboot`: this mean that the reboot is executed immediately, instead of after 3 minutes. This cause some issue on the visualization. Let's take an action chains with:
- package installation
- reboot
- package installation
- reboot

The action chain is executed correctly, but `System History` shows that these execution order (first action in the list is the first executed):
- package installation
- package installation
- reboot
- reboot

This happen because the time between the two reboot is too short.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/20277

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
+++ b/susemanager-utils/susemanager-sls/src/modules/mgractionchains.py
@@ -170,9 +170,12 @@ def start(actionchain_id):
         inside_transaction = os.environ.get("TRANSACTIONAL_UPDATE")
 
     if transactional_update and not inside_transaction:
-        ret = __salt__['transactional_update.sls'](target_sls, queue=True, activate_transaction=reboot_required)
+        ret = __salt__['transactional_update.sls'](target_sls, queue=True, activate_transaction=False)
     else:
         ret = __salt__['state.sls'](target_sls, queue=True)
+
+    if reboot_required :
+        __salt__['transactional_update.reboot']()
 
     if isinstance(ret, list):
         raise CommandExecutionError(ret)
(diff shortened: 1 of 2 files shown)


mcalmer (approved):

mbussolotto:
LIMITATION:
`transactional_update.reboot` does not have the option `at_time` as `system.reboot`: this mean that the reboot is executed immediately, instead of after 3 minutes. This cause some issue on the visualization. Let's take an action chains with:
- package installation
- reboot
- package installation
- reboot

The action chain is executed correctly, but `System History` shows that these execution order (first action in the list is the first executed):
- package installation
- package installation
- reboot
- reboot

This happen because the time between the two reboot is too short.

admd:
> LIMITATION: `transactional_update.reboot` does not have the option `at_time` as `system.reboot`: this mean that the reboot is executed immediately, instead of after 3 minutes. This cause some issue on the visualization. Let's take an action chains with:
> 
> * package installation
> * reboot
> * package installation
> * reboot
> 
> The action chain is executed correctly, but `System History` shows that these execution order (first action in the list is the first executed):
> 
> * package installation
> * package installation
> * reboot
> * reboot
> 
> This happen because the time between the two reboot is too short.

hmm this is bit tricky for user to figure out when looking at event history... I wonder what we can do about it besides documenting it.

mbussolotto:
@admd , right, I'm trying right now to find a solution

mbussolotto:
the function used for mark the reboot as completed is `handleUptimeUpdate` . I'm looking for a way to call it before the fist event after a reboot (or before collecting result from this first action). It should fix the current issue

PR #6617, opened 2023-02-15
Do not execute immediately Package Refresh action for the SSH minion

mackdk:
## What does this PR change?

This PR removes the direct execution of the package refresh action, when collecting the result of an action execution. Before the action was scheduled and then executed immediately and this led to a wrong state in the `rhnaction` and `rhnserveraction` tables. Now the action is only scheduled and it will picked up by taskomatic when appropriate.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/20500

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/spacewalk-java.changes
```diff
@@ -1,3 +1,4 @@
+- Do not execute immediately Package Refresh action for the SSH minion
```
juliogonzalez, line 1: Don't you need a bsc here? @srbarrios mentioned https://github.com/SUSE/spacewalk/issues/20500?

mackdk: Yes, it was created after I wrote the changelog. I will add it.

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
```diff
@@ -2515,10 +2514,8 @@ else if (sa.getStatus().equals(ActionFactory.STATUS_QUEUED)) {
                                 Optional.of(Xor.right(function)), Optional.of(jsonResult))) {
                             LOG.info("Scheduling a package profile update");
 
-                            Action pkgList;
                             try {
-                                pkgList = ActionManager.schedulePackageRefresh(minion.getOrg(), minion);
-                                executeSSHAction(pkgList, minion);
```
srbarrios, line 2521: Out of scope of the PR:

Hi, just to me to understand how things are working, even the SSH Minions, using the scheduleActionExecution will execute the actions on the SSH Minion?
Or this is only for Salt Minion and Traditional clients?

Btw, from Uyuni/Master didn't we removed all about traditional?
I see that here:
https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/manager/action/ActionManager.java#L1171

mackdk: > using the scheduleActionExecution will execute the actions on the SSH Minion?

TL;DR yes. Calling `TaskomatikApi::scheduleActionExecution()` will result in an execution of the `MinionActionExecutor` which in turn (on [line 160](https://github.com/uyuni-project/uyuni/blob/61a6e77c909d84895020a14f0e4d081b9c2d6fe9/java/code/src/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java#L160)) will call `SaltServerActionService::execute()`. This method [splits the minion in regular and ssh](https://github.com/uyuni-project/uyuni/blob/61a6e77c909d84895020a14f0e4d081b9c2d6fe9/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java#L496-L502) and executes the action on both sets. For the latter it will call  `TaskomatikApi::scheduleSSHActionExecution()`

But without knowing the details of all of these, I'm wondering if you have a point and all this back and forth is actually needed. Probably @lucidd knows better.

> Uyuni/Master didn't we removed all about traditional?

Not sure what's the state of this but it seems to me we have multiple places that still refers to traditional stack on master. @mcalmer is this expected to be already removed?

mcalmer: I mostly dropped the client side. Removing all the code from the server side needs much more time and is potential dangreous. There is still a lot of time we can spend in removing traditional stack from Uyuni.

lucidd: > But without knowing the details of all of these, I'm wondering if you have a point and all this back and forth is actually needed. Probably @lucidd knows better.

There is a lot of moving parts and i guess technically not all of that is needed if you already know you are working with salt ssh. But i think for now its better if everything goes though the same codepath for executing actions otherwise we risk the chance of having code introduced in on path but not the other.

lucidd (approved):

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
srbarrios: Thanks for all the info!

PR #6619, opened 2023-02-16
Allow for flexible Pgsql14 installation

sbluhm:
## What does this PR change?

The installation structure of PGSQL14 from the official PostgreSQL repository is different. This repository is being used for the Enterprise Linux implementation.

This change tries to detect the service and data folder names.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
Manually tested installation on Leap 15.4

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacewalk/admin/spacewalk-service
+++ b/spacewalk/admin/spacewalk-service
@@ -70,7 +70,7 @@ start() {
     echo "  Checking DB schema and running DB schema upgrade if needed. This may take a while."
     echo "  Call the following command to see progress: journalctl -f -u uyuni-check-database.service"
     if grep -E -m1 "^db_host[[:space:]]*=[[:space:]]*localhost" /etc/rhn/rhn.conf >/dev/null && \
-       grep "md5" /var/lib/pgsql/data/pg_hba.conf >/dev/null; then
+       grep "md5" `sudo -Hiu postgres env | grep PGDATA | cut -f2- -d=`/pg_hba.conf >/dev/null; then
          echo
          echo "INFO: Migrating password encryption mechanism to scram-sha-256"
     fi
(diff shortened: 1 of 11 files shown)


mcalmer:
I started our testsuite on this PR to see if the standard setup is working.
https://ci.nue.suse.com/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/2316/

@cbosdo please have a look if this influence somehow the container work you do.

SchoolGuy (commented):
I am not a fan of using `sudo` inside a script. I am not aware ad-hoc of a better solution but I am aware that you can supplement the environment of the PostgreSQL user with systemd environment files.

@sbluhm do you have any idea how to get this data without using `sudo`?

mcalmer:
As reference: the testsuite still setup successfully with these changes.

mcalmer (approved):
In principle I am fine. The testsuite worked. If you can find a different way than using sudo would be really good. But for me it would be ok

sbluhm:
I believe I did try to find a better way than `sudo` before but I don't remember if that is really true. So am happy to double check and will revert in a bit.

sbluhm:
ok, the way I understand it, it is not possible to get the data folder without `pgsql` running, unless you know it or grab it from the `postgres` user config.
1. During setup `pgsql` is not running. So we have to get the folder from the user config.
2. When starting the `pgsql` server from `spacewalk-service`, you can't get the data folder as it is not running...

Currently, the only improvement I would see is querying the folder via `sudo` once during setup,  write it into rhn.conf and then use it via context manager. Or instead of querying generally hard coding an Uyuni default location instead of using an OS/Pgsql package default location.

To me this would be an enhancement I would be happy to work on later (atm, I would like to focus on closing this task).

sbluhm:
@mcalmer , thank you. I was not aware of this command and after some playing, it seems to work. Will swap sudo.

PR #6539, opened 2023-01-27
Testsuite: test restart of postgres

nadvornik:
## What does this PR change?

This test restarts postgres service and verifies, that it is handled correctly in java and salt - applying highstate must work.

## Links

This should catch the bugs discussed in  https://github.com/SUSE/spacewalk/issues/20143

- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #6621, opened 2023-02-16
Retry loading of pillars from DB on connection error

nadvornik:
## What does this PR change?

After postgres restart, suma_minion.py module gets empty pillars, because the connection is closed.
This PR detects this situation and tries again.

This was detected by this test: https://github.com/uyuni-project/uyuni/pull/6539

The problem was introduced in PR https://github.com/uyuni-project/uyuni/pull/6033

## GUI diff

No difference.

## Documentation

## Test coverage
- Cucumber tests WIP: https://github.com/uyuni-project/uyuni/pull/6539

- [ ] **DONE**

## Links

Partial fix for https://github.com/SUSE/spacewalk/issues/20143

Testcase PR: https://github.com/uyuni-project/uyuni/pull/6539

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py
+++ b/susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py
@@ -14,7 +14,6 @@ Retrieve SUSE Manager pillar data for a minion_id.
 # Import python libs
 from __future__ import absolute_import
 from enum import Enum
-from contextlib import contextmanager
 import os
 import logging
 import yaml
@@ -87,8 +86,7 @@ def _is_salt_ssh(opts):
     return "__master_opts__" in opts
 
 
-@contextmanager
-def _get_cursor():
+def _get_cursor(func):
     def _connect_db():
         options = {
             "host": "localhost",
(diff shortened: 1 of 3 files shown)


agraul (changes requested):
Thanks for the PR! It would be great if we could keep the retry-logic inside the context manager so that the caller does not require duplicating the access. As it was before this change, the retrying is already quite complicated. Having another round of retries outside of the context manager only adds to it, I believe there is a cleaner way to fix the issue.

susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py
```diff
@@ -132,9 +130,18 @@ def _connect_db():
             log.error("Error on getting database pillar: %s", err.args)
             return
     try:
-        yield cursor
+        func(cursor)
     except psycopg2.DatabaseError as err:
-        log.error("Error on getting database pillar: %s", err.args)
+        log.error("Error on getting database pillar, trying again: %s", err.args)
```
rjmateus, line 135: instead of implementing in this way, should it be put in a loop with the number of retries (in this case 2) and a break as soon we are able to get the func?

nadvornik: ok, I changed the code

susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py
```diff
@@ -131,13 +129,28 @@ def _connect_db():
...
             return
-    try:
-        yield cursor
-    except psycopg2.DatabaseError as err:
-        log.error("Error on getting database pillar: %s", err.args)
-    finally:
-        if _is_salt_ssh(__opts__):
-            cnx.close()
+    retry = 0
+    while True:
```
rjmateus, line 133: I think that instead of while true you could test the number of retries left. Then after the cycle, you can also test if the retry was reached and print an error.
I'm not a big fan of infinite loops :)

rjmateus (commented):
Just one comment from my side. But I would like to ear your opinion and also from python experts

susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py
rjmateus: fine for me :)

nadvornik:
Ok, now it does not use regular expressions at all

cbosdo (approved):
It's good to me but requires a changelog entry

meaksh (approved):
LGTM, thanks! Alex will be back on Monday

PR #6624, opened 2023-02-16
Fix missing module when bootstraping transactional systems

wweellddeerr:
## What does this PR change?

The util.syncstates is not being included when bootstraping transactional systems and it is causing error when the systems has local repositories to be disabled. This PR adds the missing module to fix the problem.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/20349

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
@@ -230,6 +230,9 @@ salt-install-contextvars:
 
 {%- if not transactional %}
 {% include 'bootstrap/remove_traditional_stack.sls' %}
+{% else %}
+include:
+  - util.syncstates
 {%- endif %}
 
 # Manage minion key files in case they are provided in the pillar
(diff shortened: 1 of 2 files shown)


meaksh (approved):
For now I think this is OK, LGTM.

But I think we should start also thinking on how to cleanup this mess we already have with "includes" and SLS-dependencies and prevent this keeps growing in the future.

mcalmer:
@wweellddeerr please run the PR testsuite on this. This is the easiest way to find out if the normal clients are least work.

mcalmer:
> But I think we should start also thinking on how to cleanup this mess we already have with "includes" and SLS-dependencies and prevent this keeps growing in the future.

Agreed any maybe this is easy.

What about adding the include to
- bootstrap/init.sls
- ssh_bootstrap/init.sls
- service/salt-minion.sls

Remove the include from 
- bootstrap/remove_traditional_stack.sls

Btw: in bootstrap/init.sls we have this:
```
{%- if not transactional %}
{% include 'bootstrap/remove_traditional_stack.sls' %}
{%- endif %}
```
but not in ssh_bootstrap/init.sls . So we execute it when we bootstrap for salt ssh?

But maybe this can be done in an extra PR

mcalmer:
@wweellddeerr I agree that more improvements should be done in another PR. This can go in for the next MU

PR #6629, opened 2023-02-17
Added missing make build requirements

sbluhm:
## What does this PR change?

Added "make" as build requirement for some packages.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


cbosdo (approved):
LGTM

mcalmer:
adding @juliogonzalez also for merging this. Maybe he need changelog entries.

juliogonzalez:
> Added "make" as build requirement for some packages.

Well, I read at the PR desc `Added "make" as build requirement for some packages.`, but for the changelog I'd like to have a real reason. Is it because `make` is not preinstalled in some OS such as Alma9?


What I did for other package was adding:
```
- Add make as a BuildRequire for those operating systems that do
  not have it installed by default
```
Does this correctly explain the situation, @sbluhm ?

sbluhm:
@juliogonzalez as far as I understand, `make` is also not installed by default on Leap. It seems that it gets pulled in by some other package. most likely `rpm-build`. This does not seem to be the case for Enterprise Linux 9.
As `make` is a build requirement for either OS, I would not make it a conditional statement as it is a bulid requirement. In my opinion a statement like the following would be more true:
`- Ensure installation of make for building`

Now making my comment long :)
I am still a bit confused on when to add a changelog entry and when not. As far as I understand, this change has no interest to any end user. It doesn't change any functionality. Probably one of the most minimum changes possible.
Based on that, I intentionally skipped the changelog entry. I am happy to add them if required. Just trying to do the right thing :)

juliogonzalez:
> @juliogonzalez as far as I understand, `make` is also not installed by default on Leap. It seems that it gets pulled in by some other package. most likely `rpm-build`. This does not seem to be the case for Enterprise Linux 9. As `make` is a build requirement for either OS, I would not make it a conditional statement as it is a bulid requirement. In my opinion a statement like the following would be more true: `- Ensure installation of make for building`

ACK, I will use that.

> Now making my comment long :) I am still a bit confused on when to add a changelog entry and when not. As far as I understand, this change has no interest to any end user. It doesn't change any functionality. Probably one of the most minimum changes possible. Based on that, I intentionally skipped the changelog entry. I am happy to add them if required. Just trying to do the right thing :)

Well, what you did is usually right. Problem is that some packages usually do not change at all between releases, and in such cases we need a changelog entry so we can tag.

Usually you can figure that out by checking when was the last time we tagged it, or if the last tagging contained a comment about translations only.

And in general, the traditional packages are not maintained anymore, even if we still did not remove then, so those would require a changelog entry for sure (we'll remove it at some point in the future)

juliogonzalez:
In any case, do not add changelogs now, I will handle it.

PR #6632, opened 2023-02-19
get uptime with package profile update

mcalmer:
## What does this PR change?

A Package Profile Update should also report and handle the uptime of the system.
In case the salt start event gets lost, it is currently not possible to update this value with an action.
As the kernel live version and grains are also handled in pkgprofile update, it makes sense to send and update the uptime as well.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were updated

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/pull/20537

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


cbosdo (commented):
LGTM, but I would love to have your PR description in the commit description too. Digging the PR from a commit ID is a real pain so better store it in the git history directly for future use.

mcalmer:
> LGTM, but I would love to have your PR description in the commit description too. Digging the PR from a commit ID is a real pain so better store it in the git history directly for future use.

done

PR #6640, opened 2023-02-21
Testsuite: Remove use of synced Rocky 8 product

elariekerboull:
## What does this PR change?

Following https://github.com/SUSE/spacewalk/issues/18386#issuecomment-1449566326
This PR removes Rocky 8 sync and adapts related features - namely reverting to using an injected repository for openscap.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Cucumber tests were edited
- [x] **DONE**

## Links
Fixes #
Tracks # No ports needed - only on Uyuni
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


testsuite/features/secondary/min_rhlike_openscap_audit.feature
```diff
@@ -13,19 +13,7 @@ Feature: OpenSCAP audit of Red Hat-like Salt minion
...
-    And I check radio button "no-appstream-8-result-RHEL8-Pool for x86_64"
-    And I check "no-appstream-8-result-Custom Channel for Rocky 8 DVD"
-    And I wait until I do not see "Loading..." text
-    And I click on "Next"
-    Then I should see a "Confirm Software Channel Change" text
-    When I click on "Confirm"
-    Then I should see a "Changing the channels has been scheduled." text
-    And I wait until event "Subscribe channels scheduled by admin" is completed
-    When I enable repository "Rocky-BaseOS" on this "rhlike_minion"
+    When I enable the repositories "Rocky-BaseOS Rocky-AppStream" on this "rhlike_minion"
```
srbarrios, line 16: I might be wrong but, how you can have these channels "Rocky-BaseOS Rocky-AppStream" available if we don't trigger a channel sync for those first?
Even if we kill them later, but I think at least, we need to trigger them to have them available. 🤔

srbarrios (commented):
This PR doesn't seems to have a proper title.
This is not about disabling Rocky 8 sync.

But removing the bootstrap of Rocky 8 using RH Expanded Support.

Also, it seems we are moving to Rocky 8 Pure product, to bootstrap it, which can be a good strategy, but we need to confirm it works and announce/agree it with the squad.

testsuite/features/secondary/min_rhlike_openscap_audit.feature
elariekerboull: These are Sumaform injected repositories present by default when we deploy the environment - hence the lack of needed syncing

srbarrios: Ah that make sense!

testsuite/features/reposync/srv_sync_channels.feature
```diff
@@ -48,32 +48,19 @@ Feature: Be able to list available channels and enable them
...
-@susemanager
-  Scenario: Enable RHEL 8 channels for Rocky 8
-    When I execute mgr-sync "add channel rhel8-pool-x86_64"
-    And I execute mgr-sync "add channel res-8-updates-x86_64"
-    And I execute mgr-sync "add channel res-as-8-updates-x86_64"
-    And I execute mgr-sync "add channel res-cb-8-updates-x86_64"
-    And I execute mgr-sync "add channel res8-manager-tools-pool-x86_64"
-    And I execute mgr-sync "add channel res8-manager-tools-updates-x86_64"
+  Scenario: Enable a real base channel
+    When I execute mgr-sync "add channel sles12-sp4-pool-x86_64"
```
nodeg, line 52: Is there a reason to add those SLES12 SP4 channels or were they simply moved and GitHub does not display this correctly?

nodeg (approved):

testsuite/features/reposync/srv_add_rocky8_repositories.feature
```diff
@@ -1,119 +0,0 @@
-# Copyright (c) 2022 SUSE LLC
-# Licensed under the terms of the MIT license.
-
-@rhlike_minion
-Feature: Add the Rocky 8 distribution custom repositories
```
srbarrios, line 5: Let's give it a try.
But I'm not fully sure if we might need to keep the CLM filters. And so to keep using the mounted DVD iso. Even if we don't synchronized anything via network.

testsuite/features/reposync/srv_sync_channels.feature
elariekerboull: Good catch, I don't think this is needed as we don't have it on other branches anymore, and it would sync SLES12 SP4 which we don't especially need
I'll remove that.

PR #6656, opened 2023-02-25
speedup ubuntu errata processing

mcalmer:
## What does this PR change?

Currently Ubuntu Errata Parsing break the testsuite as it takes ~30-60 minutes to run for a fake channel with 8 packages where no of the errata match.

The problem is, that all errata are added to the DB, also when we have no packages for the update.

1. if no packages are found, skip the errata
2. try to create a full package only when at least the package name appears in the channel.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests TBD

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
+++ b/java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
@@ -28,6 +28,7 @@ import com.redhat.rhn.domain.org.Org;
 import com.redhat.rhn.domain.product.Tuple3;
 import com.redhat.rhn.domain.rhnpackage.Package;
 import com.redhat.rhn.domain.rhnpackage.PackageEvr;
+import com.redhat.rhn.domain.rhnpackage.PackageFactory;
 import com.redhat.rhn.manager.content.ContentSyncManager;
 import com.redhat.rhn.manager.content.MgrSyncUtils;
 import com.redhat.rhn.manager.errata.ErrataManager;
@@ -222,6 +223,12 @@ public class UbuntuErrataManager {
                 .filter(c -> c.isTypeDeb() && !c.isCloned())
                 .collect(Collectors.groupingBy(e -> Optional.ofNullable(e.getOrg()), Collectors.toSet()));
 
+        Map<Channel, Set<String>> channelPackageNames = TimeUtils.logTime(LOG, "prepare package name list for channels",
+                () -> orgDebChannels.entrySet().stream()
+                        .flatMap(e -> e.getValue().stream())
+                        .map(c -> Map.entry(c, PackageFactory.listPackageNamesInChannel(c)))
+                        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
+
         List<String> uniqueCVEs = ubuntuErrataInfo.stream()
                 .flatMap(e -> e.getCves().stream().filter(c -> c.startsWith("CVE-")))
                 .distinct()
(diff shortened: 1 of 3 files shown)


java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
```diff
@@ -236,22 +236,38 @@ public static void processUbuntuErrata(Set<Channel> channels, List<Entry> ubuntu
...
+                                                                        packageEvr.getRelease(),
+                                                                        packageEvr.getEpoch(),
+                                                                        packageEvr.getType()).stream()
+                                                        );
+                                            }).collect(Collectors.toSet())
+                            )).entrySet().stream()
+                            .filter(e -> !e.getValue().isEmpty())
+                            .collect(Collectors.toMap(
+                                    e -> e.getKey(), e -> e.getValue())
+                            );
```
mcalmer, line 265: filter all channels which have an empty package set

java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
```diff
@@ -236,22 +236,38 @@ public static void processUbuntuErrata(Set<Channel> channels, List<Entry> ubuntu
...
+                    if (channelPackages.isEmpty()) {
+                        LOG.debug("no packages found - skip errata");
+                        return Stream.empty();
+                    }
```
mcalmer, line 269: If we have no matches, we can skip this errata. No need for further processing

java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
```diff
@@ -236,22 +236,38 @@ public static void processUbuntuErrata(Set<Channel> channels, List<Entry> ubuntu
...
-                        ));
+                            .map(c -> Map.entry(c,
+                                    c.getPackages().stream()
+                                            .map(p -> p.getPackageName().getName())
+                                            .collect(Collectors.toSet())
+                            ))
+                            .collect(Collectors.toMap(
+                                    e -> e.getKey(),
+                                    c -> entry.getPackages().stream()
+                                            .filter(e -> c.getValue().contains(e.getA()))
```
mcalmer, line 247: Only process packages which really appear in this channel

mcalmer:
@cbosdo @lucidd - here is my idea to fix this. Maybe you have other ideas or suggestions for improvements

cbosdo (approved):
LGTM, but I'm not the Ubuntu errata expert around. I'll let @lucidd have the final word

mackdk (approved):

mackdk:
Question, mostly for @lucidd. In the `sync()` method we do:

```java
    public static void sync(Set<Long> channelIds) throws IOException {
        LOG.debug("sync started - get and parse errata");
        List<Entry> ubuntuErrataInfo = parseUbuntuErrata(getUbuntuErrataInfo());
        LOG.debug("get and parse errata finished - process Ubuntu Errata By Id");
        processUbuntuErrataByIds(channelIds, ubuntuErrataInfo);
        LOG.debug("process Ubuntu Errata By Id finished");
    }
```

This means  we will do  `getUbuntuErrataInfo()` and `parseUbuntuErrata()` every time, even when we don't have any debian channels. In particular, `getUbuntuErrataInfo()` downloads (if not in a mirrored configuration) a big file from the ubuntu server, and it takes alone a lot of time. Do we really need to do this when we are processing RPMs channels?

Moreover, can we maybe cache the JSON file locally? In that way we could avoid downloading it multiple times in a row when there are multiple reposync scheduled. We could in theory do a HEAD request and check the `last-modified` header before downloading the full file, to verify if the database has to be updated. We need to double check though if ubuntu is handling the header correctly.

lucidd:
> Question, mostly for @lucidd. In the `sync()` method we do:
> 
> ```java
>     public static void sync(Set<Long> channelIds) throws IOException {
>         LOG.debug("sync started - get and parse errata");
>         List<Entry> ubuntuErrataInfo = parseUbuntuErrata(getUbuntuErrataInfo());
>         LOG.debug("get and parse errata finished - process Ubuntu Errata By Id");
>         processUbuntuErrataByIds(channelIds, ubuntuErrataInfo);
>         LOG.debug("process Ubuntu Errata By Id finished");
>     }
> ```
> 
> This means we will do `getUbuntuErrataInfo()` and `parseUbuntuErrata()` every time, even when we don't have any debian channels. In particular, `getUbuntuErrataInfo()` downloads (if not in a mirrored configuration) a big file from the ubuntu server, and it takes alone a lot of time. Do we really need to do this when we are processing RPMs channels?
> 
> Moreover, can we maybe cache the JSON file locally? In that way we could avoid downloading it multiple times in a row when there are multiple reposync scheduled. We could in theory do a HEAD request and check the `last-modified` header before downloading the full file, to verify if the database has to be updated. We need to double check though if ubuntu is handling the header correctly.

I think the downloading hasn't really been a big issue but it would still make sense to move the deb channel filtering further out to avoid any sort of processing like the download. The idea with caching the file also makes sense but it probably wouldn't be a huge improvement unless a customer has really slow internet.

java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
```diff
@@ -193,11 +194,16 @@ private static Map<String, UbuntuErrataInfo> getUbuntuErrataInfo() throws IOExce
...
-        LOG.debug("sync started - get and parse errata");
-        List<Entry> ubuntuErrataInfo = parseUbuntuErrata(getUbuntuErrataInfo());
-        LOG.debug("get and parse errata finished - process Ubuntu Errata By Id");
-        processUbuntuErrataByIds(channelIds, ubuntuErrataInfo);
-        LOG.debug("process Ubuntu Errata By Id finished");
+        LOG.debug("sync Ubuntu Errata started");
+        Set<Channel> channels = channelIds.stream().map(ChannelFactory::lookupById).collect(Collectors.toSet());
+        if (channels.stream()
+                .filter(c -> c.isTypeDeb() && !c.isCloned())
+                .findFirst().isPresent()) {
```
mackdk, line 201:
```suggestion
        if (channels.stream().anyMatch(c -> c.isTypeDeb() && !c.isCloned())) {
```

Nitpicking, but you can use `anyMatch()` instead of `filter().findFirst().isPresent()`

lucidd:
Ok after doing some test on the old code it looks like its handling this channel in 102 seconds while writing all 4043 errata to the db. I think it would be safer to revert to the previous behavior for now and go from there. There are a lot of potential optimizations we can do to get both time and memory usage down and i think we need to try those out and need some proper testing to see what works best overall. I made an issue (https://github.com/SUSE/spacewalk/issues/20604) writing down all the ideas i've heard the past few days. If anyone has any more ideas let me know and i will add it there for me to try it out.

mcalmer:
Some data about this PR:

```
Case 1: running a Ubuntu channel which has no packages which appear in the ubuntu errata
Total time: 63 seconds
    download ubuntu errata: 11 seconds
    lookup CVEs: 43 seconds
    writing Errata to DB: ~0,2 seconds (none was written)

Case2 : sync Ubuntu2204
Total time: ~21 minutes
    download ubuntu errata: 12 seconds
    lookup CVEs: 36 seconds
    writing Errata to DB: ~20 minutes
```

mcalmer:
We got for a revert first. Closing this PR for now.

PR #6669, opened 2023-03-01
Fix the mgr-inter-sync not creating valid repository metadata when dealing with empty channels

avshiliaev:
## What does this PR change?

Fixes a problem where the Inter-Server-Sync V1 doesn't create any valid meta data for an empty repository after synchronization and thus causing errors on a subscribed system. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/20368

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


avshiliaev:
Here's a couple of code points to get you a bit of context. 
The problem is that for an empty channel there's no taskomatic task to create the repomd.
1. Normally the taskomatic task gets created under [this step here](https://github.com/avshiliaev/uyuni/blob/repositories-without-packages-have-broken-repomd-with-intersync-v1/python/spacewalk/satellite_tools/satsync.py#L382). 
2. [Here we retrieve only the packages](https://github.com/avshiliaev/uyuni/blob/repositories-without-packages-have-broken-repomd-with-intersync-v1/python/spacewalk/satellite_tools/satsync.py#L1661), attach the channel name to them and [pass further down](https://github.com/avshiliaev/uyuni/blob/repositories-without-packages-have-broken-repomd-with-intersync-v1/python/spacewalk/satellite_tools/satsync.py#L1686). No packages - nothing really happens down the chain. There's no other place where the `taskomatic.add_to_repodata_queue` gets triggered in this process. Only in relation to packages. 
3. The call stack leads eventually [here where the taskomatic normally gets called](https://github.com/avshiliaev/uyuni/blob/repositories-without-packages-have-broken-repomd-with-intersync-v1/python/spacewalk/server/importlib/packageImport.py#L150). But as I said above, at this stage we only pass an empty object around, hence the `self.affected_channels` is empty. 
That's why I decided to put the call to taskomatic inside the `_step_packages`, but before the actual packages get imported since at that point there's no info about the channel name.

meaksh (approved):
Let's give this a try, thanks! :+1:

mackdk (approved):

PR #6668, opened 2023-02-28
Fix handling of http 404 (bsc#1208624)

nadvornik:
## What does this PR change?

TFTP server on containerized proxy did not differentiate between missing file and empty file.
Grub tries multiple menu files and uses the first existing. With this bug it ended up with empty menu. 

## GUI diff

No difference.

## Documentation
- No documentation needed: bugfix

## Test coverage
- No tests

## Links

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1208624

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #6670, opened 2023-03-01
Show only cobbler menu entries for given branch

nadvornik:
## What does this PR change?

This PR modifies fbtftp on containerized branch server to filter Cobbler menus on the fly and remove entries relevant to other branches.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/20556

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/containers/proxy-tftpd-image/tftp_wrapper.py
+++ b/containers/proxy-tftpd-image/tftp_wrapper.py
@@ -38,10 +38,13 @@ class HttpResponseData(ResponseData):
         self._request.raise_for_status()
         self._stream = self._request.iter_content(chunk_size=1024)
         self._content = b''
+        self._size = int(self._request.headers['content-length'])
     def read(self, requested):
         data = self._content
         read = len(self._content)
         while read < requested:
+           if self._stream is None:
+              break
            try:
               d = next(self._stream)
               data += d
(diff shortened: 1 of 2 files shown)


SchoolGuy (changes requested):
Could you please elaborate on how a failing PXE boot in a containerized proxy is linked to handling a 404 and showing only menu entries for a given branch? Cobbler has native mechanisms to handle custom menu arrangements and there are discussions on rewriting the tftpsync module for Cobbler to support them for selective syncing to proxies. If you implement this for the containerized proxy exclusively that would be suboptimal...

nadvornik:
404 is unrelated PR https://github.com/uyuni-project/uyuni/pull/6668, I am just testing it together.

Filtering menu entries is what Ondrej suggested it https://github.com/SUSE/spacewalk/issues/20556

aaannz (approved):
👍🏾

containers/proxy-tftpd-image/tftp_wrapper.py
```diff
@@ -80,26 +153,29 @@ def get_response_data(self):
...
             # server local default
-            logging.debug(f"Got request for {path}, forwarding to HTTP")
-            return HttpResponseData(f"https://{target}/tftp/{path}", capath)
+            logging.debug(f"Got request for {path}, filtering HTTP")
+            return HttpResponseDataFilteredPXE(f"https://{target}/tftp/{path}", capath, target)
         elif path.startswith("pxelinux.cfg/"):
             # ignore other pxelinux.cfg files
             logging.debug(f"Got request for {path}, ignoring")
             raise FileNotFoundError()
+        elif path.startswith("grub/x86_64_menu_items.cfg"):
```
SchoolGuy, line 162: This is not enough. There are other architectures as well. Especially in a branch, I would expect at least to have ARM and in the future possibly RISC-V present. Power and S390 probably not, however since a proxy can also mean you just have a second datacenter, I wouldn't eliminate the possibility fully.

containers/proxy-tftpd-image/tftp_wrapper.py
```diff
@@ -55,20 +58,90 @@ def read(self, requested):
...
+        new_content = ''
+        have_entry = False
+        entry = ''
+        in_entry = False
+        for line in self._content.decode('utf-8').splitlines():
+            if in_entry:
+                 entry += line + '\n'
+                 if line.rstrip().endswith('}'):
+                    in_entry = False
+                    if " MASTER=" + self._fqdn in entry:
```
SchoolGuy, line 121: This looks like a generic kernel option. There is nothing preventing a SUMA admin from adding this manually. Thus leading to "swallowed" bootloader entries.

aaannz: This is true. Maybe we can also try to find `MINION_ID_PREFIX` option and only if both are present, then filter it? To limit accidental overreach outside of retail scenario.

containers/proxy-tftpd-image/tftp_wrapper.py
aaannz: Right. For retail x86_64 and ARM are supported.

SchoolGuy: By the name of the folder, I would suspect that this image is not only used for Retail but for all machines behind a containerized proxy?

containers/proxy-tftpd-image/tftp_wrapper.py
aaannz: Hmm.. what kernel options documentation? 😅

containers/proxy-tftpd-image/tftp_wrapper.py
aaannz: Yes, but for not matched entries / architectures there is a catch all which returns unfiltered data. We need to filter only retail supported archs.

SchoolGuy: Okay good. Then consider my inquiry satisfied.

containers/proxy-tftpd-image/tftp_wrapper.py
SchoolGuy: I believe that this is a possible candidate for adding this piece of information:

https://www.uyuni-project.org/uyuni-docs/en/uyuni/specialized-guides/salt/salt-formula-pxe.html#retail.sect.formulas.pxe.kernelparams

nadvornik: PXE formula is unrelated to Cobbler, so this is not a good place.

This is better: https://github.com/uyuni-project/uyuni-docs/blob/manager-4.3/modules/retail/pages/containerized-saltboot.adoc

Anyway, this is documentation for Retail. In non-retail environment, MASTER= does nothing, it is quite unlikely that someone uses it by accident, so I think it is safe.

containers/proxy-tftpd-image/tftp_wrapper.py
nadvornik: OK, adding ARM makes sense.

containers/proxy-tftpd-image/tftp_wrapper.py
SchoolGuy: Although not happy, I will agree with the fact that it is unlikely that this is added manually.

SchoolGuy (approved):
I am still unhappy about the approach taken (as described in the issue). However, the implementation seems to work as desired and is limited to the retail scenario (although the same problem is present for non-retail setups). Let's merge it to have the feature available in principle.

PR #6675, opened 2023-03-02
reposync: catch local file not found urlgrabber error properly (bsc#1208288)

meaksh:
## What does this PR change?

In `urlgrabber` 4.1.0, current version used in Uyuni, there is a small bug [1] on `urlgrab` method that makes the expected "local file not found" exception not to be triggered and therefore producing a different exception in these cases.

Once this small bug is fixed in `urlgrabber`, we need to adapt our current "reposync" code to properly catch the right "local file not found" exception [2] and not produce and error message by default. This is exactly what this PR is doing.

[1] https://github.com/rpm-software-management/urlgrabber/pull/37
[2] https://github.com/rpm-software-management/urlgrabber/blob/master/urlgrabber/grabber.py#L745

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/20616

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


python/test/unit/spacewalk/satellite_tools/test_yum_src.py
```diff
@@ -54,7 +55,7 @@ def _make_dummy_cs(self):
 
         yum_src.get_proxy = Mock(return_value=(None, None, None))
 
-        cs = yum_src.ContentSource("http://example.com", "test_repo", org='')
+        cs = yum_src.ContentSource("http://example.com/fake_path/", "test_repo", org='')
```
ycedres, line 58: Just something I didn't quite understand, why is this needed?

meaksh: I just noticed there was some failures during test execution when using a URL like that, with no real path, as `ContentSource` URL.

There is some code around in reposync and "yum_src" that is apparentely expecting that there is always certain relative path. For example: https://github.com/uyuni-project/uyuni/blame/master/python/spacewalk/satellite_tools/reposync.py#L567

ycedres (approved):

PR #6677, opened 2023-03-02
Create a security policy (SECURITY.md)

renner:
## What does this PR change?

This patch is adding `SECURITY.md` as a security policy for the repository. Please let me know if you think any additional information should be added.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- No documentation needed.

- [X] **DONE**

## Test coverage

- No tests.

- [X] **DONE**

## Links

No links.

- [X] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository.

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,8 +2,8 @@
 
 ## Supported Versions
 
-Uyuni follows a rolling release model, so it is always the next upcoming version that will contain bugfixes for the current one as well as any new features. As there will be no maintenance of specific versions, users will generally need to update to the most recent version in order to receive the latest security updates. For very critical bugfixes and vulnerabilities though in terms of impact and risk, we might decide to release a patch, but this will always be a case by case decision.
+Uyuni follows a rolling release model, so it is always the next upcoming version that will contain bugfixes for the current one as well as any new features. As there is generally no maintenance of specific versions, users will regularly need to update to the most recent version in order to receive the latest security updates. For critical bugfixes (see [SUSE's simplified rating](https://www.suse.com/support/security/rating/)) we will release patches for the most recent released version, while flaws categorized as important, moderate or low will usually be fixed with the next rolling release.
 
 ## Reporting a Vulnerability
 
-Please report any security vulnerabilities by creating a [security advisory](https://github.com/uyuni-project/uyuni/security/advisories/new). We will be reviewing your report and get back to you after evaluation. We will be happy to give you credits for your help with making Uyuni more secure.
+Please report any security vulnerabilities by creating a [security advisory](https://github.com/uyuni-project/uyuni/security/advisories/new). We will be reviewing your report, get back to you after evaluation and happily give you credits for your help with making Uyuni more secure.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,8 +2,8 @@
 
 ## Supported Versions
 
-Uyuni follows a rolling release model, so it is always the next upcoming version that will contain bugfixes for the current one as well as any new features. As there will be no maintenance of specific versions, users will always need to update to the most recent version in order to receive the latest security updates.
+Uyuni follows a rolling release model, so it is always the next upcoming version that will contain bugfixes for the current one as well as any new features. As there will be no maintenance of specific versions, users will generally need to update to the most recent version in order to receive the latest security updates. For very critical bugfixes and vulnerabilities though in terms of impact and risk, we might decide to release a patch, but this will always be a case by case decision.
 
 ## Reporting a Vulnerability
 
-Please report any security vulnerabilities by creating a [security advisory](https://github.com/uyuni-project/uyuni/security/advisories/new). We will be reviewing your report and get back to you after evaluation. We would be happy to give you credits for your help with making Uyuni more secure.
+Please report any security vulnerabilities by creating a [security advisory](https://github.com/uyuni-project/uyuni/security/advisories/new). We will be reviewing your report and get back to you after evaluation. We will be happy to give you credits for your help with making Uyuni more secure.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,9 @@
+# Security Policy
+
+## Supported Versions
+
+Uyuni follows a rolling release model, so it is always the next upcoming version that will contain bugfixes for the current one as well as any new features. As there will be no maintenance of specific versions, users will always need to update to the most recent version in order to receive the latest security updates.
+
+## Reporting a Vulnerability
+
+Please report any security vulnerabilities by creating a [security advisory](https://github.com/uyuni-project/uyuni/security/advisories/new). We will be reviewing your report and get back to you after evaluation. We would be happy to give you credits for your help with making Uyuni more secure.


SECURITY.md
```diff
@@ -0,0 +1,9 @@
+# Security Policy
+
+## Supported Versions
+
+Uyuni follows a rolling release model, so it is always the next upcoming version that will contain bugfixes for the current one as well as any new features. As there will be no maintenance of specific versions, users will always need to update to the most recent version in order to receive the latest security updates.
```
juliogonzalez, line 5: In very specific cases, we release patches, for very critical bugfixes in terms of impact, or for some CVEs, again according to the impact.

That's a case by case decision.

SECURITY.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Reporting a Vulnerability
+
+Please report any security vulnerabilities by creating a [security advisory](https://github.com/uyuni-project/uyuni/security/advisories/new). We will be reviewing your report and get back to you after evaluation. We would be happy to give you credits for your help with making Uyuni more secure.
```
juliogonzalez, line 9:
```suggestion
Please report any security vulnerabilities by creating a [security advisory](https://github.com/uyuni-project/uyuni/security/advisories/new). We will be reviewing your report and get back to you after evaluation. We will be happy to give you credits for your help with making Uyuni more secure.
```

SECURITY.md
agraul: Would it make sense to tie such a decision to concrete [CVSS](https://en.wikipedia.org/wiki/Common_Vulnerability_Scoring_System) rating? Alternatively we could re-use SUSE's simplified rating https://www.suse.com/support/security/rating/ and say e.g. important & critical flaws are fixed by patches, moderate & low flaws are fixed by rolling releases.

The classification still is a case by case decision, just guided by a framework that helps us with the classification and sets expectations straight for our users.

renner: Thanks Alex, I like the idea, especially given the simplified ratings. After talking to Julio though, it seems he would rather decide this case by case for now instead of giving a promise, so I changed the text correspondingly.

SECURITY.md
renner: Thanks, included this suggestion in a `fixup!` commit.

SECURITY.md
```diff
@@ -0,0 +1,9 @@
+# Security Policy
+
+## Supported Versions
+
+Uyuni follows a rolling release model, so it is always the next upcoming version that will contain bugfixes for the current one as well as any new features. As there will be no maintenance of specific versions, users will generally need to update to the most recent version in order to receive the latest security updates. For very critical bugfixes and vulnerabilities though in terms of impact and risk, we might decide to release a patch, but this will always be a case by case decision.
```
juliogonzalez, line 5: > For very critical bugfixes and vulnerabilities though in terms of impact and risk, we might decide to release a patch, but this will always be a case by case decision.

We need to talk about this on Monday. I am not exactly how to explain it, but don't think this is the right wording according to what we discussed today.

juliogonzalez (changes requested):

SECURITY.md
juliogonzalez: What we discussed was a bit different. Let's talk again on Monday, this time including @agraul.

SECURITY.md
renner: Changed the text indicating that there will be patches for critical bugs only, while all others will usually be fixed with the next rolling release.

SECURITY.md
renner: Updated the text after our latest conversation and corresponding with @thesp0nge.

juliogonzalez (approved):
LGTM

thesp0nge (approved):
Maybe we just want to tell that our rolling release is scheduled each month, so we implicitly give a milestone to fix security issues month by month

juliogonzalez:
> Maybe we just want to tell that our rolling release is scheduled each month, so we implicitly give a milestone to fix security issues month by month

That is not the chase. We try to have one release per month, but there's no promise about it. We just had to skip 2023.02 and release 2023.03.

thesp0nge:
> > Maybe we just want to tell that our rolling release is scheduled each month, so we implicitly give a milestone to fix security issues month by month
> 
> That is not the chase. We try to have one release per month, but there's no promise about it. We just had to skip 2023.02 and release 2023.03.

Ok. As I told @renner I'm fine with the rolling security patching. However, I will put a statement to make clear that in case of critical vulnerabilities, a rolling update will be issued no longer than a month later than the last update. Does it make sense to you?

juliogonzalez:
> Ok. As I told @renner I'm fine with the rolling security patching. However, I will put a statement to make clear that in case of critical vulnerabilities, a rolling update will be issued no longer than a month later than the last update. Does it make sense to you?

No.

There are not enough resources to make that commitment in the current situation.

- For the current release -> We release patches
- For the next release -> the patches will be included whenever they happen.

As I said, we aim to have one release per month, but sometimes we can't avoid delaying more. One thing is aiming, another is having a compromise.

agraul:
> > Ok. As I told @renner I'm fine with the rolling security patching. However, I will put a statement to make clear that in case of critical vulnerabilities, a rolling update will be issued no longer than a month later than the last update. Does it make sense to you?
> 
> No.
> 
> There are not enough resources to make that commitment in the current situation.
> 

Critical vulnerabilities, if we use the definition from https://www.suse.com/support/security/rating/, warrant such a commitment IMO. That's as bad as it gets, whatever else do is should not be more urgent than fixing such a vulnerability. For completeness, that's the definition:

> This rating is given to flaws that could be easily exploited by a remote unauthenticated attacker and lead to system compromise (arbitrary code execution) without requiring user interaction. These are the types of vulnerabilities that can be exploited by worms. Flaws that require an authenticated remote user, a local user, or an unlikely configuration are not classed as critical impact.


I believe we can't expect users to trust us, when we don't commit to fixing such issues quickly.

juliogonzalez:
> > > Ok. As I told @renner I'm fine with the rolling security patching. However, I will put a statement to make clear that in case of critical vulnerabilities, a rolling update will be issued no longer than a month later than the last update. Does it make sense to you?
> > 
> > 
> > No.
> > There are not enough resources to make that commitment in the current situation.
> 
> Critical vulnerabilities, if we use the definition from https://www.suse.com/support/security/rating/, warrant such a commitment IMO. That's as bad as it gets, whatever else do is should not be more urgent than fixing such a vulnerability. For completeness, that's the definition:

And even for SUSE Manager, we do patches. We don't release a full new MU with a version bump.

> > This rating is given to flaws that could be easily exploited by a remote unauthenticated attacker and lead to system compromise (arbitrary code execution) without requiring user interaction. These are the types of vulnerabilities that can be exploited by worms. Flaws that require an authenticated remote user, a local user, or an unlikely configuration are not classed as critical impact.
> 
> I believe we can't expect users to trust us, when we don't commit to fixing such issues quickly.

We fix them already, hours after those bugs are disclosed.

Please read the whole thread. 

We fix them as patches that we release to be installed on top of the current released version.

Then the next version (when available) will of course include such patches.

It's how we did it since day 1. We **provide** fixes in hours. That doesn't mean we release a full new release with a version bump.

renner:
> We fix them as patches that we release to be installed on top of the current released version.
> 
> Then the next version (when available) will of course include such patches.

This should be sufficient from my perspective, anyone can install the patch immediately and wait for the next release to happen. But it needs to be obvious enough so that people are aware, especially for new installations, that they are secure only with the additional patch.

PR #4736, opened 2022-01-27
Changed to dynamic users

sbluhm:
## What does this PR change?

- Replaced static Apache Webserver users with central config.
- Updated context manager to allow default context.
- Adjusted some parts to use the context manager.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: **add explanation. This can't be used if there is a GUI diff**

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_oracle"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #6684, opened 2023-03-03
Allow default component for context manager

sbluhm:
## What does this PR change?

Allow default component for context manager in case no component is declared when using the context manager.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/uyuni/common/context_managers.py
+++ b/python/uyuni/common/context_managers.py
@@ -3,7 +3,7 @@ from contextlib import contextmanager
 from spacewalk.common.rhnConfig import CFG, initCFG
 
 @contextmanager
-def cfg_component(component, root=None, filename=None):
+def cfg_component(component=None, root=None, filename=None):
     """Context manager for rhnConfig.
 
     :param comp: The configuration component to use in this context
--- a/python/uyuni/uyuni-common-libs.changes
+++ b/python/uyuni/uyuni-common-libs.changes
@@ -1,3 +1,5 @@
+- Allow default component for context manager.
+
 -------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


meaksh (approved):
LGTM, thanks!

meaksh:
Just realized that there was a previous consideration about this change made by @agraul in this other PR: 

_When I implemented the context manager, I didn't add a default component, but my suggestions in this PR require that. I think it's better to pass None explicitly, but that requires changes in the PR, sorry again._

https://github.com/uyuni-project/uyuni/pull/4736#issuecomment-1036097662

I don't have a much strong opinion here, but I do also understand Alex's point about being explicit when calling `cfg_component(...)` and not allowing a default component.

@agraul @sbluhm feel free to revive this discussion here so we can consider reverting this change if we all agree. Thanks!

sbluhm:
Now you are bringing up my past that I ran away from :P 

After the comment that you mentioned , I updated the code to what it is now.

@agraul replied to that with:
> Ah, I see you added the default parameter yourself. That works as well

But yeah, I follow you and am indifferent about it.

agraul:
It's true, I did say that adding a default also works. I have since regretted that I didn't push back more at the time, I still think it's easier to understand when the component is explicitly named. I'd prefer if we kept the component without a default, but it's not something I will lose sleep over.

sbluhm:
I will happily revert it with the next PR and adapt the next code submission. I don't want to be responsible for any regrets in life choices :)

PR #6706, opened 2023-03-08
Add specific run set for Create clm filter for RH-like minion feature

maximenoel8:
## What does this PR change?

Create a specific run set for the feature `Create the CLM filters for RH-like minions` to support parallel BV.

## GUI diff

No difference.

## GUI diff

No difference.

## Documentation

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

CI PR : https://github.com/SUSE/susemanager-ci/pull/794/commits/36fc1155e2ad9574bafaa8af55fa21644881019d

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
+++ b/testsuite/features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
@@ -1,6 +1,7 @@
 # Copyright 2021-2023 SUSE LLC
 # Licensed under the terms of the MIT license.
 
+@ubuntu1804_minion
 Feature: Add Ubuntu common channels and schedule their synchronization
   In order to use external repositories for Ubuntu that are not in SCC
   As an authorized user
--- a/testsuite/run_sets/build_validation_prepare_clm_filters.yml
+++ b/testsuite/run_sets/build_validation_prepare_clm_filters.yml
@@ -1,0 +1,7 @@
+# This file describes the order of features in a Build Validation test suite run.
+
+## Add clm for rhel BEGIN ###
+
+- features/build_validation/add_custom_repositories/srv_prepare_clm_filters.feature
+
+## Add clm for rhel END ###
--- a/testsuite/run_sets/build_validation_add_common_channels.yml
+++ b/testsuite/run_sets/build_validation_add_common_channels.yml
@@ -3,7 +3,6 @@
 ## Add common channels and special repositories BEGIN ###
 
 - features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
-- features/build_validation/add_custom_repositories/srv_prepare_clm_filters.feature
 - features/build_validation/add_custom_repositories/add_alma9_repositories.feature
 - features/build_validation/add_custom_repositories/add_centos7_repositories.feature
 - features/build_validation/add_custom_repositories/add_liberty9_repositories.feature


testsuite/features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
```diff
@@ -1,6 +1,7 @@
 # Copyright 2021-2023 SUSE LLC
 # Licensed under the terms of the MIT license.
 
+@ubuntu1804_minion
```
srbarrios, line 4: Actually you can let only that tag at feature level and get rid of the tags at scenario level.

testsuite/features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
```diff
@@ -1,6 +1,7 @@
 # Copyright 2021-2023 SUSE LLC
```
srbarrios, line 1: If that's only for ubuntu18, it mighe be better to rename the filename?

testsuite/run_sets/build_validation_prepare_clm_filters.yml
```diff
@@ -0,0 +1,7 @@
+# This file describes the order of features in a Build Validation test suite run.
+
```
srbarrios, line 2: Are you sure this is only for RHEL?
I would say this is for any RH-like system.

testsuite/run_sets/build_validation_prepare_clm_filters.yml
```diff
@@ -0,0 +1,7 @@
...
+## Add clm for rhel BEGIN ###
+
+- features/build_validation/add_custom_repositories/srv_prepare_clm_filters.feature
```
srbarrios, line 5: Also if we mention that this is only for RH like, maybe we should rename the file acording.

srbarrios (approved):
Approving, because it solve the issue.
But take into consideration my comments.

testsuite/features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
Bischoff: It used to be for 16.04 as well.

In any case, I plan a complete renaming of those features. We currently have a mess between "custom channels" and "common channels". I delayed it because I did not want to interfere more with Maxime's work.

testsuite/features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
Bischoff: Same remark, we had 16.04 in same file in the old times.

testsuite/run_sets/build_validation_prepare_clm_filters.yml
Bischoff: it's for all RH-like, yes

testsuite/run_sets/build_validation_prepare_clm_filters.yml
Bischoff: It's on server side, as the `srv_` prefix states it.

Besides, we don't have a prefix for all RH-like. The closest would be `allcli`.

testsuite/run_sets/build_validation_prepare_clm_filters.yml
```diff
@@ -0,0 +1,7 @@
...
+
+## Add clm for rhel END ###
```
Bischoff, line 7: Missing ending linefeed.

Bischoff (changes requested):
small change requested at the end

PR #6712, opened 2023-03-09
Refactor common repositories

maximenoel8:
## What does this PR change?

Refactor add common channel features.
Move create clm filter to core. 

## GUI diff

No difference.

## GUI diff

No difference.

## Documentation

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

CI PR : https://github.com/SUSE/susemanager-ci/pull/794/commits/36fc1155e2ad9574bafaa8af55fa21644881019d

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


testsuite/features/build_validation/add_common_repositories/alma9_minion_add_common_repositories.feature
```diff
@@ -2,7 +2,7 @@
 # Licensed under the terms of the MIT license.
 
 @alma9_minion
-Feature: Add the Alma 9 distribution custom repositories
+Feature: Add the Alma 9 distribution common repositories
   In order to use Alma 9 channels with Appstream modules
   As a SUSE Manager administrator
   I want to filter them out to remove the modules information
```
srbarrios, line 8: I'm reviewing from my mobileapp, so maybe the code is in fact fine, but this line looks bad indented

testsuite/features/build_validation/add_common_repositories/alma9_minion_add_common_repositories.feature
```diff
@@ -2,7 +2,7 @@
 # Licensed under the terms of the MIT license.
 
 @alma9_minion
-Feature: Add the Alma 9 distribution custom repositories
+Feature: Add the Alma 9 distribution common repositories
```
srbarrios, line 5: Those are nor common, but custom channels.
Common channels are those created with spacewalk-common-channels command

testsuite/run_sets/build_validation_add_common_channels.yml
```diff
@@ -1,17 +1,13 @@
...
 
-- features/build_validation/add_custom_repositories/add_ubuntu_repositories.feature
-- features/build_validation/add_custom_repositories/srv_prepare_clm_filters.feature
-- features/build_validation/add_custom_repositories/add_alma9_repositories.feature
-- features/build_validation/add_custom_repositories/add_centos7_repositories.feature
-- features/build_validation/add_custom_repositories/add_liberty9_repositories.feature
-- features/build_validation/add_custom_repositories/add_oracle9_repositories.feature
-- features/build_validation/add_custom_repositories/add_rocky8_repositories.feature
-- features/build_validation/add_custom_repositories/add_rocky9_repositories.feature
+- features/build_validation/add_common_repositories/ubuntu1804_minion_add_common_repositories.feature
```
srbarrios, line 3: No, those are not common, but custom channels that has been CLM filtered initially.
I suggest you a different stage for CLM filters and a different run_set then.

srbarrios (dismissed):
Sorry but I think it was better before :)

testsuite/features/build_validation/add_common_repositories/alma9_minion_add_common_repositories.feature
Bischoff: OK, so this is basically a terminology problem.

I don't really care about the terminology (on the contrary, I would love to see it clarified).

My main concern and the reason I asked this refactoring is to have a bit of order and remove the clutter we have in `testsuite/features/build_validation/add_custom_repositories`, where we mix features from two different pipeline stages: 
```
                stage('Add MUs')
                stage('Add Common Channels')
```
matching those two build parameters:
```
must_add_custom_channels
must_add_common_channels
```
As you can see @srbarrios the name used here is "common channels". If that's improper, fair enough, but then let's fix it everywhere.

testsuite/run_sets/build_validation_add_common_channels.yml
Bischoff: That's indeed a solution. But that would leave only the Ubuntu feature in the "common channels". It would feel a bit alone :smile_cat: .

What about finding a common name for Ubuntu and the CLM-filtered features, given they are in a common bag  in the pipelines and controlled by the same build parameter?

Bischoff (commented):
See my answers to Oscar. OK to be precise on terminology, let's do that at same time we clear this mess.

testsuite/features/build_validation/add_common_repositories/alma9_minion_add_common_repositories.feature
srbarrios: I agree that we need to rename the files to follow the naming conventions.
But it could perfectly go in a different PR, as it solve a different issue.

"Add MUs"
Should contain the creation of custom repositories plus custom channels using as source of information the JSON content with the Maintenance Updates.
Nothing else.

"Add common channels":
Should contain the execution of spacewalk-common-channel command for those channels that we have officially documented to be reposynched that way, instead of using our usual WebUI Product Wizard

"Build CLM projects":
Must be a new stage after "Core" stage, that includes all the CLM projects necessary to be used later on during the creation of AKs for some products, like RH-like products containing Modules repositories that we want to filer.

testsuite/run_sets/build_validation_add_common_channels.yml
srbarrios: I would not re-use the word "common" for different scopes:
SUMA scope
Test scope

That will cause missunderstandings.

Bischoff: Absolutely agree. And it puzzled me for a long time in our pipelines where we have
```
add_custom_channels --> for MUs
add_common channels --> for common channels and CLM
```
That was both obscure and misleading.

testsuite/run_sets/build_validation_add_none_MU_channels.yml
```diff
@@ -0,0 +1,13 @@
+## Add none_MU repositories BEGIN ###
```
Bischoff, line 1: non-MU

the name of the file is also wrong

testsuite/run_sets/build_validation_add_none_MU_channels.yml
```diff
@@ -0,0 +1,13 @@
...
+
+- features/build_validation/add_none_MU_repositories/ubuntu1804_minion_add_none_MU_repositories.feature
```
Bischoff, line 3: non_MU etc.

testsuite/run_sets/build_validation_add_none_MU_channels.yml
```diff
@@ -0,0 +1,13 @@
...
+- features/build_validation/add_none_MU_repositories/alma9_minion_build_clm.feature
```
Bischoff, line 4: Strictly speaking, the CLM projects are for both `minion` and `SSH minion`. But that's OK with `minion` in the name.

Bischoff (changes requested):
In English, `none` means French `aucun` and `non` means French `pas`.

One remark I have not been able to put in the comments:
`build_validation_add_MU_repositories.yml` please also change the contents: `custom` => `MU`

PR #6749, opened 2023-03-17
QE: Include 4.2 in the list of products that use salt-bundle

elariekerboull:
## What does this PR change?

4.2 now uses the salt-bundle in most of its operations. This causes issues on certain features that are salt dependent (e.g [bootstrapping with a script with or without salt-bundle](https://github.com/SUSE/spacewalk/blob/Manager-4.3/testsuite/features/step_definitions/file_management_steps.rb#L50))

This PR adds 4.2 to the list of products to enable salt-bundle use.
## GUI diff

No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links

Fixes #
Tracks # 
4.3
4.2
- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -70,7 +70,7 @@ end
 
 def use_salt_bundle
   # Use venv-salt-minion in Uyuni, or SUMA Head and 4.3
-  $product == 'Uyuni' || %w[head 4.3].include?($product_version)
+  $product == 'Uyuni' || %w[head 4.3 4.2].include?($product_version)
 end
 
 # create salt pillar file in the default pillar_roots location
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -69,7 +69,7 @@ def product_version
 end
 
 def use_salt_bundle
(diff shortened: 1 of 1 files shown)


testsuite/features/support/commonlib.rb
```diff
@@ -70,7 +70,7 @@ def product_version
 
 def use_salt_bundle
   # Use venv-salt-minion in Uyuni, or SUMA Head and 4.3
```
vandabarata, line 72: I'd update this comment to reflect that we now use venv-salt-minion across all our product versions

vandabarata (approved):
lgtm! I'm not sure if we shouldn't just set it to true and be done with it but I guess it's safer to specify each and every version for now :thinking:

elariekerboull:
> lgtm! I'm not sure if we shouldn't just set it to true and be done with it but I guess it's safer to specify each and every version for now thinking

Would be a good idea to discuss this to see if there might be some cases where we might still be needed, if everything is alright we'll be able to switch to using just `true`

PR #6752, opened 2023-03-20
Use docstring instead  of source  comments

### LINKED ISSUE #6006: spacecmd: Add doc strings to functions and modules
Many functions in `spacecmd` don't have a [docstring](https://peps.python.org/pep-0257/#what-is-a-docstring). Some of them have a comment in place of a docstring, which is a good starting point. Still, docstrings have advantages over plain comments since they are "attached" to the function:
```python
>>> def foo():
...     # I'm a comment
...     pass
... 
>>> def bar():
...     """I'm a docstring"""
...     pass
... 
>>> help(foo)
foo()
>>> help(bar)
bar()
    I'm a docstring
```

Whole modules (i.e. Python files that are imported by other Python files) should also be documented with docstrings to explain what they are for and how they are used.

Rakielle:
## What does this PR change?

**The PR   modified the spacecmd/src/spacecmd/utils.py file to use docstring comments instead of source comments**

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation
- No documentation needed: **This PR only works on the docstrings, no code was modified or added**

- [x] **DONE**

## Test coverage
- No tests: **This PR only works on the docstrings, no code was modified or added**
- No tests: already covered

- [x] **DONE**

## Links

Fixes #6006 

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/src/spacecmd/utils.py
+++ b/spacecmd/src/spacecmd/utils.py
@@ -100,14 +100,21 @@ def parse_command_arguments(command_args, argument_parser, glob=True):
         return leftovers, opts
     except IndexError:
         return None, None
-
-
-# check if any named options were passed to the function, and if so,
-# declare that the function is non-interactive
-# note: because we do it this way, default options are not passed into
-# OptionParser, as it would make determining if any options were passed
-# too complex
+ 
 def is_interactive(options):
+    """
+    check if any named options were passed to the function, and if so,
+    declare that the function is non-interactive
+    note: because we do it this way, default options are not passed into
+    OptionParser, as it would make determining if any options were passed
+    too complex
+
+        Parameters:
+            options (SpacecmdArgumentParser): Object containing the options passed.
+
+        Returns:
+            bool: Returns false if named options were passed
+    """
     for key in options.__dict__:
         if options.__dict__[key]:
             return False
@@ -286,9 +293,8 @@ def prompt_user(prompt, noblank=False, multiline=False):
 
     return userinput
 
-
-# parse time input from the user and return xmlrpclib.DateTime
 def parse_time_input(userinput=''):
+    """Parse time input from the user and return xmlrpclib.DateTime"""
     timestamp = None
 
     if userinput == '' or re.match('now', userinput, re.I):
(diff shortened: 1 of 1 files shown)


spacecmd/src/spacecmd/utils.py
```diff
@@ -370,11 +376,21 @@ def parse_time_input(userinput=''):
...
-
-# Compares 2 package objects (dicts) and returns the newest one.
-# If the objects are the same, we return None
 def latest_pkg(pkg1, pkg2, version_key='version',
                release_key='release', epoch_key='epoch'):
+    """
+    Compares 2 package objects (dicts) and returns the newest one.
+
+        Parameters:
+            pkg1 (dict): A dictionary
```
ycedres, line 385: The `pkg1` and `pkg2` parameters are objects different from dictionaries. I think [this](https://github.com/uyuni-project/uyuni/blob/0366d51b3ab3286fd73c18f96d7080e39cefd8b7/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java#L1284) is an example. However, you could just write that pkg1 is an object describing a package.

spacecmd/src/spacecmd/utils.py
```diff
@@ -100,14 +100,21 @@ def parse_command_arguments(command_args, argument_parser, glob=True):
...
+            options (dict): A dictionary
 
-# check if any named options were passed to the function, and if so,
-# declare that the function is non-interactive
-# note: because we do it this way, default options are not passed into
-# OptionParser, as it would make determining if any options were passed
-# too complex
-def is_interactive(options):
+        Returns:
+            bool: Returns false if a key in options has value else returns true otherwise
```
ycedres, line 116: You could add a bit of semantics for the description of returned value (in addition to its type), for example "returns false if named options were passed".

spacecmd/src/spacecmd/utils.py
```diff
@@ -607,6 +621,12 @@ def parse_str(s, type_to=None):
     >>> d = parse_str('{"channelLabel": "foo-i386-5"}')
     >>> assert d["channelLabel"] == 'foo-i386-5'
 
+        Parameters:
+            s (str): A string
+            type_to (func): A function
+
+        Returns:
+
```
ycedres, line 629: With the same idea of adding semantics to the parameters description, here we could say that `s` is the string that's going to be parsed as the type `type_to`.

spacecmd/src/spacecmd/utils.py
```diff
@@ -370,11 +376,21 @@ def parse_time_input(userinput=''):
...
+    """
+    Compares 2 package objects (dicts) and returns the newest one.
+
+        Parameters:
+            pkg1 (dict): A dictionary
+            pkg2 (dict): A dictionary
+            version_key (str): A string
+            release_key (str): A string
+            epoch_key (str): A string
+
```
ycedres, line 390: Here maybe we could add the semantics of the parameters version, release and epoch  (instead of just writing the type).

ycedres (changes requested):

spacecmd/src/spacecmd/utils.py
```diff
@@ -370,11 +376,21 @@ def parse_time_input(userinput=''):
...
+        Returns:
+            pkg (dict): pkg1 or pk2 If the objects are not the same, else returns none
```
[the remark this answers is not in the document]
Rakielle: The function returns one of pkg1, pkg2, or None. The function uses the returned value from `labelCompare` to decide which pkg to return. I modified the message though

spacecmd/src/spacecmd/utils.py
```diff
@@ -370,11 +376,22 @@ def parse_time_input(userinput=''):
...
-# Compares 2 package objects (dicts) and returns the newest one.
-# If the objects are the same, we return None
 def latest_pkg(pkg1, pkg2, version_key='version',
                release_key='release', epoch_key='epoch'):
+    """
+    Compares 2 package objects (dicts) and returns the newest one.
+
+        Parameters:
+            pkg1 (dict): An object describing a package
+            pkg2 (dict): An object describing a package
```
ycedres, line 386:
```suggestion
            pkg1 (dict): A dictionary describing a package
            pkg2 (dict): A dictionary describing a package
```

ycedres (changes requested):
Thanks for  the changes! I suggested some more changes to improve the docstring documentation of the module.

ycedres (approved):
Thank you very much for your contribution!

PR #6767, opened 2023-03-22
QE: Add tests for Kickstart API

vandabarata:
## What does this PR change?
With the latest 4.3.5 release and BV, the create and update methods for `kickstart.tree` have been updated with further parameters. As such, these tests are now automated to ensure stability in the future.

The kickstart tree is RedHat's name for Distribution and these methods basically create or update Distributions in the Systems > Autoinstallation > Distributions menu.
The `create` method added 2 new parameters, kernelOptions and postKernelOptions and it should work with and without them.
The `update` method also added the chance to add these parameters and they've been added to the tests.

I have also renamed the navigation step to find fields according to value, as it uses, not the presented text as most of our steps, but the value of one of its attributes and I wanted to make sure that was clearer.

## Ports
- 4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


Bischoff (approved):

testsuite/documentation/cucumber-steps.md
```diff
@@ -355,10 +355,10 @@ The check box can be identified by name, id or label text.
 
 Note that the text area variant handles the new lines characters while the others don't.
 
-* Make sure a text is in a given input field of a form
+* Make sure a text is in a given input field of a form. The identifier can be the name or id of the element, but it's always the html label, not the shown text.
```
nodeg, line 358:
```suggestion
* Make sure a text is in a given input field of a form. The identifier can be the name or ID of the element, but it's always the HTML label, not the shown text.
```

vandabarata:
Running Uyuni PR just in case - https://ci.suse.de/view/Manager/view/Uyuni-PRs/job/uyuni-prs-ci-tests/2289

testsuite/features/support/namespaces/kickstart.rb
```diff
@@ -66,6 +66,49 @@ def initialize(api_test)
...
+  # Creates a Kickstart tree (Distribution), adding kernel options as parameters.
+  #
+  # Args:
+  #   distro:       The name of the kicktart tree (distribution) you want to create.
+  #   path:         Path to the base or root of the distribution.
+  #   label:        Label of the channel you want to associate with the disribution.
+  #   install:      Label for KickstartInstallType - options can be obtained using kickart.tree's listInstallTypes.
+  #                 Options can be: suse, sles15generic, sles12generic, sles11generic, sles10generic, rhel_9, rhel_8, rhel_7, rhel_6, generic_rpm, fedora18.
+  #   options:      Options to be passed to the kernel when booting for the installation.
+  #   post_options: Options to be passed to the kernel when booting for the installation.
```
nodeg, line 92: Shouldn`t there be a difference in the explanation here? Same for the other method below.

nodeg (approved):
Small nitpicks. Other than that it LGTM. Thank you Vanda!

testsuite/features/support/namespaces/kickstart.rb
vandabarata: it was taken directly from the documentation like that, I'm afraid :grimacing: 
image
but open to ideas

nodeg: I see. It makes sense to just copy paste the official descriptions. Then I guess all good and the API documentation should be adjusted.

PR #6785, opened 2023-03-27
Bump datepicker z-index

Etheryte:
## What does this PR change?

Fix the datetime picker dropdowns appearing behind modals.

## GUI diff

Before:

Screenshot 2023-03-27 at 15 44 54

After:

Screenshot 2023-03-27 at 15 44 34

- [x] **DONE**

## Documentation
- No documentation needed: bugfix.

- [x] **DONE**

## Test coverage
- No tests: UI bugfix.

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/branding/css/susemanager/components/date-time-picker.less
+++ b/web/html/src/branding/css/susemanager/components/date-time-picker.less
@@ -1,6 +1,6 @@
 /* This element already has `position: absolute` */
 .react-datepicker-popper {
-  z-index: 1010 !important;
+  z-index: 1050 !important;
 }
 
 .react-datepicker {
--- a/web/spacewalk-web.changes
+++ b/web/spacewalk-web.changes
@@ -1,3 +1,4 @@
+- Fix datetime picker appearing behind modal edge
 - Refactor Software / Manage / Packages to use SQL paging (bsc#1206725)
 - Fix UI inconsistencies in susemanager-light and susemanager-dark
   theme


cbbayburt (approved):

web/html/src/branding/css/susemanager/components/date-time-picker.less
```diff
@@ -1,6 +1,6 @@
 /* This element already has `position: absolute` */
 .react-datepicker-popper {
-  z-index: 1010 !important;
+  z-index: 1050 !important;
```
srbarrios, line 3: There is any convention on the values used in the z-index?
Why we go from 1 to 1010, to 1050, and so on.
Why we don't use 1,2,3,4,... ?

Etheryte: @srbarrios There's a few issues we're solving here, the main one is that most indexes don't actually come from our code, but instead from third-party dependencies and we have no control over the values they use. So we may have values from one dependency that are 1, 2, 3 and from another dependency that are 1000, 1500, 2000 etc. In our theme we want to mostly not override everything because then things tend to break when we update those dependencies. Hence the odd numbers.

The other problem we're addressing is adding new items to the stacking order after the fact. If we have say an order of 1, 2, 3, 4, 5 and we want to add an item between 2 and 3 we have to go and also change the definitions for 3, 4 and 5. If we instead have values such as 10, 20, 30, 40, 50 and we want to put an item between 20 and 30 we can add a new item with a value of 25 and not change any other code. This allows us to add quick fixes such as this without making big changes and iron out any weird values whenever we do some bigger refactoring.

Hope this answers your question.

PR #6791, opened 2023-03-27
Move build_validation run_set in folder and export cucumber results in results dir

maximenoel8:
## What does this PR change?

Currently, all the run set are at the same level. Add a new level with build_validation folder
Move build_validation run-set into a new build_validation folder.

Currently, html and json results are at testsuite root directory.
Move  html and json output into results directory.

## Documentation

- [ ] **DONE**

## Test coverage

- No tests: already covered

- [ ] **DONE**

## Links

Issue : https://github.com/SUSE/spacewalk/issues/13834

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


Bischoff:
> LGTM. Thank you Maxime! Do not forget to mark the check boxes that are done with this in the issue you linked.

The card was suggesting to put the results for example to `/tmp` or `/var/tmp` . Using `./results` is definitely cleaner, but a `grep -R` will still be cluttered...

So, it goes into the right direction, but it's not what i requested in the card :smile_cat: .

PR #6798, opened 2023-03-29
CI: Acceptance test at PR v2: add secondary parallelizable tests

jordimassaguerpla:
## What does this PR change?

Adds secondary parallelizable tests at the github action for acceptance tests.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed

- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

fixes https://github.com/SUSE/spacewalk/issues/20879

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


jordimassaguerpla:
A screenshot of the secondary parallelizable tests runs:

image

Note that running in parallel it takes less than 30' !! Note we are not running all the tests. Some tests, like openscap ones, have been disabled and I have opened an issue to our project to fix them later, meaning time will increase once we add  more coverage.

Still, this will be much much less than the 8 hours it used to take.

jordimassaguerpla:
I am going to refactor and extract the common code as explained in https://docs.github.com/en/actions/using-workflows/reusing-workflows

jordimassaguerpla:
I have added a new commit that extracts the common code into a reusable workflow https://github.com/uyuni-project/uyuni/pull/6798/commits/1fe1fa336445afca6ed2050bd8309fc2965534f7 .

srbarrios:
@jordimassaguerpla I would suggest to create a folder like "run_sets/github_action" or similar, and we create inside the necessary YAML files.
That way it could be easier to navigate through all those new yaml files.

.github/workflows/test_all_in_one_common.yml
```diff
@@ -0,0 +1,88 @@
...
+env:
+  UYUNI_PROJECT: uyuni-project
+  UYUNI_VERSION: master
+  CUCUMBER_PUBLISH_TOKEN: ${{ secrets.CUCUMBER_PUBLISH_TOKEN }}
+jobs:
+  test-uyuni:
+    runs-on: ubuntu-latest
+    # create a PR to add yourself if you want to be a beta user and
+    # run this action in your Pull Requests
+    if: github.actor == 'jordimassaguerpla' || github.actor == 'elariekerboull'
```
srbarrios, line 17:
```suggestion
    if: github.actor == 'jordimassaguerpla' || github.actor == 'elariekerboull' ||  github.actor == 'srbarrios'
```

srbarrios:
I think that instead of running all the secondary tests, it might be good to combine our github actions at this point.
And use inside this github action the suggested tests exposed by the github action https://github.com/uyuni-project/uyuni/blob/master/.github/workflows/test-covering-pr.yml

.github/workflows/test_all_in_one_p1.yml
```diff
@@ -0,0 +1,7 @@
+name: test-all-in-one-p1
+on: pull_request
+jobs:
+  test-uyuni:
+    uses: ./.github/workflows/test_all_in_one_common.yml
+    with:
+      secondary_tests: "18_run_secondary_parallelizable_tests_1.sh"
```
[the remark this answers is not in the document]
jordimassaguerpla: We have all of the parrallelizable ones ... which is a duplication and should be removed. Good point.

jordimassaguerpla:
> I think that instead of running all the secondary tests, it might be good to combine our github actions at this point. And use inside this github action the suggested tests exposed by the github action https://github.com/uyuni-project/uyuni/blob/master/.github/workflows/test-covering-pr.yml

I would first merge this one given it already adds a lot of value, and then work on combining the two, so we accomplish already a milestone here.

I was thinking that we could run first the suggested test and if it succeeds, then run of all them to ensure there are no regressions. This way we "fail fast" on the expected test... but I am 100% sure we need to run for regressions if we run the one you implemented ... what is your opinion?

jordimassaguerpla:
> @jordimassaguerpla I would suggest to create a folder like "run_sets/github_action" or similar, and we create inside the necessary YAML files. That way it could be easier to navigate through all those new yaml files.

That sounds like a good idea.

I am a bit concern on how to maintain these run_sets in sync with the secondary_parallelizable one. What if I change the Rakefile so running secondary_parallelizable will run all the new sets? Unless we can make some kind of "include" in the parallelizable.yaml file... is it possible?

srbarrios:
> > @jordimassaguerpla I would suggest to create a folder like "run_sets/github_action" or similar, and we create inside the necessary YAML files. That way it could be easier to navigate through all those new yaml files.
> 
> That sounds like a good idea.
> 
> I am a bit concern on how to maintain these run_sets in sync with the secondary_parallelizable one. What if I change the Rakefile so running secondary_parallelizable will run all the new sets? Unless we can make some kind of "include" in the parallelizable.yaml file... is it possible?

I would maybe have a new rake task that split the current yaml file in X yaml files, in a new folder. And then you consume those new yaml files in the pipeline, by iterating over all the files inside the new folder.

jordimassaguerpla:
> I think that instead of running all the secondary tests, it might be good to combine our github actions at this point. And use inside this github action the suggested tests exposed by the github action https://github.com/uyuni-project/uyuni/blob/master/.github/workflows/test-covering-pr.yml

Please vote https://github.com/SUSE/spacewalk/issues/21015

:)

.github/workflows/test_all_in_one_common.yml
```diff
@@ -0,0 +1,90 @@
...
+env:
+  UYUNI_PROJECT: uyuni-project
+  UYUNI_VERSION: master
+  CUCUMBER_PUBLISH_TOKEN: ${{ secrets.CUCUMBER_PUBLISH_TOKEN }}
+jobs:
+  test-uyuni:
+    runs-on: ubuntu-latest
+    # create a PR to add yourself if you want to be a beta user and
+    # run this action in your Pull Requests
+    if: github.actor == 'jordimassaguerpla' || github.actor == 'elariekerboull' ||  github.actor == 'srbarrios'
```
srbarrios, line 17: For now, it's fine.
But long term, it could be better extract this list in a different file.

jordimassaguerpla: It will be removed anyway, given this is only for the beta phase.

.github/workflows/test_all_in_one_common.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: create_ssh_conf
+        run: ./testsuite/podman_runner/04_setup_ssh_controller.sh 
+      - name: install_gems_in_controller
+        run: ./testsuite/podman_runner/05_install_gems_in_controller.sh
+      - name: start-server-container
+        run: ./testsuite/podman_runner/06_start_server.sh 
+      - name: mgr-setup
+        run: ./testsuite/podman_runner/07_manager_setup.sh 
+      - name: build_code
+        run: ./testsuite/podman_runner/08_build_server_code.sh
```
srbarrios, line 61: While developing having these scripts in independent files was good.
But once this works fine, I suggest to group some of them in a unique script.
As we don't need that much detail of the progress, and if so, we can have the progress in log outputs.

jordimassaguerpla: Something like https://github.com/uyuni-project/uyuni/blob/master/testsuite/podman_runner/run ?

.github/workflows/test_all_in_one_common.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: ./testsuite/podman_runner/10_setup_sshd.sh 
+      - name: run_cucumber_core
+        run: ./testsuite/podman_runner/11_run_core_tests.sh 
+      - name: sle-salt-minion
+        run: ./testsuite/podman_runner/12_run_salt_sle_minion.sh 
+      - name: rhlike-minion
+        run: ./testsuite/podman_runner/13_run_salt_rhlike_minion.sh 
+      - name: deblike-minion
+        run: ./testsuite/podman_runner/14_run_salt_deblike_minion.sh 
+      - name: accept_keys
```
srbarrios, line 82: I'm not sure if I understood the meaning of these scripts.
As we do the bootstrap in init_clients, these scripts it's more about "deploying" the instances?
It might need some better naming I guess.

.github/workflows/test_all_in_one_common.yml
jordimassaguerpla: I am calling each step separate because this way in the github action UI you can see all phases separately.

.github/workflows/test_all_in_one_common.yml
jordimassaguerpla: init_clients do not bootstrap but just checks that the minions are there. These scripts use salt cli to bootstrap.

testsuite/podman_runner/18_run_secondary_parallelizable_tests_1.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
srbarrios, line 1: You don't need those 5 scripts with identical content, except the number.
Instead use a unique parameterizable script, where you pass that number.

srbarrios (approved):

testsuite/podman_runner/18_run_secondary_parallelizable_tests_1.sh
jordimassaguerpla: done.

PR #6816, opened 2023-04-04
re-try logout detection

mcalmer:
## What does this PR change?

Let's re-try the detection.
We have often cases where this fail, but the screenshot show the button.
Not nice, but it seems nobody else has an idea what cause this. So let's try to fix it this way.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -547,7 +547,10 @@ end
 
 Then(/^I should be logged in$/) do
   xpath_query = "//a[@href='/rhn/Logout.do']"
-  raise 'User is not logged in' unless find(:xpath, xpath_query)
+  unless find(:xpath, xpath_query)
+    sleep 1
+    raise 'User is not logged in' unless find(:xpath, xpath_query)
+  end
 end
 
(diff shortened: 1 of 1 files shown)


mcalmer:
Here it seems that the page is not fully loaded:
https://ci.suse.de/view/Manager/view/Manager-Head/job/manager-Head-dev-acceptance-tests-NUE/4762/execution/node/3/ws/results/4762/cucumber_report/cucumber_report.html

As we do massive reposync in background this could be the reason.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -547,7 +547,10 @@
 
 Then(/^I should be logged in$/) do
   xpath_query = "//a[@href='/rhn/Logout.do']"
-  raise 'User is not logged in' unless find(:xpath, xpath_query)
+  unless find(:xpath, xpath_query)
```
srbarrios, line 550: But.. that would behaves the same than using timeout parameter with the double of the default capybara timeout. 🤔

PR #6817, opened 2023-04-04
python/spacewalk/satellite_tools/spacewalk-data-fsck: Fix srpm check

apatard:
The function is_orphaned_srpm() is used to check if a src.rpm has a corresponding binary package in the database. It is looking for a binary package having the same name as the src.rpm. For instance a package foo*.src.rpm should produce something like foo*.noarch.rpm. Unfortunately, it's a false idea. For instance, python packages are coming from python-foo*.src.rpm and the resulting package is python3-foo*.noarch.rpm.

To solve the issue:
1 - Try to find the src.rpm in rhnPackageSource table. If not found, the
    file can be deleted
2 - If found, look for packages having the source_rpm_id matching the
    source package.

### DIFF
--- a/python/spacewalk/satellite_tools/spacewalk-data-fsck
+++ b/python/spacewalk/satellite_tools/spacewalk-data-fsck
@@ -59,10 +59,18 @@ def db_init():
 
 
 def src_package_query():
-    query = """select p.id
-                 from rhnPackage p
+    query = """select p.source_rpm_id
+                 from rhnPackageSource p
                 where p.path like :filename"""
     return query
+
+
+def package_find_by_src_id():
+    query = """select id
+                 from rhnPackage p
+                 where source_rpm_id = :id"""
+    return query
+
 
 def package_query(options, bind_path=False):
     query = """select %s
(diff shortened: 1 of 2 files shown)


meaksh (commented):
Thanks a lot for the fix @apatard !

In order to merge this we need to add a changelog entry. We have some notes how you can do this here: https://github.com/uyuni-project/uyuni/wiki/Contributing#changelog-rules-for-packages-in-git

Could you add this changelog file? Thanks!

apatard:
I've rebase and added a line into the changelog file. I was not sure what to write exactly so please complain if the content is not fine.

python/spacewalk/spacewalk-backend.changes.rtp.spacewalk-data-fsck-srpm
```diff
@@ -0,0 +1 @@
+- Fix spacewalk-data-fsck src.rpm handling
```
meaksh, line 1: Thanks for the changes.

Just an small note here: this file should be named `spacewalk-backend.changes.apatard.spacewalk-data-fsck-srpm` so it contains your username as part of the filename.

Other than that, it looks good!

apatard: Ok. I've updated the commit to fix that.

meaksh (approved):
LGTM. Thanks for the contribution!

PR #6829, opened 2023-04-10
Force migration from Salt 3000 clients to Salt Bundle

ycedres:
## What does this PR change?

This is the implementation corresponding to the Uyuni RFC [#00096 ](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00096-automatic-migration-salt-bundle.md)

The intention of this PR is to automatically perform the migration for Salt 3000 clients to Salt Bundle, as Salt 3000 is EOL. This process should be transparent to the user.

The implementation adds a migration state to the highstate that checks the conditions to performe the migration are satisfied.

NOTE: This has been manually tested with an SLE 12 SP5 and a CentOS 7 client in the development environment.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage

- No tests: already covered

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/21145

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
+++ b/susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
@@ -1,0 +1,19 @@
+{%- set avoid_venv_salt_minion = salt['pillar.get']('mgr_avoid_venv_salt_minion') %}
+{%- set is_salt_ssh = salt['pillar.get']('contact_method') in ['ssh-push', 'ssh-push-tunnel'] %}
+{%- set is_saltboot = salt['file.file_exists']('/etc/ImageVersion') %}
+
+{%- if grains['saltversion'] == '3000' and not is_salt_ssh and 
+       not avoid_venv_salt_minion == true and not is_saltboot %}
+
+include:
+  - util.mgr_switch_to_venv_minion
+
+mgr_remove_susemanagerconf:
+  file.absent:
+    - name: /etc/salt/minion.d/susemanager.conf
+    - require:
+      - service: mgr_disable_salt_minion
+    - order: last
+
+{%- endif %}
+
--- a/susemanager-utils/susemanager-sls/modules/tops/mgr_master_tops.py
+++ b/susemanager-utils/susemanager-sls/modules/tops/mgr_master_tops.py
@@ -34,7 +34,8 @@ MANAGER_BASE_TOP = [
     "services.salt-minion",
     "services.docker",
     "services.kiwi-image-server",
-    "ansible"
+    "ansible",
+    "switch_to_bundle.mgr_switch_to_venv_minion"
 ]
 
 
--- a/susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
+++ b/susemanager-utils/susemanager-sls/salt/util/mgr_switch_to_venv_minion.sls
@@ -11,6 +11,9 @@
 {%- set venv_minion_installed = pkgs_installed.get('venv-salt-minion', False) and True %}
 {%- set venv_minion_available = venv_minion_installed or salt['pkg.latest_version']('venv-salt-minion') or False %}
 {%- if venv_minion_available %}
+include:
+  - services.salt-minion
+ 
 mgr_venv_salt_minion_pkg:
   pkg.installed:
     - name: venv-salt-minion
@@ -66,6 +69,7 @@ mgr_disable_salt_minion:
     - enable: False
     - require:
       - service: mgr_enable_venv_salt_minion
+      - sls: services.salt-minion
 
 {%- if salt['pillar.get']('mgr_purge_non_venv_salt') %}
 mgr_purge_non_venv_salt_packages:
--- a/susemanager-utils/susemanager-sls/src/tests/test_mgr_master_tops.py
+++ b/susemanager-utils/susemanager-sls/src/tests/test_mgr_master_tops.py
@@ -24,7 +24,8 @@ TEST_MANAGER_STATIC_TOP = {
         "services.salt-minion",
         "services.docker",
         "services.kiwi-image-server",
-        "ansible"
+        "ansible",
+        "switch_to_bundle.mgr_switch_to_venv_minion"
     ]
 }
 
--- a/susemanager-utils/susemanager-sls/susemanager-sls.changes
+++ b/susemanager-utils/susemanager-sls/susemanager-sls.changes
@@ -1,3 +1,4 @@
+- Include automatic migration from Salt 3000 to Salt Bundle in highstate
 - fix duplicate packages in state
 - Fix enabling bundle build via custom info
 - Rename internal state 'synccustomall' to 'syncall'


susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,129 @@
+{%- if grains.saltversion == '3000' and not grains.get('is_bundle') -%}
```
meaksh, line 1: We would need to exclude SSH minions here, as there might be Salt 3000 SSH minions (py27-compat-salt) still in 4.2, as it does not yet use Salt Bundle to drive Salt SSH actions.

meaksh (commented):
Added a comment about Salt SSH, we need to make the migration to not run in case of SSH minions running with Salt 3000.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
meaksh: Yeah, it should, as `py27-compat-salt` is actually Salt 3000.3.

But, in order to exclude SSH minions from this, I think we should use:

```jinja
{%- if salt['pillar.get']('contact_method') not in ['ssh-push', 'ssh-push-tunnel'] %}
...
{% endif %}
```

As we generally do for excluding SSH minions in other SLS files, like: https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/services/salt-minion.sls#L10

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,93 @@
+{%- if grains['saltversion'] == '3000' and salt['pillar.get']('contact_method') not in ['ssh-push', 'ssh-push-tunnel'] and 
+       (not salt['pillar.get']('custom_info:mgr_avoid_venv_salt_minion') or salt['pillar.get']('custom_info:mgr_avoid_venv_salt_minion') == 'False') %}
```
agraul, line 2: Can we simplify this condition a bit? It's hard to follow. In the end, I think we just want to abort if `custom_info:mgr_avoid_venv_salt_minion` is True, right? Is `not salt['pillar.get']('custom_info:mgr_avoid_venv_salt_minion', false) == true` not enough?

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
agraul, line 1: What's the relationship between this new state and the existing states? I thought we would re-use what we already have, was that not possible?

agraul (changes requested):
I didn't look very closely at the SLS yet, but I already have two questions.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
ycedres: Changed. However I believe that the `mgr_avoid_venv_salt_minion` pillar is the string "True" in this case instead of the `true` literal.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
ycedres: It's of course very similar to the already existing `util.mgr_switch_to_venv_minion` state but `services.salt-minion` is a requisite for `mgr_disable_salt_minion` in order to avoid the `salt-minion` service to be started again.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
agraul: That's possible, I haven't tried the condition - just read and tried to simplify it. Have you tested it after changing?

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
agraul: I'm a bit lost. Which of the two SLS files needs a requisite: `switch_to_bundle.mgr_switch_to_venv_minion` or `util.mgr_switch_to_venv_minion`? 

I would prefer these to be a single SLS for two reasons: 
1. Duplication. These are very similar, which can cause confusions for users, supporters, and might add overhead when we need to change one or the other (or both)
2. The use the same state ids, and are therefore incompatible. That's not nice, even if we don't expect anyone to add `util.mgr_switch_to_venv_minion` to their top.sls. 

Can't we use a single SLS file, and if there needs to be a slight deviation, use jinja or an [extend declaration](https://docs.saltproject.io/en/master/ref/states/highstate.html#extend-declaration) to add/remove/change the few things that need to be changed?

ycedres: Sorry for the confusion. I'm not using `util.mgr_switch_to_venv_minion` at all. The new state `switch_to_bundle.mgr_switch_to_venv_minion` is the one needing that require for `services.salt-minion`. Nevertheless, I'll have a look at extend declaration as you mention.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,93 @@
+{%- if grains['saltversion'] == '3000' and salt['pillar.get']('contact_method') not in ['ssh-push', 'ssh-push-tunnel'] and 
+       (not salt['pillar.get']('custom_info:mgr_avoid_venv_salt_minion') == 'True') %}
```
aaannz, line 2: Sorry to jump here so late in the game, I missed the rfc process.

There is however a problem with saltboot retail terminals. Saltboot, particularly dracut-saltboot, was released with support for salt bundle around October last year.

Initrd with saltboot is tied with deployed images on the terminal and we know our users are sometimes too cautious with deploying new image versions, so it is very well possible that some of our users are still using images from before saltboot dracut module understood salt bundle.

When this terminal is migrated to bundle, on next reboot dracut saltboot module will now recognize existing configuration and will try to register terminal as a new one.

Unfortunately it is not easy to determine is used initrd understands bundle or not, so I think for time being we should not migrate terminals to the bundle automatically (they are migrated automatically with new images).

To detect if minion is a terminal, look for `/etc/ImageVersion` file. If it is present, it is a retail terminal and we should IMO skip automigrating it.

aaannz: @nadvornik any thoughts?

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
ycedres: I included `util.mgr_switch_to_venv_minion` and I overwrote the `mgr_disable_salt_minion` state. This is intended to require`services.salt-minion`.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,93 @@
...
+{%- if managed_minion or venv_managed_minion %}
+{%- set pkgs_installed = salt['pkg.info_installed']() %}
+{%- set venv_minion_installed = pkgs_installed.get('venv-salt-minion', False) and True %}
+{%- set venv_minion_available = venv_minion_installed or salt['pkg.latest_version']('venv-salt-minion') or False %}
+{%- if venv_minion_available %}
+mgr_venv_salt_minion_pkg:
+  pkg.installed:
+    - name: venv-salt-minion
+    - onlyif:
+      - ([ {{ venv_minion_installed }} = "False" ])
```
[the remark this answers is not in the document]
ycedres: Right now I changed the approach to reuse and not duplicate things from the existing `util.mgr_switch_to_venv_minion`. However, that state had this check included already. I'm not sure if there's a better way to implement this state, as you mention. I'll have a look.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
ycedres: I included the check for `/etc/ImageVersion`.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,25 @@
...
+       not avoid_venv_salt_minion == 'True' and not is_saltboot %}
+
+include:
+  - util.mgr_switch_to_venv_minion
+
+extend:
+  mgr_disable_salt_minion:
+    service.dead:
+      - require:
+        - sls: services.salt-minion
```
agraul, line 15: I wonder why this is needed here, but not in the original state. Could this be a bug in the original state?

ycedres: With `state.show_highstate` this is what I get for `mgr_salt_minion_run` that is part of `services.salt-minion.sls`:
```
mgr_salt_minion_run:
        ----------
        __env__:
            base
        __sls__:
            services.salt-minion
        service:
            |_
              ----------
              name:
                  salt-minion
            |_
              ----------
              enable:
                  True
            |_
              ----------
              order:
                  last
            - running
  ```
So I assume that this is the service that's going to be enabled (in happens in the practice, the `salt-minion` service is run again after being disabled by the migration state and that's why I needed the `require`).

In `salt-minon.sls` we have:
```
{%- if salt['pkg.version']('venv-salt-minion') %}
{%- set salt_minion_name = 'venv-salt-minion' %}
{%- set susemanager_minion_config = '/etc/venv-salt-minion/minion.d/susemanager.conf' %}
{%- endif -%}
```
But `venv-salt-minion` is not installed at the time of the highstate rendering.

File: susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls (hunk 2 above)
Comment (ycedres) [reply susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls]: I tried it and it's working right now.

File: susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls (hunk 3 above)
Comment (nadvornik) [reply susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls]: I think it could work with old images by accident - this PR deletes only `susemanager.conf` and the old retail images
put `master` entry in different file, so the old configuration is preserved and dracut saltboot continues to work.

But I agree that checking `/etc/ImageVersion` is better.

File: susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls (hunk 5 above)
Comment (agraul) [reply susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls:15]: Just to check that I understand it correctly, `salt-minion.sls` needs to be overridden by the migration because it would be applied later in the highstate and "revert" part of the migration?

If that's the case, I guess this would also happen if someone added the "original migration sls" to a highstate (which causes the inclusion of `salt-minion.sls`, something that does not happen when running the migration as a "single action"). Wouldn't it then make sense to have this `require` in the original migration sls to ensure the disabling happens after `salt-minion.sls` is applied?

File: susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls (hunk 3 above)
Comment (agraul) [reply susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls]: I don't quite get the part about having `master:` in a different file @nadvornik, what change is still needed? 
As far as I understood, we want to abort saltbooted clients. I therefore think it's enough to remove `/etc/salt/minion.d/susemanager.conf`.

File: susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls (hunk 5 above)
Comment (ycedres) [reply susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls:15]: I believe that if someone added the original migration sls to the highstate that would also happen. Could be a good idea what you propose.

Comment (ycedres) [reply susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls]: I added the `require` in the original `util.mgr_switch_to_venv_minion`. I tried it and works.

**Comment by ycedres** [comment]
@meaksh @agraul Do you have any other comments on this?

File: susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls [hunk 6]
```diff
@@ -0,0 +1,19 @@
+{%- set avoid_venv_salt_minion = salt['pillar.get']('custom_info:mgr_avoid_venv_salt_minion') %}
```
meaksh, line 1: why `custom_info:mgr_avoid_venv_salt_minion` here and not `mgr_avoid_venv_salt_minion`?

In our SUSE Manager documentation we mention `mgr_venv_salt_minion` as the pillar to control this: https://documentation.suse.com/suma/4.3/en/suse-manager/client-configuration/contact-methods-saltbundle.html

To define this pillar, customer should go to `/srv/pillar/`, define `top.sls` and define this pillar manually.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,19 @@
...
+{%- set is_salt_ssh = salt['pillar.get']('contact_method') in ['ssh-push', 'ssh-push-tunnel'] %}
+{%- set is_saltboot = salt['file.file_exists']('/etc/ImageVersion') %}
+
+{%- if grains['saltversion'] == '3000' and not is_salt_ssh and 
+       not avoid_venv_salt_minion == 'True' and not is_saltboot %}
```
meaksh, line 6: Shouldn't this be like:

```suggestion
       not avoid_venv_salt_minion == True and not is_saltboot %}
```

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
ycedres: Changed. I used `custom_info:mgr_avoid_venv_salt_minion` because I was thinking that setting the `custom_info:mgr_avoid_venv_salt_minion` from the UI was the proper way to avoid the automigration from an already existing Salt 3000 client.  But if the client was bootstrapped with `mgr_venv_salt_minion` it is reasonable to use this pillar to control also this, as you mention.

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
ycedres: Done (I just used the lowercase version of the `true` literal in Jinja).

susemanager-utils/susemanager-sls/salt/switch_to_bundle/mgr_switch_to_venv_minion.sls
```diff
@@ -0,0 +1,19 @@
+{%- set avoid_venv_salt_minion = salt['pillar.get']('mgr_avoid_venv_salt_minion') %}
+{%- set is_salt_ssh = salt['pillar.get']('contact_method') in ['ssh-push', 'ssh-push-tunnel'] %}
+{%- set is_saltboot = salt['file.file_exists']('/etc/ImageVersion') %}
+
+{%- if grains['saltversion'] == '3000' and not is_salt_ssh and 
+       not avoid_venv_salt_minion == true and not is_saltboot %}
```
meaksh, line 6: Is there any particular reason to use `true` here instead of `True`? 

I think both works fine as jinja2 evaluates both as `True`, but I would prefer a more Pythonic `True` here.

meaksh: Sorry, I missed your comment mentioning that you explicitely use `true` Jinja2 literal. I'm fine with it.

meaksh (commented):
Looks good to me. Thanks! :+1:

meaksh (commented):
Looks good to me. Thanks! :+1:

meaksh (approved):

agraul (approved):

PR #6844, opened 2023-04-12
QE: Diverge product sync for SLES15 SP4

srbarrios:
## What does this PR change?

We caught the issue regarding the autoselection of the Server Applications Module.
The root cause was in the previous feature, where we are synchronizing via mgr-sync command the channel of the SUMA client tools. That action, automatically synchronize not only the client tools but all the inverted tree of dependencies for that channel, synchronizing its parents. So, when we tried to select the product in the next feature, in fact the selection was disabled, and so it prevented the automatically selection of the unrelated sub-product of SLES15 SP4 on regard of SUMA client tools, but at same time, it was one of the recommended products and so the scenario failed.

To fix it, we removed the scenario synchronizing the client tools in the previous feature, and rely only in the synchronization made on the feature adding SLES 15 SP4 product and its recommended sub-products.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Port:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/reposync/srv_sync_channels.feature
+++ b/testsuite/features/reposync/srv_sync_channels.feature
@@ -48,11 +48,6 @@ Feature: Be able to list available channels and enable them
     And I should see "Total time:" in the output
     And I should see "Repo URL:" in the output
 
-@susemanager
-  Scenario: Enable SLES15 SP4 SUMA client tools for creating bootstrap repositories
-    When I execute mgr-sync "add channel sle-manager-tools15-pool-x86_64-sp4"
-    And I execute mgr-sync "add channel sle-manager-tools15-updates-x86_64-sp4"
-
   Scenario: Let mgr-sync time out
     When I remove the mgr-sync cache file
     And I execute mgr-sync refresh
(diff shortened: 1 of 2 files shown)


vandabarata (approved):

testsuite/features/reposync/srv_sync_channels.feature
```diff
@@ -48,11 +48,6 @@ Feature: Be able to list available channels and enable them
     And I should see "Total time:" in the output
     And I should see "Repo URL:" in the output
 
-@susemanager
-  Scenario: Enable SLES15 SP4 SUMA client tools for creating bootstrap repositories
```
Bischoff, line 52: we write elsewhere `SLES 15 SP4`

Bischoff (approved):
just a nitpick please

testsuite/features/reposync/srv_sync_channels.feature
srbarrios: Ok, but as you can see this is code we are removing.

Bischoff: "oops"

PR #6882, opened 2023-04-14
Ignore mandatory channels results that don't match list of channels

parlt91:
## What does this PR change?

When clicking  on a CLM projects `Attach/Detach Sources` prevent the popup from loading forever by ignoring mandatory channel results that don't match the list of channels.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: **Bugfix**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/19250

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/manager/content-management/shared/components/panels/sources/channels/channels-api-transforms.ts
+++ b/web/html/src/manager/content-management/shared/components/panels/sources/channels/channels-api-transforms.ts
@@ -24,10 +24,12 @@ export const toCanonicalRequires = (
     if (isNaN(channelId) || !channel) {
       throw new RangeError("Invalid channel id");
     }
-
     const requiredChannels = rawRequiresMap[channelIdString]
-      ?.map((id) => channelsMap.get(id) as BaseChannelType | ChildChannelType) // We know these values exist
-      .filter((requiredChannel) => requiredChannel.id !== channel.id); // The original data includes the channel's own id, don't include it in the set
+      ?.map((id) => channelsMap.get(id))
+      .filter((requiredChannel) => {
+        // The original data includes the channel's own id, don't include it in the set
+        return requiredChannel && requiredChannel.id !== channel.id;
+      }) as (BaseChannelType | ChildChannelType)[];
     if (requiredChannels?.length) {
       requiresMap.set(channelId, new Set(requiredChannels));
       requiredChannels.forEach((requiredChannel) => {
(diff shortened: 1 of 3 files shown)


cbbayburt:
Oh just one thing: could you add as a comment in `findSyncedMandatoryChannels` a paragraph explaining our situation and the bsc#? Would be helpful if someone decides to touch it in a year or so :)

web/html/src/manager/content-management/shared/components/panels/sources/channels/channels-api-transforms.ts
```diff
@@ -27,7 +27,17 @@ export const toCanonicalRequires = (
 
     const requiredChannels = rawRequiresMap[channelIdString]
       ?.map((id) => channelsMap.get(id) as BaseChannelType | ChildChannelType) // We know these values exist
-      .filter((requiredChannel) => requiredChannel.id !== channel.id); // The original data includes the channel's own id, don't include it in the set
+      .filter((requiredChannel) => {
+        try {
```
Etheryte, line 31: This try-catch block seems a bit unclear to me, it's not immediately obvious what the `else` part does? Do I understand correctly that the main instance we want to catch is any of the values is `undefined` or `null`?

Would something like this have the same functional result?

```tsx
const requiredChannels = rawRequiresMap[channelIdString]
  ?.map((id) => channelsMap.get(id))
  .filter((requiredChannel) => {
    // The original data includes the channel's own id, don't include it in the set
    return requiredChannel && requiredChannel.id !== channel.id;
  }) as (BaseChannelType | ChildChannelType)[];
```

Etheryte: If we want to avoid the cast, we could also do the following, but that loops the result more times and since there can be a lot of data here, I don't think that tradeoff is worth it.

```tsx
const requiredChannels = rawRequiresMap[channelIdString]
  ?.map((id) => channelsMap.get(id))
  .filter(Boolean)
  .filter((requiredChannel) => {
    // The original data includes the channel's own id, don't include it in the set
    return requiredChannel.id !== channel.id;
  });
```

Etheryte (approved):

parlt91:
> Oh just one thing: could you add as a comment in `findSyncedMandatoryChannels` a paragraph explaining our situation and the bsc#? Would be helpful if someone decides to touch it in a year or so :)

Just saw your comment. I added a comment to the code explaining the situation

PR #6884, opened 2023-04-16
Use Context manager for spacewalk-backend

sbluhm:
## What does this PR change?

Use the context manager for all spacewalk-backend files.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
Build tested ok on copr for EL9 and Leap 15.4.
Installation on EL9 and basic sanity test also ok (installation, setup, creation of channel and sync).

- [X] **DONE**

## Links

- [X] **DONE**

## Changelogs

- [X] No changelog needed

Changelog already exists on this version. Otherwise will add in case of intermediate release.

## Re-run a test

### DIFF
--- a/python/spacewalk/cdn_tools/candlepin_api.py
+++ b/python/spacewalk/cdn_tools/candlepin_api.py
@@ -19,9 +19,10 @@ import tempfile
 import requests
 
 from spacewalk.cdn_tools.constants import CA_CERT_PATH
-from uyuni.common.cli import getUsernamePassword
 from spacewalk.common.rhnConfig import CFG
 from spacewalk.satellite_tools.syncLib import log, log2
+from uyuni.common.cli import getUsernamePassword
+from uyuni.common.context_managers import cfg_component
 
 
 class CandlepinApi(object):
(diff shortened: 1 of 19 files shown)


agraul (commented):
Before reviewing each change in-depth, I wonder about two things:

1. Which of these changes target the traditional stack, i.e. files we want to delete from the repository? We don't need to spent effort on converting those to the context manager
2. How did you got to the decision to use `component=None` when there was no `initCFG` call in the old code?

sbluhm:
@agraul , if there was no initCFG call, I assumed a `component=None`. But by the way you asked that question, my assumption was probably totally wrong...
So in that case I would close this PR and not pursue it any further as I have no real scope for doing it properly. Sorry about that. I thought this was just a mindless code replacement.

agraul:
> @agraul , if there was no initCFG call, I assumed a `component=None`. But by the way you asked that question, my assumption was probably totally wrong...

I think it can be wrong, the last `initCFG` call done before the access via `CFG` is "sticky", so really we would need to see the code flow to determine the actually used component at the time of `CFG`.

> So in that case I would close this PR and not pursue it any further as I have no real scope for doing it properly. Sorry about that. I thought this was just a mindless code replacement.

No problem, I have skipped this task as well when I added the context manager. It's a lot of work for not a lot of gain. Thank you for your effort nonetheless!

PR #6923, opened 2023-04-24
Datetime tooltips do not show timestamps

Etheryte:
## What does this PR change?

Make datetime value fields use machine-readable values and correctly use user configuration in titles. Also remove unused code and migrate related code out of the old `spacewalk-essentials.js`.  

Example output:  

```html
<time title="2023-04-25 09:18 PDT" datetime="2023-04-25T09:18:28.475-07:00">a few seconds ago</time>
```

**TODO:** Check if this also fixes bsc#1195190

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/18451

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/components/datetime/FromNow.tsx
+++ b/web/html/src/components/datetime/FromNow.tsx
@@ -1,6 +1,8 @@
-import * as React from "react";
+import { MomentInput } from "moment";
 
 import { localizedMoment } from "utils";
+
+import { getTimeProps } from "./timeProps";
 
 type Props = {
   value?: string | moment.Moment;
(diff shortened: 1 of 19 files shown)


web/html/src/components/datetime/FromNow.tsx
```diff
@@ -1,23 +1,32 @@
...
+    <time title={title} dateTime={dateTime}>
+      {value.fromNow()}
+    </time>
+  );
 };
 
 // It returns a simple string instead of a component in the `fromNow` localized format
 // Use case: at the moment the `t()` function (to translate strings) does not support complex components
-export const fromNow = (value) => {
+// TODO: Make this obsolete once https://github.com/SUSE/spacewalk/issues/20449 is implemented
```
cbbayburt, line 29: Just out of curiosity: how is the localization tool going to help with this?

Etheryte: Once we have a `t()` function that supports components mixed into translations, we can use the `<FromNow />` component instead of this function since then we get the tooltips etc as well, the `fromNow()` function only generates the text.

cbbayburt: oh ok, *components mixed into translations* would be really cool.

PR #6942, opened 2023-04-26
Minor Uyuni bugfixes

cbbayburt:
Fixes various minor bugs.

## Links

- Fixes https://github.com/SUSE/spacewalk/issues/21095
- Fixes https://github.com/SUSE/spacewalk/issues/20576

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/spacecmd/src/spacecmd/misc.py
+++ b/spacecmd/src/spacecmd/misc.py
@@ -462,13 +462,17 @@ def do_whoamitalkingto(self, args):
 ####################
 
 def _show_traditional_stack_message(self):
-    has_traditional_systems = self.client.system.hasTraditionalSystems(self.session)
+    try:
+        has_traditional_systems = self.client.system.hasTraditionalSystems(self.session)
 
-    if has_traditional_systems:
-        logging.warning((
-            'The traditional stack is unsupported and scheduled for removal. '
-            'Consider migrating to salt minions.'
-        ))
+        if has_traditional_systems:
+            logging.warning((
+                'The traditional stack is unsupported and scheduled for removal. '
+                'Consider migrating to salt minions.'
+            ))
+    except xmlrpclib.Fault:
+        # hasTraditionalSystems endpoint is not available prior to 4.3
+        logging.debug(_N("Skipping traditional system check"))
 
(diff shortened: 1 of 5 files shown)


spacecmd/src/spacecmd/misc.py
agraul, line 1: From what I can see in other files, we don't use `try...except` to handle availability of API endpoints, could you adjust your code to use `hasattr()`?
For inspiration you could take a look at 
 https://github.com/uyuni-project/uyuni/blob/2c841115a1115ab441e030f58cc617373331ab6e/spacecmd/src/spacecmd/activationkey.py#L1681-L1683

agraul (changes requested):
I only looked at the Python code and would prefer if we could be consistent with existing code that checks the availability of API endpoints.

spacecmd/src/spacecmd/misc.py
cbbayburt: @agraul I don't understand how you can use `hasattr` to check the availability of an API endpoint. All the existing usages check for a spacecmd function, not an API endpoint.

The only alternative I can see is to make a call to SUMAs `api.getApiNamespaceCallList` to list all the endpoints and do a second call if `hasTraditionalSystems` exists. But that means pessimistically doing 2 calls every time, and since `spacecmd` is used mostly interactively, 2 sequential calls would have a noticeable delay so I wouldn't prefer that.

agraul: You're correct, existing uses of `hasattr` only check for spacecmd functions. I was confused about that. There is, however, one instance where `hasattr`'s "sibling" `getattr` is used: https://github.com/uyuni-project/uyuni/blob/2c841115a1115ab441e030f58cc617373331ab6e/spacecmd/src/spacecmd/api.py#L88

Looking at it more closely, that does not work as I expected. I thought it would check if the method is actually available, but it does not - it just calls `api()` and catches the exception like you do. I guess we don't have an XMLRPC server that supports introspection, otherwise we could use https://docs.python.org/3/library/xmlrpc.client.html#xmlrpc.client.ServerProxy.system.listMethods

agraul (approved):

PR #6953, opened 2023-04-27
Fix ise no scc creds nor fromdir

mcalmer:
## What does this PR change?

When neither SCC credentials nor a local mirror is configured happens,
the isRefreshNeeded() method result in an NPE and we have only a Internal Server Error.
But in such a case a refresh is not needed, so we now return just false.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Port of https://github.com/SUSE/spacewalk/pull/21287

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mcalmer:
@mackdk the third commit is a fix for Project Lotus. I think the testcase was merged "inside" of another test. As a test is a function, it was seen just as inline function which was never called. I hope the fix is ok now

mackdk (approved):

PR #6974, opened 2023-05-03
Use correct origin in case of no mirror

Bischoff:
## Links

Ports:
* 4.3:
* 4.2:

## Changelogs

- [x] No changelog needed


srbarrios (changes requested):
I'd do it differently.
We need to dynamically know if we use or not the mirror, and then use one or another URL depending on it.
Otherwise here, in case we have a mirror, we will not use it.

Bischoff:
> I'd do it differently. We need to dynamically know if we use or not the mirror, and then use one or another URL depending on it. Otherwise here, in case we have a mirror, we will not use it.

Please read the description.

vandabarata (approved):

PR #6975, opened 2023-05-03
Increased timeout for the Rocky 8 case

Bischoff:
## What does this PR change?

Follow-up of #6974

Measured 777 seconds to transfer Rocky 8, the timeout was at 700 seconds -> the ISO image was truncated

## Links

Ports together with ports of #6974

## Changelogs

- [x] No changelog needed


ktsamis (approved):

srbarrios:
Just for the record.
We can improve it, by making the timeout dynamic depending in the file size.

```
 curl -sI $URL | grep -i Content-Length | awk '{print $2}'
```

PR #7001, opened 2023-05-09
Update monitoring-admin.tsx

slsnow:
# Minor Spelling Fixes

## What does this PR change?

There are a few occurrences of the word "occurred" that are misspelled, and an accidental "to" twice in a row. A very minor change.


cbbayburt (changes requested):
Thanks for the fix, @slsnow!

Even though minor, since the change is visible to users, we'll need a changelog entry for it.
Please add a short sentence on top of `/web/spacewalk-web.changes` following its syntax convention. This will satisfy the failing check.

cbbayburt (approved):

PR #7016, opened 2023-05-14
Create unique result folder

maximenoel8:
## What does this PR change?

When we rerun test on existing platform, the former results are mixing with the new results creating messy result html summary.
This PR is design for Jenkins run. It will take the build number to create a unique folder where only the test results run during this build will be store.
The html result will only have the last run results.
The former html result will still be available in the former build result directory.

This change will not affect local run.

## Links : 
CI PR : https://github.com/SUSE/susemanager-ci/pull/866

## Changelogs

- [x] No changelog needed


maximenoel8:
If I'm correct, terracumber is only calling rake generate_test_report but I will double check.

srbarrios:
Terracumber has another step that collects the results.

PR #7047, opened 2023-05-23
systemctl is-active does not always return "inactive" if inactive

Bischoff:
## What does this PR change?

This PR works around this bug https://bugzilla.redhat.com/show_bug.cgi?id=1073481
```
[root@suma-42-min-centos7 ~]# systemctl show venv-salt-minion.service -p ActiveState
ActiveState=inactive
[root@suma-42-min-centos7 ~]# systemctl is-active venv-salt-minion 
unknown
```

--> we can't test for `systemctl is-active` to return `inactive` when the service is inactive.

## Links

Ports:
* 4.2: https://github.com/SUSE/spacewalk/pull/21554
* 4.3: https://github.com/SUSE/spacewalk/pull/21553

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -248,14 +248,16 @@ When(/^I apply highstate on "([^"]*)"$/) do |host|
   $server.run_until_ok("#{cmd} #{system_name} state.highstate")
 end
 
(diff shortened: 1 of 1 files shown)


srbarrios:
I would not split it, as that generates code duplication and we need to maintain two step definitions.
But ok, already merged.

PR #7048, opened 2023-05-23
Prevent product installation from being executed before executing product migration

wweellddeerr:
## What does this PR change?

This merge request addresses an issue where, during the creation of a product migration, a new product was being installed before the product migration was executed. This led to a cascading upgrade of other products that the newly added product depends on.

In certain cases, when the client has LTSS (Long Term Service Support) product installed, this cascading upgrade will fail although the real product migration actually worked. The side effect of this is to have a successfully product migration being marked as Failed in the UI because of problems in product installation.

The changes in this pull request aim to prevent the Salt state responsible for product installation from being executed before the actual product migration. This ensures that the product migration takes place first, allowing for a smooth upgrade process without triggering unnecessary cascading upgrades or causing conflicts when removing certain dependencies. By doing so, we mitigate the issues caused by unintended cascading upgrades and avoid potential conflicts with LTSS removal during the migration process.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: bug fix

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/21161

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/channels/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/channels/init.sls
@@ -155,6 +155,9 @@ mgrchannels_install_products:
 {%- else %}
       - mgrcompat: sync_states
 {%- endif %}
+{%- if salt['pillar.get']('susemanager:distupgrade', False) %}
+      - mgrcompat: spmigration
+{%- endif %}
 {%- endif %}
 {%- endif %}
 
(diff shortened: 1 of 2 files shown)


mbussolotto:
Thanks for the explanation.  

> 
>  The other option to install these new products would be to create a new `Apply states [channels]` just after the real migration in the same way we do for the dry run.

The purpose of https://github.com/mbussolotto/uyuni/blob/c2f21ba22662d93dd8fe14b16b679db467108576/java/code/src/com/suse/manager/webui/services/SaltServerActionService.java#L1345-L1348 was to avoid the user to execute the highstate manually, but do that automatically. So maybe we should install the product automatically also after the migration

mcalmer (approved):
I think you tested it, if this works in the distupgrade case to move this state behind the migration state. I am fine with this solution as it avoid duplicating states.

PR #7051, opened 2023-05-23
Fix low contrast texts in Login page

### LINKED ISSUE #6776: Some texts in the sign in page have very low contrast
# Short description
Some texts in the sign in page of Uyuni have very low contrast. Improving the contrast levels for the texts would improve accessibility.

# Details
As part of openSUSE/mentoring#190, this issue hopes to improve on the accessibility for the sign in page. Currently, there are some texts on the page that have very low contrast, the texts and background colours have contrast ratio that are below the 4.5:1 that should be between the foreground (text) and the background.
Screenshot 2023-03-23 at 22 46 07
Screenshot 2023-03-23 at 22 45 59
 Adequate contrast of text is necessary for all users, especially users with low vision.

To fix this, we should increase the contrast between the foreground (text) colour and the background colour. This can be done by using a darker shade of the current colours for the texts.

Using `#00D18B` for the `View website` and `#B5B8BA` for the `footer` texts achieves the right contrast to improve the accessibility for the text.

btn
footer

# References
[Contrast (Minimum)](https://webaim.org/standards/wcag/checklist#sc1.4.3)

Rakielle:
## What does this PR change?

**This PR fixes an accessibility issue on the login page. Previously, some texts on the page had very low contrast making them hard to read. This PR lands a fix for those texts**

## GUI diff

No difference.

Before:
before-footer
before-button
After:
now-footer
now-button
- [x] **DONE**

## Documentation
- No documentation needed: **This is a change in the shade of the color of some tets and so does not require documentation**

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes #6776 

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/branding/css/uyuni/legacy-variables.less
+++ b/web/html/src/branding/css/uyuni/legacy-variables.less
@@ -4,8 +4,10 @@
 @green-dark:             #02A49C;
 @green:                  #00C081;
 @green-light:            #02D35F;
+@green-light-1:          #00D18B;
 @gray-dark:              #5F5F5F;
 @gray-dark-1:            #777777;
+@gray-dark-2:            #B8B8B8;
 @gray:                   #A7A9AC;
 @gray-light:             #DCDDDE;
 @gray-light-2:           #EBEBEB;
--- a/web/html/src/branding/css/uyuni/login.less
+++ b/web/html/src/branding/css/uyuni/login.less
@@ -33,7 +33,7 @@ body.login-page {
     font-size: 32px
   }
   .btn-dark {
-    color: @green;
+    color: @green-light-1;
     border: 1px solid @green;
     display: inline-block;
     text-decoration: none;
@@ -109,7 +109,7 @@ body.login-page {
     position: fixed;
     padding-bottom: 1em;
     padding-left: 0;
-    color: @gray;
+    color: @gray-dark-2;
     div.wrapper {
       background-size: auto 35px;
       min-height: 35px;


Etheryte:
I think the main question we need to address here is the branding one. Off the top of my head I don't recall exactly, but I suspect the green color is a specific branding color and we shouldn't change that one. Ccing @admd for input.

Depending on that, the current solution might suffice or we might need to only change the background darker while keeping the green or similar.

As an aside, I think it would make sense to simply update the existing variables instead of creating new ones so we get the same contrast bump everywhere they're used.

Rakielle:
> I think the main question we need to address here is the branding one. Off the top of my head I don't recall exactly, but I suspect the green color is a specific branding color and we shouldn't change that one. Ccing @admd for input.
> 
> Depending on that, the current solution might suffice or we might need to only change the background darker while keeping the green or similar.
> 
> As an aside, I think it would make sense to simply update the existing variables instead of creating new ones so we get the same contrast bump everywhere they're used. Other than that this looks good, thanks a lot for fixing this.

@Etheryte I previously discussed the options you highlighted in your comment with @deneb-alpha and @cbbayburt, we thought that modifying the values of the variables might have unintended consequences on other pages. For instance, replacing the value of the `@green` variable with `#00D18B` might cause another low contrast issue in an element that depends on the `@green` variable if it's on a lighter background. Also, making the background lighter works but that has a greater effect on branding IMHO.

So, we chose to go with adding lighter and darker shades of the base colors as that shouldn't have much effect on branding. For example, the `@green` base color has lighter and darker shades defined for them in the codebase, I didn't use those as they are either lighter or darker than what we need to fix the low contrast issue so I added a new lighter shade that is not as light as the previously added lighter shade of the base `@green` color.

cbbayburt:
> I think the main question we need to address here is the branding one. Off the top of my head I don't recall exactly, but I suspect the green color is a specific branding color and we shouldn't change that one.

@Etheryte, apart from brand.suse.com, AFAIK we don't have an officially agreed color palette for Uyuni. Do you know of any color palette to refer to for Uyuni in such cases? Otherwise, I think we can keep referring to SUSE's palette for such changes. As long as the new colors are picked from this palette, I think we should be fine.

Rakielle:
> @Rakielle how did you pick the new colors actually? Arbitrarily? To make things even simpler, can we use some of the other existing color options such as `@green-light` and `@gray-light`?

I  used the WAVE extension to generate a lighter shade of the base `green` color

Rakielle:
@cbbayburt Using `@green-light` instead of `@green-light-1` (the new shade) makes the text color uneven with the background:
`@green-light`:
Screenshot_20230524_145442
`@green-light-1`:
now-button

And for the footer texts, using  the previously set darker shades doesn't fix the low contrast issue and the lighter shades are too bright to be used on footer texts
`@gray-dark-2` (The newly added shade):
Screenshot_20230524_151624
`@gray-dark`:
Screenshot_20230524_151918
`@gray-dark-1`:
Screenshot_20230524_152131

cbbayburt:
@Rakielle fair enough 👍

Etheryte:
Thanks for the input, I didn't know you already discussed this. In that case, sounds good. 👍

Rakielle:
@cbbayburt 
I looked through the colors in brand.suse.com but couldn't find any that fixes the low contrast issue.

cbbayburt (approved):
@Rakielle in that case, this would be the best solution. Thanks!

PR #7078, opened 2023-06-01
Refactoring the step to cleanup salt minion left-overs

srbarrios:
## What does this PR change?

Refactoring the step to cleanup salt minion left-overs.

- I removed spacewalk-proxy-salt from the packages to remove, until we have a good reason to remove it. I don't find the reason we were doing it in all the minions.
- I fixed the packages we were cleaning in the case of Debian-Like minion in Uyuni
- I re-organized if-statements with the aim to make it easier to read and follow

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Ports:
- Manager-4.2
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/salt_steps.rb
+++ b/testsuite/features/step_definitions/salt_steps.rb
@@ -498,18 +498,23 @@ end
 
 When(/^I perform a full salt minion cleanup on "([^"]*)"$/) do |host|
   node = get_target(host)
-  pkgs = $use_salt_bundle ? "venv-salt-minion" : "salt salt-minion"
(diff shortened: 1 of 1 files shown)


Bischoff (approved):
you could construct a list of commands to run in a string, and only run it at the end

that would result in one `cmd.run()` instead of two

but that would be a bit less legible, so it's fine as it is

nodeg (approved):

PR #7082, opened 2023-06-01
QE: Add step to wait for salt jobs to finish on buildhost as part of cleanup

vandabarata:
## What does this PR change?

One of the likely reasons for issues on the flaky tests on the buildhost, is that we're trying to do a build action while an image profile is still being deleted, as evidenced by the presence of this in the `rhn_web_ui` logs of the server
```yaml
[salt-event-thread-6] WARN  com.suse.manager.utils.SaltUtils - Image Profile ID not found while performing: Image Build suse_real_key in handleImageBuildData
[salt-event-thread-6] WARN  com.suse.manager.utils.SaltUtils - Image Profile ID not found while performing: Image Build suse_real_key in handleImageBuildData
```
These happened after the supposed cleanup and image deletion was done, and we were already trying to build the `auth_registry_profile` image, in the feature that ran after this.
Looking at the code, it gives us a hint of what is going on
```java
            Long imageProfileId = details.getImageProfileId();
            if (imageProfileId == null) { // It happens when the image profile is deleted during a build action
                LOG.error("Image Profile ID not found while performing: {} in handleImageBuildData", action.getName());
                return;
            }
```

This means that the image profile deletion wasn't done in time, likely due to performance issues. Instead of increasing the timeout, we are opting to wait for all salt jobs to be done on the buildhost before moving to the next features, to avoid issues from one feature to another. (Discussion can be seen in [this slack thread](https://suse.slack.com/archives/C02DDMY6R0R/p1685621964169959))

## Links

- 4.2
- 4.3 https://github.com/SUSE/spacewalk/pull/21658
- Fixes https://github.com/SUSE/spacewalk/issues/21381

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/buildhost_docker_auth_registry.feature
+++ b/testsuite/features/secondary/buildhost_docker_auth_registry.feature
@@ -58,3 +58,6 @@ Feature: Build image with authenticated registry
 
   Scenario: Cleanup: delete registry image
     When I delete the image "auth_registry_profile" with version "latest" via API calls
+
+  Scenario: Cleanup: Make sure no job is left running on buildhost
+    When I wait until no Salt job is running on "build_host"
--- a/testsuite/features/secondary/buildhost_docker_build_image.feature
+++ b/testsuite/features/secondary/buildhost_docker_build_image.feature
@@ -152,3 +152,6 @@ Feature: Build container images
     And I should see a "Are you sure you want to delete selected profiles?" text
     And I click on the red confirmation button
     And I wait until I see "Image profiles have been deleted" text
+
+  Scenario: Cleanup: Make sure no job is left running on buildhost
+    When I wait until no Salt job is running on "build_host"
--- a/testsuite/features/secondary/buildhost_osimage_build_image.feature
+++ b/testsuite/features/secondary/buildhost_osimage_build_image.feature
@@ -64,3 +64,6 @@ Feature: Build OS images
     And I should see a "Are you sure you want to delete the selected profile?" text
     And I click on the red confirmation button
     And I wait until I see "Image profile has been deleted" text
+
+  Scenario: Cleanup: Make sure no job is left running on buildhost
+    When I wait until no Salt job is running on "build_host"


nodeg (approved):
Nice catch! Thank you Vanda. LGTM.

srbarrios:
I think we need to find a way to identify exactly the Salt job we want to monitor, not just all the Salt jobs.
If we wait for all of them, that will fit on a sequential pass of all the tests, but that will not work when we try to run multiple tests in parallel even if there is no relation between the components in use and it can work at same time in a real use case.

elariekerboull (approved):

vandabarata:
> I think we need to find a way to identify exactly the Salt job we want to monitor, not just all the Salt jobs. If we wait for all of them, that will fit on a sequential pass of all the tests, but that will not work when we try to run multiple tests in parallel even if there is no relation between the components in use and it can work at same time in a real use case.

but do we use the buildhost for anything else other than this purpose? I think it's fine to do it like this in this case exactly because the buildhost only has one purpose :thinking:

nodeg:
> I think we need to find a way to identify exactly the Salt job we want to monitor, not just all the Salt jobs. If we wait for all of them, that will fit on a sequential pass of all the tests, but that will not work when we try to run multiple tests in parallel even if there is no relation between the components in use and it can work at same time in a real use case.

Couldn't we take care of all the parallelization tasks and tests in a future PR? I mean we still do a lot of sequential things and this one here clearly does not work as it should. So for now we have a working solution although it is not yet future proof with respect to more parallelization.

srbarrios:
@nodeg @vandabarata We has been always postponing coding our tests thinking in parallel tests, that's the main reason we still not support it. And trying to do all in once is impossible, I already tried, and that's an Odissey.
That's why I think that if we have the chance to code everything we do in a way that will be more precise, and not go the "all jobs, all processes, all whatever" strategy.

@mcalmer Gave a smarter approach in the thread the Vanda shared in the description, where you can know the state of salt jobs, and we can even try to identify the correct id for that salt job.

So, if we can do it better today... why to wait doing better next year?
```
 salt-run jobs.last_run metadata='{"suma-action-id": 71}'
```

PR #7094, opened 2023-06-03
give cobbler a second to come up

mcalmer:
## What does this PR change?

After restarting cobblerd we directly test access which fail often just because cobblerd seems not be totally up yet.
Give it a second to start.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/pull/21685
Does not exist in 4.2

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/cobbler_steps.rb
+++ b/testsuite/features/step_definitions/cobbler_steps.rb
@@ -288,6 +288,8 @@ When(/^I start local monitoring of Cobbler$/) do
   else
     $server.run('systemctl restart cobblerd')
   end
+  # give cobbler a second to come up
+  sleep 1
 end
 
 Then(/^the local logs for Cobbler should not contain errors$/) do


testsuite/features/step_definitions/cobbler_steps.rb
```diff
@@ -288,6 +288,8 @@
   else
     $server.run('systemctl restart cobblerd')
   end
+  # give cobbler a second to come up
+  sleep 1
```
srbarrios, line 292: I don't like this approach.
We have a step that checks if a service is active, we should use it, instead of guessing a period of time that will differ depending the environment or how the machine is under load.

srbarrios (changes requested):

PR #7128, opened 2023-06-12
CI: Acceptance tests at PRv2: do not run if we only change the changelog

jordimassaguerpla:
## What does this PR change?

CI: excludes running the acceptance tests if only the changelog has changed.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed
- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links

N/A
- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/acceptance_tests_secondary.yml
+++ b/.github/workflows/acceptance_tests_secondary.yml
@@ -6,7 +6,7 @@ on:
       - 'web/html/src/**'
       - 'testsuite/**'
       - '.github/workflows/acceptance_tests_secondary.yml'
-      - '!java/*.changes'
+      - '!java/*.changes*'
 jobs:
   test-uyuni:
     uses: ./.github/workflows/acceptance_tests_common.yml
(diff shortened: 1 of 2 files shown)


jordimassaguerpla:
@cbbayburt With this we could enable the acceptance tests before the changes in the changelog rules. Can you review?

.github/workflows/acceptance_tests_secondary.yml
```diff
@@ -6,6 +6,7 @@ on:
       - 'web/html/src/**'
       - 'testsuite/**'
       - '.github/workflows/acceptance_tests_secondary.yml'
+      - '!java/*.changes'
```
cbbayburt, line 9: Existing changelog file is named `spacewalk-java.changes` but the new files going to be named like: `spacewalk-java.changes.cbbayburt.myfeature`.

So when PRs are changed, the entries must be moved from `spacewalk-java.changes` to the new files, so both files will be modified.

To cover both, I think it should be something like this:
```suggestion
      - '!java/*.changes*'
```

cbbayburt (changes requested):

PR #7154, opened 2023-06-16
Add devel repo for each minion needed

ktsamis:
## What does this PR change?

Because we are missing devel repos in our activation keys and channels we are not currently testing the devel code in the CI. This PR attempts to fix that gap. We add devel repos for each os family as a custom repo in the fake channels we use in the activation keys.

This should not only fix the 4.2 failure for init buildhost but also cover a gap we have. We are currently testing our nightly builds in CI without devel repos (added in sumaform, disabled as soon as we bootstrap) because we only use the latest released from the SUMA server when we bootstrap. 

context: https://bugzilla.suse.com/show_bug.cgi?id=1212416
more context: https://suse.slack.com/archives/C02D12TNYLS/p1686915333705209

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Cucumber tests were added

- [x] **DONE**

## Links

Tracks 4.3: https://github.com/SUSE/spacewalk/pull/21794
4.2: https://github.com/SUSE/spacewalk/pull/21793

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1573,3 +1573,4 @@ And(/^I choose the right devel url for "([^"]*)" and add it as url$/) do |host|
   end
   steps %(And I enter "#{link}" as "url")
 end
+
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1560,3 +1560,16 @@ When(/^I do a late hostname initialization of host "([^"]*)"$/) do |host|
   node.init_os_family(os_family)
   node.init_os_version(os_version)
 end
+
+And(/^I choose the right devel url for "([^"]*)" and add it as url$/) do |host|
+  node = get_target(host)
+  os_family = node.os_family
+  if os_family =~ /^opensuse/ || os_family =~ /^sles/
+    link = "grep baseurl /etc/zypp/repos.d/tools_additional_repo.repo | awk -F '='  '{print $2}'"
+  elsif os_family =~ /^centos/
+    link = "grep baseurl /etc/yum.repos.d/tools_update_repo.repo | awk -F '='  '{print $2}'"
+  elsif os_family =~ /^ubuntu/ || os_family =~ /^debian/
+    link = "awk -F ' ' '{print $3}' /etc/apt/sources.list.d/tools_update_repo.list"
+  end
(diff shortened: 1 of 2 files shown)


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -1560,3 +1560,16 @@
...
+  os_family = node.os_family
+  if os_family =~ /^opensuse/ || os_family =~ /^sles/
+    link = "grep baseurl /etc/zypp/repos.d/tools_additional_repo.repo | awk -F '='  '{print $2}'"
+  elsif os_family =~ /^centos/
+    link = "grep baseurl /etc/yum.repos.d/tools_update_repo.repo | awk -F '='  '{print $2}'"
+  elsif os_family =~ /^ubuntu/ || os_family =~ /^debian/
+    link = "awk -F ' ' '{print $3}' /etc/apt/sources.list.d/tools_update_repo.list"
+  end
+  steps %(And I enter "#{link}" as "url")
+end
```
vandabarata, line 1575:
```suggestion
end

```
missing new line

ktsamis: fixed

nodeg:
If you still want this, it should be rebased, @ktsamis.

nodeg (approved):

nodeg (approved):
LGTM. Do not forget to do a rebase!

srbarrios (changes requested):
Hi Kostas.

On this PR, we are moving the responsibility of creating development repositories to our CI Testsuite, when this responsibility is for now given to Sumaform Salt states.

Sometime ago I was proposing to @Bischoff that exact move, but this will need to made as a big refactor where we need to align what it has been done by sumaform and what we do on CI testsuite.

This will also make the BV and CI testsuite similar, as we will be using custom channels/repositories in both of them.

But I strongly recommend that we postpone this change, and we do it with calm, consensus and well discussed.

For now, I think the best we can do is to keep using the injected repositories from sumaform, and use the steps we already have to enable and disable the repositories as soon as they become disabled by the onboarding, so as a last test after onboarding.

Also, we will need to assure that they are not disabled triggering other events like applying a salt high-state.

ktsamis:
I agree with Oscar, we need to discuss this more, and this PR is pretty old so it needs so much work it's better to start from scratch. I will close and we can revisit this later.

PR #7155, opened 2023-06-16
Increase timeout after restart of service

ktsamis:
## What does this PR change?

Increase timeout after restart of service slightly

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Tracks # **add downstream PR, if any** needs ports

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/cobbler_steps.rb
+++ b/testsuite/features/step_definitions/cobbler_steps.rb
@@ -293,8 +293,8 @@ When(/^I start local monitoring of Cobbler$/) do
   else
     $server.run('systemctl restart cobblerd')
   end
-  # give cobbler a second to come up
-  sleep 1
+  # give cobbler a few seconds to come up
+  sleep 3
 end
 
 Then(/^the local logs for Cobbler should not contain errors$/) do


vandabarata (approved):

testsuite/features/step_definitions/cobbler_steps.rb
```diff
@@ -293,8 +293,8 @@
   else
     $server.run('systemctl restart cobblerd')
   end
-  # give cobbler a second to come up
-  sleep 1
+  # give cobbler a few seconds to come up
+  sleep 3
```
srbarrios, line 297: As I mentioned in the previous PR from MC, I would prefer a smarter approach.
Having a constant value to sleep, we will depend on the environment and the workload we could have on the server.

It would be much better to use something like our step:
```
When(/^I wait until "([^"]*)" service is active on "([^"]*)"$/) do |service, host|
```

nodeg (approved):
LGTM, but I agree with Oscar's comment.

testsuite/features/step_definitions/cobbler_steps.rb
ktsamis: this doesn't work unfortunately, I tried everywhere to add this step ```step %(I wait until "cobblerd" service is active on "server")``` but it didn't work, somehow systemd doesn't get the true status of cobbler.

PR #7166, opened 2023-06-20
Do not build uyuni-base server subpackage for non relevant systems

vzhestkov:
## What does this PR change?

`server` subpackage of `uyuni-base` is trying to build for CentOS 7 what is not really required.
The change is making the `server` subpackage to be built only for relevant systems.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: not required as failing on building

## Changelogs

- [ ] No changelog needed

## Re-run a test


vzhestkov:
@sbluhm could you please check if the change is OK for you?

sbluhm:
Overall it looks OK.
But as Uyuni won't run on EL8 anymore, maybe we should make the condition `rhel >=9`?
Not sure if that makes any difference in future or not.

vzhestkov:
@sbluhm yes make sense. I just set to 8 because we have an issue with 7 right now. Will change to 9 and higher :+1:

uyuni/base/uyuni-base.changes
```diff
@@ -1,3 +1,4 @@
+- Do not build uyuni-base server subpackage for client tools
```
deneb-alpha, line 1: I think it could be better to  specify which  are  the affected  client tools. this changelog hints  that  we  are not going to build  uyuni-base anymore.

deneb-alpha (changes requested):
not sure if @meaksh has other feedback (his review is still open), but  from my side I would like to  see the  .changes  clarified (together with the  description of the PR).

uyuni/base/uyuni-base.changes
vzhestkov: It actually means we will not build `uyuni-base-server` package (`server` subpackage of `uyuni-base`) for all of the client tools. Is it fine? Or better to set exact name of the package in the changelog entry?

meaksh: Maybe something like this?

```
Only build `uyuni-base-server` package for server and not for client tools
```

I would vote for having exact package name in the changelog.

meaksh (commented):
I think we should rebase this and adapt changelog file to new changelog workflow.

deneb-alpha (approved):
thanks

meaksh (approved):

Issue #7068, opened 2023-05-26
Unable to upgrade packages for MicroOs installation


rdouman:
### Problem description

I am unable to upgrade my server after successfully registering it ("openSUSE MicroOS - 20230522) using salt with the Uyuni server.  I'm am presented with a list of packages to be upgraded, but the upgrade fails once the salt state executes.

### Steps to reproduce

1. Install new opensuse microos server
2. Install salt minion 3006.0-1.1
3. Configure Channel for X86 using repo tumbleweed oss (http://download.opensuse.org/tumbleweed/repo/oss/)
4. Subscribe to the relevant channel and wait for package etc so synch up and list of packages to be upgrade to be available under the software tab for the system.
5. Select packages under the package selection and click upgrade packages
6. Confirm upgrade and wait for salt state to execute.
...

### Uyuni version

```bash
Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : uyuni-server-stable
Name           : Uyuni-Server-release
Version        : 2023.04-220400.204.2.uyuni2
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes (automatically)
Status         : up-to-date
Source package : Uyuni-Server-release-2023.04-220400.204.2.uyuni2.src
Summary        : Uyuni Server
Description    :
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.
```

### Uyuni proxy version (if used)

```bash
Not installed
```

### Additional information

Salt reports the following error when trying to apply the changes:

```bash
Problem: the installed product:MicroOS-20230522-0.x86_64 conflicts with 'distribution-release' provided by the to be installed product:openSUSE-20230524-0.x86_64 

 Problem: the to be installed product:openSUSE-20230524-0.x86_64 requires 'product(openSUSE) = 20230524-0', but this requirement cannot be provided 
 Problem: the installed product:MicroOS-20230522-0.x86_64 conflicts with 'distribution-release' provided by the to be installed product:openSUSE-20230524-0.x86_64

The target server appears to have two conflicting products, namely:
 zypper search --provides --type product
Loading repository data...
Reading installed packages...

S  | Name     | Summary             | Type
---+----------+---------------------+--------
i+ | MicroOS  | openSUSE MicroOS    | product
[...]

PR #7187, opened 2023-06-22
Don't install products on openSUSE client OSes

agraul:
## What does this PR change?
openSUSE repos often contain multiple "product packages", that conflict. The idea is that installer installs the correct product once and that's it. In Uyuni, we currently install all products we can find. That does not work because of the conflict.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: this came up when using openSUSE MicroOS. Until we test with client OSes where this happens (like MicroOS or Tumbleweed), we can't test this properly

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/7068

## Changelogs

- [ ] No changelog needed

## Re-run a test


mcalmer (approved):
I don't like that openSUSE mess around so much, but as it is a free product we do not strictly require  to have the products installed.
So if  it helps, fine with me

PR #7193, opened 2023-06-23
Allow both CentOS and Rocky on all branches

Bischoff:
## What does this PR change?

Do not make assumptions about which RH-like minion we use on each branch.

Rocky 8 uses `NetworkManager` service, CentOS uses `network` service.

Alma and others are missing but can be added easily.

## Links

Ports:
* 4.3: https://github.com/SUSE/spacewalk/pull/21865
* 4.2: https://github.com/SUSE/spacewalk/pull/21867

Fixes #21779

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/retail_steps.rb
+++ b/testsuite/features/step_definitions/retail_steps.rb
@@ -154,7 +154,13 @@ When(/^I set up the private network on the terminals$/) do
     next if node.nil?
     domain, _code = node.run("grep '^search' /etc/resolv.conf | sed 's/^search//'")
     conf = "DOMAIN='#{domain.strip}'\\nDEVICE='eth1'\\nSTARTMODE='auto'\\nBOOTPROTO='dhcp'\\nDNS1='#{proxy}'"
-    node.run("echo -e \"#{conf}\" > #{file} && echo -e \"#{conf2}\" > #{file2} && systemctl restart NetworkManager")
+    service =
+      if node.os_family =~ /^rocky/
+        'NetworkManager'
+      else
+        'network'
+      end
+    node.run("echo -e \"#{conf}\" > #{file} && echo -e \"#{conf2}\" > #{file2} && systemctl restart #{service}")
   end
   # /etc/netplan/01-netcfg.yaml
   nodes = [$deblike_minion]


testsuite/features/step_definitions/retail_steps.rb
```diff
@@ -154,7 +154,8 @@ def compute_kiwi_profile_version(host)
     next if node.nil?
     domain, _code = node.run("grep '^search' /etc/resolv.conf | sed 's/^search//'")
     conf = "DOMAIN='#{domain.strip}'\\nDEVICE='eth1'\\nSTARTMODE='auto'\\nBOOTPROTO='dhcp'\\nDNS1='#{proxy}'"
-    node.run("echo -e \"#{conf}\" > #{file} && echo -e \"#{conf2}\" > #{file2} && systemctl restart NetworkManager")
+    service = node.os_family =~ /^rocky/ ? 'NetworkManager' : 'network'
+    node.run("echo -e \"#{conf}\" > #{file} && echo -e \"#{conf2}\" > #{file2} && systemctl restart #{service}")
```
srbarrios, line 158: Can we have as retail terminal a Rocky system?
This step definition might have and do more than one responsibility?
Maybe we want to offer this into our test framework, but not for retail terminal setup?

ktsamis (changes requested):
If we want to add the alma etc later, let's make this into an if statement with setting the variables without the short form so it's easier to copy paste the same lines later

testsuite/features/step_definitions/retail_steps.rb
Bischoff: No Retail is all about SLES.

Sorry if the context is misleading. We set up the RH-like minion to work with the Retail Proxy, but it remains a normal minion, not a reformatted terminal.

Bischoff:
> If we want to add the alma etc later, let's make this into an if statement with setting the variables without the short form so it's easier to copy paste the same lines later

Indeed we don't use the ternary operator in other parts of the software. For example:
```
features/step_definitions/common_steps.rb:  elsif os_family =~ /^centos/ || os_family =~ /^rocky/
```
Let's do like you say.

ktsamis (approved):
Thanks looks great now, much easier to extend later

PR #7218, opened 2023-06-29
CI: Use "osc api" so that it uses the Multi Factor Authentication

jordimassaguerpla:
## What does this PR change?

In order to use the current user MFA setup, we better use "osc api" instead of trying to use urllib python library. Since a call to "osc" will already use the MFA that is setup for the current user. Otherwise, we would have to implement it with python urllib, which is, at least, challenging :)

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed
- [X] **DONE**

## Test coverage
- No tests

- [X] **DONE**

## Links
see https://github.com/SUSE/spacewalk/issues/21471

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/testing/automation/obs-project.py
+++ b/susemanager-utils/testing/automation/obs-project.py
@@ -2,11 +2,27 @@
 import argparse
 import os
 import sys
-import urllib.request
 import xml.etree.ElementTree as ET
 import datetime
-import configparser
 import re
+import subprocess
+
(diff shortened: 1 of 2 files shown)


ktsamis (approved):
Syntactically looks correct, not sure about different configs how it would work, I haven't tested it but looks ok overall.

PR #7233, opened 2023-07-04
Defer rhnConfig import to cfg_component

agraul:
## What does this PR change?

This import has file I/O as a side effect. That can get in the way, especially when nothing actually access the config. By defering the import to call-time of cfg_component, there is no file I/O when importing the uyuni.common.context_managers module.

## GUI diff

No difference

## Documentation

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/python/uyuni/common/context_managers.py
+++ b/python/uyuni/common/context_managers.py
@@ -1,6 +1,12 @@
 """Collection of context managers for Uyuni."""
 from contextlib import contextmanager
-from spacewalk.common.rhnConfig import CFG, initCFG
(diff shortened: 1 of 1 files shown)


agraul:
We can just not read the config at import time, it's only done for reading the `PRODUCT_NAME`. That's a constant we can set during build time (defaulting to Uyuni).

PR #7280, opened 2023-07-17
Mask uyuni roster module password on logs

rjpmestre:
## What does this PR change?

Creating Uyuni connection instances ends up logging db connection password.
This PR masks it.

 Port of https://github.com/SUSE/spacewalk/pull/22016

## Changelogs

- [ ] No changelog needed

## Re-run a test


agraul:
@rjpmestre I've updated the description to include the original description. Let's try to always include the full PR description in Uyuni PRs, not everyone has access to the downstream repository.

The change itself looks good to me. We lose the ability to see that credentials are wrong, but those are only logged in `trace`, which is not the default.  As user has to change to `trace` and reproduce the problem, which is harder than just checking the config file. And we don't log passwords, so that's definitely an upside.

agraul (approved):

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
```diff
@@ -89,7 +90,7 @@ def __init__(self, db_config, uyuni_roster_config):
                 )
             )
 
-        log.trace("db_connect string: %s", self.db_connect_str)
+        log.trace("db_connect string: %s", re.sub(r"password='[^']*'", "password='******'", self.db_connect_str))
```
vzhestkov, line 93: Oh, I was wrong in my comment in slack. Instead of using `re` I think it's better to recompose the string with `****** (hidden)` instead of the password value, or show distinct values for `dbname`, `user`, `host` in the separate lines.
I would prefer the last option.

susemanager-utils/susemanager-sls/modules/roster/uyuni.py
```diff
@@ -89,7 +89,9 @@ def __init__(self, db_config, uyuni_roster_config):
                 )
             )
 
-        log.trace("db_connect string: %s", self.db_connect_str)
+        log.trace("db_connect dbname: %s", db_config['db'])
+        log.trace("db_connect   user: %s", db_config['user'])
+        log.trace("db_connect   host: %s", db_config['host'])
```
vzhestkov, line 94: Just one nitpick. Let's keep "more modern" double quote style:
```suggestion
        log.trace("db_connect dbname: %s", db_config["db"])
        log.trace("db_connect   user: %s", db_config["user"])
        log.trace("db_connect   host: %s", db_config["host"])
```
Other than that looks good to me :+1:

vzhestkov (approved):
Approved, but please apply the suggested change.
Thanks for taking care of it :+1:

PR #7284, opened 2023-07-19
Disable default schedule tasks to improve testing stability

maximenoel8:
## What does this PR change?

Disable most of default schedule tasks to avoid instability during our BV and CI tests.
We don't want random mgr-create-bootstrap-repo to stop manual call.

For now this change is only for BV but after review I will add it for acceptance tests too.

## Links

Related to : https://github.com/SUSE/spacewalk/issues/21827

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/core/srv_disable_default_schedule_tasks.feature
+++ b/testsuite/features/build_validation/core/srv_disable_default_schedule_tasks.feature
@@ -1,0 +1,42 @@
+# Copyright (c) 2023 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Remove schedule tasks to manually trigger them when we consider the best moment
+
+  Scenario: Log in as admin user
+    Given I am authorized for the "Admin" section
+
+  Scenario: Delete scheduled reposyncs
+    When I follow the left menu "Admin > Task Schedules"
+    And I follow "mgr-sync-refresh-default"
+    And I choose "disabled"
+    And I click on "Update Schedule"
+    And I click on "Delete Schedule"
+
+  Scenario: Delete scheduled sync of custom channels
+    When I follow the left menu "Admin > Task Schedules"
+    And I follow "channel-repodata-default"
+    And I choose "disabled"
+    And I click on "Update Schedule"
+    And I click on "Delete Schedule"
+
+  Scenario: Delete scheduled sync of cobbler
+    When I follow the left menu "Admin > Task Schedules"
+    And I follow "cobbler-sync-default"
+    And I choose "disabled"
+    And I click on "Update Schedule"
+    And I click on "Delete Schedule"
+
+  Scenario: Delete scheduled errata cache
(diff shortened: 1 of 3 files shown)


testsuite/features/build_validation/core/srv_disable_default_schedule_tasks.feature
```diff
@@ -0,0 +1,42 @@
...
+  Scenario: Delete scheduled reposyncs
+    When I follow the left menu "Admin > Task Schedules"
+    And I follow "mgr-sync-refresh-default"
+    And I choose "disabled"
+    And I click on "Update Schedule"
+    And I click on "Delete Schedule"
+
+  Scenario: Delete scheduled sync of custom channels
+    When I follow the left menu "Admin > Task Schedules"
+    And I follow "channel-repodata-default"
```
mcalmer, line 18: This task stop generation of metadata and not sync of custom channels

mcalmer (commented):
Btw. mgr-create-bootstrap-repo is called at the end of reposync. AFAIK there is a config variable to disable this. Just in case you want to make is independent of reposync

testsuite/features/build_validation/core/srv_disable_default_schedule_tasks.feature
nodeg: The [documentation](https://documentation.suse.com/suma/4.3/en/suse-manager/administration/task-schedules.html) says the same:
image

testsuite/features/build_validation/core/srv_disable_default_schedule_tasks.feature
```diff
@@ -0,0 +1,42 @@
...
+
+  Scenario: Log in as admin user
+    Given I am authorized for the "Admin" section
+
+  Scenario: Delete scheduled reposyncs
+    When I follow the left menu "Admin > Task Schedules"
+    And I follow "mgr-sync-refresh-default"
+    And I choose "disabled"
+    And I click on "Update Schedule"
+    And I click on "Delete Schedule"
```
srbarrios, line 14: Thinking it twice, maybe deleting it is a too sharp approach. What about changing the cron, so instead of running every minute, it runs... every week?
That way, once our BV pipeline finish, we can consider a scenario to have the default cron values back here.

testsuite/features/build_validation/core/srv_disable_default_schedule_tasks.feature
maximenoel8: Yes, disabling the `mgr-create-bootstrap-repo` after reposync will solve a lot of our issues in BV. Could you share with me how to disable this behaviour ?

mcalmer: Set in rhn.conf:
server.susemanager.auto_generate_bootstrap_repo = 0

PR #7316, opened 2023-07-26
prevent NPE when cpu values could not fetched or parsed correctly

mcalmer:
## What does this PR change?

When data from virtual host gatherer have an old format, maybe not all fields could be parsed and stay empty/NULL.
This can cause NPE when working with the data.
Provide default values in case we do not have the data

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Tracks 

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/gatherer/HostJson.java
+++ b/java/code/src/com/suse/manager/gatherer/HostJson.java
@@ -88,10 +88,10 @@ public class HostJson {
 
     /**
      * Gets the total CPU socket count.
-     * @return the number of sockets
+     * @return the number of sockets - can be 0 if insufficient data were send
      */
     public Integer getTotalCpuSockets() {
-        return totalCpuSockets;
+        return totalCpuSockets == null ? 0 : totalCpuSockets;
     }
 
     /**
@@ -99,7 +99,7 @@ public class HostJson {
      * @return the cpu cores
      */
     public Integer getTotalCpuCores() {
-        return totalCpuCores;
+        return totalCpuCores == null ? 1 : totalCpuCores;
     }
 
     /**
@@ -107,7 +107,7 @@ public class HostJson {
      * @return the CPU thread count
      */
     public Integer getTotalCpuThreads() {
-        return totalCpuThreads;
+        return totalCpuThreads == null ? 1 : totalCpuThreads;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


java/code/src/com/suse/manager/gatherer/HostJson.java
```diff
@@ -91,23 +92,23 @@ public String getHostIdentifier() {
      * @return the number of sockets
      */
     public Integer getTotalCpuSockets() {
-        return totalCpuSockets;
+        return Optional.ofNullable(totalCpuSockets).orElse(1);
```
mackdk, line 95: Two points here, one nitpicking and one deeper question.

Starting with the nitpicking :slightly_smiling_face: but for these type of uses cases I'd rather use a simple `== null` check or `Objects.requireNonNullElse()` if we want to have a more fluent syntax. Using an `Optional` in this way means we are creating an object in the heap just to throw it away immediately. For this specific case it probably doesn't matter, but it does introduces a significant memory overhead if we use this technique for parsing large JSON files.

The other point is related to the value we are using here as default. This `totalCpuSockets` field is stored as `nrSockets` in the `CPU` object [by the VirtualHostManagerProcessor](https://github.com/uyuni-project/uyuni/blob/c3aa9878e05f6f0e4e31e4086a8926f7cc85aaea/java/code/src/com/redhat/rhn/taskomatic/task/gatherer/VirtualHostManagerProcessor.java#L281). Later, this value is passed to subscription matcher as number of cpus [by the MatcherJsonIO class](https://github.com/uyuni-project/uyuni/blob/c3aa9878e05f6f0e4e31e4086a8926f7cc85aaea/java/code/src/com/suse/manager/matcher/MatcherJsonIO.java#L163). There is a Drools rule where [null values for the cpu count are translated to 16](https://github.com/uyuni-project/subscription-matcher/blob/master/src/main/resources/com/suse/matcher/rules/drools/InputValidation.drl#L116-L127).  I'm not sure if before this PR the `null` value actually reached subscription-matcher or if the code would fail earlier, but we might change the results of subscription-matcher in these corner cases.

So the question is, is 1 really a good default value for this value?

mcalmer: good catch. I have changed this and use for socket value "0".
In case of parsing errors, I do not set any CPU value to keep the null for subscription matcher.
The problem is, that we get now total values from gatherer and we need to divide by sockets to get single numbers.
`null.longValue()` generate a NPE while also divide by zero also not work. So we better set nothing if we get bad data. For cores and threads I assume "1" which should not generate problems.

mackdk (approved):

PR #7319, opened 2023-07-28
Single target client ssh BV

maximenoel8:
## What does this PR change?

Filter bashrc env to only target the currently testing client

## Links
Related to : https://github.com/SUSE/spacewalk/issues/22131

- [x] No changelog needed

### DIFF
--- a/testsuite/Rakefile
+++ b/testsuite/Rakefile
@@ -12,59 +12,6 @@ junit_results = '-f junit -o results_junit'
 build_number = ENV['BUILD_NUMBER']
 result_folder = build_number.nil? ? 'results' : "results/#{build_number}"
 
-namespace :cucumber do
-  # Create results folder
-  sh "mkdir -p #{result_folder}", verbose: false
-  # Rake::Task['utils:extract_undesired_env_variables'].invoke
-  bashrc_content = File.read("/root/.bashrc")
-  undesired_env_variables = []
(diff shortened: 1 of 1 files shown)


srbarrios (commented):
This add an extra complexity to the rake calls that I don't think we need.
If we want to run specific tests for a specific client, we can use Cucumber Tags and our "xxx_scope" tags.

If it's about the initialization we do when we start the test framework I would handle it internally in the test framework, we already have all what we need there.

## Proposal:
**Lazy initialization of Twopence nodes**:
When we start the test framework, we load the Ruby file twopence_init.rb, this is a plain Ruby file that at the beginning of the file start declaring variables for each fqdn of every systems we handle in the test framework, inside that we call twopence_init() passing by parameter the fqdn, and spend some time establishing a connection with each client. If a twopence node can't be initialized this cause a failure in the test framework and it stops.
The proposal is to don't initialize any twopence node when the twopence_init.rb file is loaded, but instead we have a getter method to retrieve the specific twopence node we want to use in our step definitions, and the first time we call this getter, we check if the variable is null or has a value, if it has a value, we return the value, if not we call twopence_init() method to initialize that node. In a second phase, we can convert this plain Ruby file in a class or module and refactor some of this methods and variables.
That would decouple responsibilities and outside the test framework, we don't need to care about which clients we will be using in order to initialize them.

PR #7321, opened 2023-07-28
Double timeout during logging

maximenoel8:
## What does this PR change?

Double timeout during logging 

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -553,11 +553,8 @@ end
 
 Then(/^I should be logged in$/) do
   xpath_query = "//a[@href='/rhn/Logout.do']"
-  # Double the Capybara.default_max_wait_time
-  wait_time = Capybara.default_max_wait_time * 2
-
   # Check if the user is logged in, using the specified wait_time
-  raise 'User is not logged in' unless has_selector?(:xpath, xpath_query, wait: wait_time)
+  raise 'User is not logged in' unless has_selector?(:xpath, xpath_query, wait: Capybara.default_max_wait_time * 2)
 end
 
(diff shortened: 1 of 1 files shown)


testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -553,7 +553,11 @@
 
 Then(/^I should be logged in$/) do
   xpath_query = "//a[@href='/rhn/Logout.do']"
-  raise 'User is not logged in' unless find(:xpath, xpath_query)
+  if has_selector?(:xpath, xpath_query)
+    # The element is present
+  else
+    raise 'User is not logged in' unless find(:xpath, xpath_query)
+  end
```
srbarrios, line 560: I would instead append wait: X here.
As I proposed yesterday, this will setup a specific timeout, and meanwhile the find method will refresh the DOM until it find the element or raise the timeout.

srbarrios (changes requested):
I prefer to setup a specific timeout in that find.

testsuite/features/step_definitions/navigation_steps.rb
srbarrios, line 560:
```suggestion
  raise 'User is not logged in' unless find(:xpath, xpath_query, wait: 30)
```

maximenoel8:
I strongly advice against having any hard written timeout for the login. Why do you think about just doubling the globally timeout ? 
I was looking on internet about how to wait for an element and I'm a little bit surprise we are using find.
From my understanding, find is a method use when interacting with the element afterwards.
To just check if an element exist, it's better to use has_selector.
It even seems the two methods are slightly different.
Could we use the has_selector instead of find ?

srbarrios (approved):

PR #7328, opened 2023-07-31
Refactor: reuse table filter component in recurring actions feature

wweellddeerr:
## What does this PR change?

It extracts common table filter code to a separate component (`TableFilter.tsx`) and reuses it in systems list and recurring action features.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/components/table/TableFilter.tsx
+++ b/web/html/src/components/table/TableFilter.tsx
@@ -4,21 +4,27 @@ import { Select } from "components/input";
 import { Form } from "components/input/Form";
 import { SelectSearchField } from "components/table/SelectSearchField";
 
-const renderSearchField = (props) => {
-  const { field } = props;
-  const selectedOption = props.filterOptions.find((it) => it.value === field);
+const renderSearchField = ({ filterOptions, field, criteria, onSearch, placeholder, name }) => {
+  const selectedOption = filterOptions.find((it) => it.value === field);
   if (selectedOption?.filterOptions) {
-    return <SelectSearchField label={selectedOption.label} options={selectedOption.filterOptions} {...props} />;
+    return (
+      <SelectSearchField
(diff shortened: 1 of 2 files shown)


web/html/src/components/table/TableFilter.tsx
```diff
@@ -0,0 +1,52 @@
+import { useState } from "react";
+
+import { Select } from "components/input";
+import { Form } from "components/input/Form";
+import { SelectSearchField } from "components/table/SelectSearchField";
+
+const renderSearchField = (props) => {
+  const { field } = props;
```
Etheryte, line 8: This works, but feels a bit weird. I think it would make sense to either destruct everything (`const { field, criteria, placeholder, onSearch, name } = props`), or destruct nothing and use `props.field` in the line below this one. I think the latter makes more sense, but either way is fine with me as long as it's consistent.

web/html/src/components/table/TableFilter.tsx
```diff
@@ -0,0 +1,52 @@
...
+        onChange={(e) => props.onSearch?.(e.target.value)}
+        name={props.name}
+      />
+    </div>
+  );
+};
+
+export const TableFilter = (props) => {
+  // Dummy model and onChange to reuse the Select component as it requires a Form
+  let model = {};
```
Etheryte, line 29: If this doesn't change it can be `const`.

Etheryte (approved):
This looks really good, the suggestions I have are purely cosmetic and not important. Thanks a lot for cleaning this up, it looks a lot better now. 👍

PR #7334, opened 2023-08-01
QE: Lazy initialization of Twopence nodes

srbarrios:
## What does this PR change?

Related card https://github.com/SUSE/spacewalk/issues/22101

Note: This is a Prove of Concept. Might need a follow up refactor.

Lazy initialization of Twopence nodes:

When we start the test framework, we load the Ruby file twopence_init.rb, this is a plain Ruby file that at the beginning of the file start declaring variables for each fqdn of every systems we handle in the test framework, inside that we call twopence_init() passing by parameter the fqdn, and spend some time establishing a connection with each client. 

If a twopence node can't be initialized this cause a failure in the test framework and it stops, which might be good, but the side effect of this is the need to initialize all when we start and that can take a lot of time.

The proposal is to don't initialize any twopence node when the twopence_init.rb file is loaded, but instead we have a getter method to retrieve the specific twopence node we want to use in our step definitions, and the first time we call this getter, we check if the variable is null or has a value, if it has a value, we return the value, if not we call twopence_init() method to initialize that node.

That would decouple responsibilities and outside the test framework, we don't need to care about which clients we will be using in order to initialize them.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- [x] **DONE**

## Links

Ports:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


Bischoff (approved):
lgtm

srbarrios:
Waiting to pass the full acceptance tests, but looks quite good on my first tests.

testsuite/features/support/commonlib.rb
```diff
@@ -328,3 +328,100 @@ def add_context(key, value)
...
 end
+
+# This function gets the system name, as displayed in systems list
+# * for the usual clients, it is the full hostname, e.g. suma-41-min-sle15.tf.local
+# * for the PXE booted clients, it is derived from the branch name, the hardware type,
+#   and a fingerprint, e.g. example.Intel-Genuine-None-d6df84cca6f478cdafe824e35bbb6e3b
+def get_system_name(host)
+  case host
+    # The PXE boot minion and the terminals are not directly accessible on the network,
+    # therefore they are not represented by a twopence node
```
vandabarata, line 339: One less space so these line up with the `when`, unless github messed up the formatting as it sometimes does

testsuite/features/support/commonlib.rb
```diff
@@ -328,3 +328,100 @@ def add_context(key, value)
...
+  when 'pxeboot_minion'
+    output, _code = get_target('server').run('salt-key')
+    system_name =
+      output.split.find do |word|
+        word =~ /example.Intel-Genuine-None-/ || word =~ /example.pxeboot-/ || word =~ /example.Intel/ || word =~ /pxeboot-/
```
vandabarata, line 344: I want to alert that we don't always have pxeboot written, nor Intel. I'd simplify this to just start with `example.` or `pxeboot` - unless I misunderstood something
image

testsuite/features/support/commonlib.rb
```diff
@@ -328,3 +328,100 @@ def add_context(key, value)
...
+        word =~ /example.sle12sp5terminal-/
+      end
+    system_name = 'sle12sp5terminal.example.org' if system_name.nil?
+  when 'sle15sp4_terminal'
+    output, _code = get_target('server').run('salt-key')
+    system_name =
+      output.split.find do |word|
+        word =~ /example.sle15sp4terminal-/
+      end
+    system_name = 'sle15sp4terminal.example.org' if system_name.nil?
```
vandabarata, line 360: can't these be simplified above, or am I missing something?

testsuite/features/support/commonlib.rb
srbarrios: > I want to alert that we don't always have pxeboot written, nor Intel. I'd simplify this to just start with `example.` or `pxeboot` - unless I misunderstood something
> image
> 

I might just rely too much in my IDE, fixing rubocop style warnings, I will review that, thanks.

testsuite/features/support/commonlib.rb
```diff
@@ -328,3 +328,100 @@ def add_context(key, value)
...
+    rescue RuntimeError
+      # If the node for that host is not defined, just return the host parameter as system_name
+      system_name = host
+    end
+  end
+  system_name
+end
+
+# Get MAC address of system
+def get_mac_address(host)
```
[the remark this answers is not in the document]
srbarrios: I like your IDE but is out of scope for this PR

testsuite/features/support/commonlib.rb
srbarrios: It seems we handle those in next cases, in any case, I will not refactor this in this PoC.

testsuite/features/support/commonlib.rb
srbarrios: Maybe, but out of scope of this PR.

maximenoel8:
I like the changes ! So one question to be sure I understand. 
Now we will not open ssh sessions with all the nodes at the beginning but only when we want to interact with the node. And if we do more than one call during a feature, we will reuse the node ssh session correct ?
The node session will be close at the end of each Feature ( file.feature ) correct ?

maximenoel8:
Having run call with get_target('server') make sense but I'm a little bit surprise that we also create a ssh session to get the full_hostname.
Do we have situation where the get_target('server').full_hostname is different from ENV['SERVER'] ?

maximenoel8 (approved):

srbarrios:
> I like the changes ! Just few questions to be sure I understand. 
> Now we will not open ssh sessions with all the nodes at the beginning but only when we want to interact with one node ( awesome ). And if we do more than one call during a feature, we will reuse the node ssh session correct ?
> The node session will be close at the end of each Feature ( file.feature ) correct ? 
> 

First of all, clarify something, a twopence node instance will not keep a ssh session open, it just keep all the details so it can make ssh calls to a node.
Each ssh call is independent and will close the session after the reply.

And then, we use a global variable to define a hashmap that will contain the twopence node instances, so the nodes once created, will live during all test framework life, they will not be deleted per feature.

srbarrios:
> Having a method run with get_target('server') makes sense but I'm a little bit surprise that we also create a ssh session to get the full_hostname.
> Do we have situations where the get_target('server').full_hostname is different from ENV['SERVER'] ?
> If you don't want to have a direct ENV call in the features, we could create a function get_full_hostname('server').
> What do you think ?

The idea is to have all what we could need from a target contained in the Lavanda Object as attributes.
We could maybe safe one ssh call by reusing the info from the environement variable, that's fair, but also we could perfectly pass the IP instead of the fqdn in the environment variable... so that's not a fully reliable source to get the fqdn.

For now, I would not refactor that part, at least not as part of this PR. Because it could bring some unexpected corner cases.

testsuite/features/step_definitions/api_common.rb
```diff
@@ -14,9 +14,9 @@
              end
 
 $api_test = if $debug_mode
-              ApiTestXmlrpc.new($server.full_hostname)
+              ApiTestXmlrpc.new(get_target('server').full_hostname)
```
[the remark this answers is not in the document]
srbarrios: We use the Lavanda object to store the fqdn, so we only have a unique ssh call to obtain this info, during the whole life of this target.

PR #7342, opened 2023-08-02
Testsuite: Move repository creation and sync to reposync

elariekerboull:
## What does this PR change?

Uyuni CI has been failing lately on `init client` stage
image

This is caused by the product repositories not having been synced as expected:
image
Note the repository killing step being abnormally short and not showing any repo killed or left running, as well as the waiting step being 0 second.
This in turn is caused by the fact that the syncing did not start at that stage. Looking at logs and taskotop, it seems that taskomatic gets stuck for a very long time on the sync of the fake debian custom channel
image

The root cause for this is still unknown. 

This PR changes the order of the feature which syncs the problematic channel so it is synced last and so the CI can pass the init client stage.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Cucumber tests were added
- [x] **DONE**

## Links

Fixes #
Tracks # 
- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


nodeg (approved):
Fine with me.

PR #7368, opened 2023-08-04
Fix acceptance test running in podman

mackdk:
## What does this PR change?

This PR is meant to fix the acceptance test error that are currently happening when running the tests in the podman containers. These error are caused by a database upgrade SQL file `100-update-recurring-action.sql`, that was placed in the wrong folder (into `to-4.4.6` when 4.4.6 was already tagged). This means that this specific script is not part of package `susemanager-schema-4.4.6` and thus this change is not present in the database of the container.

This problem started to cause issue only when a new SQL script was added in the `to-4.4.7` folder,  because the bash  script `run_db_migration.sh` was just taking all the script from the last directory of the listing and applying them all. So the missing sql script was still applied by the script. 

This behaviour hid the wrong file placement, until the new directory was created. This PR also fixes this, by extracting the current version from the spec file and executing only the SQL files meant for the next version. I also added the upgrade of the reporting database, that was currently not updated to the newest version.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- No documentation needed: only test infrastructure changes

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/22151

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test


mackdk:
The SonarCloud issue is unrelated, it's already been fixed on master and will go away once rebased.

srbarrios (commented):
LGTM, but I don't seem the best person to approve this piece of code.

jordimassaguerpla (approved):

jordimassaguerpla:
thank you!

PR #7372, opened 2023-08-07
Check login page is correctly loaded

maximenoel8:
## What does this PR change?
During the BV testing, on some tests, the fields in the login page are empty but the testsuite is expecting to be log in. 
After investigation, it seems we fill the field too early. With this new step, we make sure the username is correctly loaded.

## Links

Related to : https://github.com/SUSE/spacewalk/issues/22176

- [x] **DONE**

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -532,6 +532,7 @@ Given(/^I am authorized as "([^"]*)" with password "([^"]*)"$/) do |user, passwd
 
   find(:xpath, "//header//i[@class='fa fa-sign-out']").click if all(:xpath, "//header//i[@class='fa fa-sign-out']", wait: 0).any?
 
+  raise 'Login page is not correctly loaded' unless has_field?('username')
   fill_in('username', with: user)
   fill_in('password', with: passwd)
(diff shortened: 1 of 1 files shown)


testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -532,6 +532,7 @@
 
   find(:xpath, "//header//i[@class='fa fa-sign-out']").click if all(:xpath, "//header//i[@class='fa fa-sign-out']", wait: 0).any?
 
+  raise 'Login page is not correctly loaded' unless has_field?('username')
```
vandabarata, line 535: what if we wait until the 'username' field loads instead of the raise? :thinking:

maximenoel8: has_XXX? method are waiting. The timeout will be the CapybaraTimeout. In PRV BV, it will be 60 secs.

vandabarata (approved):

nodeg (approved):
LGTM. Thank you, Maxime. I wonder if we could get completely rid of those issues (with Capybara) in the future.

elariekerboull (approved):

PR #7396, opened 2023-08-09
TEST with Leap 15.5 Marina's containers

jordimassaguerpla:
this is a TEST for new containers.

see https://github.com/SUSE/spacewalk/issues/22173


jordimassaguerpla:
This should not be merged. This is only for testing the new containers.

PR #7399, opened 2023-08-10
QE: Create step for database backup

nodeg:
## What does this PR change?

Since the switch from SLES to openSUSE Leap 15.4, we sync much more packages, which results in a bigger database. Moving from the default timeout of 250 s to 600 s could help solve failing in the finishing stage of our test suite. Furthermore we enable error checking to fail earlier.

See https://github.com/SUSE/spacewalk/issues/22277

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Cucumber tests were edited

- [x] **DONE**

## Links

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/finishing/srv_smdba.feature
+++ b/testsuite/features/finishing/srv_smdba.feature
@@ -58,11 +58,11 @@ Feature: SMDBA database helper tool
     Given a postgresql database is running
     And there is no such "/smdba-backup-test" directory
     When I create backup directory "/smdba-backup-test" with UID "root" and GID "root"
-    And I issue command "smdba backup-hot --enable=on --backup-dir=/smdba-backup-test"
+    And I take a backup with smdba in folder "/smdba-backup-test"
     Then I should see error message that asks "/smdba-backup-test" belong to the same UID/GID as "/var/lib/pgsql/data" directory
     And I remove backup directory "/smdba-backup-test"
     When I create backup directory "/smdba-backup-test" with UID "postgres" and GID "postgres"
-    And I issue command "smdba backup-hot --enable=on --backup-dir=/smdba-backup-test"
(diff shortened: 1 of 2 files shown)


srbarrios (approved):

Bischoff (changes requested):
We should not use "I issue command" at all - ideally, we completely remove this step.

Rather create an ad-hoc step for the thing you want to do, with a name that is explaining what is being accomplished. In that step, use `600` if needed.

srbarrios:
> We should not use "I issue command" at all - ideally, we completely remove this step.
> 
> Rather create an ad-hoc step for the thing you want to do, with a name that is explaining what is being accomplished. In that step, use `600` if needed.

I would first try to have a green in the testsuite, identifying if the issue is related with a raised timeout.
And in a second phase, refactor this step if necessary.

nodeg:
I would agree. The suggestion from @Bischoff is good, but I would do this in a future PR as an improvement for the test suite.

ktsamis (approved):

Bischoff:
> I would agree. The suggestion from @Bischoff is good, but I would do this in a future PR as an improvement for the test suite.

No, if you change it here, it will affect every feature that uses `I use command`.

Furthermore, it's easy to do: change 
```
And I issue command "smdba backup-hot --enable=on --backup-dir=/smdba-backup-test"
```
to
```
And I take a backup with smdba
```
and implement this new step: it's a one-liner, only simple copy and paste.

Bischoff (approved):
Thanks a lot Dominik. It's soo much better now!

maximenoel8 (approved):

ktsamis (approved):

srbarrios (approved):

PR #7401, opened 2023-08-10
Optimize org channel accessibility query (bsc#1211874)

nadvornik:
## What does this PR change?

This should improve the slow query on channel accessibility by org.
Main changes:
- COUNT replaced with EXISTS
- do not use views
- rhnOrgChannelTreeView uses union of separate SELECTs for parent and child channels. In this case, single SELECT is ok.
- handling of public channel family in rhnOrgChannelFamilyPermissions might be inefficient, not sure

Added unit test to make sure the query behaves the same as before.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Might fix https://bugzilla.suse.com/show_bug.cgi?id=1211874

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/channel/Channel.hbm.xml
+++ b/java/code/src/com/redhat/rhn/domain/channel/Channel.hbm.xml
@@ -600,12 +600,40 @@ PUBLIC "-//Hibernate/Hibernate Mapping DTD 3.0//EN"
     </sql-query>
 
     <sql-query name="Channel.isAccessibleBy">
-        <return-scalar type="int" column="count"/>
-        SELECT COUNT(*) AS count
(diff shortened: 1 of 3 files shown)


rjmateus (approved):
I think this can be provided to the customer as PTF. The implementation looks good, but since we have a customer to test it, it would give us good feedback.

lucidd (approved):
looks good just a clarification. I assume the test was run before and after the change and behaved the same right?

nadvornik:
> I assume the test was run before and after the change and behaved the same right?
yes

PR #7407, opened 2023-08-11
move /srv/www/htdoc/pub/repositories to /usr/share/susemanager/gpg/repositories

mbussolotto:
## What does this PR change?

Required by uyuni container, to avoid upgrade issue since ` /srv/www/htdoc/pub/repositories` is persisted

move /srv/www/htdoc/pub/repositories to /usr/share/susemanager/gpg/repositories

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager/empty-repo.conf
+++ b/susemanager/empty-repo.conf
@@ -1,0 +1,2 @@
+RewriteRule ^/pub/repositories/empty/(.*)$ /gpg/repositories/empty/$1  [L,PT]
+RewriteRule ^/pub/repositories/empty-deb/(.*)$ /gpg/repositories/empty-deb/$1  [L,PT]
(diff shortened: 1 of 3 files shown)


aaannz:
So if I understand this right, we need just empty repos there. The rest of repos will still be on the old location which in case of container will be a persistent volume, so the path `/srv/www/htdoc/pub/repositories` will be created by other means in the container.

But how about non containerized scenario? What package will create that path?

Does this work in an update scenario? Fortunately I think rpm will not remove those parent directories, because there will be files not owned by the package. Not sure scenario when update is done on unconfigured system, then rpm may very well remove `/srv/www/htdoc/pub/repositories` and we are back on my first question.

mbussolotto:
@aaannz, right. Probably the package should create the `/srv/www/htdocs/pub/repositories` (although empty), just to avoid issue on using and update it. 
Anyway, I'll put the PR in draft, we realized that might be better to move everything is owned by the package away from `/srv/www/`. I'll update the PR.

PR #7416, opened 2023-08-14
Fix action executor to prevent blocking Taskomatic for actions that are already finished

wweellddeerr:
## What does this PR change?

The server actions can be manually cancelled via Web UI, so this PR changes the code to check if all server actions have already finished before putting the Taskomatic thread to sleep. It should prevent blocking the Taskomatic for actions that will never appear in the database.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Port of https://github.com/SUSE/spacewalk/pull/22295

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
@@ -111,7 +111,8 @@ public class MinionActionExecutor extends RhnJavaJob {
         // HACK: it is possible that this Taskomatic task triggered before the corresponding Action was really
         // COMMITted in the database. Wait for some minutes checking if it appears
         int waitedTime = 0;
-        while (countQueuedServerActions(action) == 0 && waitedTime < ACTION_DATABASE_GRACE_TIME) {
+        while (countQueuedServerActions(action) == 0 && waitedTime < ACTION_DATABASE_GRACE_TIME &&
+                !allServerActionsFinished(action)) {
             action = ActionFactory.lookupById(actionId);
             try {
                 Thread.sleep(ACTION_DATABASE_POLL_TIME);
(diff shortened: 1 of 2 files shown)


java/code/src/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
```diff
@@ -192,4 +201,13 @@ private long countQueuedServerActions(Action action) {
                      .filter(serverAction -> ActionFactory.STATUS_QUEUED.equals(serverAction.getStatus()))
                      .count();
     }
+
+    private boolean allServerActionsFinished(Action action) {
+        return action != null &&
+            !CollectionUtils.isEmpty(action.getServerActions()) &&
+            action.getServerActions().stream().noneMatch(serverAction ->
```
lucidd, line 208: i think changing the condition here to all match completed or failed better matches the description of the function and is a bit easier to understand quickly.

java/code/src/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
```diff
@@ -192,4 +201,13 @@ private long countQueuedServerActions(Action action) {
                      .filter(serverAction -> ActionFactory.STATUS_QUEUED.equals(serverAction.getStatus()))
                      .count();
     }
+
+    private boolean allServerActionsFinished(Action action) {
+        return action != null &&
+            !CollectionUtils.isEmpty(action.getServerActions()) &&
```
lucidd, line 207: does this mean an action with no server actions does never count as having all server actions finished?

lucidd: i guess it makes sense given the `countQueuedServerActions(action) == 0` check already. I'm not exactly sure in what scenario an action would be saved and scheduled without server actions and then in a separate transaction server actions get added. It sounds like all kinds of problematic.

lucidd: i think its fine for now. The code suggests that there are cases where the action has no server actions initially. The whole action hand off between tomcat and taskomatic seems just very fragile so i think this needs some proper way in the future.

lucidd (approved):

PR #7419, opened 2023-08-16
Remove has_content? Systems when looking for new system

maximenoel8:
## What does this PR change?

Systems is always visible. Looking for Systems text and System overview will not wait correctly for the System Overview page to be loaded.

@elariekerboull , I would like to remove the Systems text check. For me this text is always visible on the Left Menu.

## Changelogs

- [x] No changelog needed


elariekerboull:
It's true `Systems` is always visible, didn't realise
However Uyuni does not have the `System Overview` text, we should find an alternative text for Uyuni
image

PR #7426, opened 2023-08-16
Add 'Stale PRs' GH action to label and close inactive community PRs

cbbayburt:
Add "Stale PRs" GH action to label and close inactive community PRs.

The action runs daily at 01:00 UTC to label any PR that is inactive for 60 days with `Stale` label.

After further 10 days, the action closes the PR and deletes its branch. The author has the chance to update the PR or simply remove the `Stale` label to unmark the PR for closing.

If needed, we can extend the action to issues as well.

Check out https://github.com/actions/stale for the full range of options this action supports.

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,28 @@
+name: Stale PRs
+on:
+  schedule:
+    - cron: '0 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Label/close stale PRs
+        uses: actions/stale@v8
+        with:
+          stale-pr-message: 'This PR is stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
+          close-pr-message: 'This PR was closed because it has been stalled for 10 days with no activity.'
+
+          # Stale after 60 days, close after further 10 days
+          days-before-pr-stale: 60
+          days-before-pr-close: 10
+
+          # Don't work on issues (only PRs)
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+
+          # Delete its branch after closing a PR
+          delete-branch: true


agraul:
In the announcement email / discussion for this we should probably clarify how "community" contributors can remove the "stale" label. Some users will know it from other GH repositories, but maybe not everyone does.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+          # Don't work on issues (only PRs)
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+
+          # Only consider 'community' PRs
+          only-labels: 'community'
+
+          # Delete its branch after closing a PR
+          delete-branch: true
```
renner, line 31: This might not be possible in case of a fork, I guess, but community PRs would probably mostly be created from forks? In any case I am not completely sure if we wanted to delete any branches automatically, that could be painful in case a PR is closed accidentally if for example a person is on a longer leave or similar.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Label/close stale PRs
+        uses: actions/stale@v8
+        with:
+          stale-pr-message: 'This PR is stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
```
renner, line 16: Do I understand correctly that when removing the stale label there will be another 60 days before this message will be added again?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+          # Stale after 60 days, close after further 10 days
+          days-before-pr-stale: 60
+          days-before-pr-close: 10
+
+          # Don't work on issues (only PRs)
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+
+          # Only consider 'community' PRs
+          only-labels: 'community'
```
renner, line 28: Do we really want to consider only community PRs? From my perspective we could include org members' PRs as well, so we also get a better feeling of how this automation "feels" if you are the author.

renner (approved):
Thanks a lot @cbbayburt, let's give it a try!

.github/workflows/stale.yml
cbbayburt: Otherwise, we'll have to deal with all the zombie branches manually and that's a bit frustrating because we'll need to ping people, read through a lot of commit history etc. to be able to safely delete them.

AFAIK, GH backs up deleted branch heads so it should be possible to restore them.

Forked branches won't be deleted with this action, but since they're not in our domain, we shouldn't care about them. This setting is only meant as a cleanup for branches created in our repo, so it's only relevant for PRs from us.

.github/workflows/stale.yml
cbbayburt: True, any kind of activity, including removing the label or simply commenting on the PR will reset the timer to 0.

PR #7466, opened 2023-08-26
GSOC 23 - A - Implement a fully functional CVE auditing feature based on OVAL data

HoussemNasri:
## What does this PR change?

This is the first pull request of my GSoC project. The primary goal of the PR is to implement and test the OVAL-based CVE auditing algorithm as described in the associated RFC. It also includes a partial (but sufficient given the OVAL files we aim to consume)  implementation of the OVAL definition schema specification.

This implementation addresses OVALs produced by SUSE, RedHat, and Debian. Ubuntu will be addressed in a follow-up pull request.

### TODO
- [ ] Replace `log.error()` with `log.debug()`
- [ ] Fix checkstyle errors

### What is missing
These are known limitations that will be addressed in subsequent pull requests.
- UI Integration (Python and React)
- Ubuntu support
- The downloading and synchronization of OVAL data
- Testing of the vulnerable package extractors

### Useful links
- [The RFC](https://github.com/uyuni-project/uyuni-rfc/pull/80)

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1451,6 +1451,7 @@ else if ("debian".equalsIgnoreCase(grains.getValueAsString("os"))) {
             server.setOsFamily(grains.getValueAsString("os_family"));
             server.setRunningKernel(grains.getValueAsString("kernelrelease"));
             server.setOs(grains.getValueAsString("osfullname"));
+            server.setCpe(grains.getValueAsString("cpe"));
```
HoussemNasri, line 1454: Asking for feedback: New systems will be assigned a CPE as part of the registration process, however, old systems must be migrated. The idea here is to set the CPE when users refresh their package list. The package refresh process can be slow, and users have to refresh the package list at least once for each of their systems to use the CVE auditing functionality. This is especially a concern for users with hundreds of systems.

parlt91: @agraul Could you or someone from the Ion squad have a look at Houssems proposal above.
Thank you

cbosdo:
You would need to add a changes file in a few folders. For java, just cd in the `java` folder and run `../rel-eng/bin/mkchlog -f system-update-fix -u <your username> "<your change description>"`. This will be added in the RPM changelog later. You need to run this in `java`, `schema/spacewalk`, `susemanager-utils/susemanager-sls` and `web`.

HoussemNasri:
> You would need to add a changes file in a few folders. For java, just cd in the java folder and run ../rel-> eng/bin/mkchlog -f system-update-fix -u <your username> "<your change description>". This will be added in the RPM > changelog later. You need to run this in java, schema/spacewalk, susemanager-utils/susemanager-sls and web. This will > get the Changelogs test to go green.

Done

> The SonarCloud analysis test will give your some potential security issues and bugs to look. May be some of the > security issues can just be marked as safe.

I didn't find any security issues or bugs, however, it complained about the usage of lowercase characters in enums (which I fixed) and the usage of `EnumSet` instead of plain `Set`. I'm not sure if the latter is a good idea... `EnumSet` was used in the old implementation in `CVEAuditManager`  to pass the patch statuses to scan for just fine, and I don't think using `Set` will achieve anything.

image


> A lot of the Java unit tests is failing because of org.postgresql.util.PSQLException: ERROR: column "cpe" of relation > "rhnserver" does not exist

Oops, I added the cpe change to a migration script but forgot to add it to the `rhnServer` table schema definition. Hopefully, the tests will pass now.

cbosdo:
> > The SonarCloud analysis test will give your some potential security issues and bugs to look. May be some of the > security issues can just be marked as safe.
> 
> I didn't find any security issues or bugs, however, it complained about the usage of lowercase characters in enums (which I fixed) and the usage of `EnumSet` instead of plain `Set`. I'm not sure if the latter is a good idea... `EnumSet` was used in the old implementation in `CVEAuditManager` to pass the patch statuses to scan for just fine, and I don't think using `Set` will achieve anything.

The `EnumSet` issue is not the one that needs fixing I think, however using an interface rather than an implementation of it in a function signature would be better.
You can have a look at the code smells to try to reduce them, but if those are not all fixed that is not a big issue:
https://sonarcloud.io/project/issues?id=uyuni-project_uyuni&pullRequest=7466&resolved=false&types=CODE_SMELL

Here are the issues and security hot spots lists:
* https://sonarcloud.io/project/issues?id=uyuni-project_uyuni&pullRequest=7466&resolved=false&types=BUG
* https://sonarcloud.io/project/security_hotspots?id=uyuni-project_uyuni&pullRequest=7466&resolved=false&types=SECURITY_HOTSPOT

cbosdo:
> Oops, I added the cpe change to a migration script but forgot to add it to the `rhnServer` table schema definition. Hopefully, the tests will pass now.

You still have about 20 other unit test failures in `CVEAuditManagerOVALTest`. Those I looked at seem to boil down to this error:
```
com.redhat.rhn.common.db.WrappedSQLException: Batch entry 0 call insert_product_vulnerable_packages('kernel-debug-base', '0:4.12.14-150100.197.137.2', 'cpe:/o:opensuse:leap:15.4', 'CVE-2022-2991') was aborted: ERROR: procedure insert_product_vulnerable_packages(character varying, character varying, character varying, character varying) does not exist
  Hint: No procedure matches the given name and argument types. You might need to add explicit type casts.
  Position: 6  Call getNextException to see other errors in the batch.
	at com.redhat.rhn.common.translation.SqlExceptionTranslator.sqlException(SqlExceptionTranslator.java:39)
	at com.redhat.rhn.common.db.NamedPreparedStatement.executeBatch(NamedPreparedStatement.java:243)
	at com.redhat.rhn.common.db.datasource.CachedStatement.executeBatch(CachedStatement.java:1002)
	at com.redhat.rhn.common.db.datasource.CachedStatement.lambda$executeBatchUpdates$6(CachedStatement.java:974)
```

Did you know that you could run the unit tests on your machine? This link should help you, and I hope its content is not too outdated:
https://github.com/uyuni-project/uyuni/wiki/Java-Development-Environment#optional-running-junit-tests

In your case you would ideally just run the failing unit test to fix it: this would save you a lot of time.

java/code/src/com/suse/oval/OVALCleaner.java
```diff
@@ -0,0 +1,185 @@
...
+        definition.setOsVersion(osVersion);
+    }
+
+    private static void doCleanupTest(TestType test, OsFamily osFamily, String osVersion) {
+        if (osFamily == OsFamily.DEBIAN) {
+            test.setId(convertDebianId(test.getId(), osVersion));
+            test.setObjectRef(convertDebianId(test.getObjectRef(), osVersion));
+            if (test.getStateRef().isPresent()) {
+                test.setStateRef(convertDebianId(test.getStateRef().get(), osVersion));
+            }
```
HoussemNasri, line 129: > SonarCloud: Call "Optional#isPresent()" before accessing the value. 

This is a false positive bug because `test.getStateRef().isPresent()` is being called before calling `test.getStateRef().get()`

HoussemNasri:
> You still have about 20 other unit test failures in `CVEAuditManagerOVALTest`

They all pass locally, but `ErrataFactoryTest#testAddToChannel` fails and throws the following exception `ERROR com.redhat.rhn.common.conf.Config - Could not parse file /usr/share/rhn/config-defaults`. I followed the instructions in the docs and created the `config-defaults` folder but apparently, as you said, it's outdated.

java/code/src/com/suse/oval/OVALCleaner.java
HoussemNasri: I rewrote the code using `ifPresent(stateRef -> ... )` and now SonarCloud doesn't detect it as a bug anymore.

susemanager-utils/susemanager-sls/src/grains/cpe_grain.py
```diff
@@ -0,0 +1,78 @@
...
+
+    return ret
+
+
+def _derive_cpe(grains):
+    """
+    Try to derive the CPE of the system based on the collected core grains.
+
+    PS: This function is not guaranteed to derive the correct CPE as there could be a many 
+    variances of the same OS that require different CPEs, for example, release and beta versions.
```
agraul, line 63: Do you have example os-release files that show the differences you're accounting for? Those would be good to test this code.

agraul (commented):
I think the grains code has a place in grains/core.py upstream. They don't want to add more to it, but given that the heavy lifting is already done, we might have a chance to "just return one more grain".

Other than that, the code looks good. I'd appreciate tests based on different os-release files.

susemanager-utils/susemanager-sls/src/grains/cpe_grain.py
HoussemNasri: This is an example of [RedHat 8 CPE](https://github.com/chef/os_release/blob/main/redhat_8), notice the 'GA' at the end. And this is the standard list of vulnerable RedHat 8 CPEs found in OVAL files:
```xml
   <affected_cpe_list>
    <cpe>cpe:/a:redhat:enterprise_linux:8</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::appstream</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::crb</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::highavailability</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::nfv</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::realtime</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::resilientstorage</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::sap</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::sap_hana</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::supplementary</cpe>
    <cpe>cpe:/o:redhat:enterprise_linux:8</cpe>
    <cpe>cpe:/o:redhat:enterprise_linux:8::baseos</cpe>
   </affected_cpe_list>
```

And this is [Ubuntu 22.04 CPE](https://nvd.nist.gov/products/cpe/detail/8DE10C8A-477C-42FF-8F96-D8813B729D3A?namingFormat=2.3&orderBy=CPEURI&keyword=cpe%3A2.3%3Ao%3Acanonical%3Aubuntu_linux%3A*%3A*%3A*%3A*%3Alts%3A*%3A*%3A*&status=FINAL%2CDEPRECATED) with an LTS at the end. 

The code actually doesn't account for these differences. Since OVAL files for Ubuntu and Debian don't contain CPEs we derive them in the backend. For systems to be matched successfully, the derived CPE in salt must match the derived CPE in the backend.
https://github.com/uyuni-project/uyuni/blob/513ec82697c689d2fc638eaf1f614bc494f85dbb/java/code/src/com/suse/oval/vulnerablepkgextractor/DebianVulnerablePackagesExtractor.java#L82-L90

susemanager-utils/susemanager-sls/src/grains/cpe_grain.py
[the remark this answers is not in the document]
HoussemNasri: Sounds good, so I need to open a PR in [openSUSE/salt](https://github.com/openSUSE/salt)?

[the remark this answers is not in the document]
agraul: Even better would be in the upstream repository [saltstack/salt](https://github.com/saltstack/salt/). We try to be aligned as much as possible with upstream Salt. It can take a while for them to review and merge PRs, so once we know they like the proposed change we can include it in openSUSE/salt.

java/code/src/com/suse/oval/OVALCachingFactory.java
```diff
@@ -53,6 +53,8 @@ private OVALCachingFactory() {
      *
      * @param rootType the OVAL root to extract from
      * */
+    // TODO: Following the single responsibility principle this method shouldn't have to extract vulnerable packages
```
parlt91, line 55: If you don't plan to address this in this PR please add a card for it. We should try to avoid merging the code with leftover todos.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -251,38 +251,38 @@ class CVEAudit extends React.Component<Props, State> {
                     </div>
                     <div id={`collapse-${status}`} className="panel-collapse collapse in">
                       <div className="panel-body">
-                        The client is affected by a vulnerability that received a patch, but Uyuni is unable to locate any of the patches in
-                        relevant channels.
+                        The client is affected by a vulnerability that received a patch, but Uyuni is unable to locate
```
parlt91, line 281: This message should be translated using the `t()` function

parlt91 (commented):
The code looks pretty good to me already. Most of the comments left below are minor things and nitpicks. Great job :)

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
```diff
@@ -0,0 +1,60 @@
...
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+--
+-- Red Hat trademarks are not licensed under GPLv2. No permission is
+-- granted to use or replicate Red Hat trademarks that are incorporated
+-- in this software or its documentation.
+--
+
+CREATE OR REPLACE PROCEDURE
```
rjmateus, line 16: Is there any reason to have this logic in PSQL instead of Java? What are the advantages of having i there?
We are trying to have logic in store procedures/functions because it make it much harder to test and debug.

java/code/src/com/redhat/rhn/domain/errata/Cve.java
```diff
@@ -48,4 +52,28 @@ public void setName(String nameIn) {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean equals(final Object other) {
+        if (!(other instanceof Cve)) {
+            return false;
+        }
+        Cve castOther = (Cve) other;
+        return new EqualsBuilder()
```
rjmateus, line 65: I'm not sure about this equals implementation. If the names are different is object equal?
We should not confuse the same DB entry with the Java object being equals. Hibernate uses the equals method to check if an object needs to be flushed to the database, so I think we need to also compare the name here.

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/002-store-vulnerable-packages.sql
```diff
@@ -0,0 +1,68 @@
+CREATE TABLE IF NOT EXISTS suseOVALVulnerablePackage
```
rjmateus, line 1: this table looks to be duplicated from the file 001-oval-add-tables.sql

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/002-store-vulnerable-packages.sql
```diff
@@ -0,0 +1,68 @@
...
+        CONSTRAINT suse_oval_vulnerable_pkg_id_pk PRIMARY KEY,
+    name           VARCHAR NOT NULL,
+    fix_version    VARCHAR
+);
+
+CREATE SEQUENCE IF NOT EXISTS suse_oval_vulnerable_pkg_id_seq START WITH 301;
+
+CREATE UNIQUE INDEX IF NOT EXISTS suse_oval_vulnerable_pkg_name_fix_version ON suseOVALVulnerablePackage(name, fix_version);
+
+CREATE TABLE IF NOT EXISTS suseOVALPlatformVulnerablePackage
```
rjmateus, line 13: Some about this table, looks duplicated.

schema/spacewalk/common/tables/suseOVALPlatform.sql
```diff
@@ -0,0 +1,27 @@
...
+
+CREATE TABLE suseOVALPlatform
+(
+    id        NUMERIC NOT NULL
+                 CONSTRAINT suse_oval_platform_id_pk PRIMARY KEY,
+    cpe      VARCHAR
+);
+
+
+CREATE SEQUENCE suse_oval_platform_id_seq START WITH 101;
```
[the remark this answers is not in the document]
rjmateus: That is strange and should be a signal that something is not in good space. We should at least understand why the number is growing.

rjmateus (commented):
First round of review. I will continue in the afternoon

web/html/src/branding/css/base/cveaudit.less
```diff
@@ -0,0 +1,58 @@
...
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+#accordion {
```
Etheryte, line 16: I'm not convinced this is a good idea, this way we end up having a lot of very specific styles that are only used in this specific view. The overrides here are very specific and that makes them fragile. If the underlying panel styles change, something here will surely get out of alignment as well.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -209,35 +230,62 @@ class CVEAudit extends React.Component<Props, State> {
...
-                      className={"fa fa-big " + PATCH_STATUS_LABEL[status].className}
-                      title={PATCH_STATUS_LABEL[status].label}
-                    />
-                    <span>{" " + PATCH_STATUS_LABEL[status].label}</span>
-                  </label>
-                </div>
-              );
-            })}
+          <div className="" style={{ maxWidth: "770px", marginTop: "16px" }}>
+            <div className="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
```
Etheryte, line 234: While the accordion pattern can be useful in many situations, I don't think this specific one works well. The core problem is that traditionally, clicking on a label element should toggle the relevant input, in this case the checkbox. Here it instead toggles the accordion, so it doesn't match user expectations.

I would prefer to move the descriptions into tooltips since they're pretty short and then show exactly as the user interacts with the element, that way we can also keep the default expected behavior of the label click modifying the checkbox. What do you think @parlt91?

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -0,0 +1,295 @@
...
+
+        boolean allPackagesUnpatched = unpatchedVulnerablePackages.size() == clientProductVulnerablePackages.size();
+
+        if (allPackagesUnpatched) {
+            cveAuditServerBuilder.setPatchStatus(PatchStatus.AFFECTED_PATCH_UNAVAILABLE);
+        }
+        else {
+            boolean allPackagesPatched = patchedVulnerablePackages.stream().allMatch(patchedPackage ->
+                    allInstalledPackages.stream()
+                        .filter(installedPackage ->
```
rjmateus, line 172: This filter is similar to the one at line 146. Should we use the same method?

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -0,0 +1,295 @@
...
+        // It's possible to have 2 or more patches for one package. It's necessary to apply all of them because
+        // they will have the same outcome i.e. patch the package; instead we need to choose only one.
+        // To choose the one, we rank patches based on the channel they come from .e.g.
+        // assigned, successor product, etc. And for each vulnerable package we keep only the highest ranking patch
+        results = keepOnlyPatchCandidates(results);
+
+        CVEAuditSystemBuilder cveAuditServerBuilder = new CVEAuditSystemBuilder(clientServer.getId());
+        cveAuditServerBuilder.setSystemName(clientServer.getName());
+
+        List<ShallowSystemPackage> allInstalledPackages =
```
rjmateus, line 138: If we only care about installed packages that are affected by the CVE should we load only the packages affected but the CVE that are installed, and this way avoid loading all packages to memory?

java/code/src/com/suse/oval/OVALCachingFactory.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Extracts and save the list of vulnerable packages from {@code rootType}
+     *
+     * @param rootType the OVAL root to extract from
+     * */
+    // TODO: Following the single responsibility principle this method shouldn't have to extract vulnerable packages
+    //  itself
+    public static void savePlatformsVulnerablePackages(OvalRootType rootType) {
```
rjmateus, line 57: I may be missing something in here, but where is this method called outside the unit tests?

java/code/src/com/suse/oval/OvalParser.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Parse the given OVAL file
+     *
+     * @param ovalFile the OVAL file to parse
+     * @return the parsed OVAL encapulated in a {@link OvalRootType} object=
+     * */
+    public OvalRootType parse(File ovalFile) throws OvalParserException {
+        try {
+            JAXBContext jaxbContext = JAXBContext.newInstance(OvalRootType.class);
```
rjmateus, line 42: I think it should be using "StAX" instead of Jaxb, at least according to the issue: https://github.com/uyuni-project/uyuni/issues/7108

Also, I'm probably missing something but where is this class used outside tests?

java/code/src/com/suse/oval/OVALCleaner.java
```diff
@@ -0,0 +1,183 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * This class is responsible for cleaning OVAL resources and filling up missing data. It acts as an adapter that takes
+ * OVAL data from multiple sources and make changes to it to have a more predictable format.
+ */
+public class OVALCleaner {
```
rjmateus, line 42: I'm probably missing something, but where is this class used outside the test classes?

rjmateus (commented):
A few more questions
I would need to look in more detail to find the place where the oval files are processed from upstream.

java/code/src/com/suse/oval/OVALCleaner.java
rjmateus: I think I found the reason, It's because the download and automatic extract is not yet implemented?

java/code/src/com/suse/oval/OvalParser.java
rjmateus: I think I found the reason for not being called outside the tests. Is it because the download and automatic extract are not yet implemented?

java/code/src/com/suse/oval/OVALCachingFactory.java
rjmateus: I think I found the reason for not being called outside the tests. Is it because the download and automatic extract are not yet implemented?

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
HoussemNasri: The primary reason is performance, this way, we can call procedures in batches. Even with this optimization, writing OVAL data to the database from a ~400 MB file takes 5 minutes on my machine, so it's kind of necessary.

rjmateus: Have you tried to use hibernate and flush the inserts after a certain number of inserts (the batch size)? What was the performance degradation on that?
We should consider performance vs maintenance of the code, and to do that we would need to compare performance times for each solution.

java/code/src/com/suse/oval/cpe/Cpe.java
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * Converts and returns the Cpe as plain text URI
+     *
+     * @return the cpe as URI
+     * */
+    public String asString() {
+        String cpe = "cpe:/o:" + vendor + ":" + product + ":" + version + ":" + update;
+
+        // Removing trailing colons ':'
+        return cpe.replaceAll(":*$", "");
```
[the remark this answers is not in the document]
HoussemNasri: We remove the trailing `:` notice the `$` in the regex, not the ones in the middle. 

There is a chance that I'm misusing the word 'trailing' here.

java/code/src/com/suse/oval/OVALCachingFactory.java
```diff
@@ -0,0 +1,145 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        mode.getQuery().executeBatchUpdates(new DataResult<>(batch));
+
+        LOG.warn("Ending...");
```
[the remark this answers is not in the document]
HoussemNasri: I agree, debug should be fine.

java/code/src/com/suse/oval/vulnerablepkgextractor/VulnerablePackagesExtractors.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A factory for {@link VulnerablePackagesExtractor}
+ * */
+public class VulnerablePackagesExtractors {
+
+    public static VulnerablePackagesExtractor create(DefinitionType definition, OsFamily osFamily,
+                                                     OVALLookupHelper ovalLookupHelper) {
+        switch (osFamily) {
+            case openSUSE_LEAP:
```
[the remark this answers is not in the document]
HoussemNasri: Hmm I thought I already changed it to LEAP

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -0,0 +1,310 @@
...
+                        if (allPackagesHavePatchInSuccessorChannel) {
+                            cveAuditServerBuilder
+                                    .setPatchStatus(PatchStatus.AFFECTED_PATCH_INAPPLICABLE_SUCCESSOR_PRODUCT);
+                        }
+                        else {
+                            cveAuditServerBuilder.setPatchStatus(PatchStatus.AFFECTED_PATCH_INAPPLICABLE);
+                        }
+                    }
+                    else if (somePackagesHavePatchInUnassignedChannels) {
+                        //TODO: Not sure how to handle...
```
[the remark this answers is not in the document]
HoussemNasri: Keep in mind that `AFFECTED_PATCH_INAPPLICABLE` means a full patch is available, the same name was kept for backward compatibility but is not completely accurate. But given the context, I do agree that `AFFECTED_PATCH_INAPPLICABLE` is our best option here. Would like to hear what @admd has to say about this.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -0,0 +1,310 @@
...
+     * @param user the calling user
+     * @param cveIdentifier the CVE identifier to lookup
+     * @param patchStatuses the patch statuses
+     * @return list of images records with patch status
+     * @throws UnknownCVEIdentifierException if the CVE number is not known
+     */
+    public static List<CVEAuditImage> listImagesByPatchStatus(User user,
+                                                              String cveIdentifier, EnumSet<PatchStatus> patchStatuses)
+            throws UnknownCVEIdentifierException {
+        // TODO: Audit images with OVAL
```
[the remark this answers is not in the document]
HoussemNasri: Currently, it's not. Images are audited with channels only. But I left a TODO for the future.

java/code/src/com/redhat/rhn/manager/audit/test/CVEAuditManagerOVALTest.java
```diff
@@ -0,0 +1,387 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+// TODO: Test for AFFECTED_PATCH_INAPPLICABLE_SUCCESSOR_PRODUCT
+// TODO: Test that if we get AFFECTED_PATCH_INAPPLICABLE auditServer.Channels and auditServer.Erratas are not null
```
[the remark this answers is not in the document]
HoussemNasri: I think the first one is already taken care of, the second one only needs to be handled.

java/code/src/com/suse/oval/OVALCachingFactory.java
HoussemNasri: Ah okay, I use TODOs to not forget about things, but you're right I don't plan to work on this in the current PR.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
HoussemNasri: Yep, this was just for the demo, the message also needs to vary between patch statuses. Right now, it shows the same description for all of them.

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
HoussemNasri: Yes, I tried Hibernate while flushing after inserting a batch and the performance was horrible, even with native queries. See [this](https://github.com/uyuni-project/uyuni-rfc/pull/80#discussion_r1257750349) for more details.

schema/spacewalk/common/tables/suseOVALPlatform.sql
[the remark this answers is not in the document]
HoussemNasri: I think I found the reason. Apparently `nextval('suse_oval_platform_id_seq')` is evaluated even when the platform/cpe we're trying to insert already exists (conflict detected). One solution is to not use `ON CONFLICT` and check for the existence of a platform with the same CPE before attempting to insert one.

```sql
INSERT INTO suseovalplatform(id, cpe)
VALUES (nextval('suse_oval_platform_id_seq'), product_cpe_in)
ON CONFLICT(cpe) DO NOTHING;
```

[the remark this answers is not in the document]
HoussemNasri: Or we can save the old value of the sequence using `lastval('suse_oval_platform_id_seq')` and reset it on conflict using `setval('suse_oval_platform_id_seq', saved_last_val)`

java/code/src/com/suse/oval/manager/OvalObjectManager.java
```diff
@@ -0,0 +1,41 @@
...
+    public OvalObjectManager(List<ObjectType> objects) {
+        for (ObjectType objectType : objects) {
+            objectsMap.put(objectType.getId(), objectType);
+        }
+    }
+
+    public ObjectType get(String objectId) {
+        ObjectType object = objectsMap.get(objectId);
+        if (object == null) {
+            throw new IllegalArgumentException("The object id is invalid: " + objectId);
```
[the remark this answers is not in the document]
HoussemNasri: I agree.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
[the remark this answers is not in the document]
HoussemNasri: Note to self: Remove the TODO

java/code/src/com/suse/oval/OsFamily.java
```diff
@@ -0,0 +1,38 @@
...
+    // Should consist of all lower case characters
+    private final String shortname;
+
+
+    OsFamily(String fullname, String shortname, String vendor) {
+        this.fullname = fullname;
+        this.shortname = shortname;
+        this.vendor = vendor;
+    }
+    OsFamily(String fullname, String vendor) {
```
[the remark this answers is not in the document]
HoussemNasri: True, but there have been changes to this in PR B, so to not make it hard to resolve conflicts I'll keep it for now and handle it in PR B.

java/code/src/com/suse/oval/exceptions/OvalParserException.java
```diff
@@ -0,0 +1,19 @@
+package com.suse.oval.exceptions;
+
+public class OvalParserException extends RuntimeException {
+    public OvalParserException() {
```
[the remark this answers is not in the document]
HoussemNasri: One constructor was not used in this PR but is used in PR B. I'm going to keep it to reduce conflicts.

java/code/src/com/suse/oval/ovaltypes/linux/DpkginfoState.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.suse.oval.ovaltypes.StateType;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlType;
+
+
+/**
+ *
```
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/redhat/rhn/domain/errata/Cve.java
HoussemNasri: Done.

schema/spacewalk/common/tables/suseOVALVulnerablePackage.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE suseOVALVulnerablePackage
+(
+    id             NUMERIC NOT NULL
+                      CONSTRAINT suse_oval_vulnerable_pkg_id_pk PRIMARY KEY,
+    name           VARCHAR NOT NULL,
+    fix_version    VARCHAR
+);
+
+
+CREATE SEQUENCE suse_oval_vulnerable_pkg_id_seq START WITH 301;
```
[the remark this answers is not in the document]
HoussemNasri: See above. It behaves normally because, in the procedure definition that inserts vulnerable packages, I used an if statement instead of ON CONLIFCT clause to avoid duplicates.

schema/spacewalk/common/tables/suseOVALPlatform.sql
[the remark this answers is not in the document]
HoussemNasri: Fixed the growing problem. However, the id still starts from 101, I'll update it later.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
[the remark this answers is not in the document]
HoussemNasri: I set it to `AFFECTED_PATCH_INAPPLICABLE`.

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/002-store-vulnerable-packages.sql
HoussemNasri: Nice catch, this should be fixed now.

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/002-store-vulnerable-packages.sql
HoussemNasri: This too.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
parlt91: I agree with what you mentioned that we should stick with the behavior that clicking on the text ticks the checkboxes. We use tooltips to display descriptions all over the UI so that doesn't sound like a bad solution to me either.

java/code/src/com/suse/manager/utils/SaltUtils.java
HoussemNasri: Alexander responded in another comment. I'm resolving this.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
HoussemNasri: @Etheryte @parlt91 Thanks for the review and suggestions. I removed the accordion and replaced it with showing a tooltip message per your recommendations, but I thought it would be better to display the tooltip message for each client. Although the messages are static I think it would be more convenient for users to hover over the patch status on one client to learn more about it instead of scrolling up and down.

image

web/html/src/branding/css/base/cveaudit.less
HoussemNasri: The accordion was removed altogether so I'm resolving this.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
HoussemNasri: Done. I'm assuming the generation of the translations will be done automatically.

schema/spacewalk/common/tables/suseOVALVulnerablePackage.sql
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -0,0 +1,295 @@
...
+        // assigned, successor product, etc. And for each vulnerable package we keep only the highest ranking patch
+        results = keepOnlyPatchCandidates(results);
+
+        CVEAuditSystemBuilder cveAuditServerBuilder = new CVEAuditSystemBuilder(clientServer.getId());
+        cveAuditServerBuilder.setSystemName(clientServer.getName());
+
+        List<ShallowSystemPackage> allInstalledPackages =
+                PackageManager.shallowSystemPackageList(clientServer.getId());
+
+        LOG.error("Vul packages before filtering: {}",
```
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -0,0 +1,295 @@
...
+                OVALCachingFactory.getVulnerablePackagesByProductAndCve(clientServer.getCpe(), cveIdentifier));
+
+        Set<VulnerablePackage> clientProductVulnerablePackages =
+                OVALCachingFactory.getVulnerablePackagesByProductAndCve(clientServer.getCpe(), cveIdentifier).stream()
+                        .filter(pkg -> isPackageInstalled(pkg, allInstalledPackages))
+                        .collect(Collectors.toSet());
+
+        LOG.error("Vul packages: {}", clientProductVulnerablePackages);
```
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -0,0 +1,295 @@
...
+                        cveAuditServerBuilder.setPatchStatus(PatchStatus.AFFECTED_PATCH_UNAVAILABLE_IN_UYUNI);
+                    }
+                    else {
+                        cveAuditServerBuilder.setPatchStatus(PatchStatus.AFFECTED_PATCH_UNAVAILABLE_IN_UYUNI);
+                    }
+                }
+            }
+        }
+
+        LOG.error("Patch Status: {}", cveAuditServerBuilder.getPatchStatus());
```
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
HoussemNasri: Nice catch but I don't think we can use the same method. In line 172, in addition to checking if a certain package is installed we also need to access the list of installed versions to be able to assert that all installed versions of the package are above the patched version, which is not possible with `isPackageInstalled`. I created another method to do exactly that.

Reading through the code at CVEAuditManagerOVAL after a while since I last touched it wasn't straightforward; it is not as clear as I hoped for. 

Note to self: Create smaller methods to abstract away common operations and make the code clearer.

java/code/src/com/redhat/rhn/manager/audit/test/CVEAuditManagerOVALTest.java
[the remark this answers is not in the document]
HoussemNasri: Done.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
HoussemNasri: I have a question: any idea how to replace Uyuni with SUSE Manager when the SUSE Manager theme is used?

HoussemNasri: Thanks, I rephrased the tooltip messages to not mention Uyuni.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
HoussemNasri: I'm all for the idea. There is already a query in PR D that could help in that regard. I think we can delay this until PR D is reviewed and take advantage of that query, this is especially useful to not create confusing conflicts with changes in the other PRs.


Query in PR D:
https://github.com/uyuni-project/uyuni/blob/4f5a7f459e7ff0e9577698c3f5a76e910d656790/java/code/src/com/redhat/rhn/common/db/datasource/xml/oval_queries.xml#L33

Etheryte (approved):
Frontend looks good.

rjmateus (commented):
Just a few questions on the sql scripts and a store procedure. Other then that looks good. great work on this one :)

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
```diff
@@ -0,0 +1,63 @@
...
+$$
+DECLARE
+    cve_id_val numeric;
+    product_cpe_id_val numeric;
+    vulnerable_pkg_id_val numeric;
+begin
+
+    IF NOT exists(SELECT cve FROM rhnCve cve WHERE cve.name = cve_name_in) THEN
+        INSERT INTO rhncve(id, name)
+        VALUES (nextval('rhn_cve_id_seq'), cve_name_in);
```
mcalmer, line 28: why not using `lookup_cve(cve_name_in)` - this is handling also creating of the value if it does not exist.

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
```diff
@@ -0,0 +1,63 @@
...
+    END IF;
+
+    SELECT id INTO cve_id_val FROM rhncve WHERE name = cve_name_in;
+
+    IF NOT exists(SELECT c FROM suseOVALPlatform c WHERE cpe = product_cpe_in) THEN
+        INSERT INTO suseovalplatform(id, cpe)
+        VALUES (nextval('suse_oval_platform_id_seq'), product_cpe_in);
```
mcalmer, line 35: I would suggest to create also a lookup_oval_platform function following the same ideas we already use for other tables.

schema/spacewalk/common/tables/suseOVALPlatform.sql
```diff
@@ -0,0 +1,27 @@
...
+    id        NUMERIC NOT NULL
+                 CONSTRAINT suse_oval_platform_id_pk PRIMARY KEY,
+    cpe      VARCHAR
+);
+
+
+CREATE SEQUENCE suse_oval_platform_id_seq;
+
+CREATE UNIQUE INDEX suse_oval_aff_platform_cpe_uq
+    ON suseovalplatform(cpe);
```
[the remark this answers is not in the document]
HoussemNasri: Nice catch, cpe should be NOT NULL (like the 'name' column in the rhnCve table) so I'll update the suseOVALPlatform table definition.

schema/spacewalk/common/tables/suseOVALVulnerablePackage.sql
```diff
@@ -0,0 +1,27 @@
...
+    id             NUMERIC NOT NULL
+                      CONSTRAINT suse_oval_vulnerable_pkg_id_pk PRIMARY KEY,
+    name           VARCHAR NOT NULL,
+    fix_version    VARCHAR
+);
+
+
+CREATE SEQUENCE suse_oval_vulnerable_pkg_id_seq;
+
+CREATE INDEX suse_oval_vulnerable_pkg_name_fix_version ON suseOVALVulnerablePackage(name, fix_version);
```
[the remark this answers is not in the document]
HoussemNasri: Using 'name' only in the index didn't introduce any noticeable performance degradations; it was actually slightly faster. But I'm curious why is fix_version being null and part of the index is a bad thing?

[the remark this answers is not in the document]
rjmateus: Null's in an index that has unique are considered different values. In this case, since it's not a unique index it will not be a big problem. The only problem is if someone change this index to unique.

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/002-store-vulnerable-packages.sql
```diff
@@ -0,0 +1,48 @@
+CREATE OR REPLACE PROCEDURE
+    insert_product_vulnerable_packages(package_name_in varchar,fix_version_in varchar,product_cpe_in varchar,cve_name_in varchar)
+AS
```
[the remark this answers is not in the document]
rjmateus: ah, you are right. We already had this discussion :)
Go on with it, but please look at MC comments about the look_up procedures

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
HoussemNasri: I didn't know about it. But now that I look into it and know that it calls `insert_cve`, it will lead to CVE IDs growing exponentially, because of the calling of `nextval('rhn_cve_id_seq')` even when CVE already exists.

This is an [instance of the problem](https://github.com/uyuni-project/uyuni/pull/7466#discussion_r1366811006) that is now fixed in the suseOVALPlatform table.

image

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
HoussemNasri: Done.

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
mcalmer: But insert is only called when a select failed because the entry does not exist

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/002-store-vulnerable-packages.sql
```diff
@@ -0,0 +1,61 @@
...
+$$
+DECLARE
+    cve_id_val numeric;
+    product_cpe_id_val numeric;
+    vulnerable_pkg_id_val numeric;
+begin
+
+    IF NOT exists(SELECT cve FROM rhnCve cve WHERE cve.name = cve_name_in) THEN
```
rjmateus, line 29: and about the CVE lookup?

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
```diff
@@ -0,0 +1,63 @@
...
+begin
+
+    IF NOT exists(SELECT cve FROM rhnCve cve WHERE cve.name = cve_name_in) THEN
+        INSERT INTO rhncve(id, name)
+        VALUES (nextval('rhn_cve_id_seq'), cve_name_in);
+    END IF;
+
+    SELECT id INTO cve_id_val FROM rhncve WHERE name = cve_name_in;
+
+    IF NOT exists(SELECT c FROM suseOVALPlatform c WHERE cpe = product_cpe_in) THEN
```
rjmateus, line 33: I think you forgot forgot to replace this code with the call to `lookup_oval_platform`

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/002-store-vulnerable-packages.sql
HoussemNasri: Done.

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
HoussemNasri: Done.

schema/spacewalk/postgres/procs/insert_product_vulnerable_packages.sql
HoussemNasri: Ah okay, thank you for clarifying; I made the change.

java/code/src/com/suse/oval/manager/OVALLookupHelper.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Looks up an OVAL test with an id of {@code testId}
+     *
+     * @param testId the test id to look up
+     * @return the cached {@link TestType} object that correspond to the test id
+     * */
+    public Optional<TestType> lookupTestById(String testId) {
+        // TODO: testManager#get throws an exception if testId is invalid
```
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/suse/oval/manager/OvalStateManager.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Looks up an OVAL state with an id of {@code stateId} or throws an exception if none is found.
+     *
+     * @param stateId the id of state to lookup
+     * @return the state
+     * */
+    public StateType get(String stateId) {
+        StateType state = statesMap.get(stateId);
+        if (state == null) {
+            throw new IllegalArgumentException("The state id is invalid: " + stateId);
```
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/suse/oval/manager/OvalObjectManager.java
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/suse/oval/manager/OvalTestManager.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Looks up an OVAL test with an id of {@code testId} or throws an exception if none is found.
+     *
+     * @param testId the id of test to lookup
+     * @return the test
+     * */
+    public TestType get(String testId) {
+        TestType test = testsMap.get(testId);
+        if (test == null) {
+            throw new IllegalArgumentException("The test id is invalid: " + testId);
```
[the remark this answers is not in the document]
HoussemNasri: Done.

schema/spacewalk/common/tables/suseOVALPlatform.sql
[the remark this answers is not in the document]
HoussemNasri: Also, updated the ID sequence to start from 0.

java/code/src/com/suse/oval/OVALCachingFactory.java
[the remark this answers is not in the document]
HoussemNasri: Done. I removed it altogether.

java/code/src/com/suse/oval/vulnerablepkgextractor/VulnerablePackagesExtractors.java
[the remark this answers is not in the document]
HoussemNasri: It's called LEAP now.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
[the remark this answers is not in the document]
HoussemNasri: Done.

mcalmer:
@HoussemNasri you schema migrations are currenty placed in `susemanager-schema-4.4.6-to-susemanager-schema-4.4.7` directory. We just switched version to 5.0 and before we can merge this, the migrations need to be moved into the current schema migration directory.
`susemanager-schema-5.0.3-to-susemanager-schema-5.0.4` is currently used. If it does not exist yet, just create it

mcalmer:
Another point, the package spacewalk-java failed to build in buildservice.

```
[   53s] compile-all:
[   53s]     [mkdir] Created dir: /home/abuild/rpmbuild/BUILD/spacewalk-java-git-0.7501946/build/classes
[   53s]     [javac] /home/abuild/rpmbuild/BUILD/spacewalk-java-git-0.7501946/build.xml:415: warning: 'includeantruntime' was not set, defaulting to build.sysclasspath=last; set to false for repeatable builds
[   53s]     [javac] Compiling 2969 source files to /home/abuild/rpmbuild/BUILD/spacewalk-java-git-0.7501946/build/classes
[   60s]     [javac] /home/abuild/rpmbuild/BUILD/spacewalk-java-git-0.7501946/code/src/com/suse/oval/ovaltypes/OvalRootType.java:25: error: package javax.xml.bind.annotation is not visible
[   60s]     [javac] import javax.xml.bind.annotation.XmlAccessType;
```
I would think of a missing dependency in the spec file.

srbarrios:
It would be good to develop some E2E test scenarios for this feature as part of this PR.

srbarrios:
Also, looking at the GH acceptance-tests-secondary-parallel checks, it seems there is a problem in one scenario, but probably totally unrelated with this PR:

<image src="https://github.com/uyuni-project/uyuni/assets/2827771/634c0597-d8ed-4975-8aaa-70044f92127f" width = 600>

susemanager-utils/susemanager-sls/src/grains/cpe_grain.py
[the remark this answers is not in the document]
HoussemNasri: https://github.com/saltstack/salt/pull/65905

HoussemNasri:
> Another point, the package spacewalk-java failed to build in buildservice.

The classes are compiling locally, but using `manager-build.xml` not `build.xml`. I googled the error message and found this [blogpost](https://www.logicbig.com/tutorials/core-java-tutorial/modules/root-modules.html) which suggests that it could be related to a misconfiguration in Java modules. But as far as I understand Uyuni is not modularized, in a sense it doesn't define a `module-info.java` file, so all classes whether in modules or not should be accessible to it as long as the right Jar files are in the classpath.

rjmateus (approved):
Looks good to me. Realy great work on this topic. Congratulations @HoussemNasri

cbosdo:
> > Another point, the package spacewalk-java failed to build in buildservice.
> 
> The classes are compiling locally, but using `manager-build.xml` not `build.xml`. I googled the error message and found this [blogpost](https://www.logicbig.com/tutorials/core-java-tutorial/modules/root-modules.html) which suggests that it could be related to a misconfiguration in Java modules. But as far as I understand Uyuni is not modularized, in a sense it doesn't define a `module-info.java` file, so all classes whether in modules or not should be accessible to it as long as the right Jar files are in the classpath.

I have just pushed a commit in your PR to fix this: we were lacking JAXB jars in the `build.jar` used to build the RPM

mcalmer:
@cbosdo thanks. That helped. The package seems to build now. Let's see what our integration testsuite say.

susemanager-utils/susemanager-sls/src/grains/cpe_grain.py
```diff
@@ -0,0 +1,78 @@
...
+    Currently the function exclusively derives CPEs for Debian and Ubuntu. 
+    These two operating systems are the primary focus, as they are intended to be supported by 
+    the OVAL-based CVE auditing project, for which this modification is intended.
+
+    TODO: reference the OVAL code
+    """
+    os = grains.get('os')
+    os_release = grains.get('osrelease')
+    if os == 'Debian':
+        return "cpe:/o:debian:debian_linux:" + os_release
```
agraul, line 74: This could raise an exception if, for whatever reason, `osrelease` is not in `grains`. `grains.get()` will then return`None`, which can't be added to a string. The easiest fix would be to use `os_release = grains.get('osrelease', '')`.

agraul (changes requested):
I expect a more throughout review of the Salt code in the upstream PR. I left a comment about a potential exception. It wouldn't hurt to fix it here since we'll keep the code in Uyuni for the time being.

ktsamis (approved):
Looks good for the testsuite change

PR #7503, opened 2023-09-06
Simplify and modernize password generation logic

Etheryte:
## What does this PR change?

Drops legacy fallbacks for password generation and simplifies the code we use.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/21410

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/utils/functions.ts
+++ b/web/html/src/utils/functions.ts
@@ -119,14 +119,14 @@ function capitalize(str: string): string {
 
 function generatePassword(): string {
   const length = Math.floor(Math.random() * 10) + 15;
+  // See https://stackoverflow.com/a/68617567/1470607
   const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789:-_";
+  const array = new Uint32Array(charset.length);
+  window.crypto.getRandomValues(array);
+
   let password = "";
-  if (window.crypto && window.crypto.getRandomValues) {
-    var rand = new Uint16Array(length);
-    window.crypto.getRandomValues(rand);
-    for (let i = 0; i < length; i++) password += charset.charAt(Math.floor((rand[i] * charset.length) / 65536));
-  } else {
-    for (let i = 0; i < length; i++) password += charset.charAt(Math.floor(Math.random() * charset.length));
+  for (let i = 0; i < length; i++) {
+    password += charset[array[i] % charset.length];
   }
   return password;
 }
--- a/web/spacewalk-web.changes.eth.update-generator
+++ b/web/spacewalk-web.changes.eth.update-generator
@@ -1,0 +1,1 @@
+- Simplify and modernize password generation logic


web/html/src/utils/functions.ts
```diff
@@ -119,14 +119,14 @@ function capitalize(str: string): string {
 
 function generatePassword(): string {
   const length = Math.floor(Math.random() * 10) + 15;
+  // See https://stackoverflow.com/a/68617567/1470607
   const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789:-_";
+  const array = new Uint32Array(charset.length);
+  window.crypto.getRandomValues(array);
+
   let password = "";
-  if (window.crypto && window.crypto.getRandomValues) {
```
parlt91, line 124: Can we ensure that `window.crypto` is alsways defined or should we keep the check?
I'd guess the answer is yes but it isn't immediately obvious to me.

web/html/src/utils/functions.ts
```diff
@@ -119,14 +119,14 @@ function capitalize(str: string): string {
...
-    var rand = new Uint16Array(length);
-    window.crypto.getRandomValues(rand);
-    for (let i = 0; i < length; i++) password += charset.charAt(Math.floor((rand[i] * charset.length) / 65536));
-  } else {
-    for (let i = 0; i < length; i++) password += charset.charAt(Math.floor(Math.random() * charset.length));
+  for (let i = 0; i < length; i++) {
+    password += charset[array[i] % charset.length];
```
parlt91, line 129: Is `array[i] % charset.length` always an integer without using `Math.floor`?
Or is there some typescript magic that does that automatically?

parlt91 (commented):
Two questions before I give my approval

web/html/src/utils/functions.ts
Etheryte: It's available on all the browsers we support, I think this check was originally added quite a while ago.

web/html/src/utils/functions.ts
Etheryte: We used to index by `Math.random()` in `Math.floor(Math.random() * charset.length)` which would need the floor, now we index by `i` which is from the loop body and will be `0`, `1`, `2` etc. The difference is that now `array` is already full of random values (due to `window.crypto.getRandomValues(array)`) so we can just index into it one element after the other and get random values out.

parlt91: I was talking about the index in charset[]. As it looked to me it's not always going to be an integer if we divide by `charset.length`

parlt91 (approved):
LGTM

Etheryte:
Note to self: this needs to be ported to 4.3 once the branch is unlocked.

PR #7507, opened 2023-09-07
Fix recurring state execution not using correct order

parlt91:
## What does this PR change?

Using a Set instead of a list caused the order of the states to be incorrect.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: Bugfix

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/issues/22465

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/RecurringActionJob.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/RecurringActionJob.java
@@ -32,6 +32,7 @@ import com.suse.manager.maintenance.MaintenanceManager;
 
 import org.quartz.JobExecutionContext;
 
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


mcalmer (changes requested):
Please add a changelog with the bug number.
The code looks good. Not sure if you were able to test it?

parlt91:
> Please add a changelog with the bug number. The code looks good. Not sure if you were able to test it?

Right. I thought there'd be no changelog needed, but I forgot that we have an open bug for it. I'll add one.
I also did some testing and the order seems to be fixed now (only did a couple runs, but the probability that it would randomly always have the defined order is pretty small, so I'm pretty confident it fixes the issue)

mcalmer (approved):

mcalmer:
Thanks. Don't forget to port it to 4.3 :-)

PR #563, opened 2019-02-10
No symlinks

mcalmer:
## What does this PR change?

Replace symlinks with real files to fix permanent package submission.

When creating packages the symlinks are dead and this cause differences in the created tar ball or cpio archive. When checking the next time the package is listed as changed and re-submitted.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: **internal**

- [x] **DONE**

## Test coverage
- No tests: **fixes package submissions**

- [x] **DONE**

## Links

Tracks https://github.com/SUSE/spacewalk/pull/7007

- [x] **DONE**

## Changelogs

Copy the following sentence as a new comment if you don't need changelog entries:

> gitarro no changelog needed !!!

If the test `changelog_test`already run, then add another new comment with the following text:

> gitarro rerun changelog_test !!!

PR #7509, opened 2023-09-07
[GSOC23] - B - Enable the downloading and synchronization of OVAL data

HoussemNasri:
## What does this PR change?

This is the implementation of the OVAL downloader component as stated in the linked RFC. Also adds support for SLE & Leap Micro OVAL-based CVE auditing.
## TODO
- [x] Replace `log.error()` with `log.debug()`
- [x] Fix checkstyle errors

### Useful Links
- [The RFC](https://github.com/uyuni-project/uyuni-rfc/pull/80)
## GUI diff

No difference.

Before:

After:
Screenshot from 2024-05-11 16-51-42

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -478,6 +478,24 @@ class CVEAudit extends React.Component<Props, State> {
...
+              columnKey="scan-data"
+              width="15%"
+              comparator={Utils.sortByText}
+              header={t("Scan Data")}
+              cell={(row, criteria) => {
+                return (
+                  <div
+                    className="scan-data-indicator"
+                    style={{
+                      backgroundColor: row.scannedWithOVAL ? "#009688" : "orange",
```
Etheryte, line 491: We have two different themes with their own branding colors so we can't use hardcoded color values like this. The easiest solution here would be to use classes such as `bg-success`, `bg-warning`, or `bg-danger`, would those suffice to do what you want to do here?

web/html/src/branding/css/susemanager/components/cveaudit.less
```diff
@@ -0,0 +1,10 @@
+.scan-data-indicator {
```
Etheryte, line 1: This isn't really a global component in the same sense that e.g. buttons or inputs are, it would be preferable to make this an isolated module instead. See for example `web/html/src/manager/content-management/shared/components/panels/filters-project/filters-project.module.css` and how it's used in code. Note that you will need to use your style in the component code differently once you make this change.

java/code/src/com/redhat/rhn/common/db/datasource/xml/oval_queries.xml
```diff
@@ -29,4 +29,10 @@
               AND cve.name = :cve_name;
         </query>
     </mode>
+
+    <mode name="check_oval_availability">
+        <query params="cpe">
+            SELECT 1 FROM suseOVALPlatform plat WHERE starts_with(:cpe, plat.cpe);
```
rjmateus, line 35: Can you point me to what are the expected values on to be passed and the ones existing on the table. The order of arguments in this function looks to be on inverted order.

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -2429,6 +2450,10 @@ boolean isDebian10() {
         return ServerConstants.DEBIAN.equals(getOs()) && getRelease().equals("10");
     }
 
+    boolean isRHEL() {
+        return ServerConstants.RHEL.equals(getOs());
```
rjmateus, line 2454: can't we re-use the condition `ServerConstants.REDHAT.equals(getOsFamily())` instead of creating a new variable to defined RedHat?

java/code/src/com/redhat/rhn/domain/server/ServerConstants.java
```diff
@@ -37,6 +37,8 @@ public class ServerConstants {
     public static final String ALMA = "AlmaLinux";
     public static final String AMAZON = "Amazon Linux";
     public static final String ROCKY = "Rocky";
+    public static final String SLED = "SLED";
+    public static final String RHEL = "Red Hat Enterprise Linux";
```
rjmateus, line 41: Can't we re-use the existing variable for REDHAT?

rjmateus (commented):
First round.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
```diff
@@ -0,0 +1,178 @@
...
+        return result;
+    }
+
+    private File downloadOVALFile(String vulnerabilityInfoSource) throws IOException {
+        URL vulnerabilityInfoURL = new URL(vulnerabilityInfoSource);
+        String vulnerabilityInfoOVALFilename = FilenameUtils.getName(vulnerabilityInfoURL.getPath());
+        File vulnerabilityFile =
+                new File(DOWNLOAD_PATH + vulnerabilityInfoOVALFilename);
+        // Start downloading
+        FileUtils.copyURLToFile(vulnerabilityInfoURL, vulnerabilityFile, 10_000, 10_000);
```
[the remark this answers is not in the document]
rjmateus: This needs to come from a configuration property.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        return result;
+    }
+
+    private File downloadOVALFile(String vulnerabilityInfoSource) throws IOException {
+        URL vulnerabilityInfoURL = new URL(vulnerabilityInfoSource);
+        String vulnerabilityInfoOVALFilename = FilenameUtils.getName(vulnerabilityInfoURL.getPath());
+        File vulnerabilityFile =
+                new File(DOWNLOAD_PATH + vulnerabilityInfoOVALFilename);
```
rjmateus, line 110: This will copy the file to a location under `/var/log/rhn/ovals/`. This is not a log file. If this is a permanent file we need to store we should define a better location. IF the file is temporary and will be deleted after processing, uyuni already have a cache/temporary location defined which can be used (I think is `/var/cache/rhn/`.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
```diff
@@ -0,0 +1,178 @@
...
+        // Start downloading
+        FileUtils.copyURLToFile(vulnerabilityInfoURL, vulnerabilityFile, 10_000, 10_000);
+
+        return decompressIfNeeded(vulnerabilityFile);
+    }
+
+    private File decompressIfNeeded(File file) {
+        OVALCompressionMethod compressionMethod = getCompressionMethod(file.getName());
```
rjmateus, line 118: I'm not sure if we need all these getCompression methods. It only checking the file extension and then uses it on this method only to perform and if else chain. Why not do we use it directly in a switch?

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
```diff
@@ -0,0 +1,178 @@
...
+        File uncompressedOVALFile;
+        if (compressionMethod == OVALCompressionMethod.BZIP2) {
+            uncompressedOVALFile = new File(file.getPath().replace(compressionMethod.extension(), ""));
```
rjmateus, line 121: this replace is strange. Why not just remove the extension? Normally the files have the format `...oval.xml.bz2`. If we remove the `.bz2` we get the file named. With the replace, we could be changing the file name since the string can appear on any place.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
```diff
@@ -0,0 +1,178 @@
...
+            decompressBzip2(file, uncompressedOVALFile);
+        }
+        else if (compressionMethod == OVALCompressionMethod.GZIP) {
+            uncompressedOVALFile = new File(file.getPath().replace(compressionMethod.extension(), ""));
```
rjmateus, line 125: same as above

java/code/src/com/suse/oval/config/test/oval.config.json
```diff
@@ -0,0 +1,144 @@
+{
+  "sources": {
+    "openSUSE_LEAP": {
```
rjmateus, line 3: This file I think is used just for testing. Looking at the one that will be shipped to customers the name of this OS is different. In here called `openSUSE_LEAP` and on the other file is `LEAP`. What should be the name?

rjmateus (commented):
Another round of review. Looks good to me but we can improve the download part, specially with the topics Welder mentioned.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -298,4 +326,99 @@ public static List<CVEAuditImage> listImagesByPatchStatus(User user,
...
+        }
+    }
+
+    /**
+     * Identifies the OS products to synchronize OVAL data for.
+     * */
+    private static List<OVALProduct> getProductsToSync() {
+        List<Server> allServers = ServerFactory.list(false, false);
+
+        return allServers.stream().map(Server::toOVALProduct)
```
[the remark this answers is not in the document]
HoussemNasri: This could have resulted in OVAL data being re-synchronized unnecessarily for the same product, nice catch!

susemanager-sync-data/oval.config.json
```diff
@@ -0,0 +1,144 @@
+{
+  "sources": {
```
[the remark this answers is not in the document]
HoussemNasri: At the moment, patch information is sourced from channels. Eventually, we'll start using OVAL patch info to gather additional metadata not available through channels, such as patch status (e.g., in QA or released). The patch info source was added to the JSON file, to prevent compatibility problems in the future.

java/code/src/com/redhat/rhn/manager/audit/CVEAuditManagerOVAL.java
```diff
@@ -298,4 +326,99 @@ public static List<CVEAuditImage> listImagesByPatchStatus(User user,
...
+    /**
+     * Launches the OVAL synchronization process
+     * */
+    public static void syncOVAL() {
+        List<OVALProduct> productsToSync = getProductsToSync();
+
+        LOG.debug("Detected {} products eligible for OVAL synchronization: {}", productsToSync.size(), productsToSync);
+
+        OVALDownloader ovalDownloader = new OVALDownloader(OVALConfigLoader.loadDefaultConfig());
+        for (OVALProduct product : productsToSync) {
```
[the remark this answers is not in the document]
HoussemNasri: `productsToSync` is now a set.

java/code/src/com/suse/oval/config/test/oval.config.json
HoussemNasri: Nice catch! It was a typo. Right now, I'm thinking of using a symbolic link to the OVAL config file in `susemanager-sync-data`? @rjmateus what do you think?

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
HoussemNasri: Good point, I'm removing it and deleting the OVALCompressionMethod enum altogether.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
HoussemNasri: The goal was to remove the extension but as you said this would remove the extension string if it occurred inside the filename. It's fixed now.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
HoussemNasri: Fixed.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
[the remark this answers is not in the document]
HoussemNasri: > Is 10 seconds enough?

It should be. I never hit a timeout while set to 10 seconds and I don't have the fastest network (usually below 1 MB). I'll set it to 15 seconds by default and make it configurable as Ricardo suggested.

java/code/src/com/redhat/rhn/common/db/datasource/xml/oval_queries.xml
HoussemNasri: The param cpe is the CPE of the system to scan. It is extracted as a salt grain and stored in the rhnServer table. You can find examples of it [here](https://github.com/chef/os_release). The plat.cpe is the CPE that we extract from OVAL files. 

We want to check if the CPE of the system corresponds to any CPEs we extracted from OVAL files. We could do that by searching for exact equality but that wouldn't match if the system has a CPE like `cpe:/o:redhat:enterprise_linux:8.0:GA` considering that this is the standard list of vulnerable RedHat 8 CPEs found in OVAL files:

```xml
   <affected_cpe_list>
    <cpe>cpe:/a:redhat:enterprise_linux:8</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::appstream</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::crb</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::highavailability</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::nfv</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::realtime</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::resilientstorage</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::sap</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::sap_hana</cpe>
    <cpe>cpe:/a:redhat:enterprise_linux:8::supplementary</cpe>
    <cpe>cpe:/o:redhat:enterprise_linux:8</cpe>
    <cpe>cpe:/o:redhat:enterprise_linux:8::baseos</cpe>
   </affected_cpe_list>
```

So we check for prefixes.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
HoussemNasri: A cache location would be ideal. However, I would leave it to the users to decide whether to delete the file after processing or not (through configuration). It can also be useful for debugging.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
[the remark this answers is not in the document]
HoussemNasri: To move faster, I'm increasing the read and connection timeouts to 15 seconds for now and will create an issue to make it configurable. The issue is tracked [here](https://github.com/orgs/uyuni-project/projects/3/views/1?pane=issue&itemId=59489477) and will be converted to a real GitHub issue once this PR is merged.

java/code/src/com/suse/oval/ovaldownloader/OVALDownloader.java
HoussemNasri: Done. OVAL data is now cached under `/var/cache/rhn/ovals`.

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -2468,6 +2493,49 @@ boolean isRocky9() {
...
+    /**
+     * Derives an {@link com.redhat.rhn.manager.audit.CVEAuditManagerOVAL.OVALProduct} object based on the server's
+     * information, including the operating system name and release version. Used by the OVAL synchronization process
+     * to identify the installed product and synchronize OVAL data for it.
+     *
+     * @return the information of the installed product to synchronize OVAL data for if the product is eligible for
+     * OVAL synchronization and {@code Optional.empty} otherwise
+     * (for example product maintainers don't produce OVAL data)
+     * */
+    public Optional<CVEAuditManagerOVAL.OVALProduct> toOVALProduct() {
```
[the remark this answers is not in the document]
HoussemNasri: Done.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -219,6 +219,21 @@ class CVEAudit extends React.Component<Props, State> {
     });
   };
 
+  getPatchStatusAccuracyWarning = (row) => {
+    const dataSources: string[] = row.scanDataSources;
+    if (!dataSources) {
+      return "Unknown patch status";
```
Etheryte, line 225: All of these strings need a translation wrapper `t()`.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -219,6 +219,21 @@ class CVEAudit extends React.Component<Props, State> {
...
+    }
+
+    if (dataSources.indexOf("OVAL") === -1) {
+      return "OVAL data not in sync. Possible false negatives";
+    } else if (dataSources.indexOf("CHANNELS") === -1) {
+      return "Channels not in sync for the given server product. Patches are not available.";
+    }
+
+    return "If you see this report a bug.";
```
Etheryte, line 234: This isn't a very good way to indicate an error, I think it would make more sense to show an error toast, log an error with more details about what went wrong, and then render the field empty.

Etheryte: If reasonably possible, it might be good to aggregate showing the toast though, so if the underlying data is broken and the user has 25 rows showing, they don't get 25 error toasts.

HoussemNasri: Both solutions sound good, but I'm leaning toward the logging solution. The error in question is unlikely to occur; if it does, it's likely due to an incompatibility between the frontend and backend server, so it's better to aim for the simpler solution.

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/003-oval-taskomatic.sql
```diff
@@ -0,0 +1,52 @@
...
+
+-- Insert task and map to the Java class
+INSERT INTO rhnTaskoTask (id, name, class)
+SELECT sequence_nextval('rhn_tasko_task_id_seq'), 'oval-data-sync', 'com.redhat.rhn.taskomatic.task.OVALDataSync'
+WHERE NOT EXISTS (SELECT 1
+                  FROM rhnTaskoTask
+                  WHERE name = 'oval-data-sync');
+
+-- Insert bunch
+INSERT INTO rhnTaskoBunch (id, name, description, org_bunch)
```
rjmateus, line 21: I think you are missing some sql files where this insert needs to be added. This SQL file will only run in the upgrade, but for new installations it's not called. Look at what you have add to rhnTaskoTask.sql file.

rjmateus (commented):
I made a high level overview and overall looks good. I only have one topic regarding the SQL scripts.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
Etheryte: Sounds good to me. 👍

schema/spacewalk/upgrade/susemanager-schema-4.4.6-to-susemanager-schema-4.4.7/003-oval-taskomatic.sql
HoussemNasri: Thanks, it should be fixed now.

rjmateus (approved):
looks good to me. Thank you for the great contribute on this topic. you rock :)

java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
```diff
@@ -141,15 +141,13 @@ public static Object cveAudit(Request req, Response res, User user) {
...
                     .listSystemsByPatchStatus(user, cveAuditRequest.cveIdentifier,
                             cveAuditRequest.statuses);
                     cveAuditServers.forEach(serv -> serv.setSelected(systemSet.contains(serv.getId())));
                     return json(res, ResultJson.success(cveAuditServers));
                 case IMAGE:
-                    // TODO: Use CVEAuditManagerOVAL once it's ready
-                    List<CVEAuditImage> cveAuditImages = CVEAuditManager
+                    List<CVEAuditImage> cveAuditImages = CVEAuditManagerOVAL
                     .listImagesByPatchStatus(user, cveAuditRequest.cveIdentifier,
                             cveAuditRequest.statuses);
```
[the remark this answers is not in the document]
HoussemNasri: The reason is two folds: 
1. Currently it's not clear how images would be audited using OVAL data, primarily because it's not possible to extract the CPE (or the operating system) of the image without deploying it. @cbosdo suggested that we can derive the CPE from the channels assigned to the image, but I haven't had time to experiment with the feasibility of that option and what edge cases we might encounter, so the best decision I saw, was to keep using channels for auditing images temporarily.
2. The reason for having `listImagesByPatchStatus` even though it is just delegating to `CVEAuditManager` is just a matter of taste; having a single source of truth where all the CVE auditing logic would reside seemed reasonable from an architectural pov.

java/code/src/com/suse/oval/OsFamily.java
```diff
@@ -15,25 +15,86 @@
...
-    LEAP("openSUSE Leap", "leap", "opensuse"),
-    SUSE_LINUX_ENTERPRISE_SERVER("SUSE Linux Enterprise Server", "sles", "suse"),
-    SUSE_LINUX_ENTERPRISE_DESKTOP("SUSE Linux Enterprise Desktop", "sled", "suse"),
-    REDHAT_ENTERPRISE_LINUX("Red Hat Enterprise Linux", "enterprise_linux", "redhat"),
-    UBUNTU("Ubuntu", "ubuntu", "canonical"),
-    DEBIAN("Debian", "debian", "debian");
+    LEAP("openSUSE Leap", "Leap", "opensuse",
+            oneOf("15.2", "15.3", "15.4", "15.5")),
+    LEAP_MICRO("openSUSELeap Micro", "openSUE Leap Micro", "opensuse",
+            oneOf("5.2", "5.3")),
```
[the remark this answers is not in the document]
HoussemNasri: Good catch, I can't quite remember why I didn't include it, probably just a typo, thanks!

java/code/src/com/suse/oval/config/test/oval.config.json
HoussemNasri: It looks like symlinks could cause issues during packaging: https://github.com/uyuni-project/uyuni/pull/563. I'll add a script to update the OVAL config file, similar to the approach taken in that PR.

java/code/src/com/suse/oval/OsFamily.java
```diff
@@ -15,25 +15,86 @@
...
+    LEAP("openSUSE Leap", "Leap", "opensuse",
+            oneOf("15.2", "15.3", "15.4", "15.5")),
+    LEAP_MICRO("openSUSELeap Micro", "openSUE Leap Micro", "opensuse",
+            oneOf("5.2", "5.3", "5.4", "5.5")),
+    SUSE_LINUX_ENTERPRISE_SERVER("SUSE Linux Enterprise Server", "SLES", "suse",
+            oneOf("11", "12", "15")),
+    SUSE_LINUX_ENTERPRISE_DESKTOP("SUSE Linux Enterprise Desktop", "SLED", "suse",
+            oneOf("10", "11", "12", "15")),
+    SUSE_LINUX_ENTERPRISE_MICRO("SUSE Linux Enterprise Micro", "SLE Micro", "suse",
+            oneOf("5.0", "5.1", "5.2", "5.3", "5.4", "5.5")),
```
admd, line 38: SL Micro 6.0 is will be released in coming weeks, would be good to already add entry for it.

java/code/src/com/suse/oval/OsFamily.java
```diff
@@ -15,25 +15,86 @@
...
-    LEAP("openSUSE Leap", "leap", "opensuse"),
-    SUSE_LINUX_ENTERPRISE_SERVER("SUSE Linux Enterprise Server", "sles", "suse"),
-    SUSE_LINUX_ENTERPRISE_DESKTOP("SUSE Linux Enterprise Desktop", "sled", "suse"),
-    REDHAT_ENTERPRISE_LINUX("Red Hat Enterprise Linux", "enterprise_linux", "redhat"),
-    UBUNTU("Ubuntu", "ubuntu", "canonical"),
-    DEBIAN("Debian", "debian", "debian");
+    LEAP("openSUSE Leap", "Leap", "opensuse",
+            oneOf("15.2", "15.3", "15.4", "15.5")),
+    LEAP_MICRO("openSUSELeap Micro", "openSUE Leap Micro", "opensuse",
+            oneOf("5.2", "5.3", "5.4", "5.5")),
```
admd, line 32: 6.0 will be releasing soo, we should add it here as well.

java/code/src/com/suse/oval/OsFamily.java
```diff
@@ -15,25 +15,86 @@
...
+ * */
 public enum OsFamily {
-    LEAP("openSUSE Leap", "leap", "opensuse"),
-    SUSE_LINUX_ENTERPRISE_SERVER("SUSE Linux Enterprise Server", "sles", "suse"),
-    SUSE_LINUX_ENTERPRISE_DESKTOP("SUSE Linux Enterprise Desktop", "sled", "suse"),
-    REDHAT_ENTERPRISE_LINUX("Red Hat Enterprise Linux", "enterprise_linux", "redhat"),
-    UBUNTU("Ubuntu", "ubuntu", "canonical"),
-    DEBIAN("Debian", "debian", "debian");
+    LEAP("openSUSE Leap", "Leap", "opensuse",
+            oneOf("15.2", "15.3", "15.4", "15.5")),
```
admd, line 30: Same as other 2. 15.6 will be released soon and we can already entry for it.

java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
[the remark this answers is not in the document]
admd: @HoussemNasri can you please create an issue that Oval based scanning will be limited to systems only and not to images and this will be TODO that one might pickup in the future. Maybe also add the comment in this method about this , that @wweellddeerr has pointed out.

java/code/src/com/suse/oval/config/OVALConfigLoader.java
```diff
@@ -0,0 +1,70 @@
...
+
+import static com.suse.utils.Json.GSON;
+
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Objects;
+
+public class OVALConfigLoader {
+    private static final String DEFAULT_CONFIG_PATH = "/usr/share/susemanager/scc/oval.config.json";
```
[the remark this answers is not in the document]
admd: Good point! Ideally, we should not put it here but I see why Houssem chose this place. All of our channel info resides here in this directory and it kinda made sense to keep things together.

@HoussemNasri would it be easy to move this under new directory e.g `/usr/share/susemanager/oval`

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -219,6 +219,21 @@ class CVEAudit extends React.Component<Props, State> {
...
   };
 
+  getPatchStatusAccuracyWarning = (row) => {
+    const dataSources: string[] = row.scanDataSources;
+    if (!dataSources) {
+      return "Unknown patch status";
+    }
+
+    if (dataSources.indexOf("OVAL") === -1) {
+      return "OVAL data not in sync. Possible false negatives";
```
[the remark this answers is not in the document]
HoussemNasri: I ended up with a completely different message that includes more details, please have a look, and see if it needs to be rephrased, thanks!

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -219,6 +219,21 @@ class CVEAudit extends React.Component<Props, State> {
...
+    } else if (dataSources.indexOf("CHANNELS") === -1) {
+      return "Channels not in sync for the given server product. Patches are not available.";
```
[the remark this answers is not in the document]
HoussemNasri: This one is also changed.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -219,6 +228,30 @@ class CVEAudit extends React.Component<Props, State> {
...
+    if (!dataSources) {
+      console.error("CVE audit data sources were not supplied by server.");
+      return t("Error, see console");
+    }
+
+    if (dataSources.length === 0) {
+      return t("Unknown patch status");
+    } else if (dataSources.indexOf("OVAL") === -1) {
+      return t(
+        "OVAL data was found not to be in sync, thus the server was scanned with errata metadata fetched from channels. This could mean that the operating system installed on the server is not supported by OVAL based CVE auditing or that OVAL data is not in sync, if it is the latter option please sync OVAL data as it is known to yield more accurate results."
```
admd, line 242:
```suggestion
        "OVAL data is out of sync, so the server was scanned using errata metadata from channels. This could mean that the server's OS isn't supported by OVAL-based CVE auditing, or that OVAL data needs to be synced."
```
NitPick:This is a good  message to be added in the documentation  but for UI probably a bit too long. A bit concise message might be better.

HoussemNasri: Thanks, I reverted back to the shorter message.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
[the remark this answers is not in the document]
HoussemNasri: Done.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/suse/oval/config/OVALConfigLoader.java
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/suse/manager/webui/controllers/CVEAuditController.java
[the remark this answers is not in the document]
HoussemNasri: I created an item in the project board: https://github.com/orgs/uyuni-project/projects/3/views/1?pane=issue&itemId=63725688. Will transfer into a real issue once PR is merged.

java/code/src/com/suse/oval/OsFamily.java
HoussemNasri: Done. I have added an entry to the config file too, it points to an empty URL for now.

web/html/src/manager/audit/cveaudit/cveaudit.tsx
```diff
@@ -219,6 +228,30 @@ class CVEAudit extends React.Component<Props, State> {
     });
   };
 
+  getPatchStatusAccuracyWarning = (row) => {
+    const dataSources: string[] = row.scanDataSources;
+    if (!dataSources) {
+      Loggerhead.error("CVE audit data sources were not supplied by server.");
+      return t("Error, see console");
```
Etheryte, line 235: I think here and below, the errors we log are also fine as tooltips, so `return t("CVE audit data sources were not supplied by server")` and `return t("Invalid scan data sources")` work imo. But this is a minor nitpick.

Etheryte (approved):

rjmateus (dismissed):
From my side looks good, as soon the existing comments are addressed. Thank you

java/code/src/com/suse/oval/OVALCachingFactory.java
```diff
@@ -57,33 +65,40 @@ public static void savePlatformsVulnerablePackages(OvalRootType rootType) {
 
         OVALLookupHelper ovalLookupHelper = new OVALLookupHelper(rootType);
 
-        DataResult<Map<String, Object>> batch = new DataResult<>(new ArrayList<>(1000));
-
+        List<ProductVulnerablePackages> productVulnerablePackages = new ArrayList<>();
         for (DefinitionType definition : rootType.getDefinitions()) {
```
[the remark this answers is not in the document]
HoussemNasri: The problem originates from the usage of `&gt;=`. The older OVAL data used to indicate the exact release versions of the product being vulnerable using `==`, I implemented a workaround by allowing the usage of `>=` and treating it as an `==`, this is inaccurate and needs to be treated later. I have only found a few usages of this pattern, so I think it's okay to delay handling it in order to move faster.

image

susemanager-sync-data/susemanager-sync-data.spec
```diff
@@ -49,5 +50,6 @@ install -m 0644 additional_repositories.json    %{buildroot}/usr/share/susemanag
 /usr/share/susemanager/scc/channel_families.json
 /usr/share/susemanager/scc/additional_products.json
 /usr/share/susemanager/scc/additional_repositories.json
+/usr/share/susemanager/scc/oval.config.json
```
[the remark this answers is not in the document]
HoussemNasri: Done.

java/code/src/com/suse/oval/OsFamily.java
```diff
@@ -15,25 +15,86 @@
...
-    SUSE_LINUX_ENTERPRISE_DESKTOP("SUSE Linux Enterprise Desktop", "sled", "suse"),
-    REDHAT_ENTERPRISE_LINUX("Red Hat Enterprise Linux", "enterprise_linux", "redhat"),
-    UBUNTU("Ubuntu", "ubuntu", "canonical"),
-    DEBIAN("Debian", "debian", "debian");
+    LEAP("openSUSE Leap", "Leap", "opensuse",
+            oneOf("15.2", "15.3", "15.4", "15.5", "15.6")),
+    LEAP_MICRO("openSUSELeap Micro", "openSUE Leap Micro", "opensuse",
+            oneOf("5.2", "5.3", "5.4", "5.5", "6.0")),
+    SUSE_LINUX_ENTERPRISE_SERVER("SUSE Linux Enterprise Server", "SLES", "suse",
+            oneOf("11", "12", "15")),
```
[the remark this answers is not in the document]
HoussemNasri: I went with the stripping solution. It is the same approach used for matching OVAL metadata and RHEL servers.

java/code/src/com/suse/oval/OVALCachingFactory.java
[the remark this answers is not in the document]
HoussemNasri: If the OVAL metadata indicates that versions of module SLE HPC 12 or above are vulnerable to a specific vulnerability, the current workaround only identifies SLE HPC 12 as vulnerable. Consequently, if users scan for SLE HPC 15, it will incorrectly reports it as not vulnerable.

From my analysis of several OVAL files, this pattern is typically applied to modules rather than specific distributions. In these OVAL files, the affected module is listed alongside the distributions that have the module pre-installed. Therefore, the only scenario where a vulnerability might be missed is if the module was installed manually.

image

I have considered multiple solutions, but none are quick fixes that can be integrated into this PR. While the impact of the workaround is limited, a more thorough analysis is required to assess the impact accurately.

In the meantime, I have added a "To Open" issue card item to the project board.

PR #7537, opened 2023-09-13
User soft deletion

parlt91:
## What does this PR change?

Remove functionality for users to 'hard' delete users from the webUI API and spacecmd.
Deactivate should be used instead to serve as a soft deletion mechanism that can be reverted if needed.
This way references to the deactivated user e.g. Recurring Actions will be kept if the user is disabled (as opposed to deleted).

The `UserFactory.deleteUser()` function was preserved since we use it to clean up API tests
See: `BaseHandlerTestCase.tearDown()`

The stored procedure to delete users from the db will also be kept since we still want to able to delete users when deleting an organization.

## GUI diff

Before: 
user-delete-before

After: 
user-delete-after

- [x] **DONE**

## Documentation
- Documentation issue was created: **TODO**
- **Will add documentation changes once this PR is approved**

- [ ] **DONE**

## Test coverage
- Unit tests were adapted
- Cucumber tests were adapted

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/21035

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


agraul:
Before looking at the code, I have a fundamental question: Is "soft deletion" in line with regulations such as the GDPR? 

So far I'm under the impression that Uyuni needs to have a way to delete user data and not just hide it, but I'm not law expert.

cbbayburt:
Also not an expert but to my knowledge, we're allowed to keep such data as long as it has a use for a functionality that serves customers' needs according to GDPR. I believe this is a valid case.

cbbayburt (commented):
I don't know if this has been discussed before, but from the UX perspective so far I imagined "soft delete" would be completely transparent to the user. That is, we would still have the "Delete" function in the GUI/API, but effectively that would hide the user from all UI instead of removing the data completely. This would be separate disabling.

They might miss their "Delete" button :D

Any thoughts @parlt91 @admd ?

cbbayburt:
I'm not sure if any of this applies to us at all since all the data in question is actually hosted by the user. We're not keeping any data in our domain.

They can exercise their right to delete all data basically with sql DELETE statements :)

cbbayburt:
I think @admd can ping the Legal team to address any concerns, but I don't think we need to proactively spend any extra effort on this.

agraul:
> I'm not sure if any of this applies to us at all since all the data in question is actually hosted by the user. We're not keeping any data in our domain.
> 

Right, we just provide the software that stores the data in question. It might not be our problem legally since we don't offer hosted Uyuni ourselves, but to me that's not good enough. I don't want our users to have to fight our product just to do what they are lawfully required to do.

> They can exercise their right to delete all data basically with sql DELETE statements :)

Aren't we striving to be API-first? Is this a serious proposal to ask Uyuni operators to run delete statements because we want to drop something from the API?

admd:
Personally, I don't think this is a problem honestly as we are not storing any data at our end as Can has pointed out but I will clarify with the legal team anyways.

Removing user completely gets us in different problem. There is requirement to keep track of all the actions that are performed by a user on the server. If we do allow deletion of user, we lose track of all actions that user has performed. Now again I am not sure how should we handle this ideally.  

I will clarify this with the legal team and will get back to you on this one.

srbarrios:
> Personally, I don't think this is a problem honestly as we are not storing any data at our end as Can has pointed out but I will clarify with the legal team anyways.
> 
> Removing user completely gets us in different problem. There is requirement to keep track of all the actions that are performed by a user on the server. If we do allow deletion of user, we lose track of all actions that user has performed. Now again I am not sure how should we handle this ideally.
> 
> I will clarify this with the legal team and will get back to you on this one.

If I may, a good approach would be to decouple sensitive user data from the identifier we need to track the actions of an user.
In other words, have some anonymization procedure that allow us to delete user data that can impact on GDPR rights, but allow us to keep having some identifier for our historical data.

cbbayburt:
Still waiting for a final answer from Legal.

ktsamis (approved):

Issue #7565, opened 2023-09-20
mgr-inter-sync --list-channels problem sync with slave


santix1989:
### Problem description

Hi,
I have configured 2 nodes, one master and one slave, with this step : 

https://www.uyuni-project.org/uyuni-docs/en/uyuni/administration/iss.html

when started the sync, have this issue : 

```
mgr-inter-sync --list-channels
15:58:41 Uyuni - live synchronization
15:58:41    url: https://XXXXXXX
15:58:41    debug/output level: 1
15:58:41 --- Warning: SSL connection made but no CA certificate used
HOST:  443
15:58:41 
15:58:41 
ERROR: there was a problem synchronizing the information.
       Error message: [Errno -2] Name or service not known

15:58:41 *** TRACEBACK: 
15:58:41 Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/satsync.py", line 215, in main
    self.syncer.initialize()
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/satsync.py", line 558, in initialize
    auth.checkAuth()
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/xmlWireSource.py", line 378, in checkAuth
    authYN = self._xmlrpc('authentication.check', (self.systemid,))
  File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/xmlWireSource.py", line 357, in _xmlrpc
    retval = getattr(BaseWireSource.serverObj, function)(*params)
  File "/usr/lib/python3.6/site-packages/rhn/rpclib.py", line 659, in __call__
    return self._send(self._name, args)
  File "/usr/lib/python3.6/site-packages/rhn/rpclib.py", line 391, in _request
    self._handler, request, verbose=self._verbose)
  File "/usr/lib/python3.6/site-packages/rhn/transports.py", line 177, in request
    headers, fd = req.send_http(host, handler)
  File "/usr/lib/python3.6/site-packages/rhn/transports.py", line 728, in send_http
    self._connection.connect()
  File "/usr/lib/python3.6/site-packages/rhn/connections.py", line 191, in connect
    socket.AF_UNSPEC, socket.SOCK_STREAM)
  File "/usr/lib64/python3.6/socket.py", line 745, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

[...]

Issue #5522, opened 2022-05-31
IIS not working,  Error message: While running 'authentication.check': caught <class 'KeyError'> : 'sat'


electronix001:
# Problem description

After some updates the slave server not working properly. In fact the IIS synchronization.

## Version of Uyuni Server and Proxy (if used)

2022.05-180.1.uyuni1  not proxy

## Details about the issue

Logs from the SLAVE server :

2022/05/31 15:22:36 +02:00 Commandline: ['/usr/bin/mgr-inter-sync', '--list-channels']
2022/05/31 15:22:36 +02:00 Uyuni - live synchronization
2022/05/31 15:22:36 +02:00    url: https://ni-dmz-vm-sa-sws.gnctcs.com
2022/05/31 15:22:36 +02:00    debug/output level: 1
2022/05/31 15:22:36 +02:00 4248 0.0.0.0: common/suseLib._get_proxy_from_yast("Couldn't open /root/.curlrc",)
2022/05/31 15:22:37 +02:00
2022/05/31 15:22:37 +02:00
ERROR: There was a problem communicating with the ISS Master.
       Depending on the specific error details, please review your configuration,
       basic network connectivity, and/or name resolution, and try again.
       Error message: While running 'authentication.check': caught
<class 'KeyError'> : 'sat'

2022/05/31 15:29:52 +02:00 Commandline: ['/usr/bin/mgr-inter-sync', '--list-channels', '--debuglevel=5']
2022/05/31 15:29:58 +02:00 Commandline: ['/usr/bin/mgr-inter-sync', '--list-channels', '--debug-level=5']
2022/05/31 15:29:58 +02:00 4608 0.0.0.0: rhnSQL/driver_postgresql._execute_wrapper('Executing SQL: "select distinct c.label from rhnChannel c where c.org_id is null" with bind params: {}',)
2022/05/31 15:29:58 +02:00 Uyuni - live synchronization
2022/05/31 15:29:58 +02:00    url: https://ni-dmz-vm-sa-sws.example.com
2022/05/31 15:29:58 +02:00    debug/output level: 5
2022/05/31 15:29:58 +02:00 4608 0.0.0.0: common/suseLib._get_proxy_from_rhn_conf('Could not read proxy URL from rhn config.',)
2022/05/31 15:29:58 +02:00 4608 0.0.0.0: common/suseLib._get_proxy_from_yast("Couldn't open /root/.curlrc",)
[...]

PR #7588, opened 2023-09-26
mgr-inter-sync command fix config clear cache

tanganellilore:
## What does this PR change?

In `mgr-inter-sync` if we use `--iss-parent` or if we missing to add `iss_parent` on `rhn.conf`, configuration passed or retrived from DB will be used only for some functions, not for whole run of the script.
A issue that highlight this is this one:
https://github.com/uyuni-project/uyuni/issues/7565

This PR will fix this behavior, that is generated by calculation of initialized component and `timeDiff `in same line.

Actually, `self.__config` will be cleared everytime `self.__component` is `None` and `timeDiff` is not `0`, so everytime we have new component, we simply delete all config already loaded.
This probably was introduced with the new method `with cfg_component(component=None)` that for example is used internally. Calling this we simply define new component=None and `parse` function clear everything in the `self.__config`  creating a "sort of loop", because net time that arrive with new component, this will clear everything.

This explain why we have this error on linked issue, because in that case we simply reload the `rhn.conf` for component `server.satellite`, defined by `mgr-inter-sync` command, instead use what was extracted from DB or passed by command line.

This PR simply move out the `self.is_initialized()` in a dedicated `if`, that wrap the check of `timeDiff`.
So:

1. in case of component not initialized (like component=None or any other component), we simply to load the config from disk
2. in case of component already initialized, we check if file was changed
   a. if not, we `return`
   b. if yes, we clear `self.__config `

I have some doubpts related clear everything, becase this can broke something that is setted by OPTIONS or command line, but I don't know the logic behind this code, so I leave it as is (I have in mind some method to "merge" what is cached with what is changed in a file, but I prefer for now maintain as is)

I also evaluated some differents approach but this is the simplest with less code changed

## GUI diff

No difference.

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/spacewalk/common/rhnConfig.py
+++ b/python/spacewalk/common/rhnConfig.py
@@ -129,16 +129,17 @@ class RHNOptions:
         the configuration cache self.__configs
         """
         # Speed up the most common case
-        timeDiff = self.modifiedYN()
-        if not timeDiff and self.is_initialized():
-            # Nothing to do: the config file did not change and we already
-            # have the config cached
-            return
-        else:
-            # if the timestamp changed, clear the list of cached configs
-            # and retain the new timestamp
-            self.updateLastModified(timeDiff)
-            self.__configs.clear()  # cache cleared
+        if self.is_initialized():
+            timeDiff = self.modifiedYN()
+            if not timeDiff:
+                # Nothing to do: the config file did not change and we already
+                # have the config cached
+                return
+            else:
+                # if the timestamp changed, clear the list of cached configs
+                # and retain the new timestamp
+                self.updateLastModified(timeDiff)
+                self.__configs.clear()  # cache cleared
 
         # parse the defaults.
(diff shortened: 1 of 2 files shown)


tanganellilore:
probably also resolve this issue
https://github.com/uyuni-project/uyuni/issues/5522

mbussolotto:
according to https://github.com/uyuni-project/uyuni/issues/7565 , the PR is fine and fix the issue (not sure about https://github.com/uyuni-project/uyuni/issues/5522 )

cbosdo (approved):
Looks good to me, thanks a lot for fixing this issue.

Issue #7356, opened 2023-08-03
Reposync RPM package import fails if RPM SPEC Release field is empty/null


santeri3700:
### Problem description

The reposync package import to DB fails when an RPM package has not defined a value for the "Release" field.
This problem seems to have existed since Spacewalk according to this old mailing list thread: https://listman.redhat.com/archives/spacewalk-list/2014-June/014560.html

I started to experience this bug on two separate Uyuni 2023.04 servers starting from July 22nd 2023 with the following repositories. It seems Microsoft has pushed a new package or a version of it which Uyuni can't parse properly.
- https://packages.microsoft.com/rhel/8/prod/
- https://packages.microsoft.com/rhel/9/prod/

The problematic package seems to be `microsoft-identity-diagnostics`, specifically `microsoft-identity-diagnostics-1.0.0.x86_64` as apparent from the metadata (See "Release" line):
```
$ rpm -qip microsoft-identity-diagnostics-1.0.0.x86_64.rpm 
warning: microsoft-identity-diagnostics-1.0.0.x86_64.rpm: Header V4 RSA/SHA256 Signature, key ID be1229cf: NOKEY
Name        : microsoft-identity-diagnostics
Epoch       : 0
Version     : 1.0.0
Release     : 
Architecture: x86_64
Install Date: (not installed)
Group       : (none)
Size        : 4101685
License     : 
Signature   : RSA/SHA256, Fri Jul 21 19:47:54 2023, Key ID eb3e94adbe1229cf
Source RPM  : microsoft-identity-diagnostics-1.0.0--src.rpm
Build Date  : Fri Jul 21 01:32:18 2023
Build Host  : fv-az125-367
Packager    : root
Vendor      : 
URL         : 
Summary     : microsoft-identity-diagnostics
Description :
```

The proper fix would be for Microsoft to add the Release number to the package's RPM SPEC but I think it would be wise to add checks and handling to Uyuni itself in case other repositories make same kind of mistake while packaging.
More information about the RPM SPEC can be found here: https://rpm-packaging-guide.github.io/#what-is-a-spec-file

A slightly similar problem was also reported on #6002 regarding DEB packages and missing "Installed-Size" value.

### Steps to reproduce

NOTE: Reproducing might not be possible with the repositories later if/when Microsoft fixes the package.
1. Install an Uyuni 2023.04 server (haven't tested with older versions but probably works with previous versions too)
2. Add at least one of the repositories mentioned in the description.
[...]

PR #7594, opened 2023-09-27
Prevent reposync tracebacks on packages with empty release

vzhestkov:
## What does this PR change?

In case if repo contains the package with empty release specified in the header it could lead to the traceback while running `spacewalk-repo-sync` for such repo.

The only example found was `microsoft-identity-diagnostics-1.0.0-.x86_64.rpm` in `https://packages.microsoft.com/rhel/8/prod/` but it was fixed after.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: no tests for the modified packages and the fix is for very specific corner case.

## Links

Fixes https://github.com/SUSE/spacewalk/issues/22135

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
@@ -1180,6 +1180,10 @@ password={passwd}
         for pack in pkglist:
             new_pack = ContentPackage()
             epoch, version, release = RawSolvablePackage._parse_solvable_evr(pack.evr)
+            # Prevent repeating to download the package with empty release on each run.
+            # Release should be specified in the package. Assuming empty release as 0.
+            if release == "":
+                release = "0"
             new_pack.setNVREA(pack.name, version, release, epoch, pack.arch)
             new_pack.unique_id = RawSolvablePackage(pack)
             checksum = pack.lookup_checksum(solv.SOLVABLE_CHECKSUM)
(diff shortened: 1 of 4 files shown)


agraul (commented):
Can these packages be installed? If not, it does not make sense to mirror them.

In general I wonder if skipping these packages (and logging that they are skipped because the release is unset) is a better approach. Are packages with an empty release tag valid?

agraul:
BTW, this was also reported in https://github.com/uyuni-project/uyuni/issues/7356

vzhestkov:
The fix was made exactly because of `microsoft-identity-diagnostics` also reported https://github.com/uyuni-project/uyuni/issues/7356 as @agraul mentioned. Yes, the package can be installed, but I didn't find the notes if this strictly violates any rules, but from my point of view the package should be treated as invalid in such case. And as I know the package was fixed already and not present in the repo anymore.

agraul:
> The fix was made exactly because of `microsoft-identity-diagnostics` also reported #7356 as @agraul mentioned. Yes, the package can be installed, but I didn't find the notes if this strictly violates any rules, but from my point of view the package should be treated as invalid in such case. And as I know the package was fixed already and not present in the repo anymore.

I have the same feeling, all packaging guides I have found say that a release is required - but I don't know if it's strictly required. I'd prefer not to have this "we set the release if unset" hack if we can avoid it.

m-czernek:
ping @vzhestkov - just an FYI, I am closing this PR in favor of https://github.com/uyuni-project/uyuni/pull/7820

PR #7595, opened 2023-09-27
fix low contrast in homepage

### LINKED ISSUE #7101: Fix low contrast issues in the home page
# Short description
The home page of Uyuni has five identified low-contrast issues.

# Details
This issue hopes to improve on accessibility for the home page. The focus here will be to fix the low-contrast errors identified by WAVE. Currently, WAVE identified five accessibility errors on the page. These are:

1.  Low contrast between the menu items and the green background:
err1

2. Low contrast between the action items and the gray background in the `Recently Scheduled Actions` table
err2

3. Low contrast between the texts and background of the `System` and `Base Channel` columns of the `Recently registered Systems` table
err3

Rakielle:
## What does this PR change?

**This PR fixes the low contrast between the menu items and the background color, as identified by WAVE**

## GUI diff

Before:

After:

- [X] **DONE**

## Documentation
- No documentation needed: **This changes improves the contrast color of the texts and the background to enhance accessibility, no additional documentation is needed.**

- [X] **DONE**

## Test coverage

- No tests: already covered

- [X] **DONE**

## Links

Fixes #7101 

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/web/html/src/branding/css/uyuni/legacy-theme.less
+++ b/web/html/src/branding/css/uyuni/legacy-theme.less
@@ -430,7 +430,7 @@ header, nav.navbar-pf {
       // nodeLink and leafLink
       div.nodeLink,
       div.leafLink {
-        color: @gray-light;
+        color: @white;
         border-left: 3px solid transparent;
         a {
           display: block;
--- a/web/html/src/branding/css/uyuni/legacy-variables.less
+++ b/web/html/src/branding/css/uyuni/legacy-variables.less
@@ -1,6 +1,7 @@
 // TODO: This should be merged with variables.less
 
 /* Main tags colours */
+@green-dark-1:           #008558;
 @green-dark:             #02A49C;
 @green:                  #00C081;
 @green-light:            #02D35F;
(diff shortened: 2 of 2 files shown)


Etheryte:
While this fixes the contrast issue, I think it creates another new problem we don't want to have, which is that we now have a new color we only use in one place. Given that it's right next to the old color used in the home icon, I think we might have some better options here.

What about using the same menu highlighting logic as the Suma theme uses? Over time we're aiming to bring the two themes closer together anyway and this way we can solve both the contrast issue as well as keep the color. What do you think @Rakielle @cbbayburt?

For context, this is how the highlighting is handled in the Suma theme:

Screenshot 2023-10-02 at 16 21 31

cbbayburt:
> What about using the same menu highlighting logic as the Suma theme uses? Over time we're aiming to bring the two themes closer together anyway and this way we can solve both the contrast issue as well as keep the color. What do you think @Rakielle @cbbayburt?

I think it's a good idea. @Etheryte could you point out where this SUMA style menu highlighting is implemented?

Etheryte:
If I'm not mistaken that part should come from the old theme base in `web/html/src/branding/css/base/theme.less`, grep for `@aside-menu-active`. You will probably need other changes too aside from setting just that bit though, since I'd wager the underlying styles are different.

Etheryte:
I've manually cherry-picked the changes from this PR into https://github.com/uyuni-project/uyuni/pull/8420 and https://github.com/uyuni-project/uyuni/pull/8169 so this should be safe to close. Since the underlying structure changed a fair bit the change set doesn't apply cleanly directly, but the contrast should now be fine.

PR #7598, opened 2023-09-28
Make `Loggerhead` log include error logs from third-party code

Etheryte:
## What does this PR change?

Currently our global logger only logs errors and warns from our own code which means we crucially miss errors thrown from React etc. This PR binds our logger to the global console so we should see more information from debug logs.  

Additionally I've added [performance marks](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark) so it's easier to line up events if the client makes a performance recording for us. Otherwise you need to jump between our own logs and the recording to get the full picture.  

(As a side effect, thanks to these changes I found a small bug in our login page where the locale was set too late so it was undefined for some scripts.)

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/22609

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/login/templates/login.jade
+++ b/java/code/src/com/suse/manager/webui/controllers/login/templates/login.jade
@@ -15,6 +15,9 @@ html(lang=window.preferredLocale.replace("_", "-"))
         // import styles
         link(rel='stylesheet', href='/css/#{webTheme}.css?cb=#{webBuildtimestamp}', id='web-theme')
         link(rel='stylesheet', href='/css/updated-#{webTheme}.css?cb=#{webBuildtimestamp}', id='updated-web-theme', disabled="disabled")
+        script(type='text/javascript').
+            window.preferredLocale = "#{preferredLocale}";
+            window.docsLocale = "#{docsLocale}";
         script(src='/javascript/jquery.js?cb=#{webBuildtimestamp}')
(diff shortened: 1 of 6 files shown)


java/code/src/com/suse/manager/webui/controllers/login/templates/login.jade
```diff
@@ -15,6 +15,9 @@ html(lang=window.preferredLocale.replace("_", "-"))
         // import styles
         link(rel='stylesheet', href='/css/#{webTheme}.css?cb=#{webBuildtimestamp}', id='web-theme')
         link(rel='stylesheet', href='/css/updated-#{webTheme}.css?cb=#{webBuildtimestamp}', id='updated-web-theme', disabled="disabled")
+        script(type='text/javascript').
+            window.preferredLocale = "#{preferredLocale}";
+            window.docsLocale = "#{docsLocale}";
```
parlt91, line 20: Good catch :)

parlt91 (approved):
LGTM I imagine this will come in handy

Etheryte:
Note to self: this needs to be ported to 4.3 once the branch is unlocked.

PR #7599, opened 2023-09-28
upload artifacts in github actions for debugging

jordimassaguerpla:
## What does this PR change?

CI: upload server logs as artifacts in github actions, so we can debug better. We need this for issues that are hard to reproduce, like when they only happen sometimes (flaky tests).

## GUI diff

No difference.

- [ ] **DONE**

## Documentation
- No documentation needed

- [ ] **DONE**

## Test coverage
- No tests

- [ ] **DONE**

## Links

N/A

- [ ] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

### DIFF
--- a/.github/workflows/acceptance_tests_common.yml
+++ b/.github/workflows/acceptance_tests_common.yml
@@ -5,6 +5,9 @@ on:
       secondary_tests:
         required: true
         type: string
+      server_id:
+        required: true
+        type: string
 env:
   UYUNI_PROJECT: uyuni-project
   UYUNI_VERSION: master
@@ -84,4 +87,13 @@ jobs:
         run: ./testsuite/podman_runner/18_split_secondary_p_tests.sh
       - name: run_secondary_tests
         run: ./testsuite/podman_runner/${{ inputs.secondary_tests }} 
+      - name: get_server_logs
+        if: ${{ failure() }}
+        run: ./testsuite/podman_runner/19_get_server_logs.sh ${{ inputs.server_id }}
+      - name: upload_server_log_artifacts
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: server_rhn_logs_${{ inputs.server_id }}
+          path: /tmp/test-all-in-one/server-logs/${{ inputs.server_id }}
 
--- a/testsuite/podman_runner/19_get_server_logs.sh
+++ b/testsuite/podman_runner/19_get_server_logs.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -xe
+
+if [ $# -ne 1 ];
+then
+	echo "Usage: ${0} server_id"
+	echo "server_id is used for creating a unique folder"
+	exit -1
+fi
+
+src_dir=$(cd $(dirname "$0")/../.. && pwd -P)
+server_id=${1}
+rm -rfv /tmp/test-all-in-one/server-logs/${server_id}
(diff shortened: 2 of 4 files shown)


.github/workflows/acceptance_tests_common.yml
```diff
@@ -84,4 +87,13 @@ jobs:
         run: ./testsuite/podman_runner/18_split_secondary_p_tests.sh
       - name: run_secondary_tests
         run: ./testsuite/podman_runner/${{ inputs.secondary_tests }} 
+      - name: get_server_logs
+        if: ${{ failure() }}
```
srbarrios, line 91: If it doesn't cost resources... I would save the logs always, not only when failing.

testsuite/podman_runner/19_get_server_logs.sh
```diff
@@ -0,0 +1,17 @@
...
+	echo "Usage: ${0} server_id"
+	echo "server_id is used for creating a unique folder"
+	exit -1
+fi
+
+src_dir=$(cd $(dirname "$0")/../.. && pwd -P)
+server_id=${1}
+rm -rfv /tmp/test-all-in-one/server-logs/${server_id}
+mkdir -p /tmp/test-all-in-one/server-logs/${server_id}
+sudo -i podman exec uyuni-server-all-in-one-test bash -c "tar jcvf /tmp/server-logs/${server_id}/var.log.rhn.tar.bz2 /var/log/rhn/"
```
srbarrios, line 15: We don't need those, are already included in the supportconfig.

testsuite/podman_runner/19_get_server_logs.sh
```diff
@@ -0,0 +1,17 @@
...
+sudo -i podman exec uyuni-server-all-in-one-test bash -c "supportconfig -R /tmp/server-logs/${server_id} && chmod 644 /tmp/server-logs/${server_id}/scc_uyuni-server-all-in-one-test*.txz*"
```
srbarrios, line 16: Be aware that for the supportconfig to collect proper logs from SUMA, we need to have installed the plugins for SUMA

.github/workflows/acceptance_tests_common.yml
jordimassaguerpla: It takes time ... 5 minutes aprox. There is a limit on disc space too, but I do not think  we have a problem there. I rather prefer to not take these extra 5 minutes than to have the logs when it succeeds.

testsuite/podman_runner/19_get_server_logs.sh
jordimassaguerpla: I checked the logs and I saw a spacewalk-debug folder. I assume it is installed, given we use the pattern to install the already released packages.

testsuite/podman_runner/19_get_server_logs.sh
jordimassaguerpla: ok. I have removed it .

.github/workflows/acceptance_tests_common.yml
jordimassaguerpla: What do you think?

srbarrios: Let's skip it, and if we ever find a good reason to have the logs in passed tests, we can think about adding them in some conditional way (using some label in the GitHub PR).

jordimassaguerpla: We agreed on slack that it is ok to skip.

PR #7622, opened 2023-10-02
We are constrained to sync SLES on Uyuni

Bischoff:
## What does this PR change?

The first commit is about cleanup:
* a few duplicated test lines removed
* When/And

The second commit does the real stuff:
* synchronizes SLES 15 SP 4 on Uyuni (it will go to build host and PXE booted hosts)
* makes sure to add a base channel to build host activation key
* create new activation key for terminal, with SLES 15 SP4 as base channel
* makes sure the PXE-bootstrapped hosts (both Cobbler and Retail) use the terminal activation key

The terminal and build host activation keys do not contain the fake channel because it is already on the minion, and the parent channel is different: Leap for the minion, and SLES for the build host and the terminal. If we want the fake packages like virgo-dummy, we need to create yet another fake channel (which I do in a second draft PR #7640).

It looks like it already works in a similar way in Build Validation - nothing changed there.

## Links

Ports:
* 4.3:

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/core/srv_create_activationkey.feature
+++ b/testsuite/features/core/srv_create_activationkey.feature
@@ -105,14 +105,23 @@ Feature: Create activation keys
     And I click on "Create Activation Key"
     Then I should see a "Activation key Proxy Key x86_64 has been created" text
 
+@buildhost
   Scenario: Create an activation key for the build host
     When I follow the left menu "Systems > Activation Keys"
     And I follow "Create Key"
     And I wait for child channels to appear
     And I enter "Build host Key x86_64" as "description"
     And I enter "BUILD-HOST-KEY-x86_64" as "key"
-    And I enter "20" as "usageLimit"
     And I check "Container Build Host"
     And I check "OS Image Build Host"
     And I click on "Create Activation Key"
     Then I should see a "Activation key Build host Key x86_64 has been created" text
+
+  Scenario: Create an activation key for the terminal
+    When I follow the left menu "Systems > Activation Keys"
+    And I follow "Create Key"
+    And I wait for child channels to appear
+    And I enter "Terminal Key x86_64" as "description"
+    And I enter "TERMINAL-KEY-x86_64" as "key"
+    And I click on "Create Activation Key"
+    Then I should see a "Activation key Terminal Key x86_64 has been created" text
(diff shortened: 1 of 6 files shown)


testsuite/features/core/srv_create_activationkey.feature
```diff
@@ -116,8 +116,3 @@ Feature: Create activation keys
     And I check "OS Image Build Host"
     And I click on "Create Activation Key"
     Then I should see a "Activation key Build host Key x86_64 has been created" text
-    And I should see a "Details" link
-    And I should see a "Packages" link
-    And I should see a "Configuration" link in the content area
-    And I should see a "Groups" link
-    And I should see a "Activated Systems" link
```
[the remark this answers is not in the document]
Bischoff: Yes. They are duplicate. Probably a copy and paste error.

vandabarata (approved):

testsuite/features/reposync/srv_sync_products.feature
```diff
@@ -62,6 +62,32 @@ Feature: Synchronize products in the products page of the Setup Wizard
...
+    When I select "Desktop Applications Module 15 SP4 x86_64" as a product
+    And I select "Development Tools Module 15 SP4 x86_64" as a product
+    Then I should see the "Desktop Applications Module 15 SP4 x86_64" selected
+    And I should see the "Development Tools Module 15 SP4 x86_64" selected
+    When I select "Containers Module 15 SP4 x86_64" as a product
+    Then I should see the "Containers Module 15 SP4 x86_64" selected
+    When I click the Add Product button
+    And I wait until I see "SUSE Linux Enterprise Server 15 SP4 x86_64" product has been added
+    Then the SLE15 SP4 product should be added
+
```
srbarrios, line 90: Can we consider not synchronizing the product, but just use custom repositories injected by sumaform on the build host and server?

srbarrios (changes requested):

testsuite/features/reposync/srv_sync_products.feature
Bischoff: I fear not. The problem is that the OS Image building process is based on reposynched channels, as far as I know.

When we use sumaform channels to "cheat", we are the ones to run a "zypper" command, so it works. But here we have no control on how OS image building is written.

I'll ask Ondrej to double-check that though.

Bischoff:
Closing in favour of other one

PR #7640, opened 2023-10-04
Sync sles too on uyuni 2

Bischoff:
## What does this PR change?

First commit is about cleanup and nitpicks:

* remove a few duplicated test lines removed
* When/And
* unique scenario names
* homogeneous filtering of packages

Second commit is a refactor done independantly by Oscar and me:

* isolate the reposync of fake channels

Third commit does the real stuff:

* synchronize SLES 15 SP 4 on Uyuni (it will go to build host and PXE booted hosts)
* add a base channel to build host activation key
* create new channel and activation key for terminal, with SLES 15 SP4 as base channel
* make sure the PXE-bootstrapped hosts (both Cobbler and Retail) use the build host activation key

**It should be followed by a new PR that kills SLES 15 SP4 reposync right after beginning and tests Cobbler on Leap.**

## Links

Ports:
* 4.3: https://github.com/SUSE/spacewalk/pull/22720

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/core/srv_create_activationkey.feature
+++ b/testsuite/features/core/srv_create_activationkey.feature
@@ -111,8 +111,16 @@ Feature: Create activation keys
     And I wait for child channels to appear
     And I enter "Build host Key x86_64" as "description"
     And I enter "BUILD-HOST-KEY-x86_64" as "key"
-    And I enter "20" as "usageLimit"
     And I check "Container Build Host"
     And I check "OS Image Build Host"
     And I click on "Create Activation Key"
     Then I should see a "Activation key Build host Key x86_64 has been created" text
+
+  Scenario: Create an activation key for the terminal
+    When I follow the left menu "Systems > Activation Keys"
+    And I follow "Create Key"
+    And I wait for child channels to appear
+    And I enter "Terminal Key x86_64" as "description"
+    And I enter "TERMINAL-KEY-x86_64" as "key"
+    And I click on "Create Activation Key"
+    Then I should see a "Activation key Terminal Key x86_64 has been created" text
--- a/testsuite/features/core/srv_create_activationkey.feature
+++ b/testsuite/features/core/srv_create_activationkey.feature
@@ -116,8 +116,3 @@ Feature: Create activation keys
     And I check "OS Image Build Host"
     And I click on "Create Activation Key"
     Then I should see a "Activation key Build host Key x86_64 has been created" text
-    And I should see a "Details" link
-    And I should see a "Packages" link
-    And I should see a "Configuration" link in the content area
-    And I should see a "Groups" link
-    And I should see a "Activated Systems" link
--- a/testsuite/features/init_clients/allcli_update_activationkeys.feature
+++ b/testsuite/features/init_clients/allcli_update_activationkeys.feature
@@ -148,7 +148,6 @@ Feature: Update activation keys
     Then I should see a "Activation key Proxy Key x86_64 has been modified" text
 
 @scc_credentials
-@susemanager
   Scenario: Update build host key with synced base product
     When I follow the left menu "Systems > Activation Keys"
     And I follow "Build host Key x86_64" in the content area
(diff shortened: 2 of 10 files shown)


testsuite/features/init_clients/allcli_update_activationkeys.feature
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2022-2023 SUSE LLC
 # Licensed under the terms of the MIT license.
 
+@skip_if_github_validation
```
srbarrios, line 4: I don't think there is a real need for this tag here, as we never trigger this feature in our run_sets yaml files on the context of github_validation.
But I can understand it as a valid safety net, just in case.

testsuite/features/reposync/srv_sync_fake_channels.feature
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2022-2023 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Prepare fake package channels
```
srbarrios, line 4: Actually, packages and patches.

testsuite/features/core/srv_create_activationkey.feature
```diff
@@ -116,8 +116,3 @@ Feature: Create activation keys
     And I check "OS Image Build Host"
     And I click on "Create Activation Key"
     Then I should see a "Activation key Build host Key x86_64 has been created" text
-    And I should see a "Details" link
-    And I should see a "Packages" link
-    And I should see a "Configuration" link in the content area
-    And I should see a "Groups" link
-    And I should see a "Activated Systems" link
```
srbarrios, line 123: Are we losing coverage here? Or having these fields on the WebUI is covered somewhere else?

testsuite/features/reposync/srv_sync_fake_channels.feature
```diff
@@ -0,0 +1,92 @@
...
+    And I click on "Sync Now"
+    Then I should see a "Repository sync scheduled for Fake-RPM-SUSE-Channel." text
+    And I wait until the channel "fake-rpm-suse-channel" has been synced
+    And I disable source package syncing
+
+  Scenario: Verify state of Fake-RPM-SUSE-Channel custom channel
+    Then "orion-dummy-1.1-1.1.x86_64.rpm" package should have been stored
+    And solver file for "fake-rpm-suse-channel" should reference "orion-dummy-1.1-1.1.x86_64.rpm"
+
+@uyuni
```
srbarrios, line 51: I would add the tag here (and the other scenarios related) to skip if the deployment of a build host VM is not present.

testsuite/features/reposync/srv_sync_fake_channels.feature
```diff
@@ -0,0 +1,92 @@
...
+  Scenario: Add the build host child channel to the base product channel
+    When I follow the left menu "Software > Manage > Channels"
+    And I follow "Create Channel"
+    And I enter "Fake-RPM-Buildhost-Channel" as "Channel Name"
+    And I enter "fake-rpm-buildhost-channel" as "Channel Label"
+    And I select "SLE-Product-SLES15-SP4-Pool for x86_64" from "Parent Channel"
```
srbarrios, line 57:
```suggestion
     And I select the parent channel for the "build_host" from "Parent Channel"
```

testsuite/features/reposync/srv_sync_products.feature
```diff
@@ -62,6 +62,32 @@ Feature: Synchronize products in the products page of the Setup Wizard
...
+    And I should see the "Server Applications Module 15 SP4 x86_64" selected
+    When I select "Desktop Applications Module 15 SP4 x86_64" as a product
+    And I select "Development Tools Module 15 SP4 x86_64" as a product
+    Then I should see the "Desktop Applications Module 15 SP4 x86_64" selected
+    And I should see the "Development Tools Module 15 SP4 x86_64" selected
+    When I select "Containers Module 15 SP4 x86_64" as a product
+    Then I should see the "Containers Module 15 SP4 x86_64" selected
+    When I click the Add Product button
+    And I wait until I see "SUSE Linux Enterprise Server 15 SP4 x86_64" product has been added
+    Then the SLE15 SP4 product should be added
```
srbarrios, line 89: Can't we skip this synchronized then?
I understood by your discussion with Ondrej, that having on the build host VM our zypper repositories created coming from OBS development channels would be enough (having those injected through sumaform as we were doing till not many months ago). We will need ofc to enable/disable them when necessary.

That approach would be ideal for our long-term, as if we synchronize the full SLES15SP4 product in Uyuni CI just for this need, we will spending a precious time, and might not be adding real value if what we need here is the content and not that important the fact that this content comes from SCC repositories.

testsuite/features/init_clients/allcli_update_activationkeys.feature
Bischoff: Strange then that Jordi has put it on every scenario of the end of that file.

testsuite/features/reposync/srv_sync_fake_channels.feature
Bischoff: Correct. I will add "patches".

While we are at it, we have an organization problem here:

We have other "fake" channels, for Debian-Like and Red Hat-like. But they are not covered by this file.

I'm not exactly sure why it is so.

testsuite/features/core/srv_create_activationkey.feature
Bischoff: Explained in the PR description: it's duplicated lines, probably by copy and paste accident.

So no we don't lose coverage. Same lines are above in same file.

testsuite/features/reposync/srv_sync_fake_channels.feature
Bischoff: ok.

testsuite/features/reposync/srv_sync_fake_channels.feature
Bischoff: It was a way of insisting heavily that it never changes, even on Uyuni.

Also, that construct will not work for the terminal. If we test later on Cobbler with Leap, the PXE boot minion will sometimes be a Leap (Cobbler) and sometimes a SLES (Retail). 

Finally, the corresponding entry does not exist (yet) in `BASE_CHANNEL_BY_CLIENT` in `constants.rb`. Which can be fixed, of course.

I'd vote for keeping it.

testsuite/features/reposync/srv_sync_fake_channels.feature
srbarrios: Oh no, absolutely right...
They are here... reposync/srv_create_repository.feature
(Mine would be the neverend PR, but I will try to fix it on mine)

testsuite/features/reposync/srv_sync_products.feature
Bischoff: You mean kill the synchronization ? Yes, that's the idea of a second PR after this one, as explained in the description.

Forget the problem of zypper repositories on build host, it's already solved. We cheat with a zypper install from local repos.

I also discovered that we already directly get the packages for the OS image from our mirror, so we're good there too if we don't reposync.

The problem is with Cobbler. There are two possible solutions at this time (not mutually exclusive):
 * use Leap as test base for Cobbler (1)
 * cheat in the autoyast file to get the packages directly from external repos or our mirror, like we do for Retail (2)

I'd like to first fix the Cobbler and Retail current problems, then try to shorten up the reposync with another PR.

---

 (1) but then Leap must be fully synchronized (so I'm not sure we really win something...)
 (2) but then we bypass completely the testing of autoyast variables like here:
```
 <media_url>http://$redhat_management_server/ks/dist/child/sle-product-sles15-sp4-updates-x86_64/$distrotree</media_url>
 ```
 and we don't test we have the correct packages on the suse manager server

testsuite/features/reposync/srv_sync_fake_channels.feature
srbarrios: OK, I can fix it on my PR.
So, one of the main ideas of my refactor is also to concentrate these kind of text values in the map we have on constants, that way if we every need to change it, there is only one place to change.

testsuite/features/reposync/srv_sync_fake_channels.feature
Bischoff: Please, no :smiley:  LOL... just keep it in mind, otherwise you will never end and me neither :smile:

testsuite/features/reposync/srv_sync_fake_channels.feature
Bischoff: ok.

testsuite/features/reposync/srv_sync_products.feature
srbarrios: I like option 2, but I will give a though.

testsuite/features/reposync/srv_sync_fake_channels.feature
Bischoff: I can't do it immediately because we don't have a @pxeboot tag

testsuite/features/reposync/srv_sync_products.feature
Bischoff: It's even likely it's the only one that will work out.

I need to discuss with Dominik and/or you. Didn't we kill the reposync for leap (uyuni) and sles? if we did, how could cobbler tests work? Maybe they didn't and we did never relate this to a source of instability?

nodeg (approved):
I did not find any major issues other than what Oscar already mentioned. I am waiting for the second follow-up PR.

PR #7651, opened 2023-10-06
Master - move all package files from /srv

mbussolotto:
## What does this PR change?

Use /usr/share/susemanager for all the files currently in /srv and owned by a RPM
Keep runtime, tmp and user file in /srv

Related changes :

susemanager-frontend-libs: https://build.opensuse.org/package/show/home:mbussolotto:branches:systemsmanagement:Uyuni:Master:ServerContainer/susemanager-frontend-libs
uyuni docs: https://build.opensuse.org/package/show/home:mbussolotto:branches:systemsmanagement:Uyuni:Master:ServerContainer/uyuni-docs_en
release notes: https://build.opensuse.org/package/show/home:mbussolotto:branches:systemsmanagement:Uyuni:Master:ServerContainer/release-notes-uyuni

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- Documentation issue was created https://github.com/uyuni-project/uyuni-docs/issues/2543

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/22539

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


mbussolotto:
acceptance test https://ci.suse.de/view/Manager/view/Manager-Test/job/manager-TEST-Hexagon-acceptance-tests/  >= 1042

avshiliaev (approved):
Looks good and most importantly seems to be building just fine on my side :smile:

aaannz (commented):
Overall looks good to me, I have one question about migrations.

susemanager/bin/mgr-setup
```diff
@@ -423,6 +423,26 @@ if [ -f $MANAGER_COMPLETE ]; then
...
+    sed -i 's|ScriptAlias /cgi-bin/ "/srv|ScriptAlias /cgi-bin/ "/usr/share/susemanager|g' /etc/apache2/default-server.conf
+    sed -i 's|DocumentRoot "/srv|DocumentRoot "/usr/share/susemanager|g' /etc/apache2/vhosts.d/vhost-ssl.conf
+    sed -i 's|DocumentRoot "/srv|DocumentRoot "/usr/share/susemanager|g' /etc/apache2/vhosts.d/vhost-ssl.template
+}
+
+#TODO move this on the permission check function
+setup_permission() {
+    chown tomcat:tomcat -R /usr/share/susemanager/www/htdocs
+    chmod 777 -R /usr/share/susemanager/www/htdocs
+}
```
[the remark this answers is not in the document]
mackdk: I was wondering the same

mackdk (approved):

susemanager/bin/mgr-setup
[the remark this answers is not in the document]
mbussolotto: right, currently upgrade is not managed at all. I need to add somewhere this part

branding/spacewalk-branding.spec
```diff
@@ -24,8 +24,10 @@
 %global wwwdocroot %{_var}/www/html
 %else
 %if 0%{?suse_version}
-%global tomcat_path /srv/tomcat
-%global wwwdocroot /srv/www/htdocs
+%global susemanager_shared_path /usr/share/susemanager
```
deneb-alpha, line 27: Do we  really want to use `susemanager` as name for the  folder? In this way we are  going to hardcode the same name  also in uyuni. If we really want to have the product name as name  of the folder, what about  having a conditional that set `uyuni` or `susemanager` according to the product?

cbosdo: Why not even set `uyuni` for all?

susemanager/bin/mgr-setup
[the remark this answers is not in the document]
cbosdo: Why do we need to change those permissions? Isn't it enough to set them on the folder in the spec file?

branding/spacewalk-branding.spec
mbussolotto: so do we want to replace `/usr/share/susemanager` to `/usr/share/uyuni` (also for the files that were there before my refactoring)?

susemanager/bin/mgr-setup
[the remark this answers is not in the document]
mbussolotto: It should, I just add that for keeping it easy. Once the tests will be ok, I'd like to remove `setup_permission` at all

branding/spacewalk-branding.spec
cbosdo: This is a bigger move and we should probably not consider it for this PR, but IMHO this is something we would need to think about at some point.

deneb-alpha: > please not change it twice in different official updates. This require again migration effort
I agree. I think it could be better to have the naming fixed now and not in two  steps

susemanager/bin/mgr-setup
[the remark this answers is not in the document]
mbussolotto: I think this commit should fix the concern https://github.com/uyuni-project/uyuni/pull/7651/commits/ef506192fc1cead42a0e52860c9fd2c954a1a29b

Instead of manipulating files, I created new ones (and disable the previous ones). In this way the upgrade should be ok. I'm now going to test if permission are ok. Please let me know what you think about this change

branding/spacewalk-branding.spec
rjmateus: Agree, we should define the name once and change it for good

cbosdo: What about doing that change in a separate PR, but merge the two in the same release. We can even merge them together, but that would make the whole change easier to review... I'ld bet this would add quite some diffs more

mcalmer: For me this would be ok. I only want to prevent, that we do an uyuni release with this name and have to change it later to another including a migration script

aaannz: My 2c is to keep susemanager everywhere. Even though SUSE Manager is not the best name, it is established name which our end goal is to promote the most as that is what is paying our bills at the end.

In my opinion we should rather focus on renaming all spacewalk things to susemanager then susemanager to uyuni.

deneb-alpha: my 2 cents... 
I would call it `uyuni` and have it released with the next 2023.10. it's ok to  do it  in 2 PRs, but  please, don't split this change in two releases

spacewalk/config/spacewalk-config.spec
```diff
@@ -136,6 +140,12 @@ sysconf_addword /etc/sysconfig/apache2 APACHE_SERVER_FLAGS SSL
 sysconf_addword /etc/sysconfig/apache2 APACHE_SERVER_FLAGS ISSUSE
 %endif
 
+if [ -f /etc/apache2/vhosts.d/vhost-ssl.conf ]; then
```
rjmateus, line 143: I think we can use this logic to update the DocRoot, which is your goal. If the file exists, then we can check if the value is the one we want, otherwise, we replace it.
If the file exists and has the wanted value, let it stay.

On top of this if we update the tool  `spacewalk-setup-httpd`, newly generated files will have the new location.
Calling the tool automatically can be more tricky since we don't know which parameters where used.

spacewalk/config/spacewalk-config.spec
```diff
@@ -136,6 +140,12 @@ sysconf_addword /etc/sysconfig/apache2 APACHE_SERVER_FLAGS SSL
...
+    mv  /etc/apache2/vhosts.d/vhost-ssl.conf /etc/apache2/vhosts.d/vhost-ssl.conf.bak
+fi
+if [ -f /etc/apache2/vhosts.d/vhost.template ]; then
+    mv  /etc/apache2/vhosts.d/vhost.template /etc/apache2/vhosts.d/vhost.template.bak
```
rjmateus, line 147: we would not need to touch the template, but the setup tool that uses it.

branding/spacewalk-branding.spec
cbosdo: after so many cents we may manage to build something like this one:
image

rjmateus (commented):
I have a few questions, but overall looks almost done. Thank you for the great work :+1:

lucidd (approved):

deneb-alpha:
@mbussolotto the tests are still failing. please let me know what  to do also with the SRs you prepared

mbussolotto:
I don't think error have something to do with the PR. @jordimassaguerpla can you please confirm it?

mbussolotto:
@deneb-alpha the SR should be approved as soon this PR is merged. let's see the feedback from @jordimassaguerpla

deneb-alpha:
Ok, let's sync for the merge, and if I'm not there, please coordinate with @raulillo82 

Info: the SRs for Head can be merged too, I checked with Julio already

jordimassaguerpla:
you can merge, the failures are not related to this PR but to the build service not being available

mbussolotto:
@sbluhm Thanks for all your feedback! Consider that, besides these changes, there was also this in OBS project:

- https://build.opensuse.org/package/rdiff/systemsmanagement:Uyuni:Master/uyuni-docs_en?linkrev=base&rev=65
- https://build.opensuse.org/package/rdiff/systemsmanagement:Uyuni:Master/release-notes-uyuni?linkrev=base&rev=170
- https://build.opensuse.org/package/rdiff/systemsmanagement:Uyuni:Master/susemanager-frontend-libs?linkrev=base&rev=17

mbussolotto:
@sbluhm there was some error in EL9, now everything should be fine, sorry about that. Anyway, I think your assumption are right, we can get rid of all `%if` `%else` in spec file using the macro `%_datadir`

sbluhm:
@mbussolotto , I have done some changes and a few reverts and am currently testing them on https://github.com/sbluhm/uyuni/tree/userserverdir. Let me see what you have changed.

Ideally don't change too much at the moment :)

sbluhm:
FYI: Overall, I will follow your changes for EL as it simplifies a lot. So all the SUSE specific "if"s can go out.
Only thing I am keeping in are probably the "pub" folder locations on `/srv` or `/var/www/`.

PR #7660, opened 2023-10-09
Consider server id when removing invalid erratas from rhnSet

wweellddeerr:
## What does this PR change?

**Context**
After a successful patch installation, an update to the server needed cache is scheduled to run, but it takes up to a minute to effectively run and reflect on the WebUI. In this interval, an already installed patch is available for selection, and selecting it at this moment marks it to be installed again. After some seconds, the patch disappear from the screen and there is no chance to remove the it after that, only running a database script.

**Changes in this PR:**
This PR takes the server id into consideration when removing invalid erratas from `rhnSet`. With that we avoid the situation where the errata is invalid for a server but is in the `rhnServerNeededErrataCache` because of it is valid for other server.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: legacy code

- [x] **DONE**

## Links

Fixes #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/spacewalk-java.changes.welder.improve-invalid-errata-delete
+++ b/java/spacewalk-java.changes.welder.improve-invalid-errata-delete
@@ -1,0 +1,1 @@
+- Consider server id when removing invalid erratas from rhnSet (bsc#1204235,bsc#1207012,bsc#1211560)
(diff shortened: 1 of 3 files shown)


java/spacewalk-java.changes.welder.improve-invalid-errata-delete
```diff
@@ -0,0 +1 @@
+- Consider server id when removing invalid erratas from rhnSet
```
mcalmer, line 1: Was this a bug? Can you put a bug number here?

mcalmer (approved):
as now everything is green, this looks good

Issue #4746, opened 2022-01-30
Uyuni Bugreport - Uyuni 2022.01 - reposync imports failed downloads


sbluhm:
# Additional Information

Reposync seems to continue importing and linking packages to channels despite the download failing.

## Version of Uyuni Server and Proxy (if used)
```
Name         : Uyuni-Server-release
Version      : 2022.01
Release      : 167.1.uyuni1
Architecture : x86_64
Size         : 1.4 k
Quelle       : Uyuni-Server-release-2022.01-167.1.uyuni1.src.rpm
Repository   : @System
Aus Paketque : @commandline
Summary      : Uyuni Server
```
## Details about issue
Issue was identified when using an unpatched python-urlgrabber 4.1.
Urlgrabber fails but import and linking of those packages still continue.

Expected behaviour: failed packages are not imported (and not linked).

See below log. Note that linking was manually interrupted so log is incomplete after line "Linking packages to channel":
```
[... all downloads failing ...]
2022/01/29 10:53:25 +02:00     90/92 : venv-salt-minion-3002.2-40.4.uyuni.x86_64.rpm (failed)
2022/01/29 10:53:25 +02:00 ERROR: Download failed: https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/EL8-Uyuni-Client-Tools/EL_8/x86_64/zeromq-devel-4.1.6-11.3.uyuni.x86_64.rpm - Target file isn't valid. Checksum should be 0ce6c401f9b94db304d7a22231c7e163b9812ff4266aba924324e09d5fb5d9b1 (sha256)..
2022/01/29 10:53:25 +02:00     91/92 : zeromq-devel-4.1.6-11.3.uyuni.x86_64.rpm (failed)
2022/01/29 10:53:25 +02:00 ERROR: Download failed: https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable:/EL8-Uyuni-Client-Tools/EL_8/noarch/scap-security-guide-redhat-0.1.59-1.27.1.uyuni.noarch.rpm - Target file isn't valid. Checksum should be 13e6c61e4eb60a07ff0706e43fb40631ecc4beb4059bbe4f458178afd0b841ca (sha256)..
2022/01/29 10:53:25 +02:00     92/92 : scap-security-guide-redhat-0.1.59-1.27.1.uyuni.noarch.rpm (failed)
2022/01/29 10:53:26 +02:00 Importing packages started.
[...]

PR #7666, opened 2023-10-10
fix(4746): Do not link and sync pkgs that fail to download

m-czernek:
## What does this PR change?

When a package fails to download, `reposync.py` still tries to process and link it internally. This PR fixes https://github.com/uyuni-project/uyuni/issues/4746

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [x] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/python/spacewalk/satellite_tools/download.py
+++ b/python/spacewalk/satellite_tools/download.py
@@ -153,8 +153,7 @@ class DownloadThread(Thread):
         # pylint: disable=E1101
         self.curl = pycurl.Curl()
         self.mirror = 0
-        self.failed_pkgs = []
-        self.lock = Lock()
+        self.failed_pkgs = set()
 
     @staticmethod
     def __is_file_done(local_path=None, file_obj=None, checksum_type=None, checksum=None):
@@ -285,11 +284,9 @@ class DownloadThread(Thread):
                 self.parent.log_obj.log(success, os.path.basename(params['relative_path']))
             self.queue.task_done()
             if not success:
-                self.lock.acquire()
                 package = os.path.basename(params['relative_path'])
-                # append package without extension
-                self.failed_pkgs.append(package.rsplit('.', maxsplit=1)[0])
-                self.lock.release()
+                # append package name without extension
+                self.failed_pkgs.add(package.rsplit('.', maxsplit=1)[0])
         self.curl.close()
 
 
@@ -328,7 +325,7 @@ class ThreadedDownloader:
         # WORKAROUND - BZ #1439758 - ensure first item in queue is performed alone to properly setup NSS
         self.first_in_queue_done = False
         self.first_in_queue_lock = Lock()
-        self.failed_pkgs = []
+        self.failed_pkgs = set()
 
     def set_log_obj(self, log_obj):
         self.log_obj = log_obj
@@ -383,12 +380,8 @@ class ThreadedDownloader:
                 while any(t.is_alive() for t in started_threads):
                     time.sleep(1)
                 break
-            # accumulate all failed packages, if any
-            failed_pkgs = [pkg for t in started_threads for pkg in t.failed_pkgs]
-            if bool(failed_pkgs):
-                self.lock.acquire()
-                self.failed_pkgs = failed_pkgs
-                self.lock.release()
+            # accumulate all failed packages
+            self.failed_pkgs = {pkg for t in started_threads for pkg in t.failed_pkgs}
 
         # raise first detected exception if any
         if self.exception:
--- a/python/spacewalk/satellite_tools/download.py
+++ b/python/spacewalk/satellite_tools/download.py
@@ -284,9 +284,8 @@ class DownloadThread(Thread):
                 self.parent.log_obj.log(success, os.path.basename(params['relative_path']))
             self.queue.task_done()
             if not success:
-                package = os.path.basename(params['relative_path'])
-                # append package name without extension
-                self.failed_pkgs.add(package.rsplit('.', maxsplit=1)[0])
+                package = os.path.basename(params['target_file'])
+                self.failed_pkgs.add(package)
         self.curl.close()
 
 
--- a/python/spacewalk/satellite_tools/download.py
+++ b/python/spacewalk/satellite_tools/download.py
@@ -153,6 +153,8 @@ class DownloadThread(Thread):
         # pylint: disable=E1101
         self.curl = pycurl.Curl()
         self.mirror = 0
+        self.failed_pkgs = []
+        self.lock = Lock()
 
     @staticmethod
     def __is_file_done(local_path=None, file_obj=None, checksum_type=None, checksum=None):
@@ -282,6 +284,12 @@ class DownloadThread(Thread):
                 # log_obj must be thread-safe
                 self.parent.log_obj.log(success, os.path.basename(params['relative_path']))
             self.queue.task_done()
+            if not success:
+                self.lock.acquire()
+                package = os.path.basename(params['relative_path'])
+                # append package without extension
+                self.failed_pkgs.append(package.rsplit('.', maxsplit=1)[0])
+                self.lock.release()
         self.curl.close()
 
 
@@ -320,6 +328,7 @@ class ThreadedDownloader:
         # WORKAROUND - BZ #1439758 - ensure first item in queue is performed alone to properly setup NSS
         self.first_in_queue_done = False
         self.first_in_queue_lock = Lock()
+        self.failed_pkgs = []
 
     def set_log_obj(self, log_obj):
         self.log_obj = log_obj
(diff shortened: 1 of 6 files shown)


python/spacewalk/satellite_tools/download.py
```diff
@@ -282,6 +284,12 @@ def run(self):
                 # log_obj must be thread-safe
                 self.parent.log_obj.log(success, os.path.basename(params['relative_path']))
             self.queue.task_done()
+            if not success:
+                self.lock.acquire()
```
m-czernek, line 288: Not totally sure if these locks are necessary in the threads themselves

python/spacewalk/satellite_tools/reposync.py
```diff
@@ -1152,6 +1152,9 @@ def import_packages(self, plug, source_id, url, is_non_local_repo):
         downloader.set_log_obj(logger)
         downloader.run()
 
+        log(0, 'Filtering packages that failed to download')
+        to_process = [i for i in to_process if i[0].getNVREA() not in downloader.failed_pkgs]
```
agraul, line 1156: Worst case this is `O(n^2)`, right? `downloader.failed_pkgs` will in most cases be a short list so this does not matter much, but a set could also make sense for `downloader.failed_pkgs`

agraul (commented):
I'm still trying to follow the multithreading logic, we might not need to lock these.

python/spacewalk/satellite_tools/reposync.py
m-czernek: Makes sense; it was my assumption that the list of failed pkgs won't be long, but you're totally right about the set.

python/spacewalk/satellite_tools/download.py
```diff
@@ -374,6 +383,12 @@ def run(self):
                 while any(t.is_alive() for t in started_threads):
                     time.sleep(1)
                 break
+            # accumulate all failed packages, if any
+            failed_pkgs = [pkg for t in started_threads for pkg in t.failed_pkgs]
+            if bool(failed_pkgs):
+                self.lock.acquire()
+                self.failed_pkgs = failed_pkgs
+                self.lock.release()
```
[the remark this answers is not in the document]
m-czernek: I have the same intuition, though I wasn't totally sure I'm not missing something. Removed the thread locking

python/spacewalk/satellite_tools/download.py
```diff
@@ -282,6 +283,10 @@ def run(self):
                 # log_obj must be thread-safe
                 self.parent.log_obj.log(success, os.path.basename(params['relative_path']))
             self.queue.task_done()
+            if not success:
+                package = os.path.basename(params['relative_path'])
+                # append package name without extension
+                self.failed_pkgs.add(package.rsplit('.', maxsplit=1)[0])
```
vzhestkov, line 289: It makes me a bit confused. Actually afterwards it's checking with NVREA (strange by itself as usually it's called NEVRA as Epoch has the higher priority over the Version, but anywat) the potential problem here, that some of the third party repos are not folliwing the rule to call the package files with this naming convention. For exmple: https://rpm.opera.com/rpm. It contains `opera_stable-103.0.4928.26-linux-release-x64-signed.rpm` with `NEVRA`=`google-chrome-stable-118.0.5993.70-1.x86_64`

m-czernek: Nice, makes sense! Now, instead of relying on the package naming rules, I rely on paths. This means that there's no need for removing the extension anymore, and from my understanding, the target path should always be defined.

WDYT about the current version?

python/test/unit/spacewalk/satellite_tools/test_reposync.py
```diff
@@ -212,6 +212,63 @@ def test_sync_success_regen(self):
...
+        for i, pack in enumerate(packs):
+            pack.arch = "arch1"
+            pack.checksum = ""
+            # The first package failed to download
+            pack.unique_id.relativepath = fail_pkg_name if i == 0 else "pkg"
+
+        plug =  self._mock_repo_plugin(packs)
+
+        with patch("uyuni.common.context_managers.CFG", self._mock_cfg()), \
+            patch.object(spacewalk.satellite_tools.reposync.RepoSync, "chunks", Mock(return_value=())):
```
agraul, line 264: What exactly is this doing? Why is it needed?

python/test/unit/spacewalk/satellite_tools/test_reposync.py
```diff
@@ -212,6 +212,63 @@ def test_sync_success_regen(self):
...
+            rs.import_packages(plug, None, "unused-url-string", None)
+
+        to_process = apply_async_mock.call_args_list[0][1]["args"][0]
+        self.assertTrue(len(to_process) == 1)
+
+        pack_to_process = to_process[0][0]
+        self.assertTrue(pack_to_process.unique_id.relativepath != fail_pkg_name)
```
agraul, line 271: Do I read this correctly that first `to_process` is set to a "test data" defined in this test case and then the assertions are done on `to_process`?

python/test/unit/spacewalk/satellite_tools/test_reposync.py
```diff
@@ -212,6 +212,63 @@ def test_sync_success_regen(self):
...
+    @patch("uyuni.common.context_managers.initCFG", Mock())
+    @patch("spacewalk.satellite_tools.reposync.log2", Mock())
+    @patch("spacewalk.satellite_tools.reposync.os", os)
+    @patch("spacewalk.satellite_tools.reposync.ThreadedDownloader")
+    @patch("spacewalk.satellite_tools.reposync.multiprocessing.Pool")
+    def test_import_packages_excludes_failed_pkgs(self, pool, downloader):
+        """
+        When downloader fails to download a subset of packages
+        Then the RepoSync.import_packages function should not process the failed packages
+        """
```
agraul, line 240: I find this test a bit hard to follow. It's not obvious what's the system under test (SUT) and what are the collaborators (that can be substituted by test doubles). Could you maybe re-structure this test a bit and maybe add a few comments?

agraul (changes requested):
Thanks for adding tests! I know it's not easy with our reposync test suite, but you've done a good job!

I have left a few comments about a few details, once those are resolved we can clean up the git history and merge!

python/test/unit/spacewalk/satellite_tools/test_reposync.py
m-czernek: It is absolutely hard to follow. Honestly, the function under test is hard to follow as well :/ It does multiple things with multiple objects that we do not care about in this particular test, hence the number of patches.

I have added a few comments and refactored the test to make it just a little bit less noisy. However, the main collaborators should be objects injected by pytest into the method parameter list, right?

python/test/unit/spacewalk/satellite_tools/test_download.py
```diff
@@ -117,3 +145,34 @@ def test_reposync_download_thread_fetch_url_proxy_pass():
         assert "proxy" not in url_grabber_opts_mock.mock_calls[1].kwargs
         assert "username" not in url_grabber_opts_mock.mock_calls[1].kwargs
         assert "password" not in url_grabber_opts_mock.mock_calls[1].kwargs
+
+
+def test_reposync_download_thread_sets_failed_pkg():
```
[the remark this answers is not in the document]
agraul: For me it's okay to keep 1) The tests execute quickly enough that the execution time is not impacted noticeably while 1) can help pinpointing the source of problems. :+1:

python/test/unit/spacewalk/satellite_tools/test_reposync.py
agraul: You're totally right, `import_packages` is doing too much.

Thanks for adding the comments, I think the test is better now. It's not pretty but following the function, it's understandable.

python/test/unit/spacewalk/satellite_tools/test_reposync.py
agraul: The reason that `SQLBackend` and `ChannelPackageSubscription` need to be mocked is that there is a package that "was downloaded", right? If so, I get it now.

python/test/unit/spacewalk/satellite_tools/test_reposync.py
agraul: The names are fine. I was mostly confused why `to_process` is taken from `apply_async_mock.call_args_list[0][1]["args"][0]`.

python/test/unit/spacewalk/satellite_tools/test_reposync.py
m-czernek: As per our discussions, simplified test -> removed this mock.

python/test/unit/spacewalk/satellite_tools/test_reposync.py
m-czernek: As per our discussion, simplified the test -> removed this call.

python/spacewalk/satellite_tools/download.py
vzhestkov: Thanks for fixing it :+1: Looks good for me.

vzhestkov (approved):
Looks good :+1: 
Thanks for the fix.

agraul (approved):
LGTM! :rocket:

PR #7685, opened 2023-10-13
Recognize squashfs build results from kiwi

aaannz:
## What does this PR change?

Our kiwi build inspection was missing squashfs output format. This PR adds recognition of it.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: integration tests cover for regressions, no unit tests are present

- [X] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/22772
Tracks https://github.com/SUSE/spacewalk/pull/22743

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
+++ b/susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
@@ -198,6 +198,7 @@ _compression_types = [
     {"suffix": ".gz", "compression": "gzip"},
     {"suffix": ".bz", "compression": "bzip"},
     {"suffix": ".xz", "compression": "xz"},
+    {"suffix": ".tar.xz", "compression": "xz"},
     {"suffix": ".install.iso", "compression": None},
     {"suffix": ".iso", "compression": None},
     {"suffix": ".qcow2", "compression": None},
@@ -208,6 +209,7 @@ _compression_types = [
     {"suffix": ".vhdx", "compression": None},
     {"suffix": ".vdi", "compression": None},
     {"suffix": ".raw", "compression": None},
+    {"suffix": ".squashfs", "compression": None},
     {"suffix": "", "compression": None},
 ]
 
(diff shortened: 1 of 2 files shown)


agraul (commented):
The change itself looks good, but I'm questioning the approach of having to a list of known suffixes. Not something we need to fix in this PR, but maybe we can think about it.

I noticed you've put "No tests: already covered" in the PR description. Can you point me to the test case that covers squashfs build results? I'm a bit surprised this is already covered since the changelog entry mentions a bug report.

Please keep in mind that merging this PR will close the Bugzilla-tracking GH issue. You might need to re-open it.

aaannz:
I don't know about some key how to select the correct image result from the kiwi build. File size is not correct as you can have compressed images which are smaller then original image also present in the build results directory. So we opted for suffixes list and instead of inspecting all possible kiwi results we are doing this per-bug basis as customers needs.

We have tests with regards to overall regressions. It is true that we don't have any squashfs related tests. We don't have any unit tests for this particular code, only integration tests and those tests only KIS images.

agraul:
> I don't know about correct key how to select the end image result from the kiwi build. File size is not correct as you can have compressed images which are smaller then original image also present in the build results directory. So we opted for suffixes list and instead of inspecting all possible kiwi results we are doing this per-bug basis as customers needs.
> 
I see. I haven't used Kiwi much, maybe this is a good way to save time then.


> We have tests with regards to overall regressions. It is true that we don't have any squashfs related tests. We don't have any unit tests for this particular code, only integration tests and those tests only KIS images.

The way I see it, test coverage applies to the change and not the component that this change touches. How much work would it be to add tests for "finding the correct artifact"?

aaannz:
Still valid, not enough time to focus on this.

m-czernek (approved):
:+1: LGTM, apologies for taking so long.

PR #7701, opened 2023-10-17
Drop Select2

Etheryte:
## What does this PR change?

Select2 is a dropdown plugin left over from the jQuery days, right now it's only used in a few views which haven't been updated to get rid of it. We got some Dependabot alerts for it and the upgrade path was painful so I went ahead and refactored it out to use our more modern form components instead.

## GUI diff

Cosmetic differences in how the dropdown looks, now those dropdowns look the same everywhere.

- [x] **DONE**

## Documentation
- No documentation needed: Dropping a library so we use the same one everywhere.

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

A part of https://github.com/SUSE/spacewalk/issues/22581

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


web/html/src/manager/visualization/hierarchy.tsx
```diff
@@ -262,13 +263,131 @@ class Hierarchy extends React.Component {
...
     return (
       <TopPanel title={t(window.title)} helpUrl={hurl}>
-        <button className="toggle-filter-button" onClick={this.showFilters}>
-          {t((this.state.showFilters ? "Hide" : "Show") + " filters")}
+        <button className="toggle-filter-button" onClick={() => this.toggleFilters()}>
+          {this.state.showFilters ? t("Hide filters") : t("Show filters")}
           <i className={"fa fa-caret-" + (this.state.showFilters ? "up" : "down")} aria-hidden="true"></i>
         </button>
-        <div id="visualization-filter-wrapper"></div>
+        <div id="visualization-filter-wrapper">
```
Etheryte, line 277: This is not actually new DOM, this is the DOM the old jQuery code created, lifted out and bound to React so we can use our own form components.

web/html/src/manager/visualization/hierarchy.tsx
```diff
@@ -53,140 +34,202 @@ function showFilterTab(tabIdToShow) {
...
-  filterNavTab
-    .append("li")
-    .attr("id", "partitioning-tab-selector")
-    .attr("class", "filter-tab-selector")
-    .append("a")
-    .text(t("Partitioning"))
-    .on("click", (d) => {
-      showFilterTab("partitioning-tab");
-    });
+// NB! This is a magic constant
```
Etheryte, line 37: Here and below, this is not actually new code, almost all of it has been lifted out of the old jQuery etc code and modified to work with React.

Etheryte:
Fyi for QE: I also went ahead and updated/removed the relevant code from step definitions as well, so far PR acceptance tests fail at an earlier stage so I'm still waiting for the results on those.

lucidd (approved):

testsuite/features/reposync/reference_srv_sync_products_extra.feature
```diff
@@ -37,7 +37,7 @@ Feature: Synchronize extra products in the products page of the Setup Wizard
     When I follow the left menu "Admin > Setup Wizard > Products"
     And I wait until I do not see "Loading" text
     And I enter "Ubuntu" as the filtered product description
-    And I select "amd64-deb" in the dropdown list of the architecture filter
+    And I select "amd64-deb" from "product-arch-filter"
```
srbarrios, line 40: Ideally, we better hide this technical detail in the step definition.
That way, if we change this id in the future, we only need to change it in the step definition and not in all the scenarios using it.

testsuite/features/secondary/srv_dist_channel_mapping.feature
```diff
@@ -25,8 +25,8 @@ Feature: Distribution Channel Mapping
     Then I should see a "Create Distribution Channel Map" text
     When I enter "SUSE Linux Enterprise Server 15 SP 4" as "os"
     And I enter "15.5" as "release"
-    And I select "x86_64" from "architecture" dropdown
-    And I select "SLE-Product-SLES15-SP4-Pool for x86_64" from "channel_label" dropdown
+    And I select "x86_64" from "architecture"
+    And I select "SLE-Product-SLES15-SP4-Pool for x86_64" from "channel_label"
```
srbarrios, line 29: It would be better to hide these technical details inside new step definitions too.
And personally, I think that from the user point of view, knowing that we are acting on a dropdown webelement can help.

Etheryte: In that case, do you want to add this to all other tests that use dropdowns? We do the same for dropdowns everywhere else: `And I select "equals" from "matcher"`, `And I select "1-centos7_ssh_minion_key" from "activationKeys"` etc. Currently only the steps with Select2 were different, this change makes it be the same way everywhere. Unless I misunderstand what you mean?

Etheryte: @srbarrios ping.

Etheryte:
@cbbayburt Would you have the time to look at this at some point please?

testsuite/features/secondary/srv_dist_channel_mapping.feature
srbarrios: Hi @Etheryte the Gherkin syntax and Cucumber scenarios have been designed to use natural language to describe the steps a user must follow to achieve something. Ideally, we should write the scenarios as if were on the official documentation. So, even if as you pointed out, other steps don't follow this good practice, let's try to do it properly wherever we can.

testsuite/features/reposync/reference_srv_sync_products_extra.feature
srbarrios: Hi @Etheryte the Gherkin syntax and Cucumber scenarios have been designed to use natural language to describe the steps a user must follow to achieve something. Ideally, we should write the scenarios as if were on the official documentation. So, even if other steps don't follow this good practice, let's try to do it properly wherever we can.

The good news is that's easy to fix, we only need to create a new step definition, where inside it we use this id.

Bischoff: Yes, we should avoid using internals as much as we could. The "feature" part should remain "as the end user sees it", and end users don't see ids when they are in front of the UI, they see texts...

That being said, instead of adding new specialized steps, we might try to use this existing one:
```
When(/^I select "(.*?)" from "([^"]*)" dropdown/) do |selection, label|
```
This one uses a label instead of an id. However, it assumes the label is in "name" attribute, so I am not sure it would work. To be investigated.

testsuite/features/secondary/srv_dist_channel_mapping.feature
Bischoff: Or see my proposal above.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -202,15 +202,6 @@
   end
 end
 
-# select an item from any dropdown
```
Bischoff, line 205: are you sure it's not used anywhere else ?

Bischoff (changes requested):
see comments

testsuite/features/step_definitions/setup_steps.rb
```diff
@@ -73,14 +73,6 @@
   end
 end
 
-When(/^I select "(.*?)" in the dropdown list of the architecture filter$/) do |architecture|
```
Bischoff, line 76: same, i checked it is not used anywhere else, good

Bischoff (approved):
We should not use ids nor labels, but we can do that in a separate PR. LGTM.

Issue #7707, opened 2023-10-18
Add option to customize reboot threshold


stdevel:
# Short description

Currently, the default threshold before rebooting a system is set to 3 minutes. This might be too long for some environments.
It would be nice having the possibility to customize this threshold, e.g. as parameter in `/etc/rhn/rhn.conf` or within the UI.

# Details

When scheduling a system reboot, e.g. using the UI oder the CLI, the system usually waits 3 minutes before rebooting the system. This might be too long for some people.

PR #7734, opened 2023-10-21
Enable setting system reboot delay via rhn.conf

HoussemNasri:
## What does this PR change?

This PR adds a new config entry to `rhn.conf` named `java.reboot_delay` that is used to specify the number of minutes Uyuni has to wait before performing a system reboot.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
No documentation needed: **Description of the new config was added to `rhn_java.conf`. It should be enough as the default 3 minutes reboot delay was not mentioned in Uyuni docs nor in the XML-RPC API.**

- [x] **DONE**

## Test coverage

No tests: already covered

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/7707

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
+++ b/java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
@@ -370,6 +370,11 @@ public class ConfigDefaults {
     public static final String UNIFY_CUSTOM_CHANNEL_MANAGEMENT = "java.unify_custom_channel_management";
 
     /**
+     * Specify the number of minutes to wait before performing a system reboot
+     * */
+    public static final String REBOOT_DELAY = "java.reboot_delay";
+
+    /**
      * Disable SSL redirection
      */
     public static final String NO_SSL = "server.no_ssl";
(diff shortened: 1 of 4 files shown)


java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
```diff
@@ -369,6 +369,11 @@ public class ConfigDefaults {
      */
     public static final String UNIFY_CUSTOM_CHANNEL_MANAGEMENT = "java.unify_custom_channel_management";
 
+    /**
+     * Specify the number of minutes to wait before performing a system reboot
+     * */
+    public static final String MINUTES_TO_WAIT_BEFORE_REBOOT = "minutes_to_wait_before_reboot";
```
mcalmer, line 375: Nobody knows that this option exists. I would suggest to do the following:

- use an easier option name and start it with "java." - e.g. java.reboot_wait_time
- specify the default in java/conf/rhn_java.conf and add a comment which explains this option incl. the time unit (minutes)
- take care that the value is ">= 1" - A value of 0 would cause a direct shutdown which makes it impossible for salt to return the result back. You would often have a failed action.
- check the uyuni documentation if reboot is explained and if the 3 minutes is described. If yes, that would be a good point to explain this option. Otherwise the description in the default config file is enough.

mcalmer (changes requested):

java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
HoussemNasri: > use an easier option name and start it with "java."

Why the java prefix?

mcalmer: The half is convention. The other reason is, that config values use namespaces. Te java config parser works a bit different than the python. In the python config default files you specify only the var name and the namespace is taken of the filename. When you want to overwrite them in rhn.conf, you have to specify the namespace.
To align this, the namespaces are defined in the java config files directly.
All this still comes from the spacewalk times

lucidd (approved):

mcalmer (approved):

java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
HoussemNasri: Ah, I see, thanks!

PR #7761, opened 2023-10-24
A reboot is not needed to deploy a file in /etc on a SLE micro

Bischoff:
## What does this PR change?

According to my tests, the deployment of `/etc/s-mgr/config` works even if we don't reboot afterwards.

This PR does not make stability any better, it just helps speed.

## Links

Ports:
* 4.3: https://github.com/SUSE/spacewalk/pull/22868

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/smoke_tests/smoke_tests.template
+++ b/testsuite/features/build_validation/smoke_tests/smoke_tests.template
@@ -119,7 +119,6 @@ Feature: Smoke tests for <client>
     When I click on "Deploy Files to Selected Systems"
     Then I should see a "1 revision-deploy is being scheduled." text
     And I should see a "0 revision-deploys overridden." text
-    And I reboot the "<client>" if it is a SLE Micro
     And I wait until file "/etc/s-mgr/config" exists on "<client>"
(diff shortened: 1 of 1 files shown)


ktsamis (approved):
If we don't need to do this, let's not :)

srbarrios:
In theory, the reboot is not required but necessary to make a backup point right?

Bischoff:
> In theory, the reboot is triggered automatically by SUMA, maybe we want to check that this is reboot is actually triggered and completed in a programmatic way in our test scenario. Without a reboot, in theory, the changes are not applied.
> 
> ```
> The SLE Micro uses transactional updates. Transactional updates are atomic (all updates are applied only if all updates succeed) and support rollbacks. They do not affect a running system as no changes are activated until after the system is rebooted. This information is displayed in the Systems  Details  Overview subtab.
> ```

From what I understand, the reboot is not automatic. You have to triogger it. What this doc says is that the changes are applied only after the reboot.

You can see it: when you do changes, the SLE micro does not reboot automatically.

nodeg (approved):

srbarrios (approved):

PR #7784, opened 2023-10-27
Have patch available for upgrade in a control scenario

maximenoel8:
## What does this PR change?
Following the example of min_recurring_action, downgrade milkyway-dummy-1.0 on sle minion, switch to fake channel to have milkyway-dummy-2.0 available. Apply the upgrade and switch back the channel.

For the Remove package from Salt minion, the scenario was not removing a package but installing one. Change the scenario to remove a package.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/proxy_as_pod_basic_tests.feature
+++ b/testsuite/features/secondary/proxy_as_pod_basic_tests.feature
@@ -153,7 +153,7 @@ Feature: Register and test a Containerized Proxy
   Scenario: Install a patch on the Salt minion
     When I follow "Software" in the content area
     And I follow "Patches" in the content area
-    When I check the first patch in the list
+    When I check the row with the "milkyway-dummy" text
     And I click on "Apply Patches"
     And I click on "Confirm"
     Then I should see a "1 patch update has been scheduled for" text
@@ -162,10 +162,13 @@ Feature: Register and test a Containerized Proxy
   Scenario: Remove package from Salt minion
     When I follow "Software" in the content area
     And I follow "List / Remove"
-    And I enter the package for "milkyway-dummy-2.0" as the filtered package name
+    And I enter the package for "milkyway-dummy" as the filtered package name
     And I click on the filter button
-    And I check the package for "milkyway-dummy-2.0" in the list
+    And I wait until I see "Clear filter to see all" text
+    And I check the package for "milkyway-dummy" in the list
     And I click on "Remove Packages"
+    Then I wait until I see "Confirm Package Removal" text
+    And I should see a "milkyway-dummy" text
     And I click on "Confirm"
     Then I should see a "1 package removal has been scheduled for" text
     And I wait until event "Package Removal scheduled by admin" is completed
--- a/testsuite/features/secondary/proxy_as_pod_basic_tests.feature
+++ b/testsuite/features/secondary/proxy_as_pod_basic_tests.feature
@@ -161,14 +161,14 @@ Feature: Register and test a Containerized Proxy
 
   Scenario: Remove package from Salt minion
     When I follow "Software" in the content area
-    And I follow "Install"
+    And I follow "List / Remove"
     And I enter the package for "milkyway-dummy-2.0" as the filtered package name
     And I click on the filter button
     And I check the package for "milkyway-dummy-2.0" in the list
-    And I click on "Install Selected Packages"
+    And I click on "Remove Packages"
     And I click on "Confirm"
-    Then I should see a "1 package install has been scheduled for" text
-    And I wait until event "Package Install/Upgrade scheduled by admin" is completed
+    Then I should see a "1 package removal has been scheduled for" text
+    And I wait until event "Package Removal scheduled by admin" is completed
 
   @susemanager
   Scenario: Cleanup: subscribe system back to default base channel
--- a/testsuite/features/secondary/proxy_as_pod_basic_tests.feature
+++ b/testsuite/features/secondary/proxy_as_pod_basic_tests.feature
@@ -121,6 +121,35 @@ Feature: Register and test a Containerized Proxy
     And the uptime for "sle_minion" should be correct
     And I should see several text fields
 
+  Scenario: Pre-requisite: subscribe system to Fake Channel
+    Given I am on the Systems overview page of this "sle_minion"
+    When I follow "Software" in the content area
+    And I follow "Software Channels" in the content area
+    And I wait until I do not see "Loading..." text
+    And I check radio button "Fake-Base-Channel"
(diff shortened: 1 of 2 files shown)


srbarrios (approved):
When the acceptance tests pass, lgtm

testsuite/features/secondary/proxy_as_pod_basic_tests.feature
```diff
@@ -132,14 +161,61 @@ Feature: Register and test a Containerized Proxy
...
     And I click on the filter button
-    And I check the package for "sle_minion" in the list
-    And I click on "Install Selected Packages"
+    And I check the package for "milkyway-dummy-2.0" in the list
+    And I click on "Remove Packages"
     And I click on "Confirm"
-    Then I should see a "1 package install has been scheduled for" text
-    And I wait until event "Package Install/Upgrade scheduled by admin" is completed
+    Then I should see a "1 package removal has been scheduled for" text
+    And I wait until event "Package Removal scheduled by admin" is completed
```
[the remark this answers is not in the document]
maximenoel8: I downgrade it, but upgrade it during the Install a patch on the Salt minion scenario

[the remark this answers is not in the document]
maximenoel8: Now that's you point it out, we may have a risk of flaky test with that. I will verify it on Monday.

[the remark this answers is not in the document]
nodeg: Thank you, Maxime.

maximenoel8:
I add some safe guards around deleting a package. If the package variable is empty, for example if client is not in PACKAGE_BY_CLIENT it will delete the first package in the list. The first package in the list is `aaa_base-84.87+git20180409.04c9dae-150300.10.6.2`. Deleting this package make the minion unusable.

maximenoel8:
I'm also making sure now that I will install milky-dummy and delete milky-dummy.

nodeg:
RuboCop have some complains:

```bash
Inspecting 41 files
.........C...............................

Offenses:

features/step_definitions/navigation_steps.rb:[8](https://github.com/uyuni-project/uyuni/actions/runs/6701201498/job/18208272696?pr=7784#step:5:9)44:[9](https://github.com/uyuni-project/uyuni/actions/runs/6701201498/job/18208272696?pr=7784#step:5:10): C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
  raise "Package name is not set" if input.empty?
        ^^^^^^^^^^^^^^^^^^^^^^^^^
features/step_definitions/navigation_steps.rb:914:9: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
  raise "Package name is not set" if text.empty?
        ^^^^^^^^^^^^^^^^^^^^^^^^^

41 files inspected, 2 offenses detected
```

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -910,6 +911,7 @@
 end
 
 When(/^I check "([^"]*)" in the list$/) do |text|
+  raise 'Package name is not set' if text.empty?
```
srbarrios, line 914:
```suggestion
  raise 'The text to check can't be empty' if text.empty?
```

In this particular step definition, we could be using it to mark different kinds of information, it could be a package, but it can also be a channel for example. So, we will need to have a more generic raise error description.

testsuite/features/secondary/proxy_as_pod_basic_tests.feature
```diff
@@ -121,25 +121,105 @@ Feature: Register and test a Containerized Proxy
...
+
+  Scenario: Pre-requisite: check that there are updates available
+    Given I am on the Systems overview page of this "sle_minion"
+    And I wait until I see "Software Updates Available" text, refreshing the page
+
   Scenario: Install a patch on the Salt minion
     When I follow "Software" in the content area
     And I follow "Patches" in the content area
-    When I check the first patch in the list
+    When I check the row with the "milkyway-dummy" text
```
srbarrios, line 156: Good improvement, indeed.
Even if it will force us to run some other scenarios, this will help to have less flaky tests on the CI test suite. For the BV template, I guess we can keep using the first available patch or package.

testsuite/features/secondary/proxy_as_pod_basic_tests.feature
```diff
@@ -121,25 +121,105 @@ Feature: Register and test a Containerized Proxy
...
+    And I follow "Software Channels" in the content area
+    And I wait until I do not see "Loading..." text
+    And I check radio button "Fake-Base-Channel"
+    And I wait until I do not see "Loading..." text
+    And I check "Fake-Child-Channel"
+    And I click on "Next"
+    Then I should see a "Confirm Software Channel Change" text
+    When I click on "Confirm"
+    Then I should see a "Changing the channels has been scheduled." text
+    And I wait until event "Subscribe channels scheduled by admin" is completed
```
srbarrios, line 136: Instead of subscribing to another base channel. Would make sense to keep the same base channel, and just be sure we have marked the `Fake-RPM-SUSE-Channel` so we can test using the patches and packages of that child channel?
If that can work, we don't need to perform a clean-up scenario where we need to subscribe again to the right base channel and make sure all the child channels are ok.

testsuite/features/secondary/proxy_as_pod_basic_tests.feature
maximenoel8: The issue with this first patch step is you expect to be in a certain setup and we don't have checks or safe guards to verify that we are in the right setup. So we could have fail later related to this selection.

testsuite/features/secondary/proxy_as_pod_basic_tests.feature
maximenoel8: It's a good idea but will required a more complex refactoring. It will probably requires to create the activation key differently. It's not a bad idea to have a channel change scenario. It covers more product features.

srbarrios: Ok, indeed we will cover more that way, it could be harder to maintain for the next person jumping here, but it covers the subscription channel feature.

A separate improvement we can work on in the future is a scenario that passes an operating system, it can wait for the check of all the sub-products in a programmatic way, which will make it easier to maintain the scenarios where we explicitly mention the channel names in the scenario.

PR #7787, opened 2023-10-27
Use the full debian base channel name to avoid Ambiguous match

maximenoel8:
## What does this PR change?
Use full debian base channel name to avoid ambiguous map.
Indeed base channel is matching more than one element in the channel drop box.
We are lucky to have this issue only in this feature. Our fake base channel should never have only Fake-Base-Channel name because it is share by more than one channel. I think we should rename the fake base channel for suse as for rh and debian like channel ( Fake-Base-Channel-Suse-like )

base_channel_list

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/srv_dist_channel_mapping.feature
+++ b/testsuite/features/secondary/srv_dist_channel_mapping.feature
@@ -49,7 +49,7 @@ Feature: Distribution Channel Mapping
     When I enter "Ubuntu 22.04.01 LTS" as "os"
     And I enter "22.04" as "release"
     And I select "x86_64" from "architecture" dropdown
-    And I select "Fake-Base-Channel" from "channel_label" dropdown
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

nodeg (approved):
LGTM. Thank you, Maxime. 
> Our fake base channel should never have only Fake-Base-Channel name because it is share by more than one channel. I think we should rename the fake base channel for suse as for rh and debian like channel ( Fake-Base-Channel-Suse-like )
I fully agree with that.

deneb-alpha:
checked with @nodeg and @srbarrios and  decided to merge

PR #7795, opened 2023-10-31
Chore: Lint reposync.py

m-czernek:
This PR formats the `reposync.py` file as per the Google's pylint config file [0]. The only change to the config file was to use 4 spaces instead of 2.

[0] https://google.github.io/styleguide/pylintrc

Notes
- I did not change public API, since I am not sure whether the methods can be called externally, like by using API (for example, `getChannelRepo` should be `get_channel_repo`). We can absolutely change them if all that I have to worry about is internal callers (i.e. callers from within the uyuni repo).
- I did not want to make non-formatting changes (e.g. moving formats to f-strings). If we'd want to do that, I can do that easily.
  - a notable exception is the `repo` variable import, which I made more specific so that methods do not shadow the `repo` name.
- Some lines were too long, but I could extract variable from the expression and make it shorter. I consider that OK. 
- Some lines were too long but I didn't find an easy way to break them (that wouldn't break readability). In those cases, I simply ignored that one line. In my mind, this is "good enough".
- Some SQL statements were too long. I never modified any SQL statements (whitespace-only) but if I couldn't find an easy way to improve readability, I again disabled the "line-too-long" rule. 
- I have disabled docstrings for this file. The reason is that it was too time consuming to come up with the ~30 docstrings missing in this class. 
- I used a mix of `black` and manual editing to make Pylint happy.

Currently, executing Pylint from our test container returns a 10/10 score:

```
# pylint --rcfile python/pylintrc python/spacewalk/satellite_tools/reposync.py
Using config file /manager/python/pylintrc

------------------------------------
Your code has been rated at 10.00/10
```

## Documentation
TODO, I need to add documentation for how to execute Pylint. 
TODO, I need to remove other pylintrc files from the repo.

## Re-run a test

- [X] Re-run test "backend_unittests_pgsql"
- [X] Re-run test "java_pgsql_tests"
- [X] Re-run test "schema_migration_test_pgsql"
- [X] Re-run test "susemanager_unittests"
- [X] Re-run test "javascript_lint"
- [X] Re-run test "spacecmd_unittests"


python/spacewalk/satellite_tools/reposync.py
```diff
@@ -38,48 +40,58 @@
...
-from spacewalk.common.rhnTB import fetchTraceback
-from spacewalk.common import repo
 from uyuni.common.rhnLib import isSUSE, utc
 from uyuni.common.checksum import getFileChecksum
 from uyuni.common.context_managers import cfg_component
+
+from spacewalk.server import rhnPackage, rhnSQL, rhnChannel, suseEula
+from spacewalk.common import rhnLog, rhnMail, suseLib
+from spacewalk.common.rhnTB import fetchTraceback
+from spacewalk.common.repo import GeneralRepoException
```
m-czernek, line 51: Notable but easy to miss change; we were importing the full `repo` module but it seems that we were using only the exception from it.

I made this more specific because some methods use the `repo` name and shadow this import.

agraul:
First of all, thank you for taking the time to reformat `reposync.py`! This effort shows in what a bad shape the code currently is, a lot of these formatting changes are fixing odd indentation.

Being the lazy developer I am, I ran `black` on top of your changes to see how much work it might be to sustain this format. There are two types of errors reported now: `line-length` and `bad-continuation`. `line-length` is expected, `pylintrc` specifies `80` and black defaults to `88` (https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length). Personally, I like the 90-ish characters that black provides more, but we can configure black and set it's line length to 80.
The other error surprised me, I think the following code format is in line with https://google.github.io/styleguide/pyguide.html#34-indentation:
```
        for s in (m.group(1) for m in re.finditer(regexp, dir_html.decode())):
            if not (
                re.match(r"/", s)
                or re.search(r"\?", s)
                or re.search(r"\.\.", s)
                or re.match(r"[a-zA-Z]+:", s)
                or s.endswith(".rpm")
                or s.endswith(".mirrorlist")
                or s == "#"
            ):
                if re.search(r"/$", s):
                    file_type = "DIR"
                else:
                    file_type = "FILE"

                if s not in self.file_blacklist:
                    self.dir_content.append({"name": s, "type": file_type})
```
But `pylint` disagrees:
```
C:163, 0: Wrong hanging indentation before block (add 4 spaces).
                re.match(r"/", s)
                ^   | (bad-continuation)
C:164, 0: Wrong hanging indentation before block (add 4 spaces).
                or re.search(r"\?", s)
                ^   | (bad-continuation)
C:165, 0: Wrong hanging indentation before block (add 4 spaces).
                or re.search(r"\.\.", s)
                ^   | (bad-continuation)
C:166, 0: Wrong hanging indentation before block (add 4 spaces).
                or re.match(r"[a-zA-Z]+:", s)
                ^   | (bad-continuation)
C:167, 0: Wrong hanging indentation before block (add 4 spaces).
                or s.endswith(".rpm")
                ^   | (bad-continuation)
C:168, 0: Wrong hanging indentation before block (add 4 spaces).
                or s.endswith(".mirrorlist")
                ^   | (bad-continuation)
C:169, 0: Wrong hanging indentation before block (add 4 spaces).
                or s == "#"
                ^   | (bad-continuation)
```
I think we should set up `pylint` together with `black` so that code will be formatted according to our pylintrc automatically. 

Once we agree on what we want, we should discuss how we implement the changes. I'm in favor doing a "format once and for all" approach as explained on https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html and https://lukasz.langa.pl/36380f86-6d28-4a55-962e-91c2c959db7a/

agraul:
Looking at https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#pylint, the indentation problem is probably caused by using such an old pylint version (`1.8.2`). We should consider updating it

m-czernek:
Agreed @agraul ; I think we should accept the line length of black. I also fixed the indentation error by using black and pylint locally instead of the container. This makes it easier to integrate into dev workflow since this way, you can integrate both into IDEs.

Also added some basic documentation.

python/pylintrc
agraul, line 1: I would store the file in the root of the repository since `python/` is not the only directory with Python code. Another "actively maintained" (i.e. we change it regularly) directory is `spacecmd/`, but there are also salt modules in `susemanager-utils/susemanager-sls/**`. Can't we place it in the root of the repository?

python/README.md
```diff
@@ -40,3 +40,41 @@ docker run -t -i --rm=true -e "PYTHONPATH=/manager/python/:/manager/client/rhel/
...
+black /path/to/file.py
+
+# Check pylint
+pylint /path/to/file
+```
+
+To use the `pylintrc` file  from any directory, you can create a symbolic link to `~/.pylintrc`:
+
+```
+ln -s `pwd`/pylintrc ~/.pylintrc
```
agraul, line 79: Assuming we put it in the root of the repo:
```suggestion
ln -s `git rev-parse --show-toplevel`/pylintrc ~/.pylintrc
```

If we copy it to multiple places, we can still take it from `python/` (with CWD being something else):
```
ln -s `git rev-parse --show-toplevel`/python/pylintrc ~/.pylintrc
```

python/README.md
```diff
@@ -40,3 +40,41 @@ docker run -t -i --rm=true -e "PYTHONPATH=/manager/python/:/manager/client/rhel/
 ```
 
 Within the container shell, just run  `pytest /manager/python/test/` to run all tests.
+
+# Linting
+
+Uyuni uses `pylint` for enforcing code style, and `black` for automatic application of some of the style rules.
```
agraul, line 46:
```suggestion
Uyuni uses [`pylint`](https://pylint.pycqa.org/en/latest/index.html) for enforcing code style, and [`black`](https://black.readthedocs.io/en/stable/) for automatic application of some of the style rules.
```

python/requirements-lint.txt
agraul, line 1: I think we should version these to make sure everyone gets the same. Different pylint versions can create different results. We also need to keep in mind that not everyone is using Tumbleweed, so whatever version we specify needs to be available for Leap/Python 3.6 as well.

m-czernek:
Closing in favor of https://github.com/uyuni-project/uyuni/pull/8100

PR #7813, opened 2023-11-02
Fix the problem with the full package name of the proxy helm chart when tagging with tito

avshiliaev:
## What does this PR change?

In the following PR https://github.com/uyuni-project/uyuni/pull/7620 the name of the proxy helm chart was changed from `proxy-helm` to `proxy`. While the change is logical in itself, it creates errors when tagging the package with tito. In general, there're several things relevant for tito when tagging a package:

* its name in the `.spec` file (Chart.yml in case of a helm chart)
* the name of the `.changes` file in the package's dir
* the name on the `rel-eng/pacakges/`
* previous tags for this package

So the solution would be to either change all four parameters to `proxy` instead of `proxy-helm` here in git and only add the suffix when pushing to obs, or revert back to `proxy-helm`.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- Not tested

- [ ] **DONE**

## Links

Fixes #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/containers/proxy-helm/Chart.yaml
+++ b/containers/proxy-helm/Chart.yaml
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: MIT
-#!BuildTag: uyuni/proxy:latest
+#!BuildTag: uyuni/proxy-helm:latest
 apiVersion: v2
-name: proxy
+name: proxy-helm
 description: Uyuni proxy containers.
 type: application
 home: https://www.uyuni-project.org/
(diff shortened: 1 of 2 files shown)


deneb-alpha:
@juliogonzalez the  PR looks  good to me  and it should solve the issue while tagging, but I would like  to  have a second check from you.

deneb-alpha (approved):
LGTM

juliogonzalez (approved):

PR #7819, opened 2023-11-03
Fix build regression in FileUtils.java

sbluhm:
## What does this PR change?

This change adds back a previously removed error catching clause.
Without it, the java build can fail with:
```
    [javac] /builddir/build/BUILD/spacewalk-java-git-69657.b67eaf5/code/src/com/redhat/rhn/common/util/FileUtils.java:216: error: unreported exception IOException; must be caught or declared to be thrown
    [javac]             LineIterator it = org.apache.commons.io.IOUtils.lineIterator(fileStream,
    [javac]                                                                         ^
```

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
Additional build has been tested successfully on copr for
`epel-9-x86_64`
`opensuse-leap-15.5-x86_64`

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/common/util/FileUtils.java
+++ b/java/code/src/com/redhat/rhn/common/util/FileUtils.java
@@ -225,6 +225,10 @@ public class FileUtils {
             log.error("File not found: {}", pathToFile);
             throw new RuntimeException(e);
         }
+        catch (IOException e) {
+            log.error("Could not read from: {}", pathToFile);
+            throw new RuntimeException(e);
+        }
         finally {
             org.apache.commons.io.IOUtils.closeQuietly(fileStream);
         }
--- a/java/spacewalk-java.changes.stefan.fixes-spacewalk-java-4.4.21-1
+++ b/java/spacewalk-java.changes.stefan.fixes-spacewalk-java-4.4.21-1
@@ -1,0 +1,1 @@
+- Fix compile error "unreported IOException" in FileUtils.java.


mackdk (commented):
I understand the compilation error, but the official `commons-io` 2.11.0 code [has no `throws` declaration](https://github.com/apache/commons-io/blob/rel/commons-io-2.11.0/src/main/java/org/apache/commons/io/IOUtils.java#L1635-L1637), so this does not seem to me an issue to fix in Uyuni code. 

As a workaround, I'd rather suggest to rewrite the whole method using modern Java and get rid of the `IOUtils.lineIterator()` dependency:

```java
    public static String getTailOfFile(String pathToFile, Integer lines) {
        CircularFifoBuffer buffer = new CircularFifoBuffer(lines);
        try (BufferedReader reader = new BufferedReader(new FileReader(pathToFile, StandardCharsets.UTF_8))) {
            reader.lines().forEach(line -> buffer.add(line));
        }
        catch (IOException e) {
            log.error("File not found: {}", pathToFile);
            throw new IllegalStateException("File not found " + pathToFile, e);
        }

        // Construct a string from the buffered lines
        StringBuilder sb = new StringBuilder();
        for (Object s : buffer) {
            sb.append(s);
            sb.append(System.lineSeparator());
        }

        return sb.toString();
    }
```

sbluhm:
@mackdk thank you for that feedback. let me see if this is an issue with the version of commons-io that I am using. Will put it to draft for time being.

sbluhm:
So this is what it was. The correct dnf module was not enabled and the build happened with apache-commons-io 2.8 instead of 2.11. OBS already had that module enabled. I will submit a new change though to enforce a minimum build requirement.

Thank you again!

PR #7820, opened 2023-11-03
fix(7594): Skip syncing packages with incorrect metadata (bsc#1213738)

m-czernek:
## What does this PR change?

When an RPM package does not set required metadata headers, reposync throws errors for that repository.

In this change, we enforce that the metadata is set before the repository plugin includes the package in the `list_packages` function for further processing.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: no tests seem necessary (though I can add one for each modified plugin repository if you think it necessary)

## Links

Closes https://github.com/SUSE/spacewalk/issues/22135

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,10 +79,11 @@ reports/
 *.mypy*
 
 # VS Code
-java/.vscode
+.vscode/
 
 # Python
 venv/
+.venv/
 
 # Schema
 
--- a/.gitignore
+++ b/.gitignore
@@ -79,7 +79,8 @@ reports/
 *.mypy*
 
 # VS Code
-.vscode/
+java/.vscode
+python/.vscode
 
 # Python
 venv/
--- a/python/spacewalk/satellite_tools/repo_plugins/__init__.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/__init__.py
@@ -115,6 +115,14 @@ class ContentPackage:
             if not((checksum_type_in in self.checksums) and (self.checksums[checksum_type_in] == checksum_in)):
                 self.checksums[checksum_type_in] = checksum_in
 
+    def get_nvra_tuple(self) -> tuple:
+        """returns a tuple of name, version, release, arch"""
+        return (self.name, self.version, self.release, self.arch)
+
+    def is_metadata_valid(self) -> bool:
+        """Returns True if the package NVREA is set"""
+        return all(self.get_nvra_tuple())
+
     def __str__(self):
         return f'ContentPackage: name = {self.name}, epoch = {self.epoch}, version = {self.version}, release = {self.release}, arch = {self.arch}, checksum_type = {self.checksum_type}, checksum = {self.checksum}, checksums = {self.checksums}, path = {self.path}, a_pkg = {self.a_pkg}, unique_id = <{self.unique_id}>'
 
--- a/python/spacewalk/satellite_tools/repo_plugins/__init__.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/__init__.py
@@ -66,6 +66,12 @@ class ContentPackage:
         return nra
 
     def setNVREA(self, name, version, release, epoch, arch):
+        if not all((name, version, release, arch)):
+            raise ValueError(
+                (
+                    "Incorrect package NVRA values: "
+                    f"N: {name}, V: {version}, R: {release}, A: {arch}"
+                ))
         self.name = name
         self.version = version
         self.release = release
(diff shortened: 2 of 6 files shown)


vzhestkov:
Just not too loose the other related PR will put the link here: https://github.com/uyuni-project/uyuni/pull/7594
So in case of accepting this PR the PR by the link above makes no sense.

python/spacewalk/satellite_tools/repo_plugins/__init__.py
```diff
@@ -115,6 +115,14 @@ def set_checksum(self, checksum_type_in=None, checksum_in=None):
             if not((checksum_type_in in self.checksums) and (self.checksums[checksum_type_in] == checksum_in)):
                 self.checksums[checksum_type_in] = checksum_in
 
+    def get_nvra_tuple(self) -> tuple:
+        """returns a tuple of name, version, release, arch"""
+        return (self.name, self.version, self.release, self.arch)
```
agraul, line 120: A bit above we have `getNRA`, `getNVREA`, and `getNEVRA`. Do we really need `get_nvra_tuple` as well?

The only caller is `is_metadata_valid` which has direct access to instance members, is this level of indirection needed?

python/spacewalk/satellite_tools/repo_plugins/__init__.py
```diff
@@ -115,6 +115,14 @@ def set_checksum(self, checksum_type_in=None, checksum_in=None):
...
+
+    def is_metadata_valid(self) -> bool:
+        """Returns True if the package NVREA is set"""
+        return all(self.get_nvra_tuple())
```
agraul, line 124: The docstring mentions `NVREA` but `self.get_nvra_tuple` does not include the `E (epoch)` component. Which of the two is it :smile:  ?

python/spacewalk/satellite_tools/repo_plugins/yum_dnf_src.py
```diff
@@ -325,6 +325,11 @@ def list_packages(self, filters, latest):
             if new_pack.checksum_type == 'sha':
                 new_pack.checksum_type = 'sha1'
             new_pack.checksum = pack.returnIdSum()[1]
+            if new_pack.is_metadata_valid():
+                to_return.append(new_pack)
+            else:
+                log(0, f"WARNING: package {new_pack.getNVREA()} contains" \
+                    + "incorrect metadata. SKIPPING!")
             to_return.append(new_pack)
```
agraul, line 333: I think this needs to be removed since it was moved to the `if-branch`
```suggestion
```

agraul (commented):
While reading around the area you're modifying I got an alternative idea. I'm not sure if it's better and would like to know what you think.

Adding a package to the `to_return` list of packages in `list_packagest()` starts by initializing a `ContentPackage` and then calling `.setNVREA` on it, e.g. here: https://github.com/uyuni-project/uyuni/blob/fffc2f7ac6aa1bd8ec4f7cfb268722c40c73f679/python/spacewalk/satellite_tools/repo_plugins/yum_dnf_src.py#L319-312

Instead of accepting an NVREA without a release number and then checking it afterwards via `is_metadata_valid()`, what do you think about adding the check in the setter? `setNVREA` could raise a `ValueError` if not all required values are given.

m-czernek (commented):
@agraul I've implemented your approach; the main difference now is that we are forcing the `deb` repo plugin to adhere to the same standard. 

Looking at the database constraints, the same constraints are enforced on debs as well, so this actually seems like a desirable change since deb packages that don't set the values would run into the same issue. 

I tested the change with a deb repository as well.

.gitignore
```diff
@@ -79,10 +79,11 @@ reports/
 *.mypy*
 
 # VS Code
-java/.vscode
+.vscode/
```
agraul, line 82: I just remembered that we have https://github.com/uyuni-project/uyuni/blob/master/.vscode/settings.json. That's something @uyuni-project/frontend added, so I guess we need to exclude it?

agraul (approved):
LGTM! Left a comment about `.vscode` in `.gitignore`, I know this was my suggestion but maybe it's a problem for the .vscode/settings.json? I'm not sure since that file is already in the repo, is git still tracking it?

.gitignore
Etheryte: The frontend part of the codebase intentionally includes that file in the repo, it's necessary for Typescript support etc to work out of the box since our setup is slightly unorthodox due to historical reasons. It would be preferable if we could keep the file as-is. What's the specific conflict in this case, perhaps there's another way around the problem?

agraul: Talking about specific settings, auto formatting enabled has bitten me a few times when I used VSCode. Nothing I couldn't turn off, but maybe we should discuss IDE settings that apply to all languages.

Etheryte: I think that's a fair point, if there's cases where that is unwanted we should turn that off. Are there any other settings that cause you problems? Would [limiting formatting to specific languages](https://stackoverflow.com/questions/44831313/how-to-exclude-file-extensions-and-languages-from-format-on-save-in-vscode) work for you?

PR #7824, opened 2023-11-03
try to download compressed ubuntu usn database and skip errata without packages

mcalmer:
## What does this PR change?

Ubuntu provide the USN database also Bzip2 compressed.
Try to download the compressed first. In case this is not working fallback to raw json file.

In terms of time it does not give a big difference if the network bandwidth is high.
Maybe is makes a difference with less bandwitdth.

Download and decompress in case of bzip compressed file it takes ~ 15 seconds.
Same number when you download the uncompressed file.

The second commit skip all errata which does not have matching package names in the channel we look at.
This seems to bring down the time by 50% on the update channel i tested with.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Two points of https://github.com/SUSE/spacewalk/issues/20604
Ports: https://github.com/SUSE/spacewalk/pull/22994

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
+++ b/java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
@@ -41,6 +41,7 @@ import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
 
+import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
 import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.methods.HttpGet;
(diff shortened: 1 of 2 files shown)


lucidd (approved):

java/code/src/com/redhat/rhn/manager/content/ubuntu/UbuntuErrataManager.java
```diff
@@ -116,22 +117,40 @@ private static Optional<String> archToPackageArchLabel(String arch) {
...
             ) {
                 Map<String, UbuntuErrataInfo> errataInfo = GSON.fromJson(responseReader, ERRATA_INFO_TYPE);
                 LOG.info("download ubuntu errata end");
                 return errataInfo;
             }
         }
         else {
-            throw new IOException("error downloading " + jsonDBUrl + " status code " + statusCode);
+            LOG.info("Failed to get bzip2 DB - try plain DB");
+            httpGet = new HttpGet(jsonDBUrl);
```
admd, line 138: Small nitpick . Some of the code is duplicate in if/else claus. Maybe we can extract that in a separate method.

mcalmer: Don't know a good way to do this without making this code unreadable. Feel free to provide a solution.

PR #7855, opened 2023-11-10
Remove simple-xml dependecy in favor of JAXB

mackdk:
## What does this PR change?

This PR drops the dependency  on simple-xml in favor of standard JAXB.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/manager/audit/ScapManager.java
```diff
@@ -70,6 +67,8 @@
 import java.util.Map;
 import java.util.Optional;
 
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.JAXBException;
```
mcalmer, line 71: Hmm, didn't I just read that this was deprecated again in Java 17 ?
Or is this an external jar?

mcalmer (commented):
standard JAXB? I read a lot that this was dropped in Java17 and you need to use external jar which might need "jakarta.xml.bind" namespace. Please check

java/code/src/com/redhat/rhn/manager/audit/ScapManager.java
mackdk: Yes, standard is the wrong word here, it was more in the sense of base, plain :slightly_smiling_face: . 

But indeed, JAXB was part of Java EE which no longer exists, and now is part of jakarta so when we update the jar we will need to update the package names. I wanted to migrate away from `simple-xml` since it's a 10 years old unmaintened library. We can also investigate other frameworks, but jaxb is currently part of our dependencies so I didn't have to add add anything new:

```
        <dependency org="suse" name="jaxb-runtime" rev="2.3.1" />
        <dependency org="suse" name="jaxb-impl" rev="2.3.1" />
```

mcalmer: Let's hope the best that this provides everything what we need. This is the only path I found in the jar which I know :-)
glassfish-jaxb/jaxb-impl.jar

      35  Defl:N       37  -6% 2024-01-12 06:58 2499855a  META-INF/services/javax.xml.bind.JAXBContext

mcalmer (approved):

admd:
More for an FYI: We decided to use StAX  over JAXB but that's because we had big dataset here https://github.com/uyuni-project/uyuni/issues/7108. Here in this case, we should be fine with JAXB. Ideally, would have been nice to use the same library though.

mackdk:
> Ideally, would have been nice to use the same library though.

Indeed, that would require rewriting all the scap logic since StAX does not map XML to objects. It's totally doable, maybe even better, but requires changing how we process the results. Other drop in alternative might be Jackson but that would need an additional jar we currently don't have (jackson-dataformat-xml)

admd:
> > Ideally, would have been nice to use the same library though.
> 
> Indeed, that would require rewriting all the scap logic since StAX does not map XML to objects. It's totally doable, maybe even better, but requires changing how we process the results. Other drop in alternative might be Jackson but that would need an additional jar we currently don't have (jackson-dataformat-xml)

@mackdk For now, I would say we are fine but I would appreciate if you could create a card to revisit this so we could align it depending on the effort required.


@rjmateus any thoughts on this one?

rjmateus:
I agree with Abid, I think we are fine on moving forward with this one, and creating a card to revisit in the future.
It's great to get rid of simple-xml. Stax is a different implementation, but is the best approach for large data sets, like in the CVE case.
For small datasets jaxb is a much more convenient library, and simpler to use.

PR #7862, opened 2023-11-13
QE: Don't include a repository on the Test Base channel

srbarrios:
## What does this PR change?

In our GH acceptance tests, we are having an issue with the fact that we have duplicated patches on a API response.
This happen because, in this test environment, we are onboarding a Leap Minion that is subscribed to Test Base Channel and to the Fake RPM SUSE channel.
As both channels, base and child channel contain the same repository, we have a duplication and one of our checks becomes ambiguous.

For this reason, on this PR, we will try to empty the Test Base channel. We need to pass the full test suite to see if it can affect somewhere.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


ktsamis (approved):
I'm worried we will have other tests fail but the logic is correct. Let's test it and see

jordimassaguerpla (approved):
This fixes 4 and 5 set of secondary tests! Well done :clap:  . I am suprised this is not breaking any other think, I mean not adding the fake RPM repository :thinking:  but I trust you on that. so :+1:

PR #7866, opened 2023-11-14
QE: Fix log file extraction scenario

nodeg:
## What does this PR change?

This fixes our `Extract the logs from all our clients` scenario.
image
image

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Cucumber tests were edited

- [x] **DONE**

## Links

- Manager 4.3: 
- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -209,11 +209,7 @@ def extract_logs_from_node(node)
   os_family = node.os_family
   node.run('zypper --non-interactive install tar') if os_family =~ /^opensuse/ && !$is_container_provider
   node.run('journalctl > /var/log/messages', check_errors: false)
-  if node.full_hostname.include?('-srv')
-    node.run('salt-call --local grains.items | tee -a /var/log/salt_grains', verbose: true, check_errors: false)
-  else
-    node.run('venv-salt-call --local grains.items | tee -a /var/log/salt_grains', verbose: true, check_errors: false)
-  end
+  node.run('venv-salt-call --local grains.items | tee -a /var/log/salt_grains', verbose: true, check_errors: false) unless $host_by_node[node] == 'server'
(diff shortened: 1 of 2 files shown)


ktsamis (approved):

testsuite/features/support/commonlib.rb
```diff
@@ -209,7 +209,11 @@ def extract_logs_from_node(node)
   os_family = node.os_family
   node.run('zypper --non-interactive install tar') if os_family =~ /^opensuse/ && !$is_container_provider
   node.run('journalctl > /var/log/messages', check_errors: false)
-  node.run('venv-salt-call --local grains.items | tee -a /var/log/salt_grains', verbose: true, check_errors: false)
+  if node.full_hostname.include?('-srv')
+    node.run('salt-call --local grains.items | tee -a /var/log/salt_grains', verbose: true, check_errors: false)
+  else
+    node.run('venv-salt-call --local grains.items | tee -a /var/log/salt_grains', verbose: true, check_errors: false)
+  end
```
srbarrios, line 216: A server in theory shouldn't be a salt minion itself.
So better if we do something like:
```suggestion
  unless $host_by_node[node] == 'server'
    node.run('venv-salt-call --local grains.items | tee -a /var/log/salt_grains', verbose: true, check_errors: false)
  end
```

srbarrios (changes requested):
Thank you for this.
Please consider my suggestion. Thanks.

testsuite/features/support/commonlib.rb
nodeg: Thank you for the suggestion!

srbarrios (approved):

nodeg:
Merging because I tested this in the CI already.

PR #7868, opened 2023-11-14
Fix the evaluation of mgr_install_flavor_check in openSUSE

srbarrios:
## What does this PR change?

The evaluation of the salt state mgr_install_flavor_check was incorrect in our GH acceptance tests, where we deploy in a container an openSUSE 15.5 Salt minion.
This PR aims to fix it by removing the change to lowercase, being more explicit in the evaluation.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Ports:
- Manager-4.3 https://github.com/SUSE/spacewalk/pull/23007

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/packages/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/packages/init.sls
@@ -2,7 +2,7 @@ include:
   - util.syncstates
   - .packages_{{ grains['machine_id'] }}
 
-{%- if grains['os_family'] == 'Suse' and grains['osmajorrelease']|int > 11 and not grains['oscodename'] == 'openSUSE Leap 15.3'%}
+{%- if grains['os_family'] == 'Suse' and grains['osmajorrelease']|int > 11 and not grains['oscodename'] == 'openSUSE Leap 15.3' %}
 mgr_install_products:
   product.all_installed:
     - refresh: True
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/salt/packages/init.sls
```diff
@@ -15,7 +15,7 @@ mgr_install_products:
 {%- endif %}
 {%- endif %}
 
-{%- if grains['os_family'] == 'Suse' and grains['instance_id'] is defined and "opensuse" not in grains['oscodename']|lower%}
+{%- if grains['os_family'] == 'Suse' and grains['instance_id'] is defined and "openSUSE" not in grains['oscodename'] %}
```
mcalmer, line 18: We can give it a try. I hope that the code name will not change in the near future:-)

Alex said, that the "grains['instance_id'] is defined" might not be correct either.
`'instance_id' in grains` maybe together with a "has a value" check would be more correct.
Not sure if you want to include such a change in this as well

mcalmer (approved):

susemanager-utils/susemanager-sls/salt/packages/init.sls
srbarrios: If you give me it as a suggestion code here, I can apply the suggestion. Otherwise, I will let it out, because I'm not sure how to code it.

mcalmer: go with this now.

agraul: It's okay like that since that code does return an undefined object:
> `foo['bar']` works mostly the same with a small difference in sequence:
>
>    check for an item `'bar'` in _foo_. (`foo.__getitem__('bar')`)
>
>    if there is not, check for an attribute called _bar_ on _foo_. (`getattr(foo, 'bar')`)
>
>    if there is not, return an undefined object.

[Source](https://jinja.palletsprojects.com/en/3.0.x/templates/#variables)

srbarrios:
> @jordimassaguerpla Why are the tests not running on this PR? This would be the best test we can get

I will merge it anyway... so we can have the green in the other tests.

PR #7894, opened 2023-11-16
add support for CentOS 9

guillomovitch:
## What does this PR change?

It just allows to define a bootstrap repository for CentOS 9. The product IDs are probably wrong, but we don't use products, so...

Anyway, this kind of stuff should probably be made configurable, in order to allow arbitrary distribution management, instead of being hard-coded.

### DIFF
--- a/susemanager/src/mgr_bootstrap_data.py
+++ b/susemanager/src/mgr_bootstrap_data.py
@@ -1260,6 +1260,10 @@ DATA = {
         'PDID' : [-30, 2362], 'BETAPDID' : [2364], 'PKGLIST' : RES8,
         'DEST' : DOCUMENT_ROOT + '/pub/repositories/centos/8/bootstrap/'
     },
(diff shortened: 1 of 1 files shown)


mcalmer:
There are no centos9 repos defined. So it does not make sense to configure bootstrap repos when you cannot sync the channels.

mcalmer:
To make sense, you need to define the repositories to sync in util/spacewalk-common-channels.ini and you should also test which client tools channels work. As CentOS9 is "Stream" I am not sure if the standard EL9 Client Tools will work.

About the section with the product ids, please remove it. We will not support this in SUMA, so we do not need this section.

guillomovitch:
I pushed a new commit for removing products IDs in [own branch](https://github.com/uyuni-project/uyuni/compare/master...Renater:uyuni:feature/centos-9-bootstrap-support), but this PR didn't update, probably because it is closed.

PR #7900, opened 2023-11-17
Change srv_rename_hostname back to secondary so it does not always run

ktsamis:
## What does this PR change?

Currently it runs even if core fails, which of course the test fails and it takes way too long. At the end of secondary should be fine. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/run_sets/finishing.yml
+++ b/testsuite/run_sets/finishing.yml
@@ -9,7 +9,6 @@
 
 # IMMUTABLE ORDER
 # this feature is destructive for other features, so it is better at the end
-- features/secondary/srv_rename_hostname.feature
 - features/finishing/srv_smdba.feature
 # these features are needed for gathering log/debug info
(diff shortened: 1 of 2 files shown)


nodeg (commented):
I am not sure if this is the right solution, since all the secondary parallelizable stage is running after this disruptive test. So we could see issues in all lf those tests.

ktsamis:
until we make a firm decision on this I won't merge but I would like to move it away from finishing, it takes an hour now because it runs in timeouts when if core fails it should take ~15 minutes

ktsamis:
Since we don't see this happening often and it's a possibly destructive/disruptive change I'm going to close this and we can revisit when it happens and becomes a problem

PR #7925, opened 2023-11-21
QE: Fix ISO mounting step and add SHA256 checksum verification 

NamelessOne91:
## What does this PR change?

[spacewalk #20573](https://github.com/SUSE/spacewalk/issues/20573) Add SHA256 checksum verification when mounting images

[spacewalk #20900](https://github.com/SUSE/spacewalk/issues/20900) Fix mounting ISO step when using a mirror

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Cucumber tests were fixed and edited

- [x] **DONE**

## Links

Ports:

- Manager 4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/build_validation/add_non_MU_repositories/centos7_minion_add_iso.feature
+++ b/testsuite/features/build_validation/add_non_MU_repositories/centos7_minion_add_iso.feature
@@ -5,7 +5,7 @@
 Feature: Add the CentOS 7 distribution custom repositories
 
   Scenario: Download the iso of CentOS 7 DVD and mount it on the server
-    When I mount as "centos-7-iso" the ISO from "http://mirror.chpc.utah.edu/pub/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-2009.iso" in the server
+    When I mount as "centos-7-iso" the ISO from "http://mirror.chpc.utah.edu/pub/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-2009.iso" in the server, validating its checksum
 
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
--- a/testsuite/features/support/file_management.rb
+++ b/testsuite/features/support/file_management.rb
@@ -1,0 +1,53 @@
+# Copyright (c) 2019-2023 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+# Attempts to retrieve the SHA256 checksum for a file inside a given directory or download it
+# from the same domain the file has been downloaded from.
+# Returns the path to the checksum or nil if none has been found/downloaded
+def get_checksum_path(dir, original_file_name, file_url)
+  checksum_file_names = %W[CHECKSUM SHA256SUMS sha256sum.txt #{original_file_name}.CHECKSUM #{original_file_name}.sha256]
+
+  # When using a mirror, the checksum file should be present and in the same directory of the file
+  if $mirror
+    checksum_file_names.each do |name|
+      checksum_path = "#{dir}/#{name}"
+      _output, code = get_target('server').run("test -e #{checksum_path}", check_errors: false)
+      return checksum_path if code.zero?
+    end
+  # Attempt to download the checksum file
+  else
+    base_url = file_url.delete_suffix(original_file_name)
+
+    checksum_file_names.each do |name|
+      checksum_url = base_url + name
+      _output, code = get_target('server').run("cd #{dir} && wget --no-check-certificate #{checksum_url}", check_errors: false, timeout: 1500)
+      return "#{dir}/#{name}" if code.zero?
+    end
+  end
+
+  # No checksum file found or downloaded
+  nil
+end
+
+# Computes the SHA256 checksum for the file at the given path and verifies it against a checksum file.
+# The original file name is used to retrieve the correct checksum entry
+def validate_checksum_with_file(original_file_name, file_path, checksum_path)
+  # Search the checksum file for what should be the only non-comment line containing the original file name
+  checksum_line, _code = get_target('server').run("grep -v '^#' #{checksum_path} | grep '#{original_file_name}'")
+  raise "SHA256 checksum entry for #{original_file_name} not found in #{checksum_path}" unless checksum_line
+
+  # This relies on the fact that SHA256 hashes have a fixed length of 64 hexadecimal characters to extract the checksum
+  # and address any issue related to a checksum file having a different internal format compared to the standard
+  checksum_match = checksum_line.match(/\b([0-9a-fA-F]{64})\b/)
+  raise "SHA256 checksum not found in entry: #{checksum_line}" unless checksum_match
+
+  expected_checksum = checksum_match[1]
+  validate_checksum(file_path, expected_checksum)
+end
+
+# Computes the SHA256 checksum of the file at the given path and returns a boolean representing whether it
+# matches the expected checksum or not
+def validate_checksum(file_path, expected_checksum)
+  file_checksum, _code = get_target('server').run("sha256sum -b #{file_path} | awk '{print $1}'")
+  file_checksum.strip == expected_checksum
+end
--- a/testsuite/features/support/file_management.rb
+++ b/testsuite/features/support/file_management.rb
@@ -1,32 +1,47 @@
-# Copyright (c) 2019-2023 SUSE LLC
+# Copyright (c) 2023 SUSE LLC
 # Licensed under the terms of the MIT license.
+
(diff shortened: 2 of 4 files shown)


NamelessOne91:
ATM still working on fixing the step when using a mirror.

Regarding the SHA256 checksum verification: with @nodeg we have discussed two possible approaches to it:

1) is the current implementation proposed in this PR which relies on defining the expected checksum in the scenario definition itself. This will keep the code pretty straightforward but will require to manually retrieve and maintain the SHA256 checksum values. It also makes the scenario definition pretty long and adds another layer of complexity for non-technical users --> we can also have a map having as key an ISO image's URL  and as values the corresponding expected SHA256 checksum, somewhere, in order to not modify the features files.

2)  retrieve the checksum value during the step's execution, using wget/curl to obtain the checksum file and parsing it to have the expected value. The main issues with this approach are related to checksum files having different names, different format, potentially having different files for different ISOs (...)

I'd like to have your opinion on which approach would work best for you.
Also, if you've got other ideas on how to proceed they're welcome too.

testsuite/features/build_validation/add_non_MU_repositories/centos7_minion_add_iso.feature
```diff
@@ -5,7 +5,7 @@
 Feature: Add the CentOS 7 distribution custom repositories
 
   Scenario: Download the iso of CentOS 7 DVD and mount it on the server
-    When I mount as "centos-7-iso" the ISO from "http://mirror.chpc.utah.edu/pub/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-2009.iso" in the server
+    When I mount as "centos-7-iso" the ISO from "http://mirror.chpc.utah.edu/pub/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-2009.iso" having an expected SHA256 checksum of "e33d7b1ea7a9e2f38c8f693215dd85254c3a4fe446f93f563279715b68d07987" in the server
```
srbarrios, line 8: I think that providing the whole checksum value in the scenario is to technical, and we try to avoid too much technical detail on Gherkin side.

Instead, I would have something like:
`When I mount as "centos-7-iso" the ISO from "http://mirror.chpc.utah.edu/pub/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-2009.iso" in the server, validating its checksum``

And inside the step you do something like you did here:

```
  # Calculate the SHA256 checksum of the downloaded ISO and compare it with the expected one
  iso_checksum = get_target('server').run("sha256sum #{iso_path} | awk '{print $1}'").strip
  raise "SHA256 checksum verification failed" unless downloaded_checksum == checksum
```

But you obtain the value of `checksum` variable from `https://mirror.chpc.utah.edu/pub/centos/7/isos/x86_64/sha256sum.txt` or the `.sha256` file if exists.

nodeg: That would also be my suggestion.

NamelessOne91:
The regEx used to determine the path inside the mirror should now be working for all the images we are currently considering in our mirror.

It will strip the protocol and host/domain part of the URL and substitute it with `/mirror`, keeping everything else equal.

As an example, the mirror's path for `http://mirror.chpc.utah.edu/pub/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-2009.iso ` will be `/mirror/pub/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-2009.iso`.
The path for `http://download.opensuse.org/distribution/leap/15.4/appliances/iso/openSUSE-Leap-15.4-KDE-Live-x86_64-Media.iso` will be `/mirror/distribution/leap/15.4/appliances/iso/openSUSE-Leap-15.4-KDE-Live-x86_64-Media.iso`
So on an so forth.

NamelessOne91:
Regarding the checksum validation.

I'm working to implement it with the proposed (2nd) approach and will probably rebase once I have something ready.

NamelessOne91:
I've added an implementation of SHA256 checksum(s) validation that relies on having the needed files in the mirror or downloading them from the same domain used for the ISO.

Utility functions for this task have been moved to a new Ruby file - `file_management.rb` - which can possibly gather also files-related functions.

testsuite/features/support/file_management.rb
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2019-2023 SUSE LLC
```
vandabarata, line 1: if this is a new file completely, then
```suggestion
# Copyright (c) 2023 SUSE LLC
```

testsuite/features/support/file_management.rb
```diff
@@ -0,0 +1,53 @@
...
+      _output, code = get_target('server').run("test -e #{checksum_path}", check_errors: false)
+      return checksum_path if code.zero?
+    end
+  # Attempt to download the checksum file
+  else
+    base_url = file_url.delete_suffix(original_file_name)
+
+    checksum_file_names.each do |name|
+      checksum_url = base_url + name
+      _output, code = get_target('server').run("cd #{dir} && wget --no-check-certificate #{checksum_url}", check_errors: false, timeout: 1500)
```
vandabarata, line 23: since you have an if else, and both need the server, I would just create a variable at the beginning of the function to get the server as a target and then use that :thinking:

vandabarata:
Only some nitpicks from my side, but great work otherwise :)

testsuite/features/support/file_management.rb
NamelessOne91: Good catches, thanks !

testsuite/features/support/file_management.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    checksum_file_names.each do |name|
+      checksum_url = base_url + name
+      _output, code = server.run("cd #{dir} && wget --no-check-certificate #{checksum_url}", check_errors: false, timeout: 1500)
+      return "#{dir}/#{name}" if code.zero?
+    end
+  end
+
+  # No checksum file found or downloaded
+  nil
```
srbarrios, line 30: I would prefer to raise an exception in that case.

testsuite/features/support/file_management.rb
```diff
@@ -0,0 +1,54 @@
...
+def get_checksum_path(dir, original_file_name, file_url)
+  checksum_file_names = %W[CHECKSUM SHA256SUMS sha256sum.txt #{original_file_name}.CHECKSUM #{original_file_name}.sha256]
+
+  server = get_target('server')
+  # When using a mirror, the checksum file should be present and in the same directory of the file
+  if $mirror
+    checksum_file_names.each do |name|
+      checksum_path = "#{dir}/#{name}"
+      _output, code = server.run("test -e #{checksum_path}", check_errors: false)
+      return checksum_path if code.zero?
```
srbarrios, line 16: In the way this is designed we will be doing calls to the server for each item on the checksum_file_names.
What aboout performing an unique call to the server, using an `ls` for instance, so we retrieve all the files in `dir` and after that, we can process this output to dedice if the server contains the checksum we are looking for?

testsuite/features/support/file_management.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+  # Attempt to download the checksum file
+  else
+    base_url = file_url.delete_suffix(original_file_name)
+
+    checksum_file_names.each do |name|
+      checksum_url = base_url + name
+      _output, code = server.run("cd #{dir} && wget --no-check-certificate #{checksum_url}", check_errors: false, timeout: 1500)
```
srbarrios, line 24: Instead of trying every item in this list, as a call to the server, we will then try to download it, and finally, we process if it's ok or not.
What about, from the controller, we perform HTTP request to the different possible URLs, identifying which of them is the valid URL for this original_file_name.
And after that, we only need to make a unique call to the server to download from the valid URL, and if it fails... that we raise an exception.

srbarrios: Btw, be cautious with the value on the timeout parameter, if for some reason the server call gets stuck, we will be waiting idly for 1500 seconds, and the total time of the test suite will increase a lot.

testsuite/features/support/file_management.rb
NamelessOne91: Good point, that actually makes much more sense and speeds things up a bit.

srbarrios (approved):

NamelessOne91:
Should I squash the commits into 1 before merging or is it fine to rebase keeping the 3 of them ?

PR #7932, opened 2023-11-21
gh actions: fix directories

jordimassaguerpla:
Because of https://github.com/uyuni-project/uyuni/pull/7651, package files are moved from /srv. However, this change is done at the RPM spec file, meaning it won't be available in the gh actions until the new RPM packages for uyuni have been released and published.

Thus, meanwhile, we need to create those dirs by hand.

## Changelogs

- [X] No changelog needed


jordimassaguerpla:
@srbarrios : merging given this fixes the gh actions.

PR #7933, opened 2023-11-21
gh-actions: do not set the download.opensuse.org nor other hosts

jordimassaguerpla:
## What does this PR change?

We needed to set some of the hosts, because with avahi they were not resolvable when using containers. However, in the github actions, we use podman name resolution, so we do not have that problem, and so we do not have to set the hostname resolution.

Let's remove the "--add-host" flags. This was the root cause of why deploy stopped to work, because it had hardcoded the download.opensuse.org to the old ip address.


srbarrios:
Make sense, thank you!

This spotted a flaky tests, @vandabarata could you please check if we have a card in our rrtg board for this flaky reported inside these results?
(The one about 28 vs 27 minutes ago)
It would be ideal to track it as a card.

vandabarata:
@srbarrios I would love but I'm afraid I'm not sure which tests this affected. can you tell me the scenario or feature?

srbarrios:
> @srbarrios I would love but I'm afraid I'm not sure which tests this affected. can you tell me the scenario or feature?

The only feature failing in the GH tests:
image

image

image

PR #7940, opened 2023-11-22
Disable package normalization during installation

mbussolotto:


mbussolotto:
tested and not working, but it looks like a general problem with `install` and `normalize` (when it's used, action is always set has failed)

PR #8034, opened 2023-12-13
preflight.sh: Write SELinux policy if needed

agraul:
## What does this PR change?

SSH Push in the "tunnel" flavor can conflict with SELinux policies shipped by operating systems. The default policies are not aware of the additional port we need for the tunnel.

The preflight.sh script is executed at the beginning of bootstrapping to download the Salt Bundle and allow Salt code execution. To allow for this download, a custom policy is written when SELinux is enabled and the system is managed via a tunnel. The policy is only written and loaded once.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
+++ b/susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
@@ -14,7 +14,12 @@ if ! [ "${SH_NAME}" = "bash" ]; then
 fi
 
 REPO_HOST=$1
-REPO_PORT=$2
+if [[ $2 =~ "^[0-9]$" ]]; then
+   REPO_PORT=$2
+else
+   echo 'Error: $2 (REPO_PORT) must be an integer.' >2
+   exit 254
+fi
 FAIL_ON_ERROR=1
 if [ "$3" = "1" ]; then
     FAIL_ON_ERROR=0
@@ -199,6 +204,20 @@ elif [ "${INSTALLER}" = "apt" ]; then
         CLIENT_REPO_URL="${CLIENT_REPOS_ROOT}/${A_CLIENT_CODE_BASE}/${A_CLIENT_VARIANT_ID}/bootstrap"
     else
         CLIENT_REPO_URL="${CLIENT_REPOS_ROOT}/${A_CLIENT_CODE_BASE}/${A_CLIENT_CODE_MAJOR_VERSION}/${A_CLIENT_CODE_MINOR_VERSION}/bootstrap"
+    fi
+fi
+
+SELINUX_POLICY_FILENAME="salt_ssh_port_forwarding.cil"
+function selinux_policy_loaded {
+    semodule -l | grep -x $SELINUX_POLICY_FILENAME
+}
+
+# Our SSH tunnel uses a custom port and we must configure SELinux to account for it
+if [[ $REPO_HOST == "localhost" ]] && command -v selinuxenabled && selinuxenabled; then
+    if ! selinux_policy_loaded; then
+        echo "(portcon tcp ${REPO_PORT} (system_u object_r ssh_port_t ((s0)(s0))))" >$SELINUX_POLICY_FILENAME
+        if ! semodule -i $SELINUX_POLICY_FILENAME; then
+			exit_with_message_code "Error: Failed to install SELinux policy." 7
     fi
 fi
 
--- a/susemanager-utils/susemanager-sls/susemanager-sls.changes.agraul.ssh-tunnel-add-selinux-policy
+++ b/susemanager-utils/susemanager-sls/susemanager-sls.changes.agraul.ssh-tunnel-add-selinux-policy
@@ -1,0 +1,2 @@
+- Dynamically load an SELinux policy for "Push via SSH tunnel" for SELinux
+  enabled clients. This policy allows communication over a custom SSH port.


susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
```diff
@@ -202,6 +202,37 @@ elif [ "${INSTALLER}" = "apt" ]; then
...
+    fi
+}
+
+SELINUX_POLICY_FILENAME="salt_ssh_port_forwarding.cil"
+function selinux_policy_loaded {
+    semodule -l | grep -x $SELINUX_POLICY_FILENAME
+}
+
+function write_selinux_policy {
+    if ! [[ -f $SELINUX_POLICY_FILENAME ]]; then
```
agraul, line 219: I figured, since we make sure this script is executed by `bash`, it would be sensible to use Bash syntax. I wouldn't mind rewriting this for POSIX shell. The whole script does not really require Bashisms afaict.

susemanager-utils/susemanager-sls/susemanager-sls.changes.agraul.ssh-tunnel-add-selinux-policy
FilippoBonazziSUSE, line 1: I think from a SELinux point of view it would be more precise to say something like:

Dynamically load a SELinux policy module to allow use of a custom SSH port.

susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
```diff
@@ -202,6 +202,37 @@ elif [ "${INSTALLER}" = "apt" ]; then
...
+        echo "(portcon tcp ${REPO_PORT} (system_u object_r ssh_port_t ((s0)(s0))))" >$SELINUX_POLICY_FILENAME
+    fi
+}
+
+function load_selinux_policy {
+    semodule -i $SELINUX_POLICY_FILENAME
+}
+
+# We need a tunnel and selinux might interfere without a custom policy
```
FilippoBonazziSUSE, line 228: Also here this is not precise. I would rather say:
Our SSH tunnel uses a custom port and we must configure SELinux to account for it

FilippoBonazziSUSE (changes requested):
The solution looks good to me from a SELinux point of view. The generated CIL file is correct, minimal and sufficient.

One important observation (as already mentioned in chat): please add an explicit check that the value of `${REPO_PORT}` is a number only. If that variable could be made to contain CIL code, that would end up straight in your file.

Other minor observations and comments inline and in the chat.

susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
agraul: I've included your version, thanks for the suggestion!

susemanager-utils/susemanager-sls/susemanager-sls.changes.agraul.ssh-tunnel-add-selinux-policy
agraul: Thanks for the suggestion. I think it's important to keep the "tunnel" part since that's the cause of the custom port, but the "loading" part is missing as of now.

susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
```diff
@@ -202,6 +202,21 @@ elif [ "${INSTALLER}" = "apt" ]; then
...
+SELINUX_POLICY_FILENAME="salt_ssh_port_forwarding.cil"
+function selinux_policy_loaded {
+    semodule -l | grep -x $SELINUX_POLICY_FILENAME
+}
+
+# Our SSH tunnel uses a custom port and we must configure SELinux to account for it
+if [[ $REPO_HOST == "localhost" ]] && command -v selinuxenabled && selinuxenabled; then
+    if ! selinux_policy_loaded; then
+        if ! [[ -f $SELINUX_POLICY_FILENAME ]] && [[ $REPO_PORT =~ ^[0-9]+$ ]]; then
+            echo "(portcon tcp ${REPO_PORT} (system_u object_r ssh_port_t ((s0)(s0))))" >$SELINUX_POLICY_FILENAME
```
FilippoBonazziSUSE, line 214: I don't think this is the right place to check for a valid port. What happens if the port turns out to be invalid? You never write the policy module and then try to load a non-existing file?

You should place the check as soon as the port value has been set, and hard fail immediately if invalid.

FilippoBonazziSUSE, line 214: Also how many times do you expect to run this script on the same host? I suppose it will be ran only once at install time.

 Is there a value in checking that `$SELINUX_POLICY_FILENAME` does not already exist? Maybe you could unconditionally write the policy module, and _then_ verify that the file exists (and hard fail if it doesn't).

FilippoBonazziSUSE (changes requested):

susemanager-utils/susemanager-sls/salt-ssh/preflight.sh
FilippoBonazziSUSE: Another thought. I assume this script is running as root, and is reading/writing files into root-controlled folders. Specifically, the CIL policy module file will be written where? In the `root` user home?

If this file could be pre-created by another user, you could be tricked into loading an arbitrary policy module. Even if you unconditionally overwrite the file every time, as I suggest in the previous comment, if the file is controlled by another user it could still be modified by winning a race between the time when you write it and when you use it.

FilippoBonazziSUSE: Having looked into it, the CIL policy module file will be written in `$TEMP_DIR`, which has been safely created ahead of time with `mktemp -d -t salt-bundle-XXXXXXXXXX`. This directory is created with permissions u+rwx, and thus its contents will be fresh and root-controlled.

Therefore there is no worry about policy module pre-creation, as I was reasoning in the previous comment. 

However, this also means that the policy file will never exist upon launching the script, since the temp folder will have been created anew. So there's no point in checking if the file already exists. Since I don't see `set -x` in the script, I would rather check that the `semodule -i` call has been successful, and hard fail if not.

agraul: Of course, that check is in a bad place :facepalm:

agraul: Generally, the script will be executed once. It's executed by "Salt SSH" and "Salt SSH" does not blindly execute it multiple times. But there are situations in which it will be executed again, and I wanted to be careful about those.

You raise another good point about the file location. I was not sure if I wanted to leave it in `/root/`, which would allow a) our users to check what we've allowed (but maybe there is a better way to do it, I don't know SELinux well) and b) might help us with debugging. But then I just implemented it for `$TEMP_DIR` and there a file exists check is indeed pointless.

Generally, when `semodule -i` fails, the next function call that tries to use the tunnel will also fail (that's how we got the bug report in the first place). But I agree with adding a check for `semanage -i`'s exit code, that's a better way to do it. 

I'll revisit this tomorrow again, thanks for your review!

agraul: Moved the check to `REPO_PORT` initialization.

agraul: I removed the `-f` check, I think we don't really need to keep the file around necessarily, and added the check for `semanage -i`.

FilippoBonazziSUSE (approved):
LGTM. Thanks for involving us, happy to help

nodeg (approved):

vzhestkov (approved):
In general looks good to me, the only possible issue I see (not so many chances to get into it) if the port will be reconfigured on the server side it seems it will never be updated in the policy on the client side.

PR #8042, opened 2023-12-15
BV: We only need containers module on the POD Proxy VM

srbarrios:
## What does this PR change?

As @ktsamis pointed out here https://github.com/SUSE/spacewalk/pull/23210
We had something fishy with the container module synchronization.
We don't need to wait for this channel and we should not be synchronizing it as it's not a mandatory sub-channel.
We were manually selecting it in some cases, but actually, we will only need it as part of the Pod Proxy VM, to ensure that we have the packages to install Podman.

This proposal will remove the containers module from other cases, and keep it only on the Proxy VM.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Ports:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


testsuite/features/support/constants.rb
```diff
@@ -948,13 +936,23 @@
...
+      sle-module-containers15-sp4-updates-x86_64-proxy-4.3
     ],
   'suma-retail-branch-server-43' =>
     %w[
       sle-product-suse-manager-retail-branch-server-4.3-pool-x86_64
       sle-product-suse-manager-retail-branch-server-4.3-updates-x86_64
       sle-module-suse-manager-retail-branch-server-4.3-pool-x86_64
       sle-module-suse-manager-retail-branch-server-4.3-updates-x86_64
+      sle-module-basesystem15-sp4-pool-x86_64-smrbs-4.3
+      le-module-basesystem15-sp4-updates-x86_64-smrbs-4.3
```
nodeg, line 951:
```suggestion
      sle-module-basesystem15-sp4-updates-x86_64-smrbs-4.3
```

testsuite/features/secondary/allcli_software_channels_dependencies.feature
```diff
@@ -23,11 +23,11 @@ Feature: Channel subscription with recommended or required dependencies
...
     Then I should see the toggler "disabled"
-    And I should see a "SLE-Module-Containers15-SP4-Pool for x86_64" text
-    And I should see the child channel "SLE-Module-Containers15-SP4-Pool for x86_64" "unselected"
+    And I should see a "Desktop Applications Module 15 SP4 x86_64" text
+    And I should see the child channel "Desktop Applications Module 15 SP4 x86_64" "unselected"
     # check the a child channel selection that requires some channel trigger the selection of it
-    When I select the child channel "SLE-Module-Containers15-SP4-Pool for x86_64"
-    Then I should see the child channel "SLE-Module-Containers15-SP4-Pool for x86_64" "selected"
+    When I select the child channel "Desktop Applications Module 15 SP4 x86_64"
+    Then I should see the child channel "Desktop Applications Module 15 SP4 x86_64" "selected"
```
nodeg, line 30: Why do we sync the desktop applications instead?

testsuite/features/support/constants.rb
```diff
@@ -948,13 +936,23 @@
...
+      sle-module-server-applications15-sp4-pool-x86_64-smrbs-4.3
+      sle-module-server-applications15-sp4-updates-x86_64-smrbs-4.3
+      sle-module-suse-manager-proxy-4.3-pool-x86_64-smrbs
+      sle-module-suse-manager-proxy-4.3-updates-x86_64-smrbs
```
nodeg, line 955: I wonder why we missed those in the first place...

testsuite/features/secondary/allcli_software_channels_dependencies.feature
srbarrios: This test only use a channel to play with, this scenario is not about synchronizing anything.
Any previously selected channel would work as example.

testsuite/features/support/constants.rb
srbarrios: Because it's quite hard to make this map.

testsuite/features/support/constants.rb
```diff
@@ -944,13 +940,23 @@
       sle-product-suse-manager-proxy-4.3-updates-x86_64
       sle-module-suse-manager-proxy-4.3-pool-x86_64
       sle-module-suse-manager-proxy-4.3-updates-x86_64
+      sle-module-basesystem15-sp4-pool-x86_64-proxy-4.3
```
ktsamis, line 943: I'm pretty sure we don't need to explicitly mention basesystem here since it will mandatory channel for the product, but let's be explicit even if it's not needed.

ktsamis (approved):

testsuite/features/support/constants.rb
srbarrios: That's the thing!
This map contains all the channels that we need to wait to be synchronized.
Meaning, this map is not about triggering channels, but about waiting for those we expect to be synchronized.

NamelessOne91 (approved):

PR #8059, opened 2023-12-19
Disable failing tests in pr

jordimassaguerpla:
## What does this PR change?

CI: Skip tests that require installing RPMs from openSUSE mirror infrastructure, because they are not reliable. We know that sometimes the metadata and the RPMs from the openSUSE mirrors can be outdated, plus depending on external infra makes the tests fragile. At the PR level, we want to skip these tests because we prefer stability over coverage.

## Changelogs

- [X] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/secondary/allcli_system_group.feature
+++ b/testsuite/features/secondary/allcli_system_group.feature
@@ -1,6 +1,7 @@
 # Copyright (c) 2017-2022 SUSE LLC
 # Licensed under the terms of the MIT license.
 
+@skip_if_github_validation
 @scope_visualization
 Feature: Manage a group of systems
 
--- a/testsuite/features/secondary/srv_handle_config_channels_with_ISS_v2.feature
+++ b/testsuite/features/secondary/srv_handle_config_channels_with_ISS_v2.feature
@@ -1,6 +1,7 @@
 # Copyright (c) 2022 SUSE LLC
 # Licensed under the terms of the MIT license.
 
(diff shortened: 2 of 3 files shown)


testsuite/features/secondary/allcli_system_group.feature
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2017-2022 SUSE LLC
 # Licensed under the terms of the MIT license.
 
+@skip_if_github_validation
```
srbarrios, line 4: In this case, maybe you can just add the tag in `Scenario: Install the locale formula package on the server` as you do with `@skip_if_container_server` ?

jordimassaguerpla: Don't the tests after that scenario depend on having the package installed? It is skipped for the container server because the container server contains that RPM already... Actually, the package is installed in the container, but it is the one from the last uyuni release.

srbarrios (approved):
Approving, but maybe you can do it differently for `testsuite/features/secondary/allcli_system_group.feature`

testsuite/features/secondary/allcli_system_group.feature
srbarrios: But for your purpose.. it should be the same right?
`@skip_if_github_validation` and `@skip_if_container_server` both work with server container image, doesn't them?

PR #8097, opened 2024-01-04
Change org for orphan vendor packages via taskomatic

mcalmer:
## What does this PR change?

When vendor packages were removed from a repo, they become orphan but cannot be removed by an admin as long as they are not belong to an organization.
To solve this, a org change statement was added to reposync.
This had some side effects when a package moved from one repo to another as it become first orphan, change the org and when it appears in the new repo it does not match the org and is downloaded and created again.

This PR move the functionality to taskomatic and it should be executed only once per day.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/22944 
Ports(s): https://github.com/SUSE/spacewalk/pull/23334

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
@@ -19,25 +19,32 @@ import com.redhat.rhn.common.db.datasource.ModeFactory;
 import com.redhat.rhn.common.db.datasource.Row;
 import com.redhat.rhn.common.db.datasource.SelectMode;
 import com.redhat.rhn.common.db.datasource.WriteMode;
+import com.redhat.rhn.domain.org.OrgFactory;
 
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
 
 import java.io.File;
+import java.time.LocalTime;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Cleans up orphaned packages
  *
  */
-
 public class PackageCleanup extends RhnJavaJob {
 
     @Override
     public String getConfigNamespace() {
         return "package_cleanup";
     }
+
+    // this task is executed every 10 minutes. A daily job should be executed between
+    // these two times.
+    private static final LocalTime DAILY_JOB_START = LocalTime.parse("20:49:00");
+    private static final LocalTime DAILY_JOB_END = DAILY_JOB_START.plusMinutes(10);
 
     /**
      * {@inheritDoc}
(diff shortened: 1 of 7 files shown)


lucidd (approved):

java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
```diff
@@ -74,13 +77,29 @@ else if (log.isDebugEnabled()) {
...
         }
         catch (Exception e) {
             log.error(e.getMessage(), e);
             throw new JobExecutionException(e);
         }
     }
 
+    private void changeOrgForOrphanVendorPackages() {
+        LocalDateTime now = LocalDateTime.now();
+        if (now.getHour() == 21 && now.getMinute() >= 49) {
```
m-czernek, line 92: This feels strange to me; maybe consider making this at least a variable or a helper method instead explaining in a comment? I guess looking at this, I don't understand why we're executing this only at a specified hour after a specified minute instead of every time the method is called.

java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
```diff
@@ -74,13 +77,29 @@ else if (log.isDebugEnabled()) {
...
+            // should be executed only at 21:50 every day
+            WriteMode update = ModeFactory.getWriteMode(TaskConstants.MODE_NAME,
+                    TaskConstants.TASK_QUERY_PKGCLEANUP_ORPHAN_VENDOR_PKG_CHANGE_ORG);
+            int updates = update.executeUpdate(Map.of("org_id", OrgFactory.getSatelliteOrg().getId()));
+            if (updates > 0) {
+                log.info("{} orphan vendor packages ready to being removed", updates);
```
m-czernek, line 98: This seems wrong to me; orphaned vendor packages are ready for removal but this code doesn't remove them. What about something like this? (A better way might be to automatically remove the packages if that's what we wanted to do)

 ```suggestion
                 log.info("found {} orphaned vendor packages; remove them manually" updates);
```

**Review by vzhestkov — APPROVED** [review]
I like such python code cleanups :-) :+1:

File: python/spacewalk/satellite_tools/reposync.py [hunk 3]
```diff
@@ -1458,29 +1457,6 @@ def show_packages(self, plug, source_id):
 
             log(0, "    " + pack_status + pack_full_name + pack_size + pack_hash_info)
 
-    def _normalize_orphan_vendor_packages(self):
-        # Sometimes reposync disassociates vendor packages (org_id = 0) from
-        # channels.
-        # These orphans are then hard to work with in spacewalk (nobody has
-        # permissions to view/delete them). We workaround this issue by
-        # assigning such packages to the default organization, so that they can
-        # be deleted using the existing orphan-deleting procedure.
```
m-czernek, line 1467: Feels a shame that we're losing the context, maybe you can add this to the `changeOrgForOrphanVendorPackages` method?

m-czernek (commented):
Leaving a few nit comments, please consider them before moving with the PR

java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
mcalmer: This job run too often. It run every 10 minutes. We want this cleanup only 1 time per day to prevent the problem we have. If we would run it every time, we would just move the problem from python to java code.

Minute 49 is just to be sure that when taskomatic schedule it not at 50, but 51, the job is not skipped.
At the next run we already have hour "22" which would not match. So we can be sure that it only run 1 time per day.

java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
m-czernek: Well, "ready for being removed" is not grammatical. It's either "ready for removal" (probably what you want) or "being removed" (which to my eye is what's causing the confusion since it means the script is doing the removal).

java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
m-czernek: Maybe we could add it to the method comment? 

Ideally, we'd remove random constant numbers (like `hour == 21` && minute >= 49` -> `isDailyExecutionWindow()` though that still leaves the `21` and `49` constants being just that, random constants deep in a class) but probably a comment would be good enough since I'm not sure how to surface the `21` and `49` constants to a more readable place.

java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
mcalmer: I have changed the text and don't talk about remove at all.

java/code/src/com/redhat/rhn/taskomatic/task/PackageCleanup.java
mcalmer: This has changed as well now. Check the new code

m-czernek (approved):
Love it, much better! Esp. the `DAILY_JOB_START/END` variables are really clear now. TYVM!

PR #8106, opened 2024-01-08
Wait for "A refresh of the product data..." is gone

Bischoff:
## What does this PR change?

When you enter the Products page of the Setup Wizard, you can have this:

image

Wait for this text to disappear.

**Note: it's only one of the 7 current errors in 4.3 BV!** (happens at Debian 12)

## Links

Ports:
 * 4.3: SUSE/spacewalk#23346

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
+++ b/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
@@ -1,4 +1,4 @@
-# Copyright 2017-2023 SUSE LLC
+# Copyright 2017-2024 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 Feature: Synchronize products in the products page of the Setup Wizard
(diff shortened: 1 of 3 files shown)


ktsamis (approved):

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
```diff
@@ -13,6 +13,7 @@ Feature: Synchronize products in the products page of the Setup Wizard
   Scenario: Add SUSE Linux Enterprise Server 12 SP4
     When I follow the left menu "Admin > Setup Wizard > Products"
     And I wait until I do not see "Loading" text
+    And I wait until I do not see "currently running" text
```
srbarrios, line 16: Actually I think it works in the reverse order right?
First it shows the "...currently running..." message, and after that it starts loading the tree of products and for that it shows a "Loading" message

ktsamis: that is a possibility, yes, we only currently see the "currently running" message but the order could be reversed

Bischoff: yes, we only see the problem at time of screenshot. But it could be the "Loading" happens after.

I'll invert.

NamelessOne91 (approved):

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
Bischoff: Done.

srbarrios (approved):

PR #8108, opened 2024-01-08
Containers module needed for SLE 15 SP4 build host

Bischoff:
## What does this PR change?

Apparently, we added the containers channels to the list of channels to wait for (in `support/constants.rb`), but in the case of BV forgot to trigger the corresponding synchronization.

This resulted in a wait until timeout for the containers module.

## Links

Ports:
 * 4.3: SUSE/spacewalk#23351

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
+++ b/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
@@ -146,6 +146,8 @@ Feature: Synchronize products in the products page of the Setup Wizard
     When I open the sub-list of the product "Desktop Applications Module 15 SP4 x86_64"
     And I select "Development Tools Module 15 SP4 x86_64" as a product
     Then I should see the "Development Tools Module 15 SP4 x86_64" selected
+    When I select "Containers Module 15 SP4 x86_64" as a product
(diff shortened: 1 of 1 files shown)


srbarrios (approved):
Good catch! Thank you

NamelessOne91 (approved):

ktsamis (changes requested):
I would explicitly add them to the proxy and branch server too. The tags should take care of any strange combination of deployment and if the proxy is alone it will need those as well.

ktsamis:
> I would explicitly add them to the proxy and branch server too. The tags should take care of any strange combination of deployment and if the proxy is alone it will need those as well.

add these as well and then good to go https://github.com/SUSE/spacewalk/pull/23347/commits/9748c9291452b34ce06807f1aed3c9859f86222e

Bischoff:
> I would explicitly add them to the proxy and branch server too.

ok

> The tags should take care of any strange combination of deployment and if the proxy is alone it will need those as well.

Correct

NamelessOne91:
LGTM, thanks Eric

NamelessOne91 (approved):

PR #8110, opened 2024-01-08
QE: handle TimeoutError occurring  during a page reload

NamelessOne91:
## What does this PR change?

Introduce a reusable function to reload pages and handle the case in which a TimeoutError is raised during a page reload, resulting in no screenshot being available for a failed scenario.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- Cucumber tests were edited

- [ ] **DONE**

## Links

Issue(s):  https://github.com/SUSE/spacewalk/issues/22208
Backport Manager 4.3: 

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


maximenoel8:
Thanks for the effort Paolo. It will improve drastically our comfort of life to have this screenshot back !

NamelessOne91:
I don't like the catching and re-throwing of exceptions but ATM it should to the work and avoid introducing further complexity in `repeat_until_timeout`

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -56,32 +56,32 @@
...
 
-    begin
-      accept_prompt do
-        execute_script 'window.location.reload()'
-      end
-    rescue Capybara::ModalNotFound
-      # ignored
+      refresh_page
     end
+  rescue StandardError => e
```
Bischoff, line 65: Probably a stupid question, but why not `rescue TimeoutError => e` ?

Bischoff (commented):
just one noobish question

testsuite/features/step_definitions/navigation_steps.rb
NamelessOne91: Not stupid at all, I was actually asking myself the same question looking at this again after 1 week lol

`repeat_until_timeout` can raise 2 kind of errors and I wanted to intercept both of them, probably.

1) `raise ScriptError, "Giving up after #{attempts} attempts#{detail}" if attempts == retries`
2) `raise TimeoutError, "Timeout after #{timeout} seconds (repeat_until_timeout)#{detail}"`

But I wonder if it makes sense to have a ScriptError and TimeoutError; maybe they can be both labeled as TimeoutError

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -56,32 +56,32 @@
...
-      accept_prompt do
-        execute_script 'window.location.reload()'
-      end
-    rescue Capybara::ModalNotFound
-      # ignored
+      refresh_page
     end
+  rescue ScriptError, TimeoutError => e
+    # an error may be raised while the page still (re)loads, making no screenshot available
+    find('#page-body', wait: 3)
```
srbarrios, line 67: Why do we do this find, if we will anyway raise an error on the next line?

srbarrios, line 67: Maybe instead of waiting here, we should be waiting in the After method, just before we take the screenshot.

NamelessOne91: Right, it's much better indeed to have this kind of stuff in the After method.

Regarding the waits: when  the failure happens in cases were we are not dealing with bootstrapping minions (as when checking the events history), this and condition will probably short circuit due to lazy evaluation.

`click_button('Details') if has_content?('Bootstrap Minions') && has_content?('Details')`

Maybe we can check for the exception which lead to a failure and use it to determine whether to force another wait, or not.
Something like:

`find('#page-body', wait: 3) if scenario.exception.is_a?(TimeoutError) `

Assuming the problem is actually not having enough time for the page to load.

maximenoel8:
Hi @NamelessOne91 , what is the status of this PR. It will help a lot to have it in uyuni and spacewalk to help debugging when we hit the timeout !

NamelessOne91:
> Hi @NamelessOne91 , what is the status of this PR? It will help a lot to have it in uyuni and spacewalk to help debugging when we hit the timeout !

Hi @maximenoel8 
I've focused on fixing other testsuite issues or flaky tests lately, this PR has been on hold for a while.

I'm not sure the issue is just not having enough time to reload the page before a screenshot is taken, see Oscar's comment. 
We can still try to implement an additional wait, in the `After` method, when a scenario failed due to a `TimeoutError` and see if that makes things better.,

srbarrios (approved):
Once Rubocop issue is fixed, LGTM.

PR #8127, opened 2024-01-09
QE: replace a JS call with UI interaction

NamelessOne91:
## What does this PR change?

Closes https://github.com/SUSE/spacewalk/issues/14266

Replace the JS call to `setScheduleTime()` with actual UI interactions.

## GUI diff

No difference

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- A step used in a Cucumber test was edited

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/14266
Ports(s): Manager 4.3 https://github.com/SUSE/spacewalk/pull/23368

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/datepicker_steps.rb
+++ b/testsuite/features/step_definitions/datepicker_steps.rb
@@ -71,8 +71,18 @@ When(/^I pick (\d+) minutes from now as schedule time$/) do |arg1|
 end
 
 When(/^I schedule action to (\d+) minutes from now$/) do |minutes|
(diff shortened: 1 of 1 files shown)


maximenoel8 (approved):
It looks good.
I can see in datepicker steps that our way of describing the actions are really different from navigation for example where we are using actions from commonlib.rb.

But for homogeneity we should probably keep doing that.

srbarrios (approved):

PR #8128, opened 2024-01-10
Move destructive tests in specific rake file.

maximenoel8:
## What does this PR change?
Move samba test and rename test into secondary_destructive.yml rake file ( I was also thinking secondary _finishing.yml ).
Having those steps in finishing is wasting time when we fail on core.
We should not mix check and test.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23373

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/run_sets/secondary_destructive.yml
+++ b/testsuite/run_sets/secondary_destructive.yml
@@ -4,13 +4,10 @@
 # feature files in testsuite/docs/Guidelines.md in "Rules for features" chapter,
 # as well as guidelines about idempotency in "Idempotency" chapter.
 
+## Finishing Secondary features  BEGIN ##
 
-## Post run features BEGIN ##
+# this feature is destructive for other features, so it is better at the end
+- features/secondary/srv_rename_hostname.feature
+- features/finishing/srv_smdba.feature
 
-# IMMUTABLE ORDER
-
-# these features are needed for gathering log/debug info
-- features/finishing/srv_debug.feature
-- features/finishing/allcli_debug.feature
-
-## Post run features END ##
+## Finishing Secondary features END ##
--- a/testsuite/run_sets/secondary_finishing.yml
+++ b/testsuite/run_sets/secondary_finishing.yml
@@ -6,7 +6,7 @@
 
 ## Finishing Secondary features  BEGIN ##
 
-# this feature is destructive for other features, so it is better at the end
+# this feature is disruptive for other features, so it is better at the end
 - features/secondary/srv_rename_hostname.feature
 - features/finishing/srv_smdba.feature
 
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -619,6 +619,7 @@ Then(/^I should see a "([^"]*)" text$/) do |text|
   raise ScriptError, "Text '#{text}' not found" unless check_text_and_catch_request_timeout_popup?(text)
 end
 
+
 Then(/^I should see a "([^"]*)" text or "([^"]*)" text$/) do |text1, text2|
   raise ScriptError, "Text '#{text1}' and '#{text2}' not found" unless check_text_and_catch_request_timeout_popup?(text1, text2: text2)
 end
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -619,7 +619,6 @@ Then(/^I should see a "([^"]*)" text$/) do |text|
   raise ScriptError, "Text '#{text}' not found" unless check_text_and_catch_request_timeout_popup?(text)
 end
 
-
 Then(/^I should see a "([^"]*)" text or "([^"]*)" text$/) do |text1, text2|
   raise ScriptError, "Text '#{text1}' and '#{text2}' not found" unless check_text_and_catch_request_timeout_popup?(text1, text2: text2)
 end
--- a/testsuite/run_sets/finishing.yml
+++ b/testsuite/run_sets/finishing.yml
@@ -8,9 +8,6 @@
 ## Post run features BEGIN ##
 
 # IMMUTABLE ORDER
-# this feature is destructive for other features, so it is better at the end
-- features/secondary/srv_rename_hostname.feature
-- features/finishing/srv_smdba.feature
 # these features are needed for gathering log/debug info
 - features/finishing/srv_debug.feature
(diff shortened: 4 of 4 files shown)


testsuite/run_sets/secondary_destructive.yml
```diff
@@ -4,13 +4,10 @@
 # feature files in testsuite/docs/Guidelines.md in "Rules for features" chapter,
 # as well as guidelines about idempotency in "Idempotency" chapter.
 
+## Finishing Secondary features  BEGIN ##
 
-## Post run features BEGIN ##
+# this feature is destructive for other features, so it is better at the end
```
srbarrios, line 9: "Destructive" looks like a hard word for me...
Maybe we can have two yml:
- finishing.yml
- collect_logs.yml

srbarrios (approved):
Approving, but I proposed an alternative too.

testsuite/run_sets/secondary_destructive.yml
nodeg: I have to admit that I find the word not that hard because the server hostname rename test is indeed very destructive if it fails. Maybe it is not the proper word, but at least now, we have it in its own YAML file where everybody know it could be problematic.

nodeg (approved):
Although the new name might not be ideal, I am still satisfied with the changes.

testsuite/run_sets/secondary_destructive.yml
szachovy: Shouldn't our tests be isolated? The outcome of one test case should not impact the execution or results of others no matter what.

NamelessOne91 (approved):

testsuite/run_sets/secondary_destructive.yml
NamelessOne91: Thinking of a word which carries a similar meaning and can fit, "disruptive" comes to mind.
But, well, "destructive" does make the point pretty clear.

ktsamis (approved):

testsuite/run_sets/secondary_destructive.yml
Bischoff: I agree it's a bit violent - and it's also inexact.

What we want to state is that they are non idempotent. But they do not "destroy" anything :tongue: 

Why do we need to rename? :thinking:

szachovy (changes requested):

testsuite/run_sets/secondary_finishing.yml
```diff
@@ -4,13 +4,10 @@
 # feature files in testsuite/docs/Guidelines.md in "Rules for features" chapter,
 # as well as guidelines about idempotency in "Idempotency" chapter.
 
+## Finishing Secondary features  BEGIN ##
 
-## Post run features BEGIN ##
+# this feature is disruptive for other features, so it is better at the end
```
szachovy, line 9: Shouldn't our tests be isolated? The outcome of one test case should not impact the execution or results of others no matter what.

szachovy (changes requested):

testsuite/run_sets/secondary_destructive.yml
srbarrios: > Shouldn't our tests be isolated? The outcome of one test case should not impact the execution or results of others no matter what.

Yes, that's the theory. We need to have as close idempotent test cases as possible.

BUT, our environment setup is quite heavy, so we can't deploy a clean environment for each test, meaning we need to reuse the server, its database, its packages, and all the systems registered.

What we do, is try to roll back some actions with "Clean-Up" scenarios. But those scenarios can break, or we can fail earlier and not be able to do a proper cleanup, and then all fall in a cascade. This is something we need to live with for now.

Maybe that can improve when we move from VMs to Containers, it's something we can explore in the future.

testsuite/run_sets/secondary_finishing.yml
maximenoel8: The first reason of this change is to not  executed test in our teardown but only have the checks. Currently if core failed, secondary will not be executed but finishing yes. And we have two heavy tests in this finishing

maximenoel8: And putting those tests at the end of secondary is not a good idea and we can't run them in parallel

szachovy: Creating numerous certificates for a specific use case is usually manageable, and potential chaos can be avoided by implementing effective cleaning mechanisms. Although my perspective may be biased as I am going through this codebase for the first time, developing a mocking mechanism for this behavior (which is typical) could address parallelization concerns, especially with a number of cores available. Organizing these prolonged tasks in this manner would be a more efficient approach.

maximenoel8: Again this PR is not a fix on the potential certificate issue we can have if rename server is failing but it's about not executing those two tests during our teardown. We currently have the essential testing failing in our pipelines. It doesn't make sense to run those tests if core is failing. This PR is limited to just that. I don't touch on which order those tests are executed. I just make sure they are not executed if core fail.

maximenoel8: I think your proposition needs several servers running for the same pipeline. It would indeed be an awesome improvement but from my understanding we are a little bit limiting on the infra side.

srbarrios: In my humble opinion, we better keep the scope of this PR limited to fix our current problem.
We can improve it, following this discussion, in a new card appended in our inbox, to address when we have less pressure.

szachovy: Alright, I've identified areas for future improvements. The current process of manually turning off certain elements when a core test fails needs refinement. Ideally, this should be automated through a (piped) pipeline, something like that:
pipelines_graph_dependency_view_links_v13_12

I plan to propose a corresponding PR after reviewing our Jenkins items. That thing exploration may be beneficial further on in terms of containerization.

Closing by now.

szachovy (approved):

testsuite/run_sets/secondary_destructive.yml
szachovy: Ok, https://github.com/uyuni-project/uyuni/pull/8128#discussion_r1448667695

Closing this thread for now, feel free to reopen it.

PR #8146, opened 2024-01-11
Mention the missing sync for Retail in Uyuni

Bischoff:
## What does this PR change?

This PR addresses the reposync for Retail in the Uyuni case.

According to Ondrej, the existing Uyuni channels contain all the Retail packages, so this PRs just adds two comments stating there is nothing to do.

## Links

Ports:
 * 4.3: SUSE/spacewalk#23404

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
+++ b/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
@@ -755,6 +755,8 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I wait until I see "SUSE Manager Retail Branch Server 4.3 x86_64" product has been added
     And I wait until all synchronized channels for "suma-retail-branch-server-43" have finished
 
+# There are no channels for Retail under Uyuni
+
   Scenario: Detect product loading issues from the UI in Build Validation
     Given I am authorized for the "Admin" section
     When I follow the left menu "Admin > Setup Wizard > Products"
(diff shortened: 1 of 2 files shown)


aaannz:
IIUC, this is just for uyuni, right? Then this is not needed. There are no special channels for retail under uyuni. Everything is in the uyuni repo.
Or are you testing SUMA RBS as a client to Uyuni? That seems like a strange scenario to me. Technically ok, but with RBS subscription you get SUMA as well, I think.

Bischoff:
> IIUC, this is just for uyuni, right? Then this is not needed. There are no special channels for retail under uyuni. Everything is in the uyuni repo.

Yes, this is for Uyuni. So there is absolutely nothing special to sync when someone wants Retail in Uyuni ?

If that's the case, i'll just add a comment so that no one else starts wondering the same...


> Or are you testing SUMA RBS as a client to Uyuni? That seems like a strange scenario to me. Technically ok, but with RBS subscription you get SUMA as well, I think.

No, I did not even imagine something so weird :smile: .

Bischoff:
I redid the PR to acknowledge there are no dedicated channels in this case.

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
```diff
@@ -755,6 +755,8 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I wait until I see "SUSE Manager Retail Branch Server 4.3 x86_64" product has been added
     And I wait until all synchronized channels for "suma-retail-branch-server-43" have finished
 
+# There are no channels for Retail under Uyuni
```
srbarrios, line 758: Interesting.

srbarrios (approved):

nodeg (approved):

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
Bischoff: even surprising...

PR #8148, opened 2024-01-11
Add sleep time before rebooting transactional systems after bootstraping

wweellddeerr:
## What does this PR change?

It introduces a delay before rebooting transactional systems post-bootstrap. This addresses cases where the system would reboot before Salt had a chance to send the results of applying the bootstrap state. Such occurrences resulted in error messages appearing in the WebUI, even though the bootstrap operation was actually successful.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: sls files only

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23248

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
@@ -298,7 +298,9 @@ include:
 
 {# Change REBOOT_METHOD to systemd if it is default, otherwise don't change it #}
 
-{%- if not salt['file.file_exists']('/etc/transactional-update.conf') %}
+{%- set etc_conf_file_exists = salt['file.file_exists']('/etc/transactional-update.conf') %}
+
+{%- if not etc_conf_file_exists %}
 copy_conf_file_to_etc:
   file.copy:
     - name: /etc/transactional-update.conf
@@ -313,15 +315,21 @@ transactional_update_set_reboot_method_systemd:
     - separator: '='
     - uncomment: '# '
     - append_if_not_found: True
+{%- if not etc_conf_file_exists %}
     - require:
       - file: copy_conf_file_to_etc
+{%- endif %}
     - unless:
       - grep -P '^(?=[\s]*+[^#])[^#]*(REBOOT_METHOD=(?!auto))' /etc/transactional-update.conf
 
 {# Use transactional reboot support -> server will be rebooted according to REBOOT_METHOD #}
 reboot_transactional_server:
   mgrcompat.module_run:
-    - name: transactional_update.reboot
+    - name: cmd.run
+    - cmd: "sleep 5; transactional-update reboot"
+    - python_shell: true
     - require:
+      - file: transactional_update_set_reboot_method_systemd
       - {{ salt_minion_name }}
+
 {%- endif %}
--- a/susemanager-utils/susemanager-sls/susemanager-sls.changes.welder.bsc1218146
+++ b/susemanager-utils/susemanager-sls/susemanager-sls.changes.welder.bsc1218146
@@ -1,0 +1,1 @@
+- Add sleep time before rebooting transactional systems after bootstraping (bsc#1218146)


aaannz:
I can't say I like this. Can we assume ownership of managed system and just reboot it directly instead of relying on rebootmgr? If so, then we can just initiate delayed reboot like this for example:

```salt
reboot_transactional_server:
  cmd.run_bg:
    - name: sleep 10; reboot
    - require:
      - {{ salt_minion_name }}
```

rjmateus:
but do we safely know how much time to put on sleep?

aaannz:
We do not. It's an educated guess that this should be the last state to be executed during bootstrap. We have these guesses in multiple places:
- saltboot minion stop https://github.com/uyuni-project/retail/blob/master/saltboot-formula/_states/saltboot.py#L1933
- minion cleanup routine https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/cleanup_minion/init.sls#L67
- bootstrap proxy path update https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/bootstrap/set_proxy.sls#L23

rjmateus:
good point and I think is fair enough. upvote from my side

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -313,15 +315,21 @@ transactional_update_set_reboot_method_systemd:
...
     - unless:
       - grep -P '^(?=[\s]*+[^#])[^#]*(REBOOT_METHOD=(?!auto))' /etc/transactional-update.conf
 
 {# Use transactional reboot support -> server will be rebooted according to REBOOT_METHOD #}
 reboot_transactional_server:
   mgrcompat.module_run:
-    - name: transactional_update.reboot
+    - name: cmd.run
+    - cmd: "sleep 5; transactional-update reboot"
+    - python_shell: true
```
aaannz, line 330: Hmm.. does this work as is? I've always used `cmd.run_bg` or `cmd.run` with bg flag true.
With plain `cmd.run` I thought salt is waiting for return and thus state application may not get correct return.
With `cmd.run_bg` salt forks for the command and return event immediately.

vzhestkov: I'm not sure if we can rely on such solution. For me it looks like it will be flacky anyway.
In general I don't like 2 points here:
1. As we discussed already, on bootstrapping the state is applied with `salt-call` which is communicating with `salt-api` through the `stdout` via `ssh`, so `salt-call` just pushes the result of the state apply to the standard output in `json` format. With the reboot we see the case when the `ssh` session is just dropped due to the reboot of the client, so `salt-api` can't get the proper response (with the proper salt minion there are way more chances to get the response as the minion is a distinct service and has some headroom in time to respond on state apply with reboot). For me it's better to implement the solution for such cases on java level. We can just assume that `state.apply` with bootstrap state can just fail with no `json`, so we need to wait some time when the minion will be connected, or the other way is to split the state in 2 parts: move reboot out of bootstrap, but put in the bootstrap state additional part which will include in the response of this state apply if the client is transactional system. And if the client is transactional system, initiate reboot (also with ssh as the minion is not yet running) and it's better to ask the user if it's ok to reboot it now (this is the 2nd point I don't like)
2. It looks like there is some note in the docs which is not read by any customer and we are rebooting system explicitly with no asking the user. I would prefer to ask the user if it's ok to initiate reboot after the bootstrapping, as I described before it can be done with the part of the response on state apply. In any case it will require additional implementation on java side.

srbarrios: Any other possible solution?
We need this as soon as possible in order to have Containerized Proxy integrated in our CI and BV test suites, without this, I can't onboard the HostOS for the Proxy, and so... we are blocked.

aaannz: I might agree with the 1) that this delayed reboot in the state is not the cleanest solution. Would be great to ensure that salt is finished and reboot afterwards. Maybe instead of simple sleep we may fork a process checking for a salt-call to finish and then initiate reboot? We may also use completely different onboarding method for SLE micro like using ignition/combustion/cloud-init like rancher does with elemental.

I can't agree with 2). SLE Micro is not designed to be a pet system. It was designed in mass deployment and static configuration in mind. I don't think adding another question for user if they want to reboot is good usage and the same goes for leaving it in half configured state. I strongly feel we must get to the read state after onboarding and since it currently requires reboot, we must reboot.

agraul: We discussed the "can we reboot now or do we need to delegate" topic a few times already. The last outcome was that we set `REBOOT_METHOD` to `systemd` unless it was changed from its default. The idea was to that we'll always use `transactional_update.reboot` and unless the customer wants to use a different reboot method, it'll be instant.

The shortcoming we weren't aware about is that `systemctl reboot` does not take a parameter to delay the reboot *and* the systemd does not have a salt-minion service to wait for. Another solution might be to create a fleeting systemd service that's active as long as "salt-ssh execution is active".

>  We may also use completely different onboarding method for SLE micro like using ignition/combustion/cloud-init like rancher does with elemental.

Will this also work for already-provisioned SLE Micro systems or only for new deployments?

PR #8153, opened 2024-01-12
QE: refactor verification of a system's subscribed channels

NamelessOne91:
## What does this PR change?

Closes https://github.com/SUSE/spacewalk/issues/22908

Do not evaluate the number of enabled repositories and instead use an API call to retrieve the number of subscribed channels.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Step definitions used by some Cucumber tests were edited

- [x] **DONE**

## Links

Issue(s):  https://github.com/SUSE/spacewalk/issues/22908
Ports(s): Manager 4.3 https://github.com/SUSE/spacewalk/pull/23619

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/api_common.rb
+++ b/testsuite/features/step_definitions/api_common.rb
@@ -1,4 +1,4 @@
-# Copyright (c) 2015-2023 SUSE LLC
+# Copyright (c) 2015-2024 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 ### This file contains the definitions for all steps concerning the API.
@@ -176,6 +176,33 @@ Then(/^channel "([^"]*)" should not have attribute "([^"]*)"$/) do |label, attr|
   ret = $api_test.channel.software.get_details(label)
   assert(ret)
   assert_equal(false, ret.key?(attr))
+end
+
+Then(/^channel "([^"]*)" should be (enabled|disabled) on "([^"]*)"$/) do |channel, state, host|
+  node = get_target(host)
+  system_id = get_system_id(node)
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+
+  assert_equal(state == 'enabled', channels.include?(channel))
+end
+
+Then(/^"(\d+)" channels should be enabled on "([^"]*)"$/) do |count, host|
+  node = get_target(host)
+  system_id = get_system_id(node)
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+
+  assert_equal(count, channels.size)
+end
+
+Then(/^"(\d+)" channels with prefix "([^"]*)" should be enabled on "([^"]*)"$/) do |count, prefix, host|
+  node = get_target(host)
+  system_id = get_system_id(node)
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+
+  assert_equal(count, channels.select { |channel| channel.start_with?(prefix) }.size)
 end
 
 ## activationkey namespace
--- a/testsuite/features/support/namespaces/channel.rb
+++ b/testsuite/features/support/namespaces/channel.rb
@@ -160,7 +160,7 @@ class NamespaceChannelSoftware
   end
 
   ##
-  # Lists the labels of channels the system with the given system ID is subscribed to
+  # Lists the name of channels the system with the given system ID is subscribed to
   #
   # Args:
   #   system_id: The ID of the system.
--- a/testsuite/features/support/namespaces/channel.rb
+++ b/testsuite/features/support/namespaces/channel.rb
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2023 SUSE LLC.
+# Copyright (c) 2022-2024 SUSE LLC.
 # Licensed under the terms of the MIT license.
 
 # Channel namespace
@@ -157,5 +157,15 @@ class NamespaceChannelSoftware
   def list_user_repos
     repos = @test.call('channel.software.listUserRepos', sessionKey: @test.token)
     repos.map { |key| key['label'] }
+  end
+
+  ##
+  # Lists the labels of channels the system with the given system ID is subscribed to
+  #
+  # Args:
+  #   system_id: The ID of the system.
+  def list_system_channels(system_id)
+    channels = @test.call('channel.software.listSystemChannels', sessionKey: @test.token, sid: system_id)
+    channels.map { |channel| channel['name'] }
   end
 end
--- a/testsuite/features/secondary/allcli_software_channels.feature
+++ b/testsuite/features/secondary/allcli_software_channels.feature
@@ -215,7 +215,7 @@ Feature: Channel subscription via SSM
     Then I should see a "Changing the channels has been scheduled." text
     When I follow "scheduled" in the content area
     And I wait until I see "1 system successfully completed this action." text, refreshing the page
-    Then channel "SLE15-SP4-Installer-Updates for x86_64" should not be enabled on "sle_minion"
+    Then channel "SLE15-SP4-Installer-Updates for x86_64" should be disabled on "sle_minion"
 
 @sle_minion
 @uyuni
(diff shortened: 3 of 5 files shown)


NamelessOne91:
A few question on this one

1)  Some step definitions use assert statements, others raise errors. We should probably stick to one standard, which one is it ?

2)  This step definition seems to not be used at all, shall we remove it ?

`Then(/^"(\d+)" channels with prefix "([^"]*)" should be enabled on "([^"]*)"$/) do |count, prefix, host|`

3)  

> B) Instead of relying in a number to evaluate if our state is correct, let's check one by one the channels subscribed to the registered system.

How would you go about this ?
Check that all the starting  CHANNEL_TO_SYNCH_BY_OS_PRODUCT_VERSION are there and then keep track of what is added/removed ?

nodeg:
1. I am not aware of a standard, so we should decide on one.
2. We normally remove steps we do not actively use
3. This constant could be one possible approach. When we then remove channels in one scenario, we exactly know what we remove and only need to verify it.

srbarrios (approved):

testsuite/features/step_definitions/api_common.rb
```diff
@@ -178,6 +178,36 @@
   assert_equal(false, ret.key?(attr))
 end
 
+Then(/^channel "([^"]*)" should (not )?be enabled on "([^"]*)"$/) do |channel, disabled, host|
```
Bischoff, line 181: I'm not a big fan of grouping the enabled and the disabled case under one step.

The assert_equal test that results might be clever, but it's an horror to read and understand.

Besides, there's no performance benefit in factorizing.

testsuite/features/step_definitions/api_common.rb
```diff
@@ -178,6 +178,36 @@
...
+  node = get_target(host)
+  system_id = get_system_id(node)
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+  assert_equal(false, channels.empty?)
```
Bischoff, line 186: are we sure it's an invalid case when this list is empty?

testsuite/features/step_definitions/api_common.rb
```diff
@@ -178,6 +178,36 @@
...
+
+  assert_equal(disabled.nil?, channels.include?(channel))
+end
+
+Then(/^"(\d+)" channels should be enabled on "([^"]*)"$/) do |count, host|
+  node = get_target(host)
+  system_id = get_system_id(node)
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+  assert_equal(false, channels.empty?)
```
Bischoff, line 196: same

testsuite/features/step_definitions/api_common.rb
```diff
@@ -178,6 +178,36 @@
   assert_equal(false, ret.key?(attr))
 end
 
+Then(/^channel "([^"]*)" should (not )?be enabled on "([^"]*)"$/) do |channel, disabled, host|
+  node = get_target(host)
+  system_id = get_system_id(node)
```
Bischoff, line 183: get_target is an ssh (twopence) node

don't we have an API function to get the system id, while we're at using the API ?

testsuite/features/step_definitions/api_common.rb
```diff
@@ -178,6 +178,36 @@
...
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+  assert_equal(false, channels.empty?)
+
+  assert_equal(disabled.nil?, channels.include?(channel))
+end
+
+Then(/^"(\d+)" channels should be enabled on "([^"]*)"$/) do |count, host|
+  node = get_target(host)
+  system_id = get_system_id(node)
```
Bischoff, line 193: same, it sounds to me like we're mixing ssh and api

testsuite/features/step_definitions/api_common.rb
```diff
@@ -178,6 +178,36 @@
...
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+  assert_equal(false, channels.empty?)
+
+  assert_equal(count, channels.size)
+end
+
+Then(/^"(\d+)" channels with prefix "([^"]*)" should be enabled on "([^"]*)"$/) do |count, prefix, host|
+  node = get_target(host)
+  system_id = get_system_id(node)
```
Bischoff, line 203: same

testsuite/features/step_definitions/api_common.rb
```diff
@@ -178,6 +178,36 @@
...
+
+  channels = $api_test.channel.software.list_system_channels(system_id)
+  assert_equal(false, channels.empty?)
```
Bischoff, line 206: same

testsuite/features/support/namespaces/channel.rb
```diff
@@ -158,4 +158,14 @@ def list_user_repos
     repos = @test.call('channel.software.listUserRepos', sessionKey: @test.token)
     repos.map { |key| key['label'] }
   end
+
+  ##
+  # Lists the name of channels the system with the given system ID is subscribed to
+  #
+  # Args:
+  #   system_id: The ID of the system.
+  def list_system_channels(system_id)
```
Bischoff, line 167: thanks for the addition

Bischoff (commented):
some doubts

please apologize if unjustified

testsuite/features/step_definitions/api_common.rb
srbarrios: In opposite direction, I like it to use regex of that kind in our step definitions.
I found it easier to maintain aligned the code for the same kind of action.

NamelessOne91: It's mostly to avoid having nearly identical code blocks with just a reversed check.
Beside that, there's no particular improvement as you say, so I have no strong preference on reverting it back if that's what we agree on.

testsuite/features/step_definitions/api_common.rb
srbarrios: I don't really understood the question, but keep in mind this method is defined as:
```
def get_system_id(node)
  $api_test.system.search_by_name(node.full_hostname).first['id']
end
```

testsuite/features/step_definitions/api_common.rb
srbarrios: Sure we are mixing.
From one side, we want to obtain the twopence node, and from the other side we need the system id of that node.
We could be storing the id inside the Lavanda class, but that would be another refactor, and we need to be sure to keep this id value up to date in all moment (when we delete and recreate systems)

testsuite/features/step_definitions/api_common.rb
NamelessOne91: Good point. I'll look into that and see if we can refactor.

testsuite/features/step_definitions/api_common.rb
srbarrios: I like it, if I would complain in something about it is in the way the sentence is made, we have other examples with this kind of format:

```
When(/^I (enable|disable) repositories (before|after) installing branch server$/) do |action, when|
```

testsuite/features/step_definitions/api_common.rb
NamelessOne91: Looking also at the other comment about API calls, I think we're fine with the current implementation and can refactor this part later on if needed.
This PR is really more about handling the checks on channels .

testsuite/features/step_definitions/api_common.rb
NamelessOne91: uhm, I'd say this check may even be a problem if we want to check for a disabled channel on a minion which has no channels at all.

Probably it is worth to remove it from all the these step definitions, as it also redundant in the other cases.
`channels.includes` should return false anyway if the list is empty and checking a count against the channels.size will always work.

NamelessOne91:
Refactored a bit & rebased

PR #8162, opened 2024-01-15
CI: fix run script

jordimassaguerpla:
## What does this PR change?

Some scripts were missing in the run script, so running the tests locally was not producing the same results as running them in github actions.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation
- No documentation needed

- [ ] **DONE**

## Test coverage
- No tests

- [ ] **DONE**

## Links

fixes https://github.com/SUSE/spacewalk/issues/23398

- [ ] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/podman_runner/run
+++ b/testsuite/podman_runner/run
@@ -12,16 +12,22 @@ set -e
 ./03_run_controller.sh
 ./04_setup_ssh_controller.sh
 ./05_install_gems_in_controller.sh
-./06_start_server.sh
-./07_manager_setup.sh
-./08_build_server_code.sh
-./09_run_sshminion.sh
-./10_setup_sshd.sh
-./11_run_core_tests.sh
-./12_run_salt_sle_minion.sh
-./13_run_salt_rhlike_minion.sh
-./14_run_salt_deblike_minion.sh
-./15_accept_all_keys.sh
-./16_run_init_clients_tests.sh
-./17_run_secondary_tests.sh
-
+./06_collect_and_tag_flaky_tests_in_controller.sh
+./07_start_server.sh
+./08_manager_setup.sh
+./09_build_server_code.sh
+./10_run_sshminion.sh
+./11_setup_sshd.sh
+./12_run_core_tests.sh
+./13_run_salt_sle_minion.sh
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

nodeg (approved):

testsuite/podman_runner/run
```diff
@@ -12,16 +12,22 @@ set -e
...
+./15_run_salt_deblike_minion.sh
+./16_accept_all_keys.sh
+./17_run_init_clients_tests.sh
+./18_run_secondary_tests.sh
+./19_split_secondary_p_tests.sh
+for i in 1 2 3 4 5;do
+  ./19_run_secondary_parallelizable_tests_subset.sh ${i}
+done
+./20_get_server_logs.sh
+./21_get_client_logs.sh
```
szachovy, line 33: It is not supposed to be a run file executing shell scripts as a modules non-directly?

szachovy (changes requested):

testsuite/podman_runner/run
srbarrios: I don't understand the question, could you please give us an extended description of what do you mean and what you propose?

szachovy: Sure.
When altering the sequence of tasks, or opting to separate cases, renaming files becomes necessary. What's worse is that it can lead to inconsistency if we attempt to use something from there somewhere else.

Perhaps a more effective approach would be to create packages containing functions from all these files. Then, in the run script, you can selectively choose which packages to execute.

For example:

`run.sh`:
```
#!/bin/bash
export UYUNI_PROJECT=uyuni-project
export UYUNI_VERSION=master
set -ex

setup_environment ...
run_environment ...
get_logs ...
...
```
Each individual file, such as `setup_environment.sh`, would contain bash **function**s responsible for creating the environment. These files could be derived from some base co-routines to ensure that everything displays properly.

szachovy: Ok, I will just reference it here as an issue TODO https://github.com/SUSE/spacewalk/issues/23427

PR #8210, opened 2024-01-24
QE: Force FQDN check to run locally on the host

nodeg:
# What does this PR change?
There are issues when using only rum() where the command could potentially be executed inside a container instead of the host. With this change, we force the execution on the host.

`hotname -f` on our 5.0 does not return anything useful according to Cedric.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Cucumber tests were edited

- [x] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!](<## What does this PR change?

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -13,7 +13,7 @@ require 'date'
 
 Then(/^"([^"]*)" should have a FQDN$/) do |host|
   node = get_target(host)
-  result, return_code = node.run('date +%s; hostname -f; date +%s', check_errors: false)
+  result, return_code = node.run_local('date +%s; hostname -f; date +%s', check_errors: false)
   lines = result.split("\n")
   initial_time = lines[0]
   result = lines[1]


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -13,7 +13,7 @@
 
 Then(/^"([^"]*)" should have a FQDN$/) do |host|
   node = get_target(host)
-  result, return_code = node.run('date +%s; hostname -f; date +%s', check_errors: false)
+  result, return_code = node.run_local('date +%s; hostname -f; date +%s', check_errors: false)
```
vandabarata, line 16: shouldn't this be an if with this command for a local server and an alternative for a container? :thinking:  I assume the container also needs a hostname. unless I'm wrong and then disregard everything I just said

nodeg: We discussed that with Cedric. 
```ruby
  def run(cmd, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, user: 'root', successcodes: [0], buffer_size: 65_536, verbose: false)
    cmd_prefixed = @in_has_mgrctl ? "mgrctl exec -i '#{cmd.gsub(/'/, '\'"\'"\'')}'" : cmd
    run_local(cmd_prefixed, separated_results: separated_results, check_errors: check_errors, timeout: timeout, user: user, successcodes: successcodes, buffer_size: buffer_size, verbose: verbose)
  end
```
`run()` is a wrapper for `run_local()`. The container has a static hostname, but not a proper FQDN. So the suggestion from Cedric was to adjust it this way. We could also remove the FQDN check that we have in our build validation code.
```cucumber
Then "server" should have a FQDN
```
https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/build_validation/core/allcli_sanity.feature#L8-L9
Cedric did the same for the CI 10 months ago.

Bischoff: i prefer if we keep the check and do the present change.

Bischoff (approved):

testsuite/features/step_definitions/command_steps.rb
nodeg: Cedric and me, too. We can always remove the test later on if this does not work. But it would be good to have this check for the host OS.

cbosdo (approved):

vandabarata (approved):

PR #8227, opened 2024-01-27
improve update systems via recurring states

mcalmer:
## What does this PR change?

Implement some improvements to update systems via recurring states.

- provide a standalone salt update state
- improve update stack update state
- check pillar to skip reboot_if_needed if not wanted

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Not needed, but added https://github.com/SUSE/spacewalk/issues/23573

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/22483
Ports(s): https://github.com/SUSE/spacewalk/pull/23577

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/uptodate.sls
+++ b/susemanager-utils/susemanager-sls/salt/uptodate.sls
@@ -1,6 +1,14 @@
 include:
   - channels
 
+{%- if grains['os_family'] == 'Suse' %}
+keep_system_up2date_updatestack:
+  pkg.uptodate:
+    - onlyif: 'zypper patch-check --updatestack-only; r=$?; test $r -eq 100 || test $r -eq 101'
+    - require:
+      - sls: channels
+{% else %}
+
 mgr_keep_system_up2date_updatestack:
   pkg.latest:
     - pkgs:
@@ -23,6 +31,7 @@ mgr_keep_system_up2date_updatestack:
 {%- endif %}
     - require:
       - sls: channels
+{%- endif %}
 
 mgr_keep_system_up2date_pkgs:
   pkg.uptodate:
@@ -31,6 +40,7 @@ mgr_keep_system_up2date_pkgs:
       - sls: channels
       - pkg: mgr_keep_system_up2date_updatestack
 
+{%- if salt['pillar.get']('mgr_reboot_if_needed', True) %}
 mgr_reboot_if_needed:
   cmd.run:
     - name: shutdown -r +5
@@ -46,4 +56,5 @@ mgr_reboot_if_needed:
       - test -e /boot/do_purge_kernels
 {%- else %}
     - onlyif: 'zypper ps -s; [ $? -eq 102 ]'
+{%- endif %}
 {%- endif %}
--- a/susemanager-utils/susemanager-sls/salt/uptodate.sls
+++ b/susemanager-utils/susemanager-sls/salt/uptodate.sls
@@ -40,7 +40,7 @@ mgr_keep_system_up2date_pkgs:
       - sls: channels
       - pkg: mgr_keep_system_up2date_updatestack
 
-{%- if salt['pillar.get']('mgr_reboot_if_needed', True) %}
+{%- if salt['pillar.get']('mgr_reboot_if_needed', True) and salt['pillar.get']('custom_info:mgr_reboot_if_needed', 'true')|lower in ('true', '1', 'yes', 't') %}
 mgr_reboot_if_needed:
   cmd.run:
     - name: shutdown -r +5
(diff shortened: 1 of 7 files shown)


rjmateus (approved):
Lgtm. thank you for implementing this one.

cbbayburt (commented):
How is the `mgr_reboot_if_needed` pillar set?

mcalmer:
> How is the `mgr_reboot_if_needed` pillar set?

I think there is no automatic way yet. So by default nothing change compared to what we have now. reboot is always executed if needed.

But a customer can manual set it via `/srv/pillar/` or maybe also via "custom info" (AFAIK we convert them into pillar? - I havn't tried this yet.)

cbbayburt (approved):
Thanks. LGTM :+1: 

Related issue just for reference: https://github.com/SUSE/spacewalk/issues/22483

mcalmer:
When using custom info, it would look like this:
```
    custom_info:
        ----------
        mgr_reboot_if_needed:
            False
```
And I think "False" is here a string. I think we can enhance the check to also react on this.

mcalmer:
I am also looking for some hints if the UI configutation is complete? AFAIK there is somewhere a decription which I cannot identify where it comes from.

cbbayburt (changes requested):
> I am also looking for some hints if the UI configuration is complete? AFAIK there is somewhere a description which I cannot identify where it comes from.

Oh right, descriptions are translated strings in [`java/StringResource_en_US.xml`](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/frontend/strings/java/StringResource_en_US.xml#L9309-L9311).

The UI picks it up automatically, looking at the message key `internal_state.<state_label>`

So you need to add the following:

```xml
      <trans-unit id="internal_state.update-salt">
        <source>...</source>
      </trans-unit>
```

susemanager-utils/susemanager-sls/salt/uptodate.sls
```diff
@@ -31,6 +40,7 @@ mgr_keep_system_up2date_pkgs:
       - sls: channels
       - pkg: mgr_keep_system_up2date_updatestack
 
+{%- if salt['pillar.get']('mgr_reboot_if_needed', True) %}
```
parlt91, line 43: Just for understanding. The default value of an unset pillar is `True`? Couldn't find anything on it in the docs

mcalmer: Yes, the `True` is the default value if this pillar does not exist
https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.pillar.html#salt.modules.pillar.get

mcalmer:
@parlt91 @cbbayburt I added the translation and I have added handling of pillar created via custom_info.
Please check again

mcalmer:
@admd should this be ported to 4.3 as well? If @parlt91 port the one shot execution for recurring states, this may make sense as well.

susemanager-utils/susemanager-sls/salt/uptodate.sls
parlt91: Now I understand. The `True` is the default not the expected value. Thank you

parlt91 (approved):
Tested it out and it looks good to me. Just left one comment about the schema upgrade dir above.

schema/spacewalk/upgrade/susemanager-schema-5.0.3-to-susemanager-schema-5.0.4/100-add-internal-state-update-salt.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO suseInternalState (id, name, label)
```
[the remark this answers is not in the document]
mcalmer: yesterday this dir was correct - today maybe not :-)

cbbayburt:
> @admd should this be ported to 4.3 as well? If @parlt91 port the one shot execution for recurring states, this may make sense as well.

IMO yes. https://github.com/SUSE/spacewalk/issues/22483 is also going to be implemented on top of this.

cbbayburt (approved):

schema/spacewalk/upgrade/susemanager-schema-5.0.3-to-susemanager-schema-5.0.4/100-add-internal-state-update-salt.sql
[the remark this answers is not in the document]
mcalmer: moved

admd:
Generally, we plan not to add anything major in 4.3 anymore.

But in this case, this would be a nice change to add in 4.3. It's not that big from code wise so less chance of introducing any regressions.

rjmateus:
@admd @mcalmer @cbbayburt I was discussion this with MC on the chat. COnsidering that update salt with salt is the a root of problems, and MC (thank you so much) added this state to address the issue, would make sense to keep updating salt on the `mgr_keep_system_up2date_updatestack` state? [1]
If the user uses this new state before update, we would not have have any problems. However, if salt gets updated, it restarts the agent and forgets about the remaining tasks.

https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/uptodate.sls#L1-L26

PR #8229, opened 2024-01-29
fix: Revert py2.7 incompatible formatting

m-czernek:
## What does this PR change?

This PR reverts linting and formatting, which turns out to be incompatible with py2.7. This is because we still support running the `spacecmd` code on py2.7.


meaksh (approved):

deneb-alpha (approved):
LGTM, thanks

PR #8281, opened 2024-02-08
Opt to iterate through the list items for the 'leap15.5-x86_64' Uyuni product, avoiding manual input

szachovy:
## What does this PR change?

Iterating over the constraint list instead of manually inputting its items in certain sections.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23395
Port(s): https://github.com/SUSE/spacewalk/pull/23662

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


nodeg:
Some comment on the formalia of a PR:

- please use the template as is and do not remove the checkboxes. Example:
  image
- Inside the link section, please leave the checkbox unchecked, until you prepared the back port to spacewalk and link the PR there, too. Example:
  image
- It would make sense to create a pre-commit hook for RuboCop or run RuboCop with the correct version locally, before pushing/creating a PR.
- assign yourself to your PRs
- not strictly needed but welcome: add the PR to the QE project on the right side (Projects)
- tip for the future: prepare the back port to spacewalk only when your Uyuni PR was reviewed and e.g. merged. Then you can cherry-pick your commit(s) to spacewalk which saves some time.

nodeg:
[RuboCop](https://github.com/uyuni-project/uyuni/actions/runs/7830457647/job/21364531715?pr=8281) complains about:

```bash
cd testsuite
  rubocop features/*
  shell: /usr/bin/bash -e {0}
Inspecting 4[3](https://github.com/uyuni-project/uyuni/actions/runs/7830457647/job/21364531715?pr=8281#step:5:3) files
..W........................................

Offenses:

features/step_definitions/command_steps.rb:185:83: W: Lint/ShadowingOuterLocalVariable: Shadowing outer local variable - channel.
    CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION[product][product_version_type].each do |channel|
                                                                                  ^^^^^^^
features/step_definitions/command_steps.rb:187:100: C: [Correctable] Style/StringLiteralsInInterpolation: Prefer single-quoted strings inside interpolations.
      command = "spacewalk-common-channels -u admin -p admin -a #{arch} #{channel.gsub("-#{arch}", "")}"
                                                                                                   ^^
features/step_definitions/command_steps.rb:188:[4](https://github.com/uyuni-project/uyuni/actions/runs/7830457647/job/21364531715?pr=8281#step:5:4)3: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
      $command_output, _code = get_target("server").run(command)
                                          ^^^^^^^^

43 files inspected, 3 offenses detected, 2 offenses auto-correctable
```

Bischoff (changes requested):
Good first start :smile_cat:  .

Please see the comments.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,19 @@
   $command_output, _code = get_target('server').run(command)
 end
 
+When(/^I use spacewalk-common-channel to add all "([^"]*)" channels with arch "([^"]*)"$/) do |channel, arch|
+  product_version_type = "#{channel}-#{arch}"
```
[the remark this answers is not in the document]
szachovy: The current format is easier to read as it separates the channel from the architecture, which can be helpful for those who might have difficulty differentiating between the two. However, I see the advantage of merging it to accommodate examples like "fake", so I am going to amend the changes.

testsuite/features/support/constants.rb
```diff
@@ -1141,6 +1141,8 @@
         opensuse_leap15_5-x86_64-backports-updates
         opensuse_leap15_5-x86_64-sle-updates
         opensuse_leap15_5-uyuni-client-x86_64
+        opensuse_leap15_5-uyuni-client-devel-x86_64
```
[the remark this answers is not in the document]
szachovy: Yes, I manually removed these two channels by doing:
```
spacewalk-remove-channel -c uyuni-proxy-stable-leap-155-x86_64
spacewalk-remove-channel -c opensuse_leap15_5-uyuni-client-devel-x86_64
```
Then when I selected only these two channels to be added, it took `0m2.686s` without any other constraints. I don't know if extending timeout is needed in that case if we have some space left here.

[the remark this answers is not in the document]
szachovy: I also noticed highly repetitive fragments in `constants.rb`. Issue created https://github.com/SUSE/spacewalk/issues/23658

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,19 @@
...
+  if CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, product_version_type)
+    CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION[product][product_version_type].each do |channel|
+      log "Adding channel: #{channel}"
+      command = "spacewalk-common-channels -u admin -p admin -a #{arch} #{channel.gsub("-#{arch}", "")}"
+      $command_output, _code = get_target("server").run(command)
+    end
+  else
+    log "Version type #{product_version_type} in #{product} product not found, nothing to be synchronized, continuing"
```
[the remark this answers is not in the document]
szachovy: I agree if this channels won't be synchronized, it may cause some issues in the later stages. Now it raises exception.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,19 @@
   $command_output, _code = get_target('server').run(command)
 end
 
+When(/^I use spacewalk-common-channel to add all "([^"]*)" channels with arch "([^"]*)"$/) do |channel, arch|
+  product_version_type = "#{channel}-#{arch}"
+  if CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, product_version_type)
```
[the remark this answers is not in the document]
szachovy: I think in this scenario I just run [this](https://github.com/uyuni-project/uyuni/blob/bc0e6411067410d632aa87ca0523642ba867394b/testsuite/features/support/commonlib.rb#L24) non-parametrized function that way.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,19 @@
...
+    CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION[product][product_version_type].each do |channel|
+      log "Adding channel: #{channel}"
+      command = "spacewalk-common-channels -u admin -p admin -a #{arch} #{channel.gsub("-#{arch}", "")}"
+      $command_output, _code = get_target("server").run(command)
```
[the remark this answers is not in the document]
szachovy: Right, I do not keep the output now.

nodeg (approved):
I like what I see so far. Nice work, Wiktor, and thank you for your first contribution.
Before merging (by you or releng), please squash all commits into one, either locally with git, or here in GitHub via the webUI.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,17 @@
   $command_output, _code = get_target('server').run(command)
 end
 
+When(/^I use spacewalk-common-channel to add all "([^"]*)" channels$/) do |os_product_version|
+  raise ScriptError, "Synchronization error, version type #{os_product_version} in #{product} product not found" unless CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, os_product_version)
```
NamelessOne91, line 183: Just a nitpick.

You can probably save the return of 

`CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, os_product_version)`

to a variable and use it both in this guard clause and in the following `each` loop

NamelessOne91 (approved):
Nitpick aside, LGTM.

Thanks Wiktor

testsuite/features/step_definitions/command_steps.rb
szachovy: Good catch! It improved code readability.
Done

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,18 @@
   $command_output, _code = get_target('server').run(command)
 end
 
+When(/^I use spacewalk-common-channel to add all "([^"]*)" channels$/) do |os_product_version|
+  channels_to_synchronize = CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, os_product_version)
+  raise ScriptError, "Synchronization error, version type #{os_product_version} in #{product} product not found" unless channels_to_synchronize
```
Bischoff, line 184: According to this thread,

https://softwareengineering.stackexchange.com/questions/302530/ruby-if-variable-vs-if-variable-nil
```
  if channels_to_synchronize.nil?
```
is a better coding style than 
```
  unless channels_to_synchronize
```

I'm not sure how well accepted this is. It kind of makes sense to me because it uses `.nil?` explicitely.

It might also be a matter of tastes...

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,18 @@
...
+
+  channels_to_synchronize.each do |os_product_version_channel|
+    log "Adding channel: #{os_product_version_channel}"
+    architecture = 'x86_64' # Only x86_64 is supported due to https://github.com/SUSE/spacewalk/issues/23663
```
Bischoff, line 188: Ouch. I missed you need to pass them separated to the command. My bad.

I have been influenced by the SCC part where you pass them together, as a whole. Please accept my apologies.

However this is not true that you always is x86_64 . For example, as you noticed, it's `amd64-deb` for Debian or somthing like that.

I'm not sure about the solution. You could revert to your previous solution where you concatenate. Or you could introduce new arrays.

First solution is not valid for all "product names", like "fake", but it does not happen in your use case.

Second solution is overengineered.

I guess you initial solution was better, and you need to concatenate again. Sorry.

testsuite/features/support/constants.rb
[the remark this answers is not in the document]
Bischoff: No, when I said timing the duration, I meant using the timestamps in `/var/log/rhn/reposync/<channel>.log`.

It works in two stages: spacewalk-remove-channel only triggers the reposync, and takes only a few milliseconds. Then taskomatic schedules the reposync, which can happer later, and literally take hours.

Bischoff (changes requested):
Sorry, my bad, your initial version was better. See comments.

testsuite/features/support/constants.rb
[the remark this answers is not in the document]
szachovy: Okay, I think I understand what you mean.
- Analizing `/var/log/rhn/reposync/uyuni-proxy-stable-leap-155-x86_64.log`:

  From first: `2024/02/08 10:18:37 +02:00 Command: ['/usr/bin/spacewalk-repo-sync', '--channel', 'uyuni-proxy-stable-leap-155-x86_64', '--type', 'yum', '--non-interactive']`

  To the first line of: `2024/02/08 10:18:46 +02:00 Sync completed.`

  It took around 9 seconds.

- Analizing `/var/log/rhn/reposync/opensuse_leap15_5-uyuni-client-devel-x86_64.log`

  From first: `2024/02/08 10:17:43 +02:00 Command: ['/usr/bin/spacewalk-repo-sync', '--channel', 'opensuse_leap15_5-uyuni-client-devel-x86_64', '--type', 'yum', '--non-interactive']`

  To the first line of: `2024/02/08 10:17:46 +02:00 Sync completed.`

  It took around 3 seconds.

testsuite/features/step_definitions/command_steps.rb
szachovy: Yes, it is correct. I am not expecting `false` values so having it that way is better.
Done.

testsuite/features/step_definitions/command_steps.rb
szachovy: No worries, I reverted this fragments to have the variable values explicitly defined by the user. I also created https://bugzilla.suse.com/show_bug.cgi?id=1219776 with regard to the discussion.

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,17 @@
   $command_output, _code = get_target('server').run(command)
 end
 
+When(/^I use spacewalk-common-channel to add all "([^"]*)" channels with arch "([^"]*)"$/) do |channel, architecture|
+  channels_to_synchronize = CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, "#{channel}-#{architecture}")
```
Bischoff, line 183: what does `.dig` bring with respect to `[][]` ?

maybe we should be a bit conservative here and stay aligned with the other similar uses of  `CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION` which use the sequence of square brackets

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,17 @@
...
+  raise ScriptError, "Synchronization error, version type #{channel}-#{architecture} in #{product} product not found" if channels_to_synchronize.nil?
```
Bischoff, line 184: speaking of aligning code between implementations, this guard clause is missing from the other similar step definitions :disappointed: , if you have time feel free to add it there

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -179,6 +179,17 @@
...
+
+  channels_to_synchronize.each do |os_product_version_channel|
+    log "Adding channel: #{os_product_version_channel}"
+    command = "spacewalk-common-channels -u admin -p admin -a #{architecture} #{os_product_version_channel.gsub("-#{architecture}", '')}"
```
Bischoff, line 188: `delete_suffix()` might be a bit more legible than `gsub()`

testsuite/features/support/constants.rb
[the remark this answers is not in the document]
Bischoff: Thank you, it's the information I needed. Could you please add `60` seconds for both entries in the TIMEOUTS array please? Admitting it's not already there, of course. We round to next minute up.

Bischoff (commented):
last nitpicks and suggestions and we will be good

testsuite/features/step_definitions/command_steps.rb
szachovy: Taking the example:
```
        opensuse_leap15_5-x86_64
        opensuse_leap15_5-x86_64-non-oss
        opensuse_leap15_5-x86_64-non-oss-updates
        opensuse_leap15_5-x86_64-updates
```
the `x86_64` is not always in the end of given channel, so `delete_suffix` method is not applicable here.

testsuite/features/step_definitions/command_steps.rb
szachovy: The `dig` method is useful for safely navigating through nested hashes without raising an error if any intermediate key is `nil`. It returns `nil` if any key in the chain is not found.

This way we exactly know what the type of `channels_to_synchronize` would be (a `list` or `nil`), what makes it more predictable in further stages if we do not have a confidence that the keys exist at each level of our dictionary.

testsuite/features/step_definitions/command_steps.rb
szachovy: I have found two additional occurrences, one I adjusted to the relevant `next if` statement, and the second now raises exception if a given path does not exist.

testsuite/features/step_definitions/command_steps.rb
Bischoff: there's not much risk at first level ("SUSE Manager" or "Uyuni").

but ok

testsuite/features/step_definitions/command_steps.rb
Bischoff: ah! ok

testsuite/features/support/constants.rb
```diff
@@ -1242,6 +1244,7 @@
   'el9-manager-tools-updates-x86_64-ol9' => 60,
   'el9-manager-tools-updates-x86_64-rocky' => 60,
   'el9-pool-x86_64' => 60,
+  'opensuse_leap15_5-uyuni-client-devel-x86_64' => 60,
```
Bischoff, line 1247: thank you very much

Bischoff (approved):
thanks for your patience

testsuite/features/step_definitions/command_steps.rb
Bischoff: thanks

PR #8286, opened 2024-02-08
Coco attestation java

mcalmer:
## What does this PR change?

Implement java backend for Confidential Compute Attestation.
- create and update configuration
- schedule a attestation action
- handling the report and result objects
- implement the state to gather the data from the system
- parse the result and insert it into the database
- API for get/set the configuration and schedule the action

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Part of the big doc task

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23551

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
@@ -36,6 +36,7 @@ import com.redhat.rhn.common.validator.ValidatorResult;
 import com.redhat.rhn.domain.action.Action;
 import com.redhat.rhn.domain.action.ActionFactory;
 import com.redhat.rhn.domain.action.ActionType;
+import com.redhat.rhn.domain.action.CoCoAttestationAction;
 import com.redhat.rhn.domain.action.salt.ApplyStatesActionDetails;
 import com.redhat.rhn.domain.action.salt.ApplyStatesActionResult;
 import com.redhat.rhn.domain.action.salt.StateResult;
@@ -116,6 +117,7 @@ import com.redhat.rhn.frontend.dto.SystemOverview;
 import com.redhat.rhn.frontend.dto.VirtualSystemOverview;
 import com.redhat.rhn.frontend.events.SsmDeleteServersEvent;
 import com.redhat.rhn.frontend.xmlrpc.BaseHandler;
+import com.redhat.rhn.frontend.xmlrpc.EntityNotExistsFaultException;
 import com.redhat.rhn.frontend.xmlrpc.InvalidActionTypeException;
 import com.redhat.rhn.frontend.xmlrpc.InvalidChannelException;
 import com.redhat.rhn.frontend.xmlrpc.InvalidChannelLabelException;
@@ -189,6 +191,10 @@ import com.redhat.rhn.taskomatic.TaskomaticApi;
 import com.suse.cloud.CloudPaygManager;
 import com.suse.manager.api.ApiIgnore;
 import com.suse.manager.api.ReadOnly;
+import com.suse.manager.attestation.AttestationDisabledException;
+import com.suse.manager.attestation.AttestationManager;
+import com.suse.manager.model.attestation.CoCoEnvironmentType;
+import com.suse.manager.model.attestation.ServerCoCoAttestationConfig;
 import com.suse.manager.virtualization.VirtualizationActionHelper;
 import com.suse.manager.webui.controllers.virtualization.gson.VirtualGuestSetterActionJson;
 import com.suse.manager.webui.controllers.virtualization.gson.VirtualGuestsBaseActionJson;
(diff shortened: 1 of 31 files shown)


jordimassaguerpla:
Hi ! We fixed a spelling mistake in the name of the tests and you will have to rebase your PR for the correct tests to run. Sorry for the inconvenience.

mackdk (commented):
Looks mostly good to me. I added some comments. 

By the way, I see you defined explicit values for enums instead of relying on `ordinal()`. While it's kind of duplication, I would say it's more future proof to not depend on the declaration order.

Moreover, in the past we used `@Type` instead of `@Convert` for enums. We had already defined a `CustomEnumType` to handle that in hibernate, which supports any possible mapping in the database. I personally don't have any strong preference on that, a part that you can use a type also from the xml mappings. Maybe we should use `@Convert` consistently also for the `BillingDimension` which is currently the only enum to use `@Type`.

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1711,6 +1715,53 @@ private static String packageToKey(Map.Entry<String, Pkg.Info> entry) {
...
+            }
+            else {
+                msg += getJsonResultWithPrettyPrint(jsonResult);
+            }
+            serverAction.setResultMsg(msg);
+            if (report.getResults().isEmpty()) {
+                // if we have results, we leave this up to the attestation container
+                // to evaluate the status and set it to failed if something important is missing
+                report.setStatus(CoCoAttestationStatus.FAILED);
+            }
```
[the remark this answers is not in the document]
mcalmer: Right. I will try to phrase this different to make it more clear

java/code/src/com/suse/manager/model/attestation/CoCoAttestationStatus.java
```diff
@@ -0,0 +1,47 @@
...
+ * in this software or its documentation.
+ */
+package com.suse.manager.model.attestation;
+
+public enum CoCoAttestationStatus {
+    PENDING("PENDING"),
+    SUCCEEDED("SUCCEEDED"),
+    FAILED("FAILED");
+
+    private final String label;
```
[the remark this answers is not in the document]
mcalmer: Right. Initially I had other values. But now it does not make sense.

java/code/src/com/suse/manager/webui/utils/salt/custom/CoCoAttestationRequestData.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Gets all info returned as Map for inserting into the report entry
+     * @return returns data as Map.
+     */
+    public Map<String, Object> asMap() {
+        Map<String, Object> out = new HashMap<>();
+        getSnpguestReport()
+                .map(StateApplyResult::getChanges)
+                .ifPresent(c -> {
+                    if (c.getRetcode() == 0) {
```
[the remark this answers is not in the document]
mcalmer: This highly depends on how the container will implement the the attestation.
The value for the key "mgr_snpguest_report" should always be the report and nothing else. Otherwise you have a hard time to find out if you got an error message or a report.
We would need to define a new key for an error message.
We can do it if this is useful. Maybe we do it with a later PR when we know that the container use it somehow.

mcalmer:
> Moreover, in the past we used `@Type` instead of `@Convert` for enums. We had already defined a `CustomEnumType` to handle that in hibernate, which supports any possible mapping in the database. I personally don't have any strong preference on that, a part that you can use a type also from the xml mappings. Maybe we should use `@Convert` consistently also for the `BillingDimension` which is currently the only enum to use `@Type`.

Funny, I was not aware of this type, but I found out that I implemented the first use of it `AdvisoryStatus.java` and `AdvisoryStatusEnumType.java` :-)

I also think we should use one part only. I am also not sure what is better. 

@lucidd do you have a preference? `@Type`  or `@Convert` ?

mcalmer:
> > Funny, I was not aware of this type, but I found out that I implemented the first use of it `AdvisoryStatus.java` and `AdvisoryStatusEnumType.java` :-)
> > I also think we should use one part only. I am also not sure what is better.
> > @lucidd do you have a preference? `@Type` or `@Convert` ?
> 
> No strong preference here either. I guess the biggest difference is the `UserType` is hibernate specific while `@Convert` is not. We might find some limitations when we use it and can think about it again then.

I found the same when I checked this. I think we should go with `@Convert` and I will create a separate PR to change the 2 existing implementations to this.

java/code/src/com/suse/manager/model/attestation/CoCoAttestationStatus.java
```diff
@@ -0,0 +1,49 @@
...
+     *
+    public static CoCoAttestationStatus lookupByLabel(String label) {
+        for (CoCoAttestationStatus value : values()) {
+            if (value.label.equals(label)) {
+                return value;
+            }
+        }
+        throw new IllegalArgumentException("Unsupported label: " + label);
+    }
+    */
```
mackdk, line 48: I guess we can remove this now.

mackdk (approved):

java/code/src/com/suse/manager/attestation/AttestationManager.java
```diff
@@ -0,0 +1,181 @@
...
+    private static final Logger LOG = LogManager.getLogger(AttestationManager.class);
+    private final AttestationFactory factory;
+    private TaskomaticApi taskomaticApi;
+
+    /**
+     * Constructor
+     */
+    public AttestationManager() {
+        factory = new AttestationFactory();
+        taskomaticApi = new TaskomaticApi();
```
lucidd, line 59: we should take these in as parameters and configure a shared global instance like we do for others already. This way we can also get rid of the `setTaskomaticApi` and make it final.

java/code/src/com/suse/manager/attestation/AttestationManager.java
```diff
@@ -0,0 +1,181 @@
...
+            throw new PermissionException(msg);
+        }
+    }
+
+    private void ensureSystemConfigured(Server serverIn) {
+        if (serverIn == null) {
+            LOG.error("Server not found");
+            throw new LookupException("Server not found");
+        }
+        Optional<ServerCoCoAttestationConfig> cnf = serverIn.getOptCocoAttestationConfig();
```
lucidd, line 170: `orElseThrow` can be used here to avoid having to call `get` later.

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -5158,6 +5164,96 @@ public Integer scheduleScriptRun(User loggedInUser, String label, Integer sid, S
...
+     * @apidoc.param #session_key()
+     * @apidoc.param #param_desc("int", "sid", "ID of the server to get the config for.")
+     * @apidoc.returntype $CoCoAttestationConfigSerializer
+     */
+    @ReadOnly
+    public ServerCoCoAttestationConfig getCoCoAttestationConfig(User loggedInUser, Integer sid) {
+        MinionServer minionServer = SystemManager.lookupByIdAndUser(sid.longValue(), loggedInUser).asMinionServer()
+                .orElseThrow(NoSuchSystemException::new);
+        return minionServer.getOptCocoAttestationConfig()
+                .orElse(new ServerCoCoAttestationConfig());
```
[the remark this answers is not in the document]
mcalmer: `new ServerCoCoAttestationConfig()` return just the default config. I think this is ok.

lucidd (approved):

java/code/src/com/suse/manager/model/attestation/CoCoAttestationStatus.java
mcalmer: right

java/code/src/com/suse/manager/attestation/AttestationManager.java
mcalmer: Not sure if we need to add this to the GlobalInstanceHolder.
We do not have TaskomaticApi there yet.

I changed this now. Please check if this is better now

PR #8353, opened 2024-02-21
Fix lists of channels to sync - Uyuni

Bischoff:
## What does this PR change?

This PR fixes constants `CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION` and `TIMEOUT_BY_CHANNEL_NAME` based on experimental data from a test suite run on the uyuni BV test suite.

Note: in `srv_sync_all-products` we must use these arrays in all cases, not only for SLES. @szachovy will take care of that, see SUSE/spacewalk#23395.

## Links

Issue(s): SUSE/spacewalk#23572
Port(s):
 * 4.3: SUSE/spacewalk#23772

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/constants.rb
+++ b/testsuite/features/support/constants.rb
@@ -634,21 +634,23 @@ PKGARCH_BY_CLIENT = {
   'sle15sp5s390_ssh_minion' => 's390x'
 }.freeze
 
+# Explanations:
+# - beta channels were removed because they are not selected and not currently synced, add them again when we will use them
+# - SLED channels for SUMA tools were removed as we are not currently synchronizing them
+# - 'default' is required for auto-installation tests.
+# - '# CHECKED' means that we verified that the list of channels matches the results in /var/log/rhn/reposync,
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

NamelessOne91 (approved):

Bischoff:
> I mean, we should always verify what we are doing, so distinguishing between `CHECKED` and not `CHECKED` may suggest that something somewhere might have been added blindly. I would remove all `# CHECKED` comments (optional).

The ones without # CHECKED have indeed been added blindly, and are very likely to be wrong, indeed. I don't think we can remove until we checked all of them.

szachovy:
> > I mean, we should always verify what we are doing, so distinguishing between `CHECKED` and not `CHECKED` may suggest that something somewhere might have been added blindly. I would remove all `# CHECKED` comments (optional).
> 
> The ones without # CHECKED have indeed been added blindly, and are very likely to be wrong, indeed. I don't think we can remove until we checked all of them.

Ok

PR #8371, opened 2024-02-27
Bump Node to 20

Etheryte:
## What does this PR change?

See title.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/web/spacewalk-web.changes.eth.node20
+++ b/web/spacewalk-web.changes.eth.node20
@@ -1,0 +1,1 @@
+- Update build tooling
--- a/.github/workflows/automatic-frontend-lint-fix-pr.yml
+++ b/.github/workflows/automatic-frontend-lint-fix-pr.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [20.x]
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/frontend-dependency-audit.yml
+++ b/.github/workflows/frontend-dependency-audit.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [20.x]
 
     steps:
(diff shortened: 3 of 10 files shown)


cbbayburt (commented):
Don't we also need to specify the node version in the specfile?

Etheryte:
@cbbayburt We're currently having that discussion [on Slack](https://suse.slack.com/archives/C02D78LLS04/p1709207014953059). We use `nodejs-default` which should give us the current LTS Node version, but it's unclear whether we can pass it a specific version directly. If you have any ideas your input would be appreciated.

cbbayburt:
> @cbbayburt We're currently having that discussion [on Slack](https://suse.slack.com/archives/C02D78LLS04/p1709207014953059). We use `nodejs-default` which should give us the current LTS Node version, but it's unclear whether we can pass it a specific version directly. If you have any ideas your input would be appreciated.

Well, dependabot doesn't care about LTS at all :see_no_evil: but I think it's important to have the same version here and in the specfile.

Etheryte:
I think I wasn't maybe the clearest, right now we don't specify a version in the specfile at all, see https://github.com/uyuni-project/uyuni/blob/master/web/spacewalk-web.spec#L53, and I'm not even sure if we can specify a version using this package.

web/spacewalk-web.changes.eth.node20
```diff
@@ -0,0 +1 @@
+- Update build tooling
```
deneb-alpha, line 1:
```suggestion
- Build with at least Node.js 20
```

deneb-alpha (approved):
From what I saw, switching to a direct buildrequires to nodejs >=20 fixes the issue of having a different nodejs default version between Uyuni and SUMA. 

we should be good :)

cbbayburt (approved):

PR #8382, opened 2024-02-29
Make mgr_events engine non-blocking on reading events

vzhestkov:
## What does this PR change?

While investigating root causes of high memory utilisation by salt EventPublisher was found that memory consumption is growing when the subscribers are not reading the events from the stream fast enough.

One of the subscirbers is `mgr_events` salt engine, it's made `async` and in most cases not causing blocking, but in case of having large amount of events in the publisher this module could cause additional delays on performing `COMMIT` to the DB.

This change is intended to move DB operations to the distinct thread to make it non-blocking for all the cases.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: already covered, high workload tests not really possible to perform in CI

## Links

Tracks: https://github.com/SUSE/spacewalk/issues/23526

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
+++ b/susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
@@ -5,7 +5,7 @@ LISTEN/NOTIFY mechanism to alert SUSE Manager of newly available events.
 
 mgr_events.py tries to keep the I/O low in high load scenarios. Therefore
 events are INSERTed with the separate thread with no blocking the event bus
-amd COMMITted every `commit_interval`.
+and COMMITted every `commit_interval`.
 
 .. versionadded:: 2018.3.0
 
--- a/susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
+++ b/susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
@@ -1,22 +1,11 @@
-# -*- coding: utf-8 -*-
 """
 mgr_events.py is a SaltStack engine that writes selected events to SUSE
 Manager's PostgreSQL database. Additionally, it sends notifications via the
 LISTEN/NOTIFY mechanism to alert SUSE Manager of newly available events.
 
 mgr_events.py tries to keep the I/O low in high load scenarios. Therefore
-events are INSERTed once they come in, but not necessarily COMMITted
-immediately.
-
-The algorithm is an implementation of token bucket:
- - a COMMIT costs one token
- - initially, commit_burst tokens are available
- - every commit_interval seconds, one new token is generated
-   (up to commit_burst)
- - when an event arrives and there are tokens available it is COMMITted
-   immediately
- - when an event arrives but no tokens are available, the event is INSERTed but
-   not COMMITted yet. COMMIT will happen as soon as a token is available
+events are INSERTed with the separate thread with no blocking the event bus
+amd COMMITted every `commit_interval`.
 
 .. versionadded:: 2018.3.0
 
@@ -58,10 +47,12 @@ default for host is 'localhost'.
 
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
+import json
 import logging
-import time
-import fnmatch
 import hashlib
+import re
+import threading
+import time
 
 try:
     import psycopg2
@@ -71,10 +62,8 @@ except ImportError:
     HAS_PSYCOPG2 = False
 
 # Import salt libs
-import salt.version
 import salt.ext.tornado
 import salt.utils.event
-import json
 
 log = logging.getLogger(__name__)
 
@@ -96,11 +85,30 @@ class Responder:
         self.config.setdefault("postgres_db", {})
         self.config["postgres_db"].setdefault("host", "localhost")
         self.config["postgres_db"].setdefault("notify_channel", "suseSaltEvent")
-        self.counters = [0 for i in range(config["events"]["thread_pool_size"] + 1)]
-        self.tokens = config["commit_burst"]
+        self._commit_interval = config["commit_interval"]
+        self._commit_burst = config["commit_burst"]
+        self._thread_pool_size = config["events"]["thread_pool_size"]
+        self.counter = 0
+        self.counters = [0] * (self._thread_pool_size + 1)
+        # Pass salt job returns: salt/job/*/ret/*
+        self.re_job_ret = re.compile(r"salt\/job\/\d+\/ret\/.*")
+        # Pass all additional significant events
+        self.re_additional = [
+            re.compile(r"salt\/minion\/([^\/]+)\/start"),
+            re.compile(r"salt\/beacon\/.*"),
+            re.compile(
+                r"salt\/engines\/libvirt_events\/([^\/]+)\/(domain|pool|network)\/lifecycle"
+            ),
+            re.compile(r"salt\/engines\/libvirt_events\/([^\/]+)\/pool\/refresh"),
+            re.compile(r"salt\/batch\/([^\/]+)\/start"),
+            re.compile(r"suse\/manager\/(image_deployed|image_synced|pxe_update)"),
+            re.compile(r"suse\/systemid\/generate"),
+        ]
+        self._queue = []
         self.event_bus = event_bus
         self._connect_to_database()
-        self.event_bus.io_loop.call_later(config["commit_interval"], self.add_token)
+        self._queue_thread = threading.Thread(target=self.queue_thread)
+        self._queue_thread.start()
 
     def _connect_to_database(self):
         db_config = self.config.get("postgres_db")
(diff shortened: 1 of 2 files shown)


susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -114,45 +122,38 @@ def _connect_to_database(self):
...
+            # Pass all salt/job/*/ret/* events
+            self.re_job_ret.match(tag)
+            and not (
+                # Except mine updates
+                data.get("fun") == "mine.update"
+                or (
+                    # And presence test.ping returns
+                    data.get("fun") == "test.ping"
+                    and data.get("metadata", {}).get("batch-mode")
+                )
```
meaksh, line 153: Is there a particular reason behind removing the `_is_salt_mine_update`, `_is_test_ping` and `_is_batch_mode` helpers and just go with the unwrapped approach here?

meaksh (commented):
This looks good, thanks! I just made few comments

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
vzhestkov: Actually these methods just adds some readability but adds complexity for the python strack, in general it's a fight for some milliseconds, but in case of having large amount of events even if they are not passing the check it could save some CPU time.

meaksh (approved):
LGTM :+1:

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -118,57 +124,69 @@ def _connect_to_database(self):
...
             )
-            and not self._is_salt_mine_event(tag, data)
-            and not self._is_presence_ping(tag, data)
-        ):
+        ) or self.match_additional_events(tag):
+            # Pass all significant events also
+            queue = 0
+            if "id" in data:
+                hash_sum = hashlib.md5(
```
juliogonzalez: Sonarcloud is sending a failure about this:
> Make sure that hashing data is safe here.

If in the end this is safe, remember to configure sonarcloud.

vzhestkov: yup, it cought my eye also, but for different reason, there were some turbulence on rebase with https://github.com/uyuni-project/uyuni/pull/10551, I've cleaned up the code, so there shouldn't be such part anymore.

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -194,60 +203,58 @@ def _get_queue(self, minion_id):
...
-        )
-
-    def _is_test_ping(self, data):
-        return data.get("fun") == "test.ping"
-
-    def _is_batch_mode(self, data):
-        return data.get("metadata", {}).get("batch-mode")
+    def push_events_from_queue(self):
+        while self._queue and (
+            self.counter < self._commit_burst or self._commit_burst == 0
```
rjmateus, line 213: I'm not getting what the goal/behavior of _commit_burst == 0? If it's 0 then we should always be getting events from the queue? Then when are they commited?

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -194,60 +203,58 @@ def _get_queue(self, minion_id):
...
+        ):
+            tag, data = self._queue.pop(0)
+            try:
+                self._insert(tag, data)
+            # pylint: disable-next=broad-exception-caught
+            except Exception:
+                # Exception while inserting the data, put the data back to the queue
+                # and repeat the attempt of pushing it with the next cycle of queue_thread
+                # The exception was logged before in _insert
+                self._queue.insert(0, (tag, data))
```
rjmateus: In theory should not happen, but what if an event is recurrently causing insert errors? It will block all events from being get from the event queue right?

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -10,18 +10,8 @@
...
-The algorithm is an implementation of token bucket:
- - a COMMIT costs one token
- - initially, commit_burst tokens are available
- - every commit_interval seconds, one new token is generated
-   (up to commit_burst)
- - when an event arrives and there are tokens available it is COMMITted
-   immediately
- - when an event arrives but no tokens are available, the event is INSERTed but
-   not COMMITted yet. COMMIT will happen as soon as a token is available
+events are INSERTed with the separate thread with no blocking the event bus
```
m-czernek:
```suggestion
events are INSERTed with the separate thread without blocking the event bus
```

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -194,60 +203,58 @@ def _get_queue(self, minion_id):
...
-            and self._is_batch_mode(data)
-        )
-
-    def _is_test_ping(self, data):
-        return data.get("fun") == "test.ping"
-
-    def _is_batch_mode(self, data):
-        return data.get("metadata", {}).get("batch-mode")
+    def push_events_from_queue(self):
+        while self._queue and (
```
rjmateus, line 212: If I got it right, it should take events from the queue until it reaches the `_commit_burst`? In case we have a small number of events, it can take a long time until the event gets to the database?

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -194,60 +203,58 @@ def _get_queue(self, minion_id):
...
+                self.db_keepalive()
+                self.push_events_from_queue()
+                self.trace_log()
+                self.attempt_commit()
+            # pylint: disable-next=broad-exception-caught
+            except Exception as e:
+                log.error(
+                    "Exception while processing the events queue: %s", e, exc_info=True
+                )
+            time.sleep(self._commit_interval)
```
rjmateus, line 268: this thread sleep always happens, even if we have event on the queue to get processed. I think we could have a different approach, and commit data after a `_commit_burst` or `_commit_interval` is reached. What do you think?

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
```diff
@@ -194,60 +203,58 @@ def _get_queue(self, minion_id):
...
-        if salt.version.SaltStackVersion(*salt.version.__version_info__).major < 3004:
-            tag, data = self.event_bus.unpack(raw, self.event_bus.serial)
-        else:
+        try:
             tag, data = self.event_bus.unpack(raw)
-        self._insert(tag, data)
+        # pylint: disable-next=broad-exception-caught
+        except Exception as e:
+            log.warning("Unable to unpack the event data: %s", e)
+        self._queue.append((tag, data))
```
m-czernek: This will fail if `self.event_bus.unpack(raw)` throws an exception. 

Should we not move this line to L#228, right after the unpacking?

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
rjmateus: answered

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
rjmateus: It only stays in here if the queue has events

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
rjmateus: keeping the sleep in here, and giving some room for Java side of the process

susemanager-utils/susemanager-sls/susemanager-sls.changes.vzhestkov.improve-mgr_events-engine
```diff
@@ -0,0 +1,2 @@
+- Make mgr_events salt engine non-blocking on reading events
+- Avoid loosing the events on DB connection issues (bsc#1252098)
```
m-czernek:
```suggestion
- Avoid losing the events on DB connection issues (bsc#1252098)
```

m-czernek (dismissed):
+1, LGTM, adding minor comments.

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
vzhestkov: Yes, you are right, https://github.com/uyuni-project/uyuni/pull/8382/commits/f55d05119d0ad73622baeaf288b6c0ac0c982b1b should fix such situation

susemanager-utils/susemanager-sls/modules/engines/mgr_events.py
vzhestkov: Thanks @m-czernek, good catch, fixed in https://github.com/uyuni-project/uyuni/pull/8382/commits/d9a8a00d38e7fdd8dcf957df7457aa7d8c7bb040

PR #8407, opened 2024-03-06
Refactor Cucumber features to run through a containerized proxy

srbarrios:
## What does this PR change?

This PR refactor some cucumber features that were skipped to facilitate the first migration from using a traditional proxy to use a containerized proxy in our test suite.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

No ports.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


testsuite/features/step_definitions/setup_steps.rb
```diff
@@ -206,16 +206,10 @@
     And I wait until I see the name of "#{host}", refreshing the page
     And I follow this "#{host}" link
     And I wait until I see "System Status" text
+    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Apply states" is completed
```
Bischoff, line 209: we should have used "at most" in this step's name...

testsuite/features/step_definitions/setup_steps.rb
```diff
@@ -206,16 +206,10 @@
...
+    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Hardware List Refresh" is completed
+    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Package List Refresh" is completed
```
Bischoff, line 211: Don't we have something already to go through these 3 steps ?

"I wait until onboarding is completed" or something like that.

Bischoff (commented):
LGTM but see comments first

testsuite/features/step_definitions/setup_steps.rb
nodeg:
```ruby
When(/^I wait until onboarding is completed for "([^"]*)"$/) do |host|
  steps %(
    When I follow the left menu "Systems > System List > All"
    And I wait until I see the name of "#{host}", refreshing the page
    And I follow this "#{host}" link
  )
  if get_client_type(host) == 'traditional' && is_salt.empty?
    get_target(host).run('rhn_check -vvv')
  else
    steps %(
      And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Apply states" is completed
      And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Hardware List Refresh" is completed
      And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Package List Refresh" is completed
    )
  end
end
```
Would indeed be a good candidate.

testsuite/features/support/lavanda.rb
```diff
@@ -151,15 +151,16 @@ def os_version
   #
   # Args:
   #   cmd: The command to run.
+  #   runs_in_container: Whether the command should be run in the the container or in the host. Defaults to true.
```
nodeg, line 154:
```suggestion
  #   runs_in_container: Whether the command should be run in the container or on the host. Defaults to true.
```
Since `run()` is a wrapper for `run_local()` it makes sense to set this new parameter to `true` as default.

testsuite/features/secondary/min_ssh_tunnel.feature
```diff
@@ -94,7 +89,7 @@ Feature: Register a Salt system to be managed via SSH tunnel
     Then "ssh_minion" should not be registered
 
   Scenario: Cleanup: register a Salt minion after SSH tunnel tests
-    When I bootstrap "ssh_minion" using bootstrap script with activation key "1-SUSE-SSH-KEY-x86_64" from the proxy
+    When I bootstrap "ssh_minion" using bootstrap script with activation key "1-SUSE-SSH-KEY-x86_64" from the server
```
nodeg, line 92: Why do we use the server instead of the proxy here?

testsuite/features/step_definitions/setup_steps.rb
srbarrios: As you can see, I did not change these steps, I just removed the if statement for the traditional stack. So, I prefer to don't do any change on these steps on this PR.

testsuite/features/step_definitions/setup_steps.rb
srbarrios: As you can see, I did not change these steps, I just removed the if statement for the traditional stack. So, I prefer to don't do any change on these steps on this PR.

testsuite/features/secondary/min_ssh_tunnel.feature
srbarrios: This is a cleanup step, so we don't really care if we use the server or the proxy, but to answer your question, this is changed to `# workaround for bsc#1220864
`

Basically, mgr-bootstrap command is not available in the proxy container anymore, only  in the server container. They want to figure out a better implementation, returning the script via API instead of generating it through the command mgr-bootstrap.

testsuite/features/step_definitions/setup_steps.rb
Bischoff: ok, approving then

Bischoff (approved):

nodeg (approved):

PR #8414, opened 2024-03-07
Remove clients from private network

Bischoff:
## What does this PR change?

Remove traditional and salt client from the private network.

It would have been too complicated to keep them with the new sumaform setup (see uyuni-project/sumaform#1525).

Also prepare a general infrastructure for Cobbler and Retail in case of containers.

## Context

In a non-containerized setup, the private network will be completely empty at the beginning. This is why we remove the ping tests, as we have nothing to ping anymore. It is okay to not have these tests, because:
 * in Uyuni and Head the non-containerized setup with go away soon
 * for 4.3, those tests have never caught an error in the past

For the containerized proxy, I will re-add ping checks, but with a slightly different name, and with a completely different implementation, as we will now ping the DHCP and DNS branch server, which will be the only VM on the private network at the beginning.

## Links

Issue(s): uyuni-project/sumaform#1451
Port(s): 
* 4.3: https://github.com/SUSE/spacewalk/pull/23949

## Changelogs

- [x] No changelog needed


testsuite/features/step_definitions/retail_steps.rb
```diff
@@ -91,27 +91,54 @@
   end
 end
 
-Then(/^terminal "([^"]*)" should have got a retail network IP address$/) do |host|
+Then(/^"([^"]*)" should communicate with the server using public interface/) do |host|
```
srbarrios, line 94: Then xxx **host** should yada yada

testsuite/features/step_definitions/retail_steps.rb
```diff
@@ -91,27 +91,54 @@
...
+  unless return_code.zero?
+    sleep 2
+    puts 're-try ping'
+    node.run("ping -c 1 -I #{node.public_interface} #{get_target('server').public_ip}")
+  end
+  get_target('server').run("ping -c 1 #{node.public_ip}")
 end
 
-Then(/^name resolution should work on terminal "([^"]*)"$/) do |host|
+Then(/^"([^"]*)" should not communicate with the server using private interface$/) do |host|
```
srbarrios, line 105: Same, Then xxx **host**

testsuite/features/step_definitions/retail_steps.rb
```diff
@@ -91,27 +91,54 @@
...
+  node = get_target('proxy')
+  static_dns = net_prefix + PRIVATE_ADDRESSES['dhcp_dns']
+  cmd = 'echo -e "BOOTPROTO=dhcp\nSTARTMODE=auto\n" > /etc/sysconfig/network/ifcfg-eth1 && ' \
+        'ifup eth1 && ' \
+        "sed -i 's/^NETCONFIG_DNS_STATIC_SERVERS=\".*\"/NETCONFIG_DNS_STATIC_SERVERS=\"#{static_dns}\"/' /etc/sysconfig/network/config && " \
+        'netconfig update -f'
+  node.run(cmd)
+end
+
+Then(/^"([^"]*)" should be present on private network$/) do |host|
```
srbarrios, line 123: Same, Then xxx **host**

testsuite/features/step_definitions/retail_steps.rb
Bischoff: It will go into next PR anyway, I will restrict the scope of this PR

testsuite/features/step_definitions/retail_steps.rb
Bischoff: Same

testsuite/features/step_definitions/retail_steps.rb
Bischoff: Same

Bischoff:
Tested with success with new sumaform code on 4.3 branch.

I could not test on uyuni master branch to other bugs.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
```diff
@@ -12,10 +12,6 @@
 # * if there is no private network ($private_net is nil)
 # * if there is no PXE boot minion ($pxeboot_mac is nil)
 
-# TODO: This feature needs to be refactored for RBS containerized,
-#       as it does not support Salt formulas.
-#       Additionally, we have this bug that impedes to use mgr-bootstrap command https://bugzilla.suse.com/show_bug.cgi?id=1220864
```
srbarrios, line 17: I still reading from the top to the bottom this PR, letting this message here, because for now I don't understand why we can remove this message, in particular, how we solved this issue: **_We have this bug that impedes to use mgr-bootstrap command https://bugzilla.suse.com/show_bug.cgi?id=1220864_**

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -42,24 +42,6 @@
...
-  get_target('server').run("ping -c 1 #{node.public_ip}")
-end
-
-Then(/^"([^"]*)" should not communicate with the server using private interface/) do |host|
-  node = get_target(host)
-  node.run_until_fail("ping -c 1 -I #{node.private_interface} #{get_target('server').public_ip}")
-  # commented out as a machine with the same IP address might exist somewhere in our engineering network
-  # get_target('server').run_until_fail("ping -c 1 #{node.private_ip}")
-end
-
```
srbarrios, line 62: I miss some extended information in the PR description about the strategy we followed here, and why this is not anymore necessary.

testsuite/features/build_validation/retail/proxy_traditional_branch_network.feature
```diff
@@ -130,6 +131,7 @@ Feature: Prepare the branch server for PXE booting
     And I enter "proxy" in first A name field of example.org zone
     And I enter the local IP address of "proxy" in first A address field of example.org zone
     And I press "Add Item" in A section of example.org zone
+    # TODO: don't hardcode anymore the names in the private network once we have them in .bashrc
```
srbarrios, line 134: Leftover or this still pending?
Have we a card to address and link this dependency?

testsuite/features/step_definitions/command_steps.rb
Bischoff: It is not called anymore in non-containerized proxy context.

Since we don't have traditional client and salt minion on the private network, we have nothing to ping anymore... As a matter of fact, in a non-containerized setup, the private network will be completely empty at the beginning.

For the containerized proxy, I will re-add ping checks, but with a slightly different name, and with a completely different implementation, as we will now ping the DHCP and DNS branch server, which will be the only VM on the private network at the beginning.

The non-containerized setup with remain without ping sanity checks, it will go away soon enough in Uyuni, and for 4.3 it never caught an error in the past, so...

testsuite/features/support/constants.rb
```diff
@@ -72,12 +72,14 @@
   'salt_migration_minion' => 'SALT_MIGRATION_MINION'
 }.freeze
 
-ADDRESSES = {
+# TODO: the values for pxeboot_minion, sle12sp5_terminal, sle15sp4_terminal, and proxy can now be set in sumaform
+#       remove them from this array when we read them from .bashrc
+PRIVATE_ADDRESSES = {
   'network'           => '0',
-  'sle_minion'        => '3',
```
srbarrios, line 77: Aside of retail, I wonder if we don't use this value for sle_minion in another scope. 👀

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: I am removing this message because this is precisely what this series of PRs is about: refactoring for Cobbler and Retail in a containerized setup.

In all logic, this removal would have to be done in next PR, not this one, but I was afraid of forgetting it, so I am doing it a bit in advance.

As for the bug, yes I know about it. If I am not wrong, it will only make the second half of this file (the mass import) fail. BTW it already fails in 4.3 for other reasons.

testsuite/features/build_validation/retail/proxy_traditional_branch_network.feature
Bischoff: Yes, it's rather long term and I don't have a card. It would need a second sumaform refactor in order to pass these values in bashrc.

What do you think? We say YAGNI and I remove this remark?

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
```diff
@@ -12,10 +12,6 @@
...
-
 @skip_if_containerized_server
```
srbarrios, line 15: I would assume we also need to split this feature in two, one for traditional and another for containerized. Make sense?

testsuite/features/support/constants.rb
Bischoff: sle minion does not even have a networking card in the private network anymore. So it is strictly impossible.

And also, I checked with a grep ;-) . And the tests passed.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
srbarrios: Reverse YAGNI :)
Let's remove it only when is not anymore meaningful, together with the changes that invalidate it.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: That is a very good and important question.

My first tests seem to show I can cope with one single code for both. But I'm not advanced enough, so yes, we might need to duplicate it :cry:  .

testsuite/features/step_definitions/command_steps.rb
srbarrios: It would be awesome if you can explain this in the PR description.

testsuite/features/support/constants.rb
srbarrios: Cool.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: I will re-add the remark about the bug, because my second PR will not make the issue go away.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
srbarrios: I doubt we can make it with one, the reason is that for a traditional RBS, the doc explains that we need to use its Salt Formulas to configure some stuff, while in a containerized RBS the doc tells you to handle it differently.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
srbarrios: Thank you.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: Naturally, the tag `@skip_if_containerized_server` at feature level will go away in both hypothesis.

testsuite/features/step_definitions/command_steps.rb
Bischoff: ok

srbarrios (approved):
LGTM. Just please link the TODO with a card, and please describe in the PR description what you commented in the threads for better track of our changes :)

testsuite/features/build_validation/retail/proxy_traditional_branch_network.feature
Bischoff: Fair enough. Let's do a card.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: Strictly speaking it is still meaningful in the context of this PR. I will re-add the whole comment.

If I forget to remove it in next PR it will be your fault :trollface: .

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: I am not that sure about the formulas: in my understanding, we drop the DHCP and DNS formulas, but we keep the other ones.

But I might be wrong here. I lack experience.

testsuite/features/secondary/proxy_retail_pxeboot_and_mass_import.feature
Bischoff: Done

PR #8428, opened 2024-03-08
Confidential computing attestation container

mackdk:
## What does this PR change?

This PR defines a new microservice in java to handle attestation results and process them.  It runs in a new separate container which is as well defined in this PR. The design is modular and the service can handle multiple attestation modules dynamically loaded at runtime. Each module is able to handle a specific attestation result type.

The java project is defined in Maven. I created a parent pom `uyuni-java-parent` where all the dependency versions are centralized. Then there are the following modules:

- `uyuni-java-common`, which contains some basic functionality meant to be reusable across multiple java projects
- `coco-attestation`, root project for the real attestation service divided into:
     - `attestation-core`, the main component handling the queue processor logic, module loading and process result handling.
     - `attestation-module`, contains the classes and the interfaces needed to implement an attestation module.
     - `attestation-module-snpguest`, the attestation module for snpguest (real implementation is still WIP).

Additional attestation modules will be implemented. The current implementation uses Postgresql notification to avoid polling the database table for data.

### Open discussion points

- **Naming and modularization**: The modularization was created in this way because this was meant to be a base that can be reused to write new services in java when they do not require the full uyuni codebase, but it's not currently strictly needed (a part from the separation attestation core/modules). The whole code is now contained in a `microservices` subfolder, which is not ideal. The right place should be of course `java`, but to my understanding it's not possible without moving the existing spacewalk-java package to a subfolder (which I guess we strongly don't want to do :slightly_smiling_face:). Any suggestion regarding better naming and modularization is welcome.
- **Container**: I tried to follow the structure and naming consistent with what we have in the `containers` folder. Please double check it.
- **Packaging**: this is currently split into 3 OBS projects. All of them build with maven with a minor tweaking on the pom configuration.
     - `uyuni-java-parent`: which only contains the parent pom for building
     - `uyuni-java-common`: which packages the corresponding maven project
     - `uyuni-coco-attestation`: packages the actual service. The main package contains `attestation-core` and `attestation-module`, all attestation module will be separated into sub-packages. Currently only `uyuni-coco-attestation-module-snpguest` exists.
- **Versioning**: for now I kept the default `1.0.0-SNAPSHOT`. We need to figure out a versioning scheme and how to handle versioning when tagging.
- **Java dependencies**: the code only uses libraries that were already part of uyuni. Currently the project depends on `log4j 2`, `c3p0` for connection pooling, `mybatis` for database access and `postgresql-jdbc` for the database driver. Currently, uyuni relies on `pgjdbc-ng` for receiving notifications from the database, but the current driver offers a similar functionality and the former is stale and has been basically discontinued since three years ago since the authors are no longer using jdbc so they have no motivation to maintain it (https://github.com/impossibl/pgjdbc-ng/issues/579). Tests are using `junit 5` and `mockito` (for which I'm using version 5.10.0 that is not provided as a package in OBS but it's not needed by the package build process nor of course at runtime). Additional dependencies might be considered to simplify further developments. In particular a dependency injection mechanism could greatly improve the code. Currently, class dependencies are injected through constructors, but a framework could be used to enforce the practice. We could even implement something basic on our own  (something like [feather](https://github.com/zsoltherpai/feather)). Another nice thing to have IMO would be `@CheckForNull/@Nonnull` annotations to help static analysis.
- **GitHub actions**: further configuration will be needed to run unit tests and static analysis on this code. That's still not part of this PR.

This PR is currently **NOT** targeting master. It will be merged in the coco attestation feature branch.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- Documentation to be discussed

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23552

- [X] **DONE**

## Changelogs

- [X] No changelog needed

## Re-run a test

# Before you merge


microservices/coco-attestation/attestation-core/src/main/resources/mappers/attestation-result.xml
```diff
@@ -0,0 +1,58 @@
...
+                    , attested
+            FROM suseCoCoAttestationResult
+           WHERE id = #{id}
+                    AND status = 'PENDING'
+        FOR NO KEY UPDATE NOWAIT
+    </select>
+
+    <select id="updateAndRetrieve" parameterType="long" resultType="com.suse.coco.model.AttestationResult">
+          UPDATE suseCoCoAttestationResult
+             SET status = 'PROCESSING'
```
mcalmer, line 45: We need to add this status to the DB and to the tomcat Java code

microservices/coco-attestation/attestation-module/src/main/java/com/suse/coco/modules/AttestationModule.java
```diff
@@ -0,0 +1,51 @@
...
+     * Retrieves the name of the attestation module
+     * @return the unique name of this module
+     */
+    String getName();
+
+    /**
+     * Retrieves attestation result type that this module is able to verify.
+     * @return the supported result type
+     */
+    int getSupportedType();
```
mcalmer, line 36: Maybe it would be a good to tell where the types are specified and point to com/suse/manager/model/attestation/CoCoResultType.java

mcalmer (commented):
just a few points for now

microservices/coco-attestation/attestation-core/src/main/resources/mappers/attestation-result.xml
mackdk: Sorry, this is no longer needed. I will remove it

lucidd (approved):

microservices/uyuni-java-parent/uyuni-java-parent.spec
```diff
@@ -0,0 +1,46 @@
...
+# license for the pristine package is not an Open Source License, in which
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
+#
+
+Name:           uyuni-java-parent
+Version:        1.0.0
```
deneb-alpha, line 19:
```suggestion
Version:        5.0.0
```
please, let's align the versions  already

microservices/uyuni-java-common/uyuni-java-common.spec
```diff
@@ -0,0 +1,59 @@
+#
+# spec file for package coco-attestation
```
deneb-alpha, line 2: this is supposed to be the name of the spec
```suggestion
# spec file for package uyuni-java-common
```

microservices/uyuni-java-parent/uyuni-java-parent.spec
```diff
@@ -0,0 +1,46 @@
+#
+# spec file for package coco-attestation
```
deneb-alpha, line 2: this is supposed to be the name of the spec
```suggestion
# spec file for package uyuni-java-parent
```

rel-eng/packages/server-attestation-image
```diff
@@ -0,0 +1 @@
+5.0.4 containers/server-attestation-image/
```
deneb-alpha, line 1: If you change the image version in the docker file, please, adjust also this. 
From my  POV they should both  be 5.0.0, not  5.0.4

microservices/coco-attestation/uyuni-coco-attestation.spec
```diff
@@ -0,0 +1,113 @@
...
+# license for the pristine package is not an Open Source License, in which
+# case the license is the MIT License). An "Open Source License" is a
+# license that conforms to the Open Source Definition (Version 1.9)
+# published by the Open Source Initiative.
+
+# Please submit bugfixes or comments via https://bugs.opensuse.org/
+#
+
+Name:           uyuni-coco-attestation
+Version:        1.0.0
```
deneb-alpha, line 19: Like for the other spec files  in this  PR,  let's align with the other  sources of SUSE Manager 5.0

```suggestion
Version:        5.0.0
```

you can also go with a beta release  or something similar,  but  please, let's  use  a 5.X

microservices/coco-attestation/uyuni-coco-attestation.spec
```diff
@@ -0,0 +1,113 @@
+#
+# spec file for package coco-attestation
```
deneb-alpha, line 2: like for the others, it should be the spec file  name
```suggestion
# spec file for package uyuni-coco-attestation
```

containers/server-attestation-image/server-attestation-image.changes
```diff
@@ -0,0 +1,9 @@
+Tue Feb 13 16:59:23 CET 2024 - kwalter@suse.com
```
deneb-alpha, line 1: this is wrong folks.

the changelog must  follow a chronological  order  with  oldest entries on the bottom.

lucidd: ignore the last commit for now it's just temporary to see what is needed to make the test setup happy

microservices/coco-attestation/uyuni-coco-attestation.spec
```diff
@@ -0,0 +1,126 @@
...
+Package containing the Javadoc API documentation for %{name}.
+
+%prep
+%setup -q
+
+# Disable the common module it will be provided by the installed dependency
+%pom_disable_module '../uyuni-java-common'
+
+# Make sure there are not dependencies on the modules projects
+%pom_remove_dep 'org.uyuni-project.coco-attestation.module:' attestation-core/pom.xml
```
rjmateus, line 65: Is this line removing the dependency for the core module only or for all?

rjmateus (commented):
First round about the code structure and maven. So far looks realy good from my point of view. Congrants on all the effort put on this

microservices/uyuni-java-parent/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+            </dependency>
+            <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter</artifactId>
+                <version>5.8.2</version>
+            </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-core</artifactId>
+                <version>5.10.0</version>
```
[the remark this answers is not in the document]
mackdk: This is just the `<depedencyManagement>` declaration, so I did not include the scope, but  yes we can define it directly here as `test` as well. I will fix it. Same applies to the other 2 entries.

microservices/uyuni-java-common/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.mchange</groupId>
+            <artifactId>c3p0</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <scope>test</scope>
```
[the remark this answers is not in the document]
mackdk: For compilation of the `uyuni-java-common` library, only the `log4j-api` is needed. Of course you will need to provide an implementation at runtime to actually get any logging. Since this is only a common jar, though, I let the decision of that dependency to the consumer of the library. On second thought, I could probably define it as runtime optional.

microservices/coco-attestation/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.uyuni-project</groupId>
+        <artifactId>uyuni-java-parent</artifactId>
+        <version>5.0.0</version>
+        <relativePath>../uyuni-java-parent</relativePath>
+    </parent>
+
+    <artifactId>coco-attestation</artifactId>
+    <version>5.0.0</version>
```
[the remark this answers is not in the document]
mackdk: I actually wanted the version to be defined in a single place, but I probably made something wrong and it wasn't working. I have a look at the link you provided.

microservices/coco-attestation/uyuni-coco-attestation.spec
mackdk: It's only for  `attestation-core`. This lines make sure the maven pom for attestation-core.jar does not declare any dependency on a module implementation. This is because if the pom contains such dependency, the `uyuni-coco-attestation` package we create in OBS will require the corresponding module package, and we don't want that, as the modules should be pluggable in any combination (it should be the module that requires the core, not the opposite).

On the other hand, during development it's convenient to declare such runtime dependency, otherwise by default you won't have any module on the classpath when you run the app through an IDE. I could probably define a configuration for the maven exec plugin to achieve the same though. I will look into that.

microservices/coco-attestation/attestation-module/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <groupId>org.uyuni-project.coco-attestation</groupId>
+    <artifactId>attestation-module</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mybatis</groupId>
```
rjmateus, line 36: Is mybatis used in here?

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/SNPGuestWorker.java
```diff
@@ -0,0 +1,125 @@
...
+                return false;
+            }
+
+             LOGGER.debug("Loaded report {}", report);
+             if (report.getCpuGeneration() == EpycGeneration.UNKNOWN) {
+                LOGGER.error("Unable to identify Epyc processor generation for attestation report {}", report.getId());
+                return false;
+            }
+
+            try (var workingDir = directoryProvider.createDirectoryFor(result.getId(), report)) {
```
rjmateus, line 83: Does this directory need to be clean/removed after processing?

microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/attestation/AttestationQueueProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+
+        // Shut down the executor
+        shutdownExecutor();
+
+        // Interrupt the listener thread if still alive
+        stopListenerThread(listenerThread);
+    }
+
+    private Thread createListenerThread() {
+        Thread listenerThread = new Thread(this::listenForAttestationResults, "attestation-processor-listener");
```
rjmateus, line 160: In here we are starting a new thread to listen to changes and notify. 
What will happen if for any reason this thread dies? Will the program wait forever for a notification and stop processing stuff silently?

microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/attestation/AttestationQueueProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+
+        try {
+            while (!Thread.currentThread().isInterrupted() && listenerRunning.get()) {
+                List<Long> results = service.getPendingResultByType(moduleLoader.getSupportedResultTypes(), batchSize);
+                if (results.isEmpty()) {
+                    LOGGER.info("No attestation result to process - Waiting");
+                    synchronized (dataAvailable) {
+                        dataAvailable.wait();
+                    }
+                    continue;
```
rjmateus, line 121: great trick :)

rjmateus (commented):
Just some final comments, but over all I really like the code structure and organization. thank you for the great work

microservices/uyuni-java-common/src/main/java/com/suse/common/configuration/FileConfigurationSource.java
```diff
@@ -0,0 +1,213 @@
...
+     */
+    public FileConfigurationSource(List<Path> fileSourcesIn, String filePrefixIn, List<String> fallbackNamespacesIn) {
+        configValues = new Properties();
+
+        commonFilePrefix = filePrefixIn;
+        fallbackNamespaces = fallbackNamespacesIn;
+
+        fileSourcesIn.stream()
+            .flatMap(path -> getConfigurationFiles(path))
+            .sorted(CONFIG_PATH_COMPARATOR)
```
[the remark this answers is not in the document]
mackdk: No, this is currently not used. It's a re-implementation of the `Config` class we use in Uyuni and it's almost backwards compatible (a part for a slightly different management of the missing values). Initially I thought we might need to read the same configuration files, hence I implemented this class. In the end, it turned out we didn't need it at all. 

Anyway, this is implementing the same sorting logic we currently have, see the [Config class](https://github.com/uyuni-project/uyuni/blob/f8ee365162741cc8172794de08b8ed9500b4fc43/java/code/src/com/redhat/rhn/common/conf/Config.java#L89-L102) for reference.

microservices/uyuni-java-common/src/main/java/com/suse/common/configuration/FileConfigurationSource.java
```diff
@@ -0,0 +1,213 @@
...
+        catch (Exception ex) {
+            LOGGER.error("Could not parse file {}", file, ex);
+            return new Properties();
+        }
+
+        String namespace = makeNamespace(file);
+        LOGGER.debug("Adding namespace: {} for file: {}", namespace, file.toAbsolutePath());
+        for (Object key : properties.keySet()) {
+            String propertyName = (String) key;
+            if (!propertyName.startsWith(namespace)) {
```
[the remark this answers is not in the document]
mackdk: Not fully sure to understand. This code is meant to ensure all the properties defined in a file named for example `rhn_web.conf` starts with `web.`. In case the property does not start in the correct way, the if branch removes it from the set and then reinsert the value with the correct key. Otherwise, if the property has already the correct prefix, it is kept as is.

microservices/uyuni-java-common/src/main/java/com/suse/common/configuration/FileConfigurationSource.java
```diff
@@ -0,0 +1,213 @@
...
+        String propertyName = property;
+
+        int lastDot = property.lastIndexOf('.');
+        if (lastDot > 0) {
+            propertyName = property.substring(lastDot + 1);
+            namespace = property.substring(0, lastDot);
+        }
+
+        LOGGER.debug("getString() - Getting property: {}", propertyName);
+        String result = configValues.getProperty(propertyName);
```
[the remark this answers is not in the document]
mackdk: If a property is defined without a namespace, it will always override the namespaced one. That's the current behaviour of our configuration logic. Check the [test `testOverride()` in original `ConfigTest`](https://github.com/uyuni-project/uyuni/blob/f8ee365162741cc8172794de08b8ed9500b4fc43/java/code/src/com/redhat/rhn/common/conf/test/ConfigTest.java#L102-L110).

microservices/coco-attestation/attestation-module/pom.xml
mackdk: Yes, because the `SqlSession` object is passed to the module to allow querying the database within the same session where the attestation result row is locked. 

We could consider abstracting the MyBatis object behind an interface and hide the implementation details. This would allow the code to not depend on a specific DAO layer. It would essentially mean to rewrite some MyBatis interfaces, and I'm not sure if it's worth. 

The search server code does something similar though, by providing the [`DatabaseManager`](https://github.com/uyuni-project/uyuni/blob/f8ee365162741cc8172794de08b8ed9500b4fc43/search-server/spacewalk-search/src/java/com/redhat/satellite/search/db/DatabaseManager.java) and [`Query`](https://github.com/uyuni-project/uyuni/blob/f8ee365162741cc8172794de08b8ed9500b4fc43/search-server/spacewalk-search/src/java/com/redhat/satellite/search/db/Query.java).

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/SNPGuestWorker.java
mackdk: The cleanup is currently done [in the `close()` method of `VerificationDirectory`](https://github.com/uyuni-project/uyuni/blob/7c04887039ba03ca2cfa110ae416068786dd083c/microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/io/VerificationDirectory.java#L71-L92) and called outmatically by the `try-with-resource`.

microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/attestation/AttestationQueueProcessor.java
mackdk: If the listener thread dies, the main thread should stop as [the loop is guarded by the `listenerRunning` flag](https://github.com/uyuni-project/uyuni/blob/7c04887039ba03ca2cfa110ae416068786dd083c/microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/attestation/AttestationQueueProcessor.java#L114) which is [toggled when the listener stops](https://github.com/uyuni-project/uyuni/blob/7c04887039ba03ca2cfa110ae416068786dd083c/microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/attestation/AttestationQueueProcessor.java#L237-L238). Also there is [a test case checking that](https://github.com/uyuni-project/uyuni/blob/7c04887039ba03ca2cfa110ae416068786dd083c/microservices/coco-attestation/attestation-core/src/test/java/com/suse/coco/attestation/AttestationQueueProcessorTest.java#L99-L127). 

I have to  expand the coverage of this class though, to verify all the possible edge cases betweem the two threads.

mcalmer (approved):

mcalmer (approved):

PR #8446, opened 2024-03-12
QE: Run a containerized proxy in our GH Validation

srbarrios:
## What does this PR change?

We want to run our GH Validation using a Containerized Proxy.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- GH Validation tests

- [x] **DONE**

## Links

No ports

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


testsuite/features/github_validation/init_clients/proxy_container.feature
```diff
@@ -0,0 +1,76 @@
...
+    And I wait until "uyuni-proxy-tftpd" service is active on "uyuni-proxy-test"
+    And I wait until port "8022" is listening on "uyuni-proxy-test" container
+    And I wait until port "80" is listening on "uyuni-proxy-test" container
+    And I wait until port "443" is listening on "uyuni-proxy-test" container
+    And I visit "uyuni-Proxy-test" endpoint of this "uyuni-proxy-test"
+
+  @skip
+  Scenario: containerized proxy should be registered automatically
+    When I follow the left menu "Systems"
+    And I wait until I see the name of "uyuni-proxy-test", refreshing the page
```
srbarrios, line 76: This part is not relying on Proxy Host, this is Proxy containerized.

testsuite/podman_runner/12_run_proxy.sh
```diff
@@ -0,0 +1,152 @@
...
+
+  ssh-keygen -t rsa -b 4096 -C "salt@uyuni-server-all-in-one-test" -f /tmp/test-all-in-one/id_openssh_rsa -N ""
+
+  cat <<EOF > $PROXY_UTILS/config.yaml
+server: uyuni-server-all-in-one-test
+ca_crt: |
+$(sudo --login sed 's/^/  /' /tmp/test-all-in-one/RHN-ORG-TRUSTED-SSL-CERT)
+proxy_fqdn: proxy-httpd
+max_cache_size_mb: 2048
+server_version: 5.0.0 Beta1
```
srbarrios, line 25: I'm not sure if the value this parameter it has some impact, or it's just informative.
But I would better parametrize this in a way that we don't need to commit a change here for every new version.

testsuite/features/github_validation/init_clients/proxy_container.feature
```diff
@@ -0,0 +1,76 @@
...
+    And I enter "Proxy Key x86_64" as "description"
+    And I enter "PROXY-KEY-x86_64" as "key"
+    And I click on "Create Activation Key"
+    Then I should see a "Activation key Proxy Key x86_64 has been created" text
+
+  @skip
+  Scenario: Bootstrap the proxy host as a salt minion
+    When I follow the left menu "Systems > Bootstrapping"
+    Then I should see a "Bootstrap Minions" text
+    When I enter the hostname of "uyuni-proxy-test" as "hostname"
```
srbarrios, line 44: I wonder if we could simplify this name, not sure how many files are affected by this, but as hostname, it would be simpler to call it "proxy".

testsuite/podman_runner/12_run_proxy.sh
```diff
@@ -0,0 +1,152 @@
...
+    --detach \
+    --tty \
+    --pod uyuni-proxy-test \
+    --network uyuni-network-1 \
+    --cgroupns host \
+    --volume $PROXY_UTILS/:/etc/uyuni \
+    --volume $PROXY_UTILS/proxy-rhn-cache/:/var/cache/rhn \
+    --volume $PROXY_UTILS/proxy-tftpboot/:/srv/tftpboot \
+    --name proxy-httpd \
+      registry.opensuse.org/uyuni/proxy-httpd:2024.02
```
srbarrios, line 73: We need to parametrize the tag "2024.02" (or just use "latest" or none).
Otherwise we will need to do a commit change in these files everytime we have a new version.
Versions should have a unique source of truth, otherwise maintenance becomes a mess.

testsuite/podman_runner/12_run_proxy.sh
jordimassaguerpla: You could do it similarly to what we do for the containerized server. See https://github.com/uyuni-project/uyuni/blob/master/testsuite/podman_runner/07_start_server.sh#L31

testsuite/features/github_validation/init_clients/proxy_container.feature
jordimassaguerpla: the server is called uyuni-server-test, so I guess this is why. I would prefer to have a similar name for the proxy , as uyuni-proxy-test.

testsuite/podman_runner/12_run_proxy.sh
```diff
@@ -0,0 +1,152 @@
...
+    --publish 8088:8080 \
+    --add-host uyuni-server-all-in-one-test:$(sudo --login podman exec uyuni-server-all-in-one-test bash -c 'hostname -I | cut -d" " -f1')
+
+  sudo --login podman run \
+    --privileged \
+    --rm \
+    --detach \
+    --tty \
+    --pod uyuni-proxy-test \
+    --network uyuni-network-1 \
```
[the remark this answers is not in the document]
jordimassaguerpla: This is how the network is setup in the first scripts, can't be changed here. We could change that in all scripts, but it is not in the scope of this PR in my opinion.

testsuite/podman_runner/12_run_proxy.sh
jordimassaguerpla: You could do it similarly to what we do for the containerized server. See https://github.com/uyuni-project/uyuni/blob/master/testsuite/podman_runner/07_start_server.sh#L31

testsuite/features/github_validation/init_clients/proxy_container.feature
szachovy: Yes, exactly. I think I will leave it like this and eventually we will rename both server and proxy one day.

testsuite/podman_runner/12_run_proxy.sh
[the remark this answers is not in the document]
szachovy: Ok, the network name remain as it is.

testsuite/podman_runner/12_run_proxy.sh
szachovy: Now I point them with `$VERSION` like above.

testsuite/features/github_validation/init_clients/proxy_container.feature
szachovy: There is a problem with the current testsuite as it tests whether `systemd` services for booting proxies are active [link](https://github.com/uyuni-project/uyuni/blob/ffa2bea566e8e8b72541eec986582437e3a98139/testsuite/features/step_definitions/command_steps.rb#L283). Here, I purposely boot proxies via Podman directly, so I do not use `systemd` routines but plain proxy configuration based on `uyuni-server-all-in-one-test`. Hence, in the current state, the same tests won't work here. Therefore, there should be some extensions in these methods to accept running containers and check their availability. Rewriting the testsuite would be much more time-consuming, as initial attempts with onboarding failed. I have pinned the latest [compare diff](https://github.com/uyuni-project/uyuni/compare/d74f40587087d58530521107275b66c16761a989..09a021bbd8d432446595e790a829226de9fdec1e) from attempts. Currently, I have found the configuration where all the containers are up and running. At the moment, I can't find a convenient route to introduce others.

testsuite/podman_runner/12_run_proxy.sh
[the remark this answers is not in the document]
srbarrios: Ok, I opened a Uyuni card [here](https://github.com/uyuni-project/uyuni/issues/8564), for the future.

testsuite/features/github_validation/init_clients/proxy_container.feature
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2024 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+# The scenarios in this feature are skipped if:
+# * there is no proxy ($proxy is nil)
+#
+# Bootstrap the proxy as a Pod
```
srbarrios, line 7: This test is not bootstrapping a proxy, but just checking its status.
As we are bypassing all the bootstrap procedure. As first step to introduce the proxy, I'm fine with this approach, but let's make it clear in this feature.

```suggestion
# Check pod and containers of the proxy
```

testsuite/features/github_validation/init_clients/proxy_container.feature
```diff
@@ -0,0 +1,23 @@
...
+
+@proxy
+Feature: Setup containerized proxy
+  In order to use a containerized proxy with the server
+  As the system administrator
+  I want to register the containerized proxy on the server
```
srbarrios, line 13:
```suggestion
Feature: Check containerized proxy
  In order to use a containerized proxy with the server
  As the system administrator
  I want to check the status of the containerized proxy
```

testsuite/podman_runner/12_run_proxy.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
```
srbarrios, line 1: Maybe to make it clear, we should rename this script and mention that we are also creating the configuration inside it, not only running the container.

testsuite/podman_runner/12_run_proxy.sh
```diff
@@ -0,0 +1,148 @@
...
+set -euxo pipefail
+
+create_proxy_configuration() {
+
+  PROXY_UTILS=$HOME/proxy
```
srbarrios, line 6: IMHO, the word PROXY_UTILS could bring some misunderstandings, as we already have the word tools for other official components, mgr-tools (aka uyuni-tools) and also for client-tools.

testsuite/podman_runner/12_run_proxy.sh
```diff
@@ -0,0 +1,148 @@
...
+  server_ssh_key_pub: |
+$(sudo --login sed 's/^/    /' /tmp/test-all-in-one/ssh_host_rsa_key.pub)
+  server_ssh_push: |
+$(sudo --login sed 's/^/    /' /tmp/test-all-in-one/id_openssh_rsa)
+  server_ssh_push_pub: |
+$(sudo --login sed 's/^/    /' /tmp/test-all-in-one/id_openssh_rsa.pub)
+EOF
+}
+
+run_proxy_containers() {
```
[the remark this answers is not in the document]
szachovy: Isn't that the main requirement to do not use mgr tools? https://github.com/SUSE/spacewalk/issues/20679

testsuite/run_sets/github_validation/github_validation_init_clients.yml
```diff
@@ -6,4 +6,5 @@
 - features/github_validation/init_clients/sle_minion.feature
 - features/github_validation/init_clients/min_rhlike_salt.feature
 - features/github_validation/init_clients/min_deblike_salt.feature
+- features/github_validation/init_clients/proxy_container.feature
```
srbarrios, line 9: In fact, we are not initializing it... but.. in any case, this feature make more sense on top of this list.
As having the proxy correctly working it's a dependency for all the clients, which must be connected thru the proxy.

testsuite/run_sets/github_validation/github_validation_init_clients.yml
```diff
@@ -6,4 +6,5 @@
 - features/github_validation/init_clients/sle_minion.feature
```
srbarrios, line 6: I think that we need to adapt `sle_ssh_minion.feature` in order to bootstrap using the proxy.

testsuite/features/github_validation/init_clients/proxy_container.feature
szachovy: done

testsuite/features/github_validation/init_clients/proxy_container.feature
szachovy: done

testsuite/podman_runner/12_run_proxy.sh
szachovy: You are right, I renamed it to `12_setup_and_run_proxy.sh`

testsuite/podman_runner/12_run_proxy.sh
szachovy: Ok, changed to `PROXY_HOME_DIRECTORY`

testsuite/podman_runner/12_run_proxy.sh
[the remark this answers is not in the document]
srbarrios: It seems so, I should have caught this before.
Anyway, we can iterate over it once we have something working. So for now, don't take this message into account.

testsuite/run_sets/github_validation/github_validation_init_clients.yml
szachovy: Moved to the top.
As given in my comment above https://github.com/uyuni-project/uyuni/pull/8446#discussion_r1553043951 bootstrapping proxy with minions attached is likely to take more efforts than initially assumed. I would divide it into three separate pull requests:
1. Creating proxy configuration and starting podman containers (this PR)
2. Bootstraping proxy with minions re-attached from server
3. `systemd is-active` related tests readjusted for Proxy in CI

What do you think about it?

testsuite/run_sets/github_validation/github_validation_init_clients.yml
szachovy: the same https://github.com/uyuni-project/uyuni/pull/8446#discussion_r1553578751

srbarrios:
Converting this as draft again, as no one is working on this for now.
If we find ourselves without interest on finishing it on next months, then I will close the PR.

PR #8450, opened 2024-03-13
QE: On CI, if a channel is not synchronized in the expected time, raise an error

srbarrios:
## What does this PR change?

On CI, if a channel is not synchronized in the expected time, raise an error, so we identify on time the root cause, instead of having it hidden and cascaded in other test cases.
But in BV, let's keep the behavior untouched and let's keep running the pipeline.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Ports:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -420,7 +420,11 @@ When(/^I wait until the channel "([^"]*)" has been synced$/) do |channel|
       sleep checking_rate
     end
   rescue StandardError => e
-    log e.message # It might be that the MU repository is wrong, but we want to continue in any case
+    log e.message
+    unless $build_validation
+      # It might be that the MU repository is wrong, but we want to continue in any case
+      raise ScriptError, "This channel was not fully synced: #{channel}"
+    end
   end
 end
 
@@ -453,7 +457,11 @@ When(/^I wait until all synchronized channels for "([^"]*)" have finished$/) do 
       sleep checking_rate
     end
   rescue StandardError => e
-    log e.message # It might be that the MU repository is wrong, but we want to continue in any case
+    log e.message
+    unless $build_validation
+      # It might be that the MU repository is wrong, but we want to continue in any case
+      raise ScriptError, "These channels were not fully synced:\n #{channels_to_wait}"
+    end
   end
 end
 


testsuite/features/step_definitions/command_steps.rb
```diff
@@ -420,7 +420,11 @@
       sleep checking_rate
     end
   rescue StandardError => e
-    log e.message # It might be that the MU repository is wrong, but we want to continue in any case
+    log e.message
+    # It might be that the MU repository is wrong, but we want to continue in any case
```
Bischoff, line 424: move the comment after the "unless" for clarity

Bischoff (changes requested):
small change request

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -453,7 +457,11 @@
...
```
Bischoff, line 461: same, this remark applies only in the BV case

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -420,7 +420,11 @@
...
+    unless $build_validation
+      raise ScriptError, "This channel was not fully synced: #{channel}"
```
szachovy, line 426: Because it synchronizes channels sequentially, it would be a great enhancement if we will be able to retrieve what channels were not synced. For example we can read from remote source that it consist of 5 channels of that names, if only 3 of them exist in `.../rhn/.../reposync.log`. Can we call back channels that were not make it?

szachovy (commented):
Optional suggestions.

testsuite/features/step_definitions/command_steps.rb
srbarrios: Done

testsuite/features/step_definitions/command_steps.rb
srbarrios: Done

testsuite/features/step_definitions/command_steps.rb
ktsamis: Isn't that what the step does? It checks for one channel at a time to be synced and if not finished in the time specified in the TIMEOUT_BY_CHANNEL_NAME for that specific channel, it raised an exception with the name of the channel

ktsamis (approved):
Looks good, since we log the message anyway we just need to be careful in the MU that we don't have failed channels and we check the messages.

testsuite/features/step_definitions/command_steps.rb
srbarrios: At least this step has as a scope waiting for a particular channel. For that reason, I would not add extra responsibilities to it.

But your suggestion make more sense in the other changed step `When(/^I wait until all synchronized channels for "([^"]*)" have finished$/) do |os_product_version|` where we waiting for a set of channels.

What we do in that different step, is to rely into a function that checks by the existence of a file that must be placed once the channel is fully synchronized, and we  iterate over all these channels waiting for them to be synchronized (in a order that can't be determined) until all of them are completed.

Bischoff (approved):
as a matter of fact the comment is not better off there, my bad

the problem is that the comment comments something that is not done in code... hard to solve

testsuite/features/step_definitions/command_steps.rb
szachovy: Ok, it makes sense.

szachovy (approved):

PR #8516, opened 2024-03-26
Collect enabled AppStream modules in package profile update

wweellddeerr:
## What does this PR change?

- Add table to store enabled appstream modules for a Server
- Introduce salt module for dealing with appstreams
- Change profile update sls for including enabled appstreams using the new salt module
- Populate the new table based on the results of profile update

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/23771
Fixes https://github.com/SUSE/spacewalk/issues/23766

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


cbbayburt (changes requested):
One thing first comes to mind (I should've done it in PoC): I think it's a good idea to change the word "Module" (like in class names `ServerModule` or variable name `module`, etc.) to "Appstream" and prefer this word throughout the code. The word "module" is too ambiguous and can be confusing for readers.

Also one nitpick: some files don't have newlines at the end.

Other than that it looks all good to me :+1: 
Additional thanks for the unit tests :)

cbbayburt (approved):
LGTM :+1: thanks Welder.

Please remember to set the target branch to `appstreams` as well.

Etheryte (approved):
Nothing to review for frontend.

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -0,0 +1,153 @@
...
+                nsvca_info_list.append(nsvca_info)
+
+        return nsvca_info_list
+
+    else:
+        log.error(f"Error running DNF command: {result.stderr}")
+        return []
+
+def _is_dnf_available():
+    return shutil.which("dnf") is not None
```
meaksh, line 105: I would consider refactoring this into a `__virtual__` function, like we do in other Salt modules: https://github.com/uyuni-project/uyuni/blob/f3163fc212e11d0c92cf79f0492923124d809464/susemanager-utils/susemanager-sls/src/modules/mgrnet.py#L28

If the `__virtual__` function, which is evaluated by Salt while loading the module, is not returning `True`, then the module won't be available, and you would get an error message

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -0,0 +1,153 @@
...
+
+def get_enabled_modules():
+    if not _is_dnf_available():
+        return []
```
meaksh, line 109: This won't be necessary if we introduce the `__virtual__` function.

susemanager-utils/susemanager-sls/src/states/appstreams.py
```diff
@@ -0,0 +1,50 @@
+def change(name, appstreams_to_enable, appstreams_to_disable):
```
meaksh, line 1: Hmm, there are few things here I would consider:

1) Since this is an "state" module, the function names should not be imperative, but declarative. So, I would suggested maybe something like:

```suggestion
def configured(name, appstreams_to_enable, appstreams_to_disable):
```

So, an SLS can be written like:
```yaml
configure_appstreams_repos:
  appstream.configured:
    - enabled_appstreams: ["foobar"]
    - disabled_appstreams: ["deadbeef"]
```

2) You have to make sure this module can run in "test" mode, without doing any actual change on the system. More info: https://docs.saltproject.io/en/latest/ref/states/writing.html#test-state

3) Please, add some documentation as header for this module.

4) A `__virtual__` function would be nice here, to check the necessary execution module is available, like we do here: https://github.com/uyuni-project/uyuni/blob/f3163fc212e11d0c92cf79f0492923124d809464/susemanager-utils/susemanager-sls/src/modules/virt_utils.py#L30-L39

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -0,0 +1,153 @@
...
+    if result.returncode == 0:
+        # Split the output into lines
+        module_info_output = result.stdout.splitlines()
+
+        # Parse NSVCA information for each module
+        nsvca_info_list = []
+        current_module_info = []
+        for line in module_info_output:
+            # Check if the line starts with "Name" to identify the beginning of a new module info
+            if line.startswith("Name"):
```
meaksh, line 82: Since we are parsing the output from a `dnf` execution, I wonder if `Name` wouldn't be translated in case my system has language different than English.

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -0,0 +1,153 @@
...
+
+def get_enabled_modules():
+    if not _is_dnf_available():
+        return []
+
+    enabled_module_names = _get_enabled_module_names()
+    if enabled_module_names:
+        return _get_module_info(enabled_module_names)
+    else:
+        return []
```
meaksh, line 115: Just an small suggestion/nitpick:

```suggestion
    return _get_module_info(enabled_module_names) if enabled_module_names else []
```

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -0,0 +1,153 @@
...
+        Tuple: (result, comment, changes)
+    '''
+    cmd = ["dnf", "module", "enable", "-y"] + appstreams
+    try:
+        result = subprocess.run(cmd, check=False, capture_output=True)
+        if result.returncode == 0:
+            return True, "AppStreams enabled", {"enabled": appstreams}
+        else:
+            return False, result.stderr.decode("utf-8").strip(), {}
+    except Exception as e:
```
meaksh, line 133: Can we avoid broad exception catching? https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/broad-exception-caught.html

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -0,0 +1,153 @@
...
+        Tuple: (result, comment, changes)
+    '''
+    cmd = ["dnf", "module", "disable", "-y"] + appstreams
+    try:
+        result = subprocess.run(cmd, check=False, capture_output=True)
+        if result.returncode == 0:
+            return True, "AppStreams disabled", {"disabled": appstreams}
+        else:
+            return False, result.stderr.decode("utf-8").strip(), {}
+    except Exception as e:
```
meaksh, line 152: Same avoid broad exceptions catching

meaksh (changes requested):
Thanks for this implementation! I just added some comments that should improve the Python / Salt part.

susemanager-utils/susemanager-sls/src/modules/appstreams.py
cbbayburt: Uuf this is a problem :see_no_evil:

cbbayburt: We'll need to rethink the parsing code

agraul: Salt is already parsing `dnf` output in https://github.com/saltstack/salt/blob/master/salt/modules/yumpkg.py#L201-L205, in principle parsing the output should be possible. We can also control the locale settings when calling `dnf` if needed. It could be that `systemd-run --scope` has such an effect (that's how pkg managers are called in the *pkg modules).

susemanager-utils/susemanager-sls/src/states/appstreams.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    return (
+        __virtualname__
+        if ("appstreams.enable" in __salt__
+            and "appstreams.disable" in __salt__
+            and "appstreams.get_enabled_modules" in __salt__)
+        else (False, "Module appstreams: appstreams module can't be loaded")
+    )
+
+def configured(name, appstreams_to_enable, appstreams_to_disable):
```
agraul, line 31: Like `configured` using passive voice, the arguments should also be named with a passive voice.
```
def configured(name, enabled_appstreams, disabled_appstreams):
```

Or, without repeating ourselves (it's `appstreams.configured` after all):
```
def configured(name, enabled, disabled):
```

susemanager-utils/susemanager-sls/src/states/appstreams.py
```diff
@@ -0,0 +1,101 @@
...
+    else:
+        ret["comment"] = "No changes would be made to AppStreams."
+    return ret
+
+def _do_operation(name, appstreams, operation):
+    ret = {"name": name, "result": False, "changes": {}, "comment": ""}
+    success, comment, changes =  __salt__[operation](appstreams)
+    ret["result"] = success
+    ret["comment"] = comment
+    ret["changes"] = changes
```
agraul, line 99: `ret` is filled but then unused

susemanager-utils/susemanager-sls/src/states/appstreams.py
```diff
@@ -0,0 +1,101 @@
...
+        return _test_mode(ret, appstreams_to_enable, appstreams_to_disable)
+
+    if appstreams_to_disable:
+        disabled_success, disabled_comment, disabled_changes = _do_operation(name, appstreams_to_disable, "appstreams.disable")
+        if not disabled_success:
+            ret["result"] = False
+            ret["comment"] = disabled_comment
+            return ret
+
+        ret["changes"]["ret"]["disabled"] = disabled_changes["disabled"]
```
agraul, line 56: If I follow the code correctly, this information can be wrong. When L63 re-enables a module that was disabled here, will the module be listed in both "disabled" and "enabled"?

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -0,0 +1,150 @@
...
+
+    else:
+        log.error(f"Error running DNF command: {result.stderr}")
+        return []
+
+def get_enabled_modules():
+    enabled_module_names = _get_enabled_module_names()
+    return _get_module_info(enabled_module_names) if enabled_module_names else []
+
+def enable(appstreams):
```
agraul, line 120: It would be nice if this would take a string or a list of strings, so that a user can run `salt * appstreams.enable moduleXYZ`

susemanager-utils/susemanager-sls/src/states/appstreams.py
cbbayburt: Actually, there is a conceptual issue: the method is supposed to update the selection of appstreams. So the `enabled` and `disabled` parameters do not reflect the state, but the changes to be made. So I think this method should be better written as an execution module.

cbbayburt (approved):

PR #8527, opened 2024-03-27
Translate server fqdn to proxy fqdn for pxe boot menus

aaannz:
## What does this PR change?

To avoid clients behind proxy connecting directly to the server, we need to translate server fqdn to the proxy one. This PR adds this translation.
Also fixes preselection of correct entry for saltboot branch.

This PR does 2 translations/lookout simultaneously. It, as before looks for saltboot entries and filtering out the rest. Parallel to it it translate server fqdn to proxy fqdn for other entries.
At the end depending if saltboot entry was matched or not it presents either saltboot profile or translated cobbler profiles.

Tested this on grub 30k profiles and there is no noticeable delay. tftp request for filtered saltboot output was in approximately 0.3s.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24023
Port(s): https://github.com/SUSE/spacewalk/pull/24039

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/proxy-tftpd-image/proxy-tftpd-image.changes.oholecek.cobbler_translations
+++ b/containers/proxy-tftpd-image/proxy-tftpd-image.changes.oholecek.cobbler_translations
@@ -1,0 +1,1 @@
+- translate server fqdn to proxy fqdn for pxe boot menus
--- a/containers/proxy-tftpd-image/tftp_wrapper.py
+++ b/containers/proxy-tftpd-image/tftp_wrapper.py
@@ -78,9 +78,10 @@ class HttpResponseDataFiltered(HttpResponseData):
     Saltboot entries have MASTER= option, we want only those mathing our proxy
     """
 
-    def __init__(self, url, capath, fqdn):
+    def __init__(self, url, capath, proxyFqdn, serverFqdn):
         # request file by url and store it
-        self._proxyFqdn = fqdn
+        self._proxyFqdn = proxyFqdn
+        self._serverFqdn = serverFqdn
         self._request = requests.get(url, stream=True, verify=capath)
         if self._request.status_code == 404:
             raise FileNotFoundError()
@@ -97,9 +98,11 @@ class HttpResponseDataFiltered(HttpResponseData):
 
 class HttpResponseDataFilteredPXE(HttpResponseDataFiltered):
     def contentFilter(self):
-        new_content = ""
+        saltboot_content = ""
+        cobbler_content = ""
         have_entry = False
         entry = ""
+        entry_name = ""
         in_entry = False
         for line in self._content.decode("utf-8").splitlines():
             if in_entry:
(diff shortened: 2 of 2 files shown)


containers/proxy-tftpd-image/proxy-tftpd-image.changes.oholecek.cobbler_translations
```diff
@@ -0,0 +1 @@
+- translate server fqdn to proxy fqdn for pxe boot menus
```
deneb-alpha, line 1:
```suggestion
- Translate server FQDN to proxy FQDN for PXE boot menus
```

containers/proxy-tftpd-image/tftp_wrapper.py
```diff
@@ -260,6 +304,12 @@ def get_arguments():
         default="localhost",
         help="Hostname where to forward HTTP requests",
     )
+    parser.add_argument(
+        "--serverFqdn",
+        type=str,
+        default="localhost",
+        help="Hostname of the server for cobbler filtering",
```
deneb-alpha, line 311:
```suggestion
        help="Hostname of the server for Cobbler filtering",
```

containers/proxy-tftpd-image/tftp_wrapper.py
```diff
@@ -97,9 +98,11 @@ def contentFilter(self):
 
 class HttpResponseDataFilteredPXE(HttpResponseDataFiltered):
     def contentFilter(self):
```
szachovy, line 100: Instead of initializing empty variables and appending content, I would suggest considering templating (if possible) in a style like this:

```
saltboot_template = """
...
  LABEL {{ entry_name }}
    MENU DEFAULT
    ONTIMEOUT {{ entry_name }}
...
"""
```

You can then use a `render_template` function to render the given content.

Alternatively, you could consider some form of object representation where the content would be parsed by a common engine.

This approach would be beneficial because in the current example, it's hard to determine the resulting content, especially with the presence of empty prefix spaces, special characters, and for loops over `splitlines()`.

containers/proxy-tftpd-image/tftp_wrapper.py
```diff
@@ -295,6 +345,7 @@ def main():
         with open(args.configFile) as source:
             config = yaml.safe_load(source)
             args.proxyFqdn = config["proxy_fqdn"]
+            args.serverFqdn = config["server"]
```
szachovy, line 348: ?
```suggestion
            args.serverFqdn = config["server_fqdn"]
```

In the same style as proxy above.

containers/proxy-tftpd-image/tftp_wrapper.py
```diff
@@ -307,8 +358,9 @@ def main():
         exit(1)
 
     logging.info("Starting TFTP proxy:")
-    logging.info(f"httpHost: {args.httpHost}")
-    logging.info(f"proxyFqdn: {args.proxyFqdn}")
+    logging.info(f"forward host: {args.httpHost}")
```
szachovy, line 361: From log it would be hard to guess what it is for the person who did not implemented that. I assume it is _Hostname where to forward HTTP requests_ for help above. I would suggest using:
```suggestion
    logging.info(f"HTTP API Endpoint Host: {args.httpHost}")
```

containers/proxy-tftpd-image/tftp_wrapper.py
```diff
@@ -78,9 +78,10 @@ class HttpResponseDataFiltered(HttpResponseData):
     Saltboot entries have MASTER= option, we want only those mathing our proxy
     """
 
-    def __init__(self, url, capath, fqdn):
+    def __init__(self, url, capath, proxyFqdn, serverFqdn):
```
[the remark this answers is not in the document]
agraul: snake case is preferred according to our style guide. These modules are not checked with our automation yet, once we enable it, we'll change the naming style. Syntax-only changes should be kept separate from semantic changes and go in a different PR.

vzhestkov (changes requested):
@aaannz as @agraul already pointed I would also prefer to prevent mixing snake a camel styles of naming, so could you please turn it to the snake stile?
I also don't like it in the parameters like `--serverFqdn`, but seems `--proxyFqdn` is already used, so I'm not sure maybe it could be also turned to `--server-fqdn` and `--proxy-fqdn`, but with having the alias for the case if `--proxyFqdn` is already used by something external.

agraul:
> @aaannz as @agraul already pointed I would also prefer to prevent mixing snake a camel styles of naming, so could you please turn it to the snake stile? 

Actually I meant to not change it now. The module was written before we added the style guide and needs multiple changes to get it up to date. I would rather do that in a different PR (where we also enable the python checkstyle on this path). Right now the module uses camelCase. Mixing it with snake_case now is more confusing than leaving the camelCase.

containers/proxy-tftpd-image/tftp_wrapper.py
aaannz: Can't do that unless we want to make existing configs unusable to users and also change how we generate proxy config in the java.

aaannz:
Since change to snake case will not happen in this PR, I resolved those conversations.

aaannz:
Anything more to do to unblock this?

Just to get where this is used. This script is not part of any package. The only place it is used is in the tftp container on the proxy.
With the next PR for the snake casing, I will most probably completely remove command line options, as we do not use it and only config file is used.

containers/proxy-tftpd-image/tftp_wrapper.py
szachovy: ok

vzhestkov (approved):
Approving as it's ok to keep such names for now.

PR #8559, opened 2024-04-04
Ci fix locale formula and inter server sync tests

jordimassaguerpla:
test


jordimassaguerpla:
`salt.exceptions.CommandExecutionError: timedatectl failed: System has not been booted with systemd as init system (PID 1). Can't operate. `

Our sle minion has not been booted with systemd as init systemd, but sshd.

PR #8560, opened 2024-04-04
Avoid hairpin problem with rhnpush

Bischoff:
## What does this PR change?

Don't use `--server` parameter of `rhnpush` as it triggers the hairpin problem.

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/21384

Port(s):
 * 4.3: https://github.com/SUSE/spacewalk/pull/24064

Bug: https://bugzilla.suse.com/show_bug.cgi?id=1222325

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -510,10 +510,10 @@ Given(/^metadata generation finished for "([^"]*)"$/) do |channel|
   get_target('server').run_until_ok("ls /var/cache/rhn/repodata/#{channel}/*updateinfo.xml.gz")
 end
 
-When(/^I push package "([^"]*)" into "([^"]*)" channel$/) do |arg1, arg2|
-  srvurl = "https://#{get_target('server').full_hostname}/APP"
-  command = "rhnpush --server=#{srvurl} -u admin -p admin --nosig -c #{arg2} #{arg1} "
+When(/^I push package "([^"]*)" into "([^"]*)" channel$/) do |package, channel|
+  command = "rhnpush -u admin -p admin --nosig -c #{channel} #{package}"
   get_target('server').run(command, timeout: 500)
+  # TODO: instead of next line, wait for package to appear inside /var/spacewalk/packages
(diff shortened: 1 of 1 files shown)


testsuite/features/step_definitions/common_steps.rb
```diff
@@ -510,10 +510,10 @@
   get_target('server').run_until_ok("ls /var/cache/rhn/repodata/#{channel}/*updateinfo.xml.gz")
 end
 
-When(/^I push package "([^"]*)" into "([^"]*)" channel$/) do |arg1, arg2|
-  srvurl = "https://#{get_target('server').full_hostname}/APP"
-  command = "rhnpush --server=#{srvurl} -u admin -p admin --nosig -c #{arg2} #{arg1} "
+When(/^I push package "([^"]*)" into "([^"]*)" channel$/) do |package, channel|
+  command = "rhnpush -u admin -p admin --nosig -c #{channel} #{package}"
   get_target('server').run(command, timeout: 500)
+  # TODO: instead of next line, wait for package to appear inside /var/spacewalk/packages
```
szachovy, line 516: What does it mean, refering to `ls`?

Bischoff: Yes, the "ls" command on next line does basically nothing.

I did not want to remove it though, in case it had the implicit effect to wait until the operation is done.

szachovy: Ok, instead of listing all the directory it is probably more suitable to search for specific entry in that directory, something like `find /var/spacewalk/packages -name #{package}`.
Feel free to close this comment anytime you want.

szachovy (approved):

testsuite/features/step_definitions/common_steps.rb
Bischoff: That's exactly what I was trying to suggest in my TODO comment: wait for the package to appear. I did not specify "with a find" in case people have some other method in mind, but that was my idea.

Closing this conversation.

srbarrios (approved):

PR #8571, opened 2024-04-08
QE: Build Host entitlement adds some extra minutes to apply the salt high-state

srbarrios:
## What does this PR change?

Build Host entitlement adds some extra minutes to apply the salt high-state.
So we need to wait some more minutes until the onboarding is completed. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/step_definitions/setup_steps.rb
+++ b/testsuite/features/step_definitions/setup_steps.rb
@@ -205,16 +205,20 @@ When(/^I wait at most (\d+) seconds until I see the name of "([^"]*)", refreshin
   end
 end
 
-When(/^I wait until onboarding is completed for "([^"]*)"$/) do |host|
+When(/^I wait at most (\d+) seconds until onboarding is completed for "([^"]*)"$/) do |seconds, host|
   steps %(
     When I follow the left menu "Systems > System List > All"
     And I wait until I see the name of "#{host}", refreshing the page
     And I follow this "#{host}" link
     And I wait until I see "System Status" text
-    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Apply states" is completed
-    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Hardware List Refresh" is completed
-    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Package List Refresh" is completed
+    And I wait 180 seconds until the event is picked up and #{seconds} seconds until the event "Apply states" is completed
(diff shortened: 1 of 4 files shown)


testsuite/features/step_definitions/setup_steps.rb
```diff
@@ -205,18 +205,22 @@
...
     When I follow the left menu "Systems > System List > All"
     And I wait until I see the name of "#{host}", refreshing the page
     And I follow this "#{host}" link
     And I wait until I see "System Status" text
-    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Apply states" is completed
-    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Hardware List Refresh" is completed
-    And I wait 180 seconds until the event is picked up and #{DEFAULT_TIMEOUT} seconds until the event "Package List Refresh" is completed
+    And I wait 180 seconds until the event is picked up and #{seconds} seconds until the event "Apply states" is completed
+    And I wait 180 seconds until the event is picked up and #{seconds} seconds until the event "Hardware List Refresh" is completed
+    And I wait 180 seconds until the event is picked up and #{seconds} seconds until the event "Package List Refresh" is completed
```
nodeg, line 216: I like to have this additional step for defining the time. However, one step with e.g. 2 input values (`default` for the default timeout and a `500` (an integer) for the timeout you would like to have) would also be a good solution.

nodeg (approved):
I guess in this special case, it is ok to increase the timeout. Thank you for fixing this, Oscar!

testsuite/features/step_definitions/setup_steps.rb
Bischoff: I prefer like this (simplicity).

Bischoff (approved):
ok

testsuite/features/step_definitions/setup_steps.rb
srbarrios: My main concern when I wrote it, it was to don't need to change the text in the steps that are not affected. But I guess we could also solve this with a different regex.

I think that's more a matter of taste and we already have few steps following a similar implementation.

So, for now, I prefer to keep it like that.

nodeg: Sure. As I said, I also like to have 2 steps.

NamelessOne91 (approved):

PR #8580, opened 2024-04-09
Add "expect" package to proxy

Bischoff:
## What does this PR change?

Temporary PR so `expect` package is available on the SLE micro based containerized proxy. This package is needed for Cobbler and Retail tests.

Might be removed in favour of a sumaform solution based on combustion.

## Links

No ports, 4.3 does not use sle micro containerized proxy.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/reposync/allcli_update_activationkeys.feature
+++ b/testsuite/features/reposync/allcli_update_activationkeys.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2023 SUSE LLC
+# Copyright (c) 2022-2024 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 @skip_if_github_validation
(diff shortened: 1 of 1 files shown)


testsuite/features/reposync/allcli_update_activationkeys.feature
```diff
@@ -127,6 +127,13 @@ Feature: Update activation keys
     And I check "SUSE-Manager-Proxy-5.0-Updates for x86_64"
     And I click on "Update Activation Key"
     Then I should see a "Activation key Proxy Key x86_64 has been modified" text
+    # TODO: remove when we use combustion
+    # 'expect' is needed for Cobbler and Retail tests
```
srbarrios, line 131: As temporary solution, lgtm.
But please, could you already prepare a PR to include this package in our [combustion config](https://github.com/uyuni-project/sumaform/blob/master/backend_modules/libvirt/host/combustion)?

srbarrios (approved):

testsuite/features/reposync/allcli_update_activationkeys.feature
Bischoff: I can, but how would i test it works?

BTW, I did not even test this code yet. It could be it does not work (missing repos, etc.). If that's the case i need to create a first version for cloud-init ::cry: .

Bischoff:
ok let's forget about this one, the idea could not work in the case of proxies as foreign systems...

PR #8614, opened 2024-04-18
Coco attestation container - Part 2

mackdk:
## What does this PR change?

This PR introduces the following improvements to the Confidential computing attestation microservice:

- Allow compilation with both java 11 and 17, depending on the jdk used
- Refactor the `module` package name to be consistent (before it was wrongly called `modules` in some places)
- Improve the pom.xml definitions to do not repeat dependencies version numbers
- Review modularity to facilitate tagging with tito during release

With the new module definition, during tagging all version will be updated automatically by tito. In order to tag correctly the order must be:

1) uyuni-java-parent
2) uyuni-java-common
3) coco-attestation

Since tagging an upstream module changes the pom.xml downstream, those projects must be tagged even if they don't have any other changes a part from the pom.xml updated.

The downside of the new structure is that, in order to build only coco-attestation locally (i.e. running `mvn package` from the `microservices/coco-attestation` directory), the parent and common module must be installed in the local maven repository (i.e. run `mvn install` for both). Otherwise it is possible to run the maven commands from the directory `microservices/uyuni-java-parent` to build everything top-down. When using an IDE, though, it should just work fine out of the box.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


microservices/uyuni-java-parent/pom.xml
```diff
@@ -28,6 +28,35 @@
...
+    </modules>
+
+    <!-- Profiles based on the jdk Maven is running on.
+       ~ to ensure the compilation happens with the proper release target
+      -->
+    <profiles>
+        <profile>
+            <id>java-11</id>
+            <activation>
+                <jdk>[11,17)</jdk>
```
mcalmer, line 43: Are the brackets correct?

microservices/uyuni-java-parent/pom.xml
```diff
@@ -28,6 +28,35 @@
...
+            </activation>
+            <properties>
+                <maven.compiler.release>11</maven.compiler.release>
+            </properties>
+        </profile>
+        <profile>
+            <id>java-17</id>
+            <activation>
+                <jdk>[17,)</jdk>
```
mcalmer, line 52: Are the brackets correct?

mcalmer (commented):
I have no idea about pom files,  but there are two things which looks wrong

microservices/uyuni-java-parent/pom.xml
mackdk: Yes, that maven way to define version ranges (see https://maven.apache.org/guides/introduction/introduction-to-profiles.html#implicit-profile-activation). This basically defines that for JDK 11 and up to 17 excluded we use the release = 11 flag, otherwise we use release = 17

mcalmer: Ok, funny :-)

mcalmer (approved):

lucidd (approved):

deneb-alpha (approved):
Thanks @mackdk it should  be  ok now.   I hope  I didn't miss  anything :)

PR #8636, opened 2024-04-22
QE: Improve apply patch in BV

srbarrios:
## What does this PR change?

Sometimes we don't have any available non-critical patches.
Let's perform a different test strategy, let's don't filter by the kind of patch but let's assure the patch doesn't require a reboot, if so, we select the next patch.

Note: If we have two consecutive patches requiring reboot, bad luck 🤣

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- Cucumber tests were changed

- [x] **DONE**

## Links

Ports:
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/build_validation/smoke_tests/smoke_tests.template
+++ b/testsuite/features/build_validation/smoke_tests/smoke_tests.template
@@ -40,9 +40,7 @@ Feature: Smoke tests for <client>
     And I wait until I see "Upgrade Packages" text
     And I follow "Patches" in the content area
     When I wait until I see "Relevant Patches" text
-    And I select "Non-Critical" from "type"
-    And I click on "Show"
-    When I check the first patch in the list
+    When I check the first patch in the list, that does not require a reboot
     And I click on "Apply Patches"
     And I click on "Confirm"
     Then I should see a "1 patch update has been scheduled for" text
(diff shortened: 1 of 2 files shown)


nodeg (approved):
I like the idea.

ktsamis (approved):

NamelessOne91:
It would be good indeed to have this fixed.

Would it be possible to have something like the following ?

1) Check if there are patches to apply - I've seen the step fail quite a lot of times just 'cause there are no patches available.

2) Look for, and use, the first patch that does not require a reboot.

3) Fallback to use the first available patch or just fail

srbarrios:
> It would be good indeed to have this fixed.
> 
> Would it be possible to have something like the following ?
> 
> 1) Check if there are patches to apply - I've seen the step fail quite a lot of times just 'cause there are no patches available.
> 
> 2) Look for, and use, the first patch that does not require a reboot.
> 
> 3) Fallback to use the first available patch or just fail

We expect some patch to be available for any kind of minion, if none, I prefer to show a red, as it could be a symptom of a failure in the software package update process or a problem with subscribed channels.

NamelessOne91 (approved):

PR #8665, opened 2024-04-25
Fix skipped status text color

bisht-richa:
## What does this PR change?
- Fixed skipped tasks and read messages color contrast in table

**add description**

## GUI diff

No difference.

**Before:**

Screenshot 2024-04-25 at 16 08 39

Screenshot 2024-04-25 at 16 11 27

**After:**

Screenshot 2024-04-25 at 16 09 03

Screenshot 2024-04-25 at 16 09 33

- [x] **DONE**

## Documentation
- No documentation needed

- [ ] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): #[#24136](https://github.com/SUSE/spacewalk/issues/24136)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/web/html/src/branding/css/susemanager/components/text.less
+++ b/web/html/src/branding/css/susemanager/components/text.less
@@ -1,7 +1,0 @@
-// These are all leftovers from the old theme, hopefully we can drop most of this over time
-.text-muted {
-  &, &:hover, &:focus {
-    color: @eos-bc-gray-100;
-  }
-}
-
--- a/web/html/src/branding/css/susemanager/components/text.less
+++ b/web/html/src/branding/css/susemanager/components/text.less
@@ -1,0 +1,7 @@
+// These are all leftovers from the old theme, hopefully we can drop most of this over time
+.text-muted {
+  &, &:hover, &:focus {
+    color: @eos-bc-gray-700;
+  }
+}
+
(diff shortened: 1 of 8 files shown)


rjmateus:
Maybe instead of having the text in a different color as before (which makes it hard to read), maybe we have have a full row with a different (maybe gray) color?

bisht-richa:
> While this fixes the contrast issue, the two types of entries are now practically indistinguishable, e.g. I can't tell that there's any difference between the rows in the after image. Is this intentional? If so, we might want to use some kind of a different indicator for these statuses.

We already have `open` and 'closed` envelop icons to show the two types of entries on the Notification page. And in another table, we have a clear indicator of `Skipped` status. 
I don't know if we need more indicators here.

Etheryte:
I don't have a strong opinion on how we indicate the status, however if we want to remove the muted text visual here, it would be more appropriate to remove the class from the elements, rather than override the style for it. Right now we override a style back to the base it already was.

web/html/src/branding/css/susemanager/components/text.less
```diff
@@ -0,0 +1,7 @@
+// These are all leftovers from the old theme, hopefully we can drop most of this over time
+.text-muted {
```
Etheryte, line 2: Do we still need this if we're removing the class from the DOM? Same for the scss one.

bisht-richa: Yes, we still have use cases where we are using this class from scss file. But I will remove this text.less file I think we don't need it.

Etheryte (approved):

PR #8720, opened 2024-05-14
Retail - fix proxy name

Bischoff:
## What does this PR change?

This PR sets the name of the proxy to its name on the private network in prevision for pxeboot via Cobbler or Retail.

## Links

No ports.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/retail/proxy_container_branch_network.feature
+++ b/testsuite/features/build_validation/retail/proxy_container_branch_network.feature
@@ -9,9 +9,12 @@ Feature: Prepare the containerized branch server for PXE booting
   As the system administrator
   I prepare the branch network in containerized setup
 
-  Scenario: Activate the branch network on the proxy
-    When I connect the second interface of the proxy to the private network
-    And I restart all proxy containers to let them pick new network configuration
+  Scenario: Adapt the proxy for Retail
+    When I rename the proxy for Retail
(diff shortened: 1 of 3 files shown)


vandabarata (approved):

NamelessOne91 (approved):

Bischoff:
> This looks good, only one thing that I'm not sure about, if we use the renamed restart step anywhere else.

No, only at those two places (CI + BV).

srbarrios:
Quick question @Bischoff , once we do this rename, the rest of clients registered to the proxy will stop working properly?
Should we have a clean-up scenario or that's unnecessary?

aaannz:
> Quick question @Bischoff , once we do this rename, the rest of clients registered to the proxy will stop working properly? Should we have a clean-up scenario or that's unnecessary?

They will continue working. What this does is it changes `ServerName` in the httpd container config and also to what address `tftp` container translates the queries.

Bischoff:
> Quick question @Bischoff , once we do this rename, the rest of clients registered to the proxy will stop working properly? Should we have a clean-up scenario or that's unnecessary?

To be very honest, I had exactly same fears.

From my first tests, the already registered clients continue to work, and it only has implications on the way Cobbler and Retail work.

I'm still not feeling 100 % reassured, but merging all the same.

PR #8782, opened 2024-05-20
run all command locally and use mgradm to mount the image

maximenoel8:
## What does this PR change?
Correct `I mount as "([^"]+)" the ISO from "([^"]+)" in the server, validating its checksum` step.

If I correctly understood the expected behavior, we want to copy the image on the server host and do the checksum on server host and then copy the distro using `mgradm distro copy`.

The current behavior was a mix of downloading the image and checksum in the container and doing the checksum on the host. I fix this behavior to have everything done on the server host and then push the image using mgradm command.

I also simplify some of the steps where we wanted to always run the command on local host but use different method depending on the server configuration. In conclusion I simplified the steps with `$is_container_provider`.

I also removed the direct calls of `run_local` and use `run` with the option `runs_in_container: false`. We are doing both in the code. We should probably stick to one method.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/support/file_management.rb
+++ b/testsuite/features/support/file_management.rb
@@ -139,7 +139,7 @@ def get_checksum_path(dir, original_file_name, file_url)
       next unless response.is_a?(Net::HTTPSuccess)
 
       checksum_url = base_url + name
-      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url}-o #{name}", runs_in_container: false, timeout: 10)
+      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url} -o #{name}", runs_in_container: false, timeout: 10)
       return "#{dir}/#{name}" if code.zero?
     end
 
--- a/testsuite/features/support/file_management.rb
+++ b/testsuite/features/support/file_management.rb
@@ -115,7 +115,7 @@ def get_checksum_path(dir, original_file_name, file_url)
   cmd = "ls -1 #{dir}"
   # when using a mirror, the checksum file should be present and in the same directory of the file
   if $mirror
-    output, _code = server.run_local(cmd)
+    output, _code = server.run(cmd, runs_in_container: false)
     files = output.split("\n")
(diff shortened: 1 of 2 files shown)


testsuite/features/support/file_management.rb
```diff
@@ -139,7 +139,7 @@ def get_checksum_path(dir, original_file_name, file_url)
       next unless response.is_a?(Net::HTTPSuccess)
 
       checksum_url = base_url + name
-      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url}-o #{name}", timeout: 10)
+      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url}-o #{name}",runs_in_container: false, timeout: 10)
```
mcalmer, line 142: Looks like here is the space missing

srbarrios, line 142:
```suggestion
      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url}-o #{name}", runs_in_container: false, timeout: 10)
```

srbarrios (approved):

testsuite/features/support/file_management.rb
```diff
@@ -139,7 +139,7 @@ def get_checksum_path(dir, original_file_name, file_url)
       next unless response.is_a?(Net::HTTPSuccess)
 
       checksum_url = base_url + name
-      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url}-o #{name}", timeout: 10)
+      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url}-o #{name}", runs_in_container: false, timeout: 10)
```
mcalmer, line 142:
```suggestion
      _output, code = server.run("cd #{dir} && curl --insecure #{checksum_url} -o #{name}", runs_in_container: false, timeout: 10)
```

PR #8798, opened 2024-05-23
Change order of Retail features

Bischoff:
## What does this PR change?

With containerized server and proxy, one cannot prepare that Saltboot group formula before he/she has built the images. It is not possible anymore to prepare the proxy, then build the images.

The reason why it worked in a non-containerized setup is that Retail used an image-sync state at the end.

## Links

No port, 5.0 only.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/run_sets/build_validation/build_validation_retail_proxy.yml
+++ b/testsuite/run_sets/build_validation/build_validation_retail_proxy.yml
@@ -1,5 +1,10 @@
 # This file describes the order of features in a Build Validation test suite run.
 
+## Build Kiwi images
+## (expected time of building is around 20 minutes per image)
+- features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature
+- features/build_validation/retail/sle15sp4_buildhost_build_kiwi_image.feature
+
 ## Prepare configuration for branch server
 - features/build_validation/retail/proxy_traditional_branch_network.feature
 - features/build_validation/retail/proxy_container_branch_network.feature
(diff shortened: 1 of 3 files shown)


nodeg (approved):
As far as I can see it, this is only a restructure of when we built the images, so LGTM.

testsuite/run_sets/build_validation/build_validation_retail_sle12.yml
```diff
@@ -5,9 +5,5 @@
 ## Prepare configuration for terminal
 - features/build_validation/retail/sle12sp5_terminal_prepare.feature
```
szachovy, line 6: Shouldn't it be before `features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature`?

testsuite/run_sets/build_validation/build_validation_retail_proxy.yml
```diff
@@ -1,5 +1,10 @@
 # This file describes the order of features in a Build Validation test suite run.
 
+## Build Kiwi images
+## (expected time of building is around 20 minutes per image)
+- features/build_validation/retail/sle12sp5_buildhost_build_kiwi_image.feature
+- features/build_validation/retail/sle15sp4_buildhost_build_kiwi_image.feature
+
```
srbarrios, line 7: In the future, we might want to have a separated yaml file for the building images stage.
I think it merits its own stage because it has a responsibility very well defined and separated from retail. And also, this is a strong dependency for the next features, so it's better that if one of these features fail, we don't run the features currently below these ones.

Bischoff: Beware of the risk to run them after the proxy if you do that. If we do that, we should explicit the order in the comments.

Also, would be a bit less convenient to use with rake.

Other than that, no strong feelings.

srbarrios: Yes. That's what I mean, have a new run_set `build_validation_retail_build_images.yml` and run them through Rake, in a new Jenkins pipeline stage, just before `build_validation_retail_proxy.yml`

PR #8833, opened 2024-05-28
Fix timezone issues with recurring action datetime pickers (bsc#1225196)

cbbayburt:
Fixes multiple issues about timezone handling in recurring action details page and datetime pickers:

- Fix the time displayed in the picker after selection (bsc#1225196)
- In recurring action details, show "Created At" in the correct timezone instead of UTC
- Append server timezone to "Execution Time" in recurring action details

Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1225196

## GUI diff

Before (**`Created At` time shown is actually in UTC, not in ACST**):

Screenshot from 2024-05-28 21-12-58

After (**`Created At` is shown correctly in ACST**):

Screenshot from 2024-05-28 21-13-16

## Documentation

## Test coverage
- No tests: already covered

## Links

https://bugzilla.suse.com/show_bug.cgi?id=1225196

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


web/html/src/manager/recurring/recurring-actions-details.tsx
```diff
@@ -143,6 +146,12 @@ class RecurringActionsDetails extends React.Component<RecurringActionsDetailsPro
     details.cronTimes.hour = details.cronTimes.hour.padStart(2, "0");
     details.cronTimes.minute = details.cronTimes.minute.padStart(2, "0");
 
+    // Creation time is always retrieved in server's timezone
+    // Parse the date and show it in user's timezone
+    const createdAt = localizedMoment(
+      Date.parse(`${details.createdAt} ${localizedMoment.serverTimeZoneAbbr}`)
```
Etheryte, line 152: This looks like we could maybe standardize this away inside `localizedMoment`? Although it would be preferable if we could make `details.createdAt` an ISO string rather than a humanized date string, then we could pass it directly to `localizedMoment`. Would that be possible or does it seem like too much work?

If the latter, we could extend the interface we provide in `localizedMoment`, maybe something along these lines?

```diff
  Object.defineProperties(moment, {
    // ...
+   fromServerString: {
+     get() {
+       return function fromServerString(input: string): moment.Moment {
+         return moment.tz(input, config.serverTimeZone);
+       };
+     },
+   },
  });
```

What do you think?

Etheryte:
When I create a new recurring action with a weekly schedule without touching any of the fields with this diff, I get an error about a null value when trying to look at the details of the created action.

Other than that this looks good, thanks a lot for taking the time to look into this.

web/html/src/manager/recurring/recurring-actions-details.tsx
cbbayburt: Yes, I actually wanted to open this to discussion. The ideal solution would be to send ISO dates from the backend, but we would need to do a whole sweep since this is how we get the DB dates all around. OTOH, the current format is consistent, so I guess your suggestion is the best way for now. I haven't been able to make it work without parsing it with JS `Date` so far, but I'll try it and see if it works.

Etheryte: Well the above internally uses `Date` anyway since we don't specify any format information. But if we know the server format ahead of time we could add that and make that check strict. That way we could use this for pages that use the current format, but use ISO for everything new we make and over time if we touch stuff, move that to ISO too.

Of course, if there was an easy way to swap everything to ISO, that would be preferred, but I suspect that would be a beast with many hidden surprises.

PR #8854, opened 2024-05-31
QE: tidy some channels management scenarios

NamelessOne91:
## What does this PR change?

This PR aims to:

- make clear when API calls are used
- replace the naming SLES minion with openSUSE minion in some instances
- (the meaningful part) make these scenarios less flaky without adding complexity and/or maintenance overload

The scope of this feature should be to:

- verify the state of the channels before the action for changing channels is scheduled
- verify the state remains consistent after an action has been scheduled
- verify the state changes as expected once the action/event is complete
- verify both the UI and  API return the correct state of the channels for a given system before and after the action

To do so counting the initial number of enabled channels we would need one of:

1.  global variable
2.  hardcoded number which has been previously verified
3. some logic that can account for the presence of BETA channels, channels that should be present for a given system, added fake channels ....

1 is better to be avoided as long as possible, 2 is the current approach which is inherently flaky and subject to change, 3 would make this more robust but add complexity for little to no gains. 
We can still use the hardcoded  number for the Fake channels because that is probably not subject to changes anytime soon and is something strictly related to the context of this feature.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24432 https://github.com/SUSE/spacewalk/issues/21342
Port(s): (4.3) https://github.com/SUSE/spacewalk/pull/24445

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/allcli_software_channels.feature
+++ b/testsuite/features/secondary/allcli_software_channels.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2018-2023 SUSE LLC
+# Copyright (c) 2018-2024 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 @scope_changing_software_channels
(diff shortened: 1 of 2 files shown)


srbarrios (approved):

nodeg (approved):
Nice improvement. If this fixes the need to count all the channels, this is very useful. Thank you, Paolo.

ktsamis (approved):
I also really like the context of the PR and the description, thanks :)

Bischoff (approved):
:green_heart:

maximenoel8 (approved):

PR #8892, opened 2024-06-08
Incremental sax parsing

waterflow80:
## Description
In this PR I tried to convert the `get_packages_metadata()` function to make it return a generator of batches (packages' batches) instead of doing the whole job, including importing, internally.

The approach was mainly proposed by @agraul [here](https://github.com/uyuni-project/uyuni/pull/8840#discussion_r1631044577).

This initial implementation is still not returning 100% correct results, and I'm still debugging it.

The main problem is that we're using SAX parsers, which is an event-driven xml parser, and the `handler` passed into it doesn't return any value before the whole xml file (`InputSource` data of the xml file) is completely parsed.

For example, when parsing an xml file like this (using SAX):
```python
def parse_metadata_file(self, md_file):
        with gzip.GzipFile(fileobj=md_file, mode="rb") as gzip_fd:
            parser = xml.sax.make_parser()
            parser.setContentHandler(self.handler)
            parser.setFeature(xml.sax.handler.feature_namespaces, True)
            input_source = InputSource()
            input_source.setByteStream(gzip_fd)
            parser.parse(input_source)  <------ WE SHOULD WAIT FOR THIS TO FINISH !
```
We cannot get any data from the parser before the `parser.parse(input_source)` returns (the whole file is parsed). So we cannot `yield` each batch of packages using this approach.

## Current Solution (not guaranteed)
To solve this issue, I tried to get use of the [xml.sax.xmlreader.IncrementalParser](https://docs.python.org/3.8/library/xml.sax.reader.html#xml.sax.xmlreader.IncrementalParser) interface, and its `send()` method, that allows to send the xml content in **chunks** to the parser and parse each chunk at a time, and the parser will update its `Handler`'s internal variables, for example the `self.batch`, each time.

And each time, we check if the size of the `batch` is equal to 20 (batch_size), if so return it. And the caller function will `yield` that batch.

### Problems
Identifying the best chunk size is critical and hard to do:
```python
chunk = gzip_fd.read(1024)
```
In our implementation we used 1024 bytes as a small value, because when we tried bigger values, we've noticed that the parser parsed more than one batch (more than 20 packages), so all the batches before the last one parsed are lost.

Currently the generator is returning the same batch multiple times, but I'm still debugging that. 

Another thing is that I'm using a method `feed()` on `self._parser.feed(data)` in the primary_parser.py, and it's surprisingly working, because  in the `XMLReader` docs, it doesn't have a method `feed()`.  (I mistakenly tried it and it worked).

## Alternative
As I mentioned in this [comment](https://github.com/uyuni-project/uyuni/pull/8840#discussion_r1631785972), there are multiple sources that recommended using `xml.etree.ElementTree.iterparse()` to achieve our desired behavior (parsing in batched and yield each batch in a generator like function). Apparently, it is passing the data into the handler, but we should verify that.

`xml.dom.pulldom` might also be a good way of doing it.

But this will require changing the current parser completely.

I'll start implementing those approaches in different PRs, and we can compare the different approaches and pick the best one.


rjmateus:
I have a question. Disclaimer: I'm not a python expert, so I may say something stupid :)
Instead of doing this with chunks, can we use the sax parser since it's already streaming the content (like in java)? 
We can play around with the `ContentHandler` class implementation and save data in an internal object, controlling it on with the start element and end element (and maybe also with start/end document). Save the data on memory, and then pass it to a processor in batches.
Would this work and make it simpler to implement and maintain?
@agraul @waterflow80  any comments?

waterflow80:
> I have a question. Disclaimer: I'm not a python expert, so I may say something stupid :) Instead of doing this with chunks, can we use the sax parser since it's already streaming the content (like in java)? We can play around with the `ContentHandler` class implementation and save data in an internal object, controlling it on with the start element and end element (and maybe also with start/end document). Save the data on memory, and then pass it to a processor in batches. Would this work and make it simpler to implement and maintain? @agraul @waterflow80 any comments?

If I well understood your point, In that case we'll have to parse the whole file and buffer the saved objects (`Packages`) into the memory before proceeding with the importing in batches (`pass it to a processor in batches`).

The problem is that the parsed metadata files can sometimes be very huge, and saving all the parsed packages' data in memory might not be very optimal.

So , the initial solution was to put counter in the `ContentHandler` class, and when it gets to the given `batch_size`, it calls the `import_package_batch(..)` function. See [implementation](https://github.com/uyuni-project/uyuni/blob/14f829f5a92f39f8cce11c15d674f5e80da9166a/python/lzreposync/src/lzreposync/primary_handler.py#L177-L182).

The problem with that implementation is that it is an in-bound approach and make the code too coupled, as mentioned in @agraul 's [comment](https://github.com/uyuni-project/uyuni/pull/8840#discussion_r1631044577).

So our goal is to make the parsing function return a **generator** of packages, and get data from that generator as needed, in a separate location. See [implementation](https://github.com/uyuni-project/uyuni/blob/ed4b07423a57697e4f5d52d16bc74d5caf570e08/python/lzreposync/src/lzreposync/primary_parser.py) (PR#8894).

**Note:** the approach in the current PR has been abandoned, due to the big complexity, and replaced with the one in PR[#8894](https://github.com/uyuni-project/uyuni/pull/8894/files#diff-f297e63ceea95c67a956eab9ad10d585f831a12d59878bc6f2e5a56089bbf818). In this PR (Incremental sax parsing) we were trying to change the `ContentHandler`'s behavior to make it `yield` packages (be a generator of packages), but it seems like it wasn't best fit for that use case.

I hope that answered your question :) And if not, please let me know.

rjmateus:
That answers my questions, thank you.
In implementation in [1], you don't need to save all the data in memory, as soon you save it to the database you can clear the handler cache, since data will be in memory, or an error will be returned.

I don't dislike this approach, but understand Alex's concerns and suggestions. I'm not sure if make sax parser be able to return a generator will be simpler, but we will find out soon :)


[1] https://github.com/uyuni-project/uyuni/blob/14f829f5a92f39f8cce11c15d674f5e80da9166a/python/lzreposync/src/lzreposync/primary_handler.py#L177-L182

agraul:
Closing, we prefer the approach in https://github.com/uyuni-project/uyuni/pull/8894/

PR #8893, opened 2024-06-09
fix hairping problem during ssh push with tunnel

mcalmer:
## What does this PR change?

When configuring ssh cascade for ssh push with tunnel via proxies, the last proxy command has a reverse port forwarding rule using the hostname of the proxy. That hostname is the name of the "host", but we land inside of the container which has its own network. 
To reach the given hostname, a connection is made to the host which redirect again inside of the container.

Also in case if the RPM based installed proxy, localhost should work as we want to redirect the local port always.
This should handle also the case of containerized SUSE Manager Server 5.0 with a plain SUSE Manager 4.3 proxy.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24510 https://github.com/SUSE/spacewalk/issues/24290

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
+++ b/java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
@@ -588,8 +588,7 @@ public class SaltSSHService {
     private Optional<String> remotePortForwarding(List<String> proxyPath,
                                                   String sshContactMethod) {
         if (ContactMethodUtil.SSH_PUSH_TUNNEL.equals(sshContactMethod)) {
-            return Optional.of(getSshPushRemotePort() + ":" +
-                    ConfigDefaults.get().getJavaHostname() + ":" + SSL_PORT);
+            return Optional.of(getSshPushRemotePort() + ":" + "localhost" + ":" + SSL_PORT);
         }
         return Optional.empty();
     }
--- a/java/spacewalk-java.changes.mcalmer.fix-sshpush-tunnel
+++ b/java/spacewalk-java.changes.mcalmer.fix-sshpush-tunnel
@@ -1,0 +1,1 @@
+- fix hairpin problem in ssh push with tunnel prevent bootstrapping (bsc#1226098, bsc#1223970)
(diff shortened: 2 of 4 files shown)


srbarrios (commented):
LGTM, if you tested and fixed the issue.
And about your last paragrah in the PR description, just remember that for now we also support both containerized and RPM flavour for both server/proxy in Uyuni, but from what I understood, it should be fine.

srbarrios (approved):

mcalmer:
> LGTM, if you tested and fixed the issue. And about your last paragrah in the PR description, just remember that for now we also support both containerized and RPM flavour for both server/proxy in Uyuni, but from what I understood, it should be fine.

Yes, I hope, but I would like to let other people review my assumption that it also work without containerized proxy.

Bischoff (approved):
thank you

cbosdo (approved):

java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
```diff
@@ -588,8 +588,7 @@ private Optional<Map<String, Object>> minionOpts(String minionId,
     private Optional<String> remotePortForwarding(List<String> proxyPath,
                                                   String sshContactMethod) {
         if (ContactMethodUtil.SSH_PUSH_TUNNEL.equals(sshContactMethod)) {
-            return Optional.of(getSshPushRemotePort() + ":" +
-                    ConfigDefaults.get().getJavaHostname() + ":" + SSL_PORT);
```
aaannz, line 592: I wonder how this worked even before with chain of proxies. This would always point to the server, wouldn't it? So not a local redirection. Do we have some other postprocessing step? e.g. translating the server to the proxy fqdn?

aaannz (commented):
I am probably missing something, but this looks to me that idea was to forward from proxy to another proxy or server. So we should not be forwarding to ourselves and this not have a problem with hairpin.

Not disproving that the fix does not work btw.

java/code/src/com/suse/manager/webui/services/impl/SaltSSHService.java
mcalmer: I think `ConfigDefaults.get().getJavaHostname()` is not what 4.3 has. This was already tried to be fixed.
But I think we did not test for a long time chain proxies

aaannz (approved):
Thanks for the explanation. LGTM

PR #8913, opened 2024-06-11
TEST

jordimassaguerpla:
test https://github.com/SUSE/spacewalk/issues/20676

### DIFF
--- a/testsuite/features/secondary/min_salt_minions_page.feature
+++ b/testsuite/features/secondary/min_salt_minions_page.feature
@@ -5,7 +5,6 @@
 # - features/secondary/min_salt_mgrcompat_state.feature
 # If the minion fails to bootstrap again.
 
-@skip_if_github_validation
 @sle_minion
 @scope_salt
 Feature: Management of minion keys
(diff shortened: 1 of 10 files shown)


jordimassaguerpla:
I remember this problem! We can't use sle_minion but then, when I tried to use something else like sleminion, sle_minion was the expected hostname in other tests ... @srbarrios : How this works when we run the tests for master?

jordimassaguerpla:
Trying to use sleminionone , instead of sle_minion ...

jordimassaguerpla:
it won't work because the test restarts the salt-minion service and the minion has not been started with systemd.

srbarrios:
> I remember this problem! We can't use sle_minion but then, when I tried to use something else like sleminion, sle_minion was the expected hostname in other tests ... @srbarrios : How this works when we run the tests for master?

we must use the fqdn there, as the server don't understand our test framework internal keys, but... I'm not sure what's the fqdn in the case of that GitHub environment.

The best way to discover it, is to run `hostname -f` on that minion machine.

PR #8916, opened 2024-06-13
Fix `AttributeError: ENABLE_NVREA` for reposync

nodeg:
## What does this PR change?

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1226273.

I am also needing a changelog, which I currently do not know how to properly add.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**
- [ ] 
## Test coverage
- No tests: already covered
- [x] **DONE**

## Links

Issue(s): #
Port(s): # Manager 4.3: 

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


nodeg:
Do we need/want a backport for this?

nodeg:
I am not sure why, but the `java_pgsql_tests` fail with
image

python/spacewalk/server/importlib/backend.py
```diff
@@ -27,6 +27,7 @@
 from uyuni.common.usix import raise_with_tb
 from uyuni.common import rhn_rpm
 from spacewalk.common.rhnConfig import CFG
+from spacewalk.common.rhnConfig import cfg_component
```
vzhestkov, line 30:
```suggestion
from spacewalk.common.rhnConfig import CFG, cfg_component
```

vzhestkov (dismissed):
1 suggestion in line, but other than that looks good to me :+1:

python/spacewalk/server/importlib/backend.py
nodeg: Makes sense, thank you. Fixed.

python/spacewalk/server/importlib/backend.py
```diff
@@ -26,7 +26,7 @@
 
 from uyuni.common.usix import raise_with_tb
 from uyuni.common import rhn_rpm
-from spacewalk.common.rhnConfig import CFG
```
agraul, line 29: Keeping `CFG` is a bit dangerous, we might want to remove all usages of it. It could lead to the same kind of bug (wrong component is loaded when `CFG` is accessed).

nodeg: Could you or one of you drive this further, please? I am too inexperienced with that. You can use this PR or do it after this one will get merged.

agraul: Since it's a bugfix, I think it will go to the bug squad, either @ycedres or @vzhestkov can extend the fix to cover all `CFG` usages.

nodeg: Makes sense. So what about this PR?

python/spacewalk/server/importlib/backend.py
```diff
@@ -3295,10 +3292,11 @@ def _do_diff(self, data, table_name, uq_fields, fields):
     def validate_pks(self):
         # If nevra is enabled use checksum as primary key
         tbs = self.tables["rhnPackage"]
-        if not CFG.ENABLE_NVREA:
-            # remove checksum from a primary key if nevra is disabled.
-            if "checksum_id" in tbs.pk:
-                tbs.pk.remove("checksum_id")
+        with cfg_component("server") as cfg:
+            if not cfg.ENABLE_NVREA:
```
vzhestkov, line 3296: For me looks like `AttributeError: ENABLE_NVREA` is still possible here if it's missing in rhn.conf or default configs for some reason.

agraul: We have it in our default `rhn_server.conf`, if there is an `AttributeError` there is a bigger problem (default configs missing or modified), IMO it's okay to show a backtrace in that situation.

m-czernek:
@agraul @nodeg I see that you improved the initial PR, using cfg_component instead of CFG directly, correct? Is there something more we have to do here, or is the PR just waiting for a review?

ycedres:
> @m-czernek I think it's just a PR review now. It would be good if someone double checked the `component=None` arguments

@agraul IIUC when `component=None` the `RHNOptions` object is initialised with the "global" component here https://github.com/uyuni-project/uyuni/blob/a560f41e5e805ff256d254e94e33e23aee453d25/python/spacewalk/common/rhnConfig.py#L96

But I'm not sure if that's what you'd like to double check.

nodeg:
Since this is currently an issue on HEAD and Uyuni, could we please prioritize this again?

python/spacewalk/server/importlib/backend.py
```diff
@@ -171,11 +171,9 @@ def processCapabilities(self, capabilityHash):
 
     # pylint: disable-next=invalid-name
     def processChangeLog(self, changelogHash):
-        if (
-            CFG.has_key("package_import_skip_changelog")
-            and CFG.package_import_skip_changelog
-        ):
-            return
+        with cfg_component(None) as cfg:
```
ycedres, line 174: @agraul last thing you commented was about checking if `component=None` was the right component when accessing `package_import_skip_changelog` (if I understood you correctly). I believe this property is not in the configuration defaults  (and I believe not documented either).

m-czernek: Either way, it does seem to be used at the top level of `/etc/rhn/rhn.conf`, e.g. in BZ#1129748. So IMO `component=None` sounds correct to me.

python/spacewalk/server/importlib/backend.py
m-czernek: Note here, I cannot find `rhn_server.conf` in SUMA 5.0, nor can I find any `ENABLE_NVREA` property defined on my Uyuni server as well as on `suma-ci-50-server`.

m-czernek (approved):

ycedres (approved):

susemanager-utils/susemanager-sls/susemanager-sls.changes.mczernek.context_manage_fix_reposync
```diff
@@ -0,0 +1 @@
+- Fix python sonarlint issue S5754
```
agraul, line 1: I'd drop this, users are not interested in this change and they never saw it in the first place (since the code was introduced just a day ago)

agraul (approved):
LGTM, I'd just drop the new changelog since it's lint fixes don't need to be in the user-facing changelog

susemanager-utils/susemanager-sls/susemanager-sls.changes.mczernek.context_manage_fix_reposync
m-czernek: I needed to add this changelog; otherwise, the chlog check was failing that we're changing a file without a chlog entry in a package. Is there a way to disable the check?

nodeg:
Thank you for taking over and driving this PR!

PR #8943, opened 2024-06-18
Adapt channel change

mcalmer:
## What does this PR change?

Currently we have 2 ways in the code how channels are changed.
1. One is changing the channels first in the Database and send it out to the client after it was changed.
2. The other way (new) we send the new channels out to the client first and only when the change was performed on the client successfully, we adapt the channels in the database.

The first (old) way has the benefit, that also offline clients change the channels and it can be changed on the client also later by just applying the channel state. 

The 2nd way is complicated as new access tokens needs to be generated, but the channel pillar data needs to be manually compiled and send via the state.apply instead of using the standard way by reading the database.
Offline minions will not change the channels at all which is a requirement of some users.

This PR try to go back to the 1st (old) way to just perform the changes in the Database and send out the result to the client.
If that fail, we do not invalidate the tokens anymore as a new state.apply can fix the situation later.

For this we have to remove asynchronous handling of of things which were done in ChannelsChangeEventMessageAction as the following state.apply for the channels require to have the data in the database and this event is waiting on the main event to be finished.
In this one case we execute now this Action Syncronous, while on Distupgrade and other cases we still use it via the message queue.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were adapted

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23251
Port(s): ???

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/server/MinionServer.java
+++ b/java/code/src/com/redhat/rhn/domain/server/MinionServer.java
@@ -15,6 +15,8 @@
 package com.redhat.rhn.domain.server;
 
 import com.redhat.rhn.domain.channel.AccessToken;
+import com.redhat.rhn.domain.channel.AccessTokenFactory;
+import com.redhat.rhn.domain.channel.Channel;
 import com.redhat.rhn.domain.config.ConfigChannel;
 import com.redhat.rhn.domain.user.User;
 import com.redhat.rhn.manager.configuration.SaltConfigSubscriptionService;
@@ -29,6 +31,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * MinionServer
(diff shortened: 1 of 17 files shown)


java/code/src/com/redhat/rhn/domain/server/MinionServer.java
```diff
@@ -219,6 +222,20 @@ public boolean equals(Object other) {
...
     }
 
+
+    /**
+     * @return Return true when all assigned software channels have valid access tokens.
+     */
+    public boolean hasValidTokensForAllChannels() {
+
+        Set<Channel> tokenChannels = AccessTokenFactory.listByMinion(this)
+                .stream()
```
lucidd, line 232: does this require to check it the token is valid or is this done elsewhere already?

mcalmer: This is currently used only for a unit test, but it does not check for valid token. Just that every channel has one.
I think it makes sense to check also if the tokens are valid.
I will look into this.

mackdk (dismissed):

lucidd (approved):

PR #8999, opened 2024-07-03
Fix CoCo Attestation API errors

mackdk:
## What does this PR change?

This PR fixes:

- The absence of the `attestOnBoot` field in the `getCoCoAttestationConfig` API
- The error that happens when serializing for the XMLRPC API a `ServerCoCoAttestationReport` when the `attested` date is `null`. This might happen when the report is still PENDING
- The field `attestOnBoot` is ignored by `setCoCoAttestationConfig`  API when no configuration exists
- The error that happens when `getCoCoAttestationConfig` is called and no configuration is present. Now if the system has coco support an empty configuration is returned.
- Prevent all the API to work with systems that do not have support for coco attestation. An `UnsupportedOperationException` is now returned by the API.
- Return an `EntityNotExistsFaultException` when the system specified in the call does not match the one from the attestation result when calling `getCoCoAttestationResultDetails`

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/CoCoAttestationConfigSerializer.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/CoCoAttestationConfigSerializer.java
@@ -44,6 +44,7 @@ public class CoCoAttestationConfigSerializer extends ApiResponseSerializer<Serve
                 .add("enabled", src.isEnabled())
                 .add("environment_type", src.getEnvironmentType().name())
                 .add("system_id", src.getServer().getId())
+                .add("attest_on_boot", src.isAttestOnBoot())
                 .build();
     }
 }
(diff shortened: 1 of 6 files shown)


mcalmer (commented):
I just found another bug.
SystemHandler.setCoCoAttestationConfig() support the attestOnBoot parameter,
but it set it only when the config is already present. If a config needs to be created that parameter is not set.
Maybe you can fix this as well in this PR?

mackdk:
> I just found another bug. SystemHandler.setCoCoAttestationConfig() support the attestOnBoot parameter, but it set it only when the config is already present. If a config needs to be created that parameter is not set. Maybe you can fix this as well in this PR?

Oh, I didn't notice that. I will fix it.

lucidd (dismissed):

mackdk:
I added two more fix, I updated the PR description to describe them accordingly.

mcalmer (approved):

lucidd (approved):

Issue #8972, opened 2024-06-25
Problem importing modular packages from https://rpms.remirepo.net/


bamorrow:
### Problem description

I am getting failures in trying to import he modular repo's from https://rpms.remirepo.net/ for EL8 and EL9.
A sample message I get is:

 File "/usr/lib/python3.6/site-packages/spacewalk/satellite_tools/appstreams.py", line 294, in _parse_rpm_name
    raise ValueError(f"The value {nevra} cannot be parsed as a NEVRA string.")
ValueError: The value php-pecl-gmagick-0:2.0.6~RC1-3.el9.remi.7.4.x86_64 cannot be parsed as a NEVRA string.

If I remove that file and test, it just happens the same on the rest of the files in the repo.

### Steps to reproduce

1.  add repo baseurl=http://rpms.remirepo.net/enterprise/8/modular/x86_86/
2. try and run a sync
3.
...

### Uyuni version

```bash
Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : Uyuni Server Stable
Name           : Uyuni-Server-release
Version        : 2024.05-230900.217.1.uyuni3
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes
Status         : up-to-date
Source package : Uyuni-Server-release-2024.05-230900.217.1.uyuni3.src
Summary        : Uyuni Server
```

### Uyuni proxy version (if used)

```bash
N/A
```

### Useful logs

```bash
2024/06/25 10:13:27 -05:00 Importing packages finished.
2024/06/25 10:13:27 -05:00 
2024/06/25 10:13:27 -05:00   Linking packages to the channel.
2024/06/25 10:13:30 -05:00     1000 packages linked
2024/06/25 10:13:32 -05:00     1463 packages linked
2024/06/25 10:13:32 -05:00   Importing module metadata:
2024/06/25 10:13:32 -05:00     Modules in repo: 3
2024/06/25 10:13:32 -05:00 2 new package(s) imported for module composer:1:20240614071236:00000000:x86_64.
2024/06/25 10:13:32 -05:00 2 new package(s) imported for module composer:2:20240614071236:00000000:x86_64.
2024/06/25 10:13:32 -05:00 Unexpected error: <class 'ValueError'>
2024/06/25 10:13:32 -05:00 Traceback (most recent call last):
[...]

PR #9003, opened 2024-07-03
Fix importing module metadata from REMI's repositories

cbbayburt:
Some packages in REMI's 3rd party repositories include a '~' character as part of the version string. This patch fixes the module import during repo-sync to recognize this character.

## Documentation

## Test coverage
- Unit tests were added

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/8972

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/satellite_tools/appstreams.py
+++ b/python/spacewalk/satellite_tools/appstreams.py
@@ -286,7 +286,7 @@ class ModuleMdImporter:
         pattern = re.compile(
             r"(?P<name>[a-zA-Z0-9._+-]+)-"
             r"(?P<epoch>\d+:)?"
-            r"(?P<version>[a-zA-Z0-9._-]+)-"
+            r"(?P<version>[a-zA-Z0-9._-~]+)-"
             r"(?P<release>[a-zA-Z0-9._+-]+)\."
             r"(?P<arch>[a-zA-Z0-9._-]+)"
         )
--- a/python/spacewalk/spacewalk-backend.changes.cbbayburt.appstreams-reposync-8972
+++ b/python/spacewalk/spacewalk-backend.changes.cbbayburt.appstreams-reposync-8972
@@ -1,0 +1,1 @@
+- Support more NEVRA types when importing module metadata
--- a/python/test/unit/spacewalk/satellite_tools/test_appstreams.py
+++ b/python/test/unit/spacewalk/satellite_tools/test_appstreams.py
@@ -178,6 +178,14 @@ def test_get_modules(importer):
             "396.module_el8.1.0+6019+b22674e1",
             "noarch",
         ),
+        (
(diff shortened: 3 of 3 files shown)


python/spacewalk/satellite_tools/appstreams.py
```diff
@@ -286,7 +286,7 @@ def _parse_rpm_name(nevra):
         pattern = re.compile(
             r"(?P<name>[a-zA-Z0-9._+-]+)-"
             r"(?P<epoch>\d+:)?"
-            r"(?P<version>[a-zA-Z0-9._-]+)-"
+            r"(?P<version>[a-zA-Z0-9._-~]+)-"
```
agraul, line 289: Is a hyphen/dash inside the version string valid? openSUSE's packaging documentation says it's not available, does that differ in the RH world?
>RPM uses the dash to separate the package name, the package version and the distro-level re-issue ("release") counter. (`rpm -qa` for example emits coreutils-9.0-1.x86_64.) The dash itself is unavailable for use within the version or release substring, as are a few other characters like the asterisk. [...]

https://en.opensuse.org/openSUSE:Package_versioning_guidelines#Unusual_characters

cbbayburt: The real problem is that there's no actual standard between distros and like also mentioned there in the document, some distros are pretty liberal with this.

For "hyphens in version" case, Ubuntu/Debian is the offender. Here are some examples:
```
ipxe-qemu-256k-compat-efi-roms-1.0.0+git-20150424.a25a16d-0ubuntu4.all-deb.deb
ipxe-1.0.0+git-20190109.133f4c4-0ubuntu3.all-deb.deb
ipxe-qemu-1.0.0+git-20190109.133f4c4-0ubuntu3.all-deb.deb
grub-ipxe-1.0.0+git-20190109.133f4c4-0ubuntu3.all-deb.deb
lib32gcc1-10-20200411-0ubuntu1.amd64-deb.deb
libx32gcc1-10-20200411-0ubuntu1.amd64-deb.deb
```

I should probably add some more test cases reflecting this.

agraul: Is appstreams.py also used by Debian repos? I thought this was RPM-only.

Package version schemas definitely differ between packaging formats and we can't have one way to handle all of them. The same applies to version comparisons, where has their own rules.

ycedres: `libsolv` has some version comparison algorithms from which  maybe some reasonable expectations for a version string can be extracted:

https://github.com/openSUSE/libsolv/blob/master/src/evr.c

cbbayburt: I can have a look at it if it covers all the necessary formats. But in any case, further changes are not in scope for this fix and I don't want to risk any regressions without testing extensively. So I'm just going to merge this one as is.

PR #9014, opened 2024-07-05
QE: correctly reset the API client

NamelessOne91:
## What does this PR change?

Related to https://github.com/SUSE/spacewalk/issues/23475

The return of this function call was never used, causing the API client to still be the previous one.
Yep, that's all.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- A Cucumber step definition was modified

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23475
Port(s): 4.3

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1642,7 +1642,7 @@ When(/^I run spacewalk-hostname-rename command on the server$/) do
 
   # Reset the API client to take the new CA into account
   log 'Resetting the API client'
-  new_api_client
+  $api_test = new_api_client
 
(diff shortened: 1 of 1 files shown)


nodeg (approved):
Really nice catch, Paolo! Thank you very much for looking into that. We have the correct assignment here
https://github.com/uyuni-project/uyuni/blob/489c01e6cac9f3ae074e2679a909eb372fd09688/testsuite/features/support/env.rb#L117
so I am not sure why I missed it in `command_steps.rb`.

Bischoff (approved):

ktsamis (approved):

PR #9032, opened 2024-07-09
deb_src: introduce timeout when syncing DEB channels (bsc#1225960)

meaksh:
## What does this PR change?

As mentioned in the title, this PR introduces a timeout when running "reposync" against DEB channels. This would prevent getting reposync stuck in case connections with the remote repositories gets stuck.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **no tests for this corner case**

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24459

Port(s): https://github.com/SUSE/spacewalk/pull/24755

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/common/repo.py
+++ b/python/spacewalk/common/repo.py
@@ -84,7 +84,13 @@ class DpkgRepo:
                 )
             return self[key]
 
-    def __init__(self, url: str, proxies: dict = None, gpg_verify: bool = True):
+    def __init__(
+        self,
+        url: str,
+        proxies: dict = None,
+        gpg_verify: bool = True,
+        timeout: typing.Optional[int] = None,
+    ):
         self._url = url
         self._flat_checked: typing.Optional[int] = None
         self._flat: bool = False
@@ -95,6 +101,7 @@ class DpkgRepo:
         self._release = DpkgRepo.EntryDict(self)
         self.proxies = proxies
         self.gpg_verify = gpg_verify
+        self.timeout = timeout
 
     def append_index_file(self, index_file: str) -> str:
         """
@@ -156,7 +163,11 @@ class DpkgRepo:
                             exc_info=True,
                         )
                 else:
-                    resp = requests.get(packages_url, proxies=self.proxies)
+                    resp = requests.get(
+                        packages_url,
+                        proxies=self.proxies,
+                        timeout=self.timeout,
+                    )
                     if resp.status_code == http.HTTPStatus.OK:
                         self._pkg_index = cnt_fname, resp.content
                         break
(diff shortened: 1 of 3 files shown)


python/spacewalk/common/repo.py
```diff
@@ -84,7 +84,7 @@ def get(self, key: typing.Any) -> typing.Optional[typing.Any]:  # type: ignore
                 )
             return self[key]
 
-    def __init__(self, url: str, proxies: dict = None, gpg_verify: bool = True):
+    def __init__(self, url: str, proxies: dict = None, gpg_verify: bool = True, timeout: int = None):
```
agraul, line 87: `None` is not an `int`, either the type hint or the default value is wrong.

meaksh: fixed

python/spacewalk/common/repo.py
```diff
@@ -84,7 +84,13 @@ def get(self, key: typing.Any) -> typing.Optional[typing.Any]:  # type: ignore
                 )
             return self[key]
 
-    def __init__(self, url: str, proxies: dict = None, gpg_verify: bool = True):
+    def __init__(
+        self,
+        url: str,
+        proxies: dict = None,
+        gpg_verify: bool = True,
+        timeout: typing.Optional[int] = None,
```
agraul, line 92: Is `int` the correct type? In the python-requests [documentation](https://docs.python-requests.org/en/latest/api/#requests.request), the `timeout` parameter is documented as follows:

> timeout ([float](https://docs.python.org/3/library/functions.html#float) or [tuple](https://docs.python.org/3/library/stdtypes.html#tuple)) – (optional) How many seconds to wait for the server to send data before giving up, as a float, or a [(connect timeout, read timeout)](https://docs.python-requests.org/en/latest/user/advanced/#timeouts) tuple.

meaksh: In the context of reposync, the timeout is considered as integer value, so I guess that is what motivated the use of integer also for the timeout parameter of `DpkgRepo`  and `DebRepo` now.

There shouldn't be a problem passing this integer value to `requests` module, it should be evaluated properly.

agraul (changes requested):
Please squash the second and third commit into the first. From my POV we can merge it afterwards.

meaksh:
Thanks @agraul. I've squashed the commits as you suggested but I would need an explicit approval here as your review is still "requesting changes", so GitHub won't allow to merge this.

agraul (commented):
I don't really think the new commit message containing the summaries of the squashed commits is useful. Are they included intentionally?

ycedres (approved):
LGTM.

PR #9036, opened 2024-07-09
Fix failing server action

mcalmer:
## What does this PR change?

In the past we did not use TypeAdapters for JSON parsing. That encoded optional values different than the new type adapter.
Reading the old values cause now Json Parsing Exceptions.

This PR fix parsing the legacy entries and also writing out the new entries.
Additionally the context return now sometimes no timezone which result in a NullPointerException.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/server/ServerAction.java
+++ b/java/code/src/com/redhat/rhn/domain/action/server/ServerAction.java
@@ -24,6 +24,8 @@ import com.suse.manager.maintenance.MaintenanceManager;
 
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.io.Serializable;
 import java.util.Date;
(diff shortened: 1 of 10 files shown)


mackdk (commented):
Looks good to me. Maybe instead of using `Optional` everywhere for the issue of the null `timezone`, we could directly fix the `Context.getTimezone()` method:

```java
    /**
     * @return Returns the timezone.
     */
    public TimeZone getTimezone() {
        if (timezone == null) {
            return TimeZone.getDefault();
        }
        
        return timezone;
    }
```

In this way `Context` will always return a non-null value and we will be more safe for future code.

mcalmer:
> In this way `Context` will always return a non-null value and we will be more safe for future code.

Yes, but we need to check if there is code which is checking for NULL and than use a different value than TimeZone.getDefault().
I think I saw such code.

lucidd (dismissed):

lucidd (approved):

mackdk (dismissed):

admd:
Tests are failing here. Can one of you please take a look before we can proceed with the merge.

PR #9057, opened 2024-07-15
Autoinstallation: prevent issues with duplicate IP address due to some networks (bsc#1226461)

meaksh:
## What does this PR change?

This PR fixes an issue seen when dealing with Autoinstallation and some of your systems contain `docker*`, `virbr*` or `cni-podman*` interfaces with the same IP address.

Then, when triggering autoinstallation for those systems, an ISE is raised (due duplicate IP address on Cobbler) and autoinstallation is not triggered properly:

```
...
Caused by: org.cobbler.XmlRpcException: XmlRpcException calling cobbler.
        at org.cobbler.CobblerConnection.invokeMethod(CobblerConnection.java:160) ~[rhn.jar:?]
        at org.cobbler.CobblerConnection.invokeTokenMethod(CobblerConnection.java:190) ~[rhn.jar:?]
        at org.cobbler.SystemRecord.invokeModify(SystemRecord.java:310) ~[rhn.jar:?]
        at org.cobbler.CobblerObject.modify(CobblerObject.java:339) ~[rhn.jar:?]
        at org.cobbler.CobblerObject.modify(CobblerObject.java:322) ~[rhn.jar:?]
        at org.cobbler.SystemRecord.setNetworkInterfaces(SystemRecord.java:822) ~[rhn.jar:?]
...
Caused by: redstone.xmlrpc.XmlRpcFault: <class 'ValueError'>:IP address duplicate found "172.17.0.1". Object with the conflict has the name "my-system-host:1"
        at redstone.xmlrpc.XmlRpcClient.handleResponse(XmlRpcClient.java:444) ~[redstone-xmlrpc-client.jar:?]
        at redstone.xmlrpc.XmlRpcClient.endCall(XmlRpcClient.java:376) ~[redstone-xmlrpc-client.jar:?]
        at redstone.xmlrpc.XmlRpcClient.invoke(XmlRpcClient.java:165) ~[redstone-xmlrpc-client.jar:?]
```

The issue is caused as Java is not properly excluding these interfaces when preparing the Cobbler systems for autoinstallation, where some types of interfaces are being excluded to avoid this.

As a workaround, if we set `allow_duplicate_ips: true` on `/etc/cobbler/settings.yaml` then this issue is not happening.

So, to summarize, this PR does the following fixes:
- Fix string to parse when Cobbler raises IP address exception, to avoid a general ISE as it was expected in these cases.
- Make `sumautil` Salt module able to identify "bridge" network devices, as this is needed for Java for proper matching of network device types.
- Do not process container network devices: `docker*` and `cni-podman*`, when preparing an autoinstallation.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24580
Port(s): https://github.com/SUSE/spacewalk/pull/24815

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/server/NetworkInterface.java
+++ b/java/code/src/com/redhat/rhn/domain/server/NetworkInterface.java
@@ -318,7 +318,7 @@ Serializable {
      * @return true if the nic is a container network
      */
     public boolean isContainerNetwork() {
-        return ("bridge".equals(module) && (getName().startsWith("docker") || getName().startsWith("cni-podman")));
+        return ("bridge".equals(module) && (getName().startsWith("docker") || getName().startsWith("cni-podman") || getName().startsWith("podman")));
     }
 
     /**
--- a/java/code/src/com/redhat/rhn/domain/server/NetworkInterface.java
+++ b/java/code/src/com/redhat/rhn/domain/server/NetworkInterface.java
@@ -314,6 +314,14 @@ Serializable {
     }
 
     /**
+     * true if the nic is a container network
+     * @return true if the nic is a container network
+     */
+    public boolean isContainerNetwork() {
+        return ("bridge".equals(module) && (getName().startsWith("docker") || getName().startsWith("cni-podman")));
+    }
+
+    /**
      * isVirtBridge tells if nic is a virtual bridge
      * @return true if the nic is a virtual bridge, false otherwise
      */
--- a/java/code/src/com/redhat/rhn/domain/server/NetworkInterface.java
+++ b/java/code/src/com/redhat/rhn/domain/server/NetworkInterface.java
@@ -318,7 +318,13 @@ Serializable {
      * @return true if the nic is a container network
      */
     public boolean isContainerNetwork() {
-        return ("bridge".equals(module) && (getName().startsWith("docker") || getName().startsWith("cni-podman") || getName().startsWith("podman")));
+        return (
+            "bridge".equals(module) && (
+                    getName().startsWith("docker") ||
+                    getName().startsWith("cni-podman") ||
+                    getName().startsWith("podman") ||
+                    getName().startsWith("cni")
+            ) || getName().startsWith("flannel"));
     }
 
     /**
--- a/java/spacewalk-java.changes.meaksh.Manager-4.3-prevent-issues-on-duplicate-ips-bsc1226461
+++ b/java/spacewalk-java.changes.meaksh.Manager-4.3-prevent-issues-on-duplicate-ips-bsc1226461
@@ -1,0 +1,1 @@
+- Autoinstallation: prevent "duplicate IP address" issues cause by container networks (bsc#1226461)
--- a/susemanager-utils/susemanager-sls/src/modules/sumautil.py
+++ b/susemanager-utils/susemanager-sls/src/modules/sumautil.py
@@ -115,11 +115,15 @@ def get_net_module(iface):
 
         salt '*' sumautil.get_net_module eth0
     """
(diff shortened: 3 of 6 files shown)


java/code/src/com/redhat/rhn/domain/server/NetworkInterface.java
```diff
@@ -313,6 +313,14 @@ public boolean isBond() {
         return "bonding".equals(module);
     }
 
+    /**
+     * true if the nic is a container network
+     * @return true if the nic is a container network
+     */
+    public boolean isContainerNetwork() {
+        return ("bridge".equals(module) && (getName().startsWith("docker") || getName().startsWith("cni-podman")));
```
mbussolotto, line 321: Is there any other way to do it?  I see 2 problems:
- `netavark` is missing (but probably you just need to add `getName().startsWith("cni-netavark")`
-  I don't think this would work for K3s (I know, it's not released yet, but we're trying to do our best to not introduce something that might work for podman and not for kubernetes)

meaksh: Thanks for the review @mbussolotto !

Ok, good point about `netavark` and Kubernetes. I'll try to enhance the mechanism.

In the end, I just mimic here what we already do for the `virbr*` devices: https://github.com/uyuni-project/uyuni/pull/9057/files#diff-dd63c17a089292848e40dfb8c02de48edb6b619425601d9fe81d02aa2cf3ba5eR329

java/spacewalk-java.changes.meaksh.Manager-4.3-prevent-issues-on-duplicate-ips-bsc1226461
```diff
@@ -0,0 +1 @@
+- Autoinstallation: prevent issues with duplicate IP address due to some networks (bsc#1226461)
```
agraul, line 1: I would rephrase this slightly to make it clearer. "due to some networks" is hard to understand IMO.

```suggestion
- Autoinstallation: prevent "duplicate IP address" issues cause by container networks (bsc#1226461)
```

susemanager-utils/susemanager-sls/src/modules/sumautil.py
```diff
@@ -117,9 +117,12 @@ def get_net_module(iface):
     """
     sysfspath = os.path.join(SYSFS_NET_PATH, iface, "device/driver")
 
-    return (
-        os.path.exists(sysfspath) and os.path.split(os.readlink(sysfspath))[-1] or None
-    )
+    if os.path.exists(sysfspath):
+        return os.path.split(os.readlink(sysfspath))[-1] or None
+    elif os.path.exists(os.path.join(SYSFS_NET_PATH, iface, "bridge")):
```
agraul, line 122: Can we give this a name? Above we use `sysfspath`, which is now ambiguous. For consistency it would be good to give both of them a name

agraul (commented):
I agree that the approach taken on the Java side might require quite a bit of maintenance for keeping up with new strings to compare the "module" to.

Other than that, I'd like us to be a bit more consistent with the different branches in sumautil by giving case each an easy-to-understand name

susemanager-utils/susemanager-sls/src/modules/sumautil.py
meaksh: fixed

mbussolotto (approved):

PR #9065, opened 2024-07-17
Improve SQL queries and performance to check for PTF packages (bsc#1225619)

meaksh:
## What does this PR change?

This PR improves the performance to determine which packages are PTFs or part of a PTF, as in large environments with tons of packages, these SQL operations, and particularly the `susePackageExcludingPartOfPtf` SQL view can cause very huge delays in the context of Content Lifecycle Management (CLM) operations.

In order to achieve a better performance, this PR is doing the following changes:
- Add new columns `is_ptf` and `is_part_of_ptf` to `rhnPackage` table.
- These attributes are set once during package import, so no need to calculate them on each request.
- The `susePackageExcludingPartOfPtf` view and other SQL queries have been adapted to use the new attributes.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24426
Port(s): https://github.com/SUSE/spacewalk/pull/24882

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

- [x] Re-run test "spacecmd_unittests" (Test skipped, there are no changes to test)

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/common/db/datasource/xml/Package_queries.xml
+++ b/java/code/src/com/redhat/rhn/common/db/datasource/xml/Package_queries.xml
@@ -860,27 +860,15 @@ SELECT P.id, P.summary, PP.name as provider
         SELECT pid as package_id
           FROM suseServerChannelsRetractedPackagesView
          WHERE sid = :sid
-    ), ptfPackages AS (
-        SELECT pp.package_id
-          FROM rhnpackagecapability pc
-                  INNER JOIN rhnpackageprovides pp ON pc.id = pp.capability_id
-         WHERE pc.name = 'ptf()'
-    ), partOfPtfPackages AS (
-        SELECT pp.package_id
-          FROM rhnpackagecapability pc
-                  INNER JOIN rhnpackageprovides pp ON pc.id = pp.capability_id
-         WHERE pc.name = 'ptf-package()'
     )
     SELECT pk.id AS package_id
               -- Convert the ids to 0/1 flags
               , SIGN(COALESCE(rp.package_id, 0)) AS retracted
-              , SIGN(COALESCE(mp.package_id, 0)) AS master_ptf_package
-              , SIGN(COALESCE(pp.package_id, 0)) AS part_of_ptf
+              , SIGN(COALESCE(CAST(pk.is_ptf AS INTEGER), 0))::int AS master_ptf_package
+              , SIGN(COALESCE(CAST(pk.is_part_of_ptf AS INTEGER), 0))::int AS part_of_ptf
       FROM rhnpackage pk
               LEFT JOIN retractedPackages rp ON rp.package_id = pk.id
-              LEFT JOIN ptfPackages mp ON mp.package_id = pk.id
-              LEFT JOIN partOfPtfPackages pp ON pp.package_id = pk.id
-     WHERE pk.id IN (%s) AND ( rp.package_id IS NOT NULL OR mp.package_id IS NOT NULL OR pp.package_id IS NOT NULL )
+     WHERE pk.id IN (%s) AND ( rp.package_id IS NOT NULL OR pk.is_ptf = TRUE OR pk.is_part_of_ptf = TRUE )
 </query>
 
 <mode name="system_available_packages"  class="com.redhat.rhn.frontend.dto.PackageListItem">
@@ -2447,9 +2435,7 @@ WITH ptfInstalled AS (
     SELECT sp.*
       FROM rhnserverpackage sp
                 INNER JOIN rhnpackage pk ON sp.name_id = pk.name_id AND sp.evr_id = pk.evr_id AND sp.package_arch_id = pk.package_arch_id
-                INNER JOIN rhnpackageprovides pp ON pk.id = pp.package_id
-                INNER JOIN rhnpackagecapability pc ON pp.capability_id = pc.id
-     WHERE sp.server_id = :sid AND pc.name = 'ptf()'
+     WHERE sp.server_id = :sid AND pk.is_ptf = TRUE
 ), highest_id_packages AS (
     SELECT p.name_id, p.evr_id, p.package_arch_id, max(p.id) AS id
       FROM ptfInstalled pi
@@ -2489,9 +2475,7 @@ ORDER BY nvre
 WITH ptf_packages AS (
     SELECT DISTINCT pk.*
       FROM rhnpackage pk
-                INNER JOIN rhnpackageprovides pp ON pk.id = pp.package_id
-                INNER JOIN rhnpackagecapability pc ON pp.capability_id = pc.id
-     WHERE pc.name = 'ptf()'
+     WHERE pk.is_ptf = TRUE
 )
 SELECT rp.id AS package_id,
        pn.name || '-' || evr_t_as_vre_simple(latest.evr) AS NVRE,
--- a/java/code/src/com/redhat/rhn/domain/rhnpackage/Package.java
+++ b/java/code/src/com/redhat/rhn/domain/rhnpackage/Package.java
@@ -55,6 +55,8 @@ public class Package extends BaseDomainHelper {
     private String path;
     private String headerSignature;
     private String copyright;
+    private Boolean isPtfPackage = false;
+    private Boolean isPartOfPtfPackage = false;
     private String cookie;
     private Date lastModified;
     private Boolean lockPending = Boolean.FALSE;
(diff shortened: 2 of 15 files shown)


python/spacewalk/server/importlib/packageImport.py
```diff
@@ -338,6 +339,9 @@ def _processPackage(self, package):
             if fchecksumTuple not in self.checksums:
                 self.checksums[fchecksumTuple] = None
 
+        package['is_ptf'] = any([x[0] == "ptf()" for x in self.capabilities.keys()])
+        package['is_part_of_ptf'] = any([x[0] == "ptf-package()" for x in self.capabilities.keys()])
```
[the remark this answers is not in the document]
meaksh: Right and it would be more efficient. I  just fixed it.

schema/spacewalk/common/tables/rhnPackage.sql
```diff
@@ -59,6 +59,14 @@ CREATE TABLE rhnPackage
...
     copyright        VARCHAR(128),
     cookie           VARCHAR(128),
+    is_ptf           NUMERIC(1)
+                         DEFAULT (0)
+                         CONSTRAINT rhn_package_is_ptf_check
+                             CHECK (is_ptf in ( 1 , 0 )),
+    is_part_of_ptf   NUMERIC(1)
+                         DEFAULT (0)
+                         CONSTRAINT rhn_package_is_part_of_ptf_check
+                             CHECK (is_part_of_ptf in ( 1 , 0 )),
```
aaannz, line 69: I wonder what is the reason to use `NUMERIC(1)` instead of [boolean]()?
Is it only because of "historic reason"?

There is quite size differences there:
```sql
# select pg_column_size(boolean 'TRUE'), pg_column_size(NUMERIC(1) '1');
pg_column_size | pg_column_size
----------------+----------------
             1 |              8
```

meaksh: This was just for historical reasons. I'll give it a try to use "boolean". Thanks for the note!

aaannz: Looking into some blogs wrt. postgres datatype performance, numeric type is one of the worst performant datatype. And I see we have plenty of them. 😄

cbbayburt: We also have plenty of `Y/N` char values :laughing:

meaksh:
> The change looks fine for me. Since the PR claims to improve performance, can you please provide data that backs the claim up? How much performance is gained?

Thanks for the review!

During the investigations on this L3, customer confirmed that due `susePackageExcludingPartOfPtf` view calculations was actually causing a bottleneck [1]. A TEST package was created to bypass `susePackageExcludingPartOfPtf` view and the performace increases significantly:

```
<<< before PTF >>> : 18816 | 01:32:12.56153  | admsmdb | INSERT INTO rhnServerNeededCache (server_id, package_id, channel_id) 

<<< with PTF  >>> : 18816 | 00:00:02.508874 | admsmdb | INSERT INTO rhnServerNeededCache (server_id, errata_id, package_id, channel_id)
```

There was later a discussion with Ricardo and it was agreed to go with this proposed solution here (moving calculation of ptf to import package time and store it in DB) [2]

[1] https://bugzilla.suse.com/show_bug.cgi?id=1225619#c26
[2] https://bugzilla.suse.com/show_bug.cgi?id=1225619#c27

cbbayburt (dismissed):
Thanks a lot for the fix, @meaksh. This saves us a huge sequential scan on `rhnPackage` on every operation involving packages. The performance gain is obvious.

cbbayburt:
Thinking out loud here: Indexing on boolean fields where the values are roughly 50/50 is generally useless, but in our case, only a tiny fraction of the packages would be PTFs. In this case, we can further benefit from some partial indexing (`WHERE NOT is_ptf`) but of course this needs some more research and testing.

schema/spacewalk/upgrade/susemanager-schema-5.0.8-to-susemanager-schema-5.0.9/300-replace-susePackageExcludingPartOfPtf-view.sql
```diff
@@ -0,0 +1,34 @@
...
+	CONSTRAINT rhn_package_is_ptf_check
+	    CHECK (is_ptf in ( 1 , 0 )),
+    ADD COLUMN IF NOT EXISTS is_part_of_ptf
+        NUMERIC(1)
+        DEFAULT (0)
+        CONSTRAINT rhn_package_is_part_of_ptf_check
+            CHECK (is_part_of_ptf in ( 1 , 0 ));
+
+CREATE INDEX IF NOT EXISTS rhn_package_is_ptf_idx ON rhnPackage (is_ptf);
+CREATE INDEX IF NOT EXISTS rhn_package_is_part_of_ptf_idx ON rhnPackage (is_part_of_ptf);
```
cbbayburt, line 14: These indexes are missing in the original definition files.

java/code/src/com/redhat/rhn/common/db/datasource/xml/Package_queries.xml
```diff
@@ -861,15 +861,13 @@ SELECT P.id, P.summary, PP.name as provider
           FROM suseServerChannelsRetractedPackagesView
          WHERE sid = :sid
     ), ptfPackages AS (
-        SELECT pp.package_id
-          FROM rhnpackagecapability pc
-                  INNER JOIN rhnpackageprovides pp ON pc.id = pp.capability_id
-         WHERE pc.name = 'ptf()'
+        SELECT id as package_id
```
rjmateus, line 864: you don't need this with clause for the ptfPackages and partOfPtfPackages. the select statement of the return query is already accessing the table `rhnpackage´, see we can just filter in the where clause of the main query, and let it use any index it have.

java/code/src/com/redhat/rhn/domain/rhnpackage/Package.java
```diff
@@ -55,6 +55,8 @@ public class Package extends BaseDomainHelper {
     private String path;
     private String headerSignature;
     private String copyright;
+    private Long isPtfPackage = 0L;
```
rjmateus, line 58: why do we need both flags? What is the difference between the two flags?

schema/spacewalk/common/tables/rhnPackage.sql
rjmateus: I had the same question as Ondrej. If we want to save a boolean flag we should use what database gives us for that :)

meaksh: This is now switched to use boolean in the DB

java/code/src/com/redhat/rhn/domain/rhnpackage/Package.java
meaksh: There are two different flags here:

1) `is_ptf` -> only for packages which provides `ptf()` attribute. This is used for the kind of metapackage for the PTF, named like: `ptf-ABCDE-VER-REL.PTF.xxxxxx.noarch.rpm`

2) `is_part_of_ptf` -> for those packages that are parf of a PTF (the actual patched package), like `salt-3006.0-xx.PTF.xxxxx.noarch.rpm`

I see our SQL queries are looking / filtering based on those two, so that is why added the two attributes so we can improve performance for the two attributes.

rjmateus: ok, got your point. When we are looking at the package capabilities we need to look for both to know if it's a PTF. 
But my question was more in the direction of whether we need to save both values separately fields or just one database field would be enough (like is_ptf, that would be true if the package have the capability `ptf()` of `ptf-package()`)

meaksh: In Uyuni I see there are currently different behaviors for PTF metapackages (the ones providing `ptf()` as capabilities) and the packages that are part of a PTF (the other ones with only `ptf-package()` capability).

Since we do things differently for each type of package, IMO we should store those two attributes separetely to gain on performance anytime we need to differentiate between "normal packages / metapackages / packages part of ptfs".

I hope this makes sense to you and I'm not missing anything! :)

schema/spacewalk/upgrade/susemanager-schema-5.0.8-to-susemanager-schema-5.0.9/300-replace-susePackageExcludingPartOfPtf-view.sql
meaksh: Added!

java/code/src/com/redhat/rhn/domain/rhnpackage/Package.java
rjmateus: ok, makes sense to me. All god from my side :)
 If we have both we can better control and decide to which we are looking for. BUt we must be aware that we need to look at both flags when we are implementing a feature related to PTF.

rjmateus (dismissed):

agraul (approved):
The Git history could use some cleanup, there are commits that just fix earlier commits. I recommend squashing them together (without keeping messages like "Fix Python code formatting issues raised by black")

aaannz (approved):

cbbayburt (approved):

PR #9079, opened 2024-07-22
BV: Add possibility to generate bootstrap repository without flush option for s390 arch

maximenoel8:
### Context

During BV, the sles15sp5s390 bootstrap is always failing because the bootstrap repository is missing for this architecture. The issue is caused by the fact that the sles15sp5 x86_64 and s390 architectures are sharing the same bootstrap repository. When the testsuite creates the bootstrap repository, it always flushes the bootstrap repository before generating it. @Bischoff changed the create bootstrap step definition to remove the flush option, but this option is not used during the create bootstrap stage in BV.
### Changes

This PR changes the way we generate the create bootstrap repository feature for the clients in BV. All the s390 architecture clients will now create the bootstrap repository without flushing.
### Dependencies

To make this work, we also need to ensure that the s390 architecture creates the bootstrap repository after the x86_64 "twins" have created their own bootstrap repositories. This will be managed at the CI level.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24866
Depends on : https://github.com/SUSE/susemanager-ci/pull/1279

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


testsuite/Rakefile
```diff
@@ -147,11 +153,14 @@ namespace :utils do
     clients.reject! { |client| client.include? 'ssh' }
     clients.reject! { |client| (client.include? 'buildhost') || (client.include? 'terminal') }
     clients.each do |client|
+      # disable flush option for s390 minions
+      flush = client.include?('s390') ? 'false' : 'true'
```
NamelessOne91, line 157: Nitpick: if `include?` returns a boolean, can't we just use that value instead of strings ?

```ruby
flush = !client.include?('s390')
```

testsuite/features/build_validation/create_bootstrap_repositories/create_bootstrap_repository.template
```diff
@@ -8,4 +8,4 @@ As the system administrator
 I create all bootstrap repos with --with-custom-channels option
 
   Scenario: Create the bootstrap repository for <client>
-    When I create the bootstrap repository for "<client>" on the server
+    When I create the bootstrap repository for "<client>" on the server <flush>
```
NamelessOne91, line 11: I don't think  this step definition will match in the case we add 'without flushing'.

```ruby
When(/^I create the bootstrap repository for "([^"]*)" on the server$/) do |host|
```

Is that intended because what's needed for s390 is already created by the bootstrap for a standard minion ?

testsuite/Rakefile
```diff
@@ -90,6 +90,15 @@ namespace :utils do
     output_filepath = "#{args[:output_path]}/#{args[:client_name]}_#{File.basename(args[:template_path], '.*')}.feature"
     sh "sed 's/<client>/#{args[:client_name]}/g' #{args[:template_path]} > #{output_filepath}", verbose: false
     sh "sed -i 's/<type>/salt/g' #{output_filepath}", verbose: false
+    # When two minions with the same version but different architectures create a bootstrap repository, they will share the same directory
+    # To prevent one minion to delete the other minion bootstrap repository, disable the flush option for s390 minion.
+    if args[:flush]
+      # This will generate `minion_create_bootstrap_repository.feature` with the flush option active
```
srbarrios, line 96: If we know that this only handled on `minion_create_bootstrap_repository` , I'd have another condition in that `if` statement, so we check if `output_filepath` is `minion_create_bootstrap_repository`, if so, we apply the `sed`, otherwise we don't need to apply this `sed` command for every single feature file running on that task.

PR #9081, opened 2024-07-22
QE: Use new Timeout::Error instead of deprecated TimeoutError

vandabarata:
## What does this PR change?
- Replace TimeoutError (deprecated) with Timeout::Error

## Links

Port(s): # https://github.com/SUSE/spacewalk/pull/24967

- [x] **DONE**


srbarrios:
Could we please explain in the PR the reasons to enable these two chromedriver features?
They are gonna be permanent or they are intended to debug some particular issue?

srbarrios (approved):

vandabarata:
> Could we please explain in the PR the reasons to enable these two chromedriver features? They are gonna be permanent or they are intended to debug some particular issue? A link to the stackoverflow page plus some details of our issue it's enough.

After some investigation, I ended up not really understanding the advantages it should bring, so I ended up going back on that decision

ktsamis (dismissed):

srbarrios (dismissed):
Thank you!

NamelessOne91 (approved):

testsuite/features/support/env.rb
```diff
@@ -136,8 +136,8 @@ def capybara_register_driver
       path = "screenshots/#{scenario.name.tr(' ./', '_')}.png"
       # only click on Details when we have errors during bootstrapping and more Details available
       click_button('Details') if has_content?('Bootstrap Minions') && has_content?('Details')
-      # a TimeoutError may be raised while a page is still (re)loading
-      find('#page-body', wait: 3) if scenario.exception.is_a?(TimeoutError)
+      # a Timeout::Error may be raised while a page is still (re)loading
+      find('#page-body', wait: 3) if scenario.exception.is_a?(Timeout::Error)
```
srbarrios, line 140: Note: I think is a good practice to `require` the dependency we use on every Ruby file using it. So, imho, it would be better to also include in env.rb, this line `require 'timeout'`

nodeg (approved):

srbarrios (approved):

nodeg:
Since there are some fixup commits, please squash the commits and use the correct message before merging this PR!

nodeg (approved):

PR #9090, opened 2024-07-24
QE : use current user instead of admin during after step authentification

maximenoel8:
### Context After step scenario
During the BV smoke test, we are using client user to execute the test.
If a scenario is failing, in `After do |scenario|`, the code block inside `if scenario.failed?` is going to be executed.
A step of this rescue scenario is to log back to susemanager with admin.
Unfortunately, this scenario switch back to admin user and doesn't use our client user.

### Context I reboot the "([^"]*)" minion through the web UI
During BV, we have a user for each smote tests scenario. Currently, `I reboot the "([^"]*)" minion through the web UI` is looking at reboot event triggered by admin ( only during bootstrap ) and not reboot event from the current user. Because the testsuite was not looking at the correct event, we were not waiting for the slemicro to correctly reboot and starting new test during the slemicro reboot.

### Changes
This change will correctly log with the client user when a scenario is failing.
This change will make sure we are looking for the correct reboot event.

## GUI diff

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
ℹ️ If a major new functionality is added, it is **strongly recommended** that tests for the new functionality are added to 
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24894

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -45,6 +45,8 @@ $server_http_proxy = ENV.fetch('SERVER_HTTP_PROXY', nil) if ENV['SERVER_HTTP_PRO
 $custom_download_endpoint = ENV.fetch('CUSTOM_DOWNLOAD_ENDPOINT', nil) if ENV['CUSTOM_DOWNLOAD_ENDPOINT']
 $no_auth_registry = ENV.fetch('NO_AUTH_REGISTRY', nil) if ENV['NO_AUTH_REGISTRY']
 $auth_registry = ENV.fetch('AUTH_REGISTRY', nil) if ENV['AUTH_REGISTRY']
+$current_user = 'admin'
+$current_password = 'admin'
 
 # maximal wait before giving up
 # the tests return much before that delay in case of success
@@ -144,7 +146,7 @@ After do |scenario|
     ensure
       print_server_logs
       previous_url = current_url
-      step 'I am authorized for the "Admin" section'
+      step %(I am authorized as "#{$current_user}" with password "#{$current_password}")
       visit previous_url
     end
   end
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1584,7 +1584,7 @@ When(/^I reboot the "([^"]*)" minion through the web UI$/) do |host|
     And I should see a "Reboot system" button
     When I click on "Reboot system"
     Then I should see a "Reboot scheduled for system" text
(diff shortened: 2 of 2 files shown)


testsuite/features/support/env.rb
```diff
@@ -45,6 +45,8 @@
 $custom_download_endpoint = ENV.fetch('CUSTOM_DOWNLOAD_ENDPOINT', nil) if ENV['CUSTOM_DOWNLOAD_ENDPOINT']
 $no_auth_registry = ENV.fetch('NO_AUTH_REGISTRY', nil) if ENV['NO_AUTH_REGISTRY']
 $auth_registry = ENV.fetch('AUTH_REGISTRY', nil) if ENV['AUTH_REGISTRY']
+$current_user = 'admin'
+$current_password = 'admin'
```
nodeg, line 49: Will this be changed somewhere? ATM, I do not get how your changes are different to what we already have? I mean, with the changes, we still use the `admin` user. Could you please elaborate this further? Thank you :wink:

maximenoel8: `current user` and `password` are set in the create user step ( only use during BV smoke test so far ).
https://github.com/uyuni-project/uyuni/blob/a2118e66782e3a880ea7defb5c08792b16bffed7/testsuite/features/step_definitions/api_common.rb#L139-L152
Because I want to keep those steps compatible with CI using admin, I need to setup a default current_user and current_password in this env file that will be overwritten by the create user step when needed.

NamelessOne91: This step does change those global variables and IIRC there are other PRs from Maxime which are going to use it

```ruby
Given(/^I create a user with name "([^"]*)" and password "([^"]*)"/) do |user, password|
  $current_user = user
  $current_password = password
  next if $api_test.user.list_users.to_s.include? user

  $api_test.user.create(user, password, user, user, "#{user}@mail.com")
  roles = %w[org_admin channel_admin config_admin system_group_admin activation_key_admin image_admin]
  roles.each do |role|
    $api_test.user.add_role(user, role)
  end
  add_context('user', user)
  add_context('password', 'linux')
  log "New user #{user} created"
end
```

NamelessOne91 (approved):

testsuite/features/support/env.rb
nodeg: I see. With the above linked step, that makes sense. Thank you for the clarification, Maxime.

nodeg (approved):

PR #9101, opened 2024-07-25
Parse and import both rpm and deb packages metadata

waterflow80:
## What does this PR change?

Parse both rpm and deb packages/repositories' metadata files and import the data to the database. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: still in progress

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


schema/spacewalk/common/tables/rhnPackage.sql
```diff
@@ -56,6 +56,7 @@ CREATE TABLE rhnPackage
                          CONSTRAINT rhn_package_compat_check
                              CHECK (compat in ( 1 , 0 )),
     path             VARCHAR(1000),
+    remote_path	     VARCHAR(1000),
```
aaannz, line 59: Is there a reason, except 'we did it like that before', to limit path to 1000 characters?
In postgresql varchar is implemented using TEXT datatype with string length check on top of it. So using limited VARCHAR is actually a tiny little slower then just using TEXT and has no space or indexing benefit.

So if there is no reason to limit this to 1000 characters, please use TEXT as a datatype here.

waterflow80: Absolutely..thanks for the review,
I did not have any specific reason for using `VARCHAR(1000)`...I'll change it to `TEXT` instead.

python/lzreposync/src/lzreposync/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+from lzreposync.import_utils import (
+    import_package_batch,
+    batched,
+    import_repository_packages_in_batch,
+)
+from lzreposync.rpm_repo import RPMRepo
+
+
+def main():
+    parser = argparse.ArgumentParser(
```
agraul, line 24: In the future, i.e. not in this PR, we should align the CLI interface as much as possible with `spacewalk-repo-sync`. We don't have to support all the flags, a subset is fine. Right now we have conflicting flags (e.g. `-d` means _debug_ here but _dry-run_ in `spacewalk-repo-sync`)

python/lzreposync/src/lzreposync/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+                f"Info: successfully created channel: {channel_label} -> id={channel.get_id()}, name={channel.get_label()}"
+            )
+        except ChannelAlreadyExistsException:
+            print(f"Warn: failed to create channel {channel_label}. Already exists !!")
+        return
+
+    arch = args.arch
+    if arch != ".*":
+        # pylint: disable-next=consider-using-f-string
+        arch = "(noarch|{})".format(args.arch)
```
agraul, line 141: Instead of silencing pylint, we should follow the rules we have set it for it. In this case it's as simple as:
```suggestion
        arch = f"(noarch|{args.arch})"
```

python/lzreposync/src/lzreposync/deb_metadata_parser.py
```diff
@@ -0,0 +1,70 @@
+#  pylint: disable=missing-module-docstring
+
+import logging
+import sys
+
+from lzreposync.packages_parser import PackagesParser
+from lzreposync.translation_parser import TranslationParser
+from spacewalk.satellite_tools.syncLib import log, log2
```
agraul, line 8: Just a heads-up: logging in reposync currently suffers from a split-brain problem. We have `rhnLog` (with `syncLib` defining wrapper functions) one the one hand and `logging` on the other. The latter is also used by 3rd-party libraries (e.g. `requests`). It's one of the things on our list to clean up, so it does not matter much what you use now. That said, I suggest to stick to one of them to make it easy for yourself.

python/lzreposync/src/lzreposync/deb_metadata_parser.py
```diff
@@ -0,0 +1,70 @@
...
+from spacewalk.server.importlib import mpmSource
+
+
+class BadParserException(Exception):
+    def __init__(self, parser):
+        super().__init__(f"Bad Parser {parser}")
+
+
+#  pylint: disable-next=missing-class-docstring
+class DEBMetadataParser:
```
agraul, line 19: This class only has one method, I would move that to a "free" function at the module level. Adding another layer makes the name longer but does not give us any benefit

python/lzreposync/src/lzreposync/deb_repo.py
```diff
@@ -0,0 +1,236 @@
...
+from uyuni.common import fileutils
+
+RETRIES = 10
+RETRY_DELAY = 1
+FORMAT_PRIORITY = [".xz", ".gz", ""]
+log = logging.getLogger(__name__)
+
+
+#  pylint: disable-next=missing-class-docstring
+class DebRepo(Repo):
```
agraul, line 27: We should consolidate the debian repo parsing. Right now it's split into `repo_plugins.deb_src`, `common.repo` and this `deb_repo` (which appears to be a copy of the class in `deb_src`

agraul (commented):
This is the first part, I will continue but want to send some comments already. Besides what I wrote, I suggest to go over the `pylint` comments and try to resolve the complaints instead of disabling the checks. `lzreposync` is a new program and should meet our current standards better than the old code :sweat_smile:

python/lzreposync/src/lzreposync/deb_metadata_parser.py
waterflow80: I wanted to add some abstraction when using OOP...
For example in function [import_packages_in_batch_from_parser()](https://github.com/uyuni-project/uyuni/blob/8a49313f35c903cb79926e6a5a3bcb7264b43612/python/lzreposync/src/lzreposync/import_utils.py#L89) (used in the tests) we're making use of this abstraction to import packages using abstract `parser` objects that implement the `parse_packages_metadata()` method.

And I'm afraid that using "free functions might add some complexity especially when passing a function as a parameter to another function instead of an object, in the above scenario.    

However I'm not sure whether it's the best way to approach things in our case.
(`pylint` also complains about using one method within a class `R0903: Too few public methods (1/2)`)

agraul: To use `import_packages_in_batch_from_parser`, you first instantiate a parser object. The parser object is only used to then call the single public method it has. In other words, at the call time of `import_packages_in_batch_from_parser`, the caller has to decide which parser to pass (rpm or deb). Therefore the caller could also pass specific functions (e.g. `parse_rpm_metadata()` or `parse_deb_metadata()`) to `import_packages_in_batch_from_parser`, or call these generator functions themselves and pass the generator object instead.

In simplified/pseudo code:
```python
# parsing code
def parse_rpm_metadata(primary_xml, repository_url, arch_filter, filelists_xml, cache_dir):
    primary_parser = PrimaryParser(primary_xml, repository_url, arch_filter)
    filelists_parser = FilelistsParser(filelists_xml, cache_dir, arch_filter)
    filelitsts_parser.parse_filelists()
    for package in primary_parser.parse():
        ...
        yield rpm_package

# import code
def import_in_batches(packages, batch_size, channel=None, compatible_archs=None):
    total = 0
    for i, batch in enumerate(batched(packages, batch_size)):
        failed, *_ = import_package_batch(
            to_process=batch,
            compatible_archs=compatible_archs,
            batch_index=i,
            to_link=True,
            channel=channel
        )
        total += failed
    return total

# test code
def test_parse_and_import():
    packages = parse_rpm_metadata(primary, repo_url, arch_filter, filelists)
    failed = import_in_batches(packages, 20)
    assert ...
```

I like this more than passing a function because we decouple the different concerns. Import code shouldn't need to bother with getting the data to import, that's someone else's job.
That said, in Python it's not a problem to pass functions around, they are first-class objects, e.g.:
```python
>>> import random
>>> def foo():
...     print("Hello, I'm Foo")
... 
>>> def maybe(f):
...     if random.randint(0, 1) == 1:
...        f() 
...     else:
...        print("No luck. Maybe next time")
... 
>>> for _ in range(10):
...     maybe(foo)
... 
Hello, I'm Foo
No luck. Maybe next time
No luck. Maybe next time
No luck. Maybe next time
No luck. Maybe next time
Hello, I'm Foo
No luck. Maybe next time
Hello, I'm Foo
Hello, I'm Foo
No luck. Maybe next time

waterflow80: Very good explanation..and the approach is much better too.
Thanks.

PR #9103, opened 2024-07-25
Fix error in 'kickstart_delete' when using wildcards (bsc#1227578)

cbbayburt:
In `kickstart_delete`, the code first resolves wildcard arguments and returns matching labels. Then it tries to make sure if every arg actually matched a result. But since the original wildcard args don't literally match the resulting labels, the second check throws an error. This PR removes the second check (only exists in this method) to match the method with the rest of the system.

See bug: https://bugzilla.suse.com/show_bug.cgi?id=1227578

## Documentation

## Test coverage
- Unit tests were added

## Links

Fixes https://github.com/SUSE/spacewalk/issues/24761

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/spacecmd/tests/test_kickstart.py
+++ b/spacecmd/tests/test_kickstart.py
@@ -374,9 +374,9 @@ echo 'some more hello'
         assert_args_expect(shell.do_kickstart_list.call_args_list,
                            [(('', True), {})])
 
-    def test_kickstart_delete_some_invalid_profile(self, shell):
+    def test_kickstart_delete_wildcard_arg(self, shell):
         """
-        Test do_kickstart_delete invalid profile (not found).
+        Test do_kickstart_delete with a wildcard argument
 
         :param shell:
         :return:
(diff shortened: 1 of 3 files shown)


spacecmd/tests/test_kickstart.py
```diff
@@ -374,9 +374,9 @@ def test_kickstart_delete_invalid_profile(self, shell):
         assert_args_expect(shell.do_kickstart_list.call_args_list,
                            [(('', True), {})])
 
-    def test_kickstart_delete_some_invalid_profile(self, shell):
```
agraul, line 377: Shouldn't we keep this test?

cbbayburt: No, this one tests for the case I removed in the code. Basically, if **some** of the args are not found, it used to throw an error, which didn't work with wildcards. Now it works with only the args it could match. The removed part doesn't exist in any other methods that deal with the globs in the same way, so I removed it from this as well.

agraul: How does the user know that some of the passed profiles were not found after your change?

meaksh (approved):

PR #9122, opened 2024-07-30
QE: Add retry mechanism for click preview button in salt steps

maximenoel8:
### Context
During BV smoke tests, we have a recurrent issue with the scenario `Run a remote command on <client> minion`. The step `I should see a "Target systems (1)" text` is failing because the search is not started.
I can see the hostname in the search field ( minion correctly exist ) but the search is not started ( the `Find targets` doesn't change to `Run Command`).
After some investigation, I found out clicking on `Find targets` is done too quickly.
Image

### Change
To improve the scenario stability, add a retry mechanism that make sure the search has been started.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24893

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/step_definitions/salt_steps.rb
+++ b/testsuite/features/step_definitions/salt_steps.rb
@@ -167,7 +167,27 @@ end
 
 # user salt steps
 When(/^I click on preview$/) do
-  find('button#preview').click
+  # Define the maximum number of attempts
+  max_attempts = 2
+  attempts = 0
+
+  while attempts < max_attempts
+    attempts += 1
+
+    # Click the preview button
+    find('button#preview').click
+
+    # Wait for up to 3 seconds for the run button to be visible
+    if page.has_button?('run', visible: true, wait: 3)
+      puts "The run button is visible."
+      return true
+    else
+      puts "The run button is not visible after clicking preview (attempt #{attempts})."
+    end
+  end
+
+  # If the run button is still not visible after the maximum attempts, raise an error
+  raise "Preview button not working: the run button is not visible after #{max_attempts} attempts."
 end
 
 When(/^I click on stop waiting$/) do


NamelessOne91 (dismissed):

nodeg (dismissed):
LGTM. Thank you for the improvement, Maxime.

maximenoel8:
I had a new error for the preview button : 
image
It means the button can have three ids : preview / stop ( Stop Waiting text) / run. When we have `stop` button, preview button is not clickable.
I added a check to not click `preview` when `stop` visible. I also increase a bit the wait time for `run` button.

maximenoel8:
I could also only check that button preview is visible that would cover run and stop states but I prefer to have an explanation if failing.

srbarrios:
I'm not fun of retrials, as they could hide issues.
Historically in this project we avoided this kind of strategy, I would like the opinion of @Bischoff on that change.

maximenoel8:
I understand the risks of retry actions. But if you look at the PR, it's not just a retry after few seconds but a verification of the current stat ( preview / stop / run ). In this situation, I didn't find a way of checking that we are ready to click on the button. 
The field is correctly fill, the button is visible and clickable and if you run the scenario alone or do it manually, you will never have an issue with it. 

If we do nothing, we will still have half the smoke tests failing for a testsuite issue.
The other solution I see is a sleep before clicking ( ugly in my point of view ) or slowing down the clicks.

srbarrios:
Maybe we can contact with devs, explain them this behavior, and understand the reason why this button doesn't reacts properly on the first click. Knowing the reason, we might be able to find a way to wait until a pre-condition is done, before clicking there.

srbarrios (approved):
We had a huddle, we agreed on merging this as a quick workaround, we will add a comment in the code to warn this is a workaround, and this comment will include a link to a GH Card, to analyze it deeper in parallel.

srbarrios (approved):

srbarrios:
The card in the comment https://github.com/SUSE/spacewalk/issues/24893 has been closed together with that merge. That's not what we agreed in order to approve and merge this PR.

What we agreed in our huddle is to have a GH card where @maximenoel8 will analyze deeper this issue but having this workaround in-place to don't have a failure in cascade on next BV.

Bischoff:
@maximenoel8 Yeah, as Oscar said, I'm not a big fan of retrying actions.

However, this is merged. Let's see how it evolves.

maximenoel8:
It will improve the BV stability and improve the result visibility during parallel testing. If you can prove me that this *SPECIFIC* function is going to hide an issue, I'm happy to discuss.

srbarrios:
In our discussion on huddle, we had been discussing about it, but better if I wrote it here:

From your comment in the issue:

_> During BV smoke tests, we have a recurrent issue with the scenario Run a remote command on <client> minion. The step I should see a "Target systems (1)" text is failing because the search is not started.
> I can see the hostname in the search field ( minion correctly exist ) but the search is not started ( the Find targets doesn't change to Run Command).
> After some investigation, I found out clicking on Find targets is done too quickly._

It's easy to say that the product should not show as clickable or accessible that button, as the comment says the search is not started. So, instead of re-try an action of clicking the button until this button shows the expected behavior, we need to discuss with the development team so this button is not clickable or accessible from the WebUI, showing another kind of text or some helper for the user, so the user can't click on this button as the result it will not be the expected yet.

PR #9168, opened 2024-08-13
Partial revert of f21851be1e50

Bischoff:
## What does this PR change?

Missing call to "make" in the profiles.

## Links

No ports, Uyuni only.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/profiles/cloud_aws/Docker/Dockerfile
+++ b/testsuite/features/profiles/cloud_aws/Docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ip-172-16-1-175.eu-central-1.compute.internal/opensuse/leap:15.5
+FROM ip-172-16-1-175.eu-central-1.compute.internal/opensuse/leap:15.4
 MAINTAINER Admin User "noemail@example.com"
 
 ARG repo
--- a/testsuite/features/profiles/internal_nue/Docker/Dockerfile
+++ b/testsuite/features/profiles/internal_nue/Docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.mgr.suse.de/opensuse/leap:15.5
+FROM registry.mgr.suse.de/opensuse/leap:15.4
 MAINTAINER Admin User "noemail@example.com"
 
 ARG repo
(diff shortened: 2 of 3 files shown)


NamelessOne91 (approved):

nodeg (commented):
Is this really correct? The [last change](https://github.com/uyuni-project/uyuni/commits/master/testsuite/features/profiles/internal_nue/Docker/Dockerfile) is in place since 1 year, and we only notice issues with it now?

Bischoff:
> Is this really correct? The [last change](https://github.com/uyuni-project/uyuni/commits/master/testsuite/features/profiles/internal_nue/Docker/Dockerfile) is in place since 1 year, and we only notice issues with it now?

I think that it has small impact.

Anyway, the fix was tested with success.

nodeg:
> [...]  if you look at Docker file, you will notice that it was half 15.4, half 15.5 😸 .

I do not see it actually. See e.g.:
https://github.com/uyuni-project/uyuni/blob/42a654be0a93c5b33817cd41d10b03f0f10dd79e/testsuite/features/profiles/internal_nue/Docker/Dockerfile#L1-L16

nodeg:
But I see the Dockerfile 2 folders above still using Leap 15.4:
https://github.com/uyuni-project/uyuni/blob/42a654be0a93c5b33817cd41d10b03f0f10dd79e/testsuite/features/profiles/Docker/Dockerfile#L1-L16

jordimassaguerpla:
Hi! Both solutions look reasonable to me. You might want to do it in 2 steps, first the revert, so tests pass, and then update the docker images to 15.5. Whatever makes more sense to you.

Bischoff:
> I do not see it actually. See e.g.:

line 1:  `FROM registry.mgr.suse.de/opensuse/leap:15.5 `
line 13:  `ADD sles15sp4.repo /etc/zypp/repos.d/sles15sp4.repo `

Bischoff:
> But I see the Dockerfile 2 folders above still using Leap 15.4:

yes, these are the "sources" for the two files changed in this PR

when you run "make" the ones in internal_nue, internal_prv, and internal_aws are derived from these


> So why not bump that file to Leap 15.5, too instead of going back to 15.4?

There's an objective reason for not bumping Kiwi files: the build host and terminals have to be of same version as those templates, and currently the build host is 15.4

With Docker we just align to the version for Kiwi - that choice might be discussed, but I would find simpler to bump everything to 15.6 or 15.7 when we have time.

Bischoff:
> Hi! Both solutions look reasonable to me. You might want to do it in 2 steps, first the revert, so tests pass, and then update the 
> docker images to 15.5. Whatever makes more sense to you.

I would rather first do the revert, then update all templates (including Kiwi, build host, and terminal) directly to 15.6 or 15.7, to spare work.

nodeg (approved):

PR #9182, opened 2024-08-20
enforce directory permissions at repo-sync when creating directories

mcalmer:
## What does this PR change?

When using umask 077, reposync create directories with 0700 permissions which prevent tomcat from reading these directories.
os.mkdir seems to apply the umask on the given mode. This require an extra os.chmod to enforce the permissions

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **manual**

- [x] **DONE**

## Links

Port(s): https://github.com/SUSE/spacewalk/pull/25109 https://github.com/SUSE/spacewalk/pull/25110

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/uyuni/common/fileutils.py
+++ b/python/uyuni/common/fileutils.py
@@ -312,6 +312,7 @@ def makedirs(path, mode=int("0755", 8), user=None, group=None):
         dirname = dirs_to_create.pop()
         try:
             os.mkdir(dirname, mode)
+            os.chmod(dirname, mode)
         except OSError:
             e = sys.exc_info()[1]
             if e.errno != 17:  # File exists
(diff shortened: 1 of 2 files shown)


mcalmer:
> It's strange to me that we need this (though maybe I misunderstand). Per https://docs.python.org/3/library/os.html#os.mkdir, mkdir should first subtract the umask from the mode (so 700 - 077 should still be 700), which should give the full perms to the user.

Well, but `700` does not work. If the permissions are 700 (which they are), tomcat cannot verify anymore if the file exists or not.
The default is 755 and if you now have a umask of 077, you end up with 700 which break Uyuni.

I wonder more, if I tell python to create a directory with permissions "755", why does it think to apply magic to it and change the explicit value I have selected. But it is at least documented that it happens :-)

m-czernek:
Aah, understood. So you have `755` default, `755 - 077` = `700`, which breaks tomcat (which relies on group perms probably?). Makes sense, then it is a documented behavior :) (though unexpected, I agree :/ )

PR #9188, opened 2024-08-21
Add custom krb5.conf which moved keytab under krb5.conf.d

aaannz:
## What does this PR change?

See title.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- Documentation issue was created: https://github.com/SUSE/spacewalk/issues/25033

- [X] **DONE**

## Test coverage
- No tests: not testing kerberos integration

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25033

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/server-image/Dockerfile
+++ b/containers/server-image/Dockerfile
@@ -122,6 +122,11 @@ RUN uyuni-configfiles-sync init /srv/tftpboot/
 RUN uyuni-configfiles-sync init /srv/www/
 RUN uyuni-configfiles-sync init /var/lib/cobbler/
 
+# Set /etc/krb5.conf.d/ as a link to krb5
+COPY krb5.conf /etc/krb5.conf
+COPY krb5-conf-setup.sh /usr/bin/krb5-conf-setup.sh
+RUN /bin/bash krb5-conf-setup.sh
+
 # LABELs
 ARG PRODUCT=Uyuni
 ARG VENDOR="Uyuni project"
--- a/containers/server-image/krb5-conf-setup.sh
+++ b/containers/server-image/krb5-conf-setup.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+set -e
+
+KRB5CONFD="krb5.conf.d"
+
+rmdir "/etc/$KRB5CONFD"
+mkdir "/etc/rhn/$KRB5CONFD"
+ln -s "/etc/rhn/$KRB5CONFD" "/etc/$KRB5CONFD"
--- a/containers/server-image/krb5.conf
+++ b/containers/server-image/krb5.conf
@@ -1,0 +1,32 @@
+includedir  /etc/rhn/krb5.conf.d
+
+[libdefaults]
+    # "dns_canonicalize_hostname" and "rdns" are better set to false for improved security.
+    # If set to true, the canonicalization mechanism performed by Kerberos client may
+    # allow service impersonification, the consequence is similar to conducting TLS certificate
+    # verification without checking host name.
+    # If left unspecified, the two parameters will have default value true, which is less secure.
+    dns_canonicalize_hostname = false
+    rdns = false
+    # "verify_ap_req_nofail" is enabled to protect against KDC spoofing. After obtaining the
+    # initial credentials the client library will attempt to verify if the KDC that issued them
+    # is the same that issued the keys stored in the local keytab. If the client machine does
+    # not have a keytab, it cannot be read or there is no host key in it, the verification will
+    # fail if "verify_ap_req_nofail" is set to true. If it is set to false and the client machine
+    # does not have a keytab, the verification is skipped.
+    verify_ap_req_nofail = true
+#   default_realm = EXAMPLE.COM
+    default_ccache_name = KEYRING:persistent:%{uid}
+    # move default keytab to the persistent /etc/rhn/krb5.conf.d directory
+    default_keytab_name = FILE:/etc/rhn/krb5.conf.d/krb5.keytab
+
+[realms]
+#       EXAMPLE.COM = {
+#                kdc = kerberos.example.com
+#               admin_server = kerberos.example.com
+#       }
+
+[logging]
+    kdc = FILE:/var/log/krb5/krb5kdc.log
+    admin_server = FILE:/var/log/krb5/kadmind.log
+    default = SYSLOG:NOTICE:DAEMON
--- a/containers/server-image/server-image.changes.oholecek.krb5-persistent-fix
+++ b/containers/server-image/server-image.changes.oholecek.krb5-persistent-fix
@@ -1,0 +1,2 @@
+- use /etc/krb5.conf.d for all kerberos related configurations
+  (bsc#1229077)


rjmateus (approved):

cbosdo (approved):

mbussolotto (approved):

ppanon2022:
Upgraded to 2024.10 to pick up this PR. That got the symbolic link /etc/krb5.conf.d=>/etc/rhn/krb5.conf.d, but it took me a while to figure out that adcli was still breaking because it was a broken link since /etc/rhn/krb5.conf.d didn't exist. After I created the latter directory, it mostly worked in that I was successfully able to run adcli to join the domain. 

adcli join -vvv -D mydomain -U myadminaccount -S dcname.mydomain -N MYHOSTNAME -H MYHOSTNAME.domain -K /etc/krb5.conf.d/krb5.keytab -W --domain-ou "OU=Computersou..."

to use the actual host's name and not the container's "uyuni-server" hostname. Then using ad_hostname = in sssd.conf to match.
getent passwd and getent group work fine after that, but Uyuni portal authentication  still fails with the following error in the krb5_child log.

```
(2024-10-24 22:28:16): [krb5_child[3078]] [create_ccache] (0x0020): [RID#3] 1432: [38][Function not implemented]
********************** PREVIOUS MESSAGE WAS TRIGGERED BY THE FOLLOWING BACKTRACE:
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [main] (0x0400): [RID#3] krb5_child started.
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [unpack_buffer] (0x1000): [RID#3] total buffer size: [164]
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [unpack_buffer] (0x0100): [RID#3] cmd [241 (auth)] uid [640746932] gid [932800513] validate [true] enterprise principal [true] offline [false] UPN [t1_ppanon@SIERRAWIRELESS.LOCAL]
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [unpack_buffer] (0x2000): [RID#3] No old ccache
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [unpack_buffer] (0x0100): [RID#3] ccname: [KEYRING:persistent:640746932] old_ccname: [not set] keytab: [/etc/krb5.conf.d/krb5.keytab]
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [check_use_fast] (0x0100): [RID#3] Not using FAST.
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [k5c_precreate_ccache] (0x4000): [RID#3] Recreating ccache
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [become_user] (0x0200): [RID#3] Trying to become user [640746932][932800513].
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [main] (0x2000): [RID#3] Running as [640746932][932800513].
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [set_lifetime_options] (0x0100): [RID#3] No specific renewable lifetime requested.
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [set_lifetime_options] (0x0100): [RID#3] No specific lifetime requested.
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [set_canonicalize_option] (0x0100): [RID#3] Canonicalization is set to [true]
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [main] (0x0400): [RID#3] Will perform auth
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [main] (0x0400): [RID#3] Will perform online auth
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [tgt_req_child] (0x1000): [RID#3] Attempting to get a TGT
   *  (2024-10-24 22:28:14): [krb5_child[3078]] [get_and_save_tgt] (0x0400): [RID#3] Attempting kinit for realm [SIERRAWIRELESS.LOCAL]
   *  (2024-10-24 22:28:15): [krb5_child[3078]] [sss_krb5_responder] (0x4000): [RID#3] Got question [password].
   *  (2024-10-24 22:28:15): [krb5_child[3078]] [sss_krb5_expire_callback_func] (0x2000): [RID#3] exp_time: [11235334]
   *  (2024-10-24 22:28:15): [krb5_child[3078]] [validate_tgt] (0x2000): [RID#3] Found keytab entry with the realm of the credential.
   *  (2024-10-24 22:28:16): [krb5_child[3078]] [validate_tgt] (0x0400): [RID#3] TGT verified using key for [CARMD-NV-UYUNI1$@SIERRAWIRELESS.LOCAL].
   *  (2024-10-24 22:28:16): [krb5_child[3078]] [sss_send_pac] (0x4000): [RID#3] NSS return code [1], request return code [0][Success].
   *  (2024-10-24 22:28:16): [krb5_child[3078]] [sss_send_pac] (0x0400): [RID#3] PAC responder contacted. It might take a bit of time in case the cache is not up to date.
   *  (2024-10-24 22:28:16): [krb5_child[3078]] [get_and_save_tgt] (0x2000): [RID#3] Running as [640746932][932800513].
   *  (2024-10-24 22:28:16): [krb5_child[3078]] [sss_get_ccache_name_for_principal] (0x4000): [RID#3] Location: [KEYRING:persistent:640746932]
   *  (2024-10-24 22:28:16): [krb5_child[3078]] [sss_get_ccache_name_for_principal] (0x2000): [RID#3] krb5_cc_cache_match failed: [38][Function not implemented]
   *  (2024-10-24 22:28:16): [krb5_child[3078]] [create_ccache] (0x0020): [RID#3] 1432: [38][Function not implemented]
********************** BACKTRACE DUMP ENDS HERE *********************************

(2024-10-24 22:28:16): [krb5_child[3078]] [map_krb5_error] (0x0020): [RID#3] 2399: [38][Function not implemented]

```

aaannz:
Indeed I missed the upgrade scenario. So `/etc/rhn` is mounted from the volume. Thanks for the report.

aaannz:
However one note. This fix was done by me when I did not fully understood supported scenario. Using standalone krb5 is not officially supported, instead krb5 using sssd should be used. ( https://www.uyuni-project.org/uyuni-docs/en/uyuni/administration/auth-methods-pam.html )

I have no particular experience here, but can you try to switch over to sssd configuration? See also https://manpages.opensuse.org/Tumbleweed/sssd-krb5/sssd-krb5.5.en.html

ppanon2022:
Well, that auth-methods-pam.html page is very light on details, but I am indeed using it with sssd. Placing the krb5.keytab file in /etc/sssd is very non-standard though, so I do much prefer the /etc/krb5.conf.d => /etc/rhn/krb5.conf.d approach you've taken.

Those krb5_child log entries are from the /var/log/sssd/krb5_child.log file for that sssd process (thread?). The domain section in the sssd.conf file looks like the following

[domain/domain.fqdn]
default_shell = /bin/bash
krb5_store_password_if_offline = True
cache_credentials = True
krb5_realm = REALM.NAME
#realmd_tags = manages-system joined-with-adcli
id_provider = ad
fallback_homedir = /home/%u@%d
ad_domain = domain.fqdn
ad_hostname = HOSTNAME
dyndns_update = false
krb5_keytab = /etc/krb5.conf.d/krb5.keytab
#use_fully_qualified_names = True
use_fully_qualified_names = False
ldap_id_mapping = True
ad_site = ADSITENAME
ldap_sasl_mech = GSS-SPNEGO
access_provider = simple
simple_allow_groups = comma-separated-list-of-Uyuni-user-groups

with actual system-specific values changed to generic descriptions. That section is based on what is created by the realm command on other systems, with some manual additions (ad_hostname, ad_site, dyndns_update, ldap_sasl_mech)

I see sssd.conf has settings for krb5_ccachedir and krb5_ccname_template. Even though that page indicates krb5_ccachedir defaults to /tmp, could it be that it's actually trying to use a problematic unwriteable/non-persisting directory and the create_ccache error is just an incorrect exception handling report? If so, what should we be using for krb5_ccachedir? Does anybody have this working?

aaannz:
Finally managed to resolve this. The underlying problem is that our default ccache is not namespace aware. We have as `default_ccache_name = KEYRING:persistent:%{uid}` however this does not work inside container. As a workaround you can set this to `default_ccache_name = FILE:/tmp/krb5.ccache` for example.

The rest should work as was before with 4.3.

Note that there may be also needed a change in `/etc/pam.d/susemanager*` files which contains `pam_ldap.so` which is however no longer shipped as it was deprecated. Unfortunately changes to these files are not persisted, so after container reset it will revert back. See https://github.com/uyuni-project/uyuni/pull/9422

I will create new PRs with required changes and also new docs.

ppanon2022:
Thank you for your perseverance. I've tried to change my pam.d susemanager conf to match https://github.com/uyuni-project/uyuni/pull/9422/commits/b4da4583dfaf5c0feda899a5d376f1d8637c75f9, as well as doing an sssd restart after adding default_ccache_name = FILE:/tmp/krb5.ccache to sssd.conf. I then tried to do a spacewalk-service restart in case it was needed to load the new pam config but I'm still getting some errors. Could you show which domain settings you are using in your sssd.conf (masking/anonymizing any sensitive internal values)?

aaannz:
Sure, I am in the process of writing new docs with that examples anyway:

New `/etc/krb5.conf` which will be part of the image:
```
# cat /etc/krb5.conf
includedir  /etc/rhn/krb5.conf.d

[libdefaults]
    # "dns_canonicalize_hostname" and "rdns" are better set to false for improved security.
    # If set to true, the canonicalization mechanism performed by Kerberos client may
    # allow service impersonification, the consequence is similar to conducting TLS certificate
    # verification without checking host name.
    # If left unspecified, the two parameters will have default value true, which is less secure.
    dns_canonicalize_hostname = false
    rdns = false
    # "verify_ap_req_nofail" is enabled to protect against KDC spoofing. After obtaining the
    # initial credentials the client library will attempt to verify if the KDC that issued them
    # is the same that issued the keys stored in the local keytab. If the client machine does
    # not have a keytab, it cannot be read or there is no host key in it, the verification will
    # fail if "verify_ap_req_nofail" is set to true. If it is set to false and the client machine
    # does not have a keytab, the verification is skipped.
    verify_ap_req_nofail = true
    # KEYRING type cannot be used inside container
    default_ccache_name = FILE:/tmp/krb5.ccache
    # move default keytab to the persistent /etc/rhn/krb5.conf.d directory
    default_keytab_name = FILE:/etc/rhn/krb5.conf.d/krb5.keytab

[realms]

[logging]
    kdc = FILE:/var/log/krb5/krb5kdc.log
    admin_server = FILE:/var/log/krb5/kadmind.log
    default = SYSLOG:NOTICE:DAEMON
```

My `/etc/sssd/sssd.conf` with AD domain:
```
# cat /etc/sssd/sssd.conf
[sssd]
config_file_version = 2
services = nss, pam
# SSSD will not start if you do not configure any domains.
# Add new domain configurations as [domain/<NAME>] sections, and
# then add the list of domains (in the order you want them to be
# queried) to the "domains" attribute below and uncomment it.
#domains = LDAP, KRB5
domains = KRB5

[nss]

[pam]

[domain/KRB5]
id_provider = ad
chpass_provider = ad
access_provider = ad
auth_provider = ad

ad_domain = ad.example.com
ad_server = addc15.ad.example.com
ad_hostname = test-server.uyuni.net

ad_gpo_map_network = +susemanager
```

I am not expert in this field so I am not sure about `ad_gpo_map_network = +susemanager`. It would not work without it, but I don't have any policies in AD itself.

Also note `ad_hostname` option. Without it, it will use `uyuni-server.mgr.internal` which might conflict if you have multiple uyuni instances.

Also joining the domain I used:
```
adcli join -D ad.example.com -H test-server.uyuni.net -S addc15.ad.example.com
```

ppanon2022:
Ah,  I was placing default_ccache_name = FILE:/tmp/krb5.ccache in sssd.conf instead of krb5.conf. Oops. We're using a lot more options in sssd.conf but it's a superset of those (except for using ad_site instead of ad_server), and after adding default_ccache_name  and ad_gpo_map_network  in their respective config files. It now works. Looking forward to the next release that includes your PR to make it permanent.

Thank you very much for your efforts in finding the solution.

aaannz:
Thanks for confirming it's working for you.

Indeed `default_ccache_name` is `[libdefaults]` section. To have that in sssd.conf under domain, `krb5_ccname_template` should be used AFAIK.

ppanon2022:
Thank you. What is the base interface for sssd? Is it named pipes, loopback networking, or something else? Can the ssh sever process on the host use the sssd running in the container? Or is the sssd in the container unusable from the host? https://sssd.io/docs/architecture.html talks about a wire protocol between the client libraries and the responders, but it doesn't say what the "wire" is. I'm not seeing any obvious TCP ports so I'm guessing it's something which normally wouldn't cross the container boundary, but perhaps you've got it mapped somehow? 

I'm expecting running an sssd service both in the container and the host probably would conflict with each other in talking to the AD domain controllers even if you used different machine names/accounts due to callbacks for machine account password changes if nothing else. Is there a way to make sshd AD also work or is it mutually exclusive with Uyuni portal sssd auth?

aaannz:
I'm sorry, but I don't know the answers to this yet. For now I expect that sssd will need to run on both host and inside container.
But it is a good question and we'll look into it later.

ppanon2022:
Thanks. I think it might be challenging to have both the host and the container run sssd for AD integration. I think domain members need to run a listener so that they can carry out callback operations like the machine account password change that are initiated by the domain controller. Since the container ports are mapped onto the host ports, I expect that both the host and the container would try to use the same port for the listener and result in a conflict. Hopefully there's a way to map whatever responder interfaces the sssd service/process exposes to the host.

ppanon2022:
Hmm. Have you tried to have more than one user login? We might need something like the following for the template:
FILE:/tmp/krb5.ccache.%{uid}

I'm getting these errors in the service status
Nov 06 12:23:11 uyuni-server.mgr.internal krb5_child[31106]: Internal credentials cache error
Nov 06 12:23:11 uyuni-server.mgr.internal krb5_child[31106]: Internal credentials cache error
I think the issue is that without that %{uid} extension, the file is owned by the first user/group to log in and subsequent users don't have permissions to it. Adding the extension should create a different cache file for each user.

aaannz:
You are absolutely right. Thanks for catching this. I indeed tried only single user. I looked into actual defaults and according to [doc](https://web.mit.edu/kerberos/krb5-1.20/doc/admin/conf_files/krb5_conf.html#libdefaults) default is `DEFCCNAME` which according to [paths](https://web.mit.edu/kerberos/krb5-1.20/doc/mitK5defaults.html#paths) is `FILE:/tmp/krb5cc_%{uid}`. So I will use this and set it explicitly in the configs.

PR #9204, opened 2024-08-22
QE: Add tests for channels.appstreams API namespace

NamelessOne91:
## What does this PR change?

**WIP**

This PR depends on having a proper modular repository in place

It adds API namespace support for the currently available AppStreams related methods and feature to test them.

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


maximenoel8 (approved):
LGTM, thanks

testsuite/features/step_definitions/api_common.rb
```diff
@@ -630,3 +630,32 @@
 When(/^I delete profile and distribution using the API for "([^"]*)" kickstart tree$/) do |distro_name|
   $api_test.kickstart.tree.delete_tree_and_profiles(distro_name)
 end
+
+When(/I verify channel "([^"]*)" is( not)? modular via the API/) do |channel_label, not_modular|
```
vandabarata, line 634: is it nitpicky to ask for `is (not)?` or did it have to be this way due to some constraint?

testsuite/features/step_definitions/api_common.rb
```diff
@@ -630,3 +630,32 @@
...
+  is_modular = $api_test.channel.appstreams.modular?(channel_label)
+  expected = not_modular.nil?
+
+  raise ScriptError "channel '#{channel_label}' is modular? Expected: #{expected} - got: #{is_modular}" unless is_modular == expected
+end
+
+When(/channel "([^"]*)" is( not)? present in the modular channels listed via the API/) do |channel, not_present|
```
vandabarata, line 641: same as above comment

testsuite/features/step_definitions/api_common.rb
NamelessOne91: It gets a bit tricky with whitespaces.

`is (not)? modular` would expect two of them between `is` and `modular` when we do not write the `not`

I don't like it either, but I did not want to add other step definitions just to negate the results

vandabarata (approved):

srbarrios (approved):

PR #9205, opened 2024-08-23
QE: Fix wrong return value in scp method

srbarrios:
## What does this PR change?

Fixing remaining issue on the new Remote Node class

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- Cucumber tests were fixed

- [x] **DONE**

## Links

No ports, for this particular PR, as all will go in a unique port from qe-develop.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/support/remote_node.rb
+++ b/testsuite/features/support/remote_node.rb
@@ -30,6 +30,8 @@ class RemoteNode
     end
 
     @target = ENV.fetch(ENV_VAR_BY_HOST[@host], nil).to_s.strip
+    # Remove /etc/motd, or any output from run will contain the content of /etc/motd
+    ssh('rm -f /etc/motd && touch /etc/motd', host: @target)
     out, _err, _code = ssh('hostname', host: @target)
     @hostname = out.strip
     raise LoadError, "We can't connect to #{@host} through SSH." if @hostname.empty?
@@ -38,6 +40,8 @@ class RemoteNode
     if @host == 'server'
       _out, _err, code = ssh('which mgrctl', host: @target)
       @has_mgrctl = code.zero?
+      # Remove /etc/motd inside the container, or any output from run will contain the content of /etc/motd
+      run('rm -f /etc/motd && touch /etc/motd')
       out, _code = run('sed -n \'s/^java.hostname *= *\(.\+\)$/\1/p\' /etc/rhn/rhn.conf')
     else
       out, _err, _code = ssh('hostname -f', host: @target)
@@ -45,8 +49,6 @@ class RemoteNode
     @full_hostname = out.strip
     raise StandardError, "No FQDN for '#{@hostname}'. Response code: #{code}" if @full_hostname.empty?
 
-    # Remove /etc/motd, or any output from run will contain the content of /etc/motd
-    run('rm -f /etc/motd && touch /etc/motd')
     $stdout.puts "Host '#{@host}' is alive with determined hostname #{@hostname} and FQDN #{@full_hostname}" unless $build_validation
     @os_version, @os_family = get_os_version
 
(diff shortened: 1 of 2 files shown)


testsuite/features/support/remote_node.rb
```diff
@@ -167,47 +167,47 @@ def wait_while_process_running(process)
...
-  def inject(local_file, remote_file)
+  def inject(test_runner_file, remote_node_file)
     if @has_mgrctl
       tmp_folder, _code = run_local('mktemp -d')
-      tmp_file = File.join(tmp_folder.strip, File.basename(local_file))
-      code, _remote = scp(local_file, tmp_file)
-      if code.zero?
-        _out, code = run_local("mgrctl cp --user #{user} #{tmp_file} server:#{remote_file}")
+      tmp_file = File.join(tmp_folder.strip, File.basename(test_runner_file))
+      success = get_target('localhost').scp(test_runner_file, tmp_file, host: self)
```
nodeg, line 179: This command is basically the same as below in the `else` condition. I wonder if we do not need to use a `mgrctl` command like before? How does it know it is in the container? Is it due to the fact, we run this `.inject()` method on a server node object? Then the `localhost` value would make sense.

testsuite/features/support/remote_node.rb
```diff
@@ -167,47 +167,47 @@ def wait_while_process_running(process)
...
+  # @param remote_node_file [String] The path in the destination.
   # @return [Integer] The exit code.
-  def inject(local_file, remote_file)
+  def inject(test_runner_file, remote_node_file)
     if @has_mgrctl
       tmp_folder, _code = run_local('mktemp -d')
-      tmp_file = File.join(tmp_folder.strip, File.basename(local_file))
-      code, _remote = scp(local_file, tmp_file)
-      if code.zero?
-        _out, code = run_local("mgrctl cp --user #{user} #{tmp_file} server:#{remote_file}")
```
nodeg, line 181: Does this really need to be removed?

Bischoff (approved):
You're starting to discover the issues that twopence protected you against :) .

LGTM

testsuite/features/support/remote_node.rb
srbarrios: This line?
`        _out, code = run_local("mgrctl cp --user #{user} #{tmp_file} server:#{remote_file}")
`

testsuite/features/support/remote_node.rb
srbarrios: So, on the first case, we want to copy a file from the controller to the "container" **host**, and then we move this file from the "container" **host** to inside the container.
In the second case, we copy a file from the controller to the host.

srbarrios: `get_target('localhost')` will return a node that will point to our controller or test runner (if you run this in your local IDE)

testsuite/features/support/remote_node.rb
nodeg: Yes. But I am not sure. I only saw that you kept it in the `extract()` method below.

testsuite/features/support/remote_node.rb
nodeg: Ok, that makes sense. I somehow thought it worked the other way around. Thanks.

vandabarata (approved):

testsuite/features/support/remote_node.rb
srbarrios: I moved it to this line:
```
      if success
        _out, code = run_local("mgrctl cp --user #{user} #{tmp_file} server:#{remote_node_file}")
```

nodeg: Ah, yes. Sorry my bad. I missed that. All good then.

srbarrios:
> =~

!! Rubocop doesn't complain about it, weird.

nodeg:
As already discussed in Slack, yes, we can use Ruby 3.1 for now, until `net-ssh` has proper support for newer Ruby versions. One thing to keep in mind, Ruby 3.1 will already be EOL after 31.03.2025 and currently only receives security fixse.

PR #9229, opened 2024-08-30
Fixes related to RemoteNode implementation

srbarrios:
## What does this PR change?

- Add the possibility to force the API protocol that we want to use during a test execution
- Add an extra exception to handle unreachable hosts
- Fix an issue handle the exit code of a ssh call.
- Refactor some remaining calls to Twopence methods
- Remove a not-used step definition

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

No ports. It will be ported all together with qe-develop branch.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -64,6 +64,7 @@ $is_using_build_image = ENV.fetch('IS_USING_BUILD_IMAGE', false)
 $is_using_scc_repositories = (ENV.fetch('IS_USING_SCC_REPOSITORIES', 'False') != 'False')
 $catch_timeout_message = (ENV.fetch('CATCH_TIMEOUT_MESSAGE', 'False') == 'True')
 $beta_enabled = (ENV.fetch('BETA_ENABLED', 'False') == 'True')
+$api_protocol = ENV.fetch('API_PROTOCOL', nil) if ENV['API_PROTOCOL'] # force the API protocol to be used. You can use 'http' or 'xmlrpc'
 
 # QAM and Build Validation pipelines will provide a json file including all custom (MI) repositories
 custom_repos_path = "#{File.dirname(__FILE__)}/../upload_files/custom_repositories.json"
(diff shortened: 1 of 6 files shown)


testsuite/features/support/env.rb
```diff
@@ -64,6 +64,7 @@
 $is_using_scc_repositories = (ENV.fetch('IS_USING_SCC_REPOSITORIES', 'False') != 'False')
 $catch_timeout_message = (ENV.fetch('CATCH_TIMEOUT_MESSAGE', 'False') == 'True')
 $beta_enabled = (ENV.fetch('BETA_ENABLED', 'False') == 'True')
+$api_protocol = ENV.fetch('API_PROTOCOL', nil) if ENV['API_PROTOCOL'] # force the API protocol to be used. You can use 'http' or 'xmlrpc'
```
ktsamis, line 67: I would set this to one of the two by default instead of nil. To avoid cases where it's empty and will cause problems down the line. Probably xmlrpc as it was before. It would allow the possibility to use http if we want without changing the current setup

ktsamis (changes requested):

testsuite/features/support/env.rb
srbarrios: When this is `nil` it will fall into the `else` section of the `case` statement.
On such case, it will decide which protocol it will use depending the product version, as it was behaving before adding this variable to force the protocol.

The root cause of this PR is that some commits ago, we had defined that if `$debug_mode` was true, then we use XMLRPC API (no SSL certificates needed), and that allowed me to run my tests from my IDE without worry on SSL certificates. We removed this option, because at some point Dominik need to debug a test case using HTTP API with SSL.

To keep being able to debug on HTTP API, but please me to work without SSL certificates in my local environment... I added this variable to force the protocol, if the variable is defined.

Bischoff (approved):
Why not.

I found it simpler to hardcode by branch (master = http/json, 5.0 and 4.3 = http/xml), but that's indeed more flexible.

testsuite/features/support/env.rb
ktsamis: ok thanks for the explanation :+1:

ktsamis (approved):

vandabarata (approved):

PR #9290, opened 2024-09-24
QE: Fix path for openscap

vandabarata:
## What does this PR change?
openSCAP tests are failing due to a filename change - this fixes that.


NamelessOne91 (approved):

ktsamis (approved):

mcalmer:
@vandabarata please check. I think more tests are failing because of this with other OSes. There is also the path incorrect.

vandabarata:
> @vandabarata please check. I think more tests are failing because of this with other OSes. There is also the path incorrect. Check openSCAP tests for SLES and Ubuntu

@mcalmer the other failing tests aren't directly related to the path itself, but to other flaky parts of the testsuite. 
This one seems to be the only one path-related. As for the right path, I'm not sure how to get there so I followed your suggestion and was simply going to confirm on the next CI runs

PR #9340, opened 2024-10-10
Add a step to repository synchronization feature to check reposync logs.

maximenoel8:
## What does this PR change?
Add the log verification step at the end of the repository synchronization features.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Port :

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
+++ b/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
@@ -821,3 +821,6 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I wait until I do not see "Loading" text
     Then I should not see a "Operation not successful" text
     And I should not see a warning nor an error sign
+
+  Scenario: Detect repository synchronization issues in reposync logs
+    Then the reposync logs should not report errors
--- a/testsuite/features/reposync/srv_sync_products.feature
+++ b/testsuite/features/reposync/srv_sync_products.feature
@@ -179,6 +179,9 @@ Feature: Synchronize products in the products page of the Setup Wizard
     Then I should not see a "Operation not successful" text
     And I should not see a warning nor an error sign
 
+  Scenario: Detect repository synchronization issues in reposync logs
+    Then the reposync logs should not report errors
+
 @scc_credentials
   Scenario: Trigger a refresh of the products synched from SCC
     When I execute mgr-sync refresh


testsuite/features/build_validation/reposync/srv_sync_all_products.feature
```diff
@@ -821,3 +821,6 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I wait until I do not see "Loading" text
     Then I should not see a "Operation not successful" text
     And I should not see a warning nor an error sign
+
+  Scenario: Detect repository synchronization issues in reposync logs
+    Then the reposync logs should not report errors
```
ktsamis, line 826: I think scenarios should start with When no?
```suggestion
    When the reposync logs should not report errors
```

testsuite/features/reposync/srv_sync_products.feature
```diff
@@ -179,6 +179,9 @@ Feature: Synchronize products in the products page of the Setup Wizard
     Then I should not see a "Operation not successful" text
     And I should not see a warning nor an error sign
 
+  Scenario: Detect repository synchronization issues in reposync logs
+    Then the reposync logs should not report errors
```
ktsamis, line 183:
```suggestion
    When the reposync logs should not report errors
```

ktsamis (changes requested):
The bigger problem for this I think is that it will always be red because the grep will catch always some packages that have "error" in the name. So we will never pass reposync. I don't think we should do this here, maybe in finishing

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
srbarrios: This is a post-condition so 'Then', in my opinion.

testsuite/features/reposync/srv_sync_products.feature
```diff
@@ -179,6 +179,9 @@ Feature: Synchronize products in the products page of the Setup Wizard
     Then I should not see a "Operation not successful" text
     And I should not see a warning nor an error sign
 
+  Scenario: Detect repository synchronization issues in reposync logs
```
srbarrios, line 182: Let's please don't have it on CI, we can keep checking this on finishing.
Sometimes, we might have some error at thatog, but we might prefer to keep running the rest of CI stages

maximenoel8:
> The bigger problem for this I think is that it will always be red because the grep will catch always some packages that have "error" in the name. So we will never pass reposync. I don't think we should do this here, maybe in finishing

I tried this change with sle maintenance update 4.3 that sync proxy and sles15sp4 and had no error. This check will only with for `ERROR` in uppercase so the packages should not be catch. Maybe let's have a try on CI and see the results. I will try that tomorrow.

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
srbarrios, line 826:
```suggestion
  @susemanager
  Scenario: Detect repository synchronization issues in reposync logs
    Then the reposync logs should not report errors
```
About having this on BV, I think we should distinguish Uyuni than Susemanager, the reason behind it is that for Uyuni BV that will run periodically and unattended I think is better if we don't stop the pipeline due to issues at that point, as sometimes it can recover itself or we can catch other important issues.

srbarrios (changes requested):
Let's please discuss the different questions on the PR.

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
Bischoff: Yes, everything with "should" should be "Then".

"Then it should be so..."

Bischoff: I disagree, I'd rather detect issues in Uyuni as well.

Especially that we tend to detect issues early when we release Uyuni.

testsuite/features/reposync/srv_sync_products.feature
Bischoff: Hmmm, we run CI more often than BV, it could be nice to catch things like  https://bugzilla.suse.com/show_bug.cgi?id=1228701

testsuite/features/reposync/srv_sync_products.feature
Bischoff: No, verifications should always be Then

Bischoff (approved):
LGTM

I'm sorry I seem to be disagreeing with everyone this time :trollface:

PR #9344, opened 2024-10-10
Explicitly handle timeouts when syncing channels

NamelessOne91:
## What does this PR change?

Related to https://github.com/SUSE/spacewalk/issues/25102

This PR forces failures in the case of an explicit Timeout Error when trying to sync existing repositories.
Cases in which a repository does not exist should still be handled by the StandardError branch

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Issue(s): # https://github.com/SUSE/spacewalk/issues/25102
Port(s): 

- 5.0
- 4.3 (?)

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -406,6 +406,9 @@ When(/^I wait until the channel "([^"]*)" has been synced$/) do |channel|
       log "#{time_spent / 60.to_i} minutes out of #{timeout / 60.to_i} waiting for '#{channel}' channel to be synchronized" if ((time_spent += checking_rate) % 60).zero?
       sleep checking_rate
     end
+  rescue Timeout::Error => e
+    log e.message
+    raise ScriptError, "Timeout occured while syncing channel: #{channel}"
   rescue StandardError => e
     log e.message
     unless $build_validation
@@ -445,6 +448,9 @@ When(/^I wait until all synchronized channels for "([^"]*)" have finished$/) do 
       log "#{time_spent / 60.to_i} minutes out of #{timeout / 60.to_i} waiting for '#{os_product_version}' channels to be synchronized" if ((time_spent += checking_rate) % 60).zero?
       sleep checking_rate
     end
+  rescue Timeout::Error => e
+    log e.message
+    raise ScriptError, "Timeout occured while syncing channels for: #{os_product_version}"
   rescue StandardError => e
     log e.message
     unless $build_validation


maximenoel8:
Be careful with rising an error. We don't want to rise error if no  repository is declared or no package is in the repository.
Example: Empty MU https://github.com/SUSE/spacewalk/blob/19866877e62433aadce9d077297c5718e794b5e5/testsuite/features/build_validation/add_MU_repositories/add_maintenance_update_repositories.template

maximenoel8:
Probably having no packages to sync is more an exception than a normal case. We could have an enable disable option to check fail on reposync. By digging into the CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION, I can see some channel are not correct.

NamelessOne91:
> Be careful with rising an error. We don't want to rise error if no repository is declared or no package is in the repository. Example: Empty MU https://github.com/SUSE/spacewalk/blob/19866877e62433aadce9d077297c5718e794b5e5/testsuite/features/build_validation/add_MU_repositories/add_maintenance_update_repositories.template

Will this case result in an error being raised ?
If so, I'd expect it to not be a TimeOutError but something else that enters the branch where we rescue for StandardError, where it seems we suppress errors if we're on a  build validation run

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -406,6 +406,9 @@
       log "#{time_spent / 60.to_i} minutes out of #{timeout / 60.to_i} waiting for '#{channel}' channel to be synchronized" if ((time_spent += checking_rate) % 60).zero?
       sleep checking_rate
     end
+  rescue Timeout::Error => e
+    log e.message
+    raise ScriptError, "Timeout occured while syncing channel: #{channel}"
```
srbarrios, line 411: Why we need this begin statement catching `StandardError` and `Timeout` while in fact that's already addressed as part of `repeat_until_timeout` function?


See this:

https://github.com/NamelessOne91/uyuni/blob/master/testsuite/features/support/commonlib.rb#L121-L126

NamelessOne91: Right, we can probably remove the error handling from the caller and just pass `$build_validation` as `dont_raise', if we intend to suppress errors

srbarrios: Exactly.

PR #9349, opened 2024-10-10
Prevent a crash on Liberty 6 and RHEL 6 clones (bsc#1231404)

meaksh:
## What does this PR change?

This PR just prevent a crash when scheduling "Package List Refresh" action, as `reboot_info.reboot_required` function was raising an Exception:

```
Error running 'reboot_info.reboot_required': Unable to run command '['dnf', '-q', 'needs-restarting', '-r']' with the context '......', reason: command not found
```

The reason is that `osmajorrelease` grain in old Salt 2016.11.10 (used in this OS) is rendered as string instead of integer, making the current logic not to work.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **no active testing on rhel6 and clones**

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25460

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/src/modules/reboot_info.py
+++ b/susemanager-utils/susemanager-sls/src/modules/reboot_info.py
@@ -51,7 +51,8 @@ def reboot_required():
     elif __grains__["os_family"] == "RedHat":
         cmd = (
             "dnf -q needs-restarting -r"
-            if __grains__["osmajorrelease"] >= 8
+            # In RHEL6 and clones 'osmajorrelease' is a string
+            if int(__grains__["osmajorrelease"]) >= 8
             else "needs-restarting -r"
         )
         result = _check_cmd_exit_code(cmd, 1)
--- a/susemanager-utils/susemanager-sls/susemanager-sls.changes.meaksh.master-bsc1231404
+++ b/susemanager-utils/susemanager-sls/susemanager-sls.changes.meaksh.master-bsc1231404
@@ -1,0 +1,1 @@
+- Prevent a crash on "reboot_info" module for Liberty 6, RHEL 6 & clones (bsc#1231404)


susemanager-utils/susemanager-sls/src/modules/reboot_info.py
```diff
@@ -51,7 +51,8 @@ def reboot_required():
     elif __grains__["os_family"] == "RedHat":
         cmd = (
             "dnf -q needs-restarting -r"
-            if __grains__["osmajorrelease"] >= 8
+            # In RHEL6 and clones 'osmajorrelease' is a string
+            if int(__grains__["osmajorrelease"]) >= 8
             else "needs-restarting -r"
```
agraul, line 56: Is `needs-restarting -r` always available on RHEL 6 or can this cause another error?

meaksh: Just for clarifying here, this changes are not meant to bring the "reboot info" feature to the unsupported RHEL6 clients, but at least to avoid making the function to crash, which leads to failed actions.

So, `needs-restarting` is provided by `yum-utils` package and available and installed in RHEL 6 registered clients.

OTOH the `-r` argument is not supported in RHEL 6, leading to this error:

```
[root@6dbee07d70e3 /]# rpm -qf /usr/bin/needs-restarting 
yum-utils-1.1.30-40.el6.noarch
[root@6dbee07d70e3 /]# needs-restarting -r
Usage: 
    needs-restarting: Report a list of process ids of programs that started 
                    running before they or some component they use were updated.
    

needs-restarting: error: no such option: -r
[root@6dbee07d70e3 /]# echo $?
2
```

The logic in `reboot_info` module just tries to check if retcode from `needs-restarting` is equal to `1`, which won't happen for RHEL 6 (always retcode 2 when passing `-r`).

I wanted to avoid adding a particular condition for the case of unsupported RHEL 6.

ycedres (approved):

PR #9357, opened 2024-10-14
Issv3 db schema

mcalmer:
## What does this PR change?

Initial ISSv3 Database Schema

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- https://github.com/SUSE/spacewalk/issues/25359

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25347

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/schema/spacewalk/common/tables/suseISSHub.sql
+++ b/schema/spacewalk/common/tables/suseISSHub.sql
@@ -1,0 +1,25 @@
+--
+-- Copyright (c) 2024 SUSE LLC
+--
+-- This software is licensed to you under the GNU General Public License,
+-- version 2 (GPLv2). There is NO WARRANTY for this software, express or
+-- implied, including the implied warranties of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+
+CREATE TABLE suseISSHub
+(
+    id 	                BIGINT CONSTRAINT suse_iss_hub_id_pk PRIMARY KEY
+                          GENERATED ALWAYS AS IDENTITY,
+    fqdn                VARCHAR(253) NOT NULL
+                          CONSTRAINT suse_iss_hub_fqdn_uq UNIQUE,
+    root_ca             TEXT,
+    mirror_creds_id     NUMERIC NULL
+                        CONSTRAINT suse_iss_hub_mirrcreds_fk
+                          REFERENCES suseCredentials (id) ON DELETE SET NULL,
+    created             TIMESTAMPTZ
+                          DEFAULT (current_timestamp) NOT NULL,
+    modified            TIMESTAMPTZ
+                          DEFAULT (current_timestamp) NOT NULL
+);
--- a/schema/spacewalk/common/tables/suseISSPeripheralChannels.sql
+++ b/schema/spacewalk/common/tables/suseISSPeripheralChannels.sql
@@ -1,0 +1,30 @@
+--
+-- Copyright (c) 2024 SUSE LLC
+--
+-- This software is licensed to you under the GNU General Public License,
+-- version 2 (GPLv2). There is NO WARRANTY for this software, express or
+-- implied, including the implied warranties of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+--
+
+CREATE TABLE suseISSPeripheralChannels
+(
+    id                          BIGINT CONSTRAINT suse_issperchan_id_pk PRIMARY KEY
+                                  GENERATED ALWAYS AS IDENTITY,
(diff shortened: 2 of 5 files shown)


schema/spacewalk/common/tables/suseISSHub.sql
```diff
@@ -0,0 +1,25 @@
...
+-- This software is licensed to you under the GNU General Public License,
+-- version 2 (GPLv2). There is NO WARRANTY for this software, express or
+-- implied, including the implied warranties of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+
+CREATE TABLE suseISSHub
+(
+    id 	                NUMERIC NOT NULL
```
aaannz, line 13:
```suggestion
    id 	                BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY
```

Please no NUMERIC.
Using this you can also drop custom sequence and also will not require insert to have `sequence_nextval(...)`, on contrary it will forbid custom values there so there is no risk of accidental sequence bug.

Hibernate will need to have `@GeneratedValue(strategy=GenerationType.IDENTITY)` for this to work AFAIK, I did not test it yet.

aaannz: And same for all other `id` columns here.

schema/spacewalk/common/tables/suseISSPeripheralChannels.sql
```diff
@@ -0,0 +1,24 @@
...
+--
+
+CREATE TABLE suseISSPeripheralChannels
+(
+    id                          NUMERIC NOT NULL
+                                  CONSTRAINT suse_issperchan_id_pk PRIMARY KEY,
+    peripheral_id               NUMERIC NOT NULL
+                                  CONSTRAINT suse_issperchan_pid_fk
+                                    REFERENCES suseISSPeripheral(id) ON DELETE CASCADE,
+    peripheral_org_id           NUMERIC,
```
aaannz, line 19: We can use `INTEGER` (4B) or even `SMALLINT` (2B) here.

schema/spacewalk/common/tables/suseISSHub.sql
rjmateus: I agree with Ondrej, we should not use Numeric. In this case, I think an Integer is enough, since it can save a number up to 2147483647

aaannz: There is indeed a storage saving (4B vs 8B per row), but performance wise these two are very close to each other.
Given how we regularly copy paste existing implementations when adding new, I argue we should default to `BIGINT` for all `id` columns, regardless expected table size.

rjmateus: we, I think it makes sense. Thank you for bring it Ondrej :)

mcalmer: What about mixing? The foreign key to mirror_creds_id is also NUMERIC. Do we have to keep it as the ID column is defined that way?

schema/spacewalk/common/tables/suseISSPeripheralChannels.sql
mcalmer: I used INTEGER

schema/spacewalk/common/tables/suseISSHub.sql
aaannz: We can mix types in FK. (SQL standard AFAIK forbid it completely, pgsql allows something like SMALLINT->INTEGER, etc.)
So with FK to existing tables we will have to use what is already there.

schema/spacewalk/common/tables/suseISSHub.sql
```diff
@@ -0,0 +1,23 @@
...
+(
+    id 	                BIGINT CONSTRAINT suse_iss_hub_id_pk PRIMARY KEY
+                          GENERATED ALWAYS AS IDENTITY,
+    fqdn                VARCHAR(253) NOT NULL
+                          CONSTRAINT suse_iss_hub_fqdn_uq UNIQUE,
+    root_ca             TEXT,
+    mirror_creds_id     NUMERIC NULL
+                        CONSTRAINT suse_iss_hub_mirrcreds_fk
+                          REFERENCES suseCredentials (id) ON DELETE SET NULL,
+    created             TIMESTAMPTZ
```
rjmateus, line 21: Should we also have the modify field as some of the older tables?

mcalmer: I think I will add it. It makes the handling in Java easier

schema/spacewalk/common/tables/suseISSHub.sql
mcalmer: I think this is fixed now.

lucidd (approved):

mackdk (approved):

mcalmer:
I think the discussions are done and we can still adapt the schema later if needed. Merging now in our feature branch

PR #9368, opened 2024-10-15
Ux buttons - Deprecating the classes btn-success, btn-info, and btn-warning

bisht-richa:
## What does this PR change?
This PR partially fixes [#24324](https://github.com/SUSE/spacewalk/issues/24324) - Deprecating the classes btn-success, btn-info, btn-link, and btn-warning.
- Removed the `btn-success` class from CSS and pages
- Removed the `btn-info` class from CSS and pages
- Removed the `btn-warning` class from CSS and page
- Removed the `btn-link` class from some pages.  (Some instances still remain and will be addressed in future updates.)
- Modify `<rhn:toolbar> `component -  `ToolbarTag.java` file
    - Added the `btn-primary` CSS class to **createType** button, 
    - Applied the `btn-default` CSS classe to the **updateType** and **cloneType** buttons.
    -  Applied the `btn-danger` CSS class to the **deleteType** button.
    - Modify the code to ensure that the `createType` button is always positioned at the far right, 
- Modify `<FilterEdit>` component
  - Replaced `ModalLink>` with `<ModalButton>` component to launch a modal dialog
  - Added `className` props 

## GUI diff

No difference.

**Before:**
Screenshot 2024-10-16 at 11 42 18
Screenshot 2024-10-16 at 13 12 45

Screenshot 2024-10-16 at 11 18 02

**After:**

Screenshot 2024-10-16 at 11 17 03
Screenshot 2024-10-16 at 11 17 30

Screenshot 2024-10-16 at 11 18 13

Screenshot 2024-10-16 at 11 19 59

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links
Fixes: [#25737](https://github.com/SUSE/spacewalk/issues/25737)
Issue(s): [#24324](https://github.com/SUSE/spacewalk/issues/24324)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


web/html/src/components/buttons.stories.tsx
```diff
@@ -1,10 +1,10 @@
-import { StoryRow, StripedStorySection } from "manager/storybook/layout";
+import { StoryRow, StorySection } from "manager/storybook/layout";
 
 import { Button } from "components/buttons";
 
 export default () => {
   return (
-    <StripedStorySection>
```
Etheryte, line 7: Is there a reason to remove this here? The stripes are useful so you can see which elements have transparency and which don't.

java/code/webapp/WEB-INF/pages/admin/multiorg/configuration.jsp
```diff
@@ -72,11 +72,18 @@
...
+                  <div class="col-lg-offset-3 offset-lg-3 col-lg-6">
+                    <html:submit styleClass="btn btn-primary">
+                      <bean:message key="orgdetails.jsp.submit"/>
+                    </html:submit>
+                  </div>
+                </div>
             </div>
         </div>
-        <html:submit styleClass="btn btn-success pull-right">
+        <!-- <html:submit styleClass="btn btn-primary">
```
Etheryte, line 84: This should probably not be commented out?

java/spacewalk-web.changes.bisht-richa.ux-ui-buttons-component
```diff
@@ -0,0 +1,2 @@
+- Deprecating the classes btn-success, btn-info, and btn-warning.
```
Etheryte, line 1: Here and below, I think we need different descriptions for the changelog. While this describes what's changed from the technical side, the changelog is for customers, not engineers. For a customer, things like `btn-success` etc don't say anything meaningful. I think something along the lines of "Updated user interfaces for pages X, Y, Z" or something similar would work.  

Maybe @deneb-alpha has a good suggestion, you're usually pretty good with changelog entries?

web/html/src/manager/audit/subscription-matching/subscription-matching-matcher-run-panel.tsx
```diff
@@ -144,8 +144,7 @@ class MatcherScheduleButton extends React.Component<MatcherScheduleButtonProps>
   };
 
   render() {
-    const buttonClass =
-      "btn spacewalk-btn-margin-vertical " + (!this.props.matcherRunning ? "btn-success" : "btn-default");
```
Etheryte, line 148: If I'm reading this correctly, the button class used to essentially show loading state here. Is this something we want to display to the user in some other way here perhaps? I'm not familiar with this specific page.

web/html/src/manager/content-management/list-filters/list-filters.tsx
```diff
@@ -134,11 +135,11 @@ const ListFilters = (props: Props) => {
     </div>
   );
 
-  const unusedFilter = <Button className="btn-link" handler={onSelectUnused} text={t("Select unused")}></Button>;
+  const unusedFilter = <Button className="btn-default" handler={onSelectUnused} text={t("Select unused")}></Button>;
 
   const deleteSelected = (
     <Button
-      className={`${selectedIdentifiers.length ? "btn-danger" : "btn-disabled"}`}
+      className={`${selectedIdentifiers.length ? "btn-danger" : "btn-danger disabled"}`}
```
Etheryte, line 142: I think the standalone class `disabled` should be unnecessary here, since we already set and unset the `disabled` attribute?

web/html/src/manager/content-management/list-filters/list-filters.tsx
```diff
@@ -152,6 +153,7 @@ const ListFilters = (props: Props) => {
       button={panelButtons}
       helpUrl="reference/clm/clm-filters.html"
     >
+      test
```
Etheryte, line 156: Guessing this is obsolete.

Etheryte:
I think this is a very welcome change, we have too many different buttons and they're used in a variety of ways, so simplifying all of that is very nice. One thing to consider once we're done with the cleanup is that we might want to revisit some of the names afterwards. Right now we have for example `class="btn"` and `class="btn btn-default"`, which can be confusing as a name, as you would think that default is what you get out of the box. There's historical context here of course, but maybe once we're done with this, we can think of a better name?

In general though, this looks good, there were a few nitpicks here and there, but nothing major, this looks good. 👍

web/html/src/manager/audit/subscription-matching/subscription-matching-matcher-run-panel.tsx
Etheryte: Okay, I see what you mean now. As a small aside, since the button class is now static, we could move it onto the element itself:

```tsx
<button type="button" className="btn space..." ... />
```

Etheryte (dismissed):
This looks good! I like the new change log entry as well. 👍

deneb-alpha (changes requested):
.changes on spacewalk-web LGTM but the changes for spacewalk-java are  wrong.

the file name in java  subfolder is `spacewalk-web.changes.bisht-richa.ux-ui-buttons-component` but  it should  be  called `spacewalk-java.changes.bisht-richa.ux-ui-buttons-component`.

if  in spacewalk-java we have the .changes called spacewalk-web the mechanism tagging implemented by @cbbayburt  that unifies the intermediate .changes into the  main one is not  going to work. 

And I think this is also why we have this leftover changelog
https://github.com/uyuni-project/uyuni/blob/master/java/spacewalk-web.changes.bisht-richa.table-filters-description-fix

Etheryte:
Fyi you can avoid the above issue if you use `mkchlog` from the [uyuni-releng-tools](https://github.com/uyuni-project/uyuni-releng-tools) repo instead of manually creating the change files. The process is straightforward:

```sh
cd web
mkchlog "My changelog message"
```

bisht-richa:
> .changes on spacewalk-web LGTM but the changes for spacewalk-java are wrong.
> 
> the file name in java subfolder is `spacewalk-web.changes.bisht-richa.ux-ui-buttons-component` but it should be called `spacewalk-java.changes.bisht-richa.ux-ui-buttons-component`.
> 
> if in spacewalk-java we have the .changes called spacewalk-web the mechanism tagging implemented by @cbbayburt that unifies the intermediate .changes into the main one is not going to work.
> 
> And I think this is also why we have this leftover changelog https://github.com/uyuni-project/uyuni/blob/master/java/spacewalk-web.changes.bisht-richa.table-filters-description-fix

Ohh my mistake. I will correct that. Thanks :)

java/code/webapp/WEB-INF/pages/admin/multiorg/configuration.jsp
```diff
@@ -1,82 +1,78 @@
 <%@ taglib uri="http://rhn.redhat.com/rhn" prefix="rhn" %>
-<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
-<%@ taglib uri="http://struts.apache.org/tags-html" prefix="html" %>
-<%@ taglib uri="http://struts.apache.org/tags-bean" prefix="bean" %>
+  <%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+    <%@ taglib uri="http://struts.apache.org/tags-html" prefix="html" %>
+      <%@ taglib uri="http://struts.apache.org/tags-bean" prefix="bean" %>
```
Etheryte, line 4: Here and in the rest of this file, the formatting is wrong. It seems whatever autoformatter you're using doesn't recognize that these are references, not actual opening tags.

@cbbayburt Do you happen to know if we can autofix formatting in jsp files as well somehow conveniently? Would be nice to just have a linter do all of this automatically.

web/html/src/branding/css/base/responsive-rules.less
```diff
@@ -0,0 +1,31 @@
+@media (max-width: @screen-tablet) {
```
Etheryte, line 1: This file is safe to drop, we don't include or build `.less` files anymore since we dropped the old theme. The `.scss` files are now the only source of styles.

web/html/src/branding/css/base/theme.less
Etheryte, line 1: This file can be deleted, I'm guessing this got mixed up when you rebased the branch.

Etheryte:
Since we now only allow a narrow set of class names etc, I would recommend updating the prop type definitions in the button component. E.g. instead of `foo?: string`, I think we should now have `foo?: "btn-primary" | "..."` and list all allowed values. That way no one can go and use the wrong values again.

Etheryte:
This looks good, thanks a lot for putting what must've been a massive effort into this.  

Two things we need to do before we merge this: one, ensure we remove the `.less` files that I mentioned above, none should now exist, and two, fix the jsp formatting.

Outside of that this looks very nice. 👍

When merging this, I would recommend squashing, since that way it's easier to reason about this commit later.

bisht-richa:
> Since we now only allow a narrow set of class names etc, I would recommend updating the prop type definitions in the button component. E.g. instead of `foo?: string`, I think we should now have `foo?: "btn-primary" | "..."` and list all allowed values. That way no one can go and use the wrong values again.

I suggest fixing this use case in a separate PR which we can use to resolve other issues/modifications needed in the Button component.

deneb-alpha (dismissed):
From releng side, LGTM! thanks

PR #9372, opened 2024-10-16
Improve appstreams context selection

parlt91:
## What does this PR change?

Fix an issue where parsing appstream modules would from clients would parse all modules not just active ones.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **Bugfix**

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25484

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/webapp/WEB-INF/nav/system_detail.xml
+++ b/java/code/webapp/WEB-INF/nav/system_detail.xml
@@ -69,7 +69,7 @@
       </rhn-tab>
     </rhn-tab>
 
-    <rhn-tab name="AppStreams" acl="user_role(satellite_admin); system_has_modular_channels()" url="/rhn/manager/systems/details/appstreams" />
+    <rhn-tab name="AppStreams" acl="system_has_modular_channels()" url="/rhn/manager/systems/details/appstreams" />
 
     <rhn-tab name="PTF" acl="has_ptf_repositories(); system_supports_ptf_removal() or system_feature(ftr_package_updates)">
       <rhn-tab-url>/rhn/manager/systems/details/ptf/overview</rhn-tab-url>


cbbayburt:
Thanks, @parlt91 looks good.

I am going to test it with some corner cases before approving though.

cbbayburt (dismissed):
There's still no Rocky 8 update available since the last time, so I cannot see if context values change, but I tested everything else and it works as expected.

In the rare case that a context value changes with a module update, (RH-like 8 only) we might still have problems suggesting the correct package for install. There's no easy fix for that, but at least the workaround would be to install such packages from the CLI directly. Hopefully, we won't encounter this in real life.

Thanks for the fix :+1:

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -92,6 +92,26 @@ def _get_module_info(module_names):
         return []
 
     module_info_output = result.stdout.splitlines()
+    active_modules = []
```
meaksh, line 95: We should also change this: https://github.com/uyuni-project/uyuni/pull/9372/files#diff-0afe89c35f776dab892a3a79e2f20c5db733ec5c7c2002251a6b205145e7bf43L86 as now the function get the info for active modules only.

Also, before starting parsing the output, I would add an example output to understand better what the loop is trying parse/achieve.

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -92,6 +92,26 @@ def _get_module_info(module_names):
...
+            )
+            if end != -1:
+                active_modules+=module_info_output[start-1:end+start]
+                module_info_output = module_info_output[end+start:-1]
+            else:
+                active_modules+=module_info_output[start-1:end]
+                module_info_output=active_modules
+                break
+        else:
+            break
```
meaksh, line 114: I would probably rather prefer to iterate `module_info_output` outside of a `while True` loop, for example, iterating `module_info_output` line-by-line (if that makes sense), and then filtering and collecting only the content of active modules we want.

susemanager-utils/susemanager-sls/src/modules/appstreams.py
```diff
@@ -92,6 +92,26 @@ def _get_module_info(module_names):
...
+            end = next(
+                (i for i, e in enumerate(module_info_output[start:-1])
+                if e.startswith("Name")), -1
+            )
+            if end != -1:
+                active_modules+=module_info_output[start-1:end+start]
+                module_info_output = module_info_output[end+start:-1]
+            else:
+                active_modules+=module_info_output[start-1:end]
+                module_info_output=active_modules
```
meaksh, line 111: Styling nitpick:

```suggestion
                (i for i, e in enumerate(module_info_output[start : -1])
                if e.startswith("Name")), -1
            )
            if end != -1:
                active_modules += module_info_output[start - 1 : end + start]
                module_info_output = module_info_output[end + start : -1]
            else:
                active_modules += module_info_output[start - 1 : end]
                module_info_output = active_modules
```

susemanager-utils/susemanager-sls/src/modules/appstreams.py
meaksh: Sorry for the bad explanation! I was thinking on something a bit more easy to follow, maybe using "regex", like:

```python
    module_info_output = []
    for module in re.findall(r"(Name\s+:.*?)(?=\n\s*\n|$)", result.stdout, re.DOTALL):
        if re.search(r"Stream\s+:.*\[a\]", module):
            module_info_output += module.splitlines()
```

meaksh (approved):
LGTM, thanks!

PR #9384, opened 2024-10-17
Refactor salt cleanup step

ktsamis:
## What does this PR change?

port of https://github.com/SUSE/spacewalk/pull/25556/ . See there for explanation.

Additional changes from the 5.0 port:

Remove only some files, not whole directories.
Use existing step for removing packages since we have it.
Change the order of file removal and package removal. That is the important part because it saves us from a reboot inbetween these actions. Now with this order the transaction should not be in an inconsistent state after only one reboot. Testing currently to make sure we don't break anything

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24035
Port(s): 
5.0: https://github.com/SUSE/spacewalk/pull/25556/ together with https://github.com/SUSE/spacewalk/pull/25754
4.3: https://github.com/SUSE/spacewalk/pull/25565

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/build_validation/init_clients/proxy.feature
+++ b/testsuite/features/build_validation/init_clients/proxy.feature
@@ -17,7 +17,6 @@ Feature: Setup containerized proxy
 
   Scenario: Clean up sumaform leftovers on the containerized proxy
     When I perform a full salt minion cleanup on "proxy"
-    And I reboot the "proxy" host through SSH, waiting until it comes back
 
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
--- a/testsuite/features/build_validation/init_clients/proxy.feature
+++ b/testsuite/features/build_validation/init_clients/proxy.feature
@@ -17,6 +17,7 @@ Feature: Setup containerized proxy
 
   Scenario: Clean up sumaform leftovers on the containerized proxy
     When I perform a full salt minion cleanup on "proxy"
+    And I reboot the "proxy" host through SSH, waiting until it comes back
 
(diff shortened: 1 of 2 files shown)


vzhestkov (approved):
LGTM. I hope we can improve the bootstrap state to avoid such corner cases with file cleanups, but I haven't yet tried to fix it.

vzhestkov (changes requested):
I think we can improve the way of cleanup instead of rebooting in each step. I just checked the bootstrap state and it's checking for the case of presence the minion config directory, but as you remember while testing after reboot we saw empty config directory even if it was deleted explicitly before reboot.

So suggestion is the following:
- Do not delete `/etc/venv-salt-minion` and `/etc/salt` completely
- Delete just the following from the config dir:
    ```
    grains
    minion_id
    minion.d/* (except 00-venv.conf)
    pki/minion/*
    ```
    `minion.d/* (except 00-venv.conf)` can be done with `find /etc/venv-salt-minion/minion.d/ -type f ! -name '00-venv.conf' -delete`

`00-venv.conf` is only relevant for the salt bundle, for classic salt package it could be `transactional_update.conf`, but not sure if it's used in the testsuite, but it shouldn't be a problem to not delete it.

NamelessOne91 (approved):

PR #9401, opened 2024-10-23
QE: refresh packages list before CVE audit

NamelessOne91:
## What does this PR change?

This PR adds a packages list refresh after installing the old package otherwise the system is not aware of its presence on the minion and does not suggest the patch for it after a CVE audit.

It also move the errata cache refresh to its own scenario, as I think that should not be considered strictly part of the pre-requisite one.

## GUI diff

- None

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- [ ] **DONE**

## Links

Port(s): 

- 4.3:
- 5.0:

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/secondary/min_cve_audit.feature
+++ b/testsuite/features/secondary/min_cve_audit.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2015-2022 SUSE LLC
+# Copyright (c) 2015-2024 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 @sle_minion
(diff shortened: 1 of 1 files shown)


srbarrios (approved):

nodeg (approved):
Thank you, Paolo.

Bischoff (changes requested):
For me, developers should fix it. We should not hide the dust under the carpet.

NamelessOne91:
Closing.
The issue was caused by a product bug which should have been fixed, the package list was not correctly kept updated when packages where installed/removed by not using SUMA (e.g using zypper in the VM).

PR #9416, opened 2024-10-27
QE: correctly upgrade Salt in product migrations

NamelessOne91:
## What does this PR change?

It takes care of installing the latest OS Salt on the SLE 15 SP3 SSH minion we use to test product migration.
This is needed because we use a pristine image with an old version of it and the Product Migration feature requires the latest to be installed.

It also removes scenarios relying on a Salt file that is not present on the SLES 15 SP3 minions.
`/root/salt` and the `repos` Salt states are only found on the server 

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Port(s): 

- 4.3: https://github.com/SUSE/spacewalk/pull/25655
- 5.0: https://github.com/SUSE/spacewalk/pull/25656

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/build_validation/migration/sle15sp3_ssh_minion_migration.feature
+++ b/testsuite/features/build_validation/migration/sle15sp3_ssh_minion_migration.feature
@@ -46,12 +46,6 @@ Feature: Migrate a SLES 15 SP3 Salt SSH minion to 15 SP4
     Then I should see a "SUSE Linux Enterprise Server 15 SP4" text
     And vendor change should be enabled for product migration on "sle15sp3_ssh_minion"
 
-  Scenario: Install the latest Salt on this SSH minion
-    When I migrate the non-SUMA repositories on "sle15sp3_ssh_minion"
-    And I enable repositories before installing Salt on this "sle15sp3_ssh_minion"
-    And I install Salt packages from "sle15sp3_ssh_minion"
-    And I disable repositories after installing Salt on this "sle15sp3_ssh_minion"
-
   Scenario: Subscribe the SSH-managed SLES minion to a SLES 15 SP4 child channel
     Given I am on the Systems overview page of this "sle15sp3_ssh_minion"
     When I follow "Software" in the content area
--- a/testsuite/features/build_validation/migration/sle15sp3_ssh_minion_migration.feature
+++ b/testsuite/features/build_validation/migration/sle15sp3_ssh_minion_migration.feature
@@ -7,6 +7,21 @@ Feature: Migrate a SLES 15 SP3 Salt SSH minion to 15 SP4
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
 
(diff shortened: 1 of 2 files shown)


testsuite/features/build_validation/migration/sle15sp3_ssh_minion_migration.feature
```diff
@@ -7,6 +7,21 @@ Feature: Migrate a SLES 15 SP3 Salt SSH minion to 15 SP4
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
 
+  # pristine images for SSH minions use an old version of OS Salt
+  # Product Migration will be possible only if we update to latest
```
srbarrios, line 11: Why we don't need the same scenario in a regular Salt Minion?

testsuite/features/build_validation/migration/sle15sp3_ssh_minion_migration.feature
```diff
@@ -7,6 +7,21 @@ Feature: Migrate a SLES 15 SP3 Salt SSH minion to 15 SP4
...
+  Scenario: Prerequisite: update OS Salt to the latest version
+    Given I am on the Systems overview page of this "sle15sp3_ssh_minion"
+    When I follow "Software" in the content area
+    And I follow "Packages"
+    And I follow "Upgrade"
+    And I enter "salt" as the filtered package name
```
srbarrios, line 17: I wonder if this filter could be selecting not only the classic salt, but also the salt bundle, as both packages contains the word `salt`.

testsuite/features/build_validation/migration/sle15sp3_ssh_minion_migration.feature
NamelessOne91: TBH, I am puzzled by that as well, but it seems regular Salt minions are all using the latest venv-salt.
I don't know if that is a just a coincidence or we are actively doing something that keeps it updated in Sumaform (or the testsuite).

srbarrios: Latest venv-salt-minion it's expected, we are installing it through cloud-init phase.
But I'm talking about classic salt (aka OS salt)

NamelessOne91: Having OS Salt up to date is not needed on the regular minions then, as the Product Migration correctly detects that they are using venv-salt and it is the latest version available.

SSH Salt minions have only an old version of OS Salt installed, that's the issue with them.

srbarrios: Oh right, sorry, I was thinking on Salt migration test case, not product migration.
All good, approving :)

srbarrios (approved):

Issue #9403, opened 2024-10-23
/usr/bin/spacewalk-watch-channel-sync.sh is missing on 2024.10


rkoktan:
### Problem description

I just patched to 2024.10 and discovered /usr/bin/spacewalk-watch-channel-sync.sh is missing.

### Steps to reproduce

1. Attempted to launch /usr/bin/spacewalk-watch-channel-sync.sh from inside the Uyuni container (mgrctl term)
2. /usr/bin/spacewalk-watch-channel-sync.sh
bash: /usr/bin/spacewalk-watch-channel-sync.sh: No such file or directory
3. Looked in /usr/bin/ and spacewalk-watch-channel-sync.sh is not present.
...

### Uyuni version

```bash
Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : @System
Name           : Uyuni-Server-release
Version        : 2024.10-241000.224.1.uyuni5
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : unknown
Installed Size : 1.4 KiB
Installed      : Yes (automatically)
Status         : up-to-date
Source package : Uyuni-Server-release-2024.10-241000.224.1.uyuni5.src
Upstream URL   : https://www.uyuni-project.org/
Summary        : Uyuni Server
Description    : 
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.
```

rjmateus:
We have removed it from the container by mistake with this PR: https://github.com/uyuni-project/uyuni/pull/9251/commits
I will revert this part in a new PR next week. Thank you for noticing it.

mcalmer:
I think a simple `tail -f /...../*.log` is doing the same.
It was moved to a contrib repository

PR #9424, opened 2024-10-31
Remove snaposhot and export tools

rjmateus:
## What does this PR change?

Install the spacewalk-extra-utils since it still contains some utility tools. Several of those tools needs to be migrated, but until that, we must install the package.

Partially reverts: https://github.com/uyuni-project/uyuni/pull/9251

## GUI diff

No difference.

## Documentation
- No documentation needed. I didn't removed it in the old documentation. Also, those tools where not documented.

- [ ] **DONE**

## Test coverage

- No tests: old toos, so no tests.

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni/issues/9403
Port(s): Needs to be ported to 5.0

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/utils/spacewalk-utils.changes.rmateus.install_utils_extra_clean
+++ b/utils/spacewalk-utils.changes.rmateus.install_utils_extra_clean
@@ -1,0 +1,1 @@
+- Remove unmaintained snapshot and export tools from package `spacewalk-utils-extras`


mcalmer (changes requested):
No, we do not want these tools. They are not supported.
I need to check all of them, but from my point of view there is a reason why they are in an unsupported extra package

rjmateus:
@mcalmer I was following what was discussed with  @admd in this confluence page: https://confluence.suse.com/display/~RDiasMateus/SUMA+cmd+tools

I would love to completely remove the remaining tools from this package (tasktop is already moved to the regular package, so that can be kept). But to remove all I would need @admd agreement, and I'm not sure if we must first document that those are deprecated. 
Also, what should we do to these tools on 5.0 and 5.1?

rjmateus:
Maybe I can split the PR in 2, one to remove the tools we know we can remove, and another one to see if we want to install the extras package in the container

rjmateus:
@juliogonzalez had the idea of moving it to the contrib repository. And looking with more attention to the confluence page, I think that was what @admd and I had discussed for the majority of the tool.
Let's talk next week, but maybe I can move must of the scripts to the contrib repo, and the one or two we want to support, move to the spacewalk-utils package instead the extra, like we did with tasktop.

admd:
@rjmateus I just saw the description of spacewalk-utils-extra package that states

`Extra utilities that may be run against a SUSE Manager server (unsupported) or an Uyuni server`

I didn’t know about this before. So, @mcalmer is correct—we can safely remove the entire package. The only regularly used component is taskotop, and if I understand correctly, it’s already covered.

rjmateus:
Ok, that makes sense. 
Taskotop was moved to the package spacewalk-utils, so it means it is supported.
I will then change the PR to not install the extras on the image, and remove the other tools and extras package. At same time I will move some of the tools to the contrib repo (according to the document in confluence).
@mcalmer @admd makes sense for you?

juliogonzalez:
>  I will then change the PR to not install the extras on the image, and remove the other tools and extras package. At same time I will move some of the tools to the contrib repo (according to the document in confluence).

Should we comment it at the release notes, at least for Uyuni?

rjmateus:
@mcalmer I will keep this PR like it's now, just removing the dead tools. I will open a new one with the other tools, and then it can be announced to uyuni release notes.
Could you review it? Thank you

mcalmer (approved):

PR #9431, opened 2024-11-04
Quality Intelligence

srbarrios:
## What does this PR change?

This PR implement some handlers to report system metrics with the aim to achieve this card: https://github.com/SUSE/spacewalk/issues/24507

Additionally, it implements the necessary steps to extract the metrics from our Continuous Integration test suite.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Documentation needed: I will include it as part of this PR

- [x] **DONE**

## Test coverage

- Cucumber tests were added

- [x] **DONE**

## Links

Ports:
- Manager-5.0

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


testsuite/features/support/system_monitoring.rb
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) 2024 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
```
szachovy, line 3: This is not `system_monitoring`. These are supplementary functions to be used by Quality Intelligence exclusively. I would integrate them to `QualityIntelligence` class and use them directly there.

testsuite/features/step_definitions/system_monitoring_steps.rb
```diff
@@ -0,0 +1,25 @@
...
+# Licensed under the terms of the MIT license.
+
+### This file contains the definitions for the steps extracting and reporting information from the system
+
+When(/^I report the bootstrap duration for "([^"]*)"$/) do |host|
+  next unless $quality_intelligence_mode
+
+  duration = last_bootstrap_duration(host)
+  $quality_intelligence.push_bootstrap_duration(host, duration)
+end
```
szachovy, line 11: If all the duration functions were integrated to `QualityIntelligence` (see ...) class you wouldn't need to invoke that each time and do one liner instead.
```suggestion
When(/^I report the bootstrap duration for "([^"]*)"$/) do |host|
  $quality_intelligence.push_bootstrap_duration(host) if $quality_intelligence_mode
end
```

testsuite/features/support/database_handler.rb
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2024 SUSE LLC.
+# Licensed under the terms of the MIT license.
+require 'redis'
+
```
szachovy, line 4: If we are not sure if `DatabaseHandler` and `PrometheusHandler` will be used anywhere except `QualityIntelligence` at the moment. I would put these classes to `quality_intelligence.rb`. That's optional, I don't think we have a standards one class per file, I am just not a fan having this many files for one feature.

testsuite/features/support/system_monitoring.rb
```diff
@@ -0,0 +1,64 @@
...
+    if line.include?('Total time: ') && channel_to_evaluate
+      match = line.match(/Total time: (\d+):(\d+):(\d+)/)
+      hours, minutes, seconds = match.captures.map(&:to_i)
+      total_seconds = (hours * 3600) + (minutes * 60) + seconds
+      duration += total_seconds
+      channel_to_evaluate = false
+    end
+  end
+  duration
+end
```
szachovy, line 64: That's risky if the file structure changes or something goes wrong with the file content. I don’t recommend adding file checks, as that could quickly require dozens lines of code. Instead, have you considered alternative options, like tracking the total process duration? That might be more reliable.

szachovy (approved):
Comments are optional, but please respond to them.

testsuite/Gemfile
```diff
@@ -29,7 +30,7 @@ gem 'redis', '~> 5.2'
 gem 'redis-client', '~> 0.22'
 gem 'require_all', '~> 3.0'
 gem 'rubocop', '~>1.66'
-gem 'selenium-webdriver', '4.24'
+gem 'selenium-webdriver', '~> 4'
```
nodeg, line 33: Good idea. This way, we are up to date with the chromedriver version we are using that automatically will use the latest version.

szachovy: Haven't we some issues with the latest chromerdiver btw?

testsuite/features/support/code_coverage.rb
```diff
@@ -1,20 +1,16 @@
-# Copyright (c) 2016-2023 SUSE LLC.
+# Copyright (c) 2024 SUSE LLC.
```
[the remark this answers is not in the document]
srbarrios: No, not really, it was a mistake when I implemented code_coverage.rb this year.

testsuite/Gemfile
srbarrios: Not anymore, we were waiting for the latest that included the fix for our issue, and few weeks ago Maxime upgraded it in sumaform, and worked fine.

testsuite/features/step_definitions/system_monitoring_steps.rb
srbarrios: I prefer to follow the Single Responsibility Principle and separate the fact of extracting the duration and the action of pushing the metric into different functions.

As example, just imagine that some day we decide to re-use the QualityIntelligence handler in another project, we would not like too much to have specific code related to SUMA as part of that class.

Also, that might be a matter of taste, but I prefer to don't encapsulate more that necessary the actions we do inside a Cucumber step definition, otherwise is harder to understand all the involved components.

testsuite/features/support/code_coverage.rb
```diff
@@ -1,20 +1,16 @@
...
-    @database = Redis.new(host: redis_host, port: redis_port, username: redis_username, password: redis_password)
-  end
+  include Nokogiri::XML
 
-  # Close the connection with the Redis database
-  def close
-    @database.close
+  # Initialize the CodeCoverage handler
+  def initialize
+    @db_handler = DatabaseHandler.new(ENV.fetch('REDIS_HOST', nil), ENV.fetch('REDIS_PORT', nil), ENV.fetch('REDIS_USERNAME', nil), ENV.fetch('REDIS_PASSWORD', nil))
```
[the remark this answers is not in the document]
srbarrios: The ruby style that we follow recommends the usage of `ENV.fetch`, so I'm gonna keep it. And I prefer to don't provide any other default.

But you are right that handling the raised error when we initialize the database is something missing, I will add it.

testsuite/features/support/database_handler.rb
srbarrios: DatabaseHandler it was used previously for code coverage, and I prefer to keep those handlers independent from QualityIntelligence, in case we want to re-use them for something else.

testsuite/features/support/system_monitoring.rb
srbarrios: I really prefer to keep QualityIntelligence simpler and less coupled with the rest of our test framework code.

testsuite/features/support/system_monitoring.rb
srbarrios: There is not a unique process involved, as there are many channels synchronized.
The order how those channels are processes is not deterministic.
Tracking this it would require a different thread in our test framework or block the current thread processing it.
Additionally, that would require embedding that tracking in the middle of our tests from the moment we trigger the synchronization until it ends, while that way, it's much more independent, and it only need to process a file once we know all is synchronized.

testsuite/features/support/system_monitoring.rb
szachovy: ok

testsuite/features/support/database_handler.rb
szachovy: ok

testsuite/features/support/system_monitoring.rb
szachovy: I am a bit worried what in case of timeout interruptions where file got cut, or packages woudn't be found and the time would be 0, and other corner cases like that. Can we at least use logging if there won't be regex match or it would be suspiciously different?

testsuite/features/step_definitions/system_monitoring_steps.rb
szachovy: My initial assumption was that we rather won't use it in other projects because tech stack is too specific for this particular use case. But I am ok with it how it is now.
Closing

testsuite/features/support/system_monitoring.rb
srbarrios: I'm gonna raise an error if the duration is equal to zero, as that should never be the case.

testsuite/features/support/quality_intelligence.rb
```diff
@@ -0,0 +1,47 @@
...
+
+# Quality Intelligence handler to produce, parse and report Quality Intelligence from the test suite
+class QualityIntelligence
+  QI = 'quality_intelligence'.freeze
+  private_constant :QI
+
+  # Initialize the QualityIntelligence handler
+  def initialize
+    @db_handler = DatabaseHandler.new(ENV.fetch('REDIS_HOST', nil), ENV.fetch('REDIS_PORT', nil), ENV.fetch('REDIS_USERNAME', nil), ENV.fetch('REDIS_PASSWORD', nil))
+    @metrics_collector_handler = MetricsCollectorHandler.new(ENV.fetch('PROMETHEUS_PUSH_GATEWAY_URL', 'http://nsa.mgr.suse.de:9091'))
```
szachovy, line 14: What happens if Prometheus/Database cannot be contacted, not even if the URL is bad, but when it's after firewall/cert validation error etc?

srbarrios: It gonna raise an error and the cucumber process will fail, as expected.
I mean, we don't want to keep running under such conditions.

szachovy: My concert is that errors that will raise may be unclear, like connection timeout error, but it's likely to be managable.

Closing

szachovy (approved):

testsuite/documentation/software-components.md
```diff
@@ -34,6 +34,25 @@ to run lists of Cucumber features stored in YAML format
...
 * [`net-scp`] is a Ruby library that allows to copy files to and from remote machines via SCP
 
+### Database interaction
+
+* [`database_handler`] is a custom Ruby class that allows to interact with a No-SQL database. So you can store a Map of key-value pairs in the database.
+
+### Metrics Collection
+
+* [`metrics_collector_handler`] is a custom Ruby class that allows to push metrics from the system tested to a Metrics Collector.
+  As default this handler is configured to connect to a Prometheus Push Gateway instance located in `nsa.mgr.suse.de:9091`, if you want to change it you need to set the `PROMETHEUS_PUSH_GATEWAY_URL` environment variable.
```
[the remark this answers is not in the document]
srbarrios: Test VMs are the VMs we have as part of the system we test.
But there, this is communicating with a component in production in our internal infrastructure.

szachovy (approved):

testsuite/features/support/database_handler.rb
```diff
@@ -0,0 +1,47 @@
...
+# Database handler to interact with a database
+class DatabaseHandler
```
NamelessOne91, line 6: Nitpicking a bit.

This class seems to be really specific for a key-value store, Redis in particular.
DatabaseHandler makes me think at something that's supposed to interact with e.g PostgreSQL, or act as a common "interface" for specific implementations.
I'd probably just call it RedisHandler.

srbarrios: I renamed DatabaseHandler to be more generic, as for MetricsCollectorHandler.
So, in case we change the implementation in the future. 😅

PR #9475, opened 2024-11-14
Increase complexity for generated passwords (bsc#1231983)

bisht-richa:
## What does this PR change?

Update password generation, see https://bugzilla.suse.com/show_bug.cgi?id=1231983 for discussion.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Fixes: https://github.com/SUSE/spacewalk/issues/25664

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1231983

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


web/html/src/utils/functions.ts
```diff
@@ -116,16 +116,34 @@ function capitalize(str: string): string {
...
+  const lowercase = "abcdefghijklmnopqrstuvwxyz";
+  const uppercase = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+  const numbers = "0123456789";
+  const specials = ":-_!@#$%^&*()"; // extended set of special characters
+
+  // Helper function to get a random character from a given set
+  const allChars = lowercase + uppercase + numbers + specials;
+
+  // Ensure at least one character from each category
+  const getRandomChar = (charset) => charset[Math.floor(Math.random() * charset.length)];
```
Etheryte, line 129: `Math.random()` is a pseudorandom number generator, I don't think we want to use it here. Similarly below in the sort function, I think this is not the way we want to approach this. The whole idea of using `getRandomValues()` in the original code is that we don't want to use `Math.random()`.

web/html/src/utils/functions.ts
```diff
@@ -116,16 +116,34 @@ function capitalize(str: string): string {
 }
 
 function generatePassword(): string {
-  const length = Math.floor(Math.random() * 10) + 15;
-  // See https://stackoverflow.com/a/68617567/1470607
-  const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789:-_";
-  const array = new Uint32Array(charset.length);
-  window.crypto.getRandomValues(array);
+  const length = 24; // fixed length for better security
```
Etheryte, line 119: I think this is not true? Using a fixed length reduces security, since it makes the space you need to brute force smaller. To exemplify, if we use only passwords of length 25, then there are `X^25` combinations where `X` is the number of symbols in our charset, but if we allow for example passwords of length 15-25, then there are `X^15 + X^16 + X^17 ... X^25` combinations.

Etheryte (changes requested):
I think we need to revisit this code, I don't think as is this is a correct way to generate passwords. Is there a ticket behind this change, what are the requirements?

Also pinging @szachovy for opinions, I think this is something up your alley.

szachovy (commented):
Can't we use some JS built-in function instead of creating the whole logic for this purpose? I am not a JS expert, but I would be very surprised if there is no cryptographic module (or derivative) available.

bisht-richa:
> I think we need to revisit this code, I don't think as is this is a correct way to generate passwords. Is there a ticket behind this change, what are the requirements?
> 
> Also pinging @szachovy for opinions, I think this is something up your alley.

@Etheryte 

These are the requirements:

- Use a fixed length of (24 or 32 characters), instead of a random length
- Guarantee that the password contains at least:
   - one lowercase
   - one uppercase
   - one number
   - one special char
- Replace Math.random() with window.crypto.getRandomValues()

```
Apart from that, the length can vary between 15 and 24 characters, which may be
considered too short in high-security environments, where longer passwords
(e.g., 32 characters or more) are preferred to increase complexity and
resistance to brute-force attacks. This is important also because even though
you use a broad character set, there is no guarantee that every character type
will be present in each password.

As an improvement, we can ensure the presence of each character category by
taking at least one character from each group (lowercase, uppercase, numbers,
special characters) and then filling in the rest of the password randomly.

The chosen character set is relatively narrow (only 64 characters: uppercase
letters, lowercase letters, numbers, and 3 special characters). A broader
character set with additional special symbols and other characters could make
the password harder to crack.
```

web/html/src/utils/functions.ts
```diff
@@ -116,16 +116,34 @@ function capitalize(str: string): string {
...
+
+  const array = new Uint8Array(length - 4); // Remaining characters
+  window.crypto.getRandomValues(array);
+
+  for (let i = 0; i < array.length; i++) {
+    password += allChars[array[i] % allChars.length];
   }
+
+  // Shuffle the password to avoid guaranteed character positions
+  password = password.split('').sort(() => 0.5 - Math.random()).join('');
```
nodeg, line 145: For shuffling, you could also use the [Fisher–Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle).

Etheryte (approved):

PR #9507, opened 2024-11-22
Changing OS Image build via kiwi to running via podman and adding eib pre-processing

Martin-Weiss:
## What does this PR change?

We worked on this as part of https://hackweek.opensuse.org/projects/build-edge-image-builder-iso-with-suse-manager

In this PR we have several changes:

1. switch from local installed kiwi to running kiwi via image
2. run eib pre-processing in case eib/eib.yaml exists

We hope someone else can take this forward as this is just an initial "PoC" even though it is working well for us.
Limitations that might need to be addressed as well:

- did not remove the installation of kiwi-ng
- did not test on SLES 12 build host
- did not add "air-gapped" - so the build-hosts need to be able to pull these images from the internet:

```
docker.io/dgiebert/edge-image-builder:1.2.7
registry.suse.com/bci/kiwi:10.1.10

```
- the edge-image-builder used is based on a new PR against edge-image-builder and we might need a final merge and new version of the image https://github.com/suse-edge/edge-image-builder/pull/618
- the kiwi image might not be available "officially", yet
 
We have been testing with https://github.com/Martin-Weiss/kiwi-image-micro-gpu-60 

## GUI diff

No difference.

## Documentation
- We need to explain the git repo eib/eib.yaml and point to the official EIB docs...
- We need to explain Podman requirements

## Test coverage
- We did not add test-cases so far - but build on x86, aarch64 and with repos having eib/eib.yaml and without need to be added.

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/images/kiwi-image-build.sls
+++ b/susemanager-utils/susemanager-sls/salt/images/kiwi-image-build.sls
@@ -39,10 +39,26 @@ mgr_buildimage_prepare_activation_key_in_source:
           susemanager:
             activation_key: {{ activation_key }}
 
+mgr_buildimage_prepare_kpartx_kiwi_yml:
+  file.managed:
+    - name: /etc/kiwi.yml
+    - contents: |
+        mapper:
+          - part_mapper: kpartx
+
+# EIB
+#
+mgr_buildimage_eib:
+  cmd.run:
+# only run eib and extract in case eib.yaml exists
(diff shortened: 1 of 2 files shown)


aaannz:
I don't think we can merge this as is.

- it changes build system for everyone using kiwi-ng. Build systems are critical part and IMO we cannot change it just like that
- using kiwi10 will fail our retail templates and everyone with kiwi schema older then 7.4 and would require manual profile upgrade

After discussion with @Martin-Weiss we agreed on different approach which I prepared at https://github.com/SUSE/spacewalk/pull/25986

rjmateus:
@aaannz @Martin-Weiss do you think this PR still makes sense, now that Ondrej have merged this one: https://github.com/SUSE/spacewalk/pull/25986

aaannz:
Indeed this PR was superseded by https://github.com/SUSE/spacewalk/pull/25986 and https://github.com/uyuni-project/uyuni/pull/10552
Let's close it.

PR #9535, opened 2024-12-03
Force the generation of bootstrap repositories in CT

srbarrios:
## What does this PR change?

This PR aims to have back the proper onboarding of salt bundle on Continuous Testing for at least those minions that have a product synchronized, or at least the Client Tools.
For that, we are force to regenerate the bootstrap repository, so the venv-enabled-x86_64.txt is generated.

Until now, we live without it, because we were installing salt-bundle manually through sumaform, in particular we do it at cloud-init or combustion stage, we do like that so sumaform can run the salt states to configure all the system after the VM is deployed.
But what it happen now is that our server detects a salt-bundle package already installed and don't do a proper onboarding of the minion, installing the package on the custom client tools repo, instead it's using the one previously installed, that's why we need to run a pre-condition scenario that clean the salt-bundle used during the deployment.

That's easy to observe when we look on the metrics of the time it takes the onboard in Uyuni compared with MLM 5.0 https://grafana.mgr.suse.de/d/ae3c5r75ow8owb/fitness-functions?from=now-7d&to=now&timezone=browser

This is a continuation of this revert: https://github.com/uyuni-project/uyuni/pull/9532

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- Cucumber tests were added

- [x] **DONE**

## Links

Ports:
- Manager-5.0
- Manager-4.3

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/init_clients/sle_minion.feature
+++ b/testsuite/features/init_clients/sle_minion.feature
@@ -4,6 +4,9 @@
 @sle_minion
 Feature: Bootstrap a Salt minion via the GUI
 
+  Scenario: Clean up sumaform leftovers on SLES minion
+    When I perform a full salt minion cleanup on "sle_minion"
+
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
 
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -1076,6 +1076,11 @@ When(/^I create the bootstrap repository for "([^"]*)" on the server((?: without
   get_target('server').run(cmd)
 end
 
+When(/^I create the bootstrap repositories including custom channels$/) do
+  get_target('server').wait_while_process_running('mgr-create-bootstrap-repo')
+  get_target('server').run('mgr-create-bootstrap-repo --auto --force --with-custom-channels', check_errors: false, verbose: true)
+end
+
 When(/^I install "([^"]*)" product on the proxy$/) do |product|
   out, = get_target('proxy').run("zypper ref && zypper --non-interactive install --auto-agree-with-licenses --force-resolution -t product #{product}")
   log "Installed #{product} product: #{out}"
--- a/testsuite/features/init_clients/buildhost_bootstrap.feature
+++ b/testsuite/features/init_clients/buildhost_bootstrap.feature
@@ -4,6 +4,9 @@
 @buildhost
 Feature: Bootstrap a build host via the GUI
 
+  Scenario: Clean up sumaform leftovers on build host
+    When I perform a full salt minion cleanup on "build_host"
+
   Scenario: Log in as admin user
     Given I am authorized for the "Admin" section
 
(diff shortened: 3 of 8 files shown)


testsuite/features/init_clients/sle_minion.feature
```diff
@@ -4,6 +4,9 @@
 @sle_minion
 Feature: Bootstrap a Salt minion via the GUI
 
+  Scenario: Clean up sumaform leftovers on SLES minion
```
szachovy, line 7: Why exactly are we doing this cleanups now, is there a reason behind it considering that we have been living without it for years?

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -1076,6 +1076,11 @@
   get_target('server').run(cmd)
 end
 
+When(/^I create the bootstrap repositories including custom channels$/) do
+  get_target('server').wait_while_process_running('mgr-create-bootstrap-repo')
+  get_target('server').run('mgr-create-bootstrap-repo --auto --force --with-custom-channels', check_errors: false, verbose: true)
```
szachovy, line 1081:
```
  def wait_while_process_running(process)
    repeat_until_timeout(report_result: true) do
      result, code = run("pgrep -x #{process} >/dev/null", check_errors: false)
      return [result, code] if code.nonzero?

      sleep 2
      result
    end
  end
```

It seems to me that it starts waiting -> ends waiting -> runs `mgr-create-bootstrap-repo` command as I do not see any parallel mechanisms that runs next commands without waiting for return code.
Can you provide me the screenshot that these 2 processes are present simultaneously in any time for at max 2 seconds?

szachovy (changes requested):

testsuite/features/step_definitions/command_steps.rb
srbarrios: mgr-create-bootstrap-repo it gonna be automatically triggered by the server in a periodical manner, this just tries to prevent the colision, still it could happen that when we run our second line, the server tries to also trigger it automatically.

We usually don't need to care about bootstrap creation because it's triggered periodically by the server, but... the default behavior don't include custom channels, which is our case here.

Hard to give you a probe of this :)

szachovy: Okay, I understand. Closing

testsuite/features/init_clients/sle_minion.feature
srbarrios: I edited the description to clarify it.

szachovy: Ok, closing

szachovy (approved):

srbarrios:
Note:
Review which salt-bundle package it gonna install, because it might use the one from the boostrap repository which is in fact the released package, instead of the package from development coming in the injected repository through sumaform.

PR #9540, opened 2024-12-05
Upgrade mgrpxy tool before installing the containerized proxy.

maximenoel8:
## For merging

 @ releng, could you merge squash with Commit name: Upgrade mgrpxy tool before installing the containerized proxy
Thanks !

## Context
When installing the proxy container on proxy host, mgrpxy is not updated.

## What does this PR change?
After bootstrapping the new proxy, update mgrpxy tool using spacewalk and reboot the proxy.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Port(s): https://github.com/SUSE/spacewalk/pull/25946

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/build_validation/init_clients/proxy.feature
+++ b/testsuite/features/build_validation/init_clients/proxy.feature
@@ -39,6 +39,10 @@ Feature: Setup containerized proxy
   Scenario: Wait until the proxy host appears
     When I wait until onboarding is completed for "proxy"
 
+  Scenario: Upgrade mgrpxy tool
+    Then I upgrade "proxy" with the last "mgrpxy" version
+    And I reboot the "proxy" minion through the web UI
+
   Scenario: Generate containerized proxy configuration
     When I generate the configuration "/tmp/proxy_container_config.tar.gz" of containerized proxy on the server
     And I copy the configuration "/tmp/proxy_container_config.tar.gz" of containerized proxy from the server to the proxy
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -752,46 +752,13 @@ def api_unlock
   end
 end
 
-# Function to fetch the current list of events
-#
-# @param hostname String The hostname of the system from requested
-def fetch_event_history(hostname)
-  output, _code = get_target('server').run("spacecmd -u admin -p admin system_listeventhistory #{hostname}", check_errors: true)
-  events = []
-  current_event = {}
-
-  output.split("\n").each do |line|
-    case line
-    when /^Id:\s+(\d+)$/
-      current_event[:id] = Regexp.last_match(1).to_i
-    when /^History type:\s+(.+)$/
-      current_event[:history_type] = Regexp.last_match(1)
-    when /^Status:\s+(.+)$/
-      current_event[:status] = Regexp.last_match(1)
-    when /^Completed:\s+(.+)$/
-      begin
-        current_event[:completed] = Time.parse(Regexp.last_match(1))
-      rescue ArgumentError
-        current_event[:completed] = nil
-      end
-    end
-
-    if current_event.key?(:id) && current_event.key?(:completed)
-      events << current_event
-      current_event = {}
-    end
-  end
-
-  events
-end
-
 # Function to get the highest event ID (latest event)
 #
-# @param hostname String The hostname of the system from requested
-def get_last_event_id(hostname)
-  events = fetch_event_history(hostname)
-  last_event = events.max_by { |event| event[:id] }
-  last_event ? last_event[:id] : nil
+# @param host String The hostname of the system from requested
+def get_last_event(host)
+  node = get_target(host)
+  system_id = get_system_id(node)
+  $api_test.system.get_event_history(system_id, 0, 1)[0]
 end
 
 # Function to trigger the upgrade command
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -751,3 +751,53 @@ def api_unlock
     file.flock(File::LOCK_UN)
   end
 end
+
+# Function to fetch the current list of events
+#
+# @param hostname String The hostname of the system from requested
+def fetch_event_history(hostname)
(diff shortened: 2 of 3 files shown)


testsuite/features/support/commonlib.rb
```diff
@@ -751,3 +751,53 @@ def api_unlock
...
+end
+
+# Function to get the highest event ID (latest event)
+#
+# @param hostname String The hostname of the system from requested
+def get_last_event_id(hostname)
+  events = fetch_event_history(hostname)
+  last_event = events.max_by { |event| event[:id] }
+  last_event ? last_event[:id] : nil
+end
```
srbarrios, line 795: We already have implemented a quite similar code, but using API instead of spacecmd.
Please take a look into `features/support/system_monitoring.rb` on `last_onboarding_duration`

```
    node = get_target(host)
    system_id = get_system_id(node)
    events = $api_test.system.get_event_history(system_id, 0, 10)
    onboarding_events = events.select { |event| event['summary'].include? 'certs, channels, packages' }
    last_event_id = onboarding_events.last['id']
    event_details = $api_test.system.get_event_details(system_id, last_event_id)
    # Convert XMLRPC::DateTime to Ruby's Time if necessary
    completed_time = event_details['completed'].is_a?(XMLRPC::DateTime) ? event_details['completed'].to_time : Time.parse(event_details['completed'])
    picked_up_time = event_details['picked_up'].is_a?(XMLRPC::DateTime) ? event_details['picked_up'].to_time : Time.parse(event_details['picked_up'])
```

I suggest to re-use part of this code, so instead of using spacecmd we use the API.

testsuite/features/support/commonlib.rb
```diff
@@ -751,3 +751,53 @@ def api_unlock
...
+
+# Function to trigger the upgrade command
+#
+# @param hostname String The hostname of the system from requested
+# @param package String The package name where it will trigger an upgrade
+def trigger_upgrade(hostname, package)
+  get_target('server').run("spacecmd -u admin -p admin system_upgradepackage #{hostname} #{package} -y", check_errors: true)
```
srbarrios, line 802: Could you please include upgradePackage method as part of `features/support/namespaces/system.rb` so we use the API instead of spacecmd?

testsuite/features/build_validation/init_clients/proxy.feature
```diff
@@ -39,6 +39,10 @@ Feature: Setup containerized proxy
   Scenario: Wait until the proxy host appears
     When I wait until onboarding is completed for "proxy"
 
+  Scenario: Upgrade mgrpxy tool
+    Then I upgrade "proxy" with the last "mgrpxy" version
```
srbarrios, line 43: Personally, I would simplify this step, by just do an upgrade of all available packages, you can do it via WebUI, or using system.schedulePackageUpdate from the API.

testsuite/features/support/commonlib.rb
maximenoel8: I switched everything to spacecmd command because system.schedulePackageInstall or system.schedulePackageUpgrade is not in the documentation and when I tried to use it, it says it doesn't exist

maximenoel8: When using the other upgrade method:

```
xmlrpc.client.Fault: <Fault -1: 'redstone.xmlrpc.XmlRpcFault: Could not find method: upgradePackage in class: com.redhat.rhn.frontend.xmlrpc.system.SystemHandler with params: [com.redhat.rhn.domain.user.legacy.UserImpl, java.lang.Integer, java.lang.String]'>
```

testsuite/features/build_validation/init_clients/proxy.feature
maximenoel8: Two things here:

testsuite/features/support/commonlib.rb
maximenoel8: I try to find a way of getting the packageId but don't have a solution. I tried:
 - package_name = client.packages.search.name(session_key, "mgrprxy")
 - client.system.listInstalledPackages(session_key, 1000010004)
 - client.system.listAllInstalledPackages(session_key, 1000010004)
I don't find mgrpxy...

testsuite/features/build_validation/init_clients/proxy.feature
srbarrios: >  - we should avoid a full upgrade for the potential dependency issues and time consuming action

I don't agree with this, in a Build Validation we must always have the latest released packages, plus, on top, the unreleased packages we are testing.

Otherwise, we will not have the same experience than a customer.


>  - we should use the API for non-testing step to gain in time and stability.
> I would prefer to keep it like that. It also shows what we are doing.

Agree on this, that's why I was proposing the API method here, and in another comment these days.

testsuite/features/support/commonlib.rb
srbarrios: Something you can dig is on how system_upgradepackage is developed on spacecmd, as spacecmd use API calls behind it. But ok, no worries, if it becomes hard, using spacecmd should be fine too.

testsuite/features/build_validation/init_clients/proxy.feature
srbarrios: You are right, at first, I though we all agreed on having systems up-to-date when we test on BV, but I was wrong, I clarified it after discussing with more people.
I will add that point as one of our agreements in the confluence page.

About the API, fully agree, and as fallback if we can't do it with API, it's fine doing it with spacecmd, as it is using API behind it.

srbarrios (approved):

testsuite/features/build_validation/init_clients/proxy.feature
```diff
@@ -39,6 +39,10 @@ Feature: Setup containerized proxy
...
+    And I reboot the "proxy" minion through the web UI
```
srbarrios, line 44: I'm not sure if we need to wait until the reboot event is completed to avoid issue on next scenarios.

srbarrios (approved):

NamelessOne91 (approved):

PR #9548, opened 2024-12-06
Ubuntu2404 set signedby

mcalmer:
## What does this PR change?

Improve handling of Debian repository configuration.
- Set "Signed-by" also for unsigned repos to prevent warning messages
- Do not set "Trusted" when the repo should be signed

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Port(s): https://github.com/SUSE/spacewalk/pull/25979

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/channels/channels.repo
+++ b/susemanager-utils/susemanager-sls/salt/channels/channels.repo
@@ -18,7 +18,9 @@ URIs: {{protocol}}://{{ args['token'] }}@{{hostname}}:{{port}}/rhn/manager/downl
 {%- endif %}
 Suites: {{ chan }}/
 Components:
+{%- if not salt['pillar.get']('mgr_metadata_signing_enabled', false) %}
 Trusted: yes
+{%- endif %}
 {%- if args['gpgkeyurl'] is defined and args['gpgkeyurl'].startswith('file:///') %}
 Signed-By: {{ args['gpgkeyurl'][7:] }}
 {%- endif %}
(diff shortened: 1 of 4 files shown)


susemanager-utils/susemanager-sls/salt/channels/channels.repo
```diff
@@ -18,7 +18,9 @@ URIs: {{protocol}}://{{ args['token'] }}@{{hostname}}:{{port}}/rhn/manager/downl
 {%- endif %}
 Suites: {{ chan }}/
 Components:
+{%- if not salt['pillar.get']('mgr_metadata_signing_enabled', false) %}
```
CDellaGiusta, line 21: Curiosity/question: why using "not-false"? Could you have used "true" instead?
I mean, something like {%- if salt['pillar.get']('mgr_metadata_signing_enabled', true) %}
Just for readability purposes (logically it's the same)

CDellaGiusta (commented):
looks good

susemanager-utils/susemanager-sls/salt/channels/channels.repo
mcalmer: pillar.get(<key>, <default>) The first define the key we want to get, but if this key does not exist we can define a default value. The safe value is "false" it is not signed. But we want to set "Trusted: yes" only when metadata are not signed.

CDellaGiusta: I was missing the fact that "false" is the default value if the item is not found. That clarifies the issue. Thanks

PR #9557, opened 2024-12-09
Changelog validation: test duplicate entries

cbbayburt:
Adds validation for duplicates in the new changelog entries in the same package.

## Test coverage
- Unit tests were added

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/.github/workflows/changelogs/changelogs.py
+++ b/.github/workflows/changelogs/changelogs.py
@@ -17,6 +17,7 @@ DEFAULT_LINE_LENGTH = 67
 DEFAULT_GIT_REPO = "uyuni-project/uyuni"
 DEFAULT_BUGZILLA_URI = "bugzilla.suse.com"
 
+
 class RegexRules:
     """Contains the regex rules for all the changelog checks
 
@@ -49,8 +50,8 @@ class RegexRules:
                 try:
                     # Every <issue-tracker> element should
                     # contain 'name' and 'regex' as children
-                    name = tracker.find('name').text
-                    regex = tracker.find('regex').text
+                    name = tracker.find("name").text
+                    regex = tracker.find("regex").text
                 except AttributeError:
                     raise Exception(f"Error parsing '{tracker_filename}': not a tracker XML file")
                 trackers[name] = regex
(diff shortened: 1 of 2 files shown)


agraul (dismissed):
It might be useful to take presence/absence of issue trackers into account when comparing entries. But this is already nice!

I let our lint.sh run locally and got a number of reports. I suggest you take a look yourself with 
```
~uyuni/uy-review[changelog-action]
% python/linting/lint.sh .github/workflows/changelogs/
```

cbbayburt:
Thanks for the review @agraul.

> It might be useful to take presence/absence of issue trackers into account when comparing entries.

What do you mean by this?
 
> I let our lint.sh run locally and got a number of reports. I suggest you take a look yourself with

I did a run and since we don't enforce it in this dir, I applied most of the suggestions selectively. Others rather hurt readability, especially in the test code. pylint needs to chill a bit :D

agraul:
> Thanks for the review @agraul.
> 
> > It might be useful to take presence/absence of issue trackers into account when comparing entries.
> 
> What do you mean by this?
> 

Entries where the "message" is the same except that one entry has a tracker and the other does not.

> I did a run and since we don't enforce it in this dir, I applied most of the suggestions selectively. Others rather hurt readability, especially in the test code. pylint needs to chill a bit :D

We can keep this unenforced, but IMO it should not be "code owned" by uyuni-project/python if we don't want to follow our Python standards.

cbbayburt:
> Entries where the "message" is the same except that one entry has a tracker and the other does not.

The reason for this addition was that I saw a couple cases where people ran the same `mkchlog` command twice by mistake and ended up having the exact same entry multiple times in a PR. So the purpose is simply to prevent this.

> We can keep this unenforced, but IMO it should not be "code owned" by uyuni-project/python if we don't want to follow our Python standards.

I agree. It shouldn't be anyway. This dir belongs to `workflows-reviewers`.

ycedres (approved):
I think that the relevant code is at the end of the `validate_chlog_entry` function where you look for duplicates and the rest is formatting and linting. So LTGM.

Issue #8998, opened 2024-07-02
You are passing a list of packages that contains duplicated packages names


jay2theb:
### Problem description

Unable to use the WebUI to patch or update ubuntu systems since 2024.05, updating the client manually via apt works without issue

### Steps to reproduce

1. Install or update to Uyuni release 2024.05 
2. Sync all Ubuntu 22.04 repo's
3. Register a Ubuntu 22.04 client, and use the WebUI to apply package updates

### Uyuni version

```bash
Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : Uyuni Server Stable
Name           : Uyuni-Server-release
Version        : 2024.05-230900.217.1.uyuni3
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes
Status         : out-of-date (version 2023.09-230900.207.8.uyuni3 installed)
Source package : Uyuni-Server-release-2024.05-230900.217.1.uyuni3.src
Summary        : Uyuni Server
```

### Useful logs

```bash
ID: install_gnupg_debian
    Function: pkg.installed
        Name: install_gnupg_debian
      Result: true
     Comment: All specified packages are already installed
     Started: 09:18:20.072058
    Duration: 4.371
         SLS: channels
     Changed: {}
----------
          ID: pkg_installed
    Function: pkg.installed
        Name: pkg_installed
      Result: false
     Comment: An exception occurred in this state: Traceback (most recent call last):
  File "/usr/lib/venv-salt-minion/lib/python3.10/site-packages/salt/state.py", line 2401, in call
    ret = self.states[cdata["full"]](
  File "/usr/lib/venv-salt-minion/lib/python3.10/site-packages/salt/loader/lazy.py", line 149, in __call__
    return self.loader.run(run_func, *args, **kwargs)
  File "/usr/lib/venv-salt-minion/lib/python3.10/site-packages/salt/loader/lazy.py", line 1234, in run
    return self._last_context.run(self._run_as, _func_or_method, *args, **kwargs)
  File "/usr/lib/venv-salt-minion/lib/python3.10/site-packages/salt/loader/lazy.py", line 1249, in _run_as
    return _func_or_method(*args, **kwargs)
  File "/usr/lib/venv-salt-minion/lib/python3.10/site-packages/salt/loader/lazy.py", line 1282, in wrapper
    return f(*args, **kwargs)
[...]

Issue #8121, opened 2024-01-09
Duplicate package NEVRA


santeri3700:
## Problem description

Uyuni reposync seems to import duplicate packages if the upstream repository changes a package without changing the NEVRA details (so only the checksum and contents differ). \
This also causes issues with custom package install/upgrade scheduling when managing package upgrades via API. \
It also wastes disk space since the older package data shouldn't be downloaded by clients (or at least DNF seems to download the later packages only, this might be undefined behaviour).

Pulp has had a similar issue in the past with duplicate NEVRA: https://github.com/pulp/pulp_rpm/issues/2691

EDIT: This behaviour is not exclusive to RPM based repositories. I've also encountered DEB package duplicates.

## Handling duplicate NEVRA?
My personal opinion is that there should not be packages with duplicate NEVRA and Uyuni reposync should replace existing but differing packages with the ones available in the upstream repository (because that's probably what the repository maintainer intended).

I know it is wrong (and probably also against spec) to change repository's packages without updating the NEVRA details but lots of repository maintainers have proven to be naughty (CentOS, AlmaLinux, EPEL, ELRepo, Nagios, Microsoft, MariaDB and Docker CE to name a few that I've seen doing this).

**What are your views on this from Uyuni's point? Are there other ways to combat related problems caused by duplicate NEVRA? Potential issues if NEVRA was treated as unique by Uyuni?**

## Workaround
I've currently just deleted duplicate packages/NEVRA manually to avoid the mentioned issues. \
This way I also keep the Uyuni channel closer to what is actually available in the upstream repository (as intended by the repository maintainer).

## Steps to reproduce

Run an Uyuni Server with lots of RPM repositories (with daily reposyncs) for a while and you are going to see duplicate NEVRA at some point. Alternatively you could set up your own RPM repository and change packages without changing the NEVRA details to simulate the same thing.

## Uyuni version

```bash
This affects older versions of Uyuni such as 2023.09 and 2023.10 too.
---------------------------------------------
Repository     : Uyuni Server Stable
Name           : Uyuni-Server-release
Version        : 2023.12-230900.210.2.uyuni3
Arch           : x86_64
Vendor         : obs://build.opensuse.org/systemsmanagement:Uyuni
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : Yes
Status         : up-to-date
Source package : Uyuni-Server-release-2023.12-230900.210.2.uyuni3.src
Summary        : Uyuni Server
```

[...]

PR #9571, opened 2024-12-13
Make the list for package actions unique so it can be passed to salt (bsc#1232042)

mcalmer:
## What does this PR change?

The package name+arch must be unique for salt pkg states.
Take care that this is correctly set in the Action Details

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **manual tested**

- [x] **DONE**

## Links

Fixes https://github.com/uyuni-project/uyuni/issues/8121
Fixes https://github.com/uyuni-project/uyuni/issues/8998
Port(s): https://github.com/SUSE/spacewalk/pull/25985

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/action/ActionManager.java
+++ b/java/code/src/com/redhat/rhn/manager/action/ActionManager.java
@@ -2009,10 +2009,36 @@ public class ActionManager extends BaseManager {
     public static void addPackageActionDetails(Collection<Action> actions,
             List<Map<String, Long>> packageMaps) {
         if (packageMaps != null) {
+            List<Map<String, Long>> pkgMaps;
+
+            if (actions.iterator().next().getActionType().equals(ActionFactory.TYPE_PACKAGES_REMOVE)) {
+                // our packages.pkgremove state is handling duplicates
+                pkgMaps = packageMaps;
+            }
+            else {
+                long noarch = PackageFactory.lookupPackageArchByLabel("noarch").getId();
+                long all = PackageFactory.lookupPackageArchByLabel("all-deb").getId();
+                Map<String, Map<String, Long>> uPkgMap = new HashMap<>();
+                // For other salt pkg states (pkg.installed), name + arch must be unique.
+                for (Map<String, Long> p : packageMaps) {
+                    long archId = p.getOrDefault("arch_id", noarch);
+                    String name = String.valueOf(p.get("name_id"));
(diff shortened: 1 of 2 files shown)


CDellaGiusta (approved):
looks good to my limited knowledge. Quick question: do you do unit tests?

mcalmer:
> looks good to my limited knowledge. Quick question: do you do unit tests?

not for this change. The function write directly to DB using mode queries, which cannot be mocked so easy. 
As this is a bugfix, we should not do a bit re-write in this PR as this is a port. 
A Bugfix should be as small as possible to not introduce more bugs.

java/code/src/com/redhat/rhn/manager/action/ActionManager.java
```diff
@@ -2009,10 +2009,36 @@ private static List<Map<String, Long>> removeDuplicatedName(List<Map<String, Lon
...
+                    long archId = p.getOrDefault("arch_id", noarch);
+                    String name = String.valueOf(p.get("name_id"));
+                    if (archId == noarch || archId == all) {
+                        if (uPkgMap.keySet().stream().noneMatch(k -> k.startsWith(name + "."))) {
+                            uPkgMap.put(name, p);
+                        }
+                    }
+                    else if (!uPkgMap.containsKey(name)) {
+                        String key = name + "." + p.get("arch_id");
+                        uPkgMap.put(key, p);
```
rjmateus, line 2033: I think I'm missing something (maybe because it Friday :) ), but in line 2031 you check map containsKey using just the name, and then adds the key as name+arch. 
I understand that it's looking for duplicate names in arch-specific and also in no-arch (or all) packages. But shouldn't the contains also check if the same key is already inserted?

mcalmer: It is a map. Same key would just overwrite the value which is in there. It does not matter which package version is updated, as our state just update the package name and not a specific version.
The arch just matter if `x86_64` and `i586` is installed together. This is how RedHat handle 32Bit packages.

rjmateus: ok, got it. Thank you Michael

rjmateus (approved):

mbussolotto (approved):

PR #9614, opened 2025-01-11
Removes Python 2 compatibility code for spacecmd

riddhi1806:
## What does this PR change?

This PR removes Python 2 compatibility code and replaces it with modern Python 3 code. This aligns with the project's decision to deprecate Python 2 support.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
No documentation needed: only internal and user-invisible changes.

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Issue(s): #6005

- [x] **DONE**

## Changelogs

- [x] No changelog needed

### DIFF
--- a/spacecmd/src/spacecmd/activationkey.py
+++ b/spacecmd/src/spacecmd/activationkey.py
@@ -33,10 +33,9 @@
 import re
 import shlex
 import gettext
-try:
-    from xmlrpc import client as xmlrpclib
-except ImportError:
-    import xmlrpclib
+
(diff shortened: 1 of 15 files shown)


agraul:
Thank you for the PR @riddhi1806! It's a good start :smiley:  Would it be possible to drop other Python 2 compatibility code as well? The issue https://github.com/uyuni-project/uyuni/issues/6005 lists another example, but I think there are more.

PR #9640, opened 2025-01-17
Fix systemd services check

Bischoff:
## What does this PR change?

For some reason, `systemd is-enabled <service>` started addinga carriage return (`0x0d`) at the end of its output, thus breaking the sanity check.

This PR fixes that.

## Links

No ports, Head only (5.0 does not seem affected).

## Changelogs

- [x] No changelog needed


srbarrios (approved):
Fine for now. But I would better try to refactor the output that the `run` command return.

Bischoff:
> Fine for now. But I would better try to refactor the output that the `run` command return.

I don't agree. Better have commands focused in only one task, and not too intelligent.

mcalmer (approved):

PR #9662, opened 2025-01-22
[health-check] Refactor the config usage

m-czernek:
## What does this PR change?

In this PR, I'm refactoring the config usage in the health check tool such that:

- We use `.toml` conf file instead of `.ini` file
- Callers don't have to instantiate the config object to parse properties
- Code doesn't have to pass around config objects to parse properties

*NOTE* This is a PR into a feature branch, not the main branch.

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


health-check/src/uyuni_health_check/config.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def load_config(self, conf_file=CONFIG_TOML_PATH):
+        if not os.path.exists(conf_file):
+            raise ValueError(f"Config file does not exist: {conf_file}")
+
+        with open(conf_file, "rb") as f:
+            self.parsed_config = tomli.load(f)
+
+    def load_jinja_template(self, template_name):
+        return self.jinja_env.get_template(template_name)
```
agraul, line 76: `load_jinja_template` is completely different from reading the `toml` config. I would drop this and change the module-level `load_jinja_template` as follows:

```py
@functools.lru_cache
def _jinja_env():
    return jinja2.Environment(loader=jinja2.FileSystemLoader(TEMPLATES_DIR))

def load_jinja_template(template: str) -> jinja2.Template:
    return _jinja_env().get_template(template)
```

This way the jinja environment is instantiated lazily on first access instead of at import-time and the code does not need to stay in the unrelated class.

health-check/src/uyuni_health_check/config.py
```diff
@@ -0,0 +1,78 @@
...
+    return _conf.jinja_env.get_template(template)
+
+def load_dockerfile_dir(dockerfile_dir: str) -> str:
+    return os.path.join(CONTAINERS_DIR, dockerfile_dir)
+
+def get_config_dir_path(component: str) -> str:
+    return os.path.join(CONFIG_DIR, component)
+
+def load() -> Dict:
+    return _conf.parsed_config
```
agraul, line 27: Do we need this? I couldn't find any usages with a quick grep

health-check/src/uyuni_health_check/config.py
```diff
@@ -0,0 +1,78 @@
...
+CONTAINERS_DIR = os.path.join(BASE_DIR, "containers")
+CONFIG_TOML_PATH = os.environ.get("HEALTH_CHECK_TOML", os.path.join(BASE_DIR, "config.toml"))
+
+def get_json_template_filepath(json_relative_path: str) -> str:
+    return os.path.join(TEMPLATES_DIR, json_relative_path)
+
+def load_jinja_template(template: str) -> Template:
+    return _conf.jinja_env.get_template(template)
+
+def load_dockerfile_dir(dockerfile_dir: str) -> str:
```
agraul, line 20: Am I right to assume we'll drop this once we build the images in OBS?

health-check/src/uyuni_health_check/config.py
```diff
@@ -0,0 +1,78 @@
...
+def load() -> Dict:
+    return _conf.parsed_config
+
+def load_prop(property: str) -> Any:
+    res = _conf.parsed_config.copy()
+    for prop_part in property.split('.'):
+        try:
+            res = res.get(prop_part, {})
+        except (AttributeError, ValueError):
+            res = None
```
agraul, line 35: Does it make sense to turn access errors into `None`? This might silence errors caused by bad callers (or bad configs)

agraul (commented):
Thank you for the PR, looks good! We're moving towards https://www.youtube.com/watch?v=1__lNTlj1_w and I like it :smile: 

I left a few questions and suggestions inline.

health-check/src/uyuni_health_check/main.py
```diff
@@ -1,19 +1,13 @@
-import uyuni_health_check.config_loader
 import click
 from rich.markdown import Markdown
 
+import uyuni_health_check.config
```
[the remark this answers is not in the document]
m-czernek: Not necessary, it was part of backwards-compatible code I forgot to remove

health-check/src/uyuni_health_check/config.py
agraul: The caller always needs to deal with the case that it asks for something from the config that can't be found  (`load_prop(x)` returns `None`). 

The question is: do we force the caller to deal with it by raising an exception or are we okay with the caller ignoring that? Both ways are okay, but since we don't have any config validation, it might make sense to not silence lookup errors (e.g. because the config is incomplete and mandatory values are missing) at runtime

health-check/src/uyuni_health_check/config.py
```diff
@@ -0,0 +1,71 @@
...
+TEMPLATES_DIR = os.path.join(CONFIG_DIR, "templates")
+CONTAINERS_DIR = os.path.join(BASE_DIR, "containers")
+CONFIG_TOML_PATH = os.environ.get("HEALTH_CHECK_TOML", os.path.join(BASE_DIR, "config.toml"))
+
+@functools.lru_cache
+def _init_jinja_env() -> jinja2.Environment:
+    return jinja2.Environment(loader=jinja2.FileSystemLoader(TEMPLATES_DIR))
+
+@functools.lru_cache
+def parse_config() -> Dict:
```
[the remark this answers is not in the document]
m-czernek: So passing around a config object and calling this method are, in essence, equivalent in this context.

`lru_cache` will cache the first call to the function. Every other time it's called, it returns the cached object, so that we don't have to parse the config file every time.

In this case, since the function has no arguments, this is equivalent to lazy initializing the config, saving it to a variable, and referencing the variable when necessary.

The benefit is that the caller doesn't have to instantiate any config object (the callers just want a property, they don't care about a config object), and we don't have to eagerly init a config object at module load time (it'll be initialized when first accessed).

health-check/src/uyuni_health_check/config.py
m-czernek: Ack, makes sense. Fixed - instead of silencing the error, we are now  explicit as to the input we wanted to get, where we failed, and the state of config.

ycedres (approved):

agraul (approved):

PR #9681, opened 2025-01-24
Make inter-server-sync to export/import dates in UTC (bsc#1221505)

meaksh:
## What does this PR change?

In this PR, we force ISSv1 endpoints to convert the `last_modified` value for synchronized objects into UTC. This prevents cache misses when the ISS master and ISS slave have different timezone config, where in the case of activating DST in different TZs, because of the different `last_modified` values, objects might be forced to re-sync even if it is not strictly necessary, thus increasing the ISS synchronization time. 

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23977

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/uyuni/common/rhnLib.py
+++ b/python/uyuni/common/rhnLib.py
@@ -15,6 +15,7 @@
 #
 
 import os
+import calendar
 import re
 import time
 from datetime import tzinfo, timedelta
@@ -129,7 +130,7 @@ def timestamp(s):
     timeval = list(time.strptime(s, format_string))
     # No daylight information available
     timeval[8] = -1
-    return time.mktime(tuple(timeval))
+    return calendar.timegm(tuple(timeval))
 
 
 def checkValue(val, *args):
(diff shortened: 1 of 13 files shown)


vzhestkov (changes requested):
I would say we need to align the code according to check style suggestions.

meaksh:
@vzhestkov it should be done now :+1:

python/uyuni/common/rhnLib.py
```diff
@@ -129,7 +130,7 @@ def timestamp(s):
     timeval = list(time.strptime(s, format_string))
     # No daylight information available
     timeval[8] = -1
-    return time.mktime(tuple(timeval))
+    return calendar.timegm(tuple(timeval))
```
vzhestkov, line 133: This part is a bit confusing for me. I'm not fully sure if it's the best way to do it.
I don't like 3 points here:
1. New import of `calendar` just for this calculation
2. The logic of setting daylight information explicitly with `timeval[8] = -1`
3. Now it returns `int` instead of `float` as before

Maybe we can do it this way instead?:
```
return time.mktime(tuple(timeval))-time.timezone
```

vzhestkov (approved):
LGTM

PR #9703, opened 2025-01-29
Simplified Proxy Onboarding 

rjpmestre:
> [!WARNING]
>  This PR is a work in progress. While the core functionality is in place, there are still multiple scenarios to address and details to refine. This PR is intended to gather early feedback on the current implementation, considering the known issues (listed at the end). 

## Overview
This PR implements a new feature that aims to simplify the proxy onboarding process. 
Currently, setting up a proxy involves the following steps: onboarding the minion, creating a proxy configuration file, installing Podman and mgrpxy (provided by uyuni tools), and finally running mgrpxy install.

For this initial feature iteration, our goal is to streamline the process by extending the existing web UI functionality (Container Based Proxy Configuration). Once the user fills in the necessary information in a form, the system will automatically set up the minion as a proxy with minimal manual intervention.

## Pre-requisites
- minion is onboarded
- minion is reachable
- minion has access to Uyuni Client Tools (via Software Channels/spacewalk-common-channels)
- minion can be convertible to a proxy* or has a proxy entitlement

> [!NOTE]
>  Whether a minion can be converted to a proxy depends on the running manager:
> - **SUSE Manager (SUMA)** only allows transactional servers (for supportability reasons).
> - **Uyuni** allows all minions.
> - In both cases, the minion must not already be a proxy.

## Base Setup
If an onboarded minion meets the prerequisites to become a proxy, the first step is to go to the minion's Overview page. From there:

1. Navigate to **Details > Properties**.
2. Add "Proxy" under **Add-On System Types**.
3. Click **Update Properties**.
This action will make a new main tab, **Proxy**, visible.
Alternatively, the user can achieve the same result by clicking **Convert to Proxy** in the top-right corner of the **Overview** page.
The **Proxy** tab is intended to display data and provide operations related to the proxy. At this stage, it will only contain the form used to apply the proxy configurations to the minion.

## Flow
Once the user fills in the form and it is valid, the "**Apply**" button will be enabled. Clicking it will trigger the following flow:

- **Validate Inputs**: Ensure that all required inputs, including registry URLs, are provided and valid.
- **Create Proxy Configuration Files** (based on the Container-Based Proxy Configuration feature):
    - config.yaml
    - httpd.yaml
    - ssh.yaml
- **Save Configuration to Pillars** (with category = "proxy")
- **Apply Salt State** to:
    - determine the mgrpxy operation to execute (install or upgrade) and if it is a transactional server - eases following steps
    - confirm podman is installed and running;
    - confirm mgrpxy is installed;
    - copy configuration files to "/etc/uyuni/proxy";
    - if minion's local operating system is not transactional
        - execute mgrpxy install/ugrade command with custom registries if provided and log it into "/var/log/mgrpxy_install.log"
    - otherwise,
        - create a systemd "apply_mgrpxy.service" defined in "/etc/systemd/system/apply_mgrpxy.service" service file that should
            - execute mgrpxy install/ugrade command with custom registries if provided and log it into "/var/log/mgrpxy_install.log"
            - remove itself if the install/ugrade command was successful
        - enable "apply_mgrpxy.service"


web/html/src/manager/minion/proxy/proxy-config-messages.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from "react";
+
+import { Messages } from "components/messages/messages";
+
+type SuccessType = boolean | undefined;
+
+export const ContainerConfigMessages = (success: SuccessType, messagesIn: React.ReactNode[], loading: boolean) => {
+  if (success) {
```
Etheryte, line 8: This structure works, but feels a bit clunky with multiple branches all wrapped in a `Messages`.

Since `Messages` also returns null if there are no items, something along these lines would also work:

```tsx
let items: MessageType[] = [];
if (success) {
  items = [{
    severity: "success",
    text: <p>{t("Proxy configuration successfully applied.")}</p>,
  }];
} else if ... { }

return <Messages items={items} />;
```

What do you think?

web/html/src/manager/minion/proxy/proxy-config.tsx
```diff
@@ -0,0 +1,731 @@
+import * as React from "react";
```
Etheryte, line 1: The code here looks mostly fine, but I would recommend breaking this up into multiple files to make it easier to digest, right now it's a pretty big bite. One option would be types and enums in one file, utils like stuff around file readers in another, etc, but I don't have a strong preference for how to chop this up.

rjpmestre: Pushed an update addressing it. I feel it improved readability a lot. Let me know if you'd like any further adjustments to it. Ty for also suggesting solutions to it!

web/html/src/manager/minion/proxy/proxy-config.tsx
```diff
@@ -0,0 +1,527 @@
...
+  imageNames,
+  modelDefaults,
+  readFileFields,
+  restoreRegistryInputs,
+  retrieveRegistryTags,
+} from "./proxy-config-utils";
+
+export function ProxyConfig({
+  serverId,
+  isUyuni,
```
Etheryte, line 38: Recommend using `import { isUyuni } from "core/user-preferences";` instead, we have it globally available since so many places need this information.

Etheryte (dismissed):
The frontend parts look good, thank you for all the work you've put into this. 👍

web/html/src/manager/minion/proxy/proxy-config.tsx
Etheryte: Yes, this is considerably easier to digest, thank you very much. 👍

web/html/src/manager/minion/proxy/proxy-config-utils.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (!registryUrl) {
+    setErrors((prev) => ({ ...prev, [name]: [] }));
+    setTagOptions((prev) => ({ ...prev, [name]: [] }));
+    return;
+  }
+
+  try {
+    const response = await Network.post("/rhn/manager/systems/details/proxy-config/registry-url", {
+      registryUrl: registryUrl,
+      isExact: name !== "registryBaseURL",
```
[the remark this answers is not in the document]
rjpmestre: yep yep, also not a fan of having magic number around. Will address it!

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -2688,4 +2689,20 @@ public Set<SAPWorkload> getSapWorkloads() {
     public void setSapWorkloads(Set<SAPWorkload> sapWorkloadsIn) {
         sapWorkloads = sapWorkloadsIn;
     }
+
+    /**
+     * Checks if a server in convertible to a proxy.
+     * @return true if the server is convertible to a proxy, false otherwise
+     */
+    public boolean isConvertibleToProxy() {
+        return !isProxy() && (ConfigDefaults.get().isUyuni() || isSLEMicro5());
```
aaannz, line 2698: This needs to be expanded for SLEMicro6.x and also SLE15SP6 and SLE15SP7.

susemanager-utils/susemanager-sls/salt/proxy/apply_proxy_config.sls
```diff
@@ -0,0 +1,146 @@
...
+      - file: /etc/uyuni/proxy/config.yaml
+      - file: /etc/uyuni/proxy/httpd.yaml
+      - file: /etc/uyuni/proxy/ssh.yaml
+      - pkg: podman_installed
+      - pkg: mgrpxy_installed
+
+# The system will run this service to enable apply_proxy_config.service after reboot
+enable_apply_proxy_config_service:
+  service.running:
+    - name: apply_proxy_config.service
```
nadvornik, line 126: This does not work for me. Should it be service.enabled?

nadvornik: correct is
```
cmd.run:
    - name: systemctl enable apply_proxy_config.service
```

java/code/src/com/suse/manager/webui/controllers/ProxyConfigurationController.java
```diff
@@ -0,0 +1,196 @@
...
+     * @param proxyController instance to register.
+     * @param jade            Jade template engine
+     */
+    public void initRoutes(ProxyConfigurationController proxyController, JadeTemplateEngine jade) {
+        get("/manager/systems/details/proxy-config",
+                withCsrfToken(withDocsLocale(withUserAndServer(this::getProxyConfiguration))),
+                jade
+        );
+        post("/manager/systems/details/proxy-config", withUser(this::updateProxyConfiguration));
+        post("/manager/systems/details/proxy-config/get-registry-tags", withUser(this::getRegistryTags));
```
rjmateus, line 122: Ifg I got it right, it doesn0't change anything on the server side. Even the implementation method is called "get...". Shouldn't this use GET HTTP request instead of POST? The fields can be passed as query parameters.

java/code/src/com/suse/manager/webui/services/SaltServerActionService.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016--2024 SUSE LLC
```
rjmateus, line 2: Why changing this on a file that doesn't have any other changes?

java/code/src/com/suse/manager/webui/templates/system-common.jade
```diff
@@ -14,6 +14,10 @@ else
 
 .spacewalk-toolbar-h1
   .spacewalk-toolbar
+    if !server.isConvertibleToProxy()
```
rjmateus, line 17: Shouldn't this be the other way around? We want to show this link if the machine is convertible to proxy, right?

rjmateus (commented):
1th round or review

java/code/src/com/redhat/rhn/common/RhnError.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.redhat.rhn.common;
+
+import java.io.Serializable;
+
+/**
+ * Represents a generic error in scope of RHN
+ */
+public class RhnError implements Serializable {
+    private final String message;
```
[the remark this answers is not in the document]
rjpmestre: This is part of a structure I had already considered implementing in the past, but at the time, I couldn’t justify adding it because I only had a few/specific errors that only mattered to one caller.

Right now, the service only responds to the frontend, so I kept errors as simple strings to avoid introducing complexity.
But once you mentioned the possibility of providing this as an API, I saw it as the right opportunity to introduce this.
Because, for eg, if we need to standardize our error responses, it would be useful to include additional fields like an error type or a reference value. With RhnError in place, we can easily adapt this without major refactoring.

java/code/src/com/redhat/rhn/common/RhnErrorReport.java
```diff
@@ -0,0 +1,70 @@
...
+package com.redhat.rhn.common;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Represents a collection of errors that can be reported.
+ */
+public class RhnErrorReport {
```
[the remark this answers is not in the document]
rjpmestre: I'm not sure if you have a specific one in mind. I ended up creating this so it would fit with the error report strategies and what i wrote [here](https://github.com/uyuni-project/uyuni/pull/9703#discussion_r2019304512)

java/code/src/com/suse/manager/webui/controllers/ProxyConfigurationController.java
rjpmestre: Yeah, you're right. The naming isn't ideal, and I overlooked this. We don’t need a body to pass just two values. It makes more sense to follow your suggestion and convert it to a GET request with query params.

nadvornik: One of the parameters is registry url. It would look quite ugly and we don't have support for quoting in Network.get.
I'd leave it as it is now.

java/code/src/com/redhat/rhn/common/RhnError.java
```diff
@@ -0,0 +1,37 @@
...
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package com.redhat.rhn.common;
+
+import java.io.Serializable;
+
+/**
+ * Represents a generic error in scope of RHN
```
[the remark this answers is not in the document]
nadvornik: fixed

nadvornik (approved):

PR #9704, opened 2025-01-30
Experimental way of testing HubController api endpoint call

CDellaGiusta:
## What does this PR change?

**EXPERIMENTAL!**

The aim is to test a HubController API call without actually setting up a remote server and using tools like postman.
The idea is that unit tests can be done to test the behaviour and the returned values, while a single manual test can be done from the external (e.g. postman) to check that the API endpoint is really visible.

In this test I tried to test the **/hub/managerinfo** API endpoint, and of course I will use the testing environment to build and test the APIs that are still missing. 

Although I guess I will be expanding the way things are tested, I'd like somebody else to have a look at it to see if I'm on the right track before proceeding further on this road. All comments and suggestions are extremely welcome.

## GUI diff
- [x] **DONE**

## Documentation
- No documentation needed: experimental
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
- [x] **DONE**

## Changelogs
- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/hub/test/HubManagerTest.java
+++ b/java/code/src/com/suse/manager/hub/test/HubManagerTest.java
@@ -22,6 +22,7 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 import com.redhat.rhn.common.conf.Config;
 import com.redhat.rhn.common.conf.ConfigDefaults;
+import com.redhat.rhn.common.hibernate.HibernateFactory;
 import com.redhat.rhn.common.security.PermissionException;
 import com.redhat.rhn.domain.credentials.CredentialsFactory;
 import com.redhat.rhn.domain.credentials.HubSCCCredentials;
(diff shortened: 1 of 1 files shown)


mcalmer:
@CDellaGiusta This looks good. I would implement this in a new class called `HubControllerTest` to not overload HubManagerTest and we are here testing the Controller. It might be also a good idea to put "simulateHubControllerApi" and "simulateApiEndpointCall" methods in a class called like `ControllerTestUtils` and make them public and document them.
I think in future we might get endpoints similar to /hub and they need to be tested as well. Having a TestUtils class which provide the utilities needed for this might be usefull. We can also check if something like already exists where we can put this things to.

CDellaGiusta:
@mcalmer 
>  We can also check if something like already exists where we can put this things to.
I decided to split the "simulation" test method in two parts

_simulateHubControllerApi_ has all the dependencies to HubFactory and HubController
_simulateApiEndpointCall_ has no such dependencies and so SparkTestUtils class  could be a candidate class where to move this function. I expect there will be some instances of that with a non-null http request body etc.

If we decide to go for a separate _ControllerTestUtils_ class, then we can keep these two methods in the same place.

Let's see what the others think

java/code/src/com/suse/manager/hub/test/HubManagerTest.java
```diff
@@ -626,4 +641,82 @@ else if (char.class.equals(clazz)) {
...
+                throw new IllegalArgumentException("unsupported role " + role.getLabel());
+        }
+
+        return simulateApiEndpointCall(apiEndpoint, httpMethod, authBearerToken);
+    }
+
+    //maybe this method could be moved to SparkTestUtils?
+    private Object simulateApiEndpointCall(String apiEndpoint, HttpMethod httpMethod, String authBearerToken)
+            throws Exception {
+        Optional<RouteMatch> routeMatch = spark.Spark.routes()
```
rjmateus, line 702: would be a option to use the route find method, which would be able to handle wildcard?
In the URL you could use the expected URL to be used (like: `/systems/1000001/bla`) which would math the route pattern like `/systems/:ID/bla`.
You can see an example in here: https://github.com/rjmateus/uyuni/blob/uyuni_hack_authorization_check/java/code/src/com/redhat/rhn/frontend/servlets/AccessFilter.java#L116-L118

PR #9705, opened 2025-01-30
Improve debian reposync logging (bsc#1227859)

meaksh:
## What does this PR change?

This PR refactors and enhances logging around debian reposync executions.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **no tests around reposync logging**

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24816

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


python/spacewalk/common/repo.py
```diff
@@ -426,12 +427,11 @@ def _get_release_index_from_file(self) -> typing.Dict[str, "DpkgRepo.ReleaseEntr
         # Repo format is not flat
         if not self.is_flat():
             if self.gpg_verify and not self._has_valid_gpg_signature(local_path):
-                # pylint: disable-next=logging-format-interpolation,consider-using-f-string
-                logger.error("GPG verification failed: {}".format(release_file))
+                # pylint: disable-next=logging-fstring-interpolation
+                logger.error(f"{DpkgRepo.GPG_VERIFICATION_FAILED}: {release_file}")
```
agraul, line 431: Can we fix this completely?

python/spacewalk/common/rhnLog.py
```diff
@@ -309,15 +309,15 @@ def _exit():
 atexit.register(_exit)
 
 
-def log_level_to_logging_constant(rhnLog_log_level: int):
+def log_level_to_logging_constant(rhnlog_log_level: int):
```
agraul, line 312: `rhnLog` is still called `rhnLog`, the name was chosen on purpose to reflect that.

agraul (dismissed):

python/spacewalk/satellite_tools/spacewalk-repo-sync
```diff
@@ -65,100 +66,226 @@ def main():
...
-    initCFG('server.satellite')
-    CFG.set('DEBUG', log_level)
+        # if no -v flag is passed, use "debug" setting from rhn.conf
+        with cfg_component(None) as cfg:
+            log_level = cfg.get("DEBUG", 0)
+
+    # Apparently we need to call initCFG and have CFG available
+    # to prevent some errors accessing CFG later on during package
+    # import.
+    initCFG("server.satellite")
```
agraul, line 262: Can we please fix this as well? I'm okay with doing it in another PR, as we already have the patch we shouldn't leave it lying around until it does not apply anymore.

python/spacewalk/common/repo.py
m-czernek: This should be equivalent and should pass the linting error (f-strings are eagerly evaluated, which is why they are discouraged in logging, since logging may never fire)

```suggestion
                logger.error("%s: %s", DpkgRepo.GPG_VERIFICATION_FAILED, release_file)
```

python/spacewalk/common/rhnLog.py
meaksh: Ok, I'll revert this. It was raised as an issue by SonarCloud, but I agree it makes sense in this case to keep the `rhnLog_log_level` name.

python/spacewalk/satellite_tools/spacewalk-repo-sync
meaksh: I'd vote for doing a separated PR.

In our initial attempt, we noticed that other python files that are part of the reposync execution stack were relying on CFG being preinitializated to `server.satellite`, and that causes some issues in certain corner cases. Instead of going forwad with the refactor, in the context of the bug fix, we just isolated the actual logging improvement from the refactor to use context manager (that was causing some issues). I agree we can just fix them, but we didn't by then.

I would say we can prepare another PR with the refactor, plus any extra fixes required to use context managers, just after this one.

python/spacewalk/satellite_tools/spacewalk-repo-sync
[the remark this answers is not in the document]
agraul: Changing our lint.sh is only part of the story, the GH Action uses a `.py` filter. Having these executable scripts commit to the repository does not follow current best practices (where entry points are defined in project metadata and generated during package installation), but that's a different topic.

I think we should think about the filtering we do in GH Actions. The whitelist approach is getting a bit cumbersome lately where we miss out on linting (but still get a green checkmark!).

python/spacewalk/common/repo.py
```diff
@@ -466,20 +476,17 @@ def _get_release_index_from_file(self) -> typing.Dict[str, "DpkgRepo.ReleaseEntr
...
-                        logging.error(
-                            # pylint: disable-next=logging-format-interpolation,consider-using-f-string
-                            "GPG verification failed: '{}'. \
-                                           Raising GeneralRepoException.".format(
-                                release_file
-                            )
+                        logger.error(
+                            "%s: %s. Raising GeneralRepoException.",
+                            DpkgRepo.GPG_VERIFICATION_FAILED,
+                            release_file,
```
agraul, line 482: There are other instances of this as well. I'll fix it and force push with a clean history

meaksh: Thanks!

python/spacewalk/common/repo.py
```diff
@@ -315,11 +315,8 @@ def _has_valid_gpg_signature(self, uri: str, response=None) -> bool:
                         )
                     )
             else:
-                logging.error(
-                    # pylint: disable-next=logging-format-interpolation,consider-using-f-string
-                    "No release file found: '{}'. Raising GeneralRepoException.".format(
-                        uri
-                    )
+                logger.error(
+                    "No release file found: '%s'. Raising GeneralRepoException.", uri
```
m-czernek, line 319: You removed the `Raising GeneralRepoException` from the previous logger, but kept it for the rest. Should we remove the rest as well? Personally, I prefer not having it in the logger message because:

a) it's redundant
b) it's not consistent, i.e. sometimes this ends in a full stop, other times in an exclamation mark
c) it could become obsolete very easily if someone were to change the actual exception (though unlikely)

But, if you'd prefer to keep it, I'm OK with that as well, in which case you might re-add it above :)

m-czernek (approved):
+1, LGTM with some minor nit comments as usual :+1:

python/spacewalk/common/repo.py
agraul: Right, we can remove it everywhere :+1:

python/linting/lint.sh
```diff
@@ -56,14 +56,22 @@ function execute_lint() {
...
+  if [ ! -z "$LIST_OF_FILES" ]; then
+    egrep '^#!/usr/bin/python|^#!/usr/bin/env python' $LIST_OF_FILES | cut -d":" -f1 | sort -u | tr '\n' ' '
+  fi
+}
+
 function main() {
   ensure_latest_container_image
   if [[ "${CHECK_ALL_FILES}" == "true" ]]; then
-    files="$(get_all_py_files)"
+    files="$(get_all_py_files)$(get_all_files_with_python_shebang)"
```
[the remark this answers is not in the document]
agraul: I guess it makes sense to split these, thanks!

I'm sure if we want to include this commit at all, on PR we're not checking those files and I'd like to keep the number of commits that reformat down (so that we can add them to  `.git-blame-ignore-revs` easily). It's something we should revisit shortly either way.

agraul:
@m-czernek I think I addressed all your open points now. Please take another look when you have time.

BTW, the checkstyle now failed for the test (I updated it -> it get's checked) because black is reformatting parts I didn't touch. I think I saw this before when we updated the black version in the container, I just hope this doesn't happen all the time

m-czernek (approved):
Perfect, +1 from my side :) (though we need to squash fixups, right?)

Regarding the checkstyle, I've seen that before, but IMO this is unlikely to have been caused by a new Black version - trailing comma was a formatting style even when I first formatted our Python parts. I can't explain how this has been missed, honestly.

agraul:
As discussed in the squad, I've dropped the partial reformat in favor of a global reformat commit.

PR #9708, opened 2025-01-30
Check SELinux labels on container volumes

Bischoff:
## What does this PR change?

This PR officializes Vladimir's test for SELinux problems.

Feel free to object this PR. It adds 5 minutes to the sanity checks!

Another option could be to run it at the end of the testsuite, at same time we check the logs. The idea to do it at the beginning is that we can spare ourselves a run if SELinux is going to block us.

## Links

Port(s):
 * 5.0:

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/core/allcli_sanity.feature
+++ b/testsuite/features/build_validation/core/allcli_sanity.feature
@@ -25,6 +25,7 @@ Feature: Sanity checks
     And socket "tftp" is active on "server"
     And service "tomcat" is enabled on "server"
     And service "tomcat" is active on "server"
+    And files on container volumes should all have the proper SELinux label
 
 @proxy
   Scenario: The proxy is healthy
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -762,6 +762,14 @@ Then(/^socket "([^"]*)" is active on "([^"]*)"$/) do |service, host|
   raise ScriptError, "Service #{service} not active" if output != 'active'
 end
 
+Then(/^files on container volumes should all have the proper SELinux label$/) do
+  node = get_target('server')
+  cmd = '[ "$(sestatus 2>/dev/null | head -n 1 | grep enabled)" != "" ] && ' \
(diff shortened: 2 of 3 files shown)


nadvornik (approved):
LGTM

testsuite/features/build_validation/core/allcli_sanity.feature
```diff
@@ -25,6 +25,7 @@ Feature: Sanity checks
     And socket "tftp" is active on "server"
     And service "tomcat" is enabled on "server"
     And service "tomcat" is active on "server"
+    And files on container volumes should all have the proper SELinux label
```
srbarrios, line 28: I guess it right that this label must be there not only if we are under a transaction-system, but also a regular one?
I'm thinking in our AWS deployment where we deploy our server on top of Leap, or the idea of supporting non-transactional systems as Host OS in the future versions

Bischoff: The labels are on any system with SELinux.

But the container volumes are on containerized systems only.

From what I understood, the problems emerge only when we are combining SELinux and container volumes, https://blog.christophersmart.com/2021/01/31/podman-volumes-and-selinux/ .

The transactional aspect does not play a role AFAIK. It just happens to be a coincidence that those systems are all shipped with SELinux.

For your question, it depends whether you are deployer with containers (5.0 or 5.1) or without (4.3) in the cloud.
I hope I answered it?

testsuite/features/step_definitions/command_steps.rb
```diff
@@ -762,6 +762,14 @@
   raise ScriptError, "Service #{service} not active" if output != 'active'
 end
 
+Then(/^files on container volumes should all have the proper SELinux label$/) do
+  node = get_target('server')
+  cmd = 'find /var/lib/containers/storage/volumes/*/_data -exec ls -Zd {} \; | grep -v ":object_r:container_file_t:s0 "'
+  output, _code = node.run_local(cmd, check_errors: false)
+  log output if output != ''
```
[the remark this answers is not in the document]
Bischoff: a bit too verbose, but ok. done.

PR #9712, opened 2025-01-31
Avoid SELinux + container volumes clashes

Bischoff:
## What does this PR change?

Avoid using `/smdba-backup-test` in the root filesystem, not in a container volume. SELinux does not like that.

Use instead `/var/spacewalk/db-backup`, like in our documentation.

## Links

Issue(s): https://bugzilla.suse.com/show_bug.cgi?id=1223142
Port(s):
 * 5.0: https://github.com/SUSE/spacewalk/pull/26329
 * 4.3: none, containerized server only

## Changelogs

- [x] No changelog needed


nadvornik (approved):
LGTM

PR #9724, opened 2025-02-04
GH Recommended tests plus some renaming of GH tests workflows

srbarrios:
## What does this PR change?

Related card: https://github.com/SUSE/spacewalk/issues/21015

This PR make use of the recommended tests implementation, based in our code coverage to select those tests that triggers code part the Java files changed. By parsing this list of tests, we generate a new YAML file that will be executed by Cucumber, so **by default** we only run these sub-set of tests.

In addition, we tweaked the full acceptance tests so they run if we add the label `run-full-testsuite` into the Pull Request.

Bonus: We renamed some jobs to make it more readable.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- Cucumber tests were added

- [x] **DONE**

## Links

No ports.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


jordimassaguerpla:
I am not familiar with the "labeled" trigger. Will the tests only run when the label is added? What if we add a new commit to fix a tests? Why you do not want to run the full test-suite? Aren't we losing some "safeness"?

.github/workflows/recommended_tests.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - name: Checkout Pull Request code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Use Node.js 18
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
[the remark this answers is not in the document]
srbarrios: No special reason as far as I'm aware.
This code has been just move from the previous GH job, into that workflow and we did not updated its node version:

image

I can make a try with version 24

srbarrios:
> I am not familiar with the "labeled" trigger. Will the tests only run when the label is added?

The test will be skipped if the label is not found. But the tests will not only be triggered when the label is added, it will be trigger if:
```
  pull_request:
    types:
      - opened
      - reopened
      - synchronize
```

> What if we add a new commit to fix a tests? 

Then all the checks will be removed, and all the tests will be triggered again.
If the new label is not in place, the job is skipped.

> Why you do not want to run the full test-suite?

By default, if we only run those tests that we know will cover the changes, we will speed up the feedback given and unblock faster the possibility to merge the PR.

> Aren't we losing some "safeness"?

Make it mandatory to run all the tests or just the recommended tests, it's fair to discuss and agree with all the team. Maybe we can bring this into our retro, or in a e-mail to galaxy-devel.

On GH validation, I think that aside of trying to improve the amount of tests that can be executed in that system, which it must be our focus this year; we should also find ways to speed up the feedback and reduce the amount of resources consumed per PR.

srbarrios:
There is an issue on this PR that I need to discuss with an admin of the repository:

image

As you can see we expect a status for these particular checks, I see different problems here:

- For some reason, this PR is not executing the workflows that are in master branch, but instead, it's trying to use the workflows that we have in srbarrios:gh-recommended-tests branch.
- We need to review the name of these checks after merge this PR, as we are renaming them.
- We need to agree with the team if we want to still have mandatory the full list of tests or only the recommended tests.


From what I see @nodeg is an admin now. FYI ^

.github/workflows/acceptance_tests_secondary.yml
```diff
@@ -1,19 +1,25 @@
...
   cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
 jobs:
-  acceptance-tests-secondary:
-    # Skip running scheduled jobs on forks
-    if: (github.repository == 'uyuni-project/uyuni' || github.event_name != 'schedule')
+  acceptance-tests:
+    name: "0"
+    # Skip running scheduled jobs on forks and only run if "run-full-testsuite" label is present
+    if: >-
+      (github.repository == 'uyuni-project/uyuni' || github.event_name != 'schedule') &&
```
srbarrios, line 21: Refactor this comparison, so this is not skipped when we run master.

.github/workflows/acceptance_tests_common.yml
```diff
@@ -34,101 +37,114 @@ jobs:
...
+        run: python .github/scripts/get_tests.py '${{ inputs.recommended_tests }}' > testsuite/run_sets/filter.yml
+      - name: Generate recommended tests as YAML
+        if: ${{ inputs.recommended_tests != '' }}
+        run: |
+          ./testsuite/podman_runner/19_generate_recommended_tests_yml.sh
+          if [ ! -s /testsuite/run_sets/recommended_tests.yml ]; then
+            echo "⚠️ WARNING: No recommended tests found. Running the full test suite."
+            gh pr edit ${{ github.event.pull_request.number }} --add-label "run-full-testsuite"
+            exit 1 # Exit with error to stop the workflow
+          fi
```
srbarrios, line 94: elsif when we have more than 20

testsuite/podman_runner/19_generate_recommended_tests_yml.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -xe
+sudo -i podman exec controller bash -c "cd /testsuite && rake utils:filter_secondary[/testsuite/run_sets/filter.yml,/testsuite/run_sets/secondary.yml,/testsuite/run_sets/recommended_tests.yml] "
```
srbarrios, line 3: We also need to pass the `/testsuite/run_sets/secondary_parallizable.yml`

.github/workflows/acceptance_tests_common.yml
srbarrios: For now comment the label implementation

.github/workflows/acceptance_tests_common.yml
```diff
@@ -34,101 +37,115 @@ jobs:
...
+      - name: Parse recommended tests coming from the input as XML
+        if: ${{ inputs.recommended_tests != '' }}
+        run: python .github/scripts/get_tests.py '${{ inputs.recommended_tests }}' > testsuite/run_sets/filter.yml
+      - name: Generate recommended tests as YAML
+        if: ${{ inputs.recommended_tests != '' }}
+        run: |
+          ./testsuite/podman_runner/19_generate_recommended_tests_yml.sh
+          TESTS_COUNT=$(wc -l < /testsuite/run_sets/recommended_tests.yml)
+          if [ "$ROW_COUNT" -eq 0 ] || [ "$ROW_COUNT" -gt 30 ]; then
+            echo "⚠️ WARNING: $ROW_COUNT recommended tests found. Running the full test suite instead."
```
jordimassaguerpla, line 92: I would change "Running the full test suite instead" by "Look at the results of the full test suite, instead".

jordimassaguerpla: Other than this minor issue. LTGM

jordimassaguerpla (approved):

Issue #9613, opened 2025-01-10
Dot in Installed-Size when syncing RaspberryPi OS 12 packages


basdakkenhorst:
# Question

Allthough, not (or at least not yet) officially supported, I use Uyuni to sync Raspberry Pi OS 12 packages. I therefor make this a question, instead of a bug report.

Since at least a couple of months this fails with the following error:

```
2025/01/10 13:11:14 -00:00 Sync of channel started.
2025/01/10 13:11:14 -00:00 Repo URL: https://archive.raspberrypi.org/debian/dists/bookworm/main/binary-arm64
.....
2025/01/10 13:12:20 -00:00 Error during processing package wolframscript-14.1.0+20240822262-X:.arm64-deb.
invalid literal for int() with base 10: '6635.2451171875'
2025/01/10 13:12:20 -00:00 Exception: invalid literal for int() with base 10: '6635.2451171875'
2025/01/10 13:12:20 -00:00   Package batch #7 of 35 completed...
2025/01/10 13:12:26 -00:00 Error during processing package wolfram-engine-14.1.0+202408191410-X:.arm64-deb.
invalid literal for int() with base 10: '4341190.494140625'
2025/01/10 13:12:26 -00:00 Exception: invalid literal for int() with base 10: '4341190.494140625'
```

I found out that has something todo with a dot in the Installed-Size value of both packages.

In: http://archive.raspberrypi.org/debian/dists/bookworm/main/binary-arm64/Packages both packages are defined as follows:

```
Package: wolfram-engine
Version: 14.1.0+202408191410
Architecture: arm64
Installed-Size: 4341190.494140625
.....

Package: wolframscript
Version: 14.1.0+20240822262
Architecture: arm64
Installed-Size: 6635.2451171875
```

## Workaround

I added the following to ```python/spacewalk/server/importlib/debPackage.py```

```
import re
....
# pylint: disable-next=missing-class-docstring
class debBinaryPackage(headerSource.rpmBinaryPackage):
    # pylint: disable-next=dangerous-default-value
    def __init__(
        self, header, size, checksum_type, checksum, path=None, org_id=None, channels=[]
    ):
    .....
        # Fix some of the information up
        vendor = self["vendor"]
        if vendor is None:
            self["vendor"] = "Debian"
        payloadFormat = self["payload_format"]
        if payloadFormat is None:
            self["payload_format"] = "ar"
        if self["payload_size"] is None:
            self["payload_size"] = 0
        #
[...]

PR #9726, opened 2025-02-04
Enable packages metadata to be float when int is expected (gh#uyuni-project/uyuni#9613)

m-czernek:
## What does this PR change?

When we process package metadata, packages can mistakenly have float where we expect integer, for example `Installed Size`, `Size`, etc.

Though this is technically an invalid package, Uyuni can round such values to integer instead of throwing an exception.

In this PR, if a package contains a float value in metadata where we expect integer, we first round the value to the nearest integer.

In case where such value is neither float nor int, I improve the error that is displayed to the user.

## Documentation

## Links

Issue(s): https://github.com/uyuni-project/uyuni/issues/9613
Port(s): TODO

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/server/importlib/backendLib.py
+++ b/python/spacewalk/server/importlib/backendLib.py
@@ -494,15 +494,7 @@ class TableInsert(TableUpdate):
 
 def sanitizeValue(value, datatype):
     if isinstance(datatype, DBstring):
-        if value is None or value == "":
-            return None  # we really want to preserve Nones
-            # and not depend on Oracle converting
-            # empty strings to NULLs -- PostgreSQL
-            # does not do this
-        if len(value) > datatype.limit:
-            value = value[: datatype.limit]
-            # ignore incomplete characters created after truncating
-        return value
+        return _sanitize_dbstring(value, datatype)
     if isinstance(datatype, DBblob):
         if value is None:
             value = ""
@@ -518,7 +510,36 @@ def sanitizeValue(value, datatype):
     if isinstance(datatype, DBdate):
         return str(value)[:10]
     if isinstance(datatype, DBint):
-        return int(value)
+        try:
+            value = _sanitize_number_string(value)
+            return int(float(value))
+        except ValueError as e:
+            raise ValueError(f"Cannot convert {value} to int") from e
+    return value
+
+
+def _sanitize_number_string(value):
+    """
+    Enable number-like strings to be converted by
+    int() or float()
+    """
+    if not isinstance(value, str):
+        return value
+
+    # Allow "1 234"
+    value = value.replace(" ", "")
+    return value
+
+
+def _sanitize_dbstring(value, datatype):
+    if value is None or value == "":
+        return None  # we really want to preserve Nones
+        # and not depend on Oracle converting
+        # empty strings to NULLs -- PostgreSQL
+        # does not do this
+    if len(value) > datatype.limit:
+        value = value[: datatype.limit]
+        # ignore incomplete characters created after truncating
     return value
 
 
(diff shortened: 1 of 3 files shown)


python/spacewalk/server/importlib/backendLib.py
```diff
@@ -494,15 +494,7 @@ def query(self, values):
...
-        if value is None or value == "":
-            return None  # we really want to preserve Nones
-            # and not depend on Oracle converting
-            # empty strings to NULLs -- PostgreSQL
-            # does not do this
-        if len(value) > datatype.limit:
-            value = value[: datatype.limit]
-            # ignore incomplete characters created after truncating
-        return value
+        return _sanitize_dbstring(value, datatype)
```
m-czernek, line 497: I refactored this into a separate function so that sonarcloud won't fail with method complexity being too high.

agraul (commented):
Do we have any examples we can use for testing this change?

agraul:
Thanks for the pointer, I'll try to sync this to see if we can solve the issue higher up in the stack. Just to have a second approach we can compare to this one.

agraul:
The alternative that immediately came to my mind was to fix this at the outside layer instead of deep down close to the DB insertion. The downside is that this approach only works for known keys, instead of anything. But casting anything to float then int is maybe not needed either.

```diff
python/spacewalk/server/importlib/debPackage.py | 5 +++++

modified   python/spacewalk/server/importlib/debPackage.py
@@ -59,6 +59,11 @@ class debBinaryPackage(headerSource.rpmBinaryPackage):
             value = header[field]
             if key == "build_time" and isinstance(value, int):
                 value = gmtime(value)  # unix timestamp
+            elif key == "payload_size" and value:
+                # Payload size / Installed-Size is an int, but some packages have
+                # bad metadata (float). Since Installed-Size is an estimation, truncating to int is ok
+                # https://www.debian.org/doc/debian-policy/ch-controlfields.html#installed-size
+                value = int(float(value))
             elif value == []:
                 value = None
             elif value:
```

I'm not sure why we currently keep every value as a string until it hits the DB and only casts it to the expected DB type. Maybe your approach fits better with the current approach.

agraul (dismissed):
LGTM!

python/spacewalk/server/importlib/backendLib.py
```diff
@@ -518,7 +510,38 @@ def sanitizeValue(value, datatype):
...
+    Enable number-like strings to be converted by
+    int() or float()
+    """
+    if not isinstance(value, str):
+        return value
+
+    # Allow "1 234"
+    value = value.replace(" ", "")
+    # 1,234 cannot be processed by float()
+    value = value.replace(",", ".")
```
agraul, line 532: This one is dangerous, different countries assign different meaning to commas in numbers. `1,234` US is `1234`, but `1,234` in Germany is `1.234` US.

agraul (changes requested):
I don't see a problem with stripping spaces from number-strings :+1: 

Replacing a comma in a number-string needs to be locale aware, specifically aware of the locale the package metadata is in. It might be too hard to fix that at this point in the import process, I doubt we have access to that information where this code is executed. Maybe it's easier to reject such packages for now.

python/spacewalk/server/importlib/backendLib.py
m-czernek: Yes, but `1.234` is already handled by `float()` and `float()`, afaik, never accepts `1,234` regardless of the local settings?

agraul (approved):

PR #9763, opened 2025-02-12
uyuni-proxy-tftpd: fix issue replacing FQDNs for proxies (bsc#1236601)

meaksh:
## What does this PR change?

This PR does the following things:

### fix issue replacing FQDNs on `uyuni-proxy-tftpd` container

The current "replacing" mechanism was causing collitions, i.a. when server FQDN is `foobar.example.domain` and there are other references to `aafoobar.example.domain`. The current code will wrongly replace both cases.

This PR fixes the match/replace code to detect avoid this situation.

### fix wrong IP set on `susemanager-tftpsync-recv.conf` file
The apache2 configuration that allows files to be received from the upstream server need to contain the IP address of the server, and not from the proxy.

~~### drop code and references to `susemanager-tftpsync` and `susemanager-tftpsync-recv`.
These packages are not used anymore on containerized environment. The replacement is the `uyuni-proxy-tftpd` container.~~ we cannot do this until older proxy versions are deprecated.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **this should be tested in our acceptance tests**

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/26312

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


m-czernek (dismissed):
+1, thanks for additionally fixing the style/linting on the file!

agraul (dismissed):

tftpsync/susemanager-tftpsync-recv/add.wsgi
```diff
@@ -16,74 +17,93 @@ import os
...
 ch.setFormatter(formatter)
 
 # add ch to logger
 logger.addHandler(ch)
 
+
 class TftpFieldStorage(cgi.FieldStorage):
 
-    def make_file(self, binary=None):
+    def make_file(self, **_):
```
agraul, line 56: I'd prefer to stick to our style guide and handle unused arguments explicitly (as explained at the bottom of [2.1](https://google.github.io/styleguide/pyguide.html#21-lint)

agraul (commented):
Using `re.sub()` with `\b` makes sense :+1: 

BTW, it's a bit hard to review PRs that reformat files and change logic within them. Personally I'd prefer it if we kept these files as they are and reformat them at once. That way we can also add the reformat git commit to `.git-blame-ignore-revs`

tftpsync/susemanager-tftpsync-recv/add.wsgi
m-czernek: So that means
```suggestion
    def make_file(self, **kwargs):
        del kwargs
```

m-czernek (approved):
From my POV, this looks good to me. I would suggest that you squash all commits into 2 commits:

- 1st commit for functional changes
- 2nd commit for formatting only

That way, we can add the formatting commit to the `git-blame-ignore-revs` file as Alex suggests, and at the same time, we can do both in one PR.

This will dismiss my +1, but to me, that is OK and if those are the only changes, you can still take my dismissed review as +1.

I'm surprised that we first deleted part of the code and then put it back in, but I trust that this is correct and we cannot remove those parts just yet. Let's just not add commits like "remove X" and then "revert remove X" into `master` please :)

tftpsync/susemanager-tftpsync-recv/add.wsgi
meaksh: This should be done, now.

agraul (approved):
LGTM  :rocket:, just make sure to merge with a merge commit to keep the git sha1 (assuming squash & merge / rebase & merge options are still available)

PR #9423, opened 2024-10-31
25124 enforcing password requirements

Serp1co:
## What does this PR change?

Enforce password requirement for server users, comes with a new admin panel to manage local users password policy.

## GUI diff

Before:
uyuni-beofre

After:
uyuni-after
image

- [x] **DONE**

## Documentation
- Documentation issue was created: [Link for SUSE Manager contributors](https://github.com/SUSE/spacewalk/issues/new?template=ISSUE_TEMPLATE_DOCUMENTATION.md&labels=documentation&projects=SUSE/spacewalk/31), [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/new).
- API documentation added: please review the Wiki page [Writing Documentation for the API](https://github.com/uyuni-project/uyuni/wiki/Writing-documentation-for-the-API) if you have any changes to API documentation.
- (OPTIONAL) [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pulls)

- [ ] **DONE**

## Test coverage
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25124
Port(s): # **add downstream PR(s), if any**

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [x] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql" (Test skipped, there are no changes to test)
- [ ] Re-run test "java_pgsql_tests"        
- [ ] Re-run test "schema_migration_test_pgsql"      
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests" (Test skipped, there are no changes to test)

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

PR #9769, opened 2025-02-13
Implement module to get configuration values from the database

mcalmer:
## What does this PR change?

We started to move configuration values from rhn.conf into the database rhnConfiguration table.
As we need now some of the moved values in python code, we need a function which get us these
from the DB.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [ ] **DONE**

## Links

Addition to: https://github.com/uyuni-project/uyuni/pull/9423

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/server/db_config.py
+++ b/python/spacewalk/server/db_config.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025 SUSE LLC
+#
+# This software is licensed to you under the GNU General Public License,
+# version 2 (GPLv2). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+# along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+#
+
+"""A module that reads values from the database"""
+
+
+from typing import Optional, Union
+from spacewalk.server import rhnSQL
+
+
+def sanitize_value(key: str, val: str) -> Optional[Union[str, float, int]]:
+    """
+    attempt to convert a string value to the proper type
+    """
+    convert_table = {
+        "PSW_CHECK_SPECIAL_CHARACTERS": str,
+    }
+    val = val.strip()
+
+    if key in convert_table:
+        try:
+            val = convert_table[key](val)
+        except ValueError:
+            pass
+    else:
+        try:
+            val = int(val)  # make int if can.
+        except ValueError:
+            try:
+                val = float(val)  # make float if can.
+            except ValueError:
+                pass
+    if val == "":  # Empty strings treated as None
+        val = None
+    return val
+
+
+def value(key: str) -> Optional[Union[str, float, int]]:
+    """
+    Return the value of the given key. If the value is not defined, return the default value
+    When the key is not found, an AttributeError is raised
+    """
+    rhnSQL.initDB()
+
+    h = rhnSQL.prepare(
+        """
+            SELECT value, default_value
+            FROM rhnConfiguration
+            WHERE key = :key
+            """
+    )
+    h.execute(key=key)
+    data = h.fetchone_dict()
+    if not data:
+        raise AttributeError(key)
+
+    result = data["value"] if data["value"] else data["default_value"]
+    return sanitize_value(key, result)
--- a/python/spacewalk/satellite_tools/accounts/satpasswd
+++ b/python/spacewalk/satellite_tools/accounts/satpasswd
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 #
 # Copyright (c) 2014--2015 Red Hat, Inc.
+# Copyright (c) 2025 SUSE LLC
 #
 # This software is licensed to you under the GNU General Public License,
 # version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -16,7 +17,7 @@
 import getpass
 import sys
 
-from spacewalk.common.rhnConfig import initCFG, CFG
+from spacewalk.server import db_config
 from spacewalk.server import rhnSQL, rhnUser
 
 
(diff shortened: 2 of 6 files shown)


Serp1co (commented):
lgtm

Serp1co (approved):

python/spacewalk/server/dbConfig.py
```diff
@@ -0,0 +1,64 @@
...
+        attempt to convert a string value to the proper type
+        """
+        convert_table = {
+            "PSW_CHECK_SPECIAL_CHARACTERS": str,
+        }
+        val = val.strip()
+
+        if convert_table.get(key):
+            try:
+                val = convert_table.get(key)(val)
```
agraul, line 34: This will raise a TypeError when `dict.get()` returns it's default value `None`.

python/spacewalk/server/dbConfig.py
```diff
@@ -0,0 +1,64 @@
...
+from . import rhnSQL
+
+
+def value(key):
+    """
+    Return the value of the given key. If the value is not defined, return the default value
+    When the key is not found, an AttributeError is raised
+    """
+
+    def sanitize_value(key, val):
```
agraul, line 23: If we move this out of the `value()` function, it's easier to test this function.

python/spacewalk/server/dbConfig.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# This software is licensed to you under the GNU General Public License,
+# version 2 (GPLv2). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+# along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+#
+
+from . import rhnSQL
```
agraul, line 14: We prefer not to use relative imports ([style guide](https://google.github.io/styleguide/pyguide.html#22-imports))

agraul (changes requested):

python/spacewalk/server/dbConfig.py
```diff
@@ -0,0 +1,64 @@
...
+            "PSW_CHECK_SPECIAL_CHARACTERS": str,
+        }
+        val = val.strip()
+
+        if convert_table.get(key):
+            try:
+                val = convert_table.get(key)(val)
+            except ValueError:
+                pass
+        else:
```
m-czernek, line 37: So we want:

- For specific values, we want specific type defined in the inner dict
- Otherwise, we want an int,
- Unless it's a float.
- But default return from a DB is string (right?), so if it is neither an int or a float, it'll be a string anyways
- Finally, if it is an empty string, return None

Is that accurate? (I'm asking because this inner method seems confusing to me, and like we should be able to make it simpler than if/else-with-try-except-try-except)

m-czernek (commented):
A few comments while just passing by.

python/spacewalk/server/dbConfig.py
meaksh: I think we should be good here, as we are inside the `if convert_table.get(key)` statement, so value shouldn't be `None`.

meaksh (commented):
Besides of the comments made by Alex, this looks good to me.

python/spacewalk/server/dbConfig.py
mcalmer: Well, this . was introduced at python3 migration time by somebody of the python squad.
I think it has a reason. But I will try if the full import works

python/spacewalk/server/dbConfig.py
mcalmer: yes, we try to convert to the real type int, float or string.
There may special keys which always must be converted to a specified type.
Example: there is a password and you decide to use "1234" as password. We want it to be a string, but that mechanism would convert it to a int if we would not have the convert_table.

agraul (approved):

python/spacewalk/server/db_config.py
```diff
@@ -0,0 +1,66 @@
...
+#
+# Copyright (c) 2025 SUSE LLC
+#
+# This software is licensed to you under the GNU General Public License,
+# version 2 (GPLv2). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+# along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+#
```
meaksh, line 13: According to https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations, the "encoding definition" should be placed in the first or second line.

I would say:
```suggestion
# -*- coding: utf-8 -*-
#
# Copyright (c) 2025 SUSE LLC
#
# This software is licensed to you under the GNU General Public License,
# version 2 (GPLv2). There is NO WARRANTY for this software, express or
# implied, including the implied warranties of MERCHANTABILITY or FITNESS
# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
# along with this software; if not, see
# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
#

"""A module that reads values from the database"""

```

meaksh (approved):

python/spacewalk/server/db_config.py
agraul: utf-8 is the default anyway

python/spacewalk/server/db_config.py
```diff
@@ -0,0 +1,67 @@
...
+from typing import Optional, Union
+from spacewalk.server import rhnSQL
+
+
+def sanitize_value(key: str, val: str) -> Optional[Union[str, float, int]]:
+    """
+    attempt to convert a string value to the proper type
+    """
+    convert_table = {
+        "PSW_CHECK_SPECIAL_CHARACTERS": str,
```
rjmateus, line 25: instead of having this in the code, should we add an extra collum to the table to specify who the value should be interpreted?
Like int, string, etc?

Serp1co: Speaking for the Java side:
I don't think it's necessary, we have to know the data type beforehand, no dynamic type parsing.

PR #9806, opened 2025-02-19
QE: Handle 5.1 when resolving product_version

srbarrios:
## What does this PR change?

This PR includes 5.1 as a valid product version to be included in the group of versions using salt bundle, instead of classic salt. Also fix the match for Head versions, so when Head will become the next 5.2 version, we don't really need to include 5.2, until it's released ;)

I added a TODO, as we ideally should refactor this code to only use Salt Bundle for now on, as all our current supported versions are using Salt Bundle.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- Cucumber tests were fixed

- [x] **DONE**

## Links

Ports:
- Manager-4.3
- Manager-5.0

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -60,7 +60,7 @@ def product_version
   raise NotImplementedError, 'Could not determine product version'
 end
 
-# Retrieves the full product version using the 'salt-call' command.
+# Retrieves the full product version using the 'venv-salt-call' command.
 #
 # @return [String, nil] The full product version if the command execution was successful and
 #   the output is not empty, otherwise nil.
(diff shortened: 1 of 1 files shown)


testsuite/features/support/commonlib.rb
```diff
@@ -70,13 +70,15 @@ def product_version_full
...
+#       Refactoring all the code call it.
 # Determines whether to use the Salt bundle based on the product and product version.
 #
 # @return [Boolean] true if the product is 'Uyuni' or the product version is 'head', '5.0', '4.3', or '4.2'
 # - false otherwise
 def use_salt_bundle
-  # Use venv-salt-minion in Uyuni, or SUMA Head, 5.0, 4.2 and 4.3
-  product == 'Uyuni' || %w[head 5.0 4.3 4.2].include?(product_version)
+  # Use venv-salt-minion in Uyuni, or SUMA Head, 5.1, 5.0, 4.2 and 4.3
+  product == 'Uyuni' || %w[develHead 5.1 5.0 4.3 4.2].include?(product_version)
```
nodeg, line 81: Should we not also keep `head` there?

srbarrios: Nope, `head` was wrong. It was never matching `head`.
See an example of what we have on head:
```
uyuni-server:/ # rpm -q patterns-suma_server
patterns-suma_server-5.1-150750.2.2.develHead.x86_64
```

srbarrios: I could also apply lowercase to the product_version, but I preferred that way.

NamelessOne91 (approved):

testsuite/features/support/commonlib.rb
ktsamis: I would add only 5.1 here because we use the include function which would match either way. But no objection to add both really

ktsamis (approved):

testsuite/features/support/commonlib.rb
nodeg: Taking a lot at how we get the product version, I understand it now and you are correct, Oscar.
```ruby
# @return [String] The version number of the product being tested.
def product_version
(...)
  product_raw, code = get_target('server').run('rpm -q patterns-suma_server', check_errors: false)
  m = product_raw.match(/patterns-suma_server-(.*)-.*/)
  return m[1] if code.zero? && !m.nil?
(...)
end
```

nodeg (approved):

PR #9802, opened 2025-02-18
Add RBAC schema implementation

wweellddeerr:
## What does this PR change?

It adds the necessary code for RBAC schema, including changes to handle separate database schema usage.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes

- [x] **DONE**

## Test coverage
ℹ️ If a major new functionality is added, it is **strongly recommended** that tests for the new functionality are added to the Cucumber test suite
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/26409

- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

PR #9829, opened 2025-02-21
Add RBAC schema implementation

wweellddeerr:
## What does this PR change?

It is a port of https://github.com/uyuni-project/uyuni/pull/9802 containing the schema migration and the ORM implementation.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Issue(s):
Port(s): 

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/access/WebEndpoint.java
+++ b/java/code/src/com/redhat/rhn/domain/access/WebEndpoint.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) 2025 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package com.redhat.rhn.domain.access;
+
+import com.redhat.rhn.domain.BaseDomainHelper;
+
+import java.util.Objects;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.NamedNativeQuery;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "endpoint", schema = "access")
+@NamedNativeQuery(name = "WebEndpoint_user_access_endpoint_scope",
+        // Select an accessible endpoint by path, http method (GET, POST, etc.) and scope (Web UI or API)
+        query = "SELECT * FROM access.userAccessTable e WHERE user_id = :user_id AND e.endpoint = :endpoint " +
+                "AND e.http_method = :http_method AND e.scope = :scope LIMIT 1"
+)
+@NamedNativeQuery(name = "WebEndpoint_user_access_endpoint",
+        // Select an accessible endpoint by path, http method (GET, POST, etc.)
+        query = "SELECT * FROM access.userAccessTable e WHERE user_id = :user_id AND e.endpoint = :endpoint " +
+                "AND e.http_method = :http_method LIMIT 1"
+)
+@NamedNativeQuery(name = "WebEndpoint_user_access_class_method",
+        // Select an accessible endpoint by its handler class and method
+        query = "SELECT * FROM access.userAccessTable e WHERE user_id = :user_id AND e.scope = :scope " +
+                "AND e.class_method = :class_method LIMIT 1"
+)
+@NamedNativeQuery(name = "WebEndpoint_get_unauthenticated_webui",
+        // Get unauthenticated Web UI endpoints
+        query = "SELECT endpoint FROM access.endpoint WHERE authorized = false"
+)
+@NamedNativeQuery(name = "WebEndpoint_get_unauthenticated_api",
+        // Get unauthenticated API endpoints
+        query = "SELECT class_method FROM access.endpoint WHERE scope = 'A' AND authorized = false"
+)
+public class WebEndpoint extends BaseDomainHelper {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @Column(name = "class_method")
+    private String className;
+    private String endpoint;
+    private String httpMethod;
+    @Enumerated(EnumType.STRING)
+    private Scope scope;
+    private Boolean authorized;
+
+    public enum Scope {
+        A("API"),
+        W("Web application");
+
+        private final String label;
+
+        Scope(String labelIn) {
+            this.label = labelIn;
+        }
+
+        public String getLabel() {
+            return label;
+        }
+    }
+
+    /**
+     * Default constructor for WebEndpoint.
+     * Initializes an empty instance of WebEndpoint.
+     */
+    public WebEndpoint() {
+    }
+
+    /**
+     * Constructs a new WebEndpoint with the specified values.
+     *
+     * @param classNameIn the name of the class associated with the endpoint
+     * @param endpointIn the endpoint URL or path
+     * @param httpMethodIn the HTTP method (e.g., GET, POST)
+     * @param scopeIn the scope that defines where the endpoint is used (API or Web application)
+     * @param authorizedIn whether the endpoint is authorized (true) or not (false)
+     */
+    public WebEndpoint(String classNameIn, String endpointIn, String httpMethodIn, Scope scopeIn,
+                       Boolean authorizedIn) {
+        className = classNameIn;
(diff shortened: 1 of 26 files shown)


schema/spacewalk/common/views/userAccessTable.sql
```diff
@@ -0,0 +1,33 @@
...
+-- implied, including the implied warranties of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+--
+
+-- userAccessTable view
+-- User access rules on endpoints either permitted directly or through an access group
+CREATE VIEW access.userAccessTable AS
+WITH endpoints AS (
```
rjmateus, line 15: I'm not sure if we need this initial query. We should test it in terms of performance when we have all the endpoints and see if it has any performance impact.

schema/spacewalk/common/tables/namespace.sql
```diff
@@ -0,0 +1,19 @@
...
+-- version 2 (GPLv2). There is NO WARRANTY for this software, express or
+-- implied, including the implied warranties of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+--
+
+CREATE TABLE access.namespace (
+    id          BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    namespace   VARCHAR NOT NULL,
```
rjmateus, line 14: This should be unique, to avoid collisions.

schema/spacewalk/common/tables/endpoint.sql
```diff
@@ -0,0 +1,26 @@
...
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+--
+
+CREATE TABLE access.endpoint (
+    id              BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    class_method    VARCHAR NOT NULL,
+    endpoint        VARCHAR NOT NULL,
+    http_method     VARCHAR NOT NULL,
+    scope           CHAR(1) NOT NULL
```
rjmateus, line 17: Scope was moved from namespace to endpoint, when compared to the RFC. Could you elaborate on this option? Should it also be used to namespace, to make it clear when a user want to assign XML-RPC API (and json over HTTP) access only, and which methods should be shown?

schema/spacewalk/common/tables/endpoint.sql
```diff
@@ -0,0 +1,26 @@
...
+                        CHECK (scope in ('A', 'W')),
+    authorized      BOOLEAN NOT NULL DEFAULT true,
```
rjmateus, line 19: I would make it with a more explicit name, like auth_required, as proposed in the RFC.

schema/spacewalk/common/tables/namespace.sql
```diff
@@ -0,0 +1,19 @@
...
+--
+-- This software is licensed to you under the GNU General Public License,
+-- version 2 (GPLv2). There is NO WARRANTY for this software, express or
+-- implied, including the implied warranties of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+--
+
+CREATE TABLE access.namespace (
```
rjmateus, line 12: As stated in another comment, table doesn't have the space as defined in the RFC. Should we have it?

rjmateus (commented):
Just a few notes, but overall looks good.

schema/spacewalk/common/tables/endpoint.sql
rjmateus: Then we need to be disciplined to make sure we don't set a namespace starting with API for stuff that only focus on the webUI.
It would also allow to mix webUI features with API (XML-RPC/JSON over HTTP) call, which is a good thing in my opinion.

schema/spacewalk/common/views/userAccessTable.sql
cbbayburt: This one used to be for my own convenience, but I changed this view completely with the latest commit. This view is now queried from the Java filters and is an integral part of the feature, so its performance is critical and I optimized it with proper indexes and planning.

So far I tested it with what data I have, and it's running in microseconds (with about half of the predicted endpoint data).

java/code/src/com/redhat/rhn/domain/access/WebEndpoint.java
```diff
@@ -0,0 +1,176 @@
...
+                "AND e.http_method = :http_method LIMIT 1"
+)
+@NamedNativeQuery(name = "WebEndpoint_user_access_class_method",
+        // Select an accessible endpoint by its handler class and method
+        query = "SELECT * FROM access.userAccessTable e WHERE user_id = :user_id AND e.scope = :scope " +
+                "AND e.class_method = :class_method LIMIT 1"
+)
+@NamedNativeQuery(name = "WebEndpoint_get_unauthenticated_webui",
+        // Get unauthenticated Web UI endpoints
+        query = "SELECT endpoint FROM access.endpoint WHERE auth_required = false"
```
rjmateus, line 51: The next query is filtering by API, this one is only filtering by auth_required false. Do we need the two methods? Is this needed because the default handler for the XML-RPC API?

java/code/src/com/redhat/rhn/domain/access/WebEndpointFactory.java
```diff
@@ -0,0 +1,108 @@
...
+                                                                       WebEndpoint.Scope scope) {
+        NativeQuery<WebEndpoint> query = getSession().getNamedNativeQuery("WebEndpoint_user_access_class_method");
+        query.setParameter("user_id", userId);
+        query.setParameter("class_method", classMethod);
+        query.setParameter("scope", scope.name());
+        return query.uniqueResultOptional();
+    }
+
+    /**
+     * Get all unauthorized web UI endpoints
```
rjmateus, line 90:
```suggestion
     * Get all web UI endpoints that don't require authorization check
```

java/code/src/com/redhat/rhn/domain/access/WebEndpointFactory.java
```diff
@@ -0,0 +1,108 @@
...
+     * @return the set of unauthorized web endpoints
+     */
+    public static Set<String> getUnauthorizedWebEndpoints() {
+        // TODO: Cache
+        NativeQuery<String> query = getSession().getNamedNativeQuery("WebEndpoint_get_unauthenticated_webui");
+        return query.getResultStream().collect(Collectors.toUnmodifiableSet());
+    }
+
+    /**
+     * Get all unauthorized API methods
```
rjmateus, line 100:
```suggestion
     * Get all API methods that don't require authorization check
```

java/code/src/com/redhat/rhn/domain/access/WebEndpointFactory.java
```diff
@@ -0,0 +1,108 @@
...
+     * @return the set of unauthorized API methods as a string of qualified class name & method
```
rjmateus, line 101:
```suggestion
     * @return the set of API methods, that doesn't need authentication, as a string of qualified class name & method
```

java/code/src/com/redhat/rhn/domain/access/WebEndpointFactory.java
```diff
@@ -0,0 +1,108 @@
...
+        NativeQuery<WebEndpoint> query = getSession().getNamedNativeQuery("WebEndpoint_user_access_class_method");
+        query.setParameter("user_id", userId);
+        query.setParameter("class_method", classMethod);
+        query.setParameter("scope", scope.name());
+        return query.uniqueResultOptional();
+    }
+
+    /**
+     * Get all unauthorized web UI endpoints
+     * @return the set of unauthorized web endpoints
```
rjmateus, line 91:
```suggestion
     * @return the set of web endpoints that doesn't need authorization check
```

rjmateus (approved):
Some small nitpick, overall looks good to me.

schema/spacewalk/common/tables/endpoint.sql
cbbayburt: Sure, this is still up for debate. Hopefully we'll get some feedback until the GM and can decide accordingly.

java/code/src/com/redhat/rhn/domain/access/WebEndpoint.java
cbbayburt: This one is actually used with HTTP API routes as well (works the same way as a web UI endpoint). I'll fix the naming to reflect that.

java/code/src/com/redhat/rhn/domain/access/WebEndpointFactory.java
cbbayburt: I had a long argument with ChatGPT about this :rofl: 

Apparently, "unauthorized" means both "not authorized" and "doesn't require authorization"

But since you are also confused, I think it's better to change it like this

rjmateus: I not a native speaker, and even in my main language I not really good with "wording" so I would not argue on this regard :)

PR #9783, opened 2025-02-14
ISSv3 UI: Access token management and Peripheral registration

mackdk:
## What does this PR change?

This PR adds UI for managing access token and registering peripheral server. In particular, the UI allows:

- List the current access tokens, change the validity status and delete them if not currently associated to a peripheral
- Issue a new token, granting access to a remote server
- Storing an external token
- Registering a remote server as peripheral

Note: this are currently single pages, that  need to be inserted within the full Hub ISS UI. Currently, yo reach them you have manually point your browser to `https://uyuni-server.dev.local/rhn/manager/admin/hub/access-tokens` and `https://uyuni-server.dev.local/rhn/manager/admin/hub/peripherals/register`

#### Access tokens
image
image
image
image
image
image
image

#### Register peripheral
image
image

- [X] **DONE**

## Documentation
- Documentation WIP

- [X] **DONE**

## Test coverage
ℹ️ If a major new functionality is added, it is **strongly recommended** that tests for the new functionality are added to the Cucumber test suite
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25519 https://github.com/SUSE/spacewalk/issues/25518

- [X] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

PR #9830, opened 2025-02-22
ISSv3 - UI skeleton and Hub details page

mackdk:
## What does this PR change?

This PR adds the menu and  the navigation to make the [previously made pages](https://github.com/uyuni-project/uyuni/pull/9783) accessible and adds the hub details page

NOTE: all the pages and the features around peripherals are included only as mock, they will be part of an additional PR.

## GUI diff

After:
image
image
image
image
image
image
image

- [X] **DONE**

## Documentation
- Documentation WIP

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25355

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/suse/manager/webui/controllers/admin/AdminViewsController.java
+++ b/java/code/src/com/suse/manager/webui/controllers/admin/AdminViewsController.java
@@ -29,12 +29,15 @@ import com.redhat.rhn.taskomatic.TaskomaticApi;
 import com.suse.manager.admin.PaygAdminManager;
 import com.suse.manager.model.hub.HubFactory;
 import com.suse.manager.reactor.utils.OptionalTypeAdapterFactory;
+import com.suse.manager.webui.controllers.admin.beans.PeripheralResponse;
 import com.suse.manager.webui.controllers.admin.mappers.PaygResponseMappers;
 import com.suse.manager.webui.utils.FlashScopeHelper;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.reflect.TypeToken;
 
+import java.lang.reflect.Type;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -57,6 +60,7 @@ public class AdminViewsController {
 
     private static final PaygAdminManager PAYG_ADMIN_MANAGER = new PaygAdminManager(new TaskomaticApi());
 
+    private static final HubFactory HUB_FACTORY = new HubFactory();
 
     private AdminViewsController() { }
 
@@ -66,7 +70,7 @@ public class AdminViewsController {
      */
     public static void initRoutes(JadeTemplateEngine jade) {
         get("/manager/admin/config/monitoring",
-                withUserPreferences(withCsrfToken(withOrgAdmin(AdminViewsController::showMonitoring))), jade);
+            withUserPreferences(withCsrfToken(withOrgAdmin(AdminViewsController::showMonitoring))), jade);
         get("/manager/admin/config/password-policy",
                 withUserPreferences(withCsrfToken(withOrgAdmin(AdminViewsController::showPasswordPolicy))), jade);
         get("/manager/admin/setup/payg",
@@ -77,6 +81,11 @@ public class AdminViewsController {
                 withUserPreferences(withCsrfToken(withOrgAdmin(AdminViewsController::showPayg))), jade);
         get("/manager/admin/setup/proxy",
                 withUserPreferences(withCsrfToken(withOrgAdmin(AdminViewsController::showProxy))), jade);
+
+        get("/manager/admin/hub/hub-details",
+            withUserPreferences(withCsrfToken(withOrgAdmin(AdminViewsController::showISSv3Hub))), jade);
+        get("/manager/admin/hub/peripherals",
+            withUserPreferences(withCsrfToken(withOrgAdmin(AdminViewsController::showISSv3Peripherals))), jade);
         get("/manager/admin/hub/peripherals/register",
             withUserPreferences(withCsrfToken(withProductAdmin(AdminViewsController::registerPeripheral))), jade);
         get("/manager/admin/hub/access-tokens",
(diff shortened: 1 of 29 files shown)


java/code/src/com/suse/manager/webui/controllers/admin/AdminViewsController.java
```diff
@@ -93,7 +108,7 @@ public static void initRoutes(JadeTemplateEngine jade) {
     public static ModelAndView showMonitoring(Request request, Response response, User user) {
         Map<String, Object> data = new HashMap<>();
         data.put("isUyuni", ConfigDefaults.get().isUyuni());
-        return new ModelAndView(data, "controllers/admin/templates/monitoring.jade");
+        return new ModelAndView(data, "controllers/admin/templates/issv3/monitoring.jade");
```
mcalmer, line 111: maybe better use "hub" instead of "issv3" ?

mcalmer (approved):
I cannot say much about typescript, but the java code is ok - just one note.
But you should check SonarCloud findings

java/code/src/com/suse/manager/webui/controllers/admin/AdminViewsController.java
mackdk: Mmm, let me double check this I don't think we should have changed it.

mackdk:
> But you should check SonarCloud findings

Those are all related to the "fake" peripheral part. I think I'll probably just remove it, so sonarcloud won't complain.

web/html/src/components/datetime/DateTime.tsx
```diff
@@ -3,7 +3,7 @@ import { localizedMoment } from "utils";
 import { getTimeProps } from "./timeProps";
 
 type Props = {
-  value?: string | moment.Moment;
+  value?: string | moment.Moment | Date;
```
Etheryte, line 6: We would very much like to avoid this, if at all possible. The reason is timezones, when we start passing around native Date objects we will very quickly struggle to make any sense of what the related offsets should be.  

I couldn't find the diff what we need this for, perhaps I just missed it?

web/html/src/components/hub/DeregisterServer.tsx
```diff
@@ -0,0 +1,97 @@
...
+        : `/rhn/manager/api/admin/hub/peripherals/${this.props.id}`;
+    Network.del(resource)
+      .then(
+        (_response) => {
+          showInfoToastr(t("The server {fqdn} has been successfully deregistered.", { fqdn: this.props.fqdn }));
+
+          // Invoke the callback if present
+          if (this.props.onDeregistered !== undefined) {
+            this.props.onDeregistered();
+          }
```
Etheryte, line 91:
```suggestion
          this.props.onDeregistered?.();
```

web/html/src/components/datetime/DateTime.tsx
mackdk: Ok, no problem, I change it because I saw `<FromNow />` declares it in this way, I thought it was just forgotten :sweat_smile:

web/html/src/components/large-text-attachment.tsx
```diff
@@ -0,0 +1,251 @@
...
+ */
+export class LargeTextAttachment extends React.Component<Props, State> {
+  private static readonly INITAL_MODEL: EditFormModel = { method: EditMethod.Upload };
+
+  static defaultProps: Partial<Props> = {
+    filename: "attachment.txt",
+    editable: false,
+    hideMessage: false,
+    presentMessage: t("Data is present"),
+    absentMessage: t("No data available."),
```
Etheryte, line 63: This is a very small nitpick, but I would recommend considering if we can use consistent language here. E.g. data is present and not present or data is available and not available. Currently it's unclear from the user's perspective whether the distinction between present/available is important or not.

web/html/src/components/datetime/DateTime.tsx
mackdk: I'm using this in the `HubDetailsForm` by the way, There I declare both a `<FromNow />` and a `<HumanDateTime />` trying to use a `Date` field. The former works, the latter rejects the type. That's why I changed.

web/html/src/components/large-text-attachment.tsx
```diff
@@ -0,0 +1,251 @@
...
+    editDialogTitle: t("Edit"),
+    editMessage: undefined,
+    confirmDeleteMessage: "Are you sure?",
+    onEdit: (_value: string) => Promise.resolve(),
```
Etheryte, line 67: Jfyi, off the top of my head this could also be `onEdit: async (_value:string) => undefined` if you don't want to use `Promise.resolve()`, but I don't have a preference over which style we use, either is fine.

web/html/src/manager/admin/hub/peripherals.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { Button } from "components/buttons";
+import withPageWrapper from "components/general/with-page-wrapper";
+import { PeripheralsList } from "components/hub";
+import { SectionToolbar } from "components/section-toolbar/section-toolbar";
+import { HelpLink } from "components/utils";
+
+const IssPeripheral = () => {
+  const title = (
+    <div className="spacewalk-toolbar-h1">
```
Etheryte, line 13: I think most places use `TopPanel` for this, so would be nice to use that here for consistency.

web/html/src/manager/admin/hub/peripherals.tsx
```diff
@@ -0,0 +1,54 @@
...
+              icon="fa-plus"
+              className={"btn-default"}
+              text={t("Add Peripheral")}
+              handler={addPeripheral}
+            />
+          </div>
+        </div>
+      </SectionToolbar>
+      <span>
+        <h1>Known Peripherals instances</h1>
```
Etheryte, line 47:
```suggestion
        <h1>{t("Known Peripherals instances")}</h1>
```

CDellaGiusta (approved):
LGTM.Not an expert in typescrypt

web/html/src/components/datetime/DateTime.tsx
Etheryte: Ah okay, I see, so these are values we are sending over from the server, and you want to pass those in? In that case I think it's fine, we can go ahead with this change. Essentially we want to avoid any frontend code that uses `new Date()` and such, and we have a linter rule for it as well, but forbidding it in many places makes it more clear that it shouldn't be done. But for receiving backend values it's fine.

mackdk: Yes, they come from the server JSON response as dates:
```typescript
export type HubDetailData = {
  id: number;
  fqdn: string;
  rootCA: string | null;
  gpgKey: string | null;
  sccUsername: string;
  created: Date;
  modified: Date;
};
```
I can also convert them to `Moment`  before passing it to the component, though.

web/html/src/components/hub/HubDetailsForm.tsx
```diff
@@ -0,0 +1,141 @@
...
+              </div>
+              <div className="col-md-10">
+                <LargeTextAttachment
+                  value={this.state.model.rootCA}
+                  filename="root-ca.pem"
+                  presentMessage={t("A customized Root Certificate Authority is currently configured for this hub.")}
+                  absentMessage={t("This hub is currently using the default Root Certificate Authority.")}
+                  editDialogTitle={t("Root Certificate Authority")}
+                  editMessage={t(
+                    "Ensure the certificate you are providing is correct before proceeding. " +
```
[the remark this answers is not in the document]
mackdk: Good point. I fixed it.

web/html/src/components/large-text-attachment.tsx
mackdk: Fixed

web/html/src/manager/admin/hub/peripherals.tsx
mackdk: fixed

web/html/src/components/datetime/DateTime.tsx
Etheryte: I think for values we receive from the server it's not a problem per se because there we know what we're working with. The problem is the frontend where if you use `Date` and your computer timezone is different than what your server or user configuration is, then there will be an issue.

PR #9844, opened 2025-02-24
RBAC: Initial implementation

cbbayburt:
This PR introduces the layout for RBAC filtering structure, without enforcing access or adding any endpoint/namespace data.

Optionally, here's how to enable RBAC in this version:
1. Add endpoint/namespace data to DB manually (I will provide DMLs separately)
2. Add `server.susemanager.rbac = 1` config option to `rhn.conf`

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


java/code/src/com/suse/manager/api/HttpApiRegistry.java
```diff
@@ -136,7 +139,7 @@ public static Set<String> getUnautenticatedRoutes() {
      */
     private void registerAuthEndpoints() {
         registrationHelper.addPostRoute(HTTP_API_ROOT + "auth/login", LoginController::apiLogin);
-        registrationHelper.addPostRoute(HTTP_API_ROOT + "auth/logout", withUser(LoginController::logout));
+        registrationHelper.addGetRoute(HTTP_API_ROOT + "auth/logout", withUser(LoginController::logout));
```
rjmateus, line 142: Why was this changed from get to post? Will this influence the way other pages/locations call this method? Or is this endpoint only used by costumers? If that is the case, we would need to add something to release notes.

java/code/src/com/redhat/rhn/common/security/acl/Access.java
```diff
@@ -100,6 +102,41 @@ public boolean aclUserRole(Map<String, Object> ctx, String[] params) {
...
+        }
+
+        if (user.hasRole(RoleFactory.SAT_ADMIN)) {
+            return true;
+        }
+
+        boolean authorized = UserManager.getPermittedNamespaces(user)
+                .filter(ns -> params[0].equals(ns.getNamespace()) ||
+                        ns.getNamespace().startsWith(params[0] + '.')
+                ).anyMatch(ns -> params.length < 2 ||
```
rjmateus, line 129: we should have some comments explaining this magic number. Why 2? (I'm assuming that it must have and least one parent, otherwise it would match the full path)

java/code/src/com/redhat/rhn/frontend/xmlrpc/BaseHandler.java
```diff
@@ -136,6 +141,15 @@ public Object invoke(String methodCalled, List params) throws XmlRpcFault {
             }
         }
 
+        try {
```
rjmateus, line 144: We should consider if we should continue to have the read-only user logic or if that should/can be replaced by a User group. 
I think we must replace the read-only API user with a new group in the logic that is being implemented in the RBAC, and in that case, the code that is placed a few lines above (L18-L142) should be removed.

java/code/src/com/redhat/rhn/manager/user/UserManager.java
```diff
@@ -148,6 +151,62 @@ public static boolean verifyPackagesAccess(Org org, List<Long> packageIds) {
...
+    }
+
+    /**
+     * Ensures that the current user is permitted access to a namespace with the specified access mode
+     * @param user the user
+     * @param namespace the namespace
+     * @param mode the access mode (view/modify)
+     * @throws PermissionException if the user is not permitted to access the specified namespace
+     */
+    public static void ensureRoleBasedAccess(User user, String namespace, Namespace.AccessMode mode) {
```
rjmateus, line 172: I think this method is not used. It also feels strange, since it doesn't return anything and checks user access and through an exception if he doesn't have access to the endpoint.

java/code/src/com/redhat/rhn/manager/user/UserManager.java
```diff
@@ -148,6 +151,62 @@ public static boolean verifyPackagesAccess(Org org, List<Long> packageIds) {
...
     }
 
+    /**
+     * Get a concatenated stream of namespaces granted direct (to the user)
+     * or indirect (via access groups) access.
+     * @param user the user
+     * @return the stream of permitted namespaces
+     */
+    public static Stream<Namespace> getPermittedNamespaces(User user) {
+        return Stream.concat(user.getNamespaces().stream(),
```
rjmateus, line 161: I think we can have this method with a more efficient implementation, where we could get all the user namespaces with a single query execution. We don't need to change it for this PR, but we should refactor it.

java/code/src/com/redhat/rhn/frontend/servlets/AuthorizationFilter.java
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+
+        // Pass control on to the next filter
+        chain.doFilter(request, response);
+    }
+
+    private void handleAjaxAccess(HttpServletRequest hreq, Set<String> noAuthEndpoints) {
+        String path = hreq.getServletPath() + hreq.getPathInfo();
+        if (!authenticationService.requestURIRequiresAuthentication(hreq) || noAuthEndpoints.contains(path)) {
```
rjmateus, line 109: I would prefer to not have a dependency from `AuthenticationServiceFactory`, but I was looking at the endpoints in there and they are super generic (like `/css`), so would be hard to move them to the database.

java/code/src/com/redhat/rhn/frontend/servlets/AuthorizationFilter.java
```diff
@@ -0,0 +1,198 @@
...
+        // Check if the request URI needs authentication. If so, also check if the URI needs authorization as well.
+        if (!authenticationService.requestURIRequiresAuthentication(hreq) ||
+                noAuthEndpoints.contains(route.getMatchUri())) {
+            return;
+        }
+
+        RequestContext requestContext = new RequestContext(hreq);
+        User user = requestContext.getCurrentUser();
+        if (user == null) {
+            if (!noAuthEndpoints.contains(route.getMatchUri())) {
```
rjmateus, line 180: I think we can just throw the exception without the if statement, since we are already checking the `noAuthEndpoints` in line 173.
HOwever, it doesn't cause any issue if we keep it as is. Up to you to decide :)

java/code/src/com/redhat/rhn/frontend/servlets/AuthorizationFilter.java
```diff
@@ -0,0 +1,198 @@
...
+    private void handleStrutsAccess(HttpServletRequest hreq, Set<String> noAuthEndpoints) {
+        // Check if the request URI needs authentication. If so, also check if the URI needs authorization as well.
+        if (!authenticationService.requestURIRequiresAuthentication(hreq) ||
+                noAuthEndpoints.contains(hreq.getServletPath())) {
+            return;
+        }
+
+        User user = new RequestContext(hreq).getCurrentUser();
+        if (user == null) {
+            if (!noAuthEndpoints.contains(hreq.getServletPath())) {
```
rjmateus, line 143: Same comment from the `handleSparkAccess`, up to you if you want to keep this check.

rjmateus (dismissed):
Looks good to me, just some small nitpick that I think doesn't influence the overall quality and the ability to merge the code. All the comments can be addressed after the beta version.

java/code/src/com/suse/manager/api/HttpApiRegistry.java
cbbayburt: I think it was a mistake in the initial HTTP API implementation. In XML-RPC it's `@ReadOnly` which means it's supposed to be a GET request in HTTP API (also no payload). That's why I changed it now. The change only affects the HTTP API (and the testsuite, which is included in the PR).

And yes, since it's on the public API, we'll need a release note.

java/code/src/com/redhat/rhn/common/security/acl/Access.java
cbbayburt: Right, this one needs a description.

Second element in params is the optional "access mode". If not provided, either R or W would match.

java/code/src/com/redhat/rhn/frontend/xmlrpc/BaseHandler.java
cbbayburt: Yes, the read-only logic can be fully replaced with RBAC. And yes, there'll be many obsolete parts then :)

java/code/src/com/redhat/rhn/manager/user/UserManager.java
cbbayburt: I still need to test the performance of different approaches, but with eager fetching, Hibernate should perform good enough on this.

java/code/src/com/redhat/rhn/frontend/servlets/AuthorizationFilter.java
cbbayburt: Right, I guess I overlooked this one.

java/code/src/com/redhat/rhn/manager/user/UserManager.java
cbbayburt: Oh this one has a long story:

Imagine a struts page that renders a JSP table with pagination buttons, select all button, etc.
On top of that, there's a button to do some real action on selected rows.

In Struts, all the pagination and the actual action are handled in the same handler with the same POST endpoint. However, pagination is a read operation, while the action is usually a "modify".

Since the endpoint is the same, the only sensible way is to make the differentiation in the handler code.
So this method is called from these handlers. And the way it doesn't return but throws is for convenience.

Here's some example usages: https://github.com/uyuni-project/uyuni/commit/d4517b322fae994cc573dde90a28f4fc47cdc322

java/code/src/com/redhat/rhn/common/security/acl/Access.java
cbbayburt: Here are some examples: https://github.com/uyuni-project/uyuni/commit/4f50a5d6ac351a88bc6226e425b5d0a42deb86cc

PR #9865, opened 2025-02-27
Click refresh when minion is down

maximenoel8:
## What does this PR change?
Fixes an issue where the Salt service is sometimes down on the minion, causing failures when checking the event. In most cases, refreshing the event resolves the problem.

This PR introduces a default behavior to prevent flaky tests. If the "Minion down" message appears during the event check, the process will automatically click the refresh button and restart the event check. Since these actions are performed within repeat_until_timeout, there is no risk of an infinite loop (verified through testing).

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -81,12 +81,17 @@ When(/^I wait at most "([^"]*)" seconds until I do not see "([^"]*)" text$/) do 
   end
 end
 
-When(/^I wait at most (\d+) seconds until the event is completed, refreshing the page$/) do |timeout|
+When(/^I wait at most (\d+) seconds until the "([^"]*)" is completed, refreshing the page$/) do |timeout, event|
   last = Time.now
   next if has_content?('This action\'s status is: Completed.', wait: 3)
 
   repeat_until_timeout(timeout: timeout.to_i, message: 'Event not yet completed') do
     break if has_content?('This action\'s status is: Completed.', wait: 3)
+
+    if has_content?('Minion is down or could not be contacted.', wait: 3)
+      find(:xpath, "//input[@value='Reschedule']").click
+      step %(I wait 30 seconds until the event is picked up and #{timeout} seconds until the event "#{event}" is completed)
+    end
     raise SystemCallError, 'Event failed' if has_content?('This action\'s status is: Failed.', wait: 3)
 
     current = Time.now
--- a/testsuite/features/step_definitions/common_steps.rb
+++ b/testsuite/features/step_definitions/common_steps.rb
@@ -180,7 +180,7 @@ When(/^I wait (\d+) seconds until the event is picked up and (\d+) seconds until
     And I follow first "#{event}"
     And I wait until I see "This action will be executed after" text
     And I wait until I see "#{event}" text
-    And I wait at most #{complete_timeout} seconds until the event is completed, refreshing the page
+    And I wait at most #{complete_timeout} seconds until the "#{event}" is completed, refreshing the page
   )
 end
 


ktsamis (approved):

srbarrios (changes requested):
I don't like this approach. This could hide issues.
That's why we try to don't use re-tries anywhere.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -81,12 +81,17 @@
...
   last = Time.now
   next if has_content?('This action\'s status is: Completed.', wait: 3)
 
   repeat_until_timeout(timeout: timeout.to_i, message: 'Event not yet completed') do
     break if has_content?('This action\'s status is: Completed.', wait: 3)
+
+    if has_content?('Minion is down or could not be contacted.', wait: 3)
+      find(:xpath, "//input[@value='Reschedule']").click
+      step %(I wait 30 seconds until the event is picked up and #{timeout} seconds until the event "#{event}" is completed)
+    end
```
srbarrios, line 94:
```suggestion

    minion_down_str = 'Minion is down or could not be contacted.'
    raise SystemCallError, minion_down_str if has_content?(minion_down_str, wait: 3)
```

Bischoff (changes requested):
This "Minion is down" is most probably a product bug that should be fixed.

The solution is not to hide the problem in the test suite.

meaksh (commented):
I'm also not much fan of this approach, as it should not be expected that the minion is down and the action is not scheduled. Even if this change is improving CI stability, it might hide potential issues.

We should probably understand why in some cases the minions are down in our CI environment.

In the past, this helped us to detect some actual problems in the product.

For instance, in the past we detected that minion startup can be stuck for a bunch of seconds, making the minion unresponsive during that time, while calculating the FQDNS grains in environments with a wrong DNS setup. If an action is scheduled within this time, it fails with "'Minion is down or could not be contacted" message.

We should probably debug why in some cases we get this now. It could be that the minion startup is taking long that expected, or that CI is scheduling actions too fast after a minion is restarted by some previous action (timing issue).

I would try to go in this direction.

srbarrios:
If we don't have a card, we might need want for this issue, and a bug report if it differs from these others:

- [Bug 1182851](https://bugzilla.suse.com/show_bug.cgi?id=1182851) is supposed to be fixed by the move to the Salt bundle.  
- [Bug 1192510](https://bugzilla.suse.com/show_bug.cgi?id=1192510) is about systems that reboot and don't process Salt events during that time. We could be hit by a reincarnation of that one.
- [Bug 1159492](https://bugzilla.suse.com/show_bug.cgi?id=1159492) was about "Valid metadata not found at specified URL". I don't think we see that message anymore.  
- [Bug 1057870](https://bugzilla.suse.com/show_bug.cgi?id=1057870) was again about reboot – this time with pending jobs lost due to the reboot.  
- [Bug 1207869](https://bugzilla.suse.com/show_bug.cgi?id=1207869) was due to high load caused by a customer script using our API.  
- [Bug 1196081](https://bugzilla.suse.com/show_bug.cgi?id=1196081) is still open; it's our 4.3 formulas making the proxy non-operational from a Salt point of view for 10 minutes.  
- [Bug 1172282](https://bugzilla.suse.com/show_bug.cgi?id=1172282) was one of our scripts overloading the Salt queue with both system upgrades and patch applications. In addition, there was again a problem due to reboot.  
- [Bug 1159092](https://bugzilla.suse.com/show_bug.cgi?id=1159092) was a scalability problem fixed in the meantime.  
- [Bug 1163965](https://bugzilla.suse.com/show_bug.cgi?id=1163965) was due to bad configuration on the customer side.  
- [Bug 1045381](https://bugzilla.suse.com/show_bug.cgi?id=1045381) was for CaaS.

PR #9871, opened 2025-02-28
[health-check] fix performance issue and add extra enhacements

meaksh:
## What does this PR change?

This PR fixes the performance issue found on supportconfigs due unexpected tons of Salt jobs. Additionally it does a couple of other fixes. See each commit individually.

NOTE: This PR is targetting `health-check-skeleton` feature branch.

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


python/health-check/src/health_check/containers/promtail/Dockerfile
```diff
@@ -1,8 +1,7 @@
 FROM opensuse/leap:latest
-#FROM registry.suse.com/bci/python:3.11
```
m-czernek, line 2: @ycedres did we settle in on the actual base images in the end?

python/health-check/src/health_check/main.py
```diff
@@ -94,14 +109,14 @@ def run(ctx: click.Context, from_datetime: str, to_datetime: str, since: int):
...
-def clean(ctx: click.Context):
+def stop(ctx: click.Context):
     verbose = ctx.obj["verbose"]
     clean_containers(verbose=verbose)
     console.print(Markdown("# Execution Finished"))
 
 
 def main():
-    console.print(Markdown("# Health Check"))
+    console.print(Markdown("# Uyuni Health Check"))
```
m-czernek, line 119: Same as above

python/health-check/src/health_check/config/templates/grafana_dashboard/supportconfig_with_logs.template.json
```diff
@@ -70,9 +83,9 @@
         "x": 0,
         "y": 6
       },
-      "id": 109,
+      "id": 71,
       "panels": [],
-      "title": "Memory",
+      "title": "Uyuni Server and Salt",
```
m-czernek, line 88: Note, same as above - I wanted to remove all `uyuni` occurrences from the Grafana board as well.

python/health-check/src/health_check/config/templates/grafana_dashboard/supportconfig_with_logs.template.json
```diff
@@ -1380,13 +1343,13 @@
     ]
   },
   "time": {
-    "from": "",
-    "to": ""
+    "from": "now-30d",
+    "to": "now"
```
m-czernek, line 1347: Is this intentional change? We leave this empty because health-check tool sets this based on the passed-in parameters

m-czernek (commented):
Additionally, can you add just a brief description of what you're modifying in the Grafana board? Not sure if you're just moving/renaming stuff, or if you're removing/modifying the view.

python/health-check/pyproject.toml
```diff
@@ -29,7 +29,7 @@ homepage = "https://github.com/uyuni-project/uyuni"
 tracker =  "https://github.com/uyuni-project/uyuni/issues"
 
 [project.scripts]
-health-check = "health_check.main:main"
+uyuni-health-check = "health_check.main:main"
```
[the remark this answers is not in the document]
meaksh: Thanks for clarifying this was actually intentional (I thought the rename was only related to the codebase files, but not for the CLI).

It just feels a bit weird to me that it doesn't have anything on the CLI name that refers to the what "health"  this is checking :) 

Should be maybe use `mgr-health-check` for the CLI tool?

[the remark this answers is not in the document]
m-czernek: Well, in my mind, "what" health are we checking is covered by the fact that you install/deploy it from Uyuni or MLM repositories. So if you're deploying it, you know exactly what health are you checking.

That said, I really like `mgr-health-check` so that we group all MLM/Uyuni scripts under one "namespace".

meaksh:
@m-czernek I just addressed your comments from the review.

About the changes on the dashboard:
- Rename sections/rows titles.
- Refactor Salt keys panel.
- Add Java configuration metrics.
- Added "misc" and "hw" metrics.
- Adjust panels to fit the entire window width.

Here are some screenshots:

Captura desde 2025-03-03 10-11-30
Captura desde 2025-03-03 10-11-43

meaksh:
@m-czernek I just addressed your comments from the PR review:

About the changes made on the dashboards:
- Refactor Salt Keys panels
- Add Java configuration metrics
- Add "misc" and "hw" metrics
- Adjust panels to fit entire windows width
- Rename sections/rows titles

Here are some screeshots:

Captura desde 2025-03-03 10-11-30
Captura desde 2025-03-03 10-11-43

python/health-check/src/health_check/main.py
```diff
@@ -60,7 +62,7 @@ def cli(ctx: click.Context, supportconfig_path: str, verbose: bool):
     callback=utils.validate_date,
 )
 @click.pass_context
-def run(ctx: click.Context, from_datetime: str, to_datetime: str, since: int):
+def start(ctx: click.Context, from_datetime: str, to_datetime: str, since: int):
```
m-czernek, line 65: Does this imply change in the docs? I.e. right now, we say to run `health-check run ...` - do we have to change it to `health-check start ...`, or is it just `health-check ...`? Either way, we'll probably need to modify the readme, right?

python/health-check/src/health_check/containers/promtail/Dockerfile
m-czernek: (I think Yeray is on PTO, we can probably return to this later)

m-czernek (approved):
This LGTM now - adding +1, but please also check whether readme doesn't need a change, since we're changing the `run/clean` method names.

Additionally, I'm not sure we want to tackle this within this PR, but I have a few suggestions regarding the dashboard.

WDYT about:

- Removing:
  -  worker threads, socket pool size, timeout, and gather job timeout
  - Java settings

Reason for this is that I'm not sure this is very helpful to viewers. What is helpful is the rules we have based on those values.

- Then, we might want to break out `num_of_channels` from miscellaneous as a separate counter (IMO this is interesting/important info to see, similar to num of CPUs and RAM)
- I wonder if we might make some transformation on `master/proxy/client`, where `1` is essentially `true` and `0` is `false`. Note that we cannot modify these actual values since we use the `1` and `0` in math expressions in some rules (e.g. checking minimal requirements). 

  But, we could either do a frontend transformation on the table, if possible, or do something else, e.g. create a new metric for front-end users. This way, it looks like the supportconfig has 1 server, 0 proxies and 0 clients, which might be a bit confusing.

python/health-check/src/health_check/main.py
meaksh: That's true, we need to change the README and wiki documentation.

I'll include the documentation changes and some of your suggestions in a follow-up PR.

meaksh:
> WDYT about:
> 
> * Removing:
>   
>   * worker threads, socket pool size, timeout, and gather job timeout
>   * Java settings
> 
> Reason for this is that I'm not sure this is very helpful to viewers. [...]

I think we should not remove worker threads, socket pool size, timeout, and gather job timeout metrics, neither the Java configuration values. IMO it is worth to have this complete and agreggated view of the Java configurations values in the dashboard.

Of course, if we have alerts already implemented to indicate some possible known issue, that is great, and we should have as many as possible, but having the actual values displayed here I think helps when debugging to identify not known issues, where we don't have alerts already implemented.

m-czernek:
I think the problem I have is that with some metrics, you can't even guess what the proper fully-qualified name is. With some metrics, like worker threads, this is not an issue. But with something like "timeout", this is useless - what kind of timeout? 

I added the Salt config in the beginning when I didn't have a clear idea of what I'm doing. The Java configs are similarly confusing.

meaksh:
> I think the problem I have is that with some metrics, you can't even guess what the proper fully-qualified name is. With some metrics, like worker threads, this is not an issue. But with something like "timeout", this is useless - what kind of timeout?
> 

I understand your point here, but I think we could just change the panels so it is clear that these are Salt Master configuration parameters and values. To me, it is still important that you see these values in the dashboard.

> I added the Salt config in the beginning when I didn't have a clear idea of what I'm doing. The Java configs are similarly confusing.

Hmm, ok, now I realized that the names of the Java setting displayed here are not actually the real name of the Java property. Of course, this is then confusing as you mentioned. I would suggest then that we adjust the metrics names provided by the exporter to match the real Java parameter name.

PR #9884, opened 2025-03-03
SonarCloud error reduction: HTML issues

CDellaGiusta:
## What does this PR change?
Fixes SonarCloud issues with HTML. All issues were located in a single old-style license HTML file.
Since the license referred to MX4J, which is no longer used anywhere, the license has been removed.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests needed
- [x] **DONE**

## Links
- [x] **DONE**

## Changelogs
- [x] No changelog needed


juliogonzalez (commented):
I am wondering, are we even use `MX4J` these days?

`grep -ir MX4J java/` only returns this file as relevant result, and one unrelated entry at `code/src/com/suse/manager/reactor/utils/test/productdata/repositories.json`

So not sure if the fix here would be to just remove the file? Could this be a leftover from the spacewalk times?

CDellaGiusta:
> So not sure if the fix here would be to just remove the file? Could this be a leftover from the spacewalk times?
I had the confirmation this software is not pointed anywhere, so I removed the file

juliogonzalez (approved):
LGTM, but please update the PR description :-)

PR #9888, opened 2025-03-04
Improve read timeout failing

maximenoel8:
## What does this PR change?

Improve the failing the time when we have read:timeout for more than 10 times.
Currently, in the testsuite, when we have the read:timeout, the failing time is taking more than 18 hours.
This error is probably related to the server or network behind slow to response. See: https://github.com/SUSE/spacewalk/issues/26585

This change will show use the current time response of the server and fail fast the features.

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/secondary/srv_change_task_schedule.feature
+++ b/testsuite/features/secondary/srv_change_task_schedule.feature
@@ -5,7 +5,7 @@
 Feature: Change the schedule of a task
 
   Scenario: Log in as admin user
-    Given I am authorized for the "Admin" section
+    Given I am authorized
 
   Scenario: Change the schedule of task sandbox-cleanup-default to weekly
     When I follow the left menu "Admin > Task Schedules"
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -57,6 +57,8 @@ $current_user = 'admin'
 $current_password = 'admin'
 $chromium_dev_tools = ENV.fetch('REMOTE_DEBUG', false)
 $chromium_dev_port = 9222 + ENV['TEST_ENV_NUMBER'].to_i
+$timeout_failure_counter = 0
+$timeout_threshold = 10  # Set the threshold for stopping the suite
 
 # maximal wait before giving up
 # the tests return much before that delay in case of success
@@ -131,21 +133,31 @@ $quality_intelligence = QualityIntelligence.new if $quality_intelligence_mode
 # Define the current feature scope
 Before do |scenario|
   $feature_scope = scenario.location.file.split(%r{(\.feature|/)})[-2]
+  check_read_timeout_threshold
 end
 
-# Embed a screenshot after each failed scenario
+# After hook to handle scenario failures and dump feature code coverage into Redis DB
 After do |scenario|
+  check_read_timeout_threshold
+
   current_epoch = Time.new.to_i
   log "This scenario took: #{current_epoch - @scenario_start_time} seconds"
+
+  # Check if the scenario failed due to a Net::ReadTimeout
   if scenario.failed?
-    begin
-      if web_session_is_active?
-        handle_screenshot_and_relog(scenario, current_epoch)
-      else
-        warn 'There is no active web session; unable to take a screenshot or relog.'
+    if scenario.exception.is_a?(Net::ReadTimeout)
+      increase_read_timeout_count
+      log_server_response_time
+    else
+      begin
+        if web_session_is_active?
+          handle_screenshot_and_relog(scenario, current_epoch)
+        else
+          warn 'There is no active web session; unable to take a screenshot or relog.'
+        end
+      ensure
+        print_server_logs
       end
-    ensure
-      print_server_logs
     end
   end
   page.instance_variable_set(:@touched, false) if capybara_session_created?
@@ -158,11 +170,25 @@ def capybara_session_created?
   Capybara::Session.instance_created?
 end
 
-# Test if web session is open
+# Test is web session is open
 def web_session_is_active?
   return false unless capybara_session_created?
 
   page.has_selector?('header') || page.has_selector?('#username-field')
+end
+
+def check_read_timeout_threshold
+  # Check if the consecutive timeout threshold is reached
+  if $timeout_failure_counter >= $timeout_threshold
+    warn "Timeout failure threshold reached, stopping test suite."
+    exit(1)
+  end
+end
+
+def increase_read_timeout_count
+  $timeout_failure_counter += 1
+
+  warn "Net::ReadTimeout detected! Consecutive ReadTimeouts: #{$timeout_failure_counter}"
 end
 
 # Take a screenshot and try to log back at suse manager server
@@ -205,6 +231,7 @@ def relog_and_visit_previous_url
     end
   rescue Timeout::Error
     warn "Timed out while attempting to relog and visit the previous URL: #{current_url}"
+    increase_read_timeout_count
   rescue StandardError => e
     warn "An error occurred while relogging and visiting the previous URL: #{e.message}"
   end
@@ -275,6 +302,7 @@ end
 
 # Create a user for each feature
 Before do |scenario|
+  check_read_timeout_threshold
   feature_path = scenario.location.file
   $feature_filename = feature_path.split(%r{(\.feature|/)})[-2]
   next if get_context('user_created') == true
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -58,8 +58,7 @@ $current_password = 'admin'
 $chromium_dev_tools = ENV.fetch('REMOTE_DEBUG', false)
 $chromium_dev_port = 9222 + ENV['TEST_ENV_NUMBER'].to_i
 $timeout_failure_counter = 0
-$timeout_threshold = 10  # Set the threshold for stopping the suite
-
+$timeout_threshold = 10 # Set the threshold for stopping the suite
 # maximal wait before giving up
 # the tests return much before that delay in case of success
 $stdout.sync = true
@@ -177,14 +176,15 @@ def web_session_is_active?
   page.has_selector?('header') || page.has_selector?('#username-field')
 end
 
+# Check if the consecutive timeout threshold is reached
 def check_read_timeout_threshold
-  # Check if the consecutive timeout threshold is reached
-  if $timeout_failure_counter >= $timeout_threshold
-    warn "Timeout failure threshold reached, stopping test suite."
-    exit(1)
-  end
+  return unless $timeout_failure_counter >= $timeout_threshold
+
+  warn 'Timeout failure threshold reached, stopping test suite.'
+  exit(1)
 end
 
+# Increase read timeout count
 def increase_read_timeout_count
   $timeout_failure_counter += 1
 
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -248,6 +248,7 @@ end
 
 # Dump feature code coverage into a Redis DB before we run next feature
 Before do |scenario|
+  check_read_timeout_threshold
   next unless $code_coverage_mode
 
   # Initialize $feature_path if that's the first feature
--- a/testsuite/features/support/system_monitoring.rb
+++ b/testsuite/features/support/system_monitoring.rb
@@ -113,3 +113,16 @@ def channel_synchronization_duration(channel)
 
   duration
 end
+
+# Method to check server response time (could be using an HTTP request or ping)
+def log_server_response_time
+  begin
+    start_time = Time.now
+    uri = URI.parse("https://#{ENV.fetch('SERVER', nil)}")
+    Net::HTTP.get_response(uri)
+    response_time = Time.now - start_time
+    log "Server response time: #{response_time} seconds"
+  rescue => e
+    warn "Error checking server response time: #{e.message}"
+  end
+end
--- a/testsuite/features/support/system_monitoring.rb
+++ b/testsuite/features/support/system_monitoring.rb
@@ -122,7 +122,7 @@ def log_server_response_time
     Net::HTTP.get_response(uri)
     response_time = Time.now - start_time
     log "Server response time: #{response_time} seconds"
-  rescue => e
+  rescue StandardError => e
     warn "Error checking server response time: #{e.message}"
   end
 end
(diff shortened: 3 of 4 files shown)


testsuite/features/support/env.rb
```diff
@@ -73,6 +74,10 @@
 $beta_enabled = (ENV.fetch('BETA_ENABLED', 'False') == 'True')
 $api_protocol = ENV.fetch('API_PROTOCOL', nil) if ENV['API_PROTOCOL'] # force the API protocol to be used. You can use 'http' or 'xmlrpc'
 
+# Define a global counter to track consecutive ReadTimeout errors
+$timeout_failure_count = 0
```
srbarrios, line 78: Better if we set this counter using the `add_context('timeout_failure_counter', 0)` and `get_context('timeout_failure_counter')`

testsuite/features/support/env.rb
```diff
@@ -73,6 +74,10 @@
...
+$timeout_threshold = 10  # Set the threshold for stopping the suite
```
srbarrios, line 79: This value does not aim to change during the whole cucumber execution, so I would better define it as a constant than a global variable. Also, if we have it as a global variable, please move it up to the block of lines 45-59

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
...
+    start_time = Time.now
+    uri = URI.parse("https://#{ENV.fetch('SERVER', nil)}")
+    Net::HTTP.get_response(uri)
+    response_time = Time.now - start_time
+    log "Server response time: #{response_time} seconds"
+  rescue => e
+    warn "Error checking server response time: #{e.message}"
+  end
+end
+
```
srbarrios, line 211: This function, it seems to fit better on `features/support/system_monitoring.rb`

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
...
+  end
+  feature_path = scenario.location.file
+  $feature_filename = feature_path.split(%r{(\.feature|/)})[-2]
+  next if get_context('user_created') == true
+
+  # Create own user based on feature filename. Exclude core, reposync, finishing and build_validation features.
+  unless feature_path.match?(/core|reposync|finishing|build_validation/)
+    step %(I create a user with name "#{$feature_filename}" and password "linux")
+    add_context('user_created', true)
+  end
```
srbarrios, line 150: We are duplicating these steps in another Before method, please let's remove this from this Before method. We better have separated Before methods each of them having a single responsibility, it gonna be more readable and easier to maintain.

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
 # Define the current feature scope
 Before do |scenario|
   $feature_scope = scenario.location.file.split(%r{(\.feature|/)})[-2]
+  if $timeout_failure_count >= $timeout_threshold
+    warn "Timeout failure threshold reached, stopping test suite."
+    exit(1)
```
srbarrios, line 140: What about raising an error instead?

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
...
+        else
+          warn 'There is no active web session; unable to take a screenshot or relog.'
+        end
+      ensure
+        print_server_logs
       end
-    ensure
-      print_server_logs
     end
   end
```
srbarrios, line 181: Too many if/else statements embedded, I think we need to re-think it.

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
...
 
-# Embed a screenshot after each failed scenario
+# After hook to handle scenario failures and dump feature code coverage into Redis DB
 After do |scenario|
+  # Handle timeout failure threshold
+  if $timeout_failure_count >= $timeout_threshold
+    warn "Timeout failure threshold reached, stopping test suite."
+    exit(1)
+  end
+
```
srbarrios, line 160: What about this https://cucumber.io/docs/cucumber/api/ ?
```suggestion
    warn 'Timeout failure threshold reached, stopping test suite.'
    Cucumber.wants_to_quit = true
```

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
...
+    step %(I create a user with name "#{$feature_filename}" and password "linux")
+    add_context('user_created', true)
+  end
 end
 
-# Embed a screenshot after each failed scenario
+# After hook to handle scenario failures and dump feature code coverage into Redis DB
 After do |scenario|
+  # Handle timeout failure threshold
+  if $timeout_failure_count >= $timeout_threshold
```
srbarrios, line 156: Let's please don't put anything before logging the time this scenario took, otherwise we might be giving wrong timing.

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
...
-      print_server_logs
     end
   end
+
+  # Check if we need to dump code coverage into Redis DB
+  next unless $code_coverage_mode
+  next unless $feature_path != scenario.location.file
+
+  process_code_coverage
+  $feature_path = scenario.location.file
```
srbarrios, line 188: Let's keep the code coverage process in a different After method as it was.

maximenoel8:
I tried different solutions to stop the cucumber testsuite and none are working. I tried `Cucumber.wants_to_quit = true`

And the testsuite is still running.
```
  Scenario: Bootstrap a y Linux 9 Salt minion             # features/test.feature:20
Timeout failure threshold reached, stopping test suite.
      This scenario ran at: 2025-03-04 20:55:14 +0100
    When I follow the left menu "Systems > Bootstrapping" # features/step_definitions/navigation_steps.rb:369
      Unable to find xpath "//aside/div[@id='nav']/nav/ul/li/div/a[contains(.,'Systems')]/parent::div/parent::li" (Capybara::ElementNotFound)
      ./features/step_definitions/navigation_steps.rb:390:in `block (2 levels) in <top (required)>'
      ./features/step_definitions/navigation_steps.rb:382:in `each'
      ./features/step_definitions/navigation_steps.rb:382:in `each_with_index'
      ./features/step_definitions/navigation_steps.rb:382:in `/^I follow the left menu "([^"]*)"$/'
      features/test.feature:21:in `I follow the left menu "Systems > Bootstrapping"'
    Then I should see a "Bootstrap Minions" text          # features/step_definitions/navigation_steps.rb:651
Reached the timeout failure threshold of 3 consecutive ReadTimeouts. Stopping the test suite.
There is no active web session; unable to take a screenshot or relog.
=> /var/log/rhn/rhn_web_ui.log

=> /var/log/rhn/rhn_web_api.log

      This scenario took: 10 seconds

  Scenario: Log inin user                         # features/test.feature:24
Timeout failure threshold reached, stopping test suite.
      This scenario ran at: 2025-03-04 20:55:45 +0100
    Given I am authorized for the "Admin" section # features/step_definitions/navigation_steps.rb:447
      Net::ReadTimeout with #<TCPSocket:(closed)> (Net::ReadTimeout)
      ./features/step_definitions/navigation_steps.rb:580:in `/^I am authorized as "([^"]*)" with password "([^"]*)"$/'
      ./features/step_definitions/navigation_steps.rb:450:in `/^I am authorized for the "([^"]*)" section$/'
      features/test.feature:25:in `I am authorized for the "Admin" section'
Reached the timeout failure threshold of 3 consecutive ReadTimeouts. Stopping the test suite.
Net::ReadTimeout detected! Consecutive ReadTimeouts: 4
```

The only solution I found is to exit all scenarios once we reach the limit.

maximenoel8:
About having two before |scenario | and two after |scenario|. I have to regroup them to be sure of the order.
This will require probably more testing but from my limited testing, having those two setup and teardown separate is creating a mess. For the approach of skipping every, they have to be together or have the timeout condition at the beginning of each. I have the feeling having those after and before duplicating is increasing the time in case the testsuite is failing.

testsuite/features/support/env.rb
maximenoel8: Cucumber.wants_to_quit = true seems to do nothing.

testsuite/features/support/env.rb
maximenoel8: By doing that, you remove the `web_session_is_active?` from the after step. This check save lot of time if an api call is failing.

testsuite/features/support/env.rb
```diff
@@ -130,26 +135,80 @@ def capybara_register_driver
...
+  end
 end
 
-# Embed a screenshot after each failed scenario
+# After hook to handle scenario failures and dump feature code coverage into Redis DB
 After do |scenario|
+  # Handle timeout failure threshold
+  if $timeout_failure_count >= $timeout_threshold
+    warn "Timeout failure threshold reached, stopping test suite."
+    exit(1)
```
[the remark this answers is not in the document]
maximenoel8: Doesn't work on my side.

testsuite/features/support/env.rb
maximenoel8: If the condition is true, all next features are going to be skipped.

testsuite/features/support/env.rb
maximenoel8: This code was creating a hanging process when running multiple features as in a same run.

testsuite/features/support/env.rb
maximenoel8: It fills over engineer. The variable is only use in this file.

srbarrios:
> About having two before |scenario | and two after |scenario|. I have to regroup them to be sure of the order.
> This will require probably more testing but from my limited testing, having those two setup and teardown separate is creating a mess. For the approach of skipping every, they have to be together or have the timeout condition at the beginning of each. I have the feeling having those after and before duplicating is increasing the time in case the testsuite is failing.

@maximenoel8  Please give me time to help you with this, can you please pause this development until March 28?

Moving everything to one method, just now, when we had that for years, it gonna make my talk much harder to show the code of the implementation about code coverage tracking, and I hope you understand we can put this on stand-by for some weeks, and I can work with you in that refactor once that talk is done.

srbarrios:
> I tried to use context, but I'm losing the count between features. You can see the changes here: https://github.com/uyuni-project/uyuni/pull/9888/commits/7fcbda523ae79ddfda8ce78b5cd3c8a531c1d566
> 
> I will revert to using global variable.

Ok, make sense, the context was designed to work at feature level 👍Sorry for the wrong suggestion.

Maybe we can refactor the context, so we can have a global context, but for that particular case, we are gonna need to use `synchronized`, to be on the safe side.

Using global variables in our test framework is bad practice, in particular if they are not read-only. The reason why this is a bad practice is about possible collisions writing on the variable when we run tests in parallel through the same cucumber process, but ok, our current runners are using multiple cucumber processes... fair enough...

srbarrios (changes requested):
Please let's postpone this refactor some weeks if you don't mind, so we can discuss together about the After/Before method. Thank you.

testsuite/features/secondary/srv_change_task_schedule.feature
```diff
@@ -5,7 +5,7 @@
 Feature: Change the schedule of a task
 
   Scenario: Log in as admin user
-    Given I am authorized for the "Admin" section
+    Given I am authorized
```
srbarrios, line 8: In theory, running it with just `Given I am authorized` it will use a new user name "srv_change_task_schedule", and that user will not have admin rights. So, in theory, the navigation to the admin section should fail.

In any case if you still want to use a regular use, change also the Scenario title accordily.

testsuite/features/support/system_monitoring.rb
```diff
@@ -113,3 +113,16 @@ def channel_synchronization_duration(channel)
 
   duration
 end
+
+# Method to check server response time (could be using an HTTP request or ping)
+def log_server_response_time
+  begin
+    start_time = Time.now
+    uri = URI.parse("https://#{ENV.fetch('SERVER', nil)}")
```
srbarrios, line 121:
```suggestion
    uri = URI.parse(Capybara.app_host)
```
```suggestion
    uri = URI.parse("https://#{get_target('server').full_hostname}")
```

testsuite/features/support/system_monitoring.rb
```diff
@@ -113,3 +113,16 @@ def channel_synchronization_duration(channel)
...
+    Net::HTTP.get_response(uri)
+    response_time = Time.now - start_time
+    log "Server response time: #{response_time} seconds"
```
srbarrios, line 124: I would just return the response time on this function.
And handle the logging outside. That way we can re-use this function for other purpose like adding a new Fitness metric, if we find it useful at some point.

```suggestion
    response_time = Time.now - start_time
```

testsuite/features/support/system_monitoring.rb
```diff
@@ -113,3 +113,16 @@ def channel_synchronization_duration(channel)
 
   duration
 end
+
+# Method to check server response time (could be using an HTTP request or ping)
+def log_server_response_time
```
srbarrios, line 118:
```suggestion
def server_response_time
```

testsuite/features/support/env.rb
```diff
@@ -221,6 +248,7 @@ def process_code_coverage
 
 # Dump feature code coverage into a Redis DB before we run next feature
 Before do |scenario|
+  check_read_timeout_threshold
```
srbarrios, line 251: I think we don't need to run it here, as we are already running in another `Before` method. If we add it here too, we will repeating that check multiple times before running a scenario.

testsuite/features/support/env.rb
```diff
@@ -275,6 +303,7 @@ def process_code_coverage
 
 # Create a user for each feature
 Before do |scenario|
+  check_read_timeout_threshold
```
srbarrios, line 306: I think we don't need to run it here, as we are already running in another `Before` method. If we add it here too, we will repeating that check multiple times before running a scenario.

srbarrios:
Hi @maximenoel8 , I was reviewing this PR and struggling to explain my suggestions, so I just decided to make an additional commit to your  PR.
If you don't like it, feel free to drop it, it was the easier way I found to share it.

PR #9899, opened 2025-03-05
ISSv3 - Add registration notifications

mackdk:
## What does this PR change?

This PR adds a new notification type `Hub registration changed` when a server is registered/deregistered. In particular, 

1) if hub server A registers server B as peripheral, B will get a notification
2) if hub server A deregisters server B, B will get a notification
3) if peripheral B deregisters from hub A, A will get a notification

To simply the process of adding new notification types, this PR refactors the way `NotificationType`s are linked to the enum constants. With this change it is no longer  needed to edit multiple places in the source code to add the new type.

This PR also refactor the frontend component, to remove the legacy code, fix multiple bugs (like severity not displayed) and avoid the need to list again the types in the Typescript code decoupling the backend definition from the frontend presentation.

### Before:
image
image

### After:
image
image
image
image
image

- [X] **DONE**

## Documentation
- Documentation WIP TBD

- [ ] **DONE**

## Test coverage
- No tests: already covered

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


java/spacewalk-java.changes.mackdk.issv3-add-registration-notifications
```diff
@@ -0,0 +1,2 @@
+- Added Hub Registration changed notification
+- Improve the notification definition and handling
```
Etheryte, line 2: I think the second changelog entry could either use some rewording, or alternatively be removed altogether. This changelog is for our customers, and I suspect notification definition and handling without much context might not tell them much. 🙂

mackdk: Indeed, I just removed it

web/html/src/manager/notifications/notifications-list.tsx
```diff
@@ -0,0 +1,454 @@
...
+export class NotificationList extends React.Component<Props, State> {
+  private readonly typeMap: Map<string, string>;
+
+  private readonly sortedNotificationTypes: Array<NotificationType>;
+
+  public constructor(props: Props) {
+    super(props);
+
+    // Sort the notification types by description
+    this.sortedNotificationTypes = [...this.props.notificationTypes].sort((a, b) =>
```
Etheryte, line 63: Since these values are coming straight from props, is there a reason not to send them in the right order to begin with?

web/html/src/manager/notifications/notifications-list.tsx
```diff
@@ -0,0 +1,454 @@
...
+            text={"[" + t("show details") + "]"}
+          />
+        )}
+      </span>
+    );
+  }
+
+  private renderReaction(data: Notification): React.ReactNode {
+    if (!data.actionable) {
+      return <></>;
```
Etheryte, line 254: Also fine to return `null`, jfyi.

web/html/src/manager/notifications/notifications-list.tsx
```diff
@@ -0,0 +1,454 @@
...
+      popupItem: undefined,
+      typeCriteria: [],
+    };
+  }
+
+  public componentDidMount(): void {
+    this.refreshServerData(this.state.currentDataUrlTag);
+  }
+
+  public componentDidUpdate(prevProps: Readonly<Props>, prevState: Readonly<State>): void {
```
Etheryte, line 89: I see this and the related code around `dataUrlTags` was already there in the old code that was there previously. I'm pretty certain we don't want this approach, but I'll have to get back to you once I learn how our current spa engine implementation works. In the meanwhile we're fine to proceed as-is, but just so you know, we don't want to replicate this elsewhere or to new code etc.

web/html/src/manager/notifications/notifications-list.tsx
```diff
@@ -0,0 +1,454 @@
+import * as React from "react";
```
Etheryte, line 1: Would be nice to chop this file up a little bit, right now it's starting to be a bit of a mouthful. Maybe slice by types, comparators, etc, but I don't have a strong preference here.

mcalmer (approved):

web/html/src/manager/notifications/notifications-list.tsx
mackdk: Fixed.

web/html/src/manager/notifications/notifications-list.tsx
mackdk: I extracted the types and the comparators into separe typescript source file. For the comparators I used a static class, I can change it to an object literal and functions if that's a preferred implementation style.

web/html/src/manager/notifications/notifications-list.tsx
mackdk: Honestly, I just left this code as is because I was not sure to fully understand what it was doing. Now I ha a second look and I saw that if we manually tamper with url the existing logic breaks. So I now simplified how we handle the hash and the selected tab. Not sure if your comment was only referring to that or if it was also about the `componentDidMount()`/`componentDidUpdate()` strategy. If that's the case, let me know how we should handle it in the correct way :-)

Etheryte: My comment was about the hash situation, but I've created a separate ticket to look into that in a wider sense, so feel free to disregard that topic under this PR.

Etheryte (approved):
Frontend looks good, thank you very much for the clean code. 👍

web/html/src/manager/notifications/notifications-list.tsx
Etheryte: That looks good, no strong preference on which way to go on that front. 👍

java/code/src/com/redhat/rhn/domain/notification/types/ChannelSyncFailed.java
```diff
@@ -62,14 +58,6 @@ public NotificationMessage.NotificationMessageSeverity getSeverity() {
         return NotificationMessage.NotificationMessageSeverity.ERROR;
     }
```
CDellaGiusta, line 58: could it be worth to declare severity directly in the notification type? This could become something like: 
```
public NotificationMessage.NotificationMessageSeverity getSeverity() {
        return getType().getSeverity();
    }
```

CDellaGiusta (approved):
code looks good

java/code/src/com/redhat/rhn/domain/notification/types/ChannelSyncFailed.java
mackdk: I implemented it, see commit  https://github.com/uyuni-project/uyuni/pull/9899/commits/2e4042b9a8db40a2f3d8edaed889e5586f6d662c

CDellaGiusta: Awesome. Very clear.

PR #9918, opened 2025-03-07
Distinguish between OS of real host and of container

Bischoff:
## What does this PR change?

This PR makes the difference between OS family and version of real host and container.

In the case of non-containerized systems, both fields will be identical.

## Links

Port(s):
 * 5.0: https://github.com/SUSE/spacewalk/pull/26636
 * 4.3: non applicable

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -274,7 +274,7 @@ end
 # @return [Boolean] Returns true if the system is a SLE/SL Micro one
 def slemicro_host?(name)
   node = get_target(name)
-  os_family = node.os_family
+  os_family = node.local_os_family
   (name.include? 'slemicro') || (name.include? 'micro') || os_family.include?('sle-micro') || os_family.include?('suse-microos')
 end
 
@@ -284,7 +284,7 @@ end
 # @return [Boolean] Returns true if the system is a openSUSE Leap Micro one.
 def leapmicro_host?(name)
   node = get_target(name)
-  os_family = node.os_family
+  os_family = node.local_os_family
   os_family.include?('opensuse-leap-micro')
 end
 
--- a/testsuite/features/support/remote_node.rb
+++ b/testsuite/features/support/remote_node.rb
@@ -7,7 +7,7 @@ require_relative 'network_utils'
 # The RemoteNode class represents a remote node.
 # It is used to interact with the remote node through SSH.
 class RemoteNode
-  attr_accessor :host, :hostname, :port, :target, :full_hostname, :private_ip, :public_ip, :private_interface, :public_interface, :os_family, :os_version, :has_mgrctl
+  attr_accessor :host, :hostname, :port, :target, :full_hostname, :private_ip, :public_ip, :private_interface, :public_interface, :os_family, :os_version, :local_os_family, :local_os_version, :has_mgrctl
 
   # Initializes a new remote node.
   #
@@ -46,7 +46,11 @@ class RemoteNode
     raise StandardError, "No FQDN for '#{@hostname}'. Response code: #{code}" if @full_hostname.empty?
 
     $stdout.puts "Host '#{@host}' is alive with determined hostname #{@hostname} and FQDN #{@full_hostname}" unless $build_validation
+
+    # Determine OS version and OS family both inside the container and on the local host
+    # in the case of non-containerized systems, both fields will be identical:
     @os_version, @os_family = get_os_version
+    @local_os_version, @local_os_family = get_os_version(runs_in_container: false)
 
     if (PRIVATE_ADDRESSES.key? host) && !$private_net.nil?
       @private_ip = net_prefix + PRIVATE_ADDRESSES[host]
@@ -327,9 +331,9 @@ class RemoteNode
 
   # Extract the OS version and OS family
   # We get these data decoding the values in '/etc/os-release'
-  def get_os_version
-    os_family_raw, code = run('grep "^ID=" /etc/os-release', check_errors: false)
-    os_family_raw, code = run('sw_vers --productName', check_errors: false) if code.nonzero?
+  def get_os_version(runs_in_container: true)
+    os_family_raw, code = run('grep "^ID=" /etc/os-release', runs_in_container: runs_in_container, check_errors: false)
+    os_family_raw, code = run('sw_vers --productName', runs_in_container: runs_in_container, check_errors: false) if code.nonzero?
     return nil, nil unless code.zero?
 
     os_family = os_family_raw.strip
@@ -339,12 +343,12 @@ class RemoteNode
     os_family.delete! '"'
 
     if os_family == 'macOS'
-      os_version_raw, code = run('sw_vers --productVersion', check_errors: false)
+      os_version_raw, code = run('sw_vers --productVersion', runs_in_container: runs_in_container, check_errors: false)
       return nil, nil unless code.zero?
 
       os_version = os_version_raw.strip
     else
-      os_version_raw, code = run('grep "^VERSION_ID=" /etc/os-release', check_errors: false)
+      os_version_raw, code = run('grep "^VERSION_ID=" /etc/os-release', runs_in_container: runs_in_container, check_errors: false)
       return nil, nil unless code.zero?
 
       os_version = os_version_raw.strip.split('=')[1]


testsuite/features/support/commonlib.rb
```diff
@@ -274,7 +274,7 @@ def suse_host?(name)
 # @return [Boolean] Returns true if the system is a SLE/SL Micro one
 def slemicro_host?(name)
   node = get_target(name)
-  os_family = node.os_family
+  os_family = node.real_host_os_family
```
szachovy, line 277: How about calling it node? real_host sounds mysterious

testsuite/features/support/remote_node.rb
```diff
@@ -327,9 +328,9 @@ def client_public_ip
 
   # Extract the OS version and OS family
   # We get these data decoding the values in '/etc/os-release'
-  def get_os_version
-    os_family_raw, code = run('grep "^ID=" /etc/os-release', check_errors: false)
-    os_family_raw, code = run('sw_vers --productName', check_errors: false) if code.nonzero?
+  def get_os_version(real_host: false)
+    os_family_raw, code = run('grep "^ID=" /etc/os-release', run_in_container: !real_host, check_errors: false)
```
szachovy, line 332: Instead of this can we just set flag on the `run_the_container`?

szachovy (dismissed):

testsuite/features/support/commonlib.rb
Bischoff: Er, no. "node" is already taken for the object obtained by "get_target" (see just one line above).

Looking for a better name, how would you qualify the system running on the real host versus the system running in the container?

testsuite/features/support/remote_node.rb
Bischoff: `run_in_container` has sense for the `run()` function.

I don't see how this variable would make sense for ` get_os_version()`. The question is, "get OS version of what?"

testsuite/features/support/remote_node.rb
```diff
@@ -327,9 +328,9 @@ def client_public_ip
...
+    os_family_raw, code = run('sw_vers --productName', run_in_container: !real_host, check_errors: false) if code.nonzero?
```
srbarrios, line 333:
```suggestion
  def get_os_version(run_in_container: false)
    os_family_raw, code = run('grep "^ID=" /etc/os-release', run_in_container: run_in_container, check_errors: false)
    os_family_raw, code = run('sw_vers --productName', run_in_container: run_in_container, check_errors: false) if code.nonzero?
```

testsuite/features/support/remote_node.rb
```diff
@@ -339,12 +340,12 @@ def get_os_version
     os_family.delete! '"'
 
     if os_family == 'macOS'
-      os_version_raw, code = run('sw_vers --productVersion', check_errors: false)
+      os_version_raw, code = run('sw_vers --productVersion', run_in_container: !real_host, check_errors: false)
```
srbarrios, line 343:
```suggestion
      os_version_raw, code = run('sw_vers --productVersion', run_in_container: run_in_container, check_errors: false)
```

testsuite/features/support/remote_node.rb
```diff
@@ -339,12 +340,12 @@ def get_os_version
...
       return nil, nil unless code.zero?
 
       os_version = os_version_raw.strip
     else
-      os_version_raw, code = run('grep "^VERSION_ID=" /etc/os-release', check_errors: false)
+      os_version_raw, code = run('grep "^VERSION_ID=" /etc/os-release', run_in_container: !real_host, check_errors: false)
```
srbarrios, line 348:
```suggestion
      os_version_raw, code = run('grep "^VERSION_ID=" /etc/os-release', run_in_container: run_in_container, check_errors: false)
```

testsuite/features/support/remote_node.rb
```diff
@@ -7,7 +7,7 @@
 # The RemoteNode class represents a remote node.
 # It is used to interact with the remote node through SSH.
 class RemoteNode
-  attr_accessor :host, :hostname, :port, :target, :full_hostname, :private_ip, :public_ip, :private_interface, :public_interface, :os_family, :os_version, :has_mgrctl
+  attr_accessor :host, :hostname, :port, :target, :full_hostname, :private_ip, :public_ip, :private_interface, :public_interface, :os_family, :os_version, :real_host_os_family, :real_host_os_version, :has_mgrctl
```
srbarrios, line 10:
```suggestion
  attr_accessor :host, :hostname, :port, :target, :full_hostname, :private_ip, :public_ip, :private_interface, :public_interface, :os_family, :os_version, :host_os_family, :host_os_version, :has_mgrctl
```

testsuite/features/support/remote_node.rb
szachovy: See Oscar's suggestions

testsuite/features/support/remote_node.rb
```diff
@@ -46,7 +46,8 @@ def initialize(host, port: 22)
     raise StandardError, "No FQDN for '#{@hostname}'. Response code: #{code}" if @full_hostname.empty?
 
     $stdout.puts "Host '#{@host}' is alive with determined hostname #{@hostname} and FQDN #{@full_hostname}" unless $build_validation
-    @os_version, @os_family = get_os_version
+    @os_version, @os_family = get_os_version(real_host: false)
+    @real_host_os_varsion, @real_host_os_family = get_os_version(real_host: true)
```
srbarrios, line 50:
```suggestion
if @has_mgrctl
    @container_os_version, @container_os_family = get_os_version(run_in_container: true)
    @os_version, @os_family = get_os_version(run_in_container: false)
else
    @os_version, @os_family = get_os_version(run_in_container: false)
end    
```

Bischoff: 1- This test is already done in get_os_version
2 - some variables are not set in one of the cases

But I agree it's more legible

testsuite/features/support/commonlib.rb
Bischoff: I had a looooong huddle with Oscar :smile: 

The problem if we invert the log and now call `os_family` what is _outside` of the container is that it can break at every point where we use that attribute. I like Oscar's idea, but it's risky. So I think I will merge "as is".

testsuite/features/support/remote_node.rb
Bischoff: See remark above.

testsuite/features/support/remote_node.rb
Bischoff: Oscar also prefers if we keep same parameter name. I don't like bending the semantics, but okay, I will comply.

testsuite/features/support/remote_node.rb
Bischoff: See previous comment, I will comply.

testsuite/features/support/remote_node.rb
Bischoff: same.

testsuite/features/support/commonlib.rb
Bischoff: @szachovy what about using `outside` inside of `real_host` ? Meaning "outside of container" ? Would that work better for you ?

Bischoff: Or `local_`

testsuite/features/support/remote_node.rb
```diff
@@ -327,9 +328,9 @@ def client_public_ip
 
   # Extract the OS version and OS family
   # We get these data decoding the values in '/etc/os-release'
-  def get_os_version
-    os_family_raw, code = run('grep "^ID=" /etc/os-release', check_errors: false)
-    os_family_raw, code = run('sw_vers --productName', check_errors: false) if code.nonzero?
+  def get_os_version(run_in_container: true)
+    os_family_raw, code = run('grep "^ID=" /etc/os-release', run_in_container: run_in_container, check_errors: false)
```
nodeg, line 332: Where does `run_in_container` come from? I cannot find it here in the PR.

nodeg: And apparently our test do not know either:

```bash
PXE boot MAC address variable empty
unknown keyword: :run_in_container (ArgumentError)
/testsuite/features/support/remote_node.rb:102:in `run'
/testsuite/features/support/remote_node.rb:332:in `get_os_version'
/testsuite/features/support/remote_node.rb:49:in `initialize'
/testsuite/features/support/remote_nodes_env.rb:36:in `new'
/testsuite/features/support/remote_nodes_env.rb:36:in `get_target'
/testsuite/features/support/commonlib.rb:664:in `new_api_client'
```

Bischoff: Hmmm, let me look for typos

Bischoff: Fixed, thanks for the catch

testsuite/features/support/remote_node.rb
```diff
@@ -47,6 +47,7 @@ def initialize(host, port: 22)
 
     $stdout.puts "Host '#{@host}' is alive with determined hostname #{@hostname} and FQDN #{@full_hostname}" unless $build_validation
     @os_version, @os_family = get_os_version
+    @local_os_varsion, @local_os_family = get_os_version(runs_in_container: false)
```
srbarrios, line 50:
```suggestion
    @local_os_version, @local_os_family = get_os_version(runs_in_container: false)
```

Bischoff: Grrrr... Sorry. Fixed.

testsuite/features/support/commonlib.rb
```diff
@@ -274,7 +274,7 @@ def suse_host?(name)
 # @return [Boolean] Returns true if the system is a SLE/SL Micro one
 def slemicro_host?(name)
   node = get_target(name)
-  os_family = node.os_family
+  os_family = node.local_os_family
```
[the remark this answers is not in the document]
Bischoff: Because we are mixing the fact of being transactional with the fact of have a container part

=> precisely not! in my code, the local_os_family is defined even if the host is not containerized.

testsuite/features/support/remote_node.rb
```diff
@@ -47,6 +47,7 @@ def initialize(host, port: 22)
 
     $stdout.puts "Host '#{@host}' is alive with determined hostname #{@hostname} and FQDN #{@full_hostname}" unless $build_validation
     @os_version, @os_family = get_os_version
```
NamelessOne91, line 49: Would `get_os_version`  assumes a default true `runs_in_container` argument in this case ?
If so, wouldn't this cause issues for hosts with nothing running in a container ?

Bischoff: After a huddle: adding a comment in the code to make these lines a bit clearer.

NamelessOne91 (approved):

testsuite/features/support/remote_node.rb
Bischoff: done

testsuite/features/support/remote_node.rb
Bischoff: after a huddle with Oscar, too risky - rejected

testsuite/features/support/remote_node.rb
Bischoff: Functionally equivalent, with more risk, rejected, even if Paolo kind of liked the suggestion

srbarrios: I don't see more risk, and I see it better from an OOP design. Still, as said, your call to implement it as you like, because I can't commit to track these changes or help on stabilizing them.

PR #9957, opened 2025-03-12
WIP:Fix Raw Data Types with Parameterized Generics

hemantyb:
## Fix Raw Data type And Enhance Type Safety

1.  **Generics & Type Safety**
    
    -   Replaced raw types (e.g., `Set`, `DataResult`) with parameterized generics (`Set<RhnSetElement>`, `DataResult<PackageComparison>`).
    -   Removed unsafe type casts (e.g., `(PackageComparison)`), leveraging Java’s compile-time checks.

2.  **Cleaner Code**
    
    -   Used the **diamond operator** (`<>`) for collections (e.g., `new HashSet<>(dataResult)`) to reduce verbosity.
    -   Replaced `Map<String, Object>` with a custom `Row` class for better expressiveness.
- [ ] **DONE**

## Documentation

https://sonarcloud.io/project/issues?directories=java%2Fcode%2Fsrc%2Fcom%2Fredhat%2Frhn%2Fmanager%2Frhnpackage&issueStatuses=OPEN%2CCONFIRMED&id=uyuni-project_uyuni&open=AYMYhZgl5TQr2KFg-01k

https://sonarcloud.io/project/issues?fileUuids=AYMYhWJn5TQr2KFg-x_E&issueStatuses=OPEN%2CCONFIRMED&id=uyuni-project_uyuni&open=AYMYhXvK5TQr2KFg-zwd

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


mbussolotto:
Reason why GH actions are failing:
- Changelogs: you need to create a changelog entry, please have a look here https://github.com/uyuni-project/uyuni/wiki/Contributing#changelog-rules-for-packages-in-git and https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs (I strongly suggest to use `mkchlog` to make it simple)
- java_pgsql_test: these are the unit test, a ; is missing  
``` 
[javac] /manager/java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java:1760: error: ';' expected
[javac]                     release.add(release)
```
here you can find information about how to run junit locally https://github.com/uyuni-project/uyuni/wiki/Java-Development-Environment#optional-running-junit-tests but in this case it was enought to compile it with `ant -buildfile java/manager-build.xml compile`

java/spacewalk-java.changes.hemantyb.fix-generic-parameter
```diff
@@ -0,0 +1,2 @@
+- Fix: fix the generics for set, iterator, DataResult and others,
```
Etheryte, line 1: This changelog has a different purpose from a git commit, we use it to generate user-facing changelogs that are shipped with the release. As such, this changelog should reflect feature changes and be aimed at an end user, not developers. As is, I would recommend omitting this changelog altogether since there's no user-facing changes.

hemantyb:
@Etheryte @mbussolotto 
issues I am facing right now:
- if I didn't add changelog entry, the GH actions are failing.
- when I compile "ant -buildfile java/manager-build.xml compile" using this command (for test purpose)
  ```
  [javac] /home/eman/Documents/code/java/uyuni/java/code/src/com/suse/manager/webui/services/impl/SaltService.java:58: error: package com.suse.salt.netapi.calls.modules does not exist
    [javac] import com.suse.salt.netapi.calls.modules.Event;
  ```
  (more like these but change in filename)
- java_pgsql_test: don't know what can I do to pass this test case.

solutions I have done:
- added changelog entry -> passed one test case: but no need so removed it -> that one test case failed
- change the complete development enviroment to tumbleweed (previously on mint)
- check all java related build tools and other are setup properly, like ant, ivy and maven.

extra things that I want to know:
- can I continue working on the same branch and fix more issues while the PR is active?
- is there any other test related thing that I need to know before pushing the code?

mbussolotto:
> added changelog entry -> passed one test case: but no need so removed it -> that one test case failed

Right now is fine: `Changelogs / Warn user if merged (pull_request_target)` is skipped since now ` No changelog needed
` is marked, so no problem with it.

> java_pgsql_test: don't know what can I do to pass this test case.

This one now it's failing because the code cannot compile. If you click on `Java Checkstyle` (which is failing as well), you'd see the problem (here's the link https://github.com/uyuni-project/uyuni/actions/runs/13823095128/job/38699016806?pr=9957 ).

> when I compile "ant -buildfile java/manager-build.xml compile" using this command 

did you have a look on https://github.com/uyuni-project/uyuni/wiki/Java-Development-Environment ? In particular https://github.com/uyuni-project/uyuni/wiki/Java-Development-Environment#configure-the-development-machine , I think you're missing some dependencies, try to redo 
```
 sudo zypper addrepo obs://systemsmanagement:Uyuni:Utils systemsmanagement:uyuni:utils
 sudo zypper install obs-to-maven
cd <path_to_spacewalk>/java
ant -f manager-build.xml ivy
ant -f manager-build.xml refresh-branding-jar
```
then
```
ant -f manager-build.xml compile
```

> can I continue working on the same branch and fix more issues while the PR is active?

Better to use have on PR for each feature/fix: if the fix are related to this change yes, continue to use this one. Please ask for review when the work on the PR is finished, otherwise a new review would be required.

> is there any other test related thing that I need to know before pushing the code?

You can run locally on your machine almost each of the GH actions. I think `Java Development Environment` guide should contains all of this information. Feel free to ask if something it's not clear. In particular I advice to run locally (when the code compiles) `checkstyle` and junit tests, in this way you would save lots of time

Etheryte:
The changelog test should now pass, why it previously failed was because the corresponding section in the PR template was not filled out. If you tick the box in the template that says no changelog is needed for this PR, it will not fail when you don't have one.  

Regarding having more changes in one PR, it depends, if the changes are related or similar, it's good to put them in one PR, if not, having a separate PR would be better. If you want to add more changes and keep working under the same PR, it would be polite to add "WIP: ..." to the PR title and ask for reviews once you're done.

hemantyb:
> Hello @mbussolotto @Etheryte
> 
> # Dependencies related issue
> How can I install these dependencies:
> 
> * hibernate
> * org.apache.commons
> * redstone.xmlrpc
> 
> All dependecies which are mentioned in pom.xml I have added using jar file, and some of them added manually. but these three are just not working.
> 
> I tried:
> 
> * adding these dependencies in pom.xml
> * added modules
> 
> btw I am using intellij (for better developer experience), should I use eclipse?
> 
> Now to check that code is compiling or not after changes I am using act and test the java-checkstyle(GH action locally), but this is time consuming. so thats why I want to setup the project with dependencies.

hemantyb:
@Etheryte @mbussolotto 

hello, I fix some raw data types, and I am still working on it, but can you please review the code and tell me am I going right or need something to consider.

and, some GH actions are failing which is related to podman. How it relavent to this issue?

hemantyb:
I have a que, that when I'm providing the generics, so some cast are redundanct so Im removing them. 

But also if there is a for loop I am assigning type directly in for loop so that we don't need extra variable and this makes code more readable and concise.

Is it okay to do this? 

because removing code may cause issue, btw I am making sure that it will not cause any issue on some method, class or anywhere. 

so am I allowed to do that?

@Etheryte @mbussolotto

mbussolotto:
> and, some GH actions are failing which is related to podman. How it relavent to this issue?
right now you have an error just on checkstyle https://github.com/uyuni-project/uyuni/actions/runs/14018639423/job/39295445594?pr=9957

hemantyb:
> right now you have an error just on checkstyle https://github.com/uyuni-project/uyuni/actions/runs/14018639423/job/39295445594?pr=9957

I have done it.

will you review the code and give me a feedback that I am going well or not, meaning the generics I am providing will not break any functionality and also they are backward compatible. 
I am making sure that code will compile and now on I will do checkstyle also.

java/code/src/com/redhat/rhn/domain/contentmgmt/ContentProject.java
```diff
@@ -339,7 +339,7 @@ public void setProjectFilters(List<ContentProjectFilter> filtersIn) {
      *
      * @param filter the filter to attach
      */
-    public void attachFilter(ContentFilter filter) {
+    public void attachFilter(ContentFilter<FilterCriteria> filter) {
```
mackdk, line 342: This parameter type looks wrong to me.  The implementations of `ContentFilter` are:

- [ErrataFilter](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/contentmgmt/ErrataFilter.java), which is a `ContentFilter<Errata>`
- [ModularPackageFilter](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/contentmgmt/ModularPackageFilter.java), which is a `ContentFilter<Package>`
- [ModuleFilter](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/contentmgmt/ModuleFilter.java), which is a `ContentFilter<Module>`
- [PackageFilter](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/contentmgmt/PackageFilter.java), which is a `ContentFilter<Package>`
- [PtfFilter](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/domain/contentmgmt/PtfFilter.java), which is a `ContentFilter<Package>`

`FilterCriteria` is not the type of any of these and it's not correct in this context. All the actual types have nothing in common, so we can't pinpoint a common type. We could either:

- use `ContentFilter<?>` since most content management methods using `ContentFilter` do not care about  the actual type. This approach however will generate another SonarCloud issue "Generic wildcard types should not be used in return types" 
- Define a common `ContentFilterEntity` interface and have `Errata`, `Package`, `Module` implement it. Then we can change `class ContentFilter<T>` to `class ContentFilter<T extends ContentFilterEntity>` and all raw occurrences of `ContentFilter` to `ContentFilter<? extends ContentFilterEntity>`. This approach should work without any additional warning from SonarCloud. The only caveats is that `ContentFilterEntity` is going to probably be a marker interface with no methods.

mackdk (changes requested):

java/code/src/com/redhat/rhn/domain/contentmgmt/ContentProject.java
hemantyb: @mackdk  
Chages were made by creating interface,
Let me know is if there is an issue with these chages.

java/code/src/com/redhat/rhn/domain/ContentFilterEntity.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2024 SUSE LLC
```
mackdk, line 2: I this should be 2025, I guess :slightly_smiling_face: . Check your licence settings.

java/code/src/com/redhat/rhn/domain/ContentFilterEntity.java
```diff
@@ -0,0 +1,15 @@
...
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+
+package com.redhat.rhn.domain;
+
+public interface ContentFilterEntity {
```
mackdk, line 14: I think it's better to add a javadoc comment here, to explain why this interface exists and that it must be implemented to use a class in a ContentFilter.

java/code/src/com/redhat/rhn/manager/contentmgmt/ContentManager.java
```diff
@@ -1051,12 +1057,17 @@ private void alignPackageCache(Channel channel, Set<Package> oldChannelPackages)
...
     }
 
     // helper for extracting certain filter types
     // it's not optimal to run this method multiple times for same collection of filters, but at least it's clear
-    private static <T> List<T> extractFiltersOfType(Collection<ContentFilter> filters, Class<T> type) {
+    private static <T> List<T> extractFiltersOfType(Collection<ContentFilter<? extends ContentFilterEntity>> filters,
+                                                    Class<T> type) {
         return filters.stream()
                 .filter(f -> type.isAssignableFrom(f.getClass()))
                 .map(f -> (T) f)
```
mackdk, line 1073: I know this was not part of your change, but to avoid the unchecked warning this should be:

```java
                .map(f -> type.cast(f))
```

java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java
```diff
@@ -1756,8 +1753,12 @@ public static Package findDebugSource(User user, Package pack) {
 
     private static String getAssociatedRelease(Package pack) {
         for (Channel chan : pack.getChannels()) {
+            List<String> releases = new ArrayList<>();
             for (DistChannelMap map : chan.getDistChannelMaps()) {
-                return map.getRelease();
+                String release = map.getRelease();
+                if (release != null && !release.isEmpty()) {
+                    releases.add(release);
+                }
```
mackdk, line 1761: Not sure to understand this change. 

The original implementation looks broken indeed as it always returns the first release if available, not sure why. 

Now on the other hand, you add the release to the releases list which is unused and in the end the return value is now always `null`.

mackdk (commented):
Looks mostly good to me now :+1: Good work! Please, rebase your changes, so that the suite can run.

java/code/src/com/redhat/rhn/domain/ContentFilterEntity.java
mackdk: Ah, and I forgot. I think this interface should be in package `com.redhat.rhn.domain.contentmgmt` like `ContentFilter`

java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java
hemantyb: this was another issue on sonarcloud

https://sonarcloud.io/project/issues?fileUuids=AYMYhWJx5TQr2KFg-ycZ&issueStatuses=OPEN%2CCONFIRMED&id=uyuni-project_uyuni&open=AYMYhZgl5TQr2KFg-02d

I fixed it unknowingly but members said having seperate PR for different issues would be better and it makes sense.

mackdk: I agree this should be in another PR. I get why SonarCloud is complaining though, we could revrite this as:

```java
    private static String getAssociatedRelease(Package pack) {
        return pack.getChannels().stream()
            .flatMap(chan -> chan.getDistChannelMaps().stream())
            .findFirst()
            .map(DistChannelMap::getRelease)
            .orElse(null);
    }
```

This should make SonarCloud happy withouth changing the current logic.

java/code/src/com/redhat/rhn/domain/contentmgmt/ContentFilterEntity.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Marker interface for entities compatible with {@link ContentFilter}.
+ * Implemented by {@link Errata}, {@link Package}, {@link Module}, etc.,
+ * to enforce type-safe filtering operations.
+ */
+
+package com.redhat.rhn.domain.contentmgmt;
+
+public interface ContentFilterEntity {
+}
```
mackdk, line 21: The javadoc is misplaced, not sure if it will be picked up by the javadoc tool/IDE if placed like it is now. It should be:

```suggestion
package com.redhat.rhn.domain.contentmgmt;

/**
 * Marker interface for entities compatible with {@link ContentFilter}.
 * Implemented by {@link Errata}, {@link Package}, {@link Module}, etc.,
 * to enforce type-safe filtering operations.
 */
public interface ContentFilterEntity {
}
```

java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java
hemantyb: Yeah, so shall I revert my changes on this, or apply this suggestion?

and I am getting warning that this branch has a conflict that must be resolved.
What can I do about it?

mackdk: >Yeah, so shall I revert my changes on this, or apply this suggestion?

I would just revert it and leave the code as is. It's safer for now.

> and I am getting warning that this branch has a conflict that must be resolved.
>What can I do about it?

Since you have already merge commits in this branch, you should merge latest `master` into your branch and resolve the conflicts. We deleted many classes related to ISSv1, so you should resolve those conflicts by deleting them as well. For the future, IMHO it's better to rebase your branch on top of master (`git rebase --onto uyuni/master <commit id where your changes are based on>) instead of merging it, it keeps your history cleaner while working on your change.

hemantyb:
@mackdk 

can you take a look at this PR:
https://github.com/uyuni-project/uyuni/pull/10198#issue-2996079837

this branch was messed so, now I I will try to divide this issue in parts.

PR #9969, opened 2025-03-13
Sometimes we need the OS inside of the container, sometimes not

Bischoff:
## What does this PR change?

Sometimes we need the OS inside of the container, sometimes not

## Links

Port(s):
  * 5.0:
  * no 4.3 (no containers)

## Changelogs

- [x] No changelog needed


Pablogoliva (approved):

srbarrios:
In my opinion, we need to refactor RemoteNote, to have as attributes `host` and `container`, which will be both of a new class `System`, and on that class we have all these methods around which kind of system we have.

Example:

`get_target(host).host.is_suse?`
`get_target(host).container.is_suse?`

PR #9980, opened 2025-03-15
Add SLES 15 SP7 as potential base OS

Bischoff:
## What does this PR change?

Forward port of https://github.com/SUSE/spacewalk/pull/26622

## Links

5.0 fixes:

## Changelogs

- [ ] No changelog needed

### DIFF
--- a/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
+++ b/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
@@ -788,6 +788,7 @@ Feature: Synchronize products in the products page of the Setup Wizard
 
 @susemanager
 @proxy
+@transactional_server
   Scenario: Add SUSE Linux Enterprise Micro 5.5 needed for Proxy
     Given I am authorized for the "Admin" section
     When I follow the left menu "Admin > Setup Wizard > Products"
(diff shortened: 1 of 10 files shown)


testsuite/features/build_validation/reposync/srv_sync_all_products.feature
```diff
@@ -836,7 +858,25 @@ Feature: Synchronize products in the products page of the Setup Wizard
...
+    And I wait until I do not see "currently running" text
+    And I wait until I do not see "Loading" text
+    And I enter "SUSE Manager Proxy Extension 5.0 x86_64" as the filtered product description
+    When I open the sub-list of the product "SUSE Linux Enterprise Server 15 SP7 x86_64"
+    And I select "SUSE Manager Proxy Extension 5.0 x86_64" as a product
+    Then I should see the "SUSE Manager Proxy Extension 5.0 x86_64" selected
+    When I click the Add Product button
+    And I wait until I see "Selected channels/products were scheduled successfully for syncing." text
+    And I wait until I see "SUSE Manager Proxy Extension 5.0 x86_64" product has been added
+    And I wait until all synchronized channels for "suma-proxy-extension-50-sp7" have finished
```
srbarrios, line 874: This name looks super weird.
It's like mixing MLM version (5.0) even if though it should be 5.1, with the SLES 15 service pack, but without prepending the 15 part. So... very bad naming :/

Bischoff: Oops, i's of course 5.1, thanks for the catch!

Bischoff: As for the naming, I also consider it as suboptimal, but we discussed it in QE channel, I just aligned with the convention for channels, which we can't modify anymore as it is released.

ktsamis (approved):

PR #9994, opened 2025-03-18
Pin all GitHub actions to a commit

juliogonzalez:
## What does this PR change?

One of the recommendations of the [Good security practices for using GitHub Actions features](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions?learn=getting_started).

For each action I went to the repository and used the last released version for the major version we were using, and specified the exact version with a comment.

If we were using master, I pinned to the latest commit.

I added some warnings about things to be reviewed later. The purpose of this PR is pinning to commits only.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: Adjustement on the GitHub actions.

- [x] **DONE**

## Test coverage
- No tests: No changes to the functionality

- [x] **DONE**

## Links

None

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/.github/workflows/acceptance_tests_common.yml
+++ b/.github/workflows/acceptance_tests_common.yml
@@ -26,8 +26,8 @@ jobs:
     outputs:
       require_acceptance_tests: ${{ steps.filter.outputs.java == 'true' || steps.filter.outputs.web == 'true' || steps.filter.outputs.testsuite == 'true' }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: dorny/paths-filter@v3
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 #v3.0.2
         id: filter
         with:
           filters: |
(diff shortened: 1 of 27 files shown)


nodeg:
In that case, we also need to update our Dependabot workflow, to take that into account for the future when it creates PRs to update our actions.

srbarrios:
> I was wondering, can we also not simply use a specific version number (e.g. `@v3.0.2`) instead of the hash? Does anybody know if this would be a solution that our security team would allow?

Those are tags. And the last commit on the tag can be changed.

srbarrios (approved):
If that's a security team requirement Ok.
Although, I think we should iterate over this at some point... because we should be able to trust at least some of them, and use tags again.

szachovy (approved):

Etheryte (approved):
No comment from me, I don't know enough about the background here.

juliogonzalez:
Ready to be merged

Short-term, I hope dependabot will update the workflows using commit IDs (but TBH I am not sure if it was managing them at all right now)

Mid-term, we could evaluate if we can trust enough some organizations. As much as we we want to secure the secrets or prevent commits, we should be very careful about this.

As GitHub says:

> Pin actions to a tag only if you trust the creator
> Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators. The ‘Verified creator’ badge on GitHub Marketplace is a useful signal, as it indicates that the action was written by a team whose identity has been verified by GitHub. Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action.

So even verified creators can be risky.

juliogonzalez:
Just for the record, yes, dependabot is tracking our workflows: https://github.com/uyuni-project/uyuni/blob/master/.github/dependabot.yml#L12

PR #10003, opened 2025-03-19
Fix: Applying the highstate to Salt SSH minions causes HHH000179 warning at rhn_taskomatic_daemon.log (bsc#1216936)

CDellaGiusta:
## What does this PR change?
Taskomatic log file shows a warning like:

`     WARN  org.hibernate.engine.internal.StatefulPersistenceContext - HHH000179: Narrowing proxy to class com.redhat.rhn.domain.server.MinionServer - this operation breaks ==`

This fix solves this problem. The problem seems related to the Hibernate lazy initialization of class MinionServer, which is a child class from Server: as you may see from the following two queries, there seems to be an issue of what type to infer when doing lazy initialization
```
    <query name="Server.findByIds">
        <![CDATA[from com.redhat.rhn.domain.server.Server as s where s.id in (:serverIds)]]>
    </query>

    <query name="Server.findMinionsByServerIds">
        <![CDATA[from com.redhat.rhn.domain.server.MinionServer as s where s.id in (:serverIds)]]>
    </query>
```

While I was checking the taskomatic logs, I noticed that there was another recurring warning when restarting taskomatic    

```
WARN  org.hibernate.mapping.RootClass - HHH000038: Composite-id class does not override equals(): 
com.redhat.rhn.domain.action.salt.inspect.ImageInspectActionResult$ImageInspectActionResultId
WARN  org.hibernate.mapping.RootClass - HHH000039: Composite-id class does not override hashCode(): 
com.redhat.rhn.domain.action.salt.inspect.ImageInspectActionResult$ImageInspectActionResultId
```

So I decided to fix it as well (boy-scout rule). I won't squash this second commit.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: manually tested
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/22968
- [ ] **DONE**

## Changelogs
- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/salt/inspect/ImageInspectActionResult.java
+++ b/java/code/src/com/redhat/rhn/domain/action/salt/inspect/ImageInspectActionResult.java
@@ -18,6 +18,7 @@ import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 import java.io.Serializable;
+import java.util.Objects;
 
 import javax.persistence.Column;
 import javax.persistence.Embeddable;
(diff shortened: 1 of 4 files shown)


mcalmer:
@CDellaGiusta a unit test is failing

CDellaGiusta:
After consulting with Orion squad and @mbussolotto, it has been noted that:

- hibernate lazy="false" could potentially bring performance problems
- queries regarding MinionServer are scattered all around the system
- there is an ongoing hibernate upgrade refactoring, and this part has to be refactored
- this is a log warning but everything is stable and working
- the warning could be removed in log4j2.xml, but this is not recommended, since removing warnings has been a source of problems in the past
- the issue is labelled P3, so it's supposed to be "best effort"

Given the above considerations and the analysis, the bug will be left opened and this PR will be closed without merging it.

PR #10006, opened 2025-03-19
Remove old client tools instead of appending, fix sp7 string

ktsamis:
## What does this PR change?

Remove old client tools instead of appending, fix sp7 string to add `(BETA)`.
The design decision was to append the new client tools instead of replacing so we can keep constants.rb file aligned in all branches. But the problem is that we then have the old client tools empty in 5.1 and we need to add them to the list to be ignored. But that causes the constants file to be not aligned anyway so it doesn't make sense. So better to remove the old ones completely.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): no issue, directly from testsuite review
Port(s): no ports needed only for master/head
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


Bischoff (approved):
Perfect

PR #10008, opened 2025-03-19
Fix SL Micro detection in testsuite

ktsamis:
## What does this PR change?

Fix SL Micro detection in testsuite

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): no issue, directly from testsuite detection
Port(s): 5.0:
4.3:

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


NamelessOne91 (dismissed):
LGTM, thanks

I wonder if we should just reduce this to

```ruby
(name.include? 'micro) || (os_family.include? 'micro')
```

ktsamis:
> LGTM, thanks
> 
> I wonder if we should just reduce this to
> 
> ```ruby
> (name.include? 'micro) || (os_family.include? 'micro')
> ```

I thought about reducing the whole thing in length yes but I'm not sure if it matches anything else in the future, it's a bit hacky right now but accurate so I don't want to change it

testsuite/features/reposync/srv_create_bootstrap_repositories.feature
```diff
@@ -13,10 +13,16 @@ Feature: Create bootstrap repositories
...
+@containerized_server
 @proxy
   Scenario: Create the bootstrap repository for Proxy
     When I create the bootstrap repository for "proxy" on the server
 
+@skip_if_transactional_server
+@proxy
+  Scenario: Create the bootstrap repository for Proxy
+    When I create the bootstrap repository for "proxy_nontransactional" on the server
+
```
Bischoff, line 25: changes in this file are not needed, this is taken care of by the step

testsuite/features/support/constants.rb
```diff
@@ -395,6 +395,7 @@
     'SLE-Product-SLES15-SP4-Pool for x86_64' => 'sle-product-sles15-sp4-pool-x86_64',
     'SLE-Product-SLES15-SP5-Pool for x86_64' => 'sle-product-sles15-sp5-pool-x86_64',
     'SLE-Product-SLES15-SP6-Pool for x86_64' => 'sle-product-sles15-sp6-pool-x86_64',
+    'SLE-Product-SLES15-SP7-Pool for x86_64' => 'sle-product-sles15-sp7-pool-x86_64',
```
Bischoff, line 398: that would rather be for the "SP7" card but fine

testsuite/features/support/commonlib.rb
```diff
@@ -276,7 +276,7 @@ def suse_host?(name, runs_in_container: true)
 def slemicro_host?(name, runs_in_container: true)
   node = get_target(name)
   os_family = runs_in_container ? node.os_family : node.local_os_family
-  (name.include? 'slemicro') || (name.include? 'micro') || os_family.include?('sle-micro') || os_family.include?('suse-microos')
+  (name.include? 'slemicro') || (name.include? 'micro') || os_family.include?('sle-micro') || os_family.include?('suse-microos') || os_family.include?('sl-micro')
```
Bischoff, line 279: good catch

Bischoff (changes requested):
see remarks

testsuite/features/support/constants.rb
ktsamis: no I think this is used for label for the channel to be used in the activation key, either way at some point we need it

Pablogoliva (approved):

Bischoff (dismissed):

ktsamis:
After discussing with Eric, I remove the scenario "Create the bootstrap repository for Proxy" because it fails because SL Micro 6 MLM tools are missing still from SCC. The proxy bootstrap after works fine even without this though, so we need to pass into secondary to have results running, we don't know how long this will take to be added in SCC. I made a reminder for myself to check in a couple of weeks about these client tools and remove the workaround, hopefully it all goes well

Bischoff (approved):
approving as discussed

deneb-alpha:
agreed with @ktsamis  to bypass and merge

Issue #9848, opened 2025-02-25
Oracle Linux 8 - Remi Repo Sync Failure


TuxOps-lab:
### Problem description

Syncing Remi Repo fails with PREPENDED_DIR

### Steps to reproduce

1. Try to sync ol8 Repo Remi with Url: http://cdn.remirepo.net/enterprise/8/remi/x86_64/mirror
2.
3.
...

### Uyuni version

```bash
Information for package Uyuni-Server-release:
---------------------------------------------
Repository     : uyuni-server-stable
Name           : Uyuni-Server-release
Version        : 2024.12-241000.225.1.uyuni5
Arch           : x86_64
Vendor         : openSUSE
Support Level  : Level 3
Installed Size : 1.4 KiB
Installed      : No
Status         : not installed
Source package : Uyuni-Server-release-2024.12-241000.225.1.uyuni5.src
Upstream URL   : https://www.uyuni-project.org/
Summary        : Uyuni Server
Description    : 
    Uyuni lets you efficiently manage physical, virtual,
    and cloud-based Linux systems. It provides automated and cost-effective
    configuration and software management, asset management, and system
    provisioning.
```

### Uyuni proxy version (if used)

```bash

```

### uyuni-tools version used to deploy/manage the server or proxy (if relevant)

```bash

```

### Useful logs

```bash
14:18:57     6756/6758 : wordpress-6.7.1-1.el8.remi.noarch.rpm
14:19:04     6757/6758 : wordpress-6.7.2-1.el8.remi.noarch.rpm
14:19:13     6758/6758 : wkhtmltopdf-opt-0.12.4-3.el8.remi.x86_64.rpm
14:19:14 Filtering packages that failed to download
14:19:14 
14:19:14   Importing packages to DB:
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
[...]

PR #10013, opened 2025-03-20
Remove direct access to CFG when importing pkgs

m-czernek:
## What does this PR change?

We get sporadic reports of the following error during reposync:

```
14:19:13     6758/6758 : wkhtmltopdf-opt-0.12.4-3.el8.remi.x86_64.rpm
14:19:14 Filtering packages that failed to download
14:19:14 
14:19:14   Importing packages to DB:
14:19:14 PREPENDED_DIR
14:19:14 PREPENDED_DIR
```

This seems not directly reproducible (we have not determined root cause), but we suspect that this is the result of direct access to `CFG`, where one component doesn't properly close the namespace while another component accesses it with unexpected/incorrect namespace.

In this PR, we use the `cfg_component` context manager to force proper initCFG.

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni/issues/9848
Port(s): 
- 5.0 - https://github.com/SUSE/spacewalk/pull/28175
- 5.1 - https://github.com/SUSE/spacewalk/pull/28176

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/server/importlib/packageImport.py
+++ b/python/spacewalk/server/importlib/packageImport.py
@@ -30,7 +30,7 @@ from .importLib import (
 )
 from .mpmSource import mpmBinaryPackage
 from uyuni.common import rhn_pkg
-from spacewalk.common.rhnConfig import CFG
+from spacewalk.common.rhnConfig import cfg_component
 from spacewalk.server import taskomatic
 from spacewalk.server.rhnServer import server_packages
 
@@ -93,12 +93,14 @@ class ChannelPackageSubscription(GenericPackageImport):
         self.backend.lookupChecksums(self.checksums)
 
         # Fix the package information up, and uniquify the packages too
+        with cfg_component("server") as CFG:
+            enable_nvrea = CFG.ENABLE_NVREA
         uniqdict = {}
         for package in self.batch:
             if package.ignored:
                 continue
             self._postprocessPackageNEVRA(package)
-            if not CFG.ENABLE_NVREA:
+            if not enable_nvrea:
                 # nvrea disabled, skip checksum
                 nevrao = (
                     package["name_id"],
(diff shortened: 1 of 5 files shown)


mcalmer:
@m-czernek I had this in our testsuite when syncing tumbleweed repo. In the same testrun we synced also other repositories which did not show this problem. But we should really fix it. Please take care that this get reviewed, merged and also backported

python/spacewalk/server/importlib/packageImport.py
```diff
@@ -93,12 +93,14 @@ def fix(self):
         self.backend.lookupChecksums(self.checksums)
 
         # Fix the package information up, and uniquify the packages too
+        with cfg_component("server") as CFG:
+            enable_nvrea = CFG.ENABLE_NVREA
```
mcalmer, line 97: I am not so sure about how python behave here with variable scopes.
But is "enable_nvrea" available outside of the with when it is specified the first time inside of it?

m-czernek: Yes, `with` does not create a variable scope (similar to other keywords like `if` or `while`)

m-czernek:
So the issue with this PR is that in https://github.com/uyuni-project/uyuni/issues/9613, we couldn't verify that this PR helped the actual issue.

This was our best guess as to where the issue lies, but the issue creator couldn't reproduce the issue after a while.

WDYT @agraul , should we merge this? It shouldn't cause errors, but I'm not sure it targets the root cause, though merging this and the issue happening again would rule out the `CFG` access as the root cause.

mcalmer:
My 2ct:
We want to move in general to the use of "with" and this is a step forward. So it is definitive not wrong. So let's merge it and find out if it already solve the original problem.

python/spacewalk/server/importlib/packageImport.py
```diff
@@ -30,7 +30,7 @@
 )
 from .mpmSource import mpmBinaryPackage
 from uyuni.common import rhn_pkg
-from spacewalk.common.rhnConfig import CFG
+from uyuni.common.context_managers import cfg_component
```
agraul: The actual location is in `spacewalk.common.rhnConfig` nowadays, the reason for the move is in `uyuni.common.context_managers`:

```python
# This module always had a dependency on spacewalk.common.rhnConfig, which is wrong.
# Everything in uyuni.common should work on clients as well. When we add other context
# managers (that are available on clients), this import should be guarded.
# Re-export to allow users to keep their import on uyuni.common.context_managers
# pylint: disable-next=unused-import
from spacewalk.common.rhnConfig import cfg_component
```

m-czernek:
@agraul fixed your comment - could you re-check please?

agraul (approved):

m-czernek:
Merging, thank you all!

PR #10015, opened 2025-03-20
Fix: openscap audit is running immediately even when scheduled for next days (bsc#1239743)

CDellaGiusta:
## What does this PR change?
The follwing bugged behaviour is fixed: when scheduling a new OpenScap XCCDF Scan, and the earliest date was set in the future, the scan use to start immediately.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: manually tested (UI)
- [x] **DONE**

## Links
Issue(s): 
Port(s): https://github.com/SUSE/spacewalk/pull/26747, https://github.com/SUSE/spacewalk/pull/26748
- [x] **DONE**

## Changelogs
- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/struts/StrutsDelegate.java
+++ b/java/code/src/com/redhat/rhn/frontend/struts/StrutsDelegate.java
@@ -341,7 +341,7 @@ public class StrutsDelegate {
         //if it is not there, then that means we should always evaluate the
         //date picker.  Otherwise, we evaluate if it tells us to do so.
         if (!form.getMap().containsKey(DatePicker.SCHEDULE_TYPE) ||
-                form.get(DatePicker.SCHEDULE_TYPE) == null ||
+                StringUtils.isEmpty(String.valueOf(form.get(DatePicker.SCHEDULE_TYPE))) ||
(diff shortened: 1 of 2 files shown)


java/code/src/com/redhat/rhn/frontend/struts/StrutsDelegate.java
```diff
@@ -341,7 +341,7 @@ public Date readDatePicker(DynaActionForm form, String name, int yearDirection)
         //if it is not there, then that means we should always evaluate the
         //date picker.  Otherwise, we evaluate if it tells us to do so.
         if (!form.getMap().containsKey(DatePicker.SCHEDULE_TYPE) ||
-                form.get(DatePicker.SCHEDULE_TYPE) == null ||
+                StringUtils.isEmpty(String.valueOf(form.get(DatePicker.SCHEDULE_TYPE))) ||
```
lucidd, line 344: If i understand this correctly this change will now make it so that if schedule type is an empty string it will use the date picker and also since `String.valueOf(null)` returns the string "null" it will not use the date picker anymore when schedule type is null. Is this correct?

PR #10027, opened 2025-03-20
Update only minor and major versions of github-actions

srbarrios:
## What does this PR change?

Update only minor and major versions of github-actions.

Warning:
I was forced to do an additional change due to this error:
image

## Links

No ports

- [x] **DONE**

## Changelogs

- [x] No changelog needed

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "npm" # This value means either NPM or Yarn, Yarn doesn't have a separate value
-    directory: "/web/html/src/**"
+    directory: "/web/html/src"
     open-pull-requests-limit: 0
     schedule:
       interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,6 @@ updates:
     directory: "/"
     schedule: 
       interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]  # Ignore patch updates


Etheryte (approved):

jordimassaguerpla (approved):

.github/dependabot.yml
```diff
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "npm" # This value means either NPM or Yarn, Yarn doesn't have a separate value
-    directory: "/web/html/src/**"
+    directory: "/web/html/src"
     open-pull-requests-limit: 0
```
srbarrios, line 5: Note:
I'm unsure if that's made in purpose, but setting this to 0, will not open any PR.

.github/dependabot.yml
```diff
@@ -13,3 +13,6 @@ updates:
     directory: "/"
     schedule: 
       interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-patch"]  # Ignore patch updates
```
srbarrios, line 18: Doc reference:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#ignore--

.github/dependabot.yml
szachovy: I think that in the Security tab, there are suggestions, and you can open a PR manually if you find it appropriate instead of letting automation handle it. The bad thing of `>0` is that if nobody merges them on time, conflicts may arise, which need to be resolved. This way, you are manually opening and merging at the same time, but it may also reduce visibility.

.github/dependabot.yml
szachovy: We need to balance overload of PR's from dependabot from each package patch, bad thing is we don't know which patches are super important. Considering getting security notifications by other means (like security tab) it's not perfect but may be enough. I don't know what with other possibilities where (non-security involved) important patch is not applied. @juliogonzalez any ideas how to balance it?

szachovy (changes requested):

.github/dependabot.yml
juliogonzalez: @szachovy in that case I'd understand the security champions will handle that?

If so, this could be a valid approach, so then you evaluate if something is important or not.

szachovy: Yes, but there is no 100% guarantee as we have a lot of packages and I would manually need to enter each package and read latest commits. We can with this settings and see how it works.

Resolving

szachovy (approved):

juliogonzalez:
> I think that in the Security tab, there are suggestions, and you can open a PR manually if you find it appropriate instead of letting automation handle it. The bad thing of >0 is that if nobody merges them on time, conflicts may arise, which need to be resolved. This way, you are manually opening and merging at the same time, but it may also reduce visibility.

Please track it. You should have access to the tab, as you are part of the security managers.

If that works, LGTM. You may want to configure reminders for you and Carlo to have a look from time to time.

.github/dependabot.yml
Etheryte: This is intentional, because the way Dependabot tries to update e.g. frontend subdependencies is not workable and often results in PRs that build, but fail at runtime. We spent a lot of pointless time on this when we first introduced Dependabot, and came to the conclusion that it's more useful to intentionally set aside time to go through the notifications manually than have it make PRs on the go. This is a large part of what I work on every time I go to the bug squad.

deneb-alpha:
@srbarrios asked to NOT merge this PR now. he will handle the PR next week

PR #10031, opened 2025-03-21
Checkout the PR code on Tests Base workflow

srbarrios:
## What does this PR change?

As we changed the workflow calling that one, to use `pull_request_target` now that code in use by default when we checkout, is the code from the base branch. Instead, we need to checkout the code from the PR.

## Links

No ports

- [x] **DONE**

## Changelogs

- [x] No changelog needed


Bischoff (approved):

deneb-alpha:
ok from @Bischoff to bypass and merge

PR #10033, opened 2025-03-21
Refactor to run the GH validation on the fork's runner

srbarrios:
## What does this PR change?

`acceptance_tests_coordinator.yml` still runs on `pull_request_target` (so it has access to secrets).
But instead of calling workflows directly via `workflow_call` (which would inherit pull_request_target), it triggers `repository_dispatch`.
`repository_dispatch` runs as an event on the forked repository (i.e., PR head branch), NOT the base repo.
Since the actual test workflows (`acceptance_tests_recommended.yml, acceptance_tests_sequential.yml, acceptance_tests_parallel.yml`) now run on `repository_dispatch`, the GitHub runner will execute on the fork, testing the PR code.

## Links

No ports

- [x] **DONE**

## Changelogs

- [x] No changelog needed

### DIFF
--- a/.github/workflows/acceptance_tests_coordinator.yml
+++ b/.github/workflows/acceptance_tests_coordinator.yml
@@ -41,7 +41,7 @@ jobs:
           REDIS_HOST: ${{ secrets.REDIS_HOST }}
           REDIS_PORT: ${{ secrets.REDIS_PORT }}
         run: |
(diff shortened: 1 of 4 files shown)


jordimassaguerpla (approved):

jordimassaguerpla:
The issue is about 15.6 product ... has nothing to do with your PR. :thinking:  What about rebasing and retrigerring the test?  I do not see this test failing in master

PR #10034, opened 2025-03-21
Supportdata add Action Details and API

mcalmer:
## What does this PR change?

We need additional parameter for the support data action. This PR add Action Details to store the case number and additional parameter which can be used at execution time.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- Part of an extra issue

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/26621

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/supportdata/UploadGeoType.java
+++ b/java/code/src/com/redhat/rhn/domain/action/supportdata/UploadGeoType.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2025 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+package com.redhat.rhn.domain.action.supportdata;
+
+import com.redhat.rhn.common.localization.LocalizationService;
+import com.redhat.rhn.domain.Labeled;
+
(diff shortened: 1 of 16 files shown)


java/code/src/com/redhat/rhn/domain/action/supportdata/UploadGeoType.java
```diff
@@ -0,0 +1,28 @@
...
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+package com.redhat.rhn.domain.action.supportdata;
+
+import com.redhat.rhn.common.localization.LocalizationService;
+import com.redhat.rhn.domain.Labeled;
+
+public enum UploadGeoType implements Labeled {
+    EU,
```
rjmateus, line 17: The idea behind this is to allow users to select if they want to update to SCC to a server located in US or EU.
Should we take one more step and say if is SUSE or another provider? I was wondering if we should consider in the future this integration to other vendors (example RH), if we should try to incorporate this flexibility in the code (I know that is out of the scope, but we could design in a way we could better accommodate that in the future)

rjmateus (commented):
Just one small comment that doesn't block this PR. Other then that look good to me

rjmateus (approved):

java/code/src/com/redhat/rhn/domain/action/supportdata/UploadGeoType.java
mcalmer: Yes, this is just to decide if you want to upload to our EU or US server.
For different targets we would need a lot more. I expect the authentication is different and maybe also the protocol.
I would just implement this only if we really want to do it

PR #10035, opened 2025-03-21
Catch when web session instance is nil

Bischoff:
## What does this PR change?

This PR manages the case where Capybara::Session is nil .

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Cucumber tests were added

- [x] **DONE**

## Links

Port(s):
 * 5.0: https://github.com/SUSE/spacewalk/pull/26764
 * 4.3: https://github.com/SUSE/spacewalk/pull/26765

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -148,12 +148,19 @@ After do |scenario|
       print_server_logs
     end
   end
-  page.instance_variable_set(:@touched, false) if Capybara::Session.instance_created?
+  page.instance_variable_set(:@touched, false) if capybara_session_created?
 end
 
-# Test is web session is open
+# Test if capybara session instance was created
+def capybara_session_created?
+  return false if Capybara::Session.nil?
+
+  Capybara::Session.instance_created?
+end
+
+# Test if web session is open
 def web_session_is_active?
-  return false unless Capybara::Session.instance_created?
+  return false unless capybara_session_created?
 
   page.has_selector?('header') || page.has_selector?('#username-field')
 end
@@ -241,7 +248,7 @@ After('@scope_cobbler') do |scenario|
 end
 
 AfterStep do
-  next unless Capybara::Session.instance_created?
+  next unless capybara_session_created?
 
   log 'Timeout: Waiting AJAX transition' if has_css?('.senna-loading', wait: 0) && !has_no_css?('.senna-loading', wait: 30)
 end


testsuite/features/support/env.rb
```diff
@@ -148,12 +148,19 @@ def capybara_register_driver
       print_server_logs
     end
   end
-  page.instance_variable_set(:@touched, false) if Capybara::Session.instance_created?
+  page.instance_variable_set(:@touched, false) if capybara_session_created?
```
srbarrios, line 151:
```suggestion
  page.instance_variable_set(:@touched, false) if !Capybara::Session.nil? && Capybara::Session.instance_created?
```

testsuite/features/support/env.rb
```diff
@@ -148,12 +148,19 @@ def capybara_register_driver
...
+def capybara_session_created?
+  return false if Capybara::Session.nil?
+
+  Capybara::Session.instance_created?
+end
+
+# Test if web session is open
 def web_session_is_active?
-  return false unless Capybara::Session.instance_created?
+  return false unless capybara_session_created?
```
srbarrios, line 163:
```suggestion
  return false unless !Capybara::Session.nil? && Capybara::Session.instance_created?
```

testsuite/features/support/env.rb
```diff
@@ -241,7 +248,7 @@ def process_code_coverage
 end
 
 AfterStep do
-  next unless Capybara::Session.instance_created?
+  next unless capybara_session_created?
```
srbarrios, line 251:
```suggestion
  next unless !Capybara::Session.nil? && Capybara::Session.instance_created?
```

testsuite/features/support/env.rb
Bischoff: * It's not really legible
* It makes us repeat ourselves
* It will not work, if I read this correctly:

Ruby Short-Circuit Evaluation
When Ruby evaluates expressions containing boolean operators, it uses short-circuit evaluation. With ||, if the expression on the left evaluates to true, it will return true. Otherwise, it will check if the expression on the right evaluates to true. If so, the expression returns true; otherwise, it will return false.

With &&, both the expression on the left and the expression on the right have to evaluate to true in order to return true. If either expression is false, it will return false

srbarrios: :) Happy to hear that you prefer encapsulation ;P

Bischoff:
> Your PR LGTM. Still I shared a suggested code.

Yeah, I considered short-circuit evaluation, but I preferred a dedicated method. More legible, scales better.

testsuite/features/support/env.rb
Bischoff: I knew you would say that :smile_cat: .

Like all good things, it depends...

testsuite/features/support/env.rb
Bischoff: See above

testsuite/features/support/env.rb
Bischoff: See above

PR #10042, opened 2025-03-24
Improve CLM build time through query optimization (bsc#1235847)

wweellddeerr:
## What does this PR change?

The query for `suseServerAppStreamHiddenPackagesView` becomes slow as the number of records in the AppStreams-related tables increases, causing delays in the CLM build. This change replaces a `LEFT JOIN` with a `NOT EXISTS` clause to optimize the query's response time.

It also includes a new index for `suseAppstreamPackage(module_id)`

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: SQL fix.

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/26205

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/schema/spacewalk/common/tables/suseAppstreamPackage.sql
+++ b/schema/spacewalk/common/tables/suseAppstreamPackage.sql
@@ -23,3 +23,5 @@ CREATE TABLE suseAppstreamPackage(
     modified        TIMESTAMPTZ DEFAULT (current_timestamp) NOT NULL,
     CONSTRAINT uq_as_pkg_module UNIQUE (package_id, module_id)
 );
+
+CREATE INDEX suse_appstream_pkg_module_id_idx ON suseAppstreamPackage(module_id);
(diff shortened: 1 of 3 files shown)


cbbayburt (approved):
I can't tell how this change impacts performance by looking at the PR, but the code LGTM :+1:

Hopefully, it'll be enough to satisfy the customer.

rjmateus (dismissed):
LGTM
Do you have any explain plan showing both approach? From this change I think is already good that we avoid one more join, and replace it with the not exists will increase speed.
What I'm not sure is if the existing indexes are enough, since there are not index looking for the server_id. To know if the current index is enough I would need to have a look in the explain plan

rjmateus:
>One of the customers has already reported that the execution time was reduced from 8 hours to 2 hours after applying this change. 
This is already great news. At least it became usable again.

We can analyze the explain plan together and see if we would need some more indexes or refactoring. Let me know when you have it.

PR #10062, opened 2025-03-26
Sonar Cloud miscellaneous fixes 2

CDellaGiusta:
## What does this PR change?
Fixes the following SonarCloud issues, where the fixes are straightforward
After review and approval, they will be squashed into one commit only

SonarCloud fix: java:S1858 toString() should never be called on a String object
SonarCloud fix: java:S1158 Use static primitive wrapper class toString() method
SonarCloud fix: java:S1185 Overriding methods should do more than simply call the same method in the super class
SonarCloud fix: java:S1656 Remove or correct useless self-assignment
SonarCloud fix: java:S1157 Replace toUpperCase()/toLowerCase() and equals() calls with a single equalsIgnoreCase() call
SonarCloud fix: java:S1130 Remove the declaration of thrown exception as it cannot be thrown from method body
SonarCloud fix: java:S2209 Static members should be accessed statically
SonarCloud fix: java:S2097 equals(Object obj) should test the argument type

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): 
Port(s): 
- [x] **DONE**

## Changelogs
- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/reactor/SaltEvent.java
+++ b/java/code/src/com/redhat/rhn/domain/reactor/SaltEvent.java
@@ -70,7 +70,7 @@ public class SaltEvent {
      * @param minionIdIn the new id
      */
     public void setMinionId(String minionIdIn) {
-        this.minionId = minionId;
+        this.minionId = minionIdIn;
     }
 
     /**
--- a/java/code/src/com/suse/scc/model/SCCMinProductJson.java
+++ b/java/code/src/com/suse/scc/model/SCCMinProductJson.java
@@ -37,11 +37,11 @@ public class SCCMinProductJson {
      */
     public SCCMinProductJson(long idIn, String identifierIn, String versionIn,
             String releaseTypeIn, String archIn) {
-        this.id = id;
-        this.identifier = identifier;
-        this.version = version;
-        this.releaseType = releaseType;
-        this.arch = arch;
+        this.id = idIn;
+        this.identifier = identifierIn;
+        this.version = versionIn;
+        this.releaseType = releaseTypeIn;
+        this.arch = archIn;
     }
 
     /**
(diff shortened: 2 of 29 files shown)


cbosdo (approved):

mcalmer (approved):

java/code/src/com/redhat/rhn/domain/reactor/SaltEvent.java
```diff
@@ -70,7 +70,7 @@ public String getMinionId() {
      * @param minionIdIn the new id
      */
     public void setMinionId(String minionIdIn) {
-        this.minionId = minionId;
+        this.minionId = minionIdIn;
```
mackdk, line 73: Ouch, this was pretty bad. I would guess this setter was never used, otherwise it would have been a problem. 

I would evaluate if it's possible to remove all the setters and turn this pojo into an immutable object.

java/code/src/com/suse/scc/model/SCCMinProductJson.java
```diff
@@ -37,11 +37,11 @@ public class SCCMinProductJson {
...
-        this.id = id;
-        this.identifier = identifier;
-        this.version = version;
-        this.releaseType = releaseType;
-        this.arch = arch;
+        this.id = idIn;
+        this.identifier = identifierIn;
+        this.version = versionIn;
+        this.releaseType = releaseTypeIn;
+        this.arch = archIn;
```
mackdk, line 44: Looks like this constructor was never used either luckily, maybe we can just remove it.

mackdk (approved):

rjmateus (approved):
LGTM. Thank you for the cleanup

PR #10109, opened 2025-04-02
RBAC UI/API implementation

cbbayburt:
RBAC API introduces the `access` API namespace and updates the `user` namespace to manage RBAC roles.

## TODO
 - [x] WebUI permission setup
 - [x] Unit tests
 - [x] Logging
 - [x] User migrations
 - [x] Map latest endpoints merged recently
 
### Before
Screenshot From 2025-04-03 19-30-21

### After
Screenshot From 2025-04-03 19-30-16

## Documentation
- API docs were added

## Test coverage
TODO

- [ ] **DONE**

## Links

Fixes https://github.com/SUSE/spacewalk/issues/26623
Fixes https://github.com/SUSE/spacewalk/issues/26625

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


java/code/src/com/redhat/rhn/domain/access/AccessGroupFactory.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * Lists all access groups that belong to an org.
+     * @param org the org
+     * @return the list of access groups
+     */
+    public static List<AccessGroup> list(Org org) {
+        return getSession()
+                .createQuery("SELECT a FROM AccessGroup a WHERE a.org = :org OR a.org IS NULL", AccessGroup.class)
```
rjmateus, line 79: This query is not really returning all groups that belong to an Org. In fact it returns all groups from the org or the default ones (not linked to any org).

schema/spacewalk/common/tables/accessGroup.sql
```diff
@@ -20,5 +20,5 @@ CREATE TABLE access.accessGroup (
 );
 COMMENT ON TABLE access.accessGroup IS 'Access groups (roles) for RBAC';
 
-CREATE UNIQUE INDEX access_group_label_uq
-ON access.accessGroup(label);
+CREATE UNIQUE INDEX access_group_org_label_uq
+ON access.accessGroup(org_id, label);
```
rjmateus, line 24: If I remember correctly the org_id can be null. How will this unique index will behave? 
It will allow to have the same label at different organizations, but will those names collide with the default group labels?
For example, if we have a default group name "imageAdmin", with org_id null, can the user create new group with the same label assigned to an org?
I think we must prevent this.

rjmateus (commented):
First round is done. I didn't review all the files bu I will try to make it later.

schema/spacewalk/common/tables/accessGroup.sql
cbbayburt: I can do it with an extra check constraint on reserved groups.

java/code/src/com/redhat/rhn/domain/access/AccessGroupFactory.java
cbbayburt: Yes, it returns all the groups available to an org, including the defaults, which are collectively owned by every org.

java/code/src/com/redhat/rhn/frontend/xmlrpc/user/UserHandler.java
```diff
@@ -318,15 +349,29 @@ private int  modifySatAdminRole(User loggedInUser, String login, boolean grant)
...
+        // RBAC roles
+        Stream<AccessGroup> rbacRoles;
+        if (user.hasRole(RoleFactory.SAT_ADMIN)) {
+            // Satellite admin can assign any role in any org
+            rbacRoles = accessGroupManager.list().stream();
         }
-        return assignableRoles;
+        else if (user.hasRole(RoleFactory.ORG_ADMIN)) {
+            // Org admin can assign any role in the same org
+            rbacRoles = accessGroupManager.list(user.getOrg()).stream();
```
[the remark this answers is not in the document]
cbbayburt: I didn't understand this one.

Here it works like this:

1. If you're a sat_admin, this will return every role in MLM including all RBAC roles
2. If you're an org_admin, this will return `org_admin` + RBAC default groups + custom groups in org
3. If you're a regular user, according to the next comment, I'll return an empty set

Does this sound correct?

schema/spacewalk/common/tables/accessGroup.sql
cbbayburt: Fixed in `AccessGroupManager`

java/code/src/com/redhat/rhn/manager/access/AccessGroupManager.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Class to manage RBAC access groups
+ */
+public class AccessGroupManager {
+
+    private static final Logger LOG = LogManager.getLogger(AccessGroupManager.class);
```
rjmateus, line 37: The LOG variable is never used. It would be good to add several debug data in case we need to activate it and check what is happening.

java/code/src/com/redhat/rhn/manager/access/AccessGroupManager.java
```diff
@@ -0,0 +1,195 @@
...
+    public AccessGroup create(String label, String description, Org org) throws DefaultRoleException {
+        return this.create(label, description, org, Collections.emptyList());
+    }
+
+    /**
+     * Creates a new access group with the given parameters.
+     * @param label the label of the group
+     * @param description the description of the group
+     * @param org the org to which the group belongs
+     * @param inheritFrom a collection of {@code AccessGroup}s to inherit permissions from
```
rjmateus, line 55: I would not call it `inheritFrom` since it will not create a persistent relationship. I think the correct naming would be `copyFrom`.

java/code/src/com/redhat/rhn/manager/access/AccessGroupManager.java
```diff
@@ -0,0 +1,195 @@
...
+     * @return an {@code Optional} containing the access group, or an empty {@code Optional} if not found
+     */
+    public Optional<AccessGroup> lookup(String label, Org org) {
+        return AccessGroupFactory.lookupByLabelAndOrg(label, org);
+    }
+
+    /**
+     * Grants access to the given namespaces for the specified access group.
+     * <p>
+     * The asterisk (*) character can be used as a wildcard in namespace strings.
```
rjmateus, line 117: I think someone in the presentaton asked about the need for the `*` and if we should just remove it and state that anything in that namespace and "children" will be added. The reality is that in the query processing you are just replacing the `*` with `%`. By not having it in the input you could just concatenate the `%` at the end.

java/code/src/com/redhat/rhn/frontend/action/user/AdminUserEditAction.java
```diff
@@ -206,6 +211,30 @@ else if (roleSetting == null && existingRoles.contains(role.getLabel())) {
...
+    private void processRBACGroupAssignments(HttpServletRequest request, User target) {
+        var userGroups = target.getAccessGroups();
+        var currentGroupLabels = userGroups.stream()
+                .map(AccessGroup::getLabel)
+                .collect(Collectors.toUnmodifiableSet());
+
+        Iterator<AccessGroup> iterator = userGroups.iterator();
+        while (iterator.hasNext()) {
+            AccessGroup group = iterator.next();
+            String groupSetting = request.getParameter(ROLE_SETTING_PREFIX + group.getLabel());
```
rjmateus, line 223: I'm not sure if I fully understand this `ROLE_SETTING_PREFIX`, but this is old school code (17 year old) running on struts. If it worked I'm fine with it :)

rjmateus: In the meantime I think I found it in the `userdetails.jsp`, where the checkbox is populated with the role value and a prefix `role_` 
I keeped the comment so you can confirm if that is the case :)

java/code/src/com/redhat/rhn/frontend/xmlrpc/access/AccessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+     * @return 1 if successful
+     *
+     * @apidoc.doc Grant full access to the given namespace for the specified role.
+     * Return the expanded list of namespaces granted by the call.
+     * @apidoc.param #session_key()
+     * @apidoc.param #param("string", "label", "the unique label of the role")
+     * @apidoc.param #array_single_desc("string", "namespaces", "the list of namespaces to grant access to")
+     * @apidoc.returntype #return_int_success()
+     */
+    public int grantAccess(User loggedInUser, String label, List<String> namespaces) {
```
rjmateus, line 194: Should we have one method to grant read-only access to the namespace also?

rjmateus (commented):
New round. The only missing part to review is the SQL, but I will let that for tomorrow. So far look good to me. Just some small nitpicks.

rjmateus (commented):
I didn't test the full sql files, and I only made some general test. I still want to perform a full test for the data and schema upgrade files, but that will take a few more time.


On the file `schema/spacewalk/common/data/endpoint.sql`:
  - the initial inserts sets auth_require as a boolean (True/False) but the next ones are using the char `t`. It should work but would be great to keep it coherent.
  - You have some of the files with the option `on conflict`. In this initial data I think we should not expect any conflict. Is this needed? My concer is that it may be hiding some conflicts.

  File `schema/spacewalk/common/data/namespace.sql` also have all inserts with `On Conflict` clause. Is this needed or do we have the risk of hidden conflict?

  This PR are craeting the default access groups but are not asssigning any access. Is this expected and will be done in a folow up pr?

cbbayburt:
Thanks for the review, Ricardo.

> the initial inserts sets auth_require as a boolean (True/False) but the next ones are using the char t. It should work but would be great to keep it coherent.

Yes, I also realized it afterwards. I'll do the change.

> You have some of the files with the option on conflict. In this initial data I think we should not expect any conflict. Is this needed? My concer is that it may be hiding some conflicts.

I actually did it in `common/data` so that when we're are adding more data on top in the future, we don't have to refer to older namespace entries, but I'm not sure about it.

> This PR are craeting the default access groups but are not asssigning any access. Is this expected and will be done in a folow up pr?

I'm currently working on it. I'll push it in another commit as soon as I'm done with some corner cases.

java/code/src/com/redhat/rhn/frontend/action/user/AdminUserEditAction.java
cbbayburt: This comes from Welder's PR, but he's not around to explain it. It works with no problem, so I'd not touch it :D

java/code/src/com/redhat/rhn/manager/access/AccessGroupManager.java
cbbayburt: The reason I didn't want to do it is this case:

Imagine there are 3 namespaces like following:
 - foo
 - foo.baz
 - foobar.baz
 
 If `%` is added by default, a query with `foo` will return `foo`, `foo.baz`, `foobar.baz`.
 
 By making it explicit, the user will have control over it:
 
  - `foo` would return only `foo`
  - `foo.*` would return `foo.baz`
  - `foo*` would return `foo`, `foo.baz` and `foobar.baz`
  
  I made sure that this problem doesn't appear with the initial data, but since we'll be adding namespaces constantly, it's not easy to prevent this in the future, so I think making `*` explicit is a safer way.

java/code/src/com/redhat/rhn/frontend/action/user/AdminUserEditAction.java
rjmateus: if it works don't touch it :)

java/code/src/com/redhat/rhn/manager/access/AccessGroupManager.java
rjmateus: Makes sense to me. If users don't like it or don't understand it we can always revisit it.

java/code/src/com/redhat/rhn/frontend/action/channel/manage/ChannelPackageMenuAction.java
```diff
@@ -49,7 +50,7 @@ public ActionForward execute(ActionMapping mapping,
         long cid = requestContext.getRequiredParam("cid");
 
         Channel chan = ChannelFactory.lookupByIdAndUser(cid, user);
-        if (!user.hasRole(RoleFactory.CHANNEL_ADMIN) &&
+        if (!user.isMemberOf(AccessGroupFactory.CHANNEL_ADMIN) &&
                 !UserManager.verifyChannelAdmin(user, chan)) {
             throw new PermissionCheckFailureException(RoleFactory.CHANNEL_ADMIN);
```
rjmateus, line 55: Is just a nitpick, but in here is still being used the old Channel_Admin role

java/code/src/com/redhat/rhn/frontend/action/configuration/files/ManageRevisionSubmit.java
```diff
@@ -303,6 +306,7 @@ public int deleteRevision(RhnSetElement elementIn, User userIn) {
     private void checkAcl(RequestContext requestContext) {
         //Throws an exception if the user does not have permission to edit the channel.
         User user = requestContext.getCurrentUser();
+        ensureRoleBasedAccess(user, "config.files", Namespace.AccessMode.W);
```
rjmateus, line 309: Shouldn't this be done on the general web filter? Or this is one of those cases that reading and writing are mixed together?

java/code/src/com/redhat/rhn/manager/user/UserManager.java
```diff
@@ -735,6 +736,20 @@ public static boolean hasRole(Long uid, Role label) {
         return UserFactory.lookupById(uid).hasRole(label);
     }
 
+    /**
+     * Check RBAC memberships of the specified user.
+     * @param uid The id of the user to lookup.
+     * @param group the access group to check.
+     * @return {@code True} if the user is a member of the access group.
+     */
+    public static boolean isMemberOf(Long uid, AccessGroup group) {
```
rjmateus, line 745: In this file you should also check the method "verifyChannelAdmin", which I think is still using the old roles.

java/code/src/com/redhat/rhn/manager/channel/ChannelManager.java
```diff
@@ -2427,7 +2428,7 @@ public static Long lookupOriginalId(Channel channel) {
      * @return list of errataOverview objects that need to be resynced
      */
     public static List<ErrataOverview> listErrataNeedingResync(Channel c, User user) {
```
rjmateus, line 2430: I think the method `removeErrata`  is missing the same fix

java/code/src/com/redhat/rhn/domain/kickstart/builder/KickstartBuilder.java
```diff
@@ -368,14 +366,6 @@ private void storeScript(String prefix, KickstartData ksData, StringBuilder buf,
 
     }
 
-    private void checkRoles() {
```
rjmateus, line 371: Shouldn't we check in here that the user is part of these groups? Or this is not needed anymore?

schema/spacewalk/common/views/suseChannelUserRoleView.sql
```diff
@@ -64,8 +66,14 @@ AS
...
           JOIN rhnUserGroup g ON g.id = m.user_group_id
           JOIN rhnUserGroupType t ON t.id = g.group_type
-        WHERE t.label = 'channel_admin' OR t.label = 'org_admin') adminUsers
+        WHERE t.label = 'org_admin') adminUsers
      ON (adminUsers.user_id = combination.user_id)
+     LEFT JOIN
+       (SELECT DISTINCT uag.user_id, 1 AS is_channel_admin
+        FROM access.userAccessGroup uag
+          JOIN access.accessGroup ag ON ag.id = uag.group_id
+        WHERE ag.label = 'channel_admin') channelAdmins
```
rjmateus, line 75: Just as a double check, we should also verify that the org for the access group is null.

java/code/src/com/redhat/rhn/frontend/action/schedule/FailedSystemsAction.java
```diff
@@ -136,6 +139,7 @@ public ActionForward handleDispatch(ListSessionSetHelper helper,
                                         ActionForm formIn, HttpServletRequest request,
                                         HttpServletResponse response) {
 
+        ensureRoleBasedAccess(new RequestContext(request).getCurrentUser(), "schedule.details", Namespace.AccessMode.W);
```
rjmateus, line 142: Shouldn't this one be handled by the general web filter already? I don't get what this have in special (some more like this one on the PR, but I will not comment all)

rjmateus: Ignore my comment, I got it. I was looking at the wrong place.
You put the check for `w` permission on the `handleDispatch` dispatch method, which is responsible for writing. This is one those cases where the read and write are in the same class.

java/code/src/com/redhat/rhn/frontend/action/configuration/files/ManageRevisionSubmit.java
cbbayburt: Exactly, only upload and delete actions require W mode.

java/code/src/com/redhat/rhn/manager/channel/ChannelManager.java
cbbayburt: That one's also fixed in `suseChannelUserRoleView` view

java/code/src/com/redhat/rhn/domain/kickstart/builder/KickstartBuilder.java
cbbayburt: The same logic is enforced with RBAC filters now.

nodeg:
I always see
```bash
Failing Scenarios:
cucumber features/core/srv_user_preferences.feature:16 # Scenario: Change page size to 100 per page in testing user
```
failing. 
```cucumber
  Scenario: Change page size to 100 per page in testing user
    Given I am authorized as "testing" with password "testing"
    When I follow the left menu "Home > My Preferences"
    And I select "100" from "pagesize"
    And I click on "Save Preferences"
    Then I should see a "Preferences modified" text
```
I think this is related to the access rights for the `testing` user and RBAC.

cbbayburt:
@nodeg yes, there's one particular table not populated during schema migration and that's the cause. I'm debugging it now.

deneb-alpha:
just for tracking... the new PR is https://github.com/uyuni-project/uyuni/pull/10166

PR #10140, opened 2025-04-07
QE: Modifiying some problems in checking and unchecking checkboxes

Pablogoliva:
## What does this PR change?

This PR introduces a new method for checking and unchecking checkboxes to address an issue encountered on the `min_change_software_channels.feature` screen. The problem occurs when multiple checkboxes share the same base name with differing suffixes.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed.

- [x] **DONE**

## Test coverage
- Cucumber tests were edited

- [x] **DONE**

## Links

Issue(s): None, something found during the rrtg tasks
Port(s): 
 - Manager 5.0:
 - Manager 4.3:

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/secondary/min_change_software_channel.feature
+++ b/testsuite/features/secondary/min_change_software_channel.feature
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
     Given I am authorized
 
 @susemanager
-  Scenario: Check the system is still subscribed to old channels before channel change completes
+  Scenario: Pre-requisite: unsubscribe from old channels
     Given I am on the Systems overview page of this "sle_minion"
     When I follow "Software" in the content area
     And I follow "Software Channels" in the content area
     And I wait until I do not see "Loading..." text
+    And I check radio button "SLE-Product-SLES15-SP4-Pool for x86_64"
     Then radio button "SLE-Product-SLES15-SP4-Pool for x86_64" should be checked
     And I wait until I do not see "Loading..." text
+    When I uncheck "SLE15-SP4-Installer-Updates for x86_64"
     And I should see "SLE15-SP4-Installer-Updates for x86_64" as unchecked
+    And I click on "Next"
+    Then I should see a "Confirm Software Channel Change" text
+    When I click on "Confirm"
+    Then I should see a "Changing the channels has been scheduled." text
 
 @uyuni
-  Scenario: Check the system is still subscribed to old channels before channel change completes
+  Scenario: Pre-requisite: unsubscribe from old channels
     Given I am on the Systems overview page of this "sle_minion"
     When I follow "Software" in the content area
     And I follow "Software Channels" in the content area
     And I wait until I do not see "Loading..." text
+    And I check radio button "openSUSE Leap 15.6 (x86_64)"
     Then radio button "openSUSE Leap 15.6 (x86_64)" should be checked
-    And I wait until I do not see "Loading..." text
-    And I should see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" as unchecked
+    When I wait until I do not see "Loading..." text
+    And I uncheck "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" by label
+    Then I should see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" as unchecked
+    When I uncheck "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64) (Development)" by label
+    Then I should see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64) (Development)" as unchecked
+    When I click on "Next"
+    Then I should see a "Confirm Software Channel Change" text
+    When I click on "Confirm"
+    Then I should see a "Changing the channels has been scheduled." text
 
 @susemanager
-  Scenario: Check via API old channels are still the same on the system before channel change completes
+  Scenario: Pre-requisite: check via API that the system is unsubscribed from old channels
     When I refresh the metadata for "sle_minion"
     Then channel "SLE-Product-SLES15-SP4-Pool for x86_64" should be enabled on "sle_minion"
     And channel "SLE15-SP4-Installer-Updates for x86_64" should be disabled on "sle_minion"
 
 @uyuni
-  Scenario: Check via API old channels are still the same on the system before channel change completes
+  Scenario: Pre-requisite: check via API that the system is unsubscribed from old channels
     When I refresh the metadata for "sle_minion"
     Then channel "openSUSE Leap 15.6 (x86_64)" should be enabled on "sle_minion"
     And channel "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" should be disabled on "sle_minion"
@@ -69,11 +83,12 @@ Feature: Assign child channel to a system
     And I follow "Software Channels" in the content area
     And I wait until I do not see "Loading..." text
     Then radio button "openSUSE Leap 15.6 (x86_64)" should be checked
-    And I wait until I do not see "Loading..." text
-    And I check "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)"
-    And I click on "Next"
+    When I wait until I do not see "Loading..." text
+    And I check "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" by label
+    Then I should see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" as checked
+    When I click on "Next"
     Then I should see a "Confirm Software Channel Change" text
-    And I click on "Confirm"
+    When I click on "Confirm"
     Then I should see a "Changing the channels has been scheduled." text
     When I follow "scheduled" in the content area
     And I wait until I see "1 system successfully completed this action." text, refreshing the page
@@ -142,13 +157,15 @@ Feature: Assign child channel to a system
     Then radio button "openSUSE Leap 15.6 (x86_64)" should be checked
     And I wait until I do not see "Loading..." text
     And I wait until I see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" text
-    And I uncheck "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)"
-    And I check "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64) (Development)"
-    And I check "Fake-RPM-SUSE-Channel"
+    And I uncheck "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" by label
+    Then I should see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" as unchecked
+    When I check "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64) (Development)" by label
+    Then I should see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64) (Development)" as checked
+    When I check "Fake-RPM-SUSE-Channel"
     And I click on "Next"
     Then I should see a "Confirm Software Channel Change" text
     When I click on "Confirm"
-    Then I should see a "Changing the channels has been scheduled." text
-    When I follow "scheduled" in the content area
+    And I wait until I see "Changing the channels has been scheduled." text
+    And I follow "scheduled" in the content area
     And I wait until I see "1 system successfully completed this action." text, refreshing the page
     Then channel "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64)" should be disabled on "sle_minion"
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -161,6 +161,17 @@ When(/^I uncheck "([^"]*)"$/) do |identifier|
   raise ScriptError, "Checkbox #{identifier} not unchecked." if has_checked_field?(identifier)
 end
 
+When(/^I (check|uncheck) "([^"]*)" by label$/) do |action, label|
+  checkbox = find(:xpath, "//label[text()='#{label}']/preceding-sibling::input[@type='checkbox']")
+  if action == 'check'
+    checkbox.set(true)
+    raise ScriptError, "Checkbox #{label} not checked." unless has_checked_field?(label)
+  elsif action == 'uncheck'
+    checkbox.set(false)
+    raise ScriptError, "Checkbox #{label} not unchecked." unless has_unchecked_field?(label)
+  end
+end
+
 When(/^I select "([^"]*)" from "([^"]*)"$/) do |option, field|
   if has_select?(field, with_options: [option], wait: 1)
     select(option, from: field)


testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -152,13 +152,35 @@
 # Check a checkbox of the given id
 #
 When(/^I check "([^"]*)"$/) do |identifier|
-  check(identifier)
-  raise ScriptError, "Checkbox #{identifier} not checked." unless has_checked_field?(identifier)
+  begin
+    contenedor = find('div', text: identifier, exact_text: true)
```
nodeg, line 156: What does `contenedor` mean? I assume it is Spanish, so please use the english translation :wink:

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -152,13 +152,47 @@
 # Check a checkbox of the given id
 #
 When(/^I check "([^"]*)"$/) do |identifier|
```
srbarrios, line 154: I fear, we will need a new step for that particular implementation, keeping the other generic one.

Here the RubyDoc https://www.rubydoc.info/gems/capybara/Capybara/Node/Actions#check-instance_method

As you can see the check(id) tries id, name, and then associated label text.
But this proposed implementation is very specific for "text" associated.

So, I proposed you move this implementation to a new step:
```
When(/^I check "([^"]*)" by label text$/) do |text|
```

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -152,13 +152,53 @@
...
+  if has_unchecked_field?(identifier)
+    if checkbox_id.nil?
+      check(identifier)
+    else
+      check(identifier, id: checkbox_id)
+    end
+    sleep(0.5)
+    raise ScriptError, "Checkbox #{identifier} not checked." if has_unchecked_field?(identifier)
+  else
+    log "Checkbox:#{identifier} is already checked"
```
srbarrios, line 174: What about just that?

```suggestion
    checkbox = find(:xpath, "//div[. = '#{text}']//input[@type='checkbox']")
    checkbox.set(true)
  rescue StandardError => e
    log "Error message: #{e.message}"
```

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: Done

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: Done

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
     Given I am authorized
 
 @susemanager
-  Scenario: Check the system is still subscribed to old channels before channel change completes
+  Scenario: Pre-requisite: unsubscribed to old channels
```
ktsamis, line 19:
```suggestion
  Scenario: Pre-requisite: Unsubscribe from old channels
```

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
     Given I am on the Systems overview page of this "sle_minion"
     When I follow "Software" in the content area
     And I follow "Software Channels" in the content area
     And I wait until I do not see "Loading..." text
+    And I check radio button "SLE-Product-SLES15-SP4-Pool for x86_64"
```
ktsamis, line 24: This needs to be removed no? Same as next line

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
+    When I uncheck "SLE15-SP4-Installer-Updates for x86_64"
     And I should see "SLE15-SP4-Installer-Updates for x86_64" as unchecked
+    And I click on "Next"
+    Then I should see a "Confirm Software Channel Change" text
+    When I click on "Confirm"
+    Then I should see a "Changing the channels has been scheduled." text
 
 @uyuni
-  Scenario: Check the system is still subscribed to old channels before channel change completes
+  Scenario: Pre-requisite: unsubscribed to old channels
```
ktsamis, line 35:
```suggestion
  Scenario: Pre-requisite: Unsubscribe from old channels
```

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
     Given I am on the Systems overview page of this "sle_minion"
     When I follow "Software" in the content area
     And I follow "Software Channels" in the content area
     And I wait until I do not see "Loading..." text
+    And I check radio button "openSUSE Leap 15.6 (x86_64)"
```
ktsamis, line 40: Remove?

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
+    When I uncheck "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64) (Development)" by label
+    Then I should see "Uyuni Client Tools for openSUSE Leap 15.6 (x86_64) (Development)" as unchecked
+    When I click on "Next"
+    Then I should see a "Confirm Software Channel Change" text
+    When I click on "Confirm"
+    Then I should see a "Changing the channels has been scheduled." text
 
 @susemanager
-  Scenario: Check via API old channels are still the same on the system before channel change completes
+  Scenario: Pre-requisite: check via API that system is unsubscribed from old channels
```
ktsamis, line 53:
```suggestion
  Scenario: Pre-requisite: Check via API that the system is unsubscribed from old channels
```

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
     When I refresh the metadata for "sle_minion"
     Then channel "SLE-Product-SLES15-SP4-Pool for x86_64" should be enabled on "sle_minion"
     And channel "SLE15-SP4-Installer-Updates for x86_64" should be disabled on "sle_minion"
 
 @uyuni
-  Scenario: Check via API old channels are still the same on the system before channel change completes
+  Scenario: Pre-requisite: check via API that system is unsubscribed from old channels
```
ktsamis, line 59:
```suggestion
  Scenario: Pre-requisite: Check via API that the system is unsubscribed from old channels
```

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
+    When I click on "Next"
+    Then I should see a "Confirm Software Channel Change" text
+    When I click on "Confirm"
+    Then I should see a "Changing the channels has been scheduled." text
 
 @susemanager
-  Scenario: Check via API old channels are still the same on the system before channel change completes
+  Scenario: Pre-requisite: check via API that system is unsubscribed from old channels
     When I refresh the metadata for "sle_minion"
     Then channel "SLE-Product-SLES15-SP4-Pool for x86_64" should be enabled on "sle_minion"
```
ktsamis, line 55: Should that be enabled or not? I think if it's the base channel it should be there but just checking

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
     And channel "SLE15-SP4-Installer-Updates for x86_64" should be disabled on "sle_minion"
 
 @uyuni
-  Scenario: Check via API old channels are still the same on the system before channel change completes
+  Scenario: Pre-requisite: check via API that system is unsubscribed from old channels
     When I refresh the metadata for "sle_minion"
     Then channel "openSUSE Leap 15.6 (x86_64)" should be enabled on "sle_minion"
```
ktsamis, line 61: Should that be enabled or not? I think if it's the base channel it should be there but just checking

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -161,6 +161,31 @@
   raise ScriptError, "Checkbox #{identifier} not unchecked." if has_checked_field?(identifier)
 end
 
+When(/^I (check|uncheck) "([^"]*)" by label$/) do |action, identifier|
+  begin
+    checkbox = find(:xpath, "//input[@type='checkbox' and @id = //label[normalize-space(.) = '#{identifier}']/@for]")
+    if action == 'check'
```
ktsamis, line 167: There is no reason for this action == check and then make it into another boolean variable. I would suggest to remove this "translation" and keep checking for the action string below as well in line 182 and 184. It would make it easier to read

ktsamis (changes requested):

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: No because the first one is to check the checkbox for ensuring that it is checked in the configuration of the test environment and the second one is to check that it was checked correctly

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: You are right is unsubscribe from not unsubscribe to. Nice catch. But regarding the capital letter after :, I have seen in other features and we don't do it like this
Screenshot From 2025-04-09 11-24-09

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: No because I want to check the button to ensure that is checked

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -161,6 +161,31 @@
...
+    else
+      raise StandardError, "Variable indicated in step:#{action} not correctly indicated"
+    end
+    checkbox.set(action_boolean)
+  rescue Capybara::ElementNotFound => e
+    puts "ElementNotFound. Error message: #{e.message}"
+  rescue Capybara::Ambiguous => e
+    puts "Element is ambiguous. Error message: #{e.message}"
+  rescue StandardError => e
+    puts "StandardError. Error message: #{e.message}"
```
srbarrios, line 180: If we have errors, we don't want to catch them and just print a log.
If we have an error, we want to raise the error, and handle the issue.
Please remove these rescues.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -161,6 +161,31 @@
   raise ScriptError, "Checkbox #{identifier} not unchecked." if has_checked_field?(identifier)
 end
 
+When(/^I (check|uncheck) "([^"]*)" by label$/) do |action, identifier|
```
srbarrios, line 164: If we un/check by label, let's name the parameter `label` instead of `identifier`

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: Thinking from a pure software development point of view, and maybe is not correct because we are qe not pure software developers. The reason is to avoid writing two times the same key line: `checkbox.set(action_boolean)` because if we repeat the same line two times inside an if-else we can forget modifing it the second time or something like this. To summarize, for trying to not repeat myself in key code but maybe as qe we have to be focused in the readability instead of the quality of the methods.

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: The action here is checking or unchecking checkboxes and the error that should be raised is if the checkbox is checked or unchecked respectively, not the ones that I catch in these try-caths. The ones that I catch here are just to catch errors in the finding of the label, something not relevant for the purpouse of this. The one that really matters is raised bellow.

Pablogoliva: But I should use log instead of put to write the errors, that's right

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: I can do something in between using a ternary operator instead for maintaining the readability meanwhile I don't repeat key things.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -161,6 +161,31 @@
...
+  begin
+    checkbox = find(:xpath, "//input[@type='checkbox' and @id = //label[normalize-space(.) = '#{identifier}']/@for]")
```
srbarrios, line 166:
```suggestion
    checkbox = find(:xpath, ""//label[text()='#{label}']/preceding-sibling::input[@type='checkbox']"")
```

image

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: Nice catch regarding the missing of the 'the'. Somethimes after reading the same to many times the vicious makes you making mistakes hahaha

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: Good idea, I used identifier as and inheritance of the previous check and uncheks steps

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: Clarified in the channel https://suse.slack.com/archives/C02D12TNYLS/p1744192855159329

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: Done

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: Yes but checking it can be useful, not just graphically.

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: Done

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: okey

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: Outdated

testsuite/features/step_definitions/navigation_steps.rb
Pablogoliva: Rubocop doesn't like ternary operator nested, better having an if-else with a key line repeated

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: SLE-Product-SLES15-SP4-Pool for x86_64 should enable and SLE15-SP4-Installer-Updates for x86_64 should be disable

srbarrios (dismissed):

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
     Given I am authorized
 
 @susemanager
-  Scenario: Check the system is still subscribed to old channels before channel change completes
+  Scenario: Pre-requisite: unsubscribed from old channels
```
ktsamis, line 19:
```suggestion
  Scenario: Pre-requisite: check that the system is unsubscribed from the old channels
```

testsuite/features/secondary/min_change_software_channel.feature
```diff
@@ -16,33 +16,47 @@ Feature: Assign child channel to a system
...
+    When I uncheck "SLE15-SP4-Installer-Updates for x86_64"
     And I should see "SLE15-SP4-Installer-Updates for x86_64" as unchecked
+    And I click on "Next"
+    Then I should see a "Confirm Software Channel Change" text
+    When I click on "Confirm"
+    Then I should see a "Changing the channels has been scheduled." text
 
 @uyuni
-  Scenario: Check the system is still subscribed to old channels before channel change completes
+  Scenario: Pre-requisite: unsubscribed from old channels
```
ktsamis, line 35:
```suggestion
  Scenario: Pre-requisite: check that the system is unsubscribed from the old channels
```

ktsamis (changes requested):
small nitpick sorry

testsuite/features/secondary/min_change_software_channel.feature
ktsamis:
```suggestion
  Scenario: Pre-requisite: unsubscribe from old channels
```

testsuite/features/secondary/min_change_software_channel.feature
Pablogoliva: Yes hahaha this one yes hahahaha. Nice catch and thanks for carefully reading this hahaha

Pablogoliva: Done

ktsamis (approved):

PR #10154, opened 2025-04-08
Fix: Correctly register and deregister with multiple roles

CDellaGiusta:
## The first commit is from https://github.com/uyuni-project/uyuni/pull/10148, only the last commit is new

This fix avoid that peripherals already taken (subscribed to a hub) could be registered from a different hub.
Furthermore, it correctly deregister when a server is both hub and peripheral

## GUI diff
Before:
before

After:

- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: manually tested
- [x] **DONE**

## Links
Issue(s): 
Port(s): 
- [x] **DONE**

## Changelogs
- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/ServerInfoJsonSerializer.java
+++ b/java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/ServerInfoJsonSerializer.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2025 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+
+package com.redhat.rhn.frontend.xmlrpc.serializer;
+
+import com.suse.manager.api.ApiResponseSerializer;
+import com.suse.manager.api.SerializationBuilder;
+import com.suse.manager.api.SerializedApiResponse;
+import com.suse.manager.model.hub.ServerInfoJson;
+
+/**
+ * Converts a ServerInfoJson to an XMLRPC &lt;struct&gt;.
+ *
+ * @apidoc.doc
+ *  #struct_begin("server info")
+ *      #prop_desc("boolean", "isHub", "true if server is a hub")
+ *      #prop_desc("boolean", "isPeripheral", "true if server is a peripheral")
+ *  #struct_end()
+ */
+public class ServerInfoJsonSerializer extends ApiResponseSerializer<ServerInfoJson> {
+
+    @Override
+    public Class<ServerInfoJson> getSupportedClass() {
+        return ServerInfoJson.class;
+    }
+
+    @Override
+    public SerializedApiResponse serialize(ServerInfoJson src) {
+        return new SerializationBuilder()
+                .add("isHub", src.isHub())
+                .add("isPeripheral", src.isPeripheral())
+                .build();
+    }
+}
--- a/java/code/src/com/suse/manager/hub/HubController.java
+++ b/java/code/src/com/suse/manager/hub/HubController.java
@@ -44,6 +44,7 @@ import com.suse.manager.model.hub.ManagerInfoJson;
 import com.suse.manager.model.hub.OrgInfoJson;
 import com.suse.manager.model.hub.RegisterJson;
 import com.suse.manager.model.hub.SCCCredentialsJson;
+import com.suse.manager.model.hub.ServerInfoJson;
 import com.suse.manager.model.hub.UpdatableServerData;
 import com.suse.manager.webui.controllers.ECMAScriptDateAdapter;
 import com.suse.manager.webui.utils.token.TokenBuildingException;
@@ -104,6 +105,7 @@ public class HubController {
         post("/hub/ping", asJson(usingTokenAuthentication(this::ping)));
         post("/hub/sync/deregister", asJson(usingTokenAuthentication(onlyFromRegistered(this::deregister))));
         post("/hub/sync/registerHub", asJson(usingTokenAuthentication(onlyFromUnregistered(this::registerHub))));
+        get("/hub/serverInfo", asJson(usingTokenAuthentication(this::getServerInfo)));
         post("/hub/sync/replaceTokens", asJson(usingTokenAuthentication(onlyFromHub(this::replaceTokens))));
         post("/hub/sync/storeCredentials", asJson(usingTokenAuthentication(onlyFromHub(this::storeCredentials))));
         post("/hub/sync/setHubDetails", asJson(usingTokenAuthentication(onlyFromHub(this::setHubDetails))));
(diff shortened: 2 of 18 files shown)


mcalmer (commented):
The code is ok, but I think we should now have a changelog entry

mcalmer (approved):

mackdk (dismissed):
LGTM. Added some discussion points.

mcalmer (dismissed):
I am not so happy anbout the API name "canBeRegisteredAsPeripheral", but 
I have no better idea. So it is ok for me

mackdk (approved):

java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/ServerInfoJsonSerializer.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public Class<ServerInfoJson> getSupportedClass() {
+        return ServerInfoJson.class;
+    }
+
+    @Override
+    public SerializedApiResponse serialize(ServerInfoJson src) {
+        return new SerializationBuilder()
+                .add("registeredToHub", src.isRegisteredToHub())
+                .add("withPeripherals", src.isWithPeripherals())
```
mcalmer, line 39: I like more to say either:

- hasHub
- hasPeripherals

or when we want to say what this server is to turn this around with:

- isPeripheral
- isHub

From my point of view this is easier to understand when you use the API. 
As we use this as API my feeling is the second option is better as we want to know what this server is and not what objects we have in the DB.

CDellaGiusta: @mcalmer What I want to know is if the server I'm trying to register (as peripheral) has already a hub, and if it has already some peripherals subscribed to it.
When you say the server "isHub" you actually mean "hasPeripherals", and when you say "isPeripheral" you actually mean "hasHub", so the two meanings are exchanged.
With this in mind, are you still trending to the "is" approach?

mcalmer: This is what i mean. I have a tendency to "is" approach, but the other one is also fine. Maybe we get more opinions.

CDellaGiusta: I'll go with the "is" approach

java/code/src/com/suse/manager/model/hub/ServerInfoJson.java
```diff
@@ -0,0 +1,70 @@
...
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+
+package com.suse.manager.model.hub;
+
+import java.util.Objects;
+
+public class ServerInfoJson {
```
mackdk, line 16: This can be a record.

java/code/src/com/suse/manager/hub/HubController.java
```diff
@@ -105,8 +106,7 @@ public void initRoutes() {
         post("/hub/ping", asJson(usingTokenAuthentication(this::ping)));
         post("/hub/sync/deregister", asJson(usingTokenAuthentication(onlyFromRegistered(this::deregister))));
         post("/hub/sync/registerHub", asJson(usingTokenAuthentication(onlyFromUnregistered(this::registerHub))));
-        get("/hub/sync/canBeRegisteredAsPeripheral",
-                asJson(usingTokenAuthentication(this::canBeRegisteredAsPeripheral)));
+        get("/hub/sync/serverInfo", asJson(usingTokenAuthentication(this::getServerInfo)));
```
mackdk, line 109: Maybe the path should be `/hub/serverInfo` as it's not really specific to sync (like `hub/managerinfo`)?

java/code/src/com/redhat/rhn/frontend/xmlrpc/serializer/ServerInfoJsonSerializer.java
CDellaGiusta: @mcalmer done!

java/code/src/com/suse/manager/hub/HubController.java
CDellaGiusta: @mackdk changed to /hub/serverInfo

mcalmer (dismissed):

cbbayburt (changes requested):
Hi @CDellaGiusta, looks like you're the first one changing an endpoint (`/hub/serverInfo`) since we merged the RBAC feature. From now on, whenever you add or change an endpoint, you'll need to insert its RBAC mappings to the DB.

I still need to write some proper documentation for this, but here's how it works in your case:

The relevant tables endpoint mappings are:
 - [namespace](https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/common/data/namespace.sql)
- [endpoint](https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/common/data/endpoint.sql)
- [endpointNamespace](https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/common/data/endpointNamespace.sql)

In your case, these hub endpoints don't require (RBAC) authorization, so there's no need to tie them to a namespace. Therefore, you only need to add your new endpoint to the `endpoint` table. The existing hub endpoints are already defined there, so you can just copy one and add yours.

Of course, you must also write the same insert as a migration script in `/schema/spacewalk/upgrade` dir.

cbbayburt (approved):
LGTM :+1: Thanks for the change.

About the acceptance test failure: I've recently bumped the timeout for the image build scenario to 660 secs, but it seems it's not always enough.

It looks like we might have to increase it even more.

@Bischoff what do you think?

schema/spacewalk/upgrade/susemanager-schema-5.1.5-to-susemanager-schema-5.1.6/411-rbac-endpoint.sql
```diff
@@ -18,6 +18,9 @@ INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_re
 INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
     SELECT '', '/hub/sync/registerHub', 'POST', 'W', False
     WHERE NOT EXISTS (SELECT 1 FROM access.endpoint WHERE endpoint = '/hub/sync/registerHub' AND http_method = 'POST');
+INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
+    SELECT '', '/hub/serverInfo', 'GET', 'W', False
+    WHERE NOT EXISTS (SELECT 1 FROM access.endpoint WHERE endpoint = '/hub/serverInfo' AND http_method = 'GET');
```
mcalmer, line 23: This is now in the wrong directory. susemanager-schema-5.1.6 was already tagged.
This needs to be now in susemanager-schema-5.1.6-to-susemanager-schema-5.1.7/

mcalmer (changes requested):

schema/spacewalk/upgrade/susemanager-schema-5.1.5-to-susemanager-schema-5.1.6/411-rbac-endpoint.sql
CDellaGiusta: done

mcalmer (dismissed):

mackdk (approved):

PR #10180, opened 2025-04-13
get virtual instance info from physical hosts using salt virt.vm_info

mcalmer:
## What does this PR change?

After removing the virtualization feature we need to take care to still get the virtual host <=> guest relation for
SCC reporting.
This PR use salt virt.vm_info execution module to get the virtual instances running on physical systems.
The gatherer taskomatic job execute this one time per day only on physical salt managed nodes to get up to date data.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): SUSE/spacewalk#25437

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/suse/manager/utils/SaltUtils.java
+++ b/java/code/src/com/suse/manager/utils/SaltUtils.java
@@ -74,6 +74,7 @@ import com.redhat.rhn.domain.server.SAPWorkload;
 import com.redhat.rhn.domain.server.Server;
 import com.redhat.rhn.domain.server.ServerAppStream;
 import com.redhat.rhn.domain.server.ServerFactory;
+import com.redhat.rhn.domain.server.VirtualInstanceFactory;
 import com.redhat.rhn.domain.server.ansible.InventoryPath;
 import com.redhat.rhn.domain.user.User;
 import com.redhat.rhn.frontend.action.common.BadParameterException;
@@ -83,6 +84,7 @@ import com.redhat.rhn.manager.errata.ErrataManager;
 import com.redhat.rhn.manager.rhnpackage.PackageManager;
 import com.redhat.rhn.manager.system.AnsibleManager;
 import com.redhat.rhn.manager.system.SystemManager;
+import com.redhat.rhn.manager.system.VirtualInstanceManager;
 import com.redhat.rhn.taskomatic.TaskomaticApi;
 import com.redhat.rhn.taskomatic.TaskomaticApiException;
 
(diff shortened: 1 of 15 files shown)


java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -655,6 +662,19 @@ else if (action.getActionType().equals(ActionFactory.TYPE_HARDWARE_REFRESH_LIST)
...
+        else if (action.getActionType().equals(ActionFactory.TYPE_VIRT_PROFILE_REFRESH)) {
+            if (serverAction.getStatus().equals(ActionFactory.STATUS_FAILED)) {
+                var originalErrorMsg = getErrorFromStateApplyResult(jsonResult,
+                        "Error while updating the virtual instance profile.\nGot no result from the system.");
+                serverAction.setResultMsg(originalErrorMsg);
+            }
+            else {
+                serverAction.setResultMsg("Success");
+            }
+            serverAction.getServer().asMinionServer()
```
rjmateus, line 674: IF we don't find the minion in the database should we log something with this data? It should be a edge case, but can happen.

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -1936,19 +2003,15 @@ private void handleInventoryRefresh(Action action, ServerAction serverAction, Js
...
-        switch (type) {
-            case DEB:
-                return PackageEvrFactory.lookupOrCreatePackageEvr(PackageEvr.parseDebian(version));
-            case RPM:
-                return PackageEvrFactory.lookupOrCreatePackageEvr(epoch.map(StringUtils::trimToNull).orElse(null),
-                        version, release.orElse("0"), PackageType.RPM);
-            default:
-                throw new RuntimeException("unreachable");
-        }
+        return switch (type) {
```
rjmateus, line 2006: The default case that throws an exception was removed. Is this intended?

rjmateus (approved):
Just two small questions, but overall looks good to me.

java/code/src/com/suse/manager/utils/SaltUtils.java
mcalmer: `type` is an enum and has only 2 values. We have cases for both and this means "default" is never reached.
IntelliJ was marking this and it should be removed.

mackdk (approved):

java/code/src/com/suse/manager/utils/SaltUtils.java
mackdk: The `default` case for simple `switch` expressions like this one is not needed because the exhaustiveness is enforced by the compiler:

>The cases of a switch expression must be exhaustive; for all possible values there must be a matching switch label. (Obviously switch statements are not required to be exhaustive.)
>
>In practice this normally means that a default clause is required; however, **in the case of an enum switch expression that covers all known constants, a default clause is inserted by the compiler to indicate that the enum definition has changed between compile-time and runtime**. Relying on this implicit default clause insertion makes for more robust code; now when code is recompiled, the compiler checks that all cases are explicitly handled. Had the developer inserted an explicit default clause (as is the case today) a possible error will have been hidden.

https://openjdk.org/jeps/361

PR #10199, opened 2025-04-15
Reposync improvements

NamelessOne91:
## What does this PR change?

[ Placeholder ]

Related to https://github.com/SUSE/spacewalk/issues/25102

## GUI diff

No difference.

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage

- Cucumber step definitions were modified

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25102
Port(s): 

- 5.0
- 4.3

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


srbarrios (dismissed):
LGTM, thank you!

maximenoel8 (dismissed):
LGTM, thank you

srbarrios (dismissed):

NamelessOne91:
Refactored to use the new information available through channel.software.get_details API

testsuite/features/support/commonlib.rb
```diff
@@ -590,71 +590,32 @@ def update_controller_ca
...
+  timeout
+end
+
+# This method checks if the channel with the given label has been fully synced
+#
+# @param channel_label [String] the label of the channel to check
+# @return [Boolean] true if the synchronization is completed, false otherwise
+def channel_sync_completed?(channel_label)
+  channel_details = $api_test.channel.software.get_details(channel_label)
+  # 'C' for new created, 'S' for syncing and 'R' for ready
```
srbarrios, line 617: Saving few characters will not give us any additional performance.
Let's just write the full word please.

testsuite/features/support/commonlib.rb
```diff
@@ -590,71 +590,32 @@ def update_controller_ca
...
-      log "WARN: Repository metadata for #{channel_name} seems not synchronized. Even if the reposync log says it is."
-      return false
-    end
+# This method returns the timeout, in seconds, for syncing the given channel
+#
+# @param channel [String] the channel to check
+# @return [Integer] number of seconds representing the timeout
+def channel_timeout(channel)
+  if channel.include?('custom_channel') || channel.include?('ptf')
+    log 'Timeout of 10 minutes for a custom channel or PTF'
```
srbarrios: We can't use log here. `log` is a keyword from Cucumber, and this is executed out of a Cucumber step, so it will not be treat it the same way. I suggest we use  `STDOUT.puts` (No idea why we had it wrong in the previous implementation, probably we did not know this trick.)

srbarrios (dismissed):

testsuite/features/support/commonlib.rb
NamelessOne91: Agreed, but that's what's currently returned from the API itself. It will need changes from devs' side.

PR #10211, opened 2025-04-17
download.suse.de is unreacheable from SLC1

Bischoff:
## What does this PR change?

Use download.nue.suse.com instead of download.suse.de. It's same site, but it's filtered differently by our firewalls.

**IMPORTANT NOTE**: I have searched other occurrences of download.suse.de in our test suite. There's quite a bit of them, but none should have an importance on the tests in SLC1.

In particular, in the Kiwi and Docker profiles, `download.suse.de` gets changed into the domain name of the closest mirror anyways.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

Cucumber tests were added

- [x] **DONE**

## Links

Port(s):
 * 5.0: https://github.com/SUSE/spacewalk/pull/27022
 * 4.3: https://github.com/SUSE/spacewalk/pull/27023

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/step_definitions/command_steps.rb
+++ b/testsuite/features/step_definitions/command_steps.rb
@@ -67,13 +67,13 @@ Then(/^it should be possible to use the custom download endpoint$/) do
 end
 
 Then(/^it should be possible to reach the build sources$/) do
-  if product == 'Uyuni'
-    # TODO: move that internal resource to some other external location
-    log 'Sanity check not implemented, move resource to external network first'
-  else
-    url = 'http://download.suse.de/ibs/SUSE/Products/SLE-SERVER/12-SP5/x86_64/product/media.1/products.key'
-    get_target('server').run("curl --insecure --location #{url} --output /dev/null")
-  end
+  example_product =
+    if product == 'Uyuni'
+      'distribution/leap-micro/5.5/product/repo/Leap-Micro-5.5-x86_64-Media1/media.1/products'
+    else
+      'ibs/SUSE/Products/SLE-Product-SLES/15-SP6/x86_64/product/media.1/products'
+    end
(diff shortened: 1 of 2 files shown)


srbarrios (approved):
Thank you.
In the future, I think we should move `download.nue.suse.com` to an environment variable, that we can set-up through sumaform on controller's bashrc depending the value of the `mirror` salt grain.

I took the liberty to make this PR https://github.com/uyuni-project/sumaform/pull/1819/files hope it helps at some point

Bischoff:
> Thank you. In the future, I think we should move `download.nue.suse.com` to an environment variable, that we can set-up through sumaform on controller's bashrc depending the value of the `mirror` salt grain.

There are cases where we explicitely want no mirror. Like this one.

srbarrios:
> > Thank you. In the future, I think we should move `download.nue.suse.com` to an environment variable, that we can set-up through sumaform on controller's bashrc depending the value of the `mirror` salt grain.
> 
> There are cases where we explicitely want no mirror. Like this one.

Why we don't want the mirror here? 🤔 
If we rely on the mirror as part of the whole test suite, including reposync, I think it make sense to check here the mirror, as it gonna be the place where the built sources are provided on that test suite

Bischoff:
> Why we don't want the mirror here? 🤔 If we rely on the mirror as part of the whole test suite, including reposync, I think it make sense to check here the mirror, as it gonna be the place where the built sources are provided on that test suite

In fact, yes, you're correct.

srbarrios (dismissed):

srbarrios (dismissed):

srbarrios (approved):

PR #10236, opened 2025-04-24
SonarCloud fix: java:S1488 Immediately return expression instead of assigning it to a temporary variable

CDellaGiusta:
## What does this PR change?
SonarCloud error reduction fix, rule java:S1488 Immediately return expression instead of assigning it to a temporary variable

Local variables should not be declared and then immediately returned or thrown

Declaring a variable only to immediately return or throw it is considered a bad practice because it adds unnecessary complexity to the code. This practice can make the code harder to read and understand, as it introduces an extra step that doesn’t add any value. Instead of declaring a variable and then immediately returning or throwing it, it is generally better to return or throw the value directly. This makes the code cleaner, simpler, and easier to understand.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Port(s): not backported
- [x] **DONE**

## Changelogs
- [x] No changelog needed

## Re-run a test


mcalmer (approved):
Ok, but I think the change in CVEAuditManager will be reverted as soon as somebody need to debug this and want to know what is returned :-)

lucidd (approved):

PR #10239, opened 2025-04-24
SonarCloud fix: java:S1845 Methods and field names should not be the same or differ only by capitalization

CDellaGiusta:
## What does this PR change?
SonarCloud error reduction fix, rule java:S1845 Methods and field names should not be the same or differ only by capitalization

This rule raises an issue when there is a method and a field in a class with names that differ only by capitalization.

Looking at the set of methods in a class, including superclass methods, and finding two methods or fields that differ only by capitalization is confusing to users of the class. It is similarly confusing to have a method and a field which differ only in capitalization or a method and a field with exactly the same name and visibility.

In the case of methods, it may have been a mistake on the part of the original developer, who intended to override a superclass method, but instead added a new method with nearly the same name.

Otherwise, this situation simply indicates poor naming. Method names should be action-oriented, and thus contain a verb, which is unlikely in the case where both a method and a member have the same name (with or without capitalization differences). However, renaming a public method could be disruptive to callers. Therefore renaming the member is the recommended action.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Port(s): not backported
- [x] **DONE**

## Changelogs
- [x] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/domain/iss/IssSlave.java
```diff
@@ -53,7 +53,7 @@ public class IssSlave extends BaseDto {
      */
     @Override
     public Long getId() {
-        return this.id;
+        return this.idPrivate;
```
mcalmer, line 56: I do not understand this. We do this everywhere. A field called "id" and a getter called getId(). 
And is this live in domain/ package I expect this is a hibernate object. Not sure if this is a good idea to rename this.

mcalmer (changes requested):
I am not convinced about this rule and the changes

CDellaGiusta:
> I am not convinced about this rule and the changes
Here is the rule explanation: (https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS1845&rule_key=java%3AS1845)
It is issued when a field and another field/method differ only by capitalization.

In the case of IssSlave class, I changed the private fields since I didn't know where the "public static final String " ones were used

mcalmer (approved):

java/code/src/com/redhat/rhn/domain/iss/IssSlave.java
CDellaGiusta: Reverted modifications in IssSlave class as agreed

cbbayburt:
I totally agree with the objections here. It's pretty common practice to have fields named `foo` and getters `getFoo`. IMHO we don't need this level of childproofing.

CDellaGiusta:
> I totally agree with the objections here. It's pretty common practice to have fields named `foo` and getters `getFoo`. IMHO we don't need this level of childproofing.

it's not about getters/setters but about having fields with the same name but different case: e.g.
        public static final String UID = "xxx";
        private String uid;

Please have a look at the rule explanation [here](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS1845&rule_key=java%3AS1845).
We can also decide to drop the rule straight away.... as we did with other rules we reckon are not fitting our needs.... any take is very welcome!

cbbayburt:
> it's not about getters/setters but about having fields with the same name but different case: e.g.

Ah nevermind then. I got confused by the issue history.

java/code/src/com/redhat/rhn/frontend/action/systems/SystemRemoteCommandAction.java
```diff
@@ -82,29 +82,29 @@ public static final class FormData {
...
-            this.uid = "root";
-            this.gid = "root";
-            this.timeout = FormData.DEFAULT_TIMEOUT;
-            this.label = "";
-            this.scriptBody = "#!/bin/sh\n# Add your shell script below\n";
+            this.uidPrivate = "root";
+            this.gidPrivate = "root";
+            this.timeoutPrivate = FormData.DEFAULT_TIMEOUT;
+            this.labelPrivate = "";
+            this.scriptBodyPrivate = "#!/bin/sh\n# Add your shell script below\n";
```
lucidd, line 99: looking at the rule its clear that it tries to prevent confusing naming. That said having field names and their getter/setters named the same is pretty standard and anyone would expect that so changing that would be pretty confusing on its own. The confusing part in the current version are the constants that are named after the fields but contain the form field names instead. So i would suggest changing their naming to be more clear by prefixing them with something like `FIELD_` or `FIELD_NAME_`

CDellaGiusta: This is doable, of course. But being the constants "public final", the impact in the code is greater that changing the "private" fields. As a first instance I aimed at the least possible change that could fulfill the rule.
But if you reckon that the one you proposed is clearer, no problem I'll go for it.

lucidd: That makes sense but the impact should be minimal either way since we don't have external dependents and with an ide you can do a safe repo wide rename. The reason we are doing this in the first place is this rule that is about reducing confusion and increasing maintainability. If the solution does the opposite while technically shutting up the rule i'd rather just disable the rule.

CDellaGiusta: I reverted these changes, please have a look

PR #10240, opened 2025-04-24
SonarCloud fix: java:S1192 String literals should not be duplicated

CDellaGiusta:
## What does this PR change?
SonarCloud error reduction fix, rule java:S1192 String literals should not be duplicated
Duplicated string literals make the process of refactoring complex and error-prone, as any change would need to be propagated on all occurrences.
Exceptions: To prevent generating some false-positives, literals having less than 5 characters are excluded.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Port(s): not backported
- [x] **DONE**

## Changelogs
- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/BaseSearchAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/BaseSearchAction.java
@@ -248,8 +248,8 @@ public abstract class BaseSearchAction extends RhnAction {
                     boolean flag) {
         LocalizationService ls = LocalizationService.getInstance();
         Map<String, String> selection = new HashMap<>();
-        selection.put("display", (flag ? "*" : "") + ls.getMessage(key));
-        selection.put("value", StringEscapeUtils.escapeHtml4(value));
+        selection.put(DISPLAY_KEY, (flag ? "*" : "") + ls.getMessage(key));
+        selection.put(VALUE_KEY, StringEscapeUtils.escapeHtml4(value));
         options.add(selection);
     }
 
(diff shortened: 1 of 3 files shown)


java/code/src/com/redhat/rhn/manager/system/SystemManager.java
```diff
@@ -172,6 +172,10 @@
  * SystemManager
  */
 public class SystemManager extends BaseManager {
+    public static final String SYSTEM_QUERIES = "System_queries";
+    public static final String USER_ID = "user_id";
+    public static final String ORG_ID = "org_id";
+    public static final String SET_LABEL = "set_label";
```
mackdk, line 178: I'm wondering why SonarCloud is not complaining for `sid` or `server_id` which are used multiple times in this files. Moreover,  maybe we should define an interface `CommonColumns` and use it everywhere, because you have already defined the same `ORG_ID` constant in `PackageManager`. 

Another solution would be to just ignore these warnings when they are generated by column names and mark them as won't fix.

mackdk (approved):

java/code/src/com/redhat/rhn/manager/system/SystemManager.java
CDellaGiusta: Here is the reason.
>  Exceptions: To prevent generating some false-positives, literals having less than 5 characters are excluded.

I anyway aimed for the minimum solution that avoids showing the errors

mackdk: Yeah, but this create an inconsistency in how we pass the parameters to the map. Personally, I'd rather have consistent code rather then using sometimes a constant and sometimes not.

CDellaGiusta: Fair point.
I will make things consistent when I have a moment

CDellaGiusta: I reverted this change. Please have a look

mackdk (approved):

PR #10243, opened 2025-04-24
 ISSv3: Add SCC endpoints to send system and metering data

CDellaGiusta:
## What does this PR change?

This is the upstream branch referred to https://github.com/uyuni-project/uyuni/pull/10229

SUSE Manager send data about registered systems to SCC. In a Hub scenario, the peripheral server should send the data to the Hub and the Hub should send it to SCC.
This PR puts the basis of having the Hub acting as a proxy to the SCC, in particular:
- Creates DB schema on the Hub to hold data send from peripheral server 
- Add endpoints to act as if it was an SCC:
        PUT /connect/organizations/systems (create systems)
        DELETE /connect/organizations/systems/:id (delete systems)

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/25350
Port(s): not backported
- [x] **DONE**

## Changelogs
- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/test/ForwardRegistrationTaskTest.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/test/ForwardRegistrationTaskTest.java
@@ -233,6 +233,10 @@ public class ForwardRegistrationTaskTest extends BaseTestCaseWithUser {
         SaltStateGeneratorService.INSTANCE.setSkipSetOwner(true);
         Config.get().setString(ConfigDefaults.REG_BATCH_SIZE, String.valueOf(batchSize));
 
+        for (SCCRegCacheItem item : getAllSCCRegCacheItems()) {
+            SCCCachingFactory.deleteRegCacheItem(item);
+        }
+
         servers = new ArrayList<>();
         for (int i = 0; i < systemSize; i++) {
             Server testSystem = ServerTestUtils.createTestSystem();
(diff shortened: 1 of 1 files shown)


schema/spacewalk/common/tables/suseSccProxy.sql
```diff
@@ -0,0 +1,34 @@
...
+
+CREATE TABLE suseSccProxy
+(
+    proxy_id            BIGINT CONSTRAINT suse_sccproxy_proxy_id_pk PRIMARY KEY
+                                                  GENERATED ALWAYS AS IDENTITY,
+    peripheral_fqdn     VARCHAR(128),
+    scc_login           VARCHAR(64),
+    scc_passwd          VARCHAR(64),
+    scc_creation_json   TEXT,
+    scc_id              NUMERIC,
```
mcalmer, line 25: I wonder if we can also use BIGINT for scc_id.
Do we join it with another ID column which is numeric? Or do we need to make a foreign key to a NUMERIC column?
If not, we may want to use BIGINT as well.

CDellaGiusta: no, we do not. This is the id coming from the SCC

CDellaGiusta:
Moved to upstream in : https://github.com/uyuni-project/uyuni/pull/10245

PR #10271, opened 2025-04-29
cleanup old supportdata before creating new one

mcalmer:
## What does this PR change?

Remove old supportdata before creating new to prevent full disk problems.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


lucidd (approved):

susemanager-utils/susemanager-sls/src/modules/supportdata.py
```diff
@@ -24,6 +26,21 @@ def _get_supportdata_dir():
...
+        err = args[2]
+        log.error("Failed to remove %s: %s", path, err[1])
+
+    for d in os.listdir("/var/log/"):
+        fullpath = os.path.join("/var/log", d)
+        if os.path.isdir(fullpath) and re.match(r"^supportdata-[0-9]+$", d):
+            if (time.time() - os.path.getmtime(fullpath)) > 3600:
+                # older than 1 hour
+                # pylint: disable-next=deprecated-argument
+                shutil.rmtree(fullpath, onerror=_log_error)
```
meaksh, line 41: I guess it is not common, but why if the supportconfig takes more than 1 hour to be generated? Shouldn't we maybe increase this value to 1 day maybe?

mcalmer: This is not the check here. 
This is about a cleanup of supportconfig data which were generated before. 

In case of salt-ssh managed clients we cannot send the supportconfig using "push_dir".
So we return only the path to the master and the Java code is triggering an "scp" which should copy the files from the minion to the master. 
But if this fail, the supportconfig stay on the minion "forever".
This code exists to cleanup old supportconfigs before we create a new one, but only if it is older than 1 hour.

PR #10272, opened 2025-04-29
Timeouts measured on 5.0 BV

Bischoff:
## What does this PR change?

Fine-tune timeouts measured experimentally on 5.0 BV.

Note: I changed rockylinux8 to rockylinux-8 and rockylinux9 to rockylinux-9. These are the channel names I see on 5.0.
There's a possibility that the channel names are slightly different between Manager and Uyuni. If that's the case, we will need them in double, one with a caret, and once without :-( .

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

Cucumber tests were added

- [x] **DONE**

## Links

Port(s):
 * 5.0: https://github.com/SUSE/spacewalk/pull/27108
 * 4.3: https://github.com/SUSE/spacewalk/pull/27109

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/constants.rb
+++ b/testsuite/features/support/constants.rb
@@ -1074,16 +1074,16 @@ CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION = {
       ],
     'rockylinux8' => # CHECKED
       %w[
-        rockylinux8-x86_64
-        rockylinux8-appstream-x86_64
-        rockylinux8-extras-x86_64
+        rockylinux-8-x86_64
+        rockylinux-8-appstream-x86_64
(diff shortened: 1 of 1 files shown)


srbarrios (approved):
WOW. The increase is significant in some of them.

Bischoff:
> WOW. The increase is significant in some of them.

Yes, we don't measure quite often, and the updates tend to pile up.

There are also some random variations at a given time, but not that much.

There's also a selection bias, because I processed only the ones that triggered the message "XXX did not finish reposyncing" this time.

PR #10300, opened 2025-05-07
Supportdata ssh tunnel

mcalmer:
## What does this PR change?

Mainly use the proxies where the minion is connected through as jumhosts by compiling the `ProxyCommand` for the ssh options.
Special ssh tunnel options should not be needed as we do ssh commands and not use a different protocol.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests - hard to test without cucumber and a real upload. But this would pollute the upload server with useless data

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/26617

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/suse/manager/utils/SaltUtils.java
+++ b/java/code/src/com/suse/manager/utils/SaltUtils.java
@@ -981,8 +981,7 @@ public class SaltUtils {
                         Optional.empty());
                 Optional<List<String>> proxyCommand = SaltSSHService.sshProxyCommandOption(proxyPath,
                         minionServer.getContactMethod().getLabel(),
-                        minionServer.getMinionId(),
-                        Optional.ofNullable(minionServer.getSSHPushPort()).orElse(SaltSSHService.SSH_PUSH_PORT));
+                        minionServer.getMinionId(), port);
 
                 String sshOptions = "-o ConnectTimeout=2 " + String.join(" ",
                         proxyCommand
--- a/java/code/src/com/suse/manager/utils/SaltUtils.java
+++ b/java/code/src/com/suse/manager/utils/SaltUtils.java
@@ -977,16 +977,15 @@ public class SaltUtils {
                 var actionPath = MinionActionUtils.getActionPath(minionServer, actionId);
                 var user = SaltSSHService.getSSHUser();
                 var port = Optional.ofNullable(minionServer.getSSHPushPort()).orElse(SaltSSHService.SSH_PUSH_PORT);
+                String method = minionServer.getContactMethod().getLabel();
                 List<String> proxyPath = SaltSSHService.proxyPathToHostnames(minionServer.getServerPaths(),
                         Optional.empty());
-                Optional<List<String>> proxyCommand = SaltSSHService.sshProxyCommandOption(proxyPath,
-                        minionServer.getContactMethod().getLabel(),
-                        minionServer.getMinionId(), port);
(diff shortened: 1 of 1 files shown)


java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -975,15 +975,27 @@ private Optional<Path> fetchSupportFiles(MinionServer minionServer, ServerAction
...
                 var user = SaltSSHService.getSSHUser();
                 var port = Optional.ofNullable(minionServer.getSSHPushPort()).orElse(SaltSSHService.SSH_PUSH_PORT);
-                var rsync = "rsync -p --chown salt:susemanager --chmod=660 -e 'ssh -p %d -i %s' %s@%s:%s/* %s/."
-                        .formatted(port, SaltSSHService.SSH_KEY_PATH, user, hostname, supportDataDir, actionPath);
+                List<String> proxyPath = SaltSSHService.proxyPathToHostnames(minionServer.getServerPaths(),
+                        Optional.empty());
+                Optional<List<String>> proxyCommand = SaltSSHService.sshProxyCommandOption(proxyPath,
+                        minionServer.getContactMethod().getLabel(),
+                        minionServer.getMinionId(),
+                        Optional.ofNullable(minionServer.getSSHPushPort()).orElse(SaltSSHService.SSH_PUSH_PORT));
```
lucidd, line 985: this seems to be the same as port

CDellaGiusta (approved):
LGTM

java/code/src/com/suse/manager/utils/SaltUtils.java
```diff
@@ -975,15 +975,26 @@ private Optional<Path> fetchSupportFiles(MinionServer minionServer, ServerAction
...
+                List<String> proxyPath = SaltSSHService.proxyPathToHostnames(minionServer.getServerPaths(),
+                        Optional.empty());
+                Optional<List<String>> proxyCommand = SaltSSHService.sshProxyCommandOption(proxyPath,
+                        minionServer.getContactMethod().getLabel(),
+                        minionServer.getMinionId(), port);
+
+                String sshOptions = "-o ConnectTimeout=2 " + String.join(" ",
+                        proxyCommand
+                                .map(l -> l.stream().map("-o %s"::formatted).collect(Collectors.toList()))
+                                .orElse(new ArrayList<>()));
```
mackdk, line 989: Nitpicking here, but I think unwrapping immediately the `Optional` leads to a more readable code. Morevoer `minionServer.getMinionId()` is already stored in the variable `hostname` initalized before the `if`.

```java
                String method = minionServer.getContactMethod().getLabel();
                List<String> proxyCommand = SaltSSHService.sshProxyCommandOption(proxyPath, method, hostname, port)
                    .orElse(List.of());

                String sshOptions = "-o ConnectTimeout=2 " + proxyCommand.stream()
                    .map("-o %s"::formatted)
                    .collect(Collectors.joining(" "));
```

I would argue that `SaltSSHService::sshProxyCommandOption()` should just return a List<String> instead of an optional, but this would be a bigger refactoring so let's avoid it for now :slightly_smiling_face: . In general, we should not wrap collections into optionals since most of the times the same effect can be achieved with an empty collection.

mackdk (approved):

PR #10315, opened 2025-05-12
Support Data UI

mackdk:
## What does this PR change?

This PR adds a tab in the system details page, to allow the user to generate and upload the support config data.

## GUI diff

After:
image
image
image

- [X] **DONE**

## Documentation

- Documentation WIP

- [ ] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/26619

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/suse/utils/gson/RecordTypeAdapterFactory.java
+++ b/java/code/src/com/suse/utils/gson/RecordTypeAdapterFactory.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2025 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+
+package com.suse.utils.gson;
+
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.RecordComponent;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Creates the adapters for serializing and deserializing the record types
+ */
+public class RecordTypeAdapterFactory implements TypeAdapterFactory {
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> typeToken) {
+        Class<? super T> rawType = typeToken.getRawType();
+        if (!rawType.isRecord()) {
+            return null;
+        }
+
+        return (TypeAdapter<T>) new RecordTypeAdapter<>(gson, rawType);
+    }
+
+    // Adapter for a specific record type
+    private static class RecordTypeAdapter<T> extends TypeAdapter<T> {
+
+        private final String recordName;
+
+        private final RecordComponent[] components;
+
+        private final Constructor<T> constructor;
+
+        private final Map<String, Type> componentTypeMap;
+
+        private final Gson gson;
+
+        RecordTypeAdapter(Gson gsonIn, Class<T> recordClassIn) {
+            this.gson = gsonIn;
+            this.recordName = recordClassIn.getName();
+
+            try {
+                // Find all the record components
+                this.components = recordClassIn.getRecordComponents();
+                // Cache the type of each component
+                this.componentTypeMap = Arrays.stream(components)
+                    .collect(Collectors.toMap(RecordComponent::getName, RecordComponent::getGenericType));
+                // Retrieve the canonical constructor
+                this.constructor = recordClassIn.getDeclaredConstructor(
+                    Arrays.stream(components).map(RecordComponent::getType).toArray(Class<?>[]::new)
+                );
+            }
+            catch (ReflectiveOperationException | RuntimeException ex) {
+                throw new IllegalStateException("Unable to collect the information of record " + recordName, ex);
+            }
+        }
+
+        @Override
+        public void write(JsonWriter jsonWriter, T value) throws IOException {
+            jsonWriter.beginObject();
+
+            for (RecordComponent field : components) {
+                writeField(jsonWriter, value, field);
+            }
+
+            jsonWriter.endObject();
+        }
+
+        private void writeField(JsonWriter jsonWriter, T value, RecordComponent field) throws IOException {
+            String fieldName = field.getName();
+
+            try {
+                Object fieldValue = field.getAccessor().invoke(value);
+
+                // This suppression is unavoidable as we don't know at compile-time the generic type of the component
+                @SuppressWarnings("unchecked")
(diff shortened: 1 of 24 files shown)


mcalmer (changes requested):
I think you need to add RBAC schema data as you created a new endpoint

web/html/src/manager/minion/details/support-data.renderer.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+
+import SpaRenderer from "core/spa/spa-renderer";
+
+import { SupportData, UploadRegion } from "./support-data";
+
+export const renderer = (
```
Etheryte, line 7: Renderers in the rest of the codebase usually (although unfortunately not always) follow the pattern of taking two params, `id: string` and then an object with the props. It would be preferable to follow this, as it will make future changes we have planned around Jade renderers easier.

web/html/src/manager/minion/details/support-data.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+interface Props {
+  serverId: number;
+  availableRegions: UploadRegion[];
+  supportProgramName: string | null;
+}
+
+export const SupportData: React.FC<Props> = ({ serverId, availableRegions, supportProgramName }): JSX.Element => {
+  const [formModel, setFormModel] = useState(
+    () => new SupportDataRequest("", availableRegions[0].label, "", localizedMoment())
```
Etheryte, line 41: Will there always be at least one element in `availableRegions`?

Etheryte (approved):
The frontend looks good otherwise, but please fix the lint before merging.

web/html/src/manager/minion/details/support-data.renderer.tsx
mackdk: Fixed, I introduces a props object.

web/html/src/manager/minion/details/support-data.tsx
mackdk: Yes, there will always be at least one location for the upload. I introduced a type `UploadRegionArray` to enforce that through typescript, even though as we are passing the values in the jade template, it is mostly for documenting it at the moment.

mackdk:
> I think you need to add RBAC schema data as you created a new endpoint

Done. The [documentation](https://github.com/uyuni-project/uyuni/wiki/RBAC-Endpoint-Mapping#1-i-have-a-new-pagefeature)  doesn't say anything about the `accessGroupEndpoint` table, but if I don't add an entry there the new namespace will be inaccessible for all users...

mcalmer:
> > I think you need to add RBAC schema data as you created a new endpoint
> 
> Done. The [documentation](https://github.com/uyuni-project/uyuni/wiki/RBAC-Endpoint-Mapping#1-i-have-a-new-pagefeature) doesn't say anything about the `accessGroupEndpoint` table, but if I don't add an entry there the new namespace will be inaccessible for all users since it's not part of any access group

I had problems with accessGroupNamespace table. @cbbayburt wanted to update the wiki page. Seems he also need to add this table.

schema/spacewalk/upgrade/susemanager-schema-5.1.6-to-susemanager-schema-5.1.7/003-supportdata-rbac.sql
```diff
@@ -0,0 +1,33 @@
...
+    SELECT ns.id, ep.id FROM access.namespace ns, access.endpoint ep
+    WHERE ns.namespace = 'systems.details.support' AND ns.access_mode = 'W'
+    AND ep.endpoint = '/manager/api/systems/:sid/details/uploadSupportData' AND ep.http_method = 'POST'
+    ON CONFLICT DO NOTHING;
+
+INSERT INTO access.accessGroupNamespace
+    SELECT ag.id, ns.id
+    FROM access.accessGroup ag, access.namespace ns
+    WHERE ns.namespace = 'systems.details.support'
+    ON CONFLICT DO NOTHING;
```
mcalmer, line 33: you are only selecting the "namespace" but not the "group" you want to add this to.

mcalmer: And maybe you need to define it also in common/data/ for the initial setup.
It depends on when there is already a definition which will catch your new namespace.

mcalmer (changes requested):

schema/spacewalk/upgrade/susemanager-schema-5.1.6-to-susemanager-schema-5.1.7/003-supportdata-rbac.sql
mackdk: I'm not selecting a group because  I want all group to access this page [like other systems.details namespaces pages](https://github.com/uyuni-project/uyuni/blob/e26a2db54af0a069c2b0c313424e36daed811e75/schema/spacewalk/common/data/accessGroupNamespace.sql#L129-L171): 

```sql
-- Namespace: systems.ansible
-- Namespace: systems.audit.*
-- Namespace: systems.bootstrap
-- Namespace: systems.custom_data
-- Namespace: systems.details.*
-- Namespace: systems.events
-- Namespace: systems.formulas
-- Namespace: systems.groups.*
-- Namespace: systems.list
-- Namespace: systems.maintenance
-- Namespace: systems.profiles
-- Namespace: systems.proxy
-- Namespace: systems.recurring
-- Namespace: systems.search
-- Namespace: systems.snapshots
-- Namespace: systems.software.*
-- Namespace: systems.ssm
-- Namespace: systems.states.*
-- Namespace: systems.vhms
-- Permit to all
INSERT INTO access.accessGroupNamespace
    SELECT ag.id, ns.id
    FROM access.accessGroup ag, access.namespace ns
    WHERE (ns.namespace = 'systems.ansible' OR
        ns.namespace LIKE 'systems.audit.%' OR
        ns.namespace = 'systems.bootstrap' OR
        ns.namespace = 'systems.custom_data' OR
        ns.namespace LIKE 'systems.details.%' OR
        ns.namespace = 'systems.events' OR
        ns.namespace = 'systems.formulas' OR
        ns.namespace LIKE 'systems.groups.%' OR
        ns.namespace = 'systems.list' OR
        ns.namespace = 'systems.maintenance' OR
        ns.namespace = 'systems.profiles' OR
        ns.namespace = 'systems.proxy' OR
        ns.namespace = 'systems.recurring' OR
        ns.namespace = 'systems.search' OR
        ns.namespace = 'systems.snapshots' OR
        ns.namespace LIKE 'systems.software.%' OR
        ns.namespace = 'systems.ssm' OR
        ns.namespace LIKE 'systems.states.%' OR
        ns.namespace = 'systems.vhms')
    ON CONFLICT DO NOTHING;
```

Moreover, I don't need a common a definition in common data because the query above already covers it since it authorizes `systems.details.%`.

mcalmer: ahh, right. accessGroup are our "roles" like `channel_admin`. Then it makes sense to not specify it.
And that insert in common/data makes it during the initial setup.
:+1:

mcalmer (approved):

lucidd (approved):

java/code/src/com/suse/utils/gson/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,148 @@
...
+            catch (ReflectiveOperationException | RuntimeException ex) {
+                throw new IOException("Unable to process field '%s' of record %s".formatted(fieldName, recordName), ex);
+            }
+        }
+
+        @Override
+        public T read(JsonReader jsonReader) throws IOException {
+            if (jsonReader.peek() == JsonToken.NULL) {
+                jsonReader.nextNull();
+                return null;
```
lucidd, line 114: small nitpick i think we should fail on null values to make sure if we expect a value we always get one and leave the null handling to our optional type adapter so its explicit what is required and optional.

PR #10362, opened 2025-05-20
Issv3 SCC: Implement and test the update last seen mechanism when a hub is acting as an SCC proxy

CDellaGiusta:
## What does this PR change?
Implements and tests the update last seen mechanism when a hub is acting as an SCC proxy

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/27058
Port(s): not backported
- [x] **DONE**

## Changelogs
- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/scc/SCCEndpoints.java
+++ b/java/code/src/com/suse/scc/SCCEndpoints.java
@@ -39,7 +39,7 @@ import com.suse.scc.client.SCCConfigBuilder;
 import com.suse.scc.client.SCCFileClient;
 import com.suse.scc.client.SCCWebClient;
 import com.suse.scc.model.SCCOrganizationSystemsUpdateResponse;
-import com.suse.scc.model.SCCRegisterSystemJson;
+import com.suse.scc.model.SCCRegisterSystemItem;
 import com.suse.scc.model.SCCRepositoryJson;
 import com.suse.scc.model.SCCSystemCredentialsJson;
 import com.suse.scc.model.SCCVirtualizationHostJson;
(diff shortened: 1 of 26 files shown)


java/code/src/com/suse/scc/proxy/SCCProxyManager.java
```diff
@@ -44,20 +44,32 @@ public SCCProxyManager(SCCProxyFactory sccProxyFactoryIn) {
     /**
      * Creates systems from the lists and stores data to be later registered to SCC
      *
-     * @param systemsList a list of {@link SCCRegisterSystemJson} to be created
+     * @param systemsList a list of {@link SCCRegisterSystemItem} to be created
      * @param peripheralFqdnIn the fqdn of the peripheral
      * @return list of corresponding generated records
      */
-    public List<SCCProxyRecord> createSystems(List<SCCRegisterSystemJson> systemsList, String peripheralFqdnIn) {
+    public List<SCCProxyRecord> createSystems(List<SCCRegisterSystemItem> systemsList, String peripheralFqdnIn) {
```
mcalmer, line 51: I would rename this method to "createOrUpdateSystems" as it also does an update of an existing entry

java/code/src/com/suse/scc/SCCEndpoints.java
```diff
@@ -351,13 +351,13 @@ public String productTree(Request request, Response response) {
      */
     public String createSystems(Request request, Response response, HubSCCCredentials credentials) {
         try {
```
mcalmer, line 353: I would name this "createOrUpdateSystems" as it does both

java/code/src/com/suse/scc/proxy/SCCProxyManager.java
```diff
@@ -44,20 +44,32 @@ public SCCProxyManager(SCCProxyFactory sccProxyFactoryIn) {
...
+            if (system.isOnlyLastSeenAt()) {
+                //update last seen at
+                sccProxyRecord.setLastSeenAt(system.getLastSeenAt());
+            }
+            else {
+                //full creation or update
+                String sccCreationJson = Json.GSON.toJson(system);
+                sccProxyRecord.setSccCreationJson(sccCreationJson);
+                sccProxyRecord.setStatus(SccProxyStatus.SCC_CREATION_PENDING);
+            }
```
mcalmer, line 72: Do I understand this right.
Also in case of update last seen, we send back the changed system records.
We just ignore the response on the client side for this call as we do not need it.

(I am fine with this - I just want to clarify that this is wanted)

CDellaGiusta: yes. In the SCC, both `createUpdateSystems` and `updateBulkLastSeen` call the same endpoint `/connect/organizations/systems`, that always return an answer, but in the latter case we ignore the answer in the caller.
So I did when the hub acts as a proxy for the SCC

mcalmer (approved):

PR #10415, opened 2025-06-04
Ensure bootstrap.sh can write to overlay fs

agraul:
## What does this PR change?
We can't assume that the overlay dir corresponding to the transaction we open contains the minion config directory. For example, when venv-salt-minion was previously installed and bootstrap.sh is executed again, the new overlay dir might not have etc/venv-salt-minion.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/spacewalk/certs-tools/rhn_bootstrap_strings.py
+++ b/spacewalk/certs-tools/rhn_bootstrap_strings.py
@@ -948,6 +948,10 @@ if [ $VENV_ENABLED -eq 1 ]; then
     MINION_SERVICE="venv-salt-minion"
 fi
 
+if [ -n "$SNAPSHOT_ID" ]; then
+    test -d "$MINION_CONFIG_DIR" || mkdir -p "$MINION_CONFIG_DIR"
+fi
+
 if [ $REGISTER_THIS_BOX -eq 1 ]; then
     echo "* registering"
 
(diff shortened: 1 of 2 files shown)


agraul:
Changelog test stumbles over a period that's not separating two sentences :man_shrugging:

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -948,6 +948,10 @@ def getRegistrationSaltSh(productName):
     MINION_SERVICE="venv-salt-minion"
 fi
 
+if [ -n "$SNAPSHOT_ID" ]; then
+    test -d "$MINION_CONFIG_DIR" || mkdir -p "$MINION_CONFIG_DIR"
```
m-czernek, line 952: I'm not sure I understand here, but if the only problem is that we could re-execute the bootstrap script multiple times, shouldn't we just continue in the previous transaction?

In `new_transaction`, we use `/usr/sbin/tukit -q open`, we could use `/usr/sbin/tukit --continue --discard -q open` instead?

agraul: There might have been transactions in between the one that added `etc/venv-salt-minion/` and the latest. I didn't want to search all the transactions, thought it was cleaner to "onboard in a new transaction", no matter what happened before.

m-czernek: Hmm, that sounds about right. I suppose this would be situational. Perhaps at some point we could let the users decide whether they want a new transaction or continue in previous one.

m-czernek (approved):

agraul:
I think we now need a different approach, this won't work on SL Micro 6.2

PR #10437, opened 2025-06-09
Add the possibility to specify the installation of the last package v…

maximenoel8:
## Context

When installing a package via the UI, searching for a package displays all available versions.
Currently, the first package in the list is selected by default, which often does not correspond to the latest version.
This behavior can lead to outdated versions being installed unintentionally.

This PR improves the package selection logic to ensure that the latest version is selected when needed.

## What does this PR change?
**Update**
I found the toggle action was use not only in package list but in all other similar list.
I split the toggle function in two. One specific to the package list and one generic to match the first row for all the other lists.

**New**

Add a last version option that, when filtering a package, retrieves the list of available versions and automatically selects the latest version.
Update the BV smoke test to use last version option when installing a package.

*Note: I was not sure to keep the old method (first match) and decided to keep it because we also use this toggle for removing package. In this case, we don't care about the version.*

</details>

</details>

**Refactor**

The filtering and package selection steps currently rely on substeps.
Refactor these substeps into standalone functions that can be reused.

To achieve this:
 - Create a new helper file named navigation_step_helper.rb
 - Move the relevant logic into functions within this file
 - Update the step definitions to call these new helper functions directly

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/27417
Port(s): Need port if accepted

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/build_validation/smoke_tests/smoke_tests.template
+++ b/testsuite/features/build_validation/smoke_tests/smoke_tests.template
@@ -78,7 +78,7 @@ Feature: Smoke tests for <client>
     And I wait until I see "Installable Packages" text
     And I enter the package for "<client>" as the filtered package name
     And I click on the filter button
-    And I check the package for "<client>" in the list
+    And I check the package last version for "<client>" in the list
     And I click on "Install Selected Packages"
     And I click on "Confirm"
     Then I should see a "1 package install has been scheduled for" text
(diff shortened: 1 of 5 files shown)


maximenoel8:
The package order is alphabetic, that's why we have the last version last.

PR #10440, opened 2025-06-09
Use the env variables MGR_SERVER_HOSTNAME for the hostname

admd:
## What does this PR change?

This add the possibility to use the env variable for hostname, this will significantly improve the situation where users needs to have a bootstrap script for each proxy and some users have number of proxies in 4 digits

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: I will create a PR for the visibility

- [x] **DONE**

## Test coverage
- No tests:  Not covered

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/spacewalk/certs-tools/rhn_bootstrap_strings.py
+++ b/spacewalk/certs-tools/rhn_bootstrap_strings.py
@@ -107,7 +107,10 @@ REACTIVATION_KEY=${{REACTIVATION_KEY:-}}
 
 # can be edited, but probably correct:
 CLIENT_OVERRIDES={overrides}
-HOSTNAME={hostname}
+
+#If the environment variable MGR_SERVER_HOSTNAME is set, its value will be used.
+#Otherwise, or they will default to the templated value.
+HOSTNAME=${{MGR_SERVER_HOSTNAME:-{hostname}}}
 
 ORG_CA_CERT={orgCACert}
 
--- a/spacewalk/certs-tools/spacewalk-certs-tools.changes.abid.use-env-varibale-for-hostname
+++ b/spacewalk/certs-tools/spacewalk-certs-tools.changes.abid.use-env-varibale-for-hostname
@@ -1,0 +1,1 @@
+- Add the possibility to use env variable for HOSTNAME


spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -107,7 +107,10 @@
 
 # can be edited, but probably correct:
 CLIENT_OVERRIDES={overrides}
-HOSTNAME={hostname}
+
+#If the environment variable HOSTNAME is set, its value will be used.
+#Otherwise, or they will default to the templated values.
+HOSTNAME=${{HOSTNAME:-{hostname}}}
```
mcalmer: Not sure if this is a good idea as HOSTNAME is mostly only the short hostname.
Currently in our testsuites we have always the long hostname as we call `hostname -f`.

And it seems that "HOSTNAME" is the hostname of the MLM Server and not the client hostname.

aaannz: Exactly as @mcalmer said. HOSTNAME is not hostname of the client running the bootstrap script, but rather server or proxy hostname provided by the templating during the bootstrap script creation.

If you want this behavior, i.e. overriding server/proxy value on bootstrap time, I suggest to use different envvar then HOSTNAME, it is already confusing we use this variable internally like this.
How about `MGR_SERVER`?

If you want to allow envvar to modify minionid, look at [L153](https://github.com/uyuni-project/uyuni/blob/a9ee88794d3d02dd5817b67c8d22a9a721652c32/spacewalk/certs-tools/rhn_bootstrap_strings.py#L153)

admd: Michael/Ondrej, isn't it the same that we set when doing `mgr-boostrap --hostname $proxyfqdn` for example?

Some user needs to create separate bootstrap script for each proxy and end up with a lot of bootstrap scripts. I thought this is the field that is the only difference among all these bootstrap scripts. If this is the case, then user will only need to create one bootstrap script and the call by by passing the env varibale like,  export HOSTNAME=proxy.fqdn && bootstrap.sh

admd: I know, variable name isn't ideal in this case and happy to change it to something better. But what what should we do this HOSTNAME, should we completely remove it - yes that it will probably need some more refactoring.

mcalmer: HOSTNAME is part of the default environment . If you do not set it, it will try to connect to itself

aaannz: Keep the variable name as is. Just the environmental variable you are introducing should be named differently in my opinion. So e.g.:

```suggestion
HOSTNAME=${{MGR_SERVER:-{hostname}}}
```

So the actual usage on the client system would be
`MGR_SERVER=proxy.fqdn bootstrap.sh`

admd: Thank you folks, I have made change and used 'MGR_SERVER_HOSTNAME' for env variable in this case.

aaannz (approved):

PR #10458, opened 2025-06-12
Implemented Bootstrap tooltips

bisht-richa:
## What does this PR change?

Tooltips have been upgraded to Bootstrap tooltips for a cleaner and more consistent user experience.

## GUI diff

Screenshot 2025-06-12 at 13 48 02
Screenshot 2025-06-12 at 13 48 10
Screenshot 2025-06-12 at 13 48 26
Screenshot 2025-06-12 at 13 48 34
Screenshot 2025-06-12 at 13 52 12

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/webapp/WEB-INF/decorators/layout_c.jsp
+++ b/java/code/webapp/WEB-INF/decorators/layout_c.jsp
@@ -64,5 +64,22 @@
         </c:if>
       </script>
     </div>
+    <script>
+   // Dispose all existing tooltips
+    document.querySelectorAll('[data-bs-toggle="tooltip"]').forEach((el) => {
+      const existingTooltip = bootstrap.Tooltip.getInstance(el);
+      if (existingTooltip) {
+        existingTooltip.dispose();
+      }
+    });
+
+    // Initialize tooltips with default trigger
+    const tooltipTriggerList = Array.from(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
+    const tooltipList = tooltipTriggerList.map((el) => {
+      return new bootstrap.Tooltip(el, {
+        trigger: 'hover focus',
+      });
+    });
+    </script>
   </body>
 </html:html>
--- a/web/html/javascript/spacewalk-essentials.js
+++ b/web/html/javascript/spacewalk-essentials.js
@@ -449,6 +449,17 @@ function onDocumentReadyInitOldJS() {
 jQuery(document).ready(function() {
   onDocumentReadyGeneral();
   onDocumentReadyAutoBootstrapGrid();
+  //Initialize Bootstrap tooltips
+  jQuery('[data-bs-toggle="tooltip"]').tooltip();
+
+  const tooltipObserver = new MutationObserver(() => {
+    jQuery('[data-bs-toggle="tooltip"]').tooltip();
+  });
+
+  tooltipObserver.observe(document.body, {
+    childList: true,
+    subtree: true,
+  });
   registerSpacewalkContentObservers();
   initIEWarningUse();
   listenForGlobalNotificationChanges();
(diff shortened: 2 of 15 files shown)


cbbayburt (dismissed):
Java part LGTM :+1: 

I added @Etheryte as a reviewer for JS

java/code/webapp/WEB-INF/decorators/layout_c.jsp
```diff
@@ -64,5 +64,22 @@
         </c:if>
       </script>
     </div>
+    <script>
```
Etheryte: In general, we do not want to add stray global scripts like this, they don't show up in intellisense when you check scripts and they also do not have Typescript or other checks.

web/html/javascript/spacewalk-essentials.js
```diff
@@ -449,6 +449,17 @@ function onDocumentReadyInitOldJS() {
 jQuery(document).ready(function() {
   onDocumentReadyGeneral();
   onDocumentReadyAutoBootstrapGrid();
+  //Initialize Bootstrap tooltips
```
Etheryte: Same comment as `layout_c`, we want to get rid of this script file over time, not add new things to it. A good tip is that if you're adding straight js as new code to a file as opposed to ts, it's probably not the right place. (This does not apply to bugfixes, e.g. if you were fixing an existing bug in this file.)

web/html/src/branding/css/base/theme.scss
```diff
@@ -1312,3 +1312,17 @@ tr.changed > td {
     font-size: .4em;
   }
 }
+ :root {
```
Etheryte: We want to get rid of the large theme.scss file over time, not add to it. Please lift this out into a separate module as other code does.

web/html/javascript/spacewalk-essentials.js
bisht-richa: In this case, could you suggest the right place to add this new code?

Etheryte: The general entrypoint for our Typescript code is `web/html/src/manager/index.ts`. You can see find this from `web/html/src/build/webpack.config.js` line 96 and forward:

```
    entry: {
      "javascript/manager/main": path.resolve(__dirname, "../manager/index.ts"),
      "css/updated-suse-light": path.resolve(__dirname, "../branding/css/suse-light.scss"),
      "css/updated-suse-dark": path.resolve(__dirname, "../branding/css/suse-dark.scss"),
      "css/updated-uyuni": path.resolve(__dirname, "../branding/css/uyuni.scss"),
    },
```

We don't currently have other scripts there that are both global and tie to the UI in such a way. What I would recommend is to create a new file in `web/html/src/components`, e.g. `web/html/src/components/tooltips.tsx`, export a function from there and import that in `web/html/src/manager/index.ts`.

This way you can always trace back where what code comes from when you're refactoring.

web/html/src/manager/index.ts
```diff
@@ -70,6 +71,8 @@ window.spaImportReactPage = function spaImportReactPage(pageName) {
       `Found no page with name "${pageName}", did you add the renderer to \`pages\` in \`web/html/src/manager/index.ts\`?`
     );
   }
-
+  initializeTooltips();
```
Etheryte, line 74: I think this is incorrect. Since `initializeTooltips` creates a `MutationObserver`, it really only needs to be called once. Right now we create a new `MutationObserver` on every page navigation.

Etheryte (approved):

PR #10490, opened 2025-06-23
ci: fix cucumber reports not uploading when a test fails

akayiz:
## What does this PR change?

Turns out GitHub Actions by default skips all the following workflow steps if a step in the workflow failed, so previously, the cucumber reports upload step was only triggered on successful test runs. This PR ensures the upload runs in both cases, whether tests were successful or not, so that we can have the expected output, which is cucumber reports for PRs that had no failing tests, as well as for PRs that had failing scenarios/tests.

Making it always run even if any step in the workflow failed shouldn't be a problem if no reports were generated, since we have the `if-no-files-found: warn` option, which just outputs a warning and continues the workflow normally if no files were found.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links
No links.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/.github/workflows/acceptance_tests_base.yml
+++ b/.github/workflows/acceptance_tests_base.yml
@@ -215,7 +215,7 @@ jobs:
         run: ./testsuite/podman_runner/${{ inputs.tests }}
       
       - name: Upload Cucumber JSON Reports
-        if: ${{ !inputs.skip_tests && needs.filters.outputs.require_acceptance_tests == 'true' }}
+        if: ${{ always() && !inputs.skip_tests && needs.filters.outputs.require_acceptance_tests == 'true' }}
         uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 #v4.6.1
         with:
           name: cucumber_json_reports_${{ inputs.server_id }}


.github/workflows/acceptance_tests_base.yml
```diff
@@ -215,7 +215,7 @@ jobs:
         run: ./testsuite/podman_runner/${{ inputs.tests }}
       
       - name: Upload Cucumber JSON Reports
-        if: ${{ !inputs.skip_tests && needs.filters.outputs.require_acceptance_tests == 'true' }}
+        if: ${{ always() && !inputs.skip_tests && needs.filters.outputs.require_acceptance_tests == 'true' }}
```
srbarrios, line 218: I had been reading about `always()`, if I understood well, it will always return `true`.
So, I don't see how prepending it, it could change the behavior of this condition.

See https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#example-of-always

**Update**

Ok, I wrongly understood the first time.
All good: `Causes the step to always execute, and returns true, even when canceled`

akayiz: Yeah, the thing is that GitHub by default adds `success()` if we don't add any status check functions, as mentioned below, that's why this step wasn't running when a test failed.

success-status

So technically, if we don't add `always()`, this is how this line looks implicitly:

```yaml
if: ${{ success() && !inputs.skip_tests && needs.filters.outputs.require_acceptance_tests == 'true' }}
```

We don't want that, we want this step to run in `success()` and in `failure()`, so we add `always()`

PR #10496, opened 2025-06-24
WIP: RBAC UI - component steps progressbar

bisht-richa:
## What does this PR change?

- Created new component steps progressbar
- Created some UIs with dummy data for RBAC UI using steps progressbar component

## GUI diff

image

image

image

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


cbbayburt:
Some notes after initial review:

- Add RBAC endpoint entries to DB
  - /manager/admin/config/access-group, etc.
- Rename menu item "Access Group" to "Access Control"
  - Left menu bar
  - URL '../access-group'
  - Controller method: showAccessGroup
  - Template file 'access-group.jade'
  - .. etc

Role list page:
- No pagination needed
- Don't list built-in groups
  - Remove 'Type' column

Namespace list table:
- Search bar must work for descriptions as well

Add users step & add users tab in edit:
- Only users from the same org as the AG should be listed
  - Remove 'Organization' column

schema/spacewalk/common/tables/namespace.sql
```diff
@@ -20,3 +20,7 @@ COMMENT ON TABLE access.namespace IS 'Namespace definitions to provide access to
 
 CREATE UNIQUE INDEX namespace_ns_mode_uq
 ON access.namespace(namespace, access_mode);
+
+CREATE INDEX namespace_trgm_idx
+ON access.namespace
+USING gist (lower(namespace||description) gist_trgm_ops);
```
cbbayburt: Maybe we should add a "." after the namespace. Currently, "systemget_details" works, but "system.get_details" doesn't.

web/html/src/manager/admin/access-control/access-group-permissions.tsx
```diff
@@ -120,11 +128,20 @@ const AccessGroupPermissions = (props: Props) => {
...
       });
-  }, []);
+  }
+
+  useEffect(() => {
+    if (!isLoading) {
+      setIsLoading(true);
+      getNamespaces(searchValue);
+    }
+  }, [searchValue]);
```
cbbayburt: In this kind of search-as-you-type task, it's a good practice to limit the requests to the backend to prevent clogging up the network.

We already do this using the `debounce` utility from lodash. It wraps the method that does the API call and it waits for some milliseconds of inactivity before it calls the method. This way you avoid calling the API for every keystroke, and only do one call after the user pauses typing.

Here's an example usage (note the last arg: 100ms of wait time):

https://github.com/uyuni-project/uyuni/blob/fa55769c56d76af2ef586f972e47f9e8b2833ce2/web/html/src/manager/content-management/shared/components/panels/sources/channels/channels-selection.tsx#L83-L103

I think it should work if you wrap the `getNamespace` method like this and call it in `useEffect`.

web/html/src/manager/admin/access-control/access-group-list.tsx
```diff
@@ -0,0 +1,152 @@
...
+        setAccessGroups(accessGroups.filter((g) => g.id !== item.id));
+      })
+      .catch((error) => setMessages(Network.responseErrorMessage(error)));
+  };
+
+  const actionButtons = (item: AccessGroupListItem) => {
+    if (item.orgName === "-") {
+      return (
+        <div className="btn-group">
+          <Button className="btn-default btn-sm" icon="fa-user" />
```
cbbayburt: This button has no function. I think we don't need it at all. I don't even remember why this was here in the first place (also below in the `else` block).

web/html/src/manager/admin/access-control/access-group-list.tsx
```diff
@@ -0,0 +1,152 @@
...
+      );
+    } else {
+      return (
+        <div className="btn-group">
+          <Button className="btn-default btn-sm" icon="fa-user" />
+          <LinkButton
+            className="btn-default btn-sm"
+            icon="fa-pencil"
+            href={`/rhn/manager/admin/access-control/show-access-group/${item.id}`}
+          />
```
cbbayburt: This button is missing a `title`

java/core/src/main/java/com/suse/manager/webui/menu/MenuTree.java
```diff
@@ -520,6 +523,9 @@ private MenuItem getAdminNode(Map<String, Boolean> adminRoles) {
                             .withVisibility(adminRoles.get("satellite")))
                     .addChild(new MenuItem("Password Policy")
                             .withPrimaryUrl("/rhn/manager/admin/config/password-policy")
+                            .withVisibility(adminRoles.get("satellite")))
+                    .addChild(new MenuItem("Access Control")
+                            .withPrimaryUrl("/rhn/manager/admin/access-control")
```
cbbayburt: I think we can remove the item from here if there's no specific reason for it?

java/spacewalk-java.changes.bisht-richa.steps-progressbar-component
```diff
@@ -0,0 +1 @@
+- Implemented a new StepsProgressBar component for managing RBAC.
```
cbbayburt, line 1: The end user doesn't need to know about StepsProgressBar. Let's keep it simple. Maybe something like:
```suggestion
- Implemented RBAC management UI
```

java/core/src/main/java/com/redhat/rhn/domain/access/NamespaceFactory.java
```diff
@@ -46,6 +57,77 @@ public static List<Namespace> list() {
...
+     * Lists namespaces defined in MLM filter by a search parameter.
+     * @param filterParam the search parameter to filter by
+     * @return the list of all namespaces
+     */
+    public static List<Namespace> list(String filterParam) {
+        return getSession()
+                .createNativeQuery("""
+                SELECT *
+                FROM access.namespace
+                WHERE :filter <% lower(namespace||description)
```
cbbayburt: Maybe we should add a "." after the namespace (also to the index in the schema). Currently, "systemget_details" works, but "system.get_details" doesn't.

schema/spacewalk/upgrade/susemanager-schema-5.2.0-to-susemanager-schema-5.2.1/100-rbac-ui-mappings.sql
```diff
@@ -0,0 +1,112 @@
+-- Namespaces
+INSERT INTO access.namespace (namespace, access_mode, description)
+    SELECT 'admin.access-control.access-group', 'R', 'List and detail custom access groups.'
+    WHERE NOT EXISTS (SELECT 1 FROM access.namespace WHERE namespace = 'admin.access-control.access-group' AND access_mode = 'R');
+
+INSERT INTO access.namespace (namespace, access_mode, description)
+    SELECT 'admin.access-control.access-group', 'W', 'Create, modify and delete custom access groups.'
+    WHERE NOT EXISTS (SELECT 1 FROM access.namespace WHERE namespace = 'admin.access-control.access-group' AND access_mode = 'W');
```
cbbayburt: We don't plan to expand this feature much further so I think only 2 levels are enough for the namespace. Also I suggested "access" to be consistent with the `api.access` namespace.

```suggestion
INSERT INTO access.namespace (namespace, access_mode, description)
    SELECT 'admin.access', 'R', 'List and detail custom access groups.'
    WHERE NOT EXISTS (SELECT 1 FROM access.namespace WHERE namespace = 'admin.access-control.access-group' AND access_mode = 'R');

INSERT INTO access.namespace (namespace, access_mode, description)
    SELECT 'admin.access', 'W', 'Create, modify and delete custom access groups.'
    WHERE NOT EXISTS (SELECT 1 FROM access.namespace WHERE namespace = 'admin.access-control.access-group' AND access_mode = 'W');
```

Also: the data is missing in `common/data/..`

web/spacewalk-web.changes.bisht-richa.steps-progressbar-component
```diff
@@ -0,0 +1 @@
+- Implemented a new StepsProgressBar component for managing RBAC.
```
cbbayburt, line 1: The end user doesn't need to know about StepsProgressBar. Let's keep it simple. Maybe something like:
```suggestion
- Implemented RBAC management UI
```

cbbayburt (changes requested):
Thank you very much @bisht-richa, @parlt91. It looks very good overall.

I've also deployed and tested it, and everything works as expected.

Apart from my inline comments, here's a few notes I had when testing, mostly on the UX:

- The text `[API]` looks a bit raw on the table. Maybe we should pick an icon for this instead.
- Bug: when I go back to the first step and remove an item from "Copy permissions from" control, the permissions list is not recalculated.
- After a series of searching and adding, it's easy to lose track of the overall selections. We need an overview of all the selections at the final step.
- When you search for a namespace, the natural next step is always expanding some rows. We can automatically expand the whole filtered list after a search, and reduce the necessary steps taken by the user.
- Apart from that, we could also use the usual "Expand/Collapse all" controls on the table.

java/core/src/main/java/com/redhat/rhn/domain/access/NamespaceFactory.java
parlt91: Upon further testing this didn't really work out that well. I've refined the search to use a Full-text-search with prefix and trigram fallback. It should be much more reliable now. Maybe not optimized for performance that well, but with only ~1000 rows in the table (not expected to grow that much) it should be fine. On testing this was quick.
Please have a look at the updated search function if you have some time.

web/html/src/manager/admin/access-control/access-group-list.tsx
bisht-richa: Fixed

web/html/src/manager/admin/access-control/access-group-permissions.tsx
bisht-richa: I have wrapped the getNamespace in debounce.

web/html/src/manager/admin/access-control/access-group-list.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <Button className="btn-default btn-sm" icon="fa-user" />
+        </div>
+      );
+    } else {
+      return (
+        <div className="btn-group">
+          <LinkButton
+            className="btn-default btn-sm"
+            icon="fa-pencil"
+            title="Edit Access Group"
```
cbbayburt:
```suggestion
            title={t("Edit Access Group")}
```

web/html/src/manager/admin/access-control/access-group-list.tsx
```diff
@@ -0,0 +1,151 @@
...
+        setAccessGroups(accessGroups.filter((g) => g.id !== item.id));
+      })
+      .catch((error) => setMessages(Network.responseErrorMessage(error)));
+  };
+
+  const actionButtons = (item: AccessGroupListItem) => {
+    if (item.orgName === "-") {
+      return (
+        <div className="btn-group">
+          <Button className="btn-default btn-sm" icon="fa-user" />
```
cbbayburt: This Button needs to go too.

cbbayburt (approved):

Bischoff (dismissed):
approving the test suite part -- thanks!

Pablogoliva (dismissed):
LGTM

PR #10523, opened 2025-06-27
Skip password verification for PAM users (bsc#1245398)

aaannz:
## What does this PR change?

This commit changes how PAM users password verification works. Instead of relying on additional variable, which was not correctly set, always rely on user PAM settings.
Also use PAM settings only in the actual password set routine in order not to hide some potential future errors.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/27630
Port(s): https://github.com/SUSE/spacewalk/pull/27639

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/user/CreateUserCommand.java
+++ b/java/code/src/com/redhat/rhn/manager/user/CreateUserCommand.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2009--2015 Red Hat, Inc.
+ * Copyright (c) 2025 SUSE LLC
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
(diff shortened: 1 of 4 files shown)


java/code/src/com/redhat/rhn/manager/user/CreateUserCommand.java
```diff
@@ -85,7 +86,7 @@ public CreateUserCommand() {
     public ValidatorError[] validate() {
         errors = new ArrayList<>(); //clear validation errors
 
-        if (passwordErrors != null && !user.getUsePamAuthentication()) {
```
Serp1co, line 88: Shouldn't we still enforce this: !user.getUsePamAuthentication() ? (i think we already do it somewhere up the flow and here the passwordErrors array will always be null)

just a nitpick, not really needed i think.

Serp1co (approved):
LGTM

java/code/src/com/redhat/rhn/manager/user/CreateUserCommand.java
aaannz: Indeed it is done in the `setPassword` below. So we should not have any passwordErrors in case of PAM auth and if we have, it is a bug and IMO validator should report it.

PR #10525, opened 2025-06-27
Fix systems page numeric filter

wweellddeerr:
## What does this PR change?

It fixes the numeric filter by decoding HTML entities and improving the matcher selection logic. Previously, the component could incorrectly match shorter operators (like <) when longer ones (like <=) were intended.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: bug fix

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): 

https://github.com/SUSE/spacewalk/issues/27222

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/web/html/src/components/table/NumericSearchField.tsx
+++ b/web/html/src/components/table/NumericSearchField.tsx
@@ -20,12 +20,27 @@ const resultingCriteria = (matcher: string, value: string) => {
   return matcher && value && value.trim() !== "" ? matcher + value : null;
 };
 
+const decodeHTMLEntities = (str) => {
+  const txt = document.createElement("textarea");
+  txt.innerHTML = str;
+  return txt.value;
+};
+
 export const NumericSearchField = ({ name, criteria, onSearch }) => {
   let criteriaMatcher = "";
   let criteriaValue = "";
   if (criteria) {
-    criteriaMatcher = matchers.find((it) => criteria.startsWith(it.value))?.value ?? "=";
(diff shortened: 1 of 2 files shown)


Etheryte:
It's unclear to me based on the current comments and description why decoding is an issue in this context, but the current implementation works nicely.

Regarding picking longer matchers, an alternative way to achieve the same thing would be to sort the matchers at the top level first:

```ts
const matchers = ...;
// Pick longer matchers first, so we don't take "<" when it should be "<="
const sortedMatchers = matchers.toSorted((a, b) => b.value.length - a.value.length);
```

and then in the `NumericSearchField`:

```ts
const matcher = sortedMatchers.find(matcher => criteria.includes(matcher.value))?.value || "=";
```

To be honest, I'm not even sure which version is more easy to follow, I'm fine with either.

Etheryte (approved):

wweellddeerr:
> It's unclear to me based on the current comments and description why decoding is an issue in this context, but the current implementation works nicely.

Hey @Etheryte, thanks for your review! To be honest, it’s not very clear to me either. The issue was happening on the customer side, but I couldn’t reproduce it locally. I suspected it might be related to parsing HTML entities, so I added the decoding just in case. After the change, the customer confirmed it solved the problem.

PR #10555, opened 2025-07-07
Simplify action handling (step 1 - SaltServerActionService)

CDellaGiusta:
## What does this PR change?
The specific aim of this PR is to simplify the method `SaltServerActionService::callsForAction.`
This method has a list of "if" clauses that, for any given action type, return the salt calls that need to be executed for the targeted minions. Making a list of "ifs" on a type is a code smell of a misuse of object-oriented inheritance.
   
The final goal is to have a single method call `actionIn.getSaltCalls(minions)` where each action is implementing their own logic of handling, getting rid of the "if" on type, making a correct use of the Action object hierarchy.

The steps followed were:

1) moving each chunk of code executed in the "if" body part to a public static method of the relevant action class, maintaining the same name as the handler method in `SaltServerActionService` (if any). This part allows to move the relevant handling code (as well as constants and other helper methods) to the appropriate Action classes.

2) moving `SaltServerActionService` constants in a new class `SaltParameters`, to be shared among classes

3) act on `ActionFactory::createAction(ActionType typeIn) `to make sure that all action types create a different action class (it was not true for `ConfigAction` and `PackageAction` child classes). This step touched also the database discriminator-value of `ConfigAction` subclasses

4) renaming the newly created public static methods of Action derived classes with the same name (`getSaltCalls`) and unifying calling parameters

5) creating an Action base class method `getSaltCalls` and making the public static methods of Action derived classes as `@Override` methods of the base class

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**
## Links

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
+++ b/java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
@@ -15,9 +15,54 @@
 package com.redhat.rhn.domain.action;
 
 
+import com.redhat.rhn.domain.server.MinionSummary;
+
+import com.suse.manager.reactor.messaging.ApplyStatesEventMessage;
+import com.suse.salt.netapi.calls.LocalCall;
+import com.suse.salt.netapi.calls.modules.State;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
 /**
  * HardwareRefreshAction
  */
 public class HardwareRefreshAction extends Action {
+
+    /**
+     * @param minionSummaries a list of minion summaries of the minions involved in the given Action
+     * @return minion summaries grouped by local call
+     */
+    @Override
+    public Map<LocalCall<?>, List<MinionSummary>> getSaltCalls(List<MinionSummary> minionSummaries) {
+        Map<LocalCall<?>, List<MinionSummary>> ret = new HashMap<>();
+
+        // salt-ssh minions in the 'true' partition
+        // regular minions in the 'false' partition
+        Map<Boolean, List<MinionSummary>> partitionBySSHPush = minionSummaries.stream()
+                .collect(Collectors.partitioningBy(MinionSummary::isSshPush));
+
+        // Separate SSH push minions from regular minions to apply different states
+        List<MinionSummary> sshPushMinions = partitionBySSHPush.get(true);
+        List<MinionSummary> regularMinions = partitionBySSHPush.get(false);
+
+        if (!sshPushMinions.isEmpty()) {
+            ret.put(State.apply(List.of(
+                            ApplyStatesEventMessage.HARDWARE_PROFILE_UPDATE),
+                    Optional.empty()), sshPushMinions);
+        }
+        if (!regularMinions.isEmpty()) {
+            ret.put(State.apply(Arrays.asList(
+                            ApplyStatesEventMessage.SYNC_ALL,
+                            ApplyStatesEventMessage.HARDWARE_PROFILE_UPDATE),
+                    Optional.empty()), regularMinions);
+        }
+
+        return ret;
+    }
 
 }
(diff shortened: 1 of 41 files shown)


java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
```diff
@@ -15,9 +15,54 @@
...
+                .collect(Collectors.partitioningBy(MinionSummary::isSshPush));
+
+        // Separate SSH push minions from regular minions to apply different states
+        List<MinionSummary> sshPushMinions = partitionBySSHPush.get(true);
+        List<MinionSummary> regularMinions = partitionBySSHPush.get(false);
+
+        if (!sshPushMinions.isEmpty()) {
+            ret.put(State.apply(List.of(
+                            ApplyStatesEventMessage.HARDWARE_PROFILE_UPDATE),
+                    Optional.empty()), minionSummaries);
```
mcalmer: should minionSummaries here be sshPushMinions?

java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
```diff
@@ -15,9 +15,54 @@
...
+        }
+        if (!regularMinions.isEmpty()) {
+            ret.put(State.apply(Arrays.asList(
+                            ApplyStatesEventMessage.SYNC_ALL,
+                            ApplyStatesEventMessage.HARDWARE_PROFILE_UPDATE),
+                    Optional.empty()), minionSummaries);
```
mcalmer: should minionSummaries here be regularMinions ?

java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
CDellaGiusta: I had the same doubt. Probably a copy-paste leftover already present in the original code. 
Fixed

java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
CDellaGiusta: Ditto. Fixed

java/code/src/com/redhat/rhn/frontend/xmlrpc/system/test/SystemHandlerTest.java
```diff
@@ -526,6 +532,9 @@ public void testSetChildChannels() throws Exception {
                 .filter(a -> a.getActionType().equals(ActionFactory.TYPE_SUBSCRIBE_CHANNELS))
                 .findFirst();
         assertTrue(first.isPresent(), "No Subscribe Channels Action created");
+        if (first.isPresent()) {
```
Etheryte, line 535: Java is not my first language, but isn't this conditional redundant since we assert right before it?

CDellaGiusta: That's correct. Nevertheless SonarCloud complains if you're using an optional (first.get) without checking for its actual presence (first.isPresent). So that's only to avoid this hiccup.

Etheryte: My subjective preference would be turn off rules that don't actually make sense or just litter code, but I'm not that familiar with the Java side of the codebase, there may be places where the same rule is useful. From the frontend side, many of the rules generate noise without being useful so we're considering turning some off.

CDellaGiusta: I agree with you. This case can actually make sense in ordinary code (not tests) where you may not want to use an optional without checking for its existence (hence firing a null pointer exception). Some SonarCloud rules actually are on the border between being acceptable and making unwanted noise.

PR #10583, opened 2025-07-11
Adjust defaults and validations for registries in proxy configuration

rjpmestre:
## What does this PR change?

This PR updates the defaults and validations for registry related fields in the proxy configuration.

Recent discussions regarding access to registries helped clarify some initial assumptions about default URLs and validation behavior.

These adjustments align the intended workflows and current requirements:
a) Remove custom validations from registry url and tag fields (now only required)
b) Change registry tag related fields from dropdowns to text inputs
c) Set default source mode to RPM
d) Fix default MLM registry URL
e) Adjust default tag values

## GUI diff

Before / After:
a,b,d,e)
image

c) 
image

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: previous unit tests adjusted

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/27867 https://github.com/SUSE/spacewalk/issues/27897 https://github.com/SUSE/spacewalk/issues/27938

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/web/html/src/manager/minion/proxy/proxy-config.tsx
+++ b/web/html/src/manager/minion/proxy/proxy-config.tsx
@@ -349,7 +349,7 @@ export function ProxyConfig({
                   <Text
                     name="registryHttpdURL"
                     label={t("HTTPD URL")}
-                    placeholder={t("e.g., " + registryUrlExample + "/proxy-httpd")}
+                    placeholder={t("e.g., {registryUrlExample}/proxy-httpd", { registryUrlExample })}
                     required
                     labelClass="col-md-3"
                     divClass="col-md-6"
@@ -366,7 +366,7 @@ export function ProxyConfig({
                   <Text
                     name="registrySaltbrokerURL"
                     label={t("Salt Broker URL")}
-                    placeholder={t("e.g., " + registryUrlExample + "/proxy-salt-broker")}
+                    placeholder={t("e.g. {registryUrlExample}/proxy-salt-broker", { registryUrlExample })}
                     required
                     labelClass="col-md-3"
                     divClass="col-md-6"
@@ -383,7 +383,7 @@ export function ProxyConfig({
                   <Text
                     name="registrySquidURL"
                     label={t("Squid URL")}
-                    placeholder={t("e.g., " + registryUrlExample + "/proxy-squid")}
+                    placeholder={t("e.g. {registryUrlExample}/proxy-squid", { registryUrlExample })}
                     required
                     labelClass="col-md-3"
                     divClass="col-md-6"
@@ -400,7 +400,7 @@ export function ProxyConfig({
                   <Text
                     name="registrySshURL"
                     label={t("SSH URL")}
-                    placeholder={t("e.g., " + registryUrlExample + "/proxy-ssh")}
+                    placeholder={t("e.g. {registryUrlExample}/proxy-ssh", { registryUrlExample })}
                     required
                     labelClass="col-md-3"
                     divClass="col-md-6"
@@ -417,7 +417,7 @@ export function ProxyConfig({
                   <Text
                     name="registryTftpdURL"
                     label={t("TFTPD URL")}
-                    placeholder={t("e.g., " + registryUrlExample + "/proxy-tftpd")}
+                    placeholder={t("e.g. {registryUrlExample}/proxy-tftpd", { registryUrlExample })}
                     required
                     labelClass="col-md-3"
                     divClass="col-md-6"


web/html/src/manager/minion/proxy/proxy-config.tsx
```diff
@@ -487,20 +408,13 @@ export function ProxyConfig({ serverId, parents, currentConfig, initFailMessage
                     labelClass="col-md-3"
                     divClass="col-md-6"
                   />
-                  <Select
+                  <Text
                     name="registryTftpdTag"
                     label={t("TFTPD Tag")}
+                    placeholder={t("e.g., " + registryTagExample)}
```
Etheryte: It's preferable to avoid string concat when dealing with translations, the reason for this is that this leaves translators little room to work with in languages where the word order differs. A better way to achieve the same result would be

```ts
t("e.g. {registryTagExample}", {
  registryTagExample
})
```

This way the translators can work with the full string and change where the actual value is put.

srbarrios (approved):
LGTM

java/spacewalk-java.changes.rjpmestre.proxy-configuration-registry-validation-and-default
```diff
@@ -0,0 +1,2 @@
+- Adjust defaults and validations for registries in proxy
+  configuration
```
aaannz: As you are addressing bug https://bugzilla.suse.com/show_bug.cgi?id=1246634 here too, please add changelog entry about it too

aaannz:
```suggestion
- Adjust defaults and validations for registries in proxy
  configuration (bsc#1246660)
```

aaannz: I think you also fixed https://bugzilla.suse.com/show_bug.cgi?id=1246347 here too, right?

rjpmestre: Sorry, saw this after i addressed it in its own commit. The previous change was not so rigid anyway. Anyway, took the opportunity to write some basic tests for the method that verifies whether if a proxy is convertible or not

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -2682,14 +2682,15 @@ public void setSapWorkloads(Set<SAPWorkload> sapWorkloadsIn) {
...
-     * SLE Micro 5, SE Micro 6, SLE15 SP6 and SLE15 SP7.
+     * Servers that are already proxies are not convertible. For the remaining, SUSE Manager considers only:
+     * SE Micro 6.1 and SLE15 SP7.
      * @return true if the server is convertible to a proxy, false otherwise
      */
     public boolean isConvertibleToProxy() {
         return !isProxy() && (
-                ConfigDefaults.get().isUyuni() || isSLEMicro5() || isSEMicro6() ||
-                        (isSLES() && (getRelease().equals("15.6") || getRelease().equals("15.7")))
+                ConfigDefaults.get().isUyuni() ||
```
srbarrios, line 2706: Does it means that any kind of system will be convertible to proxy, if we are in Uyuni?

java/code/src/com/redhat/rhn/domain/server/test/ServerProxyTest.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    static Stream<Arguments> isConvertibleToProxyOnMLMData() {
+        return Stream.of(
+            Arguments.of(SLEMICRO, "6.1", false),
+            Arguments.of(SLMICRO, "5.5", false),
+            Arguments.of(SLMICRO, "6.0", false),
+            Arguments.of(SLMICRO, "6.1", true),
+            Arguments.of(SLMICRO, "6.2", false),
+            Arguments.of(SLES, "15.8", false),
```
srbarrios, line 156: This will never exist. SLES 15 SP7 is the latest.

srbarrios:
@rjpmestre just double-checking, we will not have any problem validating the usage of an internal registry provided by a customer right?

rjpmestre:
> @rjpmestre just double-checking, we will not have any problem validating the usage of an internal registry provided by a customer right?

Exactly. We're dropping those validations. That should not longer be a problem.

java/code/src/com/redhat/rhn/domain/server/Server.java
rjpmestre: Yes. From what I understood, the reason is because we're officially only supporting SLES 15 SP7 and SL Micro 6.1.

aaannz:
Rebased and addressed the rest of @Etheryte review finding.

Etheryte:
Regarding the build error, the rest of the context is in the Slack thread, for now it should suffice if you rebase, run `yarn build` and then commit the result.

aaannz:
> Regarding the build error, the rest of the context is in the Slack thread, for now it should suffice if you rebase, run `yarn build` and then commit the result.

But that would want me to drop license change made by @deneb-alpha I don't see any change related to build automation fix for the license

deneb-alpha:
> > Regarding the build error, the rest of the context is in the Slack thread, for now it should suffice if you rebase, run `yarn build` and then commit the result.
> 
> But that would want me to drop license change made by @deneb-alpha I don't see any change related to build automation fix for the license

Please, do NOT revert back the normalization I did from `GPL-3.0` to the correct SPDX name (`GPL-3.0-only`). 

If you revert, you will "solve" the build issue here, but the submission will be blocked when we try to work at the next maintenance update. The licence tag needs to use valid SPDX names

Etheryte:
I think there might be some ambiguity to the situation. This is not specific to this pull request and I'm not specifically suggesting to revert anything. This revert will happen automatically on every single build on every single pull request that touches frontend files. This part of the spec is generated and any manual edits you make will be overwritten on build. What you're asking me is akin to not changing the build output when someone rebuilds the product.

I've raised this issue repeatedly for roughly half a year now without much in the way of a response from anyone. If it's going to block submissions, well, then it will block submissions, maybe then it will finally catch someone's attention.

aaannz:
So what do we here next?

Should I adapt `build/licenses/index.js` build script to correctly translate GPL-3.0 to GPL-3.0-Only? We already have in our dependency tree [spxd-correct](https://www.npmjs.com/package/spdx-correct) module which should handle it. We already do some AND parsing, we can very well do this for multiple allowable licenses.

What we know for the fact is:
- GPL-3.0 symbol is deprecated and no longer to be used
- we currently have only one dependency `pwstrength-bootstrap` with this License

Etheryte:
My opinion is that this is not something we should be doing to begin with. The only reason we have the current implementation is that we used to rely on a library by Microsoft, they stopped maintaining it, and we needed something to backfill that quickly.  

Since legal and other teams already have tools around this, I don't see why we would be both building and maintaining all of this. Looking at the scripts Marina linked to, it's a very complex task with a unknown amount of edge cases. We're in the business of building server management software, not parsing licenses, so in my opinion we should focus on the former.

aaannz:
We are not in a business of parsing licenses, but no one is asking us for do this. So why are we even doing this automation in our spec file? Why are we even collecting list of licenses to the spec file?

We do not do this for the uyuni-tools which also has vendored dependencies with various licenses other then Apache-2.0?
Shouldn't we just have `GPL-2.0-only` as the rest of uyuni or `Apache-2.0` as uyuni-tools in the spec file and just package the licenses?

Etheryte:
I don't have the answer to that question, but I've asked it a number of times in the last half a year without getting a response. 😄

mcalmer:
A go binary is compiled together and should have 1 license after it.

In spacewalk-web we put different fonts, css and other javascript files from different projects together. They do not loose there license just because we put them all into 1 package.
We can build 20 different packages and every package has its license, but AFAIK @Etheryte would not be happy about it as it makes development harder

aaannz:
Since another change was pushed which overwrite spec file, this question here is moot.

I will reach to the upstream of problematic package to see if they can adjust from GPL-3.0 to valid SPDX GPL-3.0-only.

PR was rebased.

PR #10657, opened 2025-07-29
Reset the Chrome driver when it crashes

Bischoff:
## What does this PR change?

This PR stops the chrome driver after it has crashed. When it crashes, main chrome renderer process dies, and the chrome driver becomes unusable for the rest of the test suite.

Then it restarts it by visiting Manager's main web page, using Capybara's lazy loading mechanisms.

We don't attempt to re-login and return to previous web page. Reason is, it would be too complicated and too fragile.

The effect of this PR is to remove hundreds of subsequent "tab crashed" failures.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

Cucumber tests were modified

- [x] **DONE**

## Links

Issue(s): #26585

Port(s):
* 4.3:
* 5.0:
* Head and 5.1: will be ported by releng from staging branch, if I understood well

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


testsuite/features/support/env.rb
```diff
@@ -150,7 +150,12 @@ def capybara_register_driver
   log "This scenario took: #{current_epoch - @scenario_start_time} seconds"
   if scenario.failed?
     begin
-      if web_session_is_active?
+      if scenario.exception.is_a?(Selenium::WebDriver::Error::WebDriverError)
```
maximenoel8, line 153: You could move this part in restart_webdriver method. But it's also fine like that.

Bischoff: YAGNI :smile_cat: , Don't create a method that would be called only once.

testsuite/features/support/env.rb
```diff
@@ -150,7 +150,12 @@ def capybara_register_driver
...
+        log "Caught web driver error: #{scenario.exception.message}"
+        Capybara.current_session.driver.quit
+        visit Capybara.app_host
+        log 'Web driver has been restarted'
+      elsif web_session_is_active?
```
srbarrios, line 158: Could it happen that w ehave a WebDriverError but the web session is active?
I'm not sure if we should order differently these if statements.

mcalmer (approved):

testsuite/features/support/env.rb
```diff
@@ -150,7 +150,12 @@ def capybara_register_driver
   log "This scenario took: #{current_epoch - @scenario_start_time} seconds"
   if scenario.failed?
     begin
-      if web_session_is_active?
+      if scenario.exception.is_a?(Selenium::WebDriver::Error::WebDriverError)
+        log "Caught web driver error: #{scenario.exception.message}"
+        Capybara.current_session.driver.quit
```
srbarrios, line 155: What happen when Capybara.current_session is broken?

srbarrios:
I think that not trying to re-login, it breaks the design and strategy made in other parts of our test framework. And it will make anyway fail next scenarios.

What about trying something like:

```
if scenario.exception.is_a?(Selenium::WebDriver::Error::WebDriverError)
  log "Caught web driver error: #{scenario.exception.message}"
  begin
    Capybara.current_session.driver.quit
  rescue => e
    warn "Failed to quit driver: #{e.message}"
  end

  begin
    relog_and_visit_previous_url
    log 'Web driver session restarted via relogin.'
  rescue => e
    warn "Could not recover session after driver crash: #{e.message}"
  end
else
  if web_session_is_active?
    handle_screenshot_and_relog(scenario, current_epoch)
  else
    warn 'There is no active web session; unable to take a screenshot or relog.'
  end
end
```

testsuite/features/support/env.rb
Bischoff: What happens is explained in PR's description. It's not only the session, but the whole driver that is broken.

Bischoff:
> Let's discuss about it (after FCS is out), before merging please.

Er, are you suggesting Kostas works with one run out of 2 with 400 failures in the meantime?

Bischoff:
> I think that not trying to re-login, it breaks the design and strategy made in other parts of our test framework. And it will make anyway fail next scenarios.
> 
> What about trying something like:
> 
> ```
> if scenario.exception.is_a?(Selenium::WebDriver::Error::WebDriverError)
>   log "Caught web driver error: #{scenario.exception.message}"
>   begin
>     Capybara.current_session.driver.quit
>   rescue => e
>     warn "Failed to quit driver: #{e.message}"
>   end
> 
>   begin
>     relog_and_visit_previous_url
>     log 'Web driver session restarted via relogin.'
>   rescue => e
>     warn "Could not recover session after driver crash: #{e.message}"
>   end
> else
>   if web_session_is_active?
>     handle_screenshot_and_relog(scenario, current_epoch)
>   else
>     warn 'There is no active web session; unable to take a screenshot or relog.'
>   end
> end
> ```

Please read the PR's description why this was discarded.

testsuite/features/support/env.rb
Bischoff: Not according to what we saw.

nodeg:
My 2 cents on this: Since this is blocking/disturbing current RRTG work (which is paused for 2 days), but also Orion and their test pipeline where the current staging changes are being tested, I think we should merge this PR as it is good enough for now and think about reworking things later.

maximenoel8:
Some explanation here:
once `Capybara.current_session.driver.quit` has been executed, relog_and_visit_previous_url can not work because we killed the websession. To recover from there, you need to save the url before hand but doing that in a tab crashed doesn't work. That's why Eric simplify with revisiting the main url.
We need to run a `visit` command to (lazy) restart the webdriver after closing it.

I had a PR trying to recover the previous url but it adds lot of complexity and I can not assure the stability.
https://github.com/SUSE/spacewalk/pull/27921/files
It requires some protective mechanics to avoid having the after step running for a long time.

And we were not able to recover the previous url.


For the active session, it's in a begin rescue so it will be OK to fail. Using web_session_is_active? doesn't work because we do a visit command that will trigger the tab crashed exception and skip the driver restart.

testsuite/features/support/env.rb
srbarrios: What I mean is that if `Capybara.current_session` is broken/nil, you will not be able to call `quit`.

srbarrios:
> I had a PR trying to recover the previous url but it adds lot of complexity and I can not assure the stability. https://github.com/SUSE/spacewalk/pull/27921/files It requires some protective mechanics to avoid having the after step running for a long time.

@maximenoel8 I like your [PR](https://github.com/SUSE/spacewalk/pull/27921/), that's exactly what I had in mind 👍🏼 
I think we need to keep working on this, please don't discard it.
@nodeg please let's consider to re-open https://github.com/SUSE/spacewalk/issues/26585 or open a similar one to track Maxime's proposal, and only close it once we can achieve a stable version of the mechanism.

For now, we can merge this temporary fix.

Bischoff:
> Approving as temporary fix. But I think we need to keep working on [SUSE/spacewalk#27921]
> (https://github.com/SUSE/spacewalk/pull/27921)

We tried it, and abandoned it. It was adding ~ 2 hours to each test suite run. Restarting the webdriver on crashes like in this PR is a much better approach. I suppose @maximenoel8 forgot to close that draft.


> and keep open [SUSE/spacewalk#26585](https://github.com/SUSE/spacewalk/issues/26585) until we assure this is fixed.

That card is solved by this PR.

srbarrios:
> We tried it, and abandoned it. It was adding ~ 2 hours to each test suite run. Restarting the webdriver on crashes like in this PR is a much better approach. I suppose @maximenoel8 forgot to close that draft.

His draft also restart the crash like in this PR. It just try to go further, recovering previous state. Also, it handle much better the exceptions, as I suggested in a [in-line comment](https://github.com/uyuni-project/uyuni/pull/10657#discussion_r2238628375), we should keep into account that `Capybara.current_session` can be broken, and so `quit` might not be available. Maxime's PR handle this case.

Bischoff:
> His draft also restart the crash like in this PR. It just try to go further, recovering previous state. Also, it handle much better the exceptions, as I suggested in a [in-line comment](https://github.com/uyuni-project/uyuni/pull/10657#discussion_r2238628375),

There are several independant problems:

 * should we restart the driver at each feature?
 * could we cope by mechanisms at `session` level?
 * should we try to recover the error, for example by relogging and returning to current page?

Title of Maxime's drafs was "Reopen webdriver between features", that's the first problem in this list.

We tried all of the paths above before doing this PR. The outcome is:

 * restarting the driver at each feature is too slow and does not bring benefits;
 * using mechanisms at `session` level was not enough, we had to completely stop the driver;
 * maybe we could try to recover better, but it seemed complicated, fragile, and honestly not worth the effort.


> we should keep into account that `Capybara.current_session` can be broken, and so `quit` might not be available. Maxime's PR handle this case.

I'd suggest we first try to solve the problems we have, rather than trying to solve the problems we imagine.

If that really happens, we can revise.

PR #10651, opened 2025-07-28
CoCo attestation: support VLEK attestation report and new AMD cpu models in attestation settings

CDellaGiusta:
## What does this PR change?

In AWS, instances that run with SNP-SEV enabled do their attestation process using a new way for attestation (VLEK).

The Versioned Loaded Endorsement Key (VLEK) is a versioned signing key that is certified by AMD and used by the AMD CPU to sign the AMD SEV-SNP attestation reports. VLEK signatures can be validated using certificates provided by AMD.

This PR provides the following:

- adapts the state to perform fetching the VLEK certificate. If this fails, we assume VCEK.
- transfers the VLEK back to MLM Server together with the attestation report
- removes the "none" environment type (cpu), since it has no meaning anymore
- adds new processor models (bergamo, siena and turin)

With reference to https://confluence.suse.com/display/~CDellaGiusta/snpguest+version+differences

- the call to the creation of the attestation report is the same for all snpguest versions (0.9.1 not working on AWS instances at the moment, see https://bugzilla.suse.com/show_bug.cgi?id=1246910)
- if VLEK is not supported by the cpu/virtual environment, the certificate is collected as "null" and so we assume using VCEK
- downloading the VLEK certificate is not supported in snpguest version 0.3.2: as for the case above, the collected certificate is absent

## GUI diff

Before:
before

After:
after

- [x] **DONE**

## Documentation
- Documentation issue was created: https://github.com/uyuni-project/uyuni-docs/issues/4205
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Pairs with https://github.com/uyuni-project/uyuni/pull/10672

Issue(s): https://github.com/SUSE/spacewalk/issues/27154
Port(s): 
- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
[...]

PR #10681, opened 2025-08-01
CoCo verification: support VLEK attestation verification and new AMD cpu models

CDellaGiusta:
## What does this PR change?

In AWS, instances that run with SNP-SEV enabled do their attestation process using a new way for attestation (VLEK).

The Versioned Loaded Endorsement Key (VLEK) is a versioned signing key that is certified by AMD and used by the AMD CPU to sign the AMD SEV-SNP attestation reports. VLEK signatures can be validated using certificates provided by AMD.

This PR provides the following:
- to verify the VLEK, we need a different intermediate CA (asvk.pem) which we should integrate in our container
- adds new processor models (bergamo, siena and turin)

With reference to https://confluence.suse.com/display/~CDellaGiusta/snpguest+version+differences

- the call to the verification of the attestation report is the same for all snpguest versions (0.9.1 not working on AWS instances at the moment, see https://bugzilla.suse.com/show_bug.cgi?id=1246910)
- if VLEK is not supported by the cpu/virtual environment, the certificate is collected as "null" and so we assume using VCEK
- each cpu version have their VLEK intermediate certificate (asvk.pem)

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Pairs with https://github.com/uyuni-project/uyuni/pull/10651

Issue(s): https://github.com/SUSE/spacewalk/issues/27154
Port(s): 
- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test


CDellaGiusta:
Substitutes closed PR https://github.com/uyuni-project/uyuni/pull/10672

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapper.java
```diff
@@ -32,33 +32,46 @@
 /**
  * Wrapper to execute the command line tool SNPGuest.
  */
-public class SNPGuestWrapper {
+public abstract class SNPGuestWrapper {
```
mackdk: According to java standards, this should be called `AbstractSNPGuestWrapper`. Ideally, you should now extract a `SNPGuestWrapper` interface so that the code does not refer any specific implementation and it's easier to mock if needed.

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapper.java
```diff
@@ -32,33 +32,46 @@
...
 
-    private static final Marker STDOUT_MARKER = MarkerManager.getMarker("stdout");
+    protected static final Marker STDOUT_MARKER = MarkerManager.getMarker("stdout");
 
-    private static final Marker STDERR_MARKER = MarkerManager.getMarker("stderr");
+    protected static final Marker STDERR_MARKER = MarkerManager.getMarker("stderr");
 
-    private static final Logger LOGGER = LogManager.getLogger(SNPGuestWrapper.class);
+    protected static final Logger LOGGER = LogManager.getLogger(SNPGuestWrapper.class);
```
mackdk: IMO a logger in an abstract class is better implemented as a non-static field and initialized in the constructor as:

```java
    protected SNPGuestWrapper(Runtime runtimeIn) {
        this.runtime = runtimeIn;
        this.logger = LogManager.getLogger(getClass());
    }
```

In this way, all the logged entries will use the correct class name depending on the implementation used at runtime, otherwise the generic base class name will be used.

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapper.java
```diff
@@ -32,33 +32,46 @@
...
 
-    private final Runtime runtime;
+    protected final Runtime runtime;
 
     /**
      * Default constructor.
      */
-    public SNPGuestWrapper() {
+    protected SNPGuestWrapper() {
         this.runtime = Runtime.getRuntime();
```
mackdk: I know this is my code :-) but since we are improving it, it would be better to call the other constructor instead of duplicating the code:

```java
protected SNPGuestWrapper() {
        this(Runtime.getRuntime());
}
```

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
```diff
@@ -0,0 +1,113 @@
...
+
+package com.suse.coco.module.snpguest.execution;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+
+public class SNPGuestWrapperCreator {
```
mackdk: In java lingo, this class should be called `SNPGuestWrapperFactory`.

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
```diff
@@ -0,0 +1,113 @@
...
+                return new SNPGuestWrapperVer07Below();
+            }
+            case VER_09_ABOVE -> {
+                return new SNPGuestWrapperVer09Above();
+            }
+            default -> {
+                //fallback
+                return new SNPGuestWrapperVer07Below();
+            }
+        }
```
mackdk: From what I see, `theSnpguestVersion` can only be either `VER_07_BELOW` or `VER_09_ABOVE` after `findVersion()` is called so you can simply this to:

```java
        return switch (theSnpguestVersion) {
            case VER_07_BELOW -> new SNPGuestWrapperVer07Below();
            case VER_09_ABOVE -> new SNPGuestWrapperVer09Above();
        };
```

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private SNPGuestWrapperCreator() {
+        //Utility classes should not have a public or default constructor.
+    }
+
+    /**
+     * tries to find the snpguest tool version
+     */
+    public static void findVersion() {
+        synchronized (FIND_VERSION_LOCK) {
```
mackdk: I don't think you need to synchronize explicitly on a separate variable here since you are locking the whole method and you don't have any other locks in place you can just do:

```java
public static synchronized void findVersion() {
    ...
}
```

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
```diff
@@ -0,0 +1,113 @@
...
+            if (null != theSnpguestVersion) {
+                return;
+            }
+
+            try {
+                SNPGuestWrapper snpGuest = new SNPGuestWrapperVer07Below();
```
mackdk: Nitpicking here but, instantiating a specific instance to figure out what instance we need is not very clean :-) It might even break if they change the output of the version command (and we saw how much they care about backwards compatibility...).

IMO, the safer way would be to call `rpm` here to figure out what package is installed. This command will just return the version:

```
/usr/bin/rpm -q --queryformat '%{VERSION}\n' snpguest
```

microservices/coco-attestation/attestation-module-snpguest/src/test/java/com/suse/coco/module/snpguest/io/VerificationDirectoryProviderTest.java
```diff
@@ -71,15 +76,25 @@ public void tearDown() throws IOException {
...
 
-    @Test
-    @DisplayName("Verification directory is created by the provider and destroyed on resource closure")
-    void canCreateAndDestroyVerificationDirectory() throws IOException {
+    private static Stream<Arguments> listCpuAndUsingVlek() {
+        return TestHelper.listCpuAndUsingVlek();
+    }
+
+    @ParameterizedTest(name = TestHelper.CPU_USING_VLEK_NAME)
+    @MethodSource("listCpuAndUsingVlek")
```
mackdk: Here and everywhere you use `@MethodSource` I think you can use a fully qualified method name to avoid redefining the method multiple times:

```java
@MethodSource("com.suse.coco.module.snpguest.TestHelper#listCpuAndUsingVlek")
```

microservices/coco-attestation/attestation-module-snpguest/src/test/java/com/suse/coco/module/snpguest/SNPGuestWorkerTest.java
```diff
@@ -80,13 +84,35 @@ public void setup() {
...
+    }
+
+    private String getCertificateSuccessfulFetchString(boolean usingVlek) {
+
+        if (usingVlek) {
+            return """
+                    - VLEK certification retrieved successfully
+                    """;
+        }
+        else {
```
mackdk: This `else` is redundant and can be remove to avoid nesting

.github/workflows/changelogs.yml
```diff
@@ -5,6 +5,7 @@ on:
     types: [opened, edited, synchronize, reopened, closed]
     # Only execute on the main branches (feature branches are excluded)
     branches:
+      - master-staging # Temporary branch for staging until 5.1 GM
```
mackdk, line 8: This can be removed as `master-staging` no longer exists :-)

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapper.java
CDellaGiusta: Done

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapper.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/test/java/com/suse/coco/module/snpguest/io/VerificationDirectoryProviderTest.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/test/java/com/suse/coco/module/snpguest/SNPGuestWorkerTest.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperCreator.java
CDellaGiusta: it makes sense. Fixed

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapper.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/test/java/com/suse/coco/module/snpguest/execution/AbstractSNPGuestWrapperTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Mock
+    private Runtime runtime;
+
+    private AbstractSNPGuestWrapper wrapperVer07Below;
+    private AbstractSNPGuestWrapper wrapperVer09Above;
+
+    private AbstractSNPGuestWrapper getWrapperToTest(SNPGuestWrapperFactory.SNPGuestVersion type)
+            throws InvalidClassException {
+        switch (type) {
```
lucidd: i think this can be simplified the same way you have it in `SNPGuestWrapperFactory`

CDellaGiusta: good catch. Fixed

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/AbstractSNPGuestWrapper.java
```diff
@@ -32,31 +32,33 @@
...
+public abstract class AbstractSNPGuestWrapper {
 
-    private static final Marker STDOUT_MARKER = MarkerManager.getMarker("stdout");
+    protected static final Marker STDOUT_MARKER = MarkerManager.getMarker("stdout");
 
-    private static final Marker STDERR_MARKER = MarkerManager.getMarker("stderr");
+    protected static final Marker STDERR_MARKER = MarkerManager.getMarker("stderr");
 
-    private static final Logger LOGGER = LogManager.getLogger(SNPGuestWrapper.class);
+    protected Logger LOGGER;
```
mackdk: This field could be still final I guess, and it should be named lowercase. Not sure why our checkstyle does not include [the `MemberName` rule](https://checkstyle.org/checks/naming/membername.html) to enforce the standard naming conventions, maybe we can try to add it in a separate PR (not sure how many failures we would get in uyuni main codebase).

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperFactory.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        int exitCode = -1;
+        String versionString = "";
+        try {
+            // /usr/bin/rpm -q --queryformat '%{VERSION}\n' snpguest
+
+            Runtime runtime = Runtime.getRuntime();
+            String[] versionCommand = {"/usr/bin/rpm", "-q", "--queryformat", "%{VERSION}", "snpguest"};
+            Process process = runtime.exec(versionCommand);
```
mackdk, line 81: Nitpicking here, but similarly for what we do in the `AbstractSNPGuestWrapper` this `runtime` variable can be a field and initialized in the constructor. The change would allow to write a unit test for this factory, to verify the behaviour of the factory with specific outputs.

mackdk (approved):
LGTM :+1:

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/AbstractSNPGuestWrapper.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module-snpguest/src/main/java/com/suse/coco/module/snpguest/execution/SNPGuestWrapperFactory.java
CDellaGiusta: good point. Modification done

PR #10687, opened 2025-08-02
Remove BETA from product names

Bischoff:
## What does this PR change?

Remove BETA from product names

**IMPORTANT** if we are going to switch to 5.2 immediately in HEAD, this PR is wrong, and I will just close it. If that is the case, I will do another one, keeping BETA but jumping to 5.2.

In 5.1 branch, it's a different story.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

Cucumber tests were modified

- [x] **DONE**

## Links

Head and 5.1 only

Ports:
 * 5.1:  https://github.com/SUSE/spacewalk/pull/27988

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/reposync/srv_sync_channels.feature
+++ b/testsuite/features/reposync/srv_sync_channels.feature
@@ -37,8 +37,8 @@ Feature: List available channels and enable them
     When I execute mgr-sync "list products --expand"
     Then I should get "[ ] SUSE Linux Enterprise Server 15 SP4 x86_64"
     And I should get "[ ] SUSE Manager Proxy 4.3 x86_64"
-    And I should get "  [ ] (R) SUSE Multi-Linux Manager Client Tools for SUSE Liberty Linux 7, RHEL and clones 7 x86_64 (BETA)"
-    And I should get "  [ ] (R) SUSE Multi-Linux Manager Client Tools for SLE 15 x86_64 (BETA)"
+    And I should get "  [ ] (R) SUSE Multi-Linux Manager Client Tools for SUSE Liberty Linux 7, RHEL and clones 7 x86_64"
+    And I should get "  [ ] (R) SUSE Multi-Linux Manager Client Tools for SLE 15 x86_64"
 
   Scenario: List products with filter
     When I execute mgr-sync "list products --expand --filter x86_64"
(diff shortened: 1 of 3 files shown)


testsuite/features/reposync/srv_sync_channels.feature
```diff
@@ -37,8 +37,8 @@ Feature: List available channels and enable them
     When I execute mgr-sync "list products --expand"
     Then I should get "[ ] SUSE Linux Enterprise Server 15 SP4 x86_64"
     And I should get "[ ] SUSE Manager Proxy 4.3 x86_64"
-    And I should get "  [ ] (R) SUSE Multi-Linux Manager Client Tools for SUSE Liberty Linux 7, RHEL and clones 7 x86_64 (BETA)"
```
nodeg, line 40: I think for the regular CI, we want to keep the beta channels.

testsuite/features/reposync/srv_sync_products.feature
```diff
@@ -51,15 +51,15 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I open the sub-list of the product "SUSE Linux Enterprise Server 15 SP4 x86_64"
     And I open the sub-list of the product "Basesystem Module 15 SP4 x86_64"
     And I open the sub-list of the product "Desktop Applications Module 15 SP4 x86_64"
-    And I open the sub-list of the product "SUSE Multi-Linux Manager Client Tools for SLE 15 x86_64 (BETA)" if present
```
nodeg, line 54: Same for this file.

nodeg (changes requested):

testsuite/features/reposync/srv_sync_channels.feature
Bischoff: Yes, but we don't have them right now and this makes the test suite fail.

We can keep having Head fail at core until we have 5.2 beta channels - if that's not too long from now.

testsuite/features/reposync/srv_sync_products.feature
Bischoff: Same explanation.

I think we need some advice from Marina.

Bischoff:
After getting feedback from Marina, we will wait until 5.2 is ready, which should happen soon.

Head test suite will fail in the meantime. Closing the PR.

PR #10702, opened 2025-08-08
Simplify action handling (step 5 - change action type "if" clauses to action object methods)

CDellaGiusta:
## What does this PR change?

Every chunk of code where there is an "if" clause on Action type is a "code smell" that object-oriented paradigm is implemented correctly.

in fact, the following chunk of code shows a particular behaviour related to a specific action type:
```

if (action.getActionType().equals(ActionFactory.SOME_ACTION_TYPE))  {
   ... specific behaviour
}

```
in OO paradigm, this should be moved to

`action.doBehaviour(.... arguments)`

where  on the base class:

```
Action.doBehaviour(.... arguments) {
  //default does nothing
}
```
and on the specific derived class:

```
@Override
SomeActionTypeAction.doBehaviour(.... arguments) {
   ... specific behaviour
}
```

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links

Port(s): 5.1: https://github.com/SUSE/spacewalk/pull/28113
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/action/Action.java
+++ b/java/code/src/com/redhat/rhn/domain/action/Action.java
@@ -14,10 +14,12 @@
  */
 package com.redhat.rhn.domain.action;
 
+import com.redhat.rhn.common.localization.LocalizationService;
 import com.redhat.rhn.common.util.StringUtil;
 import com.redhat.rhn.domain.BaseDomainHelper;
 import com.redhat.rhn.domain.action.server.ServerAction;
 import com.redhat.rhn.domain.org.Org;
+import com.redhat.rhn.domain.server.MinionServerFactory;
 import com.redhat.rhn.domain.server.MinionSummary;
 import com.redhat.rhn.domain.server.Server;
 import com.redhat.rhn.domain.user.User;
@@ -31,6 +33,7 @@ import com.suse.salt.netapi.calls.LocalCall;
 
 import com.google.gson.JsonElement;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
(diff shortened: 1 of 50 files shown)


java/code/src/com/redhat/rhn/domain/action/rhnpackage/PackageRemoveAction.java
```diff
@@ -65,4 +66,12 @@ public Map<LocalCall<?>, List<MinionSummary>> getSaltCalls(List<MinionSummary> m
         return ret;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<Map<String, String>> createActionSpecificDetails(ServerAction serverAction) {
+        return createPackageActionSpecificDetails();
+    }
```
mcalmer, line 77: Why this? Wouldn't it be possible to rename `createPackageActionSpecificDetails()` to `createActionSpecificDetails(ServerAction serverAction)` and remove this from all classes which are derived from PackageAction?

java/code/src/com/redhat/rhn/domain/action/Action.java
```diff
@@ -580,5 +580,13 @@ public boolean shouldCleanupAction(Date bootTime, ServerAction sa) {
     public boolean canBeScheduledAnyway() {
         return false;
     }
+
+    /**
+     * Returns the pkg_parameter parameter to the schedule_action queries in Action_queries.xml
+     * @return a parameter value
+     */
+    public String getPackageParameter() {
+        return "upgrade";
```
mcalmer: This is how the original does it, but i wonder if we should improve it.
Return on every action for this method "upgrade" is wrong as only package actions should return something. 
Maybe return here just an empty string, and overwrite in PackageAction with "upgrade" and for "lock" we have it already

java/code/src/com/redhat/rhn/domain/action/rhnpackage/PackageRemoveAction.java
CDellaGiusta: The method `public List<Map<String, String>> createActionSpecificDetails(ServerAction serverAction)` is overridden from the class `Action` and by default returns an empty list.
Not all `PackageAction` derived classes do override it: actually only `PackageRemoveAction`, `PacakgeUpdateAction` and `PackageVerify` actions are overriding it: all 3 classes override it in the same way, therefore calling `createPackageActionSpecificDetails()` which is residing in the parent `PackageAction` class.
Other package action derived classes (`PackageLockAction, PackageDeltaAction, PackageRunTransactionAction, PackageRefreshListAction, PackageAutoUpdateAction`) do not override this method (hence returning the empty list)

java/code/src/com/redhat/rhn/domain/action/Action.java
CDellaGiusta: overridden getPackageParameter in class PackageAction, put the default one to empty string

java/code/src/com/redhat/rhn/domain/action/rhnpackage/PackageRemoveAction.java
mcalmer: Ok, if we have more methods which use the default than a special handling this makes sense

CDellaGiusta:
> Code looks good, but please check the sonarcloud suggestions. On the first look there are 2 which can be fixed easily.
done.

PR #10724, opened 2025-08-13
Simplify action handling (step 6 - simplify ActionManager methods)

CDellaGiusta:
## What does this PR change?
The class ActionManager has an intricated number of method calls, grown during time, that have to be simplified and merged together to avoid code repetition.
Also, the creation of actions should be moved towards the ActionFactory object

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


mcalmer:
I wonder if it would be possible to unify the way how an action is "setup".
We have "createAction" which just create the object.
We have ""createScheduledAction" which create and schedule it.
We have also "scheduleForExecution" which only schedule an action.
And we have tons of functions which create and/or schedule a specific Action type.

But maybe this is something for "step 7....N" ?

CDellaGiusta:
> But maybe this is something for "step 7....N" ?

Yes, that was my intention. My suspiction is that some of that methods have the same final effect, but they have been created in different times. So the idea was as first to simplify as much as possible, then do another chunk of modifications but with a "spare pair of eyes"
In particular, my guess is that `addServerToAction` and `createAddServerAction` are doing more or less the same thing, so they can be merged into one.
But I'd like to further do it in another step

java/code/src/com/redhat/rhn/domain/action/errata/ErrataAction.java
```diff
@@ -39,6 +39,7 @@
 import com.suse.salt.netapi.calls.LocalCall;
 import com.suse.salt.netapi.calls.modules.State;
 
+import java.util.ArrayList;
 import java.util.ArrayList;
```
mcalmer: 2 times the same import?

CDellaGiusta:
> There is one sonarcloud issue which should be easy to fix. But the rest looks good

fixed before squashing

PR #10729, opened 2025-08-14
Simplify/improve ContentSyncManager

CDellaGiusta:
## What does this PR change?
With reference to https://github.com/SUSE/spacewalk/issues/22584, this PR aims at fixing the external file lock.
The locking mechanism is external of the class implementation. This means that anyone not fully aware of how the process works might call any of the methods without locking, thus leading to concurrency problems and potential inconsistent states.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/22584
Port(s): 5.1: https://github.com/SUSE/spacewalk/pull/28116
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
+++ b/java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
@@ -18,6 +18,7 @@ import com.redhat.rhn.GlobalInstanceHolder;
 import com.redhat.rhn.common.conf.Config;
 import com.redhat.rhn.common.conf.ConfigDefaults;
 import com.redhat.rhn.common.hibernate.HibernateFactory;
+import com.redhat.rhn.common.util.FileLocks;
 import com.redhat.rhn.common.util.TimeUtils;
 import com.redhat.rhn.domain.channel.Channel;
 import com.redhat.rhn.domain.channel.ChannelFactory;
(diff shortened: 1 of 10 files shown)


OAOA
```diff
@@ -0,0 +1,84 @@
+[1mdiff --git a/java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java b/java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java[m
```
CDellaGiusta: garbage file.... removed at the last commit

java/code/src/com/redhat/rhn/manager/content/ContentSyncManager.java
```diff
@@ -2746,4 +2766,33 @@ public boolean canSyncToolsChannelViaCloudRMT() {
...
+                    HibernateFactory.commitTransaction();
+                    HibernateFactory.closeSession();
+
+                    updateRepositories(null);
+                    HibernateFactory.commitTransaction();
+                    HibernateFactory.closeSession();
+
+                    updateSubscriptions();
+                    HibernateFactory.commitTransaction();
+                    HibernateFactory.closeSession();
```
mcalmer: Will this work as every method try to lock on its own?

CDellaGiusta: > Will this work as every method try to lock on its own?

I wrongly assumed that FileLocks could automatically handle recursive lock attempts from the same process.
That assumption is wrong, my fault.
I added a small test to highlight this behaviour and fixed the issue with ContentSyncManager

PR #10743, opened 2025-08-19
Ssm product migration

mackdk:
## What does this PR change?

This PR allows the user to schedule a product migration from the UI on multiple servers using system set manager

## GUI diff

After:
image
image
image
image
image
image
image
image

- [X] **DONE**

## Documentation
- Documentation WIP

- [ ] **DONE**

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25369
Port(s): https://github.com/SUSE/spacewalk/pull/28119

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


mackdk:
SonarCloud issues are not related to the code changed/introduced by this PR. I will see if they can be addressed in a separate PR.

java/code/src/com/suse/utils/test/ListsTest.java
```diff
@@ -86,4 +86,34 @@ public void testNullSafetyUnion() {
...
     }
+
+    @Test
+    public void testMerge() {
+        List<Integer> result = Lists.merge(List.of(1, 2, 3), List.of(2, 3, 3, 5));
+        assertEquals(List.of(1, 2, 3, 5), result);
+    }
+
+    @Test
+    public void testNullSafetyMerge() {
```
CDellaGiusta, line 97: Worth doing the same test with the empty list instead of null?

java/code/src/com/redhat/rhn/domain/action/dup/DistUpgradeAction.java
```diff
@@ -69,44 +66,70 @@
...
+                .stream()
+                .collect(Collectors.partitioningBy(
+                    ct -> ct.getTask() == DistUpgradeChannelTask.SUBSCRIBE,
+                    Collectors.mapping(DistUpgradeChannelTask::getChannel, Collectors.toList())
                 ));
 
-        List<Channel> subbed = collect.get(true);
-        List<Channel> unsubbed = collect.get(false);
+            List<Channel> subbed = channelTaskMap.get(true);
+            List<Channel> unsubbed = channelTaskMap.get(false);
```
CDellaGiusta: maybe `subscribedChannels` and `unsubscribedChannels`?

java/code/src/com/redhat/rhn/domain/action/dup/DistUpgradeAction.java
```diff
@@ -157,30 +183,41 @@ public Map<LocalCall<?>, List<MinionSummary>> getSaltCalls(List<MinionSummary> m
...
-            List<Channel> unsubbed = collect.get(false);
+            Map<Boolean, List<Channel>> channelTaskMap = details.getChannelTasks()
+                .stream()
+                .collect(Collectors.partitioningBy(
+                    ct -> ct.getTask() == DistUpgradeChannelTask.SUBSCRIBE,
+                    Collectors.mapping(DistUpgradeChannelTask::getChannel, Collectors.toList())
+                ));
+
+            List<Channel> subbed = channelTaskMap.get(true);
+            List<Channel> unsubbed = channelTaskMap.get(false);
```
CDellaGiusta: ditto maybe `subscribedChannels` and `unsubscribedChannels`?

mcalmer (dismissed):
I have not checked the javascript code.
I think we are ok for the rest.

java/code/src/com/redhat/rhn/domain/action/dup/DistUpgradeAction.java
```diff
@@ -192,22 +229,21 @@ public void handleUpdateServerAction(ServerAction serverAction, JsonElement json
...
+                    distUpgradeResult.getSpmigration().getChanges().getRetOpt().orElse(""),
+                    distUpgradeResult.getSpmigration().getComment()
+                );
             }
         }
         catch (JsonSyntaxException e) {
             try {
-                DistUpgradeOldSlsResult distUpgradeSlsResult = Json.GSON.fromJson(
-                        jsonResult, DistUpgradeOldSlsResult.class);
+                var oldFormatResult = Json.GSON.fromJson(jsonResult, DistUpgradeOldSlsResult.class);
```
CDellaGiusta, line 242: Question: Is there a reason why this has been turned to var ?

web/spacewalk-web.spec
```diff
@@ -63,7 +63,7 @@ but it does generate a number of sub-packages.
 
 %package -n spacewalk-html
 Summary:        HTML document files for Spacewalk
-License:        (GPL-3.0-only OR MIT) AND (MPL-2.0 OR Apache-2.0) AND 0BSD AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND CC-BY-4.0 AND GPL-2.0-only AND ISC AND LGPL-3.0-or-later AND MIT AND MPL-2.0 AND OFL-1.1 AND Python-2.0
+License:        (GPL-3.0 OR MIT) AND (MPL-2.0 OR Apache-2.0) AND 0BSD AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND CC-BY-4.0 AND GPL-2.0-only AND ISC AND LGPL-3.0-or-later AND MIT AND MPL-2.0 AND OFL-1.1 AND Python-2.0
```
mcalmer: I lost a bit the overview what we need and what not.
Is it correct that we now lost the "only" part for GPL-3.0?

mackdk: Yes, I'm not sure were this change comes from. Calling `yarn build` will automatically add to the spec file this change. I haven't changed or added any dependency. If I omit it, the  javascript build check fails asking for the file to be committed.

java/code/src/com/redhat/rhn/manager/distupgrade/DistUpgradeManager.java
```diff
@@ -906,4 +921,128 @@ private static List<SUSEProductSet> removeIncompatibleTargets(Optional<SUSEProdu
...
+            .sorted(DistUpgradeManager.PRODUCT_SET_VERSION_COMPARATOR)
+            .toList();
+    }
+
+    // Returns true if the given server has the specified base product installed
+    private static boolean serverHasBaseProduct(Server server, SUSEProduct sourceBaseProduct) {
+        return server.getInstalledProductSet()
+            .map(SUSEProductSet::getBaseProduct)
+            .filter(installedBaseProduct -> sourceBaseProduct.equals(installedBaseProduct))
+            .isPresent();
```
CDellaGiusta, line 1023: can `anyMatch` be used instead?

java/code/src/com/redhat/rhn/manager/distupgrade/DistUpgradeManager.java
```diff
@@ -906,4 +921,128 @@ private static List<SUSEProductSet> removeIncompatibleTargets(Optional<SUSEProdu
...
+     */
+    public static Set<SUSEProduct> listInstalledTargetAddons(Optional<SUSEProductSet> installedProductSet,
+                                                             List<SUSEProductSet> migrationTargets) {
+
+        Set<SUSEProduct> result = new HashSet<>();
+        List<SUSEProduct> installedAddons = installedProductSet.map(SUSEProductSet::getAddonProducts).orElse(List.of());
+
+        for (SUSEProductSet target : migrationTargets) {
+            // Find all the targets that don't have a matching base
+            target.getAddonProducts().stream()
```
CDellaGiusta: could we do something like
migrationTargets.stream().map(target -> target.getAddonProducts()).filter(.... ?

java/code/src/com/suse/manager/webui/controllers/MinionsAPI.java
```diff
@@ -729,4 +753,90 @@ private String scheduleAllCoCoAttestation(Request request, Response response, Us
 
         return json(GSON, response, result.longValue());
     }
+
+
+    private String computeMigrationChannels(Request request, Response response, User user) {
+        var dataFactory = new MigrationDataFactory();
```
CDellaGiusta: worth setting `dataFactory` as a private final class member?

java/code/src/com/suse/manager/webui/controllers/MinionsAPI.java
```diff
@@ -729,4 +753,90 @@ private String scheduleAllCoCoAttestation(Request request, Response response, Us
...
+
+        var migrationChannelsRequest = GSON.fromJson(request.body(), MigrationChannelsRequest.class);
```
CDellaGiusta: method `scheduleMigration` below has this line surrounded by` try/catch`. Worth doing here as well?

java/code/src/com/suse/manager/webui/controllers/MinionsAPI.java
```diff
@@ -729,4 +753,90 @@ private String scheduleAllCoCoAttestation(Request request, Response response, Us
...
+
+        List<MinionServer> serverList;
+
+        SUSEProductSet source;
+        SUSEProductSet target;
```
CDellaGiusta: can these 3 variables go inside the try block?

schema/spacewalk/common/data/endpoint.sql
```diff
@@ -3587,6 +3587,16 @@ INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_re
 INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
     VALUES ('', '/manager/sso/sls', 'GET', 'W', False)
     ON CONFLICT (endpoint, http_method) DO NOTHING;
+INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
+    VALUES ('', '/manager/systems/ssm/product-migration', 'GET', 'W', True)
```
CDellaGiusta, line 3591: question: is 'W' the right value for this GET request?

schema/spacewalk/common/data/endpointNamespace.sql
```diff
@@ -5186,6 +5186,21 @@ INSERT INTO access.endpointNamespace (namespace_id, endpoint_id)
     WHERE ns.namespace = 'systems.vhms' AND ns.access_mode = 'R'
     AND ep.endpoint = '/manager/api/vhms/kubeconfig/:id/contexts' AND ep.http_method = 'GET'
     ON CONFLICT DO NOTHING;
+INSERT INTO access.endpointNamespace (namespace_id, endpoint_id)
+    SELECT ns.id, ep.id FROM access.namespace ns, access.endpoint ep
+    WHERE ns.namespace = 'systems.migration' AND ns.access_mode = 'W'
```
CDellaGiusta, line 5191: ditto

schema/spacewalk/upgrade/susemanager-schema-5.2.0-to-susemanager-schema-5.2.1/202-ssm-rbac-endpoints.sql
```diff
@@ -0,0 +1,34 @@
+
+INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
+    SELECT '', '/manager/systems/ssm/product-migration', 'GET', 'W', True
```
CDellaGiusta, line 3: ditto

schema/spacewalk/upgrade/susemanager-schema-5.2.0-to-susemanager-schema-5.2.1/202-ssm-rbac-endpoints.sql
```diff
@@ -0,0 +1,34 @@
...
+    SELECT '', '/manager/api/systems/migration/schedule', 'POST', 'W', True
+    WHERE NOT EXISTS (SELECT 1 FROM access.endpoint WHERE endpoint = '/manager/api/systems/migration/schedule' AND http_method = 'POST');
+
+INSERT INTO access.namespace (namespace, access_mode, description)
+    SELECT 'systems.migration', 'W', 'Perform product migration on systems'
+    WHERE NOT EXISTS (SELECT 1 FROM access.namespace WHERE namespace = 'systems.migration' AND access_mode = 'W');
+
+INSERT INTO access.endpointNamespace (namespace_id, endpoint_id)
+    SELECT ns.id, ep.id FROM access.namespace ns, access.endpoint ep
+    WHERE ns.namespace = 'systems.migration' AND ns.access_mode = 'W'
```
CDellaGiusta, line 24: ditto

java/code/src/com/suse/manager/model/products/migration/MigrationDataFactory.java
```diff
@@ -97,6 +102,69 @@ public MigrationTargetSelection toMigrationTargetSelection(List<MinionServer> se
...
+            .allMatch(BooleanUtils::isTrue);
+
+        Channel targetBaseChannel = findBaseChannel(newChannels);
+
+        // Get name of original base channel if channel is cloned
+        String originalName = ChannelManager.getOriginalChannel(targetBaseChannel).getName();
+        SUSEProduct targetBaseProduct = SUSEProductFactory.lookupByChannelName(originalName).get(0).getRootProduct();
+
+        // Extract the source product from the first system, they must be matching
+        var sourceSet = serverList.get(0).getInstalledProductSet();
```
CDellaGiusta, line 133: can `serverList` ever be empty?

schema/spacewalk/common/data/endpoint.sql
```diff
@@ -3590,6 +3590,9 @@ INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_re
 INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
     VALUES ('', '/manager/systems/ssm/product-migration', 'GET', 'W', True)
     ON CONFLICT (endpoint, http_method) DO NOTHING;
+INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
+    VALUES ('', '/manager/systems/ssm/product-migration/dry-run/:actionId', 'GET', 'W', True)
```
CDellaGiusta, line 3594: ditto

CDellaGiusta (approved):
java part looks good. No experience in UI code

java/code/src/com/redhat/rhn/manager/distupgrade/DistUpgradeManager.java
mackdk: This is an `Optional`, so it doesn't have a method `anyMatch()`. I could convert it to a stream to call it, not sure what would be better

java/code/src/com/suse/manager/model/products/migration/MigrationDataFactory.java
mackdk: In theory no, since you won't be able to reach this code if there are no server selected in SSM and a previous dry run was not executed. Maybe I can enforce the check somewhere before in the code

java/code/src/com/redhat/rhn/domain/action/dup/DistUpgradeAction.java
mackdk: To avoid line wrapping, because the type is already specified in the `fromJson()` call.

CDellaGiusta:
> SonarCloud issues are not related to the code changed/introduced by this PR. I will see if they can be addressed in a separate PR.

Unfortunately even if you modify a little thing in a method, SonarCloud blames you for all the original sins.
We are going to discuss what rules to keep and drop in a meeting when we're back from holidays. Cognitive complexity is one of the hot topics there. I suggest to leave these errors there at the moment.

schema/spacewalk/common/data/endpoint.sql
mackdk: While the method itself is not modifying anything, the feature itself is. I can change to R for all the calls that are not actually modifying the data, but the web page will be kind of useless and broken, because the final call to migrate will be impossible to execute. So, not sure what would be the best approach here.

CDellaGiusta: > While the method itself is not modifying anything, the feature itself is. 

ok, I thought it was a slip.

java/code/src/com/suse/utils/test/ListsTest.java
mackdk: Tests added :+1:

java/code/src/com/redhat/rhn/manager/distupgrade/DistUpgradeManager.java
mackdk: Done :+1:

java/code/src/com/suse/manager/webui/controllers/MinionsAPI.java
mackdk: Done

java/code/src/com/suse/manager/webui/controllers/MinionsAPI.java
mackdk: Done

java/code/src/com/suse/manager/webui/controllers/MinionsAPI.java
mackdk: Done

java/code/src/com/redhat/rhn/domain/action/dup/DistUpgradeAction.java
mackdk: Names changed

java/code/src/com/redhat/rhn/domain/action/dup/DistUpgradeAction.java
mackdk: Names changed

web/html/src/components/input/radio/Radio.tsx
```diff
@@ -40,11 +54,15 @@ export function Radio(props: Props) {
...
+        if (customRadioClass !== undefined) {
+          radioClass = radioClass + " " + customRadioClass;
+        }
+
         return (
           <span className={styles.radio}>
-            {props.items.map(({ label, value }) => (
-              <label className={radioClass} key={`${props.name}_${value}`}>
+            {props.items.map(({ label, value, title, disabled }) => (
+              <label className={radioClass} key={`${props.name}_${value}`} title={title}>
```
Etheryte, line 58: @bisht-richa If we want to use Bootstrap tooltips from your recent update here, do we need to do anything additional or will this work out of the box?

web/html/src/components/input/radio/Radio.tsx
```diff
@@ -6,16 +6,30 @@ import { FormContext } from "../form/Form";
...
+  items: Array<RadioOption>;
 
   /** Show the choices in a line or not */
   inline?: boolean;
 
   /** Whether to let the user input another value than the proposed ones */
   openOption?: boolean;
 
+  /** CSS class for the <label> element wrapping the <input> */
+  customRadioClass?: string;
```
Etheryte: Would recommend `labelClass`, that way it's similar to the other props.

web/html/src/components/product-migration/utils.tsx
```diff
@@ -0,0 +1,128 @@
...
+ * @param baseChannelTrees the array of channel trees
+ * @param mandatoryMap a record mapping the requirements for each channel
+ * @returns an object containing:
+ *  - channelTrees, a clone of the baseChannelTree with all the fields of {@link BaseChannelType} and {@link ChildChannelType} populated
+ *  - channelsMap, a map of {@link ChannelType} where the key is the channel id
+ *  - baseChannels, an array containing all the base channels
+ *  - requiresMap a mapping detailing for each channel id which other channels it requries
+ *  - requiredByMap a mapping detailing for each channel id, which other channels requires it
+ */
+function processChannelData(baseChannelTrees: ChannelTreeType[], mandatoryMap: Record<string, number[]>) {
```
Etheryte, line 28: I don't think this logic belongs on the frontend. What is it and why do we need to do it here?

web/html/src/components/product-migration/MigrationTargetSelectorForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+  migrationTargets,
+  targetId,
+  onTargetChange,
+}): JSX.Element => {
+  const firstSelectableTarget = migrationTargets.find((item) => item.missingChannels.length === 0);
+  const [formModel, setFormModel] = useState<FormModel>(
+    firstSelectableTarget !== undefined ? { selectedTarget: targetId ?? firstSelectableTarget.id } : {}
+  );
+
+  const onSubmit = useCallback(async (): Promise<void> => {
```
Etheryte: Here and elsewhere, why do we need to use `useCallback` here?

Etheryte:
Now that I've read through the whole pull request, I understand some of the decisions better. I know that we're rehashing a lot of older code that's already there, however if it was possible, it would be strongly advisable to not do all the data munging on the frontend. Many of our clients have a lot of data in their production systems and all of this processing will be done on the main thread, meaning the page will not be responsive while it happens. Memoing and similar help, but they don't really solve the problem.  

I understand this needs to go into 5.1, but after that I think we should take a step back and look at what we're doing. Most of the channel selection feature has grown organically over time, refactored as clients hit either bugs or performance issues, but at the scale we do things, I don't think the solution we've reached makes sense anymore.

web/html/src/components/input/radio/Radio.tsx
mackdk: ~Done~

web/html/src/components/product-migration/utils.tsx
mackdk: This logic is required for the channel display component, which needs dependency and reverse-dependency maps constructed with the complete `BaseChannelType` and `ChildChannelType` objects. The component uses this data structure to identify and show recommended and mandatory channels based on their parent-child relationships and the `recommended` field.

I will look into moving as much of this logic to the backend. However, the existing implementation of the components `BaseChannel` and `ChildChannel` ultimately needs this structure, which has an inherent circular dependency. Because this circular reference can't be represented directly in JSON from the server, some client-side processing is unavoidable.

web/html/src/components/product-migration/MigrationTargetSelectorForm.tsx
mackdk: I thought it was a good practice to use `useCallback` for functions that depend on state and/or props. If it is not needed/overkill I can remove it.

web/html/src/components/input/radio/Radio.tsx
mackdk: Actually this is a different class. With the `Radio` component there are multiple labels. There's the real label of every input component inherted from `InputBase`, and there is another label for each radio value. This is the class I needed to modify. The structure is:

```
<div class="form-group">
    <label class="control-label {labelClass}>...</label>
    <div class="{divClass}">
        <label class="{customRadioClass}">
            <input type="radio">
            Option 1
        </label>
        
        <label class="{customRadioClass}">
            <input type="radio">
            Option 2
        </label>
        
        ...
    </div>
</div>
```

On second look the naming is indeed very bad. Maybe I can avoid adding the property and use some CSS to style my component.

web/html/src/components/product-migration/MigrationTargetSelectorForm.tsx
Etheryte: While a lot of this does come down to preference, most of the time there will be little to no benefit, or even negative benefit, depending on the situation. There are of course times when it is useful, but memoizing in React land is tricky and has many footguns, so I would advise to use it only when you've made sure you need it.

To exemplify what I mean, if you have

```tsx
const MigrationTargetSelectorForm = (props) => {
  const onSubmit = useCallback(() => {
    ...
  }, [formModel, props.onTargetChange]);

  ...
}
```

but then it's used like this

```tsx
const onTargetSelection = () => { ... };

return <MigrationTargetSelectorForm onTargetSelection={onTargetSelection} />;
```

Then the dependency array for `useCallback` will invalidate on every render and there is actually an extra overhead without any benefit. (Because `onTargetSelection` will be a "new" function on every render).

web/html/src/components/input/radio/Radio.tsx
mackdk: I now removed this additional property. I was able to apply the styling I needed inside the component that is using the radio.

web/html/src/components/product-migration/MigrationTargetSelectorForm.tsx
mackdk: I removed the usages of `useCallback`

web/html/src/components/product-migration/utils.tsx
mackdk: I tried to reduce the amount for processing and logic on the fronted in a separate PR, since the change was getting pretty big: https://github.com/uyuni-project/uyuni/pull/10766.

web/html/src/components/input/radio/Radio.tsx
bisht-richa: You need to add the attribute `data-bs-toggle="tooltip"` to the element.

PR #10767, opened 2025-08-22
Add gpgverify module for signature verification

agraul:
## What does this PR change?

Previously, GPG signature verification was done by looking at the returncode of `gpg --verify`. That approach has a shortcoming, it requires that all signatures pass the verification to exit 0. Generally, that's a good approach, but in our use case it causes a reocurring issue: whenever Debian release a new version, the previous version's metadata get signed by a new GPG key. That is done in addition to the previous version's GPG key. Reposync fails the verification of existing Debian repositories due to the new key being unknown.

This new module parses the `gpg --verify` output to check each signature individually. If one passes, we consider the check successful.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28088

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


python/test/unit/spacewalk/common/test_repo.py
agraul, line 1: I need to update these tests as well, they come from a time when I looked at `gpgv`

python/spacewalk/common/gpgverify.py
```diff
@@ -0,0 +1,200 @@
...
+
+import logging
+import os
+import subprocess
+import typing
+from datetime import datetime, timezone
+
+logger = logging.getLogger(__name__)
+
+Path: typing.TypeAlias = typing.Union[str, os.PathLike[str]]
```
agraul: Looks like os.PathLike[str] does not work in Python 3.6, will fix tomorrow

python/spacewalk/common/repo.py
```diff
@@ -259,109 +261,67 @@ def _parse_release_index(self, release: str) -> "EntryDict":
...
         return self._release
 
-    def _has_valid_gpg_signature(self, uri: str, response=None) -> bool:
-        """
-        Validate GPG signature of Release file.
-
-        :return: bool
+    def _write_to_tempfile(self, content: bytes) -> str:
+        "Write content to a NamedTemporaryFile, rewind, and return the file name."
+        file = tempfile.NamedTemporaryFile()
```
ycedres: I wonder if there could be a problem with the `file` object being garbage collected too soon as it is local and the `delete` parameter for `NamedTemporaryFile` defaults to `True`.

agraul: I need to double check this. I return `file.name`, but I guess that will be a copy of the string and not a reference that keeps the garbage collector away. `delete` just means that file is deleted when it's closed. We can turn that off, the gc cleanup should be enough.

agraul: You're right, good catch!

I tried a minimal reproducer as follows:
```python
import tempfile
import time

def _write_to_tempfile(content: bytes) -> str:
    "Write content to a NamedTemporaryFile, rewind, and return the file name."
    file = tempfile.NamedTemporaryFile()
    file.write(content)
    file.seek(0)
    return file.name

def _write_to_tempfile2(content: bytes) -> str:
    "Write content to a NamedTemporaryFile, rewind, and return the file object."
    file = tempfile.NamedTemporaryFile()
    file.write(content)
    file.seek(0)
    return file

def caller():
    content = b"hello world\n"
    fname = _write_to_tempfile(content)
    print(f"file.name at: {fname}")
    file = _write_to_tempfile2(content)
    print(f"file at: {file.name}")
    time.sleep(600)

caller()
```

and here this is how they differ when used:

```sh
% python3 x.py  
file.name at: /tmp/tmp8mdn2hlc
file at: /tmp/tmpd1ekl_k1
# process keeps sleeping, from another terminal:
% cat /tmp/tmp8mdn2hlc /tmp/tmpd1ekl_k1
cat: /tmp/tmp8mdn2hlc: No such file or directory
hello world
```

PR #10774, opened 2025-08-25
testsuite: https connections tests added

alichiel:
## What does this PR change?

Adding basic https connection tests. It also check the "title" of the main page.

## GUI diff

No difference.

## Documentation
- No documentation needed.

## Changelog
- No changelog needed.

# Before you merge

### DIFF
--- a/testsuite/features/support/network_utils.rb
+++ b/testsuite/features/support/network_utils.rb
@@ -3,8 +3,8 @@
 
 require 'net/scp'
 require 'net/ssh'
-require 'stringio'
 require 'openssl'
+require 'stringio'
 
 Net::SSH::Transport::Algorithms::ALGORITHMS.each_value { |algs| algs.reject! { |a| a.match(/^ecd(sa|h)-sha2/) } }
 Net::SSH::KnownHosts::SUPPORTED_TYPE.reject! { |t| t.match(/^ecd(sa|h)-sha2/) }
@@ -116,28 +116,29 @@ end
 
 # This helper method generates dummy CA certificate
 #
-# @param [String] file name, it the CA certificate will be stored there.
+# @param [String] file name, the CA certificate will be stored there.
 # @param [String] CA certificate subject.
-def generate_dummy_cacert(filename, subject="/DC=localdomain/DC=localhost/CN=dummy CA")
+def generate_dummy_cacert(filename, subject = '/DC=localdomain/DC=localhost/CN=dummy CA')
   begin
-    root_key = OpenSSL::PKey::RSA.new 2048
+    root_key = OpenSSL::PKey::RSA.new(2048)
     root_ca = OpenSSL::X509::Certificate.new
     root_ca.public_key = root_key.public_key
-    root_ca.version = 2  # RFC 5280, "v3" certificate
+    # RFC 5280, "v3" certificate
+    root_ca.version = 2
     root_ca.serial = 1
     root_ca.subject = OpenSSL::X509::Name.parse(subject)
     root_ca.issuer = root_ca.subject
     root_ca.not_before = Time.now
     # 10 days
-    root_ca.not_after = root_ca.not_before + 864000
+    root_ca.not_after = root_ca.not_before + (60 * 60 * 24 * 10)
     ef = OpenSSL::X509::ExtensionFactory.new
     ef.subject_certificate = root_ca
     ef.issuer_certificate = root_ca
-    root_ca.add_extension(ef.create_extension("basicConstraints","CA:TRUE",true))
-    root_ca.add_extension(ef.create_extension("keyUsage","keyCertSign, cRLSign", true))
-    root_ca.sign(root_key, OpenSSL::Digest::SHA256.new)
-    File.open(filename, 'w') { |file| file.write(root_ca.to_pem) }
-  rescue => e
+    root_ca.add_extension(ef.create_extension('basicConstraints', 'CA:TRUE', true))
+    root_ca.add_extension(ef.create_extension('keyUsage', 'keyCertSign', 'cRLSign', true))
+    root_ca.sign(root_key, OpenSSL::Digest.new('SHA256'))
+    File.write(filename) { |file| file.write(root_ca.to_pem) }
+  rescue StandardError => e
     # issues to generate certificate
     puts e.message
   end
@@ -150,9 +151,9 @@ end
 def get_dummy_cacert(filename)
   begin
     certificate = File.read(filename)
-  rescue => e
+  rescue StandardError => e
     # issues to read certificate file
     puts e.message
   end
-  return(certificate)
+  certificate
 end


testsuite/features/secondary/https_connection.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  Scenario: Connecting to the server securely
+    When I connect to the server securely
+    Then the connection should be secured
+    And the page title should be "SUSE Multi-Linux Manager - Sign In" text
+
+  Scenario: Connecting to the server securely while using CA certificate file
+    When I connect to the server securely while using CA certificate file
+    Then the connection should be secured
+    And the page title should be "SUSE Multi-Linux Manager - Sign In" text
```
srbarrios: That will only work for MLM but not for Uyuni. I suggest that you find some other text that can apply for both.

testsuite/features/secondary/https_connection.feature
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2015-2025 SUSE LLC
```
srbarrios: That's a new one, so you don't need to write 2015-2025, we normally just keep the year of creation and year of last modification.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
+# Copyright 2017-2025 SUSE LLC.
```
srbarrios: Same. Just 2025.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+# Licensed under the terms of the MIT license.
+
+### SSL connection checks to server
+
+require 'open-uri'
+require 'uri'
+require 'openssl'
```
srbarrios: I think long term it gonna be easier to maintain and reuse code if we add new functions inside `testsuite/features/support/network_utils.rb`, and then you just call the functions in your new steps.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+
+When(/^I connect to the server insecurely$/) do
+  @url = "#{Capybara.app_host}".gsub("https:", "http:")
+  begin
+    @res = URI.open(@url, redirect: false, open_timeout: 5)
```
srbarrios: Ideally, we avoid using global/class/instance variables between steps.
But in case you need to pass information between steps, we use this helper: 

```
# Retrieves the value associated within the current feature scope context.
#
# @param key [Symbol] The key to retrieve the value for.
# @return [Object, nil] The value associated with the key, or nil if the key is not found.
def get_context(key)
  return unless $context.key?($feature_scope)

  $context[$feature_scope][key]
end

# Define or replace a key-value in the current feature scope context
#
# @param key [Object] The key to be added to the context hash.
# @param value [Object] The value to be associated with the key in the context hash.
def add_context(key, value)
  $context[$feature_scope] = {} unless $context.key?($feature_scope)
  $context[$feature_scope].merge!({ key => value })
end
```

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+# Licensed under the terms of the MIT license.
+
+### SSL connection checks to server
+
+require 'open-uri'
+require 'uri'
+require 'openssl'
+
+When(/^I connect to the server insecurely$/) do
+  @url = "#{Capybara.app_host}".gsub("https:", "http:")
```
srbarrios: In Ruby, for plain text, our code-style is to use single quotes.
So, `gsub('https:', 'http:')`

testsuite/features/step_definitions/https_connection_steps.rb
srbarrios: I guess we also need `, ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE` on `URI.open` method.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+  begin
+    @res = URI.open(@url, redirect: false, open_timeout: 5)
+  rescue => @e
+  end
+end
+
+Then(/^the connection should redirect to the secured channel$/) do
+  if not (@e.instance_of?(OpenURI::HTTPRedirect) and @e.message == "302 302") then
+    print("\tReturn code: #{@res.status[0]}\n\tURI: #{@res.base_uri.to_s}\n") if @res
```
srbarrios: We don't use `print` in Ruby functions but `puts`, I guess it was a matter of taste at the beginning, but I would say is code-style now.
Still, for Step definitions, as they are processed by Cucumber framework, we recommend to use `log` as that way the line it gonna appear on the Cucumber report. Otherwise, it's only printed on the console output.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+end
+
+When(/^I connect to the server securely$/) do
+  @url = "#{Capybara.app_host}".gsub("http:", "https:")
+  @res = URI.open(@url, redirect: false, open_timeout: 5, ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE)
+  visit('/')
+end
+
+Then(/^the connection should be secured$/) do
+  if not (@res and @res.status[0].to_i == 200)
```
srbarrios: We don't use `if not` but `unless`.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+    print("\tReturn code: #{@res.status[0]}\n\tURI: #{@res.base_uri.to_s}\n") if @res
+    raise ScriptError, "The return value is not OK (code: 200)."
+  end
+end
+
+And(/^the page title should be "(.*?)" text$/) do |page_title|
```
srbarrios: Better use `Then`

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+  root_ca.issuer = root_ca.subject
+  root_ca.not_before = Time.now
+  root_ca.not_after = root_ca.not_before + 864000  # 10 days
+  ef = OpenSSL::X509::ExtensionFactory.new
+  ef.subject_certificate = root_ca
+  ef.issuer_certificate = root_ca
+  root_ca.add_extension(ef.create_extension("basicConstraints","CA:TRUE",true))
+  root_ca.add_extension(ef.create_extension("keyUsage","keyCertSign, cRLSign", true))
+  root_ca.sign(root_key, OpenSSL::Digest::SHA256.new)
+  File.open(ssl_ca_cert_file, 'w') { |file| file.write(root_ca.to_pem) }
```
srbarrios: It would be good to move this piece of code as a function in `support/network_utils.rb`

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+  @url = "#{Capybara.app_host}".gsub("http:", "https:")
+  begin
+    @res = URI.open(@url, redirect: false, open_timeout: 5, ssl_ca_cert: ssl_ca_cert_file, ssl_verify_mode: OpenSSL::SSL::VERIFY_PEER)
+  rescue OpenSSL::SSL::SSLError => @e
```
srbarrios: I would say you can't set directly your variable here, if that was your intention.
Instead, if you really want to store it, you gonna need to do something like:

```
rescue OpenSSL::SSL::SSLError => error
  @e = error
end
```

Btw, matter of taste, but I bet for meaningful/readable variables, no worries if your variable name looks long, it's better than having difficulties to understand what exactly `@e` it's about, when you comeback here in 3 years.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+  visit('/')
+end
+
+When(/^I connect to the server securely while using incorrect certificate as a CA certificate file$/) do
+  ssl_ca_cert_file = '/tmp/dummy_CA.pem'
+  root_key = OpenSSL::PKey::RSA.new 2048
+  root_ca = OpenSSL::X509::Certificate.new
+  root_ca.public_key = root_key.public_key
+  root_ca.version = 2  # RFC 5280, "v3" certificate
```
srbarrios: I'm not 100% sure, but I think our lint might complain here.
I think it ask to don't put comments on the same code line.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: This is there because of the exception handling. Of course, there is dummy_ca generator, this could be moved.
In contrary if I removed that "require" and add it somewhere else, this test would be dependent on "somewhere else" which if changed, will cause this test to break. If I misunderstood your point, feel free to ping me.

testsuite/features/secondary/https_connection.feature
alichiel: Right. This needs to be ported to every version. MLM here is wrong one.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: It actually should not be required since the step here is connetcting to http and it prevents the redirection to https, the exception should raise the redirect error. But I will re-check.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: I was thinking to have it at some place like mentioned. On the other hand it's just about to have some "different" CA that has nothing to do with the server certificate. That's why I leave it as the part of the test.
If we need to do deeper cert magic, then I agree but also this CA shouldn't be that "dummy" ;).

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: About the storing the variable. Yes it is possible and works. However I'd use the recommended methods above, rather than global variables.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+    print("\tReturn code: #{@res.status[0]}\n\tURI: #{@res.base_uri.to_s}\n") if @res
+    raise ScriptError, "The connection has not redirected to the secure channel!"
+  end
+  print("\tException message: #{@e.message}\n")
+end
+
+When(/^I connect to the server securely$/) do
+  @url = "#{Capybara.app_host}".gsub("http:", "https:")
+  @res = URI.open(@url, redirect: false, open_timeout: 5, ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE)
+  visit('/')
```
srbarrios: This `visit` doesn't make to much sense here, as it's unrelated with URI.open and its returned value.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,84 @@
...
+
+When(/^I connect to the server securely while using CA certificate file$/) do
+  @url = "#{Capybara.app_host}".gsub("http:", "https:")
+  ssl_ca_cert_file = "/etc/ssl/certs/#{@url.gsub('https://', '')}.pem"
+  begin
+    @res = URI.open(@url, redirect: false, open_timeout: 5, ssl_ca_cert: ssl_ca_cert_file, ssl_verify_mode: OpenSSL::SSL::VERIFY_PEER)
+  rescue => @e
+    print("\tCA certificate file: ${ssl_ca_cert_file}\n")
+  end
+  visit('/')
```
srbarrios: Same here. This Capybara's `visit` method, has no relation with the previous result of `URI.open`.

testsuite/features/step_definitions/https_connection_steps.rb
srbarrios: I wasn't referring to the requirements in particular.
I was suggesting to move Ruby code to Ruby functions as part of `network_utils.rb`, so the functions can be reused, and the step definitions are just calling those functions.

**Example**

Instead of:
```
When(/^I connect to the server securely while using CA certificate file$/) do
  @url = "#{Capybara.app_host}".gsub("http:", "https:")
  ssl_ca_cert_file = "/etc/ssl/certs/#{@url.gsub('https://', '')}.pem"
  begin
    @res = URI.open(@url, redirect: false, open_timeout: 5, ssl_ca_cert: ssl_ca_cert_file, ssl_verify_mode: OpenSSL::SSL::VERIFY_PEER)
  rescue => @e
    print("\tCA certificate file: ${ssl_ca_cert_file}\n")
  end
end
```

Do:
```
When(/^I connect to the server securely while using CA certificate file$/) do
  ssl_ca_cert_file = "/etc/ssl/certs/#{@url.gsub('https://', '')}.pem"
  raise SystemCallError, "Connection failed." unless open_url(Capybara.app_host, secure_connection: true)
end
```

Where `open_url` is a function handling all the rest in `network_utils.rb`, and returning a boolean to assert the step in case the connection wasn't successful. Avoid the usage of a variable `@res` to process in another step.

testsuite/features/step_definitions/https_connection_steps.rb
srbarrios: I have no idea if we might reuse this piece of code in another scenario or step, but it looks interesting to me to have a function that allows us to create CAs passing some params.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: It does, otherwise the "title" won't be filled and couldn't be checked.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Same as previous, without "visit" there is no "title".

testsuite/features/step_definitions/https_connection_steps.rb
srbarrios: Checking a title coming from that `visit`, doesn't make sense, because this visit is using `Capybara.app_host`, not your connection made with `URI.open`.
Just debug it and you will see that `visit` will always use the same protocol and the same kind of SSL verification, as you don't change here the value of `Capybara.app_host` or the kind of connection used by `visit`, which comes from `capybara-3.40.0/lib/capybara/session.rb:261`

testsuite/features/secondary/https_connection.feature
```diff
@@ -11,12 +11,12 @@ Feature: Secure connection
...
     When I connect to the server securely
     Then the connection should be secured
-    And the page title should be "SUSE Multi-Linux Manager - Sign In" text
+    Then the page title should be "SUSE Multi-Linux Manager - Sign In" text
 
   Scenario: Connecting to the server securely while using CA certificate file
     When I connect to the server securely while using CA certificate file
     Then the connection should be secured
-    And the page title should be "SUSE Multi-Linux Manager - Sign In" text
+    Then the page title should be "SUSE Multi-Linux Manager - Sign In" text
```
srbarrios: Here we want `And`, as we had a previous `Then`.
While the Step definition on Ruby, should say `Then`, because this is a step validating some state.

testsuite/features/secondary/https_connection.feature
alichiel: Resolved with "Sign In" string check.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Reworked.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Done.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Done.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Reworked.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Reworked.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Reworked.

testsuite/features/step_definitions/https_connection_steps.rb
alichiel: Reworked.

testsuite/features/step_definitions/https_connection_steps.rb
```diff
@@ -0,0 +1,89 @@
...
+  unless [OpenURI::HTTPRedirect, OpenSSL::SSL::SSLError, Faraday::SSLError].any? { |err_class| error.instance_of? err_class } && error.message == '302 302'
+    uri_open_result = get_context(:uri_open_result)
+    log("Return code: #{uri_open_result.status.first}\nURI: #{uri_open_result.base_uri}\n") unless uri_open_result.nil?
+    raise ScriptError, 'The connection has not redirected to the secure channel!'
+  end
+  log("Error: #{error.message}\n") unless error.message.nil?
+end
+
+When(/^I connect to the server securely$/) do
+  uri_open_result = URI.open(Capybara.app_host.gsub('http:', 'https:'), redirect: false, open_timeout: 5, ssl_verify_mode: OpenSSL::SSL::VERIFY_NONE)
```
srbarrios, line 30: From my unknowledge on this topic, this is correct?
```
OpenSSL::SSL::VERIFY_NONE
```

alichiel: Yes, this means the https connection is used but the certificate issuer won't be checked. Technically we may want to use different word but I thought this is enough.
In other words this is the same behavior when you "add an exception" in the browser.

srbarrios (approved):

srbarrios (approved):

PR #10825, opened 2025-09-04
Optimize appstream cloning to prevent timeout during channel clone (bsc#1247990)

wweellddeerr:
## What does this PR change?

This PR fixes a timeout issue that occurs when cloning software channels containing a large number of modular appstreams.

The previous implementation in `AppStreamsManager.java` looped through each appstream and performed multiple individual INSERT operations per stream. This N+1 insertion resulted in excessive database round-trips and poor performance, leading to timeouts.

To resolve this, the cloning logic has been moved to the database. A new PostgreSQL function, `clone_channel_appstreams`, now handles the entire operation and the Java code has been refactored to make a single call to this new function.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: bug fix

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28083

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/appstreams/AppStreamsManager.java
+++ b/java/code/src/com/redhat/rhn/manager/appstreams/AppStreamsManager.java
@@ -20,9 +20,7 @@ import com.redhat.rhn.domain.action.ActionChain;
 import com.redhat.rhn.domain.action.ActionChainFactory;
 import com.redhat.rhn.domain.action.appstream.AppStreamActionDetails;
 import com.redhat.rhn.domain.channel.AppStream;
-import com.redhat.rhn.domain.channel.AppStreamApi;
 import com.redhat.rhn.domain.channel.Channel;
-import com.redhat.rhn.domain.rhnpackage.Package;
 import com.redhat.rhn.domain.user.User;
 import com.redhat.rhn.manager.action.ActionChainManager;
 import com.redhat.rhn.manager.system.SystemManager;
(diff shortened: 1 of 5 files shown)


java/code/src/com/redhat/rhn/common/db/datasource/xml/Channel_queries.xml
```diff
@@ -719,6 +719,12 @@ ORDER BY UPPER(C.name)
   </query>
 </callable-mode>
 
+<callable-mode name="clone_channel_appstreams">
+    <query params="fromChannelId, toChannelId">
+        {call clone_channel_appstreams(:fromChannelId, :toChannelId)}
```
mcalmer: I would be very interrested in trying to call the stored procedure from hibernate instead of via Mode Queries.
Can you give this a try?

mcalmer:
CC @lucidd - as you work also on cloning, you might be interrested

java/code/src/com/redhat/rhn/common/db/datasource/xml/Channel_queries.xml
mcalmer: There are multiple examples available. For example this one https://dzone.com/articles/call-stored-procedures-with-hibernate-and-postgres
I would try wit `createNativeQuery()` first

java/code/src/com/redhat/rhn/manager/appstreams/AppStreamsManager.java
```diff
@@ -182,33 +183,9 @@ public static AppStream findAppStream(Long channelId, String name, String stream
...
-            AppStreamsManager.listChannelAppStreams(from.getId()).forEach(appStream -> {
-                AppStream a = new AppStream();
-                a.setChannel(to);
-                a.setName(appStream.getName());
-                a.setStream(appStream.getStream());
-                a.setVersion(appStream.getVersion());
-                a.setContext(appStream.getContext());
-                a.setArch(appStream.getArch());
-                cloneArtifacts(a, appStream);
-                HibernateFactory.getSession().persist(a);
```
rjmateus, line 194: Did you try moving the persist/save to outside the stream and see if it makes any difference in the performance? I'm asking this because I really dislike psql to have any logic (hard to debug and maintain).
However, if it is the only possibility to have good performance, then we need to use it.

java/code/src/com/redhat/rhn/common/db/datasource/xml/Channel_queries.xml
wweellddeerr: Thank you for the tip. I had to switch from `CREATE FUNCTION` to `CREATE PROCEDURE` because calling it as a function from a native query required handling its results, which made the code a bit messy.

java/code/src/com/redhat/rhn/manager/appstreams/AppStreamsManager.java
wweellddeerr: It hasn’t been validated for a user facing timeouts during the operation; however, in my local environment, where I inserted a large amount of dummy data into the database, the performance improved from 3 minutes to 30 seconds.

java/code/src/com/redhat/rhn/common/db/datasource/xml/Channel_queries.xml
wweellddeerr: I mean, the code was messy with a function, but it’s clean now with a procedure.

java/code/src/com/redhat/rhn/manager/appstreams/AppStreamsManager.java
rjmateus: For reference, could you share how much time it takes with the psql procedure with the same amount of data?

wweellddeerr: that’s what I meant in my last comment: it took 3 minutes with the previous code and 30 seconds with the procedure, using the same amount of data, about 6x faster

rjmateus: got it. I thought the 30 seconds were with the persist/save  calls outside the stream. Thank you for clarification

PR #10840, opened 2025-09-08
Syncing Snap Channels

Kathy0704:
**PR for Syncing Snap Channels**
This PR introduces the ability to sync Snap channels. By running the command 
spacewalk-repo-sync -c <channel-label> -t snap, 
you can achieve the following:

1. Download the Snap stable channel to your local repository.
2. Import the package metadata into the rhnpackage table in the database.
3. Display the synced packages on the front-end web UI under Software > Channels > Channel List > Package Management.

**Testing the Command Line**:
**On the Database**
INSERT INTO rhnContentSourceType (id, label)
VALUES (nextval('rhn_content_source_type_id_seq'), 'snap');

Add the following entries to your database:
-INSERT INTO rhnChannelArch (id, label, arch_type_id, name, created, modified)
VALUES (525, 'channel-amd64-snap', 4, 'AMD64 Snap', NOW(), NOW());

- INSERT INTO rhnArchType (id, label, name) VALUES (4, 'snap', 'Snap');
- This allows you to create a Snap-type repository.

- INSERT INTO rhnPackageArch (id, label, name, arch_type_id) VALUES (151, 'amd64-snap', 'AMD64 Snap', 4);
- This adds a new package architecture and create a amd64-snap arch type channel.

- INSERT INTO rhnchannelpackagearchcompat (channel_arch_id, package_arch_id, created, modified) VALUES (525, 151, now(), now());
- This enables a snap channel to support syncing Snap packages.

**On the Front-end Web UI**
The process is similar to the current Uyuni logic:

1. Add a new channel.
2. Create a repository for this new channel and save it.

**On the uyuni container**
you need to add all the files on this GitHub to uyuni-server container
On this pr, you may see some duplicate files because I have added extensive debug and print logs to help track the progress of each step.
Please use the debug-enabled versions, you will need to replace the original files. Please use reposync_dev.py and packageImport_vm.py in your container.

First, remove the original files:
rm reposync.py
rm packageImport.py
Then, rename the debug files:
mv reposync_dev.py reposync.py
mv packageImport_vm.py packageImport.py
After performing these steps, you can run the command spacewalk-repo-sync -c <channel-label> -t snap as described above.
c0680018ffd58ba817c6d1ee44495d1

27abb9a8d7a112818690bd3c42a08a5

**On the terminal:**
!Note:
1. File Size: The full Snap stable repository is approximately 500GB. For testing purposes, you can limit the number of downloaded binaries by editing the python/spacewalk/satellite_tools/repo_plugins/snap_src.py file.
In the def get_package_list(self): function on line 181, change the return value at the end of the function to:
return to_return [num1: num2] 
This will only download a small subset of packages.
by this way you can not download all the snap package in this package list just the number you want to download
2. Calling function Time: In the same def get_package_list(self): function, the loop on line 192:
for ch in range(ord("a"), ord("z") + 1):
iterates through the entire alphabet, which can take up to 15 minutes. For testing, it is recommended to change the range to something smaller, such as range(ord("z"), ord("z") + 1), to save time.
4. After making these changes, run the following command:
spacewalk-repo-sync -c <channel-label> -t snap
You can then verify the results on the front-end by navigating to your target channel. You should see a result similar to the image below:
please run the spacewalk - repo - sync - c channel-label -t snap
then you will see final result:
on the front end , go to the targe channel and see
6416704e2f2a147e3d64062c1fb1271

**Design Overview**
The current binary download method is implemented in Python. However, in the future, it should be moved to Java if the chosen approach is to first list all packages from Snap stable, save only the metadata in the database, and then allow users to select which packages to download. This PR still follows the existing Uyuni logic of downloading all binary files locally first, then saving the metadata to the database.

**Key Idea**
The design for Snap channel management is structured across three key files: These three files are closely related but serve distinct purposes in the overall process.

python/spacewalk/satellite_tools/repo_plugins/snap_src.py
python/spacewalk/server/importlib/snapPackage.py
python/uyuni/common/rhn_snap.py

python/spacewalk/satellite_tools/repo_plugins/snap_src.py: 

This file handles the pre-download preparation. It fetches information from Snapcraft Canonical, which Uyuni then uses to perform the download.

python/uyuni/common/rhn_snap.py: 

After a package is downloaded from Snapcraft to the local repository, this file is responsible for reading the binary file's metadata from meta.yaml. It requires the squashfs file system to unzip the binary files.

python/spacewalk/server/importlib/snapPackage.py: 

After the metadata is read from the local binary file, this file formats the data to be suitable for the Uyuni database. It handles the necessary changes to the data structure and data types.

### DIFF
--- a/python/spacewalk/satellite_tools/repo_plugins/snap_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/snap_src.py
@@ -189,7 +189,7 @@ class SnapRepo:
             "Accept": "application/json",
         }
 
-        for ch in range(ord("z"), ord("z") + 1):
+        for ch in range(ord("a"), ord("z") + 1):
             keyword = chr(ch)
             page = 1
             same_count = 0
@@ -268,7 +268,7 @@ class SnapRepo:
                     )
                     break
 
-        return to_return[208:210]
+        return to_return
 
 
 # pylint: disable-next=missing-class-docstring


rjmateus:
I'm not sure if we should had the snap synchronization to the repo-sync process. What would be the benefit of having it integrated in the reposync process and in the same tables we have regular package repositories? The management of this kind of pacakges are funmdamentally different, and will not bne compatible.

Kathy0704:
> I'm not sure if we should had the snap synchronization to the repo-sync process. What would be the benefit of having it integrated in the reposync process and in the same tables we have regular package repositories? The management of this kind of pacakges are funmdamentally different, and will not bne compatible.

Thanks for the comment!
At the current stage, I put Snap packages into the same table mainly as a temporary solution, so that they can be displayed in the Uyuni UI without requiring frontend changes. This allows us to quickly validate the flow and see the data in the existing interface.

In the long term, I fully agree that Snap packages should be managed separately. The plan is to create a dedicated table for Snap data and adjust the React frontend logic accordingly, since the lifecycle management of Snap packages is fundamentally different from traditional package repositories.

Regarding the repo-sync process: my goal is to reuse the existing Uyuni logic for now (download files → read metadata from download files→ store metadata in the database). If we use a different approach (e.g., skipping downloads and fetching metadata directly), then we would need to introduce a new command-line tool or workflow. Keeping it within repo-sync for the time being helps reduce complexity and lets us leverage existing code paths.

PR #10849, opened 2025-09-11
Add a set -o pipefail command before the execution of system commands with pipes to manage the use of pipes in them

Pablogoliva:
## What does this PR change?

Check the pipestatus variable of a commands that contains commands chained by pipes in it to ensure that all of the commands have the expected outcome.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed:

- [x] **DONE**

## Test coverage
- Cucumber tests were edited

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23650
Port(s): 
 -  Manager 5.1: https://github.com/SUSE/spacewalk/pull/29182
 - Manager 5.0: https://github.com/SUSE/spacewalk/pull/29181
 - Manager 4.3: https://github.com/SUSE/spacewalk/pull/29183

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/support/remote_node.rb
+++ b/testsuite/features/support/remote_node.rb
@@ -110,10 +110,47 @@ class RemoteNode
   # @param successcodes [Array<Integer>] An array with the values to be accepted as success codes from the command run.
   # @param buffer_size [Integer] The maximum buffer size in bytes.
   # @param verbose [Boolean] Whether to log the output of the command in case of success.
+  # @param exec_option [Boolean] The container exec option.
   # @return [Array<String, String, Integer>] The output, error, and exit code.
   def run(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i')
     cmd_prefixed = @has_mgrctl && runs_in_container ? "mgrctl exec #{exec_option} '#{cmd.gsub('\'', '\'"\'"\'')}'" : cmd
     run_local(cmd_prefixed, separated_results: separated_results, check_errors: check_errors, timeout: timeout, successcodes: successcodes, buffer_size: buffer_size, verbose: verbose)
+  end
+
+  # Runs a command that contains commands chained by pipes in it and returns the output, error, and exit code of all the commands chained. Just for debugging purpouses.
+  #
+  # @param cmd [String] The command to run.
+  # @param expected_pipestatus_codes [Array<Integer>] Expected stderrs of the commands chained by pipes.
+  # @param runs_in_container [Boolean] Whether the command should be run in the container or on the host. Defaults to true.
+  # @param separated_results [Boolean] Whether the results should be stored separately. Defaults to false.
+  # @param check_errors [Boolean] Whether to check for errors or not. Defaults to true.
+  # @param timeout [Integer] The timeout to be used, in seconds.
+  # @param successcodes [Array<Integer>] An array with the values to be accepted as success codes from the command run.
+  # @param buffer_size [Integer] The maximum buffer size in bytes.
+  # @param verbose [Boolean] Whether to log the output of the command in case of success.
+  # @param exec_option [Boolean] The container exec option.
+  # # @return [Array<String, Integer>] The output, error, and exit code.
+  def run_pipe(cmd, expected_pipestatus_codes, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i')
+    pipestatus_file_path = '/tmp/temp_file_with_stderrs'
+    cmd += "; echo ${PIPESTATUS[*]} > #{pipestatus_file_path}"
+    cmd_read_codes = "cat #{pipestatus_file_path}; rm #{pipestatus_file_path}"
+    if @has_mgrctl && runs_in_container
+      cmd = "mgrctl exec #{exec_option} '#{cmd.gsub('\'', '\'"\'"\'')}'"
+      cmd_read_codes = "mgrctl exec #{exec_option} '#{cmd_read_codes.gsub('\'', '\'"\'"\'')}'"
+    end
+
+    out, initial_code = run_local(cmd, separated_results: separated_results, check_errors: check_errors, timeout: timeout, successcodes: successcodes, buffer_size: buffer_size, verbose: verbose)
+    stderr_of_commands, _code = run_local(cmd_read_codes)
+
+    stderr_of_commands_array = stderr_of_commands.split.map(&:to_i)
+    raise "Expected the number of expected pipestatus codes does not match the number of commands chained by pipes. Expected stderr:#{expected_pipestatus_codes}, current stderr:#{stderr_of_commands_array}" if expected_pipestatus_codes.length != stderr_of_commands_array.length
+    raise "Expected outcome does not match with current outcome. Expected stderr:#{expected_pipestatus_codes}, current stderr:#{stderr_of_commands_array}" if check_errors && stderr_of_commands_array.each_index.any? { |i| stderr_of_commands_array[i] != expected_pipestatus_codes[i] }
+
+    if separated_results
+      [out, stderr_of_commands, initial_code]
+    else
+      [out + stderr_of_commands, initial_code]
+    end
   end
 
   # Runs a command locally and returns the output, error, and exit code.


testsuite/features/support/remote_node.rb
```diff
@@ -104,8 +104,9 @@ def scp_download(remote_path, local_path, host: @full_hostname)
   # @param buffer_size [Integer] The maximum buffer size in bytes.
   # @param verbose [Boolean] Whether to log the output of the command in case of success.
   # @return [Array<String, String, Integer>] The output, error, and exit code.
-  def run(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i')
-    cmd_prefixed = @has_mgrctl && runs_in_container ? "mgrctl exec #{exec_option} '#{cmd.gsub('\'', '\'"\'"\'')}'" : cmd
+  def run(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false)
```
NamelessOne91: Did you check if this functions is called somewhere with the `exec_option` parameter set to a different value than the default of '-i' ?
If not, fine to remove it like you did.

testsuite/features/support/commonlib.rb
```diff
@@ -322,6 +322,16 @@ def deb_host?(name)
...
 end
 
+# Determines if the given command contains a pipe.
+#
+# @param cmd [String] The string to check.
+# @return [Boolean] Returns true if the command contains a pipe in it, false otherwise.
+def command_contains_a_pipe?(cmd)
+  command_splitted = cmd.split('|')
+  command_without_spaces = command_splitted.map(&:strip)
+  command_splitted.length > 1 && command_without_spaces.all? { |command| !command.empty? }
```
srbarrios: Maybe something simpler like `cmd.match?(/\S\s*\|\s*\S/)` could be enough for our needs? Or even just `cmd.include?('|')` if we just want to know if there is a pipe character in between.

testsuite/features/support/remote_node.rb
Pablogoliva: This comment is not relevant anymore.

testsuite/features/support/commonlib.rb
Pablogoliva: Simplified using this hint.

Pablogoliva:
> Not a big fan.
> 
> First, with `def command_contains_a_pipe?(cmd)`, we are doing some kind of meta-programming. Analyzing our own code at run time and acting upon it.
> 
> Second, it's likely to introduce a lot of unexpected changes that will be hard to debug.
> 
> I can only speak for myself, but I know how pipes work and very often it's intentional in my code that only the last command can fail.
> 
> I would rather vote for a case-by-case adjustment of their work by the developers themselves. For me, Domink's card plays as a way to raise attention on this bash programming detail.
> 
> (finally, on a purely technical side, I think that it's possible to change the environment variables of a SSH connection, maybe it's cleaner than prefixing).

What if we have something in between?. I have changed the approach to another approach that brings the opportunity to check the stderr of all of the pipes inside a commands but doesn't enforce it.

Bischoff (commented):
OK now it's not enforced, but I still don't like a lot this kind of meta-programming.

testsuite/features/support/remote_node.rb
```diff
@@ -110,8 +110,11 @@ def scp_download(remote_path, local_path, host: @full_hostname)
   # @param successcodes [Array<Integer>] An array with the values to be accepted as success codes from the command run.
   # @param buffer_size [Integer] The maximum buffer size in bytes.
   # @param verbose [Boolean] Whether to log the output of the command in case of success.
+  # @param exec_option [Boolean] The container exec option.
+  # @param check_stderr_of_pipes [Boolean] Check the stderr of all the commands separated by pipes or not.
   # @return [Array<String, String, Integer>] The output, error, and exit code.
-  def run(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i')
+  def run(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i', check_stderr_of_pipes: false)
+    cmd = "set -o pipefail ; #{cmd.gsub('\'', '\'"\'"\'')}" if check_stderr_of_pipes && cmd.include?('|')
```
srbarrios: Maybe name it `check_pipe_errors`:
```suggestion
  # @param exec_option [Boolean] The container exec option.
  # @param check_pipe_errors [Boolean] Check the stderr of all the commands separated by pipes or not.
# @return [Array<String, String, Integer>] The output, error, and exit code.
  def run(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i', check_pipe_errors: false)
    cmd = "set -o pipefail ; #{cmd.gsub('\'', '\'"\'"\'')}" if check_pipe_errors && cmd.include?('|')
```

srbarrios: Another approach it could be to move this to another method, like:

```
  # Runs a command that includes a pipe and returns the outputs, errors, and exit codes for each part of the pipe
  #
  # @param cmd [String] The command to run.
  # @param separated_results [Boolean] Whether the results should be stored separately.
  # @param check_errors [Boolean] Whether to check for errors or not.
  # @param timeout [Integer] The timeout to be used, in seconds.
  # @param successcodes [Array<Integer>] An array with the values to be accepted as success codes from the command run.
  # @param buffer_size [Integer] The maximum buffer size in bytes.
  # @param verbose [Boolean] Whether to log the output of the command in case of success.
  # @return [Array<String, Integer>] The output, error, and exit code.
  def run_pipe_local(cmd, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false)
    out, err, code = ssh_command(cmd, @target, timeout: timeout, buffer_size: buffer_size)
    # TODO: PROCESS FINAL ERRORS BUT RETURN ALSO `PIPESTATUS` VARIABLE.
  end
```

srbarrios (dismissed):

testsuite/features/support/remote_node.rb
Pablogoliva: Applied.

testsuite/features/support/remote_node.rb
```diff
@@ -110,12 +110,52 @@ def scp_download(remote_path, local_path, host: @full_hostname)
   # @param successcodes [Array<Integer>] An array with the values to be accepted as success codes from the command run.
   # @param buffer_size [Integer] The maximum buffer size in bytes.
   # @param verbose [Boolean] Whether to log the output of the command in case of success.
+  # @param exec_option [Boolean] The container exec option.
```
Pablogoliva, line 113: Just noticed that this parameter was missing in the description of the method.

testsuite/features/support/remote_node.rb
```diff
@@ -110,12 +110,50 @@ def scp_download(remote_path, local_path, host: @full_hostname)
...
```
Pablogoliva, line 113: Just noticed that this parameter was missing in the description of the method.

Pablogoliva, line 113: Just noticed that this parameter was missing in the description of the method.

testsuite/features/support/remote_node.rb
```diff
@@ -110,12 +110,51 @@ def scp_download(remote_path, local_path, host: @full_hostname)
...
```
Pablogoliva, line 113: Just noticed that this parameter was missing in the description of the method.

testsuite/features/support/remote_node.rb
```diff
@@ -110,12 +110,51 @@ def scp_download(remote_path, local_path, host: @full_hostname)
...
+  # @param timeout [Integer] The timeout to be used, in seconds.
+  # @param successcodes [Array<Integer>] An array with the values to be accepted as success codes from the command run.
+  # @param buffer_size [Integer] The maximum buffer size in bytes.
+  # @param verbose [Boolean] Whether to log the output of the command in case of success.
+  # @param exec_option [Boolean] The container exec option.
+  # @param raise_error_if_a_command_fails [Boolean] Raises an error if a command fails (even if is chained by pipes), otherwise returns the outcome of the commands chained by pipes.
+  # # @return [Array<String, Integer>] The output, error, and exit code.
+  def run_and_check_pipe_error(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, successcodes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i', raise_error_if_a_command_fails: false)
+    pipestatus_file_path = '/tmp/temp_file_with_stderrs'
+    cmd += "; echo ${PIPESTATUS[*]} > #{pipestatus_file_path}"
```
Pablogoliva, line 135: I know that using temporal files is not the cleanest way to do it. But for a debugging method I didn't want to modify the bashrc file and I couldn't directly execute the second command just after because the SSH connections are atomic, also modifying the way we establish and send SSH commands or create a new one was maybe too much for a debugging method.

testsuite/features/support/remote_node.rb
srbarrios: What about something like this?

```suggestion
  def run_pipe(cmd, runs_in_container: true, separated_results: false, check_errors: true, timeout: DEFAULT_TIMEOUT, expected_pipestatus_codes: [0], buffer_size: 65_536, verbose: false, exec_option: '-i')
```

- If we have `check_errors` enabled then we check one by one all the pipe status code and verify that are the expected ones in the array `expected_pipestatus_codes`. If some is not expected, we raise an error.
- If we have `check_errors` disabled, we don't raise an error in any case.
- We can keep returning all the pipestatus codes always, as in your code.

srbarrios: Just combine both no?
Read and then delete, on the same ssh call.

Pablogoliva: Good idea. And it doesn't affect to the stdout because the rm doesn't have.

Pablogoliva: Good idea

Pablogoliva: Applied with one difference. I set expected_pipestatus_codes as mandatory without default value.

srbarrios (approved):

PR #10852, opened 2025-09-12
SonarCloud error reduction fix, miscellaneous 15 rules

CDellaGiusta:
## What does this PR change?

[SonarCloud fix: java:S1185 Overriding methods should do more than simply call the same method in the super class](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS1185&rule_key=java%3AS1185)
[SonarCloud fix: java:S1319 Declarations should use Java collection interfaces rather than specific implementation classes](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS1319&rule_key=java%3AS1319)
[SonarCloud fix: java:S1488 Local variables should not be declared and then immediately returned](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS1488&rule_key=java%3AS1488)
[SonarCloud fix: java:S2293: the diamond operator should be used](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS2293&rule_key=java%3AS2293)
[SonarCloud fix: java:S2093 Try-with-resources should be used](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS2093&rule_key=java%3AS2093)
[SonarCloud fix: java:S1940 Boolean checks should not be inverted](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS1940&rule_key=java%3AS1940)
[SonarCloud fix: java:S4065 ThreadLocal.withInitial should be preferred](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS4065&rule_key=java%3AS4065)
[SonarCloud fix: java:S2446 notifyAll() should be preferred over notify()](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS2446&rule_key=java%3AS2446)
[SonarCloud fix: java:S3398 private methods called only by inner classes should be moved to those classes](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3398&rule_key=java%3AS3398)
[SonarCloud fix: java:S2387 Child class fields should not shadow parent class fields](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS2387&rule_key=java%3AS2387)
[SonarCloud fix: java:S2185 Do not perform unnecessary mathematical operations](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS2185&rule_key=java%3AS2185)
[SonarCloud fix: java:S3984 Exceptions should not be created without being thrown](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3984&rule_key=java%3AS3984)
[SonarCloud fix: java:S1751 Loops with at most one iteration should be refactored](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS1751&rule_key=java%3AS1751)
[SonarCloud fix: java:S3516 Methods returns should not be invariant](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3516&rule_key=java%3AS3516)
[SonarCloud fix: java:S2326 Unused type parameters should be removed](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS2326&rule_key=java%3AS2326)

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links

Port(s):
- [x ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/redhat/rhn/common/util/http/HttpClientAdapter.java
+++ b/java/code/src/com/redhat/rhn/common/util/http/HttpClientAdapter.java
@@ -394,7 +394,7 @@ public class HttpClientAdapter {
      * @param uri the URI to check
      * @return true if proxy should be used, else false
      */
-    private boolean useProxyFor(URI uri) {
+    protected boolean useProxyFor(URI uri) {
         if (uri.getScheme().equals("file")) {
             return false;
         }
--- a/java/code/src/com/redhat/rhn/frontend/taglibs/RhnHiddenTag.java
+++ b/java/code/src/com/redhat/rhn/frontend/taglibs/RhnHiddenTag.java
@@ -37,7 +37,6 @@ public class RhnHiddenTag extends TagSupport {
 
     private static final long serialVersionUID = -8385317358288103720L;
 
-    private String id;
     private String name;
     private String value;
 
--- a/java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java
+++ b/java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java
@@ -28,7 +28,6 @@ import com.redhat.rhn.common.util.CompressionUtil;
 import com.redhat.rhn.domain.access.AccessGroupFactory;
 import com.redhat.rhn.domain.channel.Channel;
 import com.redhat.rhn.domain.channel.ChannelFactory;
-import com.redhat.rhn.domain.channel.DistChannelMap;
 import com.redhat.rhn.domain.errata.Errata;
 import com.redhat.rhn.domain.org.Org;
 import com.redhat.rhn.domain.org.OrgFactory;
@@ -64,6 +63,7 @@ import com.redhat.rhn.manager.system.IncompatibleArchException;
 import com.suse.manager.utils.PagedSqlQueryBuilder;
 import com.suse.oval.ShallowSystemPackage;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -1764,8 +1764,8 @@ public class PackageManager extends BaseManager {
 
     private static String getAssociatedRelease(Package pack) {
         for (Channel chan : pack.getChannels()) {
-            for (DistChannelMap map : chan.getDistChannelMaps()) {
-                return map.getRelease();
+            if (CollectionUtils.isNotEmpty(chan.getDistChannelMaps())) {
+                return chan.getDistChannelMaps().iterator().next().getRelease();
             }
         }
         return null;
--- a/java/code/src/com/redhat/rhn/common/util/http/test/HttpClientAdapterTest.java
+++ b/java/code/src/com/redhat/rhn/common/util/http/test/HttpClientAdapterTest.java
@@ -31,7 +31,6 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
-import java.lang.reflect.Method;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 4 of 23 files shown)


java/code/src/com/redhat/rhn/frontend/taglibs/RhnHiddenTag.java
```diff
@@ -37,7 +37,6 @@ public class RhnHiddenTag extends TagSupport {
 
     private static final long serialVersionUID = -8385317358288103720L;
 
-    private String id;
```
rjmateus, line 40: Can we remove this field? It looks to me that is being used.

java/code/src/com/redhat/rhn/common/util/http/HttpClientAdapter.java
```diff
@@ -394,7 +394,7 @@ public List<Cookie> getCookies(String name) {
      * @param uri the URI to check
      * @return true if proxy should be used, else false
      */
-    private boolean useProxyFor(URI uri) {
+    public boolean useProxyFor(URI uri) {
```
rjmateus: why changing this to public if is not used outside (or at least as far as I understood

java/code/src/com/redhat/rhn/frontend/taglibs/RhnHiddenTag.java
CDellaGiusta: Yes, because it is a shadowed parent field (as per java:S2387 Child class fields should not shadow parent class fields)

In fact:

`public class RhnHiddenTag extends TagSupport  .....`

... and ...
```
public class TagSupport ...
    protected String id;

```

java/code/src/com/redhat/rhn/common/util/http/HttpClientAdapter.java
CDellaGiusta: Rule java:S3398 ([private methods called only by inner classes should be moved to those classes](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3398&rule_key=java%3AS3398)) should have suggested to move private method `useProxyFor` of  `HttpClientAdapter` inside the  inner class `CustomProxyRoutePlanner`, since this is the only place where it is called, and so this is where it should belong.
However, there was another place where this method is called, in `HttpClientAdapterTest.java`

 ```
private boolean callUseProxyFor(URI uri) throws Exception {
        Method method = HttpClientAdapter.class.getDeclaredMethod("useProxyFor", URI.class);
        method.setAccessible(true);
        return (Boolean) method.invoke(new HttpClientAdapter(), uri);
    }
```

You can see that the "`method.setAccessible(true);`" should be avoided, since it's bad practice.
Hence, putting `useProxyFor` as public, I can solve both problems: removing `method.setAccessible(true);` in the test and having a method not breaking the rule. I just thought it was a decent compromise, but any suggestion is welcome

Comment (mackdk) [reply java/code/src/com/redhat/rhn/common/util/http/HttpClientAdapter.java]: Since this method is used only in unit testing, the general java convention is to use package private visibility, which we cannot do because our tests do not live in the same package as the class they're testing.  

Given the method `callUseProxyFor` always calls `useProxyFor` on a new instance, I would say that a good compromise would be to set the method `protected` instead of `public` and use inheritance and not reflection to widen its visibility for the test. This approach has the advantage of not changing the accessibility of method for the consumer classes while allowing at the same time the test to check it, at the expense of creating each time an unnamed inner class (which for a test is not a big deal):

```java
    private boolean callUseProxyFor(URI uri) throws Exception {
        var httpClientAdapter = new HttpClientAdapter() {
            @Override
            public boolean useProxyFor(URI uri) {
                return super.useProxyFor(uri);
            }
        };

        return httpClientAdapter.useProxyFor(uri);
    }
```

Generally speaking though, when there is the need to test in isolation a method that should be private to a class it is a code smell: it usually means that the method should be refactored to be elsewhere.

java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java
```diff
@@ -1756,8 +1755,8 @@ public static Package findDebugSource(User user, Package pack) {
 
     private static String getAssociatedRelease(Package pack) {
         for (Channel chan : pack.getChannels()) {
-            for (DistChannelMap map : chan.getDistChannelMaps()) {
-                return map.getRelease();
+            if (chan.getDistChannelMaps().iterator().hasNext()) {
```
mackdk: Isn't this the same as?

```java
            if (CollectionUtils.isNotEmpty(chan.getDistChannelMaps())) {
```

mackdk (approved):

java/code/src/com/redhat/rhn/manager/rhnpackage/PackageManager.java
CDellaGiusta: yes, suggestion taken before squashing

java/code/src/com/redhat/rhn/common/util/http/HttpClientAdapter.java
CDellaGiusta: Your suggested implementation is nice. Implemented when squashing the code

PR #10854, opened 2025-09-12
Fix code for determining private interface

Bischoff:
## What does this PR change?

This PR fixes the code for determining private interface

It uses the same method as for the public interface: test one interface after the other, stopping at the first one that matches. This allows to consider priorities.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

Cucumber tests were modified

- [x] **DONE**

## Links

Port(s): none, Head only

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/remote_node.rb
+++ b/testsuite/features/support/remote_node.rb
@@ -54,9 +54,16 @@ class RemoteNode
 
     if (PRIVATE_ADDRESSES.key? host) && !$private_net.nil?
       @private_ip = net_prefix + PRIVATE_ADDRESSES[host]
-      @private_interface = 'eth1'
-      _out, _err, code = ssh('ip address show ens4')
-      @private_interface = 'ens4' if code.zero?
+      @private_interface = nil
+      %w[eth1 ens4].each do |dev|
+        _output, code = run_local("ip address show dev #{dev}")
+
+        if code.zero?
+          @private_interface = dev
+          break
+        end
+      end
+      raise StandardError, "No private interface for '#{@hostname}'." if @private_interface.nil?
     end
 
     ip = client_public_ip


testsuite/features/support/remote_node.rb
```diff
@@ -54,9 +54,16 @@ def initialize(host, port: 22)
 
     if (PRIVATE_ADDRESSES.key? host) && !$private_net.nil?
       @private_ip = net_prefix + PRIVATE_ADDRESSES[host]
-      @private_interface = 'eth1'
-      _out, _err, code = ssh('ip address show ens4')
-      @private_interface = 'ens4' if code.zero?
+      @private_interface = ''
```
szachovy:
```suggestion
      @private_interface = nil
```
? In case something happens and you would have empty string on the global var

testsuite/features/support/remote_node.rb
```diff
@@ -54,9 +54,16 @@ def initialize(host, port: 22)
...
+      %w[eth1 ens4].each do |dev|
```
szachovy, line 58: Is it possible to make it more dynamic with listing by `ip -o link show`?

testsuite/features/support/remote_node.rb
```diff
@@ -54,9 +54,16 @@ def initialize(host, port: 22)
...
+        _output, code = run_local("ip address show dev #{dev}")
+
+        next unless code.zero?
+
+        @private_interface = dev
+        break
```
szachovy:
```suggestion
    if code.zero?
      @private_interface = dev
      break
    end
```
maybe this way it would be more readable?

testsuite/features/support/remote_node.rb
Bischoff: i catch it with a raise afterwards...

but it's true that raise could test nil? instead of empty?

testsuite/features/support/remote_node.rb
Bischoff: no. we want hardcoded priorities

same for the public interfaces

testsuite/features/support/remote_node.rb
Bischoff: i hesitated.. ok let's take this one

testsuite/features/support/remote_node.rb
szachovy: I think in general it's better to assign to `nil` instead of empty string. More pure style thing as you catch it anyway later. I am open to counter arguments.

testsuite/features/support/remote_node.rb
szachovy: ok

testsuite/features/support/remote_node.rb
Bischoff: No, I think you are right. No counter arguments.

PR #7852, opened 2023-11-09
Raspberry Pi OS 12 support

raulillo82:
## What does this PR change?

Add Raspbian OS 12 support for both arm64 and armhf architectures.

## GUI diff

No difference.

- [X] **DONE**

## Documentation
- Documentation issue needs to be created: https://github.com/uyuni-project/uyuni-docs/pull/2607 

- [X] **DONE**

## Test coverage
- Cucumber tests to be added later

- [ ] **DONE**

## Links

Implements #7853 

- [X] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

PR #10860, opened 2025-09-15
Add ARMHF for DEB based back to fix incomplete PR#7852

vzhestkov:
## What does this PR change?

Looks like https://github.com/uyuni-project/uyuni/pull/7852 was not really complete as works fine just in case of upgrading the server, in case of installing most recent versions from scratch there is almost no data about ARMHF for Debian based distros in the DB.

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Tracks: https://github.com/SUSE/spacewalk/issues/28209

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/schema/spacewalk/susemanager-schema.changes.vzhestkov.add-deb-armhf-compat-archs
+++ b/schema/spacewalk/susemanager-schema.changes.vzhestkov.add-deb-armhf-compat-archs
@@ -1,0 +1,1 @@
+- Add leftovers of partially missing ARMHF for Debian


schema/spacewalk/common/data/rhnServerPackageArchCompat.sql
```diff
@@ -560,5 +560,17 @@ insert into rhnServerPackageArchCompat
 (server_arch_id, package_arch_id, preference) values
 (LOOKUP_SERVER_ARCH('arm64-debian-linux'), LOOKUP_PACKAGE_ARCH('all-deb'), 1000);
 
+insert into rhnServerPackageArchCompat
+(server_arch_id, package_arch_id, preference) values
+(LOOKUP_SERVER_ARCH('armhf-debian-linux'), LOOKUP_PACKAGE_ARCH('armhf-deb'), 0);
+
+insert into rhnServerPackageArchCompat
```
rjmateus, line 567: This one is not part of the previous PR, is it missing on the table?

schema/spacewalk/common/data/rhnServerServerGroupArchCompat.sql
```diff
@@ -1261,4 +1261,32 @@ insert into rhnServerServerGroupArchCompat ( server_arch_id, server_group_type )
...
+            lookup_sg_type('foreign_entitled'));
+
+insert into rhnServerServerGroupArchCompat ( server_arch_id, server_group_type )
+	values (lookup_server_arch('armhf-debian-linux'),
+            lookup_sg_type('container_build_host'));
+
+insert into rhnServerServerGroupArchCompat ( server_arch_id, server_group_type )
+	values (lookup_server_arch('armhf-debian-linux'),
+            lookup_sg_type('monitoring_entitled'));
+
```
meaksh, line 1291: Shouldn't we add also `proxy_entitled` for `armhf-debian-linux`? I see we have it `arm64-debian-linux` and other ARM related archs.

schema/spacewalk/common/data/rhnServerPackageArchCompat.sql
vzhestkov: If I get the question right it was exactly the point that in update script it was adding this data while not adding the values to original state of the table, what was causing the situation when the new installed server instance doesn't have such entries while the instance which was updated from the previous version has it.

schema/spacewalk/common/data/rhnServerServerGroupArchCompat.sql
vzhestkov: Doesn't it mean Uyuni Proxy based on armhf debian client? I don't think it's possible.

meaksh: Hmmm yeah, but even if we only support Uyuni proxy on x86_64 and aarch64, I see we already fill `proxy_entitled` for all arches here: https://github.com/uyuni-project/uyuni/blob/ec1735f3b128d3c891f59cf70df14b2df1dde32f/schema/spacewalk/common/data/rhnServerServerGroupArchCompat.sql#L1099-L1262

So, I wonder whether we should also add `armhf-debian-linux` here for consistency.

vzhestkov: For me it looks like it was made by accident some time ago and copied for some architectures by accident.
@rjmateus. what do you think? Do we need `proxy_entitled` for such systems?

rjmateus: I was looking at the data, and it looks to me that it would be good to add it, especially for consistency since all the other architectures are there.

vzhestkov: just put proxy entitlement back

PR #10925, opened 2025-10-02
Fix bug 1246035: Correct PROFILE_NAME variable definition in bootstrap script

CDellaGiusta:
## What does this PR change?

The bootstrap script has defined a variable named PROFILENAME to specify profilename for client systems. However later in the script, a variable named PROFILE_NAME is addressed. The variable name  has been changed to match the code by adding the underscore separator.

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: not needed
- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/27853
Port(s): 5.0: https://github.com/SUSE/spacewalk/pull/28535, 5.1: https://github.com/SUSE/spacewalk/pull/28536
- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java
+++ b/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java
@@ -397,6 +397,16 @@ public class RegisterMinionEventMessageAction implements MessageAction {
         });
     }
 
+    private void setMinionName(String minionId, MinionServer minion, Optional<ValueMap> grains) {
+        grains.ifPresentOrElse(
+                grain -> minion.setName(
+                        grain.getMap("susemanager")
+                                .flatMap(s -> s.getOptionalAsString("profile_name"))
+                                .orElse(minionId)),
+                () -> minion.setName(minionId)
+        );
+    }
+
     /**
      * Update information of already registered minion, in case minion_id is different
      * or machine-id is different.
@@ -410,7 +420,10 @@ public class RegisterMinionEventMessageAction implements MessageAction {
         if (!minionId.equals(oldMinionId)) {
             LOG.warn("Minion '{}' already registered, updating profile to '{}' [{}]", oldMinionId, minionId,
                     registeredMinion.getMachineId());
-            registeredMinion.setName(minionId);
+
+            Optional<ValueMap> grainsOptional = saltApi.getSystemInfoFull(minionId).map(SystemInfo::getGrains);
+            setMinionName(minionId, registeredMinion, grainsOptional);
+
             registeredMinion.setMinionId(minionId);
             ServerFactory.save(registeredMinion);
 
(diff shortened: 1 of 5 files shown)


aaannz:
As I mentioned in the bug, I do not think this is used anywhere in the code.
Before dropping legacy stuff by @mcalmer it was used only for traditional.
Then in c1a347aac267 he dropped legacy stuff and introduced this segment of the code. But I do not see any addition to the server side using this.

IMO this should be completely removed. Or am I missing something?

CDellaGiusta:
> IMO this should be completely removed. Or am I missing something?

adding mc as a reviewer, so he can shed a light on this

mcalmer:
Hmm, we set profile_name as grain to submit it during the registration in the bootstrap script:
```bash
    if [ -n "$PROFILE_NAME" ]; then
        echo "Setting profile name to: $PROFILE_NAME"
        cat <<EOF >>"$SUSEMANAGER_MASTER_FILE"
        profile_name: "$(echo $PROFILE_NAME)"
EOF
    fi
```

But it seems we do not use it in RegisterMinionEventMessageAction .
Maybe we should fix the java side and use the profile name as "Name" when it is provided as grain.

```java
           minion.setName(
                    grains.getMap("susemanager")
                            .flatMap(s -> s.getOptionalAsString("profile_name"))
                            .orElse(minionId));
```

This would make it possible that the customer define the "Name" before bootstrapping.

CDellaGiusta:
> But it seems we do not use it in RegisterMinionEventMessageAction . Maybe we should fix the java side and use the profile name as "Name" when it is provided as grain.

I made the changes, please have a look

mcalmer (dismissed):

java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java
```diff
@@ -397,6 +397,18 @@ private void reactivateSystem(String minionId, String machineId, String reActiva
...
+    private void setMinionName(String minionId, MinionServer minion, Optional<ValueMap> grains) {
+        if (grains.isPresent()) {
+            minion.setName(
+                    grains.get().getMap("susemanager")
+                            .flatMap(s -> s.getOptionalAsString("profile_name"))
+                            .orElse(minionId));
+        }
+        else {
+            minion.setName(minionId);
+        }
```
aaannz: This is only a nitpick, but we are mixing styles here.
What about:
```suggestion
        grains.ifPresentOrElse(
                grain -> {
                    minion.setName(
                        grain.getMap("susemanager")
                            .flatMap(s -> s.getOptionalAsString("profile_name"))
                            .orElse(minionId));
                },
                () -> minion.setName(minionId)
        );
```

aaannz (approved):

java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java
CDellaGiusta: I agree. Suggestion taken in commit squash

aaannz (approved):

PR #10938, opened 2025-10-03
Ensure alerts remain visible in long forms

bisht-richa:
## What does this PR change?

In some long forms (e.g., the Proxy Configuration form), the success message appears at the top of the page after submission. However, if the user submits the form while scrolled down, it may seem like nothing happened because the message isn’t immediately visible. The user has to scroll up manually to see the result.

With this update, alerts now remain “sticky” at the top of the page, ensuring users see immediate feedback even if they’ve scrolled down.

## GUI diff

Screenshot 2025-11-12 at 16 36 26

Screenshot 2025-11-12 at 16 20 37

- [x] **DONE**

## Documentation
- No documentation needed: Ensure alerts remain visible in long forms.

- [x] **DONE**

## Test coverage
- No tests: Ensure alerts remain visible in long forms.

- [x] **DONE**

## Links

Fixes: [28354](https://github.com/SUSE/spacewalk/issues/28354)
Port(s): [5.1](https://github.com/SUSE/spacewalk/pull/29754)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/web/html/src/manager/minion/proxy/proxy-config-messages.tsx
+++ b/web/html/src/manager/minion/proxy/proxy-config-messages.tsx
@@ -4,19 +4,25 @@ import { MessagesContainer, showErrorToastr, showInfoToastr, showSuccessToastr }
 type SuccessType = boolean | undefined;
 
 export const ContainerConfigMessages = (success: SuccessType, messagesIn: ReactNode[], loading: boolean) => {
+  const hasErrors = messagesIn.length > 0;
+
+  const renderErrors = () => (
+    <>
+      {messagesIn.map((msg, index) => (
+        <div key={`error-${index}`}>{msg}</div>
+      ))}
+    </>
+  );
+
   if (success) {
     showSuccessToastr(t("Proxy configuration action has been scheduled."));
-  } else if (messagesIn.length > 0) {
-    showErrorToastr(
-      <>
-        {messagesIn.map((msg, index) => (
-          <div key={`error${index}`}>{msg}</div>
-        ))}
-      </>,
-      { autoHide: false }
-    );
+  } else if (hasErrors) {
+    showErrorToastr(renderErrors(), { autoHide: false });
   } else if (loading) {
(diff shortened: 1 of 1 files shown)


Etheryte:
Is this PR ready to review? Based on the description, I'm not entirely sure I understand when and where we need this and why?

web/html/src/components/messages/messages.tsx
```diff
@@ -56,6 +57,22 @@ const _classNames = {
 };
 
 export class Messages extends React.Component<Props> {
+  componentDidUpdate(prevProps) {
+    const prevItems = Array.isArray(prevProps.items) ? prevProps.items : [prevProps.items];
+    const curItems = Array.isArray(this.props.items) ? this.props.items : [this.props.items];
+
+    const changed =
```
Etheryte: I don't think this is the right way to do this, but I'm not sure I understand what we're trying to do here? Trigger an event when the props change?

Etheryte:
I think the new approach works. Maybe we want to make it the default and omit the prop altogether, so all message containers are sticky, what do you think?

bisht-richa:
> I think the new approach works. Maybe we want to make it the default and omit the prop altogether, so all message containers are sticky, what do you think?

Yes, the current approach works fine for single-line alerts, but it becomes a bit tricky with multi-line ones — they occupy too much space at the top, and since our custom `Message` doesn’t have an auto-hide or close option, so the alerts stay visible and the user can’t dismiss them when they need more space for the form, which ends up cluttering the screen.

I’m planning to switch the custom `Message` alert to use `MessagesContainer` instead. It already supports auto-hide and a close option, so it should work better for this case.
What do you think?

Etheryte:
I think that makes sense. I'm not too familiar with the specific implementation details here, but I think the user should be able to close all alerts anyway. In general, I think successes are fine to auto-hide, we shouldn't automatically close errors though.

PR #10941, opened 2025-10-03
Check only for major version for proxy containers

aaannz:
## What does this PR change?

Port of https://github.com/SUSE/spacewalk/pull/28432

Check only for major version for proxy containers.
This was probably intended even before, but matcher was matching minor version too.

## GUI diff

No difference.

- [X] **DONE**

## Documentation

- [X] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #
Port(s): # https://github.com/SUSE/spacewalk/pull/28432

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/proxy-httpd-image/proxy-httpd-image.changes.oholecek.fix-check-only-for-majorversion
+++ b/containers/proxy-httpd-image/proxy-httpd-image.changes.oholecek.fix-check-only-for-majorversion
@@ -1,0 +1,1 @@
+- Check only for major version for proxy containers
--- a/containers/proxy-httpd-image/uyuni-configure.py
+++ b/containers/proxy-httpd-image/uyuni-configure.py
@@ -68,7 +68,7 @@ with open(config_path + "httpd.yaml", encoding="utf-8") as httpdSource:
 
     server_version = config.get("server_version")
     # Only check version for SUSE Multi-Linux Manager, not Uyuni
-    matcher = re.fullmatch(r"([0-9]+\.[0-9]+\.)[0-9]+", server_version)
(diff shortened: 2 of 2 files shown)


rjmateus:
@aaannz this means that in each Major version upgrade, users must re-generate the proxy configuration or manually edit the configuration file?
Would make more sense to check that the needed data is present in each version, and if something is missing, provide meaningful information? It would make it more flexible and avoid errors on user side, or the need to regenerate the config data.

aaannz:
I can remove this check completely. For example, uyuni is not checked at all.
For MLM it currently acts as a some form of versioning, but too strict, hence this PR.

I only vaguely remember that we added this check as a precaution for when we will fully drop old python scripts that older proxies may stop working.
However given that we generally support proxy - 1 version with server, this check seems to me superfluous and we will always need to make sure older proxies will work.

rjmateus:
My main concern is that when customers migrate to a newer major version, we still block it even when the configuration is still valid. To work around it they need to manually change the config file version or generate a new configuration. 
I think we would be in a better place if we dropped this check and instead checked that the data structure has what we need (which should be the case; otherwise, the proxy would not work).
What do you think @aaannz ? Makes sense?

aaannz:
In usecase it was originally designed, this is not only about data in the config, but if the used image (particularly httpd image) can work with the server.

E.f. Once we drop processing of the systemid on the server, even if proxy config is the same, all clients behind the proxy will start to get permission denied on repo refresh.

PR #10943, opened 2025-10-03
Make some packages to build on openSUSE Tumbleweed

meaksh:
## What does this PR change?

This PR makes the `uyuni-common-libs`, `spacewalk-client-tools`, `rhnlib` and `mgr-push` packages to build on openSUSE Tumbleweed.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: **spec file changes**

- [x] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


client/rhel/spacewalk-client-tools/spacewalk-client-tools.spec
```diff
@@ -25,7 +25,7 @@
 %global __python /usr/bin/python3
 %endif
 
-%if !(0%{?rhel} >= 8 || 0%{?sle_version} >= 150000 )
+%if !(0%{?rhel} >= 8 || 0%{?suse_version} >= 1500 )
```
juliogonzalez, line 28: Wouldn't this still build on Tumbleweed? What's the value of suse_version there?

Same for the others.

python/uyuni/uyuni-common-libs.spec
```diff
@@ -25,7 +25,7 @@
 %global build_py3 1
 %endif
 
-%if ( 0%{?rhel} && 0%{?rhel} < 8 ) || 0%{?suse_version}
+%if ( 0%{?rhel} && 0%{?rhel} < 8 ) || 0%{?suse_version} < 1600
```
juliogonzalez: In fact I wonder if we should not use this expression everywhere, for the other packages as well?

vzhestkov: `0%{?suse_version} < 1600` are you sure about it? it matches with any debian, rh etc as `0%{?suse_version}` expands there to 0

client/rhel/spacewalk-client-tools/spacewalk-client-tools.spec
meaksh: Here, since this is negation, we will exclude all SLE/openSUSE that is higher than 15.

The `suse_version` covers SLE and openSUSE, while `sle_version` would only exclude SLE (and not openSUSE).

python/uyuni/uyuni-common-libs.spec
meaksh: I wanted to be conservative here, as some are negations and the others not.. I didn't want to do a refactors here, just in case :sweat_smile:

meaksh: @vzhestkov hmm, that's correct!! good catch!

meaksh: it should be fixed now

mcalmer (approved):

PR #10945, opened 2025-10-04
Uyuni hibernate migration mc

mcalmer:
## What does this PR change?

Next step hibernate annotations migrations

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


java/code/src/com/redhat/rhn/domain/common/Checksum.java
```diff
@@ -38,7 +56,7 @@ public Long getId() {
     /**
      * @param idIn The id to set.
      */
-    public void setId(Long idIn) {
+    protected void setId(Long idIn) {
```
CDellaGiusta: since there is no <meta attribute="scope-set">protected</meta> in the xml file, why are you putting this to protected? Is it something we have to do anyway?
We have to find a way to share this tips ...

java/code/src/com/redhat/rhn/domain/common/FileList.java
```diff
@@ -23,14 +23,44 @@
...
     private Long id;
+
+    @Column
     private String label;
+
+    @ManyToOne
+    @JoinColumn(name = "org_id")
     private Org org;
+
+    @ManyToMany(cascade = {CascadeType.REFRESH, CascadeType.MERGE, CascadeType.PERSIST})
```
CDellaGiusta: xml file mentions cascade="save-update", where do you get these cascade information?

java/code/src/com/redhat/rhn/domain/common/FileList.java
```diff
@@ -23,14 +23,44 @@
...
+    @JoinTable(
+            name = "rhnFileListMembers",
+            joinColumns = { @JoinColumn(name = "file_list_id") },
+            inverseJoinColumns = { @JoinColumn(name = "config_file_name_id") }
+    )
```
CDellaGiusta, line 63: How to map the xml description to this code? I feel we have to find a way to share all these information somehow
```

<bag name="fileNames" table="rhnFileListMembers" cascade="save-update"
             order-by="created">
          <key column="file_list_id"/>
          <many-to-many column="config_file_name_id"
                class="com.redhat.rhn.domain.config.ConfigFileName" />
        </bag>
```

java/code/src/com/redhat/rhn/domain/common/FileList.java
```diff
@@ -44,7 +74,7 @@ public Long getId() {
     /**
      * @param i The id to set.
      */
-    public void setId(Long i) {
+    protected void setId(Long i) {
```
CDellaGiusta: Always protected if it's a setId?

CDellaGiusta (approved):
Code looks good as far as tests run.
There are some xml-annotation mappings not clear to me, we should have a chat about it and find a way to share this information

java/code/src/com/redhat/rhn/domain/common/Checksum.java
mcalmer: Right, there was no protected in the XML.
I will role it back.
Maybe later we can think about to make all setId protected as soon as the ID is somehow generated and should not be set from external. But that can happen after a discussion and in a separate PR.

java/code/src/com/redhat/rhn/domain/common/FileList.java
mcalmer: There is no "save-update" in JPA.
It seems that "SAVE_UPDATE" is equivalent to MERGE + PERSIST. 
See https://docs.jboss.org/hibernate/orm/6.5/javadocs/org/hibernate/Session.html#saveOrUpdate(java.lang.Object)
or https://medium.com/@himani.prasad016/jpa-hibernate-cascade-types-0490be19cc1f

I will remove the refresh.

java/code/src/com/redhat/rhn/domain/common/FileList.java
mcalmer: "bag" does not exist in JPA.
Reading hibernate docs it was an unordered list which allow duplicates.
Funny we have an "order-by" criteria and looking at the join table it has a unique index. So duplicates are also not possible.
So I tried to replace it with a simple ManyToMany relation and it seems to work.

java/code/src/com/redhat/rhn/domain/common/FileList.java
mcalmer: same as above. I will revert.

java/code/src/com/redhat/rhn/domain/common/FileList.java
mcalmer: Even better described here in the migration guide of hibernate 7
https://docs.jboss.org/hibernate/orm/7.0/migration-guide/migration-guide.html#session-save-update

CDellaGiusta: in 4.3 file UserImpl.hbm.xml I have found this snippet:
```
<set name="stateChanges" lazy="true"
      cascade="save-update" sort="natural">
      <key column="web_contact_id"/>
      <one-to-many class="com.redhat.rhn.domain.user.StateChange" />
  </set>
```
In master this was translated (UserImpl.java @ 129) as:
 ```
  @OneToMany(mappedBy = "user", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
    private Set<StateChange> stateChanges = new HashSet<>();
```
so from this translation it seems that "save-update" moves to CascadeType.ALL
Is that not correct?

Comment (mcalmer) [reply java/code/src/com/redhat/rhn/domain/common/FileList.java]: That change the behavior. This can be wanted in this case, but it is not equivalent.
ALL include a "REMOVE" and this is explicitly discarded by save_update.

File: java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java [hunk 5]
```diff
@@ -760,7 +760,11 @@ public static List<Channel> listCustomBaseChannels(User user) {
      * @return List of ChecksumTypes instances
      */
     public static List<ChecksumType> listYumSupportedChecksums() {
-        return singleton.listObjectsByNamedQuery("ChecksumType.loadAllForYum", Map.of());
+        return getSession().createQuery("""
+                 FROM com.redhat.rhn.domain.common.ChecksumType as t
```
CDellaGiusta, line 764: for my reference: is the full class name (com.redhat.rhn.domain.common.ChecksumType) necessary or is it sufficient to have the simple one (ChecksumType)?

CDellaGiusta: also: default of lookupObjectByNamedQuery is cacheable = false. Did you change it on purpose?

java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java
```diff
@@ -784,7 +788,12 @@ public static ChecksumType findChecksumTypeByLabel(String checksum) {
         if (checksum == null) {
             return null;
         }
-        return singleton.lookupObjectByNamedQuery("ChecksumType.findByLabel", Map.of(LABEL, checksum));
+        return getSession().createQuery("""
+                 FROM com.redhat.rhn.domain.common.ChecksumType AS t
+                 WHERE t.label = :label""", ChecksumType.class)
+                .setParameter(LABEL, checksum)
```
CDellaGiusta, line 794: ditto

java/code/src/com/redhat/rhn/domain/channel/ChannelFactory.java
mcalmer: I think you can use the simple one if it is unique.

ChecksumType is just a static list of pre-defined types similar to ArchType and ActionStatus. So I think cachable was just forgotten.

PR #10948, opened 2025-10-06
Uyuni hibernate migration utils

CDellaGiusta:
## What does this PR change?

There are many places in the code where there is a named query that looks for an unique object, with the selection made on one canonical field (usually id, name, label and so on).

The aim is to provide a generic method in HibernateFactory to replace the named query.

Where the code is something like:

```
ActionType-xbm.xml:

<query name="ActionType.findByLabel">
      <![CDATA[from com.redhat.rhn.domain.action.ActionType as t where t.label = :label]]>
 </query>


Anywhere in the code:

ActionType actType =  
      hibernateFactory.lookupObjectByNamedQuery("ActionType.findByLabel", Map.of("label", label), true);
```

we can be substitute it with

```
ActionType-xbm.xml:
   ... removed ...

Anywhere in the code:
  ActionType actType = 
       hibernateFactory.lookupObjectByParam("label", label, ActionType.class, true);
```
    
### Security analysis:
```
  String sanitizedParamName = paramName.replaceAll("[^A-Za-z]+", "");
  String query = "FROM %s WHERE %s = :param".formatted(objClass.getSimpleName(), sanitizedParamName);
  return session.createQuery(query, objClass)
                    .setParameter("param", paramValue, StringType.INSTANCE)
                    .setCacheable(cacheable)
                    .uniqueResult();

```
SonarCloud asks to check that the dynamically formatted SQL query is safe.
1) Calls to this method are within the code and with no user input
2) Query table name: comes from Class<T> objClass method parameter, calling objClass.getSimpleName() so it's safe
3) Parameter name: comes from String paramNam method parameter, and should be called by developers with a constant string. Anyway it is sanitized allowing only alphabetical chars (just in case), so it is safe
4) Parameter value: it is substituted through setParameter, so it is safe

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Port(s): 
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


mcalmer:
I think we should get rid of these "hibernateFactory.lookupObjectBy****".
Why not porting this to:
```
getSession().createQuery("from com.redhat.rhn.domain.action.ActionType as t where t.label = :label", ActionType.class)
        .setParameter("label", label")
        .uniqueResult()
```

CDellaGiusta:
> I think we should get rid of these "hibernateFactory.lookupObjectBy****". Why not porting this to:

The reason of this proposal is that I've seen many of simple named queries in the hbm.xml files, looking up for objects by 1 simple parameter (like id label, name etc). All these methods are very similar, and the proposed set of methods is a tentative not to repeat this pattern.
An example: ActionFactory.lookupActionTypeByLabel and ActionFactory.lookupActionStatusByName. 
Their implementation should be something like:

```
getSession().createQuery("from ActionType as t where t.label = :label", ActionType.class)
        .setParameter("label", label")
        .setCacheable(true)
        .uniqueResult()
```
 and 
 ```
getSession().createQuery("from ActionStatus as t where t.name = :name", ActionStatus.class)
        .setParameter("name", name")
        .setCacheable(true)
        .uniqueResult()   
```     

If it could be possible to have one implementation for both, maybe we could "collapse" code which is similar.
I don't know whether there could be other unexpected implications, though

**Comment by mcalmer** [comment]
I think I understand this now. I think it is a good idea.

**Comment by mcalmer** [comment]
I think the only difference is the value type Long vs String. Would it be possible to use "instanceof" ?
Do we have more cases where we query for 2 parameter? Would it make sense to have Map<String, Object> for the parameters?

File: java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java [hunk 1]
```diff
@@ -133,6 +135,108 @@ public static void registerComponentName(String componentName) {
...
+     * Anywhere in the code:
+     *      ActionType actType = hibernateFactory.lookupObjectByParam("id", id, ActionType.class, true);
+     *
+     * @param paramName the parameter name (e.g. "label", "name", "id" etc.)
+     * @param paramValue the parameter actual value
+     * @param objClass class of the query object
+     * @param cacheable if we should cache the results of this object
+     * @return Instance of the found object by the query or null if nothing is found.
+     */
+    protected <T> T lookupObjectByParam(String paramName, Long paramValue, Class<T> objClass, boolean cacheable) {
```
mackdk: Nitpicking as usual but, I think the order of the parameters should be `objClass`, `paramName`, `paramValue`, `cacheable`

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
```diff
@@ -133,6 +135,108 @@ public static void registerComponentName(String componentName) {
...
+        String sanitizedParamName = paramName.replaceAll("[^A-Za-z]+", "");
+
+        try {
+            Session session = HibernateFactory.getSession();
+            String query = "FROM %s WHERE %s = :param".formatted(objClass.getSimpleName(), sanitizedParamName);
+            return session.createQuery(query, objClass) //NOSONAR dynamically formatted SQL query is safe here
+                    .setParameter("param", paramValue, LongType.INSTANCE)
```
mackdk: Why are you explicitly passing the Hibernate type? Shouldn't `.setParameter("param", paramValue)` work as well? If so, you can replace both versions with a single one:

```
    protected <T, V> T lookupObjectByParam(Class<T> objClass, String paramName, V paramValue, boolean cacheable) {
        String sanitizedParamName = paramName.replaceAll("[^A-Za-z]+", "");
        try {
            Session session = HibernateFactory.getSession();
            String query = "FROM %s WHERE %s = :param".formatted(objClass.getSimpleName(), sanitizedParamName);
            return session.createQuery(query, objClass) //NOSONAR dynamically formatted SQL query is safe here
                    .setParameter(paramName, paramValue)
                    .setCacheable(cacheable)
                    .uniqueResult();
        }
        catch (HibernateException he) {
            throw new HibernateRuntimeException("lookupObjectByParam failed with param [%s]=[%s] on class [%s]"
                    .formatted(paramName, Objects.toString(paramValue), objClass.getSimpleName()), he);
        }
    }
```

Maybe we can even use `Object` and avoid the second type parameter.

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
```diff
@@ -133,6 +135,108 @@ public static void registerComponentName(String componentName) {
...
+        }
+    }
+
+    protected <T> T lookupObjectByParam(String paramName, Long paramValue, Class<T> objClass) {
+        return lookupObjectByParam(paramName, paramValue, objClass, false);
+    }
+
+    protected <T> T lookupObjectByParamCacheable(String paramName, Long paramValue, Class<T> objClass) {
+        return lookupObjectByParam(paramName, paramValue, objClass, true);
+    }
```
mackdk: Maybe `cacheable` should be `true` by default? Do you have already an idea on what is the most common value in our cases? 

Also I would avoid having a separate overload for the other case, calling `singleton.lookupObjectByParam(paramName, paramValue, objClass, true);` (or `false` depending on the default we choose)  could be readeable enough, and I think it's also shorter :-)

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
mackdk: Ah, one more thing moment :rofl: This won't work with `null`. I guess we won't be using when converting `IS NULL` queries, but for the future it's maybe better to add an explicit check and mention it in the javadoc:

```java
            return session.createQuery(query, objClass) // NOSONAR dynamically formatted SQL query is safe here
                    .setParameter(paramName, Objects.requireNonNull(paramValue, "null value is not supported"))
                    .setCacheable(cacheable)
                    .uniqueResult();
```

The other solution is of course to implement the null-check

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
CDellaGiusta: changed

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
CDellaGiusta: lookupObjectByNamedQuery has as default cacheable = false, so I will do it.
I'll remove the other case separate overload

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
CDellaGiusta: done

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
```diff
@@ -133,6 +134,53 @@ public static void registerComponentName(String componentName) {
...
+     *      ActionType actType = hibernateFactory.lookupObjectByParam("id", id, ActionType.class, true);
+     *
+     * @param paramName the parameter name (e.g. "label", "name", "id" etc.)
+     * @param paramValue the parameter actual value
+     * @param objClass class of the query object
+     * @param cacheable if we should cache the results of this object (default = false)
+     * @return Instance of the found object by the query or null if nothing is found.
+     */
+    protected <T, V> T lookupObjectByParam(Class<T> objClass, String paramName, V paramValue, boolean cacheable) {
+        String sanitizedParamName = paramName.replaceAll("[^A-Za-z]+", "");
```
szachovy, line 164: Are there some Java built-in methods for sanitization?

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
```diff
@@ -133,6 +134,53 @@ public static void registerComponentName(String componentName) {
...
+
+        try {
+            Session session = HibernateFactory.getSession();
+            String query = "FROM %s WHERE %s = :param".formatted(objClass.getSimpleName(), sanitizedParamName);
```
szachovy, line 168: For string based queries, I found that this is type of template that are generally used: https://docs.jboss.org/hibernate/orm/6.3/javadocs/org/hibernate/query/criteria/HibernateCriteriaBuilder.html

Does it have it's case here?

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
```diff
@@ -133,6 +134,53 @@ public static void registerComponentName(String componentName) {
...
+            return session.createQuery(query, objClass) //NOSONAR dynamically formatted SQL query is safe here
+                    .setParameter("param", Objects.requireNonNull(paramValue, "null value is not supported"))
+                    .setCacheable(cacheable)
+                    .uniqueResult();
+        }
+        catch (HibernateException he) {
+            throw new HibernateRuntimeException("lookupObjectByParam failed with param [%s]=[%s] on class [%s]"
```
szachovy, line 175: I don't know if someone who receives this error message will know what is going on, if yes (in case of Java devs) then feel free to ignore my comment.

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
mcalmer: Would `StringUtils.isAlpha(...)` be an option?

CDellaGiusta: no, unfortunately.
This method is called only internally (see protected) and it is a generic "swiss knife" method to  avoid to do the same operations again and again. It is called on developers input, not on input coming from the user. I made a sanitization just in case

mcalmer: Hmm, no. it is not the same what we need

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
CDellaGiusta: We decided to avoid CriteriaBuilder, since it results in unreadable code and these "builders" change with every major version as they do not find a good way to implement it

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
CDellaGiusta: yes, they are received by the java devs: it means that the parameters were wrong

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
szachovy: ok

java/code/src/com/redhat/rhn/common/hibernate/HibernateFactory.java
szachovy: ok

szachovy (approved):

PR #10970, opened 2025-10-08
Enhance cucumber rake namespace to export html output

maximenoel8:
## What does this PR change?

To improve the user experience with BV, expose the html report for each task (stage in Jenkins) in a txt file with fix name that can be use by Jenkins to show the report in the Jenkins interface.

Adding this new file required some refactoring, indeed the txt file was created for all tasks with the dir loop.
Use the enhance method to generate the file only once the task is executed.

image

This will help to debug a specific stage without the need of opening the full report.

Solution2: 
We could simplify the export by removing the timestamp and so having a predictable name but it will overwrite the previous results when running locally (without build number).
## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Issue(s): #
Depend on: 
Port(s): 
 - 4.3: https://github.com/SUSE/spacewalk/pull/28607
 - 5.0: https://github.com/SUSE/spacewalk/pull/28608
 - 5.1: https://github.com/SUSE/spacewalk/pull/28609

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/Rakefile
+++ b/testsuite/Rakefile
@@ -13,34 +13,45 @@ build_number = ENV.fetch('BUILD_NUMBER', nil)
 result_folder = build_number.nil? ? 'results' : "results/#{build_number}"
 
 namespace :cucumber do
+  # Define the timestamp before the loop starts.
+  timestamp = Time.now.strftime('%Y%m%d%H%M%S')
+  node_name = ENV.fetch('NODE', nil)
+  node_result_extension = node_name.nil? ? '' : "_#{node_name}"
   # Create results folder
   sh "mkdir -p #{result_folder}", verbose: false
   Dir.glob(File.join(Dir.pwd, 'run_sets', '**', '*.yml')).each do |entry|
-    Cucumber::Rake::Task.new(File.basename(entry, '.yml').to_sym) do |t|
-      features = YAML.safe_load(File.read(entry))
-      fail "Skipping rake task for #{entry}, the YAML file is empty." if features.nil? || features.empty? # rubocop:disable Style/SignalException
-
-      filename = File.basename(entry, '.yml').to_sym
-      timestamp = Time.now.strftime('%Y%m%d%H%M%S')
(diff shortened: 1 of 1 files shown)


szachovy:
Is it possible to `curl` content of this report from the outside?

maximenoel8:
@szachovy Yes, the controller workspace is accessible by Web Interface and by curl.
If you enter: http://suma-continuous-bv-controller.mgr.suse.de/ for example, you will have access to the cucumber files.
And in http://suma-continuous-bv-controller.mgr.suse.de/results/<build_number> you have access to the cucumber report for each rake task run. Like: http://suma-continuous-bv-controller.mgr.suse.de/results/285/output_20251009102220-build_validation_retail_proxy.html

maximenoel8:
The main issue is the timestamp that remove the possibility of a deterministic path.
Removing the timestamp is not really a good idea because you could run twice the same rake task and want to keep those two runs.
So to create some determinism, I'm using a txt file that point to the html report. If you have a better idea to do that like using symlink, I'm open to suggestion.

srbarrios:
@maximenoel8 why not just run a "echo" at the end of the Rake task, printing the URL of the report?
That 'echo' will be shown as part of the Jenkins console output as any other log, and we can easly see it.
I don't see a need to generate a temporary file, or move the responsibility to show this URL to the Groovy code.

maximenoel8:
> @maximenoel8 why not just run a "echo" at the end of the Rake task, printing the URL of the report? That 'echo' will be shown as part of the Jenkins console output as any other log, and we can easly see it. I don't see a need to generate a temporary file, or move the responsibility to show this URL to the Groovy code.

The idea is to have the url visible in a separate step. Just echoing the value in Rake will be part of the script step and you will need to scroll down to find it.

PR #11055, opened 2025-10-22
Make regular expressions safer to avoid SonarCloud warnings

vzhestkov:
## What does this PR change?

Fix for security hotspots found by SonarCloud related to regular expressions used in python code.

## Codespace

## GUI diff

No difference.

## Documentation

- [ ] **DONE**

## Test coverage
- Unit tests were added for the missing tests, but should work the same one for the existing test for the others
NOTE: `kiwi_info` is the only module touched which didn't have unit tests covering modified part, for the others the tests are present and should just work as expected (`mgrnet`, `byterange` etc.)

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


python/spacewalk/spacewalk-backend.changes.vzhestkov.make-regexps-bit-safer
```diff
@@ -0,0 +1 @@
+- Fix regular expressions to calm down SonarCloud
```
m-czernek: Let's not say "calm down" in chlog :))
```suggestion
- Fix regular expressions for SonarCloud
```

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
```diff
@@ -134,18 +131,14 @@ def parse_kiwi_result(dest):
...
         packages = __salt__["cp.get_file_str"](path)
-        pattern = re.compile(
-            r"^(?P<name>.*?)\|(?P<epoch>.*?)\|(?P<version>.*?)\|(?P<release>.*?)\|(?P<arch>.*?)\|(?P<disturl>.*?)(\|(?P<license>.*))?$"
-        )
+        fields = ["name", "epoch", "version", "release", "arch", "disturl", "license"]
         for line in packages.splitlines():
-            match = pattern.match(line)
-            if match:
+            line_data = line.split("|")
+            if len(line_data) in (6, 7):
```
m-czernek: Could we add a comment here that `license` is optional? (i.e. why "6 or 7")

aaannz: This is a mistake. `License` is not optional, so it should be split into 7 fields.

See https://github.com/OSInside/kiwi/blob/main/doc/source/image_types_and_results.rst#image-results

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
```diff
@@ -478,23 +458,21 @@ def inspect_bundles(dest, basename):
         if match:
             res1 = match.groupdict()
             sha256_file = f
-            # pylint: disable-next=undefined-variable
             sha256_str = __salt__["cp.get_file_str"](os.path.join(dest, sha256_file))
-            pattern2 = re.compile(r"^(?P<hash>[0-9a-f]+)\s+(?P<filename>.*)\s*$")
+            pattern2 = re.compile(r"^(?P<hash>[0-9a-f]+)\s+(?P<filename>\S.*)$")
```
aaannz, line 505: Why are you adding `\S` when we capture everything anyway?
```suggestion
            pattern2 = re.compile(r"^(?P<hash>[0-9a-f]+)\s+(?P<filename>.*)$")
```

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
```diff
@@ -478,23 +458,21 @@ def inspect_bundles(dest, basename):
...
             match = pattern2.match(sha256_str)
             if match:
                 d = match.groupdict()
-                # pylint: disable-next=consider-using-f-string
-                d["hash"] = "sha256:{0}".format(d["hash"])
+                d["filename"] = d["filename"].strip()
+                d["hash"] = f'sha256:{d["hash"]}'
```
aaannz, line 510: Is this safe to use now?

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
m-czernek: Ah the regex was wrong in that case, but that simplifies this line to:

```suggestion
            if len(line_data) == len(fields):
```

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
vzhestkov: it's exactly why SonarCloud is complaining as `.*` after `\s+` is also matching any space, the idea here is explisitly distinguish first non-space character and all the rest.

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
vzhestkov: why not?

aaannz: Just double checking there will be no surprise with old python version. I remember I had to revert such changes in saltboot because of older python version some year back.

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
aaannz: Well yes, that's why I propose to just remove `\s` and match everything to the end. What is there to complain about?

I am only imagining that year later we will wonder why there is one single non-whitespace expected.

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
vzhestkov: It will not work with python 2 for sure, but do we use `kiwi_info` somewhere where python 2 is used? at least we don't have any salt supported with python 2, for 3.6 it should work for sure

susemanager-utils/susemanager-sls/src/modules/kiwi_info.py
vzhestkov: Not sure if I get the last comment right. The change I made here in this PR is replacing `.*` with `\S.*` so basically `any sequence of character` to `first non space character`+`any character until the end of the line`, the tail of spaces is trimmed after with `d["filename"].strip()` just ot avoid the same issue of matching spaces with `.*`. The idea is to make the regexp here more strict.

PR #11070, opened 2025-10-24
Uyuni hibernate migration: step 8 fix

CDellaGiusta:
## What does this PR change?
In the context of [hibernate major version update](https://github.com/SUSE/spacewalk/issues/24556), this PR is a fix of hibernate migration of class **ServerAction**, migrated in step 8

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Port(s): no backports
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/code/src/com/redhat/rhn/domain/action/server/ServerAction.java
```diff
@@ -49,7 +48,6 @@
  */
 @Entity
 @Table(name = "rhnServerAction")
-@IdClass(ServerActionId.class)
```
mcalmer, line 53: I think removing the Id class is the wrong way.
The primary key / unique key of the table is server_id + action_id.

Maybe we should try if there is a difference when we define not only "Long serverId", but also "Long actionId" and use MapsId for "parentAction" similar as with "server".

CDellaGiusta: I put a comment at the beginning of the class as agreed

PR #11078, opened 2025-10-27
migrate Server to annotations

mcalmer:
## What does this PR change?

Migrate Server class from hbm.xml to annotations

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -78,80 +88,253 @@
...
-    private Boolean ignoreEntitlementsForMigration;
-
+    @Id
+    @GeneratedValue(generator = "rhn_server_seq")
+    @GenericGenerator(
+        name = "rhn_server_seq",
+        strategy = "org.hibernate.id.enhanced.SequenceStyleGenerator",
+        parameters = {
+                @Parameter(name = "sequence_name", value = "rhn_server_id_seq"),
+                @Parameter(name = "initial_value", value = "1000010000"),
```
CDellaGiusta, line 129: where does the initial value come from?

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -78,80 +88,253 @@
...
     private ContactMethod contactMethod;
+
+    @Column(name = "running_kernel")
     private String runningKernel;
+
+    @Column(name = "last_boot")
     private Long lastBoot;
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "server_arch_id")
```
CDellaGiusta, line 177: there is no obvious migration for lazy="proxy". Did you by chance tried `@LazyToOne(LazyToOneOption.PROXY)`?

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -78,80 +88,253 @@
...
     private ServerArch serverArch;
+
+    @ManyToOne(fetch = FetchType.LAZY, cascade = CascadeType.ALL)
```
CDellaGiusta, line 180: this one says cascade="save-update" , but I guess it doesn't work. I've found that sometimes changing it to ALL make things work

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -78,80 +88,253 @@
...
+    @OneToOne(mappedBy = "server", fetch = FetchType.LAZY, cascade = CascadeType.ALL)
     private ServerUuid serverUuid;
+
+    @OneToMany(mappedBy = "server", cascade = CascadeType.ALL, fetch = FetchType.LAZY, orphanRemoval = true)
     private Set<Note> notes;
+
+    @OneToMany(mappedBy = "server", fetch = FetchType.LAZY, cascade = CascadeType.ALL, orphanRemoval = true)
     private Set<ServerFQDN> fqdns;
+
+    @OneToOne(mappedBy = "server", fetch = FetchType.LAZY, cascade = CascadeType.ALL)
```
CDellaGiusta, line 214: note: access="field" is the default in annotations, meaning that hibernate does not set this field using the setters but directly

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -78,80 +88,253 @@
...
+            joinColumns = @JoinColumn(name = "server_id"),
+            inverseJoinColumns = @JoinColumn(name = "server_group_id")
+    )
     private Set<ServerGroup> groups = new HashSet<>();
+
+    @OneToMany(mappedBy = "server", fetch = FetchType.LAZY, cascade = CascadeType.ALL, orphanRemoval = true)
     private Set<ClientCapability> capabilities = new HashSet<>();
+
+    @ManyToMany(fetch = FetchType.LAZY, cascade = CascadeType.ALL)
+    @JoinTable(
```
CDellaGiusta, line 296: ditto `cascade="save-update"`

java/code/src/com/redhat/rhn/domain/server/Server.java
```diff
@@ -2626,14 +2802,14 @@ public boolean isOsFamilySuse() {
      * @param osFamilyIn to set
      */
     public void setOsFamily(String osFamilyIn) {
-        this.osFamily = osFamilyIn;
+        throw new NotImplementedException("os_family does not exist on Server");
```
CDellaGiusta, line 2805: I am a bit undecided if those two methods should do nothing instead of throwing an exception....

java/code/src/com/redhat/rhn/domain/server/VirtualInstance.java
```diff
@@ -30,15 +44,41 @@
...
+    @JoinColumn(name = "host_system_id")
+    private Server hostSystem;
+
+    @Column
     private String uuid;
+
+    @Column
     private Long confirmed;
+
+    @OneToOne(mappedBy = "parent")
```
CDellaGiusta, line 80: I would put `cascade = CascadeType.ALL` in `@OneToOne`, so that we can spot the `@Cascade `deprecated stuff straight away

java/code/src/com/redhat/rhn/domain/server/VirtualInstanceInfo.java
```diff
@@ -19,18 +19,47 @@
...
+
+    @Column(name = "vcpus")
     private Integer virtualCPUs;
+
+    @OneToOne
+    @JoinColumn(name = "instance_id")
+    @MapsId
     private VirtualInstance parent;
+
+    @ManyToOne
```
CDellaGiusta: (fetch = FetchType.EAGER)

java/code/src/com/redhat/rhn/domain/server/VirtualInstanceInfo.java
```diff
@@ -19,18 +19,47 @@
...
+    @JoinColumn(name = "instance_type", updatable = false, nullable = false)
     private VirtualInstanceType type;
+
+    @ManyToOne
```
CDellaGiusta: (fetch = FetchType.EAGER)

java/code/src/com/redhat/rhn/manager/system/test/SystemManagerTest.java
```diff
@@ -767,8 +767,7 @@ private void addCpuToServer(Server s) {
         cpu.setMHz(CPUTest.MHZ);
         cpu.setModel(CPUTest.MODEL);
         cpu.setNrCPU(NUM_CPUS);
-        TestUtils.saveAndFlush(cpu);
-        TestUtils.reload(s);
+        s.setCpu(cpu);
```
CDellaGiusta, line 770: this saving/flushing/reloading is really nasty! You feel you have no clue of what is really happening

java/code/src/com/redhat/rhn/domain/server/VirtualInstanceFactory.java
```diff
@@ -83,41 +85,31 @@ public void saveVirtualInstance(VirtualInstance virtualInstance) {
     public VirtualInstance lookupByGuestId(Org org, Long id) {
         Session session = HibernateFactory.getSession();
 
-        return (VirtualInstance) session.getNamedQuery("VirtualInstance.lookupGuestBySidAndOrg")
```
CDellaGiusta, line 86: looks like `lookupByGuestId` is never used. Worth removing?

java/code/src/com/redhat/rhn/domain/server/VirtualInstanceFactory.java
```diff
@@ -161,69 +153,50 @@ public void deleteVirtualInstanceOnly(VirtualInstance virtualInstance) {
...
+                             where
+                                  sgt.label = 'virtualization_host'
+                                  and sgm.server_id = host.id
+                          ))
+                        """, Tuple.class)
                 .setParameter("org_id", org.getId(), StandardBasicTypes.LONG)
-                .list();
-
-        return new HashSet<>(convertToView(results));
+                .addScalar("guest_id", StandardBasicTypes.BIG_INTEGER)
```
CDellaGiusta: I used to put `StandardBasicTypes.LONG `and then `t.get(0, Long.class)`
Worth doing so?

java/code/src/com/redhat/rhn/domain/server/MinionServerFactory.java
```diff
@@ -333,9 +335,17 @@ public static List<MinionServer> findEmptyProfilesByHostName(String hostname) {
...
-        List<Object[]> results = findByIds(serverIds, "Server.findSimpleMinionsByServerIds", "serverIds");
-        return results.stream().map(row -> new MinionIds(((BigDecimal) row[0]).longValue(), row[1].toString()))
-                .collect(toList());
+        return getSession().createNativeQuery("""
+                SELECT m.server_id, m.minion_id
+                FROM   suseMinionInfo m
+                WHERE  m.server_id IN (:serverIds)
+                """, Tuple.class)
+                .setParameterList("serverIds", serverIds)
+                .addScalar("server_id", StandardBasicTypes.BIG_INTEGER)
```
CDellaGiusta, line 344: ditto Long?

java/code/src/com/redhat/rhn/domain/server/ServerFactory.java
```diff
@@ -269,11 +282,22 @@ public static Set<OsReleasePair> listAllServersOsAndRelease() {
...
+                FROM   rhnServer s
+                JOIN   suseMinionInfo m ON s.id = m.server_id
+                WHERE  EXISTS (SELECT 1
+                               FROM   rhnUserServerPerms USP
+                               WHERE  USP.user_id = :user_id
+                               AND    USP.server_id = s.id)
+                """, Tuple.class)
+                .setParameter("user_id", id)
+                .addScalar("minion_id", StandardBasicTypes.STRING)
+                .addScalar("server_id", StandardBasicTypes.BIG_INTEGER)
```
CDellaGiusta: ditto Long?

java/code/src/com/redhat/rhn/domain/server/ServerFactory.java
```diff
@@ -547,15 +571,38 @@ public static Server lookupByCert(ClientCertificate clientcert)
...
+                                   WHERE  SA.server_id = SNPC.server_id
+                                   AND    SA.status = AST.id
+                                   AND    AST.name IN('Queued', 'Picked Up')
+                                   AND    AEU.action_id = SA.action_id
+                                   AND    AEU.errata_id = E.id )
+                ORDER BY E.id
+                """, Tuple.class)
+                .setParameterList("serverIds", serverIds)
+                .setParameter("user_id", user.getId())
+                .addScalar("serverId", StandardBasicTypes.BIG_INTEGER)
```
CDellaGiusta: ditto Long?

java/code/src/com/redhat/rhn/domain/server/Server.java
mcalmer: This is from the DB SQL file as the sequence is defined to start with this value.
I thought it would be better to have this also in the annotations.

`CREATE SEQUENCE rhn_server_id_seq START WITH 1000010000;`

java/code/src/com/redhat/rhn/domain/server/Server.java
mcalmer: Seem this proxy is outdated, maybe deprecated or at least hibernate specific.
If it works without, should be better prevent using this?

java/code/src/com/redhat/rhn/domain/server/Server.java
mcalmer: I will try

java/code/src/com/redhat/rhn/domain/server/Server.java
CDellaGiusta: Just a pure question. If it works, no problem at all!

java/code/src/com/redhat/rhn/domain/server/Server.java
mcalmer: I think exception is better as we directly see implementation errors. Otherwise we will hunt ghosts when we think an object is a Minion, but in reality it is only a Server and the value is not set.

java/code/src/com/redhat/rhn/domain/server/Server.java
CDellaGiusta: Sometimes I have to force it to Cascade.ALL to be working, so, if it does not work, leave it as it is without any problem. We'll have surprises in the migration anyway when we upgrade the hibernate version

java/code/src/com/redhat/rhn/domain/server/Server.java
CDellaGiusta: ok, it makes sense

java/code/src/com/redhat/rhn/domain/server/VirtualInstance.java
mcalmer: We would get these problems with 2 different CascadeTypes. When we are able to change SAVE_UPDATE, we can easily also change this cascade type to "the other ALL".

CDellaGiusta: OK

java/code/src/com/redhat/rhn/manager/system/test/SystemManagerTest.java
mcalmer: Yes, the old "idea" was to save CPU with server_id set and than reload this object to have it also in the cache.
But now the reload say "a different object already in cache".
So the right way is, to set the references on both ends and let hibernate only save when it is really needed.

java/code/src/com/redhat/rhn/domain/server/VirtualInstanceFactory.java
mcalmer: Not sure if this works. AFAIK I tried it at some time earlier and it produced problems.
But I can try and see if tests are failing

java/code/src/com/redhat/rhn/domain/server/Server.java
mcalmer: Failed - only "ALL" works.

java/code/src/com/redhat/rhn/domain/server/Server.java
mcalmer: same here. Only ALL works with the current test code.

java/code/src/com/redhat/rhn/domain/server/VirtualInstanceFactory.java
mcalmer: seems to work. changing

PR #11084, opened 2025-10-28
OpenID login endpoint for MCP security integration

cbbayburt:
Adds `/manager/api/oicdLogin` endpoint that accepts JWTs forwarded by the MCP server from an external idP with strict token verification and explicit user matching.

Related PR for the MCP server: https://github.com/uyuni-project/mcp-server-uyuni/pull/33

### Authentication flow

1. Following the OAuth 2.0 exchange between the user, Uyuni MCP server, and the idP, the MCP server forwards the JWT issued by the idP to the `/manager/api/oidcLogin` endpoint of MLM API (downstream resource server) in the "Authorization" header.
2. MLM validates the token against idP's public JWK (see below)
3. MLM extracts the `preferred_username` value that is included in the JWT
4. MLM matches the username to an existing active user in MLM and logs the user in
5. In response, MLM returns a session key inteded to be used for subsequent API calls
6. Uyuni MCP server puts the session key in a cookie to make subsequent API calls

### Security overview

MLM validates the JWTs according to the following rules:

- Signature verification using keys from the idP's JWKS URI.
- Allowed JWS algorithms: RSA/PSS SHA-256, SHA-384, SHA-512 and ESDCA SHA-256, SHA-384.
- Requires an expiration time claim.
- Allows for a 5-second clock skew.
- Requires a subject claim (any).
- Validates the issuer claim against the configured OIDC issuer.
- Custom audience validation against the configured OIDC audience. Uyuni MCP server and the MLM require different `aud` claims to accept the JWT: respectively `mcp-server-uyuni` and `uyuni-server`. Therefore, to explicitly permit the JWTs to be forwarded downstream, a JWT must be configured to include `aud = ['mcp-server-uyuni', 'uyuni-server']`
- Matches the `preferred_username` claim to an existing user.

### New configuration variables (with defaults)

```properties
# OpenID Connect authorization
web.oidc.enabled = false
# The identity provider URI
web.oidc.idp.issuer =
# The JWKS endpoint of the identity provider (path only)
# If not provided, the server will fetch it from the OIDC discovery endpoint of the idP.
web.oidc.idp.jwks_path =
# Expected audience to which the JWT is issued
web.oidc.jwt.audience = uyuni-server
# The name of the Uyuni username claim
web.oidc.jwt.username_claim = preferred_username
```

Minimal configuration using defaults:

```properties
# OpenID Connect authorization
web.oidc.enabled = true
# The identity provider URI
web.oidc.idp.issuer = https://auth.example.com
```

## Documentation (TODO)

- [ ] **DONE**

## Test coverage
- Unit tests were added

## Links

Fixes https://github.com/SUSE/spacewalk/issues/28483

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/suse/manager/webui/services/OidcAuthHandler.java
+++ b/java/core/src/main/java/com/suse/manager/webui/services/OidcAuthHandler.java
@@ -1,0 +1,287 @@
+/*
+ * Copyright (c) 2025 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+package com.suse.manager.webui.services;
+
+import com.redhat.rhn.common.conf.ConfigDefaults;
+import com.redhat.rhn.common.conf.ConfigException;
+import com.redhat.rhn.common.util.http.HttpClientAdapter;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.util.EntityUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jose4j.jwa.AlgorithmConstraints;
+import org.jose4j.jwk.HttpsJwks;
+import org.jose4j.jws.AlgorithmIdentifiers;
+import org.jose4j.jwt.JwtClaims;
+import org.jose4j.jwt.MalformedClaimException;
+import org.jose4j.jwt.consumer.InvalidJwtException;
+import org.jose4j.jwt.consumer.JwtConsumer;
+import org.jose4j.jwt.consumer.JwtConsumerBuilder;
+import org.jose4j.keys.resolvers.HttpsJwksVerificationKeyResolver;
+import org.jose4j.keys.resolvers.VerificationKeyResolver;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Handles OpenID Connect (OIDC) authorization.
+ */
+public class OidcAuthHandler {
+
+    public static final String OIDC_DISCOVERY_PATH = "/.well-known/openid-configuration";
+
+    private static final Logger LOG = LogManager.getLogger(OidcAuthHandler.class);
+
+    private boolean oidcEnabled;
+    private String issuer;
+    private String jwksUri;
+    private String audience;
+    private String usernameClaim;
+
+    private JwtConsumer jwtConsumer;
+    private HttpClientAdapter httpClient;
+
+    /**
+     * Constructs an OidcAuthHandler and loads configuration.
+     */
+    public OidcAuthHandler() {
+        this(null, new HttpClientAdapter());
+    }
+
+    /**
+     * Constructs an OidcAuthHandler with a custom {@link VerificationKeyResolver}.
+     *
+     * @param keyResolver the verification key resolver, if {@code null} a default one is created
+     * @param httpClientIn the HTTP client to use when fetching from the discovery endpoint
+     */
+    public OidcAuthHandler(VerificationKeyResolver keyResolver, HttpClientAdapter httpClientIn) {
+        httpClient = httpClientIn;
+
+        try {
+            loadConfiguration();
+            if (!isOidcEnabled()) {
+                return;
+            }
+        }
+        catch (URISyntaxException eIn) {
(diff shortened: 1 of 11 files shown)


java/code/src/com/suse/manager/webui/controllers/login/LoginController.java
```diff
@@ -221,6 +225,61 @@ private static String performLogin(Request request, Response response, boolean a
...
+        String authHeader = request.headers("Authorization");
+        if (authHeader == null || !authHeader.startsWith("Bearer ")) {
+            log.error("OIDC AUTH FAILURE: Missing or invalid Authorization header.");
+            response.status(HttpServletResponse.SC_UNAUTHORIZED);
+            return SparkApplicationHelper.json(response, new LoginResult(false,
+                    "Missing or invalid Authorization header."), new TypeToken<>() { });
+        }
+
+        String token = authHeader.substring(7); // Removes "Bearer "
+        Long userId;
```
rjmateus: Is correct to assume that customers must add the UserID to the OIDC system? That Id is internal in the database, and the only place where it's visible is in the URL. 
I think would make more sense to ask customers to define the username instead. It would be more user-friendly and it's still unique on the system.

java/code/src/com/suse/manager/webui/controllers/login/LoginController.java
```diff
@@ -76,6 +79,7 @@ private LoginController() { }
     public static void initRoutes(JadeTemplateEngine jade) {
         get("/manager/login", withCsrfToken(LoginController::loginView), jade);
         post("/manager/api/login", LoginController::login);
+        get("/manager/api/oidcLogin", LoginController::performOidcLogin);
```
rjmateus: Have a login method using the "GET" method looks strange to me. It's doing changes on the server side, since it creates the session. 
Do you think it would make more sense to use POST instead?

java/code/src/com/suse/manager/webui/controllers/login/LoginController.java
cbbayburt: In a discussion a while ago someone mentioned that usernames could change in theory so that's why I went with the ID, but I just realized that we don't allow this anyway, so yes it definitely makes more sense to use the usernames.

java/code/src/com/suse/manager/webui/controllers/login/LoginController.java
cbbayburt: Right, even though the login requests don't require a body, considering the state change, `POST` would be semantically correct. :+1:

java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
```diff
@@ -403,6 +403,68 @@ public class ConfigDefaults {
...
+    }
+
+    /**
+     * The name of the claim that provides the internal Uyuni username
+     * <p>
+     * This claim is used to match a user of the idP to a user in Uyuni.
+     * @return the name of the Uyuni username claim
+     */
+    public String getOidcUsernameClaim() {
+        return Config.get().getString(OIDC_JWT_USERNAME_CLAIM, "uyuni_username");
```
Firstyear: Not every provider will set this claim potentially, and it would require every provider to map some value into the field. Additionally, this will create issues with any username changes going forward.

You should probably be using `sub` and have some kind of internal mapping to an uyuni-user instead, or even consider a way to map users via an external service such as LDAP/SCIM.

Firstyear: Additionally, if you intend to use "username" this should default to preferred_username which is a standard claim per https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

java/code/src/com/suse/manager/webui/controllers/login/LoginController.java
```diff
@@ -221,6 +225,61 @@ private static String performLogin(Request request, Response response, boolean a
...
 
+    /**
+     * Perform an OpenID connect authentication, if enabled and properly configured.
+     * @param request the request object
+     * @param response the response object
+     * @return the JSON result of the login operation
+     */
+    private static String performOidcLogin(Request request, Response response) {
+        if (!OIDC_AUTH_HANDLER.isOidcEnabled()) {
+            log.error("OIDC AUTH FAILURE: OpenID Connect authentication is disabled.");
```
Firstyear: nitpick - OIDC is *authorisation* not *authentication*.

java/core/src/main/java/com/suse/manager/webui/services/OidcAuthHandler.java
```diff
@@ -0,0 +1,287 @@
...
+            if (!isOidcEnabled()) {
+                return;
+            }
+        }
+        catch (URISyntaxException eIn) {
+            throw new ConfigException("Malformed URI in the OIDC configuration.");
+        }
+        VerificationKeyResolver resolver = keyResolver;
+        if (resolver == null) {
+            HttpsJwks httpsJwks = new HttpsJwks(jwksUri);
```
Firstyear, line 91: Not sure if this is the case, but you probably can cache this public key as these are infrequently rotated.

java/code/src/com/redhat/rhn/common/conf/ConfigDefaults.java
cbbayburt: We want to keep the MCP server decoupled from the product as much as we can, so we didn't want to store "sub"s in MLM to avoid keeping an internal mapping table. Also, in MLM, usernames are not allowed to be changed.

I didn't know about `preferred_username`, though. I'll make the change to use it.

java/core/src/main/java/com/suse/manager/webui/services/OidcAuthHandler.java
cbbayburt: Yes, this library caches the key for 5 mins by default.

cbbayburt:
@Firstyear thank you very much for the review. I'll address your comments accordingly.

PR #11088, opened 2025-10-29
QE: Configure the interface of pxeboot for each product

Pablogoliva:
## What does this PR change?

TW by default is naming the interfaces as ensx instead of ethx, given that I'm going to do a different version of the step to ssh the pxeboot minion for each of the products. Maybe this is just a workaround and in the future we can rename the interfaces in TW instead, but currently is a good solution because the release of uyuni is near, we have to into account that this change in sumaform is going to be tricky and maybe we break more things in the CI with this change.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were edited

- [x] **DONE**

## Links

Issue(s): none.
Port(s): 
- Manager 5.1: https://github.com/SUSE/spacewalk/pull/28790
- Manager 5.0: https://github.com/SUSE/spacewalk/pull/28791
- Manager 4.3: https://github.com/SUSE/spacewalk/pull/28792

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/secondary/proxy_container_cobbler_pxeboot.feature
+++ b/testsuite/features/secondary/proxy_container_cobbler_pxeboot.feature
@@ -74,9 +74,21 @@ Feature: PXE boot a terminal with Cobbler and containerized proxy
     When I restart cobbler on the server
     Then service "cobblerd" is active on "server"
 
+# Workaround to ssh the pxeboot minions through different interfaces for each product,
+#  maybe in the future we can rename the interfaces directly in sumaform
+@uyuni
   Scenario: PXE boot the PXE boot minion
     When I set the default PXE menu entry to the target profile on the "server"
-    And I reboot the Cobbler terminal "pxeboot_minion"
+    And I reboot the Cobbler terminal "pxeboot_minion" through the interface "ens4"
+    And I wait for "60" seconds
+    And I set the default PXE menu entry to the local boot on the "server"
+    And I wait at most 1200 seconds until Salt master sees "pxeboot_minion" as "unaccepted"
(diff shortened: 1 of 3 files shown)


Pablogoliva:
I know that renaming the interface in TW would make more sense from a maintenance point of view, but it's not going to be that easy and let's do it as a workaround.

NamelessOne91 (commented):
Please, run a pipeline on this branch to verify that the problems are indeed caused by the interface name.

If that's the case, we should verify if we can setup the same network interface name for Tumbleweed and other OSes through Sumaform.
If not, maybe we can also change the interface name just inside the step definition, based on the Proxy OS. (although I actually don't mind being explicit in the scenario definition).

Pablogoliva:
> So in theory this looks fine. I'm worried about using two different interfaces for the same minion though and that we're working around the product maybe and hiding things by hard coding the interface. Also this might be fragile in the future. For now I'll approve so we move forward but please keep this workaround in mind to see later if it breaks

On the one hand the interfaces in TW are called like this by default, it's not a matter of sumaform I have checked it manually installing a TW and a slmicro6.1, and on the other hand  we are just sshing the pxeboot minion from the proxy via ipv6 and to do this we need to specify the interface which is connected to this private network. Therefore FMPOV we are not hiding things, I'm making this more difficult to maintain but at the same time I'm modifying less the machines to make a real testing and not overengineering the test at the same time.
But yes, I will take into account and I'm going to document it in a card inside the testsuite board to track it, and also is commented in the feature.

PR #11106, opened 2025-11-01
Improve Dockerfiles

mayrstefan:
## What does this PR change?

Referring to discussion #11073
- remove unnecessary layers through multiple run commands
- add chown or chmod to COPY commands to avoid unnecessary layers
- remove outdated add_repos and remove_unused Shell scripts
- add zypper clean and logfile deletion where missing
- add suseconnect.log to deleted log files
- remove unnecessar file deletions if files are still present in previous layers

## GUI diff

- [x] **DONE**

No difference.

## Documentation

- [x] **DONE**

## Test coverage
- No tests: the images should basically stay the same

- [x] **DONE**

## Links

Discussion #11073

- [x] **DONE**

## Changelogs

- [x] No changelog needed: if I understand the docs correctly no package is changed so no changelog is needed. But I'm not sure about that.

## Re-run a test

# Before you merge


containers/proxy-httpd-image/Dockerfile
```diff
@@ -9,40 +9,32 @@ ARG PXYRELEASENOTESPKG="release-notes-uyuni-proxy"
...
+        spacewalk-proxy-broker \
+        spacewalk-proxy-redirect \
+        spacewalk-proxy-html \
+        ${PXYRELEASENOTESPKG} \
+        python3-rhnlib \
+        python3-PyYAML && \
+    sysconf_addword /etc/sysconfig/apache2 APACHE_MODULES headers && \
+    sysconf_addword /etc/sysconfig/apache2 APACHE_MODULES proxy_http && \
+    zypper --non-interactive clean --all && \
+    rm -rf /var/log/{alternatives.log,lastlog,tallylog,suseconnect.log,zypper.log,zypp/history,YaST2}
```
cbosdo:
```suggestion
RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses  \
        spacewalk-proxy-broker \
        spacewalk-proxy-redirect \
        spacewalk-proxy-html \
        ${PXYRELEASENOTESPKG} \
        python3-rhnlib \
        python3-PyYAML && \
    sysconf_addword /etc/sysconfig/apache2 APACHE_MODULES headers && \
    sysconf_addword /etc/sysconfig/apache2 APACHE_MODULES proxy_http && \
    zypper --non-interactive clean --all && \
    rm -rf /var/log/{alternatives.log,lastlog,tallylog,suseconnect.log,zypper.log,zypp/history,YaST2}
```
I would drop the repo add: this is only to build the container images locally without OBS… which we never do anymore: we always build in OBS branches or using `osc build`

containers/proxy-httpd-image/remove_unused.sh
```diff
@@ -1,10 +0,0 @@
-#!/bin/bash
-# Removes any unnecessary files and packages before moving to the next build stage
-
-set -xe
-
-# remove locale data
-rm -rf /usr/share/locale
```
cbosdo, line 7: May be we could still remove the locale files to shrink the image a bit, but I have no idea if that is worth it

containers/proxy-salt-broker-image/Dockerfile
```diff
@@ -4,18 +4,15 @@
...
-COPY add_repos.sh /usr/bin
-RUN sh add_repos.sh ${PRODUCT_REPO}
-
 # Main packages
-RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses spacewalk-proxy-salt && zypper clean --all
+RUN [ -n "$PRODUCT_REPO" ] && zypper addrepo $PRODUCT_REPO product; \
+    zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
+        spacewalk-proxy-salt && \
+    zypper --non-interactive clean --all && \
+    rm -rf /var/log/{alternatives.log,lastlog,tallylog,suseconnect.log,zypper.log,zypp/history,YaST2}
```
cbosdo:
```suggestion
RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
        spacewalk-proxy-salt && \
    zypper --non-interactive clean --all && \
    rm -rf /var/log/{alternatives.log,lastlog,tallylog,suseconnect.log,zypper.log,zypp/history,YaST2}
```

containers/proxy-salt-broker-image/Dockerfile
```diff
@@ -25,12 +22,19 @@ RUN sh prepare_target.sh
...
 ARG REFERENCE_PREFIX="registry.opensuse.org/uyuni"
 
-RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses python3 python3-PyYAML python3-pyzmq && zypper clean --all
+RUN [ -n "$PRODUCT_REPO" ] && zypper addrepo $PRODUCT_REPO product; \
+    zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
+        python3 \
+        python3-PyYAML \
+        python3-pyzmq && \
+    zypper --non-interactive clean --all && \
+    rm -rf /var/log/{alternatives.log,lastlog,tallylog,suseconnect.log,zypper.log,zypp/history,YaST2}
```
cbosdo: I never realized that, but how comes we are installing packages in the slim layer? Probably this can be merged with the other install

containers/proxy-squid-image/Dockerfile
```diff
@@ -6,50 +6,34 @@ FROM $BASE AS base
...
-# Additional material
-COPY uyuni-configure.py /usr/bin/uyuni-configure.py
-RUN chmod +x /usr/bin/uyuni-configure.py
-
-COPY squid.conf /etc/squid/squid.conf
+RUN [ -n "$PRODUCT_REPO" ] && zypper addrepo $PRODUCT_REPO product; \
+    zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
+        python3 \
+        python3-PyYAML \
+        squid && \
```
cbosdo:
```suggestion
RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
        python3 \
        python3-PyYAML \
        squid && \
```

May be we can drop the `zypper ar` here too.

containers/proxy-squid-image/Dockerfile
```diff
@@ -6,50 +6,34 @@ FROM $BASE AS base
...
 # Ensure every run / exec uses the squid user to avoid errors like
```
cbosdo, line 12: can we have comments in the middle of a multi-line shell script?

containers/proxy-ssh-image/Dockerfile
```diff
@@ -4,40 +4,25 @@
...
 # Additional material
-COPY mgr-proxy-ssh-force-cmd /usr/sbin/mgr-proxy-ssh-force-cmd
-RUN chmod a+x /usr/sbin/mgr-proxy-ssh-force-cmd
-
-COPY uyuni-configure.py /usr/bin/uyuni-configure.py
-RUN chmod +x /usr/bin/uyuni-configure.py
-
-# Define slim image
-ARG BASE=registry.suse.com/bci/bci-base:15.7
-FROM $BASE AS slim
```
cbosdo, line 32: Why removing the slim image layer? Doesn't it reduce anything? IIRC the idea was to merge the layers into a single one here. Any trick to slim down images is welcomed… though the most important one would be the server image, but that is another pain.

containers/proxy-tftpd-image/Dockerfile
```diff
@@ -6,37 +6,25 @@ FROM $BASE AS base
...
-    python3-fbtftp \
-    python3-requests && \
-    sh remove_unused.sh
+RUN [ -n "$PRODUCT_REPO" ] && zypper addrepo $PRODUCT_REPO product; \
+    zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
+        python3-PyYAML \
+        python3-fbtftp \
+        python3-requests && \
+    zypper --non-interactive clean --all && \
+    rm -rf /var/log/{alternatives.log,lastlog,tallylog,suseconnect.log,zypper.log,zypp/history,YaST2}
```
cbosdo:
```suggestion
RUN zypper --gpg-auto-import-keys --non-interactive install --auto-agree-with-licenses \
        python3-PyYAML \
        python3-fbtftp \
        python3-requests && \
    zypper --non-interactive clean --all && \
    rm -rf /var/log/{alternatives.log,lastlog,tallylog,suseconnect.log,zypper.log,zypp/history,YaST2}
```

containers/proxy-httpd-image/Dockerfile
mayrstefan: So I guess that we can drop the PRODUCT_REPO completely

containers/proxy-httpd-image/remove_unused.sh
mayrstefan: In my tests I did not see any difference because most of that was present in the base layer. Deleting it from a different layer just hides it - it is still present in the previous layer - so this won't save any space.

containers/proxy-squid-image/Dockerfile
mayrstefan: In this case yes because it is a comment in the Dockerfile and not the shell script

containers/proxy-ssh-image/Dockerfile
mayrstefan: The trick is done wrong: 
- because fat and slim use the same base image the data from the base image is even doubled: in the layer of the base image and in the layer of the copy of the fat image
- the new image has less layers but the sum of it is bigger
- the trick could have been to use `FROM scratch` as slim layer. But because everything is copied that reduces layers, not size
- for single host deployments shared layers are good layers

The server image is a different beast: what Uyuni has done is just another anti-pattern. It moved more or less a complete VM into a single container. But slicing that into multiple containers will make things even more complicated and you will need more storage. The reason are shared dependencies that are hard to translate into shared layers. This will result in duplicated data on disk. That can be partially mitigated by a small tree of shared base images:
- uyuni-base
  -  uyuni-base-java
      - uyuni-tomcat
      - uyuni-taskomatic
  - uyuni-base-python
      - uyuni-cobbler
      - uyuni-salt-master
      - uyuni-spacewalk-reposync
      - ...

containers/proxy-salt-broker-image/Dockerfile
mayrstefan: This is an image a need to check when I can see what gets installed and copied with that /target folder. My gut feeling is that this slim layer is not worth it. The slim image installs some components to also get the native c libraries we need in the /lib folders.
On other thing we lose when we copy those files between mutiple stages is the rpm database. That could be handy when we try to scan for vulnerabilites or when we need an SBOM (the cyber resilience act will require software vendors like SUSE to provide SBOMs - the clock is already ticking)

containers/proxy-ssh-image/Dockerfile
cbosdo: > The trick is done wrong:

This is something we did a long time ago while still discovering the containers. Let's remove this thing indeed and we'll see what can be done later.

> The server image is a different beast: what Uyuni has done is just another anti-pattern. It moved more or less a complete VM into a single container. But slicing that into multiple containers will make things even more complicated and you will need more storage.

This was on purpose to get things progress. I don't even think we will ever reach a state where everything has its container: this would make tho whole thing much more complex than needed. The current idea is to split component that could bring a benefit to either the users or the devs / release engineers.

One of the possible ways to reduce the size of that container is to remove some of the useless things in it. For instance we install rpmbuild and all its dependencies just to package the SSL CA cert for kiwi image builds. Perl is probably also adding some weight there too.

containers/proxy-httpd-image/Dockerfile
mayrstefan: The `osc branch` is missing one argument. What package should I specify?

cbosdo: any one, but ideally one that is changed by your PR like `server-image`

containers/proxy-salt-broker-image/Dockerfile
cbosdo: I just pushed a change to your PR to get rid of this: we can look at the image size later.

containers/proxy-httpd-image/Dockerfile
cbosdo: I added a commit removing all those repos in your PR. I'll test it together with another on that remove the init-image

containers/proxy-httpd-image/Dockerfile
```diff
@@ -9,40 +9,31 @@ ARG PXYRELEASENOTESPKG="release-notes-uyuni-proxy"
 
 ARG PRODUCT_REPO
```
rjmateus: With the changes you added, looks to me that this variable is not needed anymore.

cbosdo: indeed, I overlooked it

cbosdo:
The images are now building fine together with the changes from https://github.com/uyuni-project/uyuni/pull/11133 in https://build.opensuse.org/project/show/home:cbosdonnat:branches:systemsmanagement:Uyuni:Master

containers/proxy-ssh-image/Dockerfile
cbosdo: slim image removed: resolving the conversation

containers/proxy-salt-broker-image/Dockerfile
cbosdo: merged together after removing the slim image

cbosdo:
The proxy images are working: I just tested them here

PR #11107, opened 2025-11-01
Fix Sonar cloud warnings

NeimadTL:
Sonar reported a bunch of non compliant variables so they've fixed with either ``@SuppressWarnings("unchecked")` or they've been renamed properly using camel case.

Fix for #9937


NeimadTL:
@CDellaGiusta @lucidd, please have a look when you'll have time, thanks.

CDellaGiusta:
When there is a SonarCloud warning, you are supposed to fix it, if it is not a false positive (as in these cases).
@SuppressWarnings("unchecked") means you tell SonarCloud not to issue any warning, and that is not the correct way of fixing a warning.
Please review your code complying and fixing the warnings

NeimadTL:
Thanks for your feedback @CDellaGiusta. Most of the @SuppressWarnings("unchecked") were added on either constants or variable that start with an acronyme and that's why I did change them. Any suggestion will be appreciated.

PR #11108, opened 2025-11-03
Ensure Proxy Extension Channel is present when converting to proxy (bsc#1249438)

rjpmestre:
## What does this PR change?

According to documentation, converting a client to a proxy requires the client to be subscribed to the appropriate Proxy Extension Channel or have the `mgrpxy` tool already installed.  

The reason is that applying a proxy configuration triggers a `mgrpxy` installation if it is not already present.
In scenarios where `mgrpxy` is missing and there is no appropriate Proxy Extension Channel subscribed, the installation will fail 

This PR ensures that when entering the proxy configuration (which triggers the same mechanism as clicking the "Convert to Proxy" button),  the system checks if mgrpxy is (or can be) installed. If is not, will look for and automatically automatically for suitable channels.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28398

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
+++ b/java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2025 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
+ */
+
+package com.suse.proxy.get.formdata;
+
+import static com.suse.proxy.get.ProxyConfigGetFacadeImpl.MGRPXY;
+import static com.suse.utils.Predicates.isAbsent;
+
+import com.redhat.rhn.common.conf.ConfigDefaults;
+import com.redhat.rhn.domain.channel.Channel;
+import com.redhat.rhn.domain.server.Server;
+import com.redhat.rhn.domain.user.User;
+import com.redhat.rhn.manager.channel.ChannelManager;
+import com.redhat.rhn.manager.entitlement.EntitlementManager;
+import com.redhat.rhn.manager.rhnpackage.PackageManager;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Pre-condition checks before handling Proxy Config Get Form Data
+ * The main goal is to detect blockers that would prevent a server from being converted to a proxy.
+ * These may be:
+ * a) Server does not have the proxy entitlement and is not entitleable for it.
+ * b) Server is not subscribed to the appropriate proxy extension channel (ie, at least one containing the mgrpxy
+ * package), it is subscribed to it.
+ */
+public class ProxyConfigGetFormDataPreConditions implements ProxyConfigGetFormDataContextHandler {
+
+    @Override
+    public void handle(ProxyConfigGetFormDataContext context) {
+        Server server = context.getServer();
+
+        if (isAbsent(server)) {
+            context.getErrorReport().register("Server not found");
+            return;
+        }
+
+        if (server.isMgrServer()) {
+            context.getErrorReport().register("The system is a Management Server and cannot be converted to a Proxy");
+            return;
+        }
+
+        if (!server.hasProxyEntitlement() &&
+                !context.getSystemEntitlementManager().canEntitleServer(server, EntitlementManager.PROXY)) {
+            context.getErrorReport().register("Cannot entitle server ID: {0}", server.getId());
+        }
+
+        if (ConfigDefaults.get().isUyuni() || isMgrpxyInstalled(server)) {
+            return;
+        }
+
+        Set<Channel> subscribableMgrpxyChannels = getSubscribableMgrpxyChannels(server, context.getUser());
+        context.setSubscribableChannels(subscribableMgrpxyChannels);
+        if (subscribableMgrpxyChannels == null) {
+            context.getErrorReport().register(
+                    "No channel with mgrpxy package found for server ID: {0}", server.getId());
+        }
+
+    }
+
+    /**
+     * Check if mgrpxy is installed on the server
+     * @param server the server to check
+     * @return true if installed, false otherwise
+     */
+    private boolean isMgrpxyInstalled(Server server) {
+        return PackageManager.shallowSystemPackageList(server.getId())
+                .stream().anyMatch(p -> p.getName().equals(MGRPXY));
+    }
+
+    /**
+     * Returns the set of channels providing the mgrpxy package that should be subscribed to.
+     * Logic:
+     * - If mgrpxy is already available from a subscribed channel, returns an empty set.
+     * - If mgrpxy is available only from non-subscribed channels, returns those channels.
+     * - If no channels provide mgrpxy at all, returns {@code null}.
+     *
+     * @param server the server to inspect
+     * @param user   the user requesting the channels
+     * @return set of subscribable channels providing mgrpxy, an empty set if already subscribed,
+     * or {@code null} if unavailable
+     */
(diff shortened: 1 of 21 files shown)


java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        if (!server.hasProxyEntitlement() &&
+                !context.getSystemEntitlementManager().canEntitleServer(server, EntitlementManager.PROXY)) {
+            context.getErrorReport().register(
+                    "Cannot entitle server ID: {1}", server.getId());
+        }
+
+        if (!ConfigDefaults.get().isUyuni() && !isMgrpxyInstalled(server) && !isMgrpxyAvailable(server)) {
+            Set<Channel> subscribableChannelsWithMgrpxy = getSubscribableChannelsWithMgrPxy(server, context.getUser());
```
mcalmer: Can you check for system being a "MLM Server" ? A Server should not be migrated to a Proxy.

java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param user the user requesting the channels
+     * @return a set of subscribable channels with the mgrpxy package
+     */
+    private Set<Channel> getSubscribableChannelsWithMgrPxy(Server server, User user) {
+        return server.getBaseChannel().getAccessibleChildrenFor(user).stream()
+                .filter(c -> !server.isSubscribed(c))
+                .filter(c -> c.getPackages()
+                        .stream().anyMatch(p -> MGRPXY.equals(p.getPackageName().getName()))
+                )
+                .collect(Collectors.toSet());
```
mcalmer: Check ChannelManager.listLatestPackagesEqual() or ChannelManager.listLatestPackagesLike()
Maybe this can be help.

This is what similar code did already in the past. It was for checking if a "kickstart package" was installed or installable on a system.
But this is just to give you an option. If this works, I am fine.

java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
rjpmestre: Good catch!

java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
rjpmestre: Yep, seems to work just fine. I ended up using that same method to check if `mgrpxy` is available on a subscribed channel. No need to have a dedicated (and different) way of checking that.
Actually, I’m now wondering if it’s even worth checking whether `mgrpxy` is installed. For all we, deploying a proxy config will guarantee it will be installed anyway. What do you think?

java/code/src/com/suse/proxy/get/formdata/test/ProxyConfigGetFormDataPreConditionsTest.java
```diff
@@ -0,0 +1,394 @@
...
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
```
mcalmer: JFYI - when we create a new file, we can omit the last paragraph about the trademark.

java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
```diff
@@ -0,0 +1,137 @@
...
+                        .collect(Collectors.partitioningBy(
+                                server::isSubscribed,
+                                Collectors.toSet()
+                        ));
+
+        Set<Channel> subscribed = channelsBySubscription.getOrDefault(true, Set.of());
+        Set<Channel> unsubscribed = channelsBySubscription.getOrDefault(false, Set.of());
+
+        if (!subscribed.isEmpty()) {
+            return Set.of();
```
mcalmer, line 116: What happens when we have channels in subscribed and unsubscribed?
For 15 SP7 we have a pool and update channel. The typical case is, that either both are subscribed or not, but when cloned or in Uyuni case it might happen that you have only one.

web/html/src/manager/minion/proxy/proxy-config-types.ts
```diff
@@ -39,18 +39,18 @@ export type ProxyConfigModel = {
...
   disabled: boolean;
-};
+}
 
 export type ProxyConfigProps = {
   serverId: string;
   parents: Parent[];
   currentConfig: ProxyConfigModel;
-  initFailMessage?: string;
+  validationErrors?: any[];
```
Etheryte: Could we possibly have a better type here? Or can it literally be anything?

web/html/src/manager/minion/proxy/proxy-config.tsx
```diff
@@ -46,9 +46,9 @@ export function ProxyConfig({
   useEffect(() => {
     setModel((prev) => ({ ...prev }));
 
-    if (initFailMessage) {
+    if (validationErrors && validationErrors.length > 0) {
       setSuccess(false);
-      setMessages([initFailMessage]);
+      setMessages(validationErrors.map((error) => error.message || error));
```
Etheryte: If I understood correctly, we're passing this data to the frontend straight from the backend? Then it would be preferable to pass it in the right format right away, so we don't need to process it as soon as we get it.

web/html/src/manager/minion/proxy/proxy-config.tsx
```diff
@@ -141,312 +141,313 @@ export function ProxyConfig({
...
-            label={t("Proxy admin email")}
-            hint={t("The container-based proxy administrator email address")}
-            placeholder={t("e.g., proxy.admin@mycompany.com")}
-            required
-            labelClass="col-md-3"
-            divClass="col-md-6"
-          />
-          <hr />
-          {hasCertificates && (
+      {!validationErrors ||
```
Etheryte: This JSX conditional and the next one _look_ like a bug, are you sure that's what you meant? It looks like the parens are incorrectly placed and create an extra nesting level.

java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
rjpmestre: In a scenario like that, as long as we detect a subscribed channel able to provide `mgrpxy`, we will not automatically subscribe to a new channel.

mcalmer: But you would maybe use an outdated package if you have pool, but not update.
It depends what you want.
If any package is ok, fine.
But finding the other might be complicated.

mcalmer (dismissed):

web/html/src/manager/minion/proxy/proxy-config.tsx
rjpmestre: I see your point. I adjusted to pass only the error strings. Let me know if it works ok

web/html/src/manager/minion/proxy/proxy-config.tsx
rjpmestre: Oh! My bad. I trusted an IDE suggestion to adjust it. Didn't notice it was creating the extra encapsulation

Etheryte (approved):
This looks good, thank you very much for the time you've put into this. 👍

java/code/src/com/suse/proxy/get/formdata/ProxyConfigGetFormDataPreConditions.java
rjpmestre: Ah, I see your point. That's a valid concern. 
Having a second look at this made me wonder if we should also enforce `podman`  (ie,  `Containers Module` channel). This is pretty much a system's pre assumption as well.
But as these might go a bit beyond the scope of this issue, i  went ahead and created a separate [issue](https://github.com/SUSE/spacewalk/issues/28928) to track and refine these points. Is that ok?

PR #11126, opened 2025-11-05
Use an inject polyfill script to fix the cucumber report error

maximenoel8:
## Description

The Cucumber HTML report is failing to load when accessed via its web URL, rendering it unusable for remote review. The page fails to initialize due to a critical JavaScript error:
```
Uncaught TypeError: crypto.randomUUID is not a function
```

This error confirms that the client browser is missing or restricting access to the crypto.randomUUID function from the modern Web Crypto API. This issue currently makes it impossible to examine the report without manually downloading and opening the file locally.

## Cause

 The report was not opening because the JavaScript code used to generate the interactive interface immediately crashed when run in a browser.

 -  **Unsupported Feature**: The Cucumber reporter relies on a modern function called crypto.randomUUID() from the browser's Web Crypto API.

  - **Browser Failure**: The specific browser environment (when accessing the file over HTTP) either did not support this feature or did not expose it correctly to the script, resulting in the error: TypeError: crypto.randomUUID is not a function.

Although the server successfully sent the complete HTML file, the broken JavaScript meant the browser could not render the content, making the page appear empty or stuck.

## Tests

I tried to use different browser with the same results. The solution I found was to remove the function from the html page.

### Main Change
The fix involves implementing a polyfill (short for "poly-filling a hole"), which is a block of JavaScript code that provides the missing functionality.

Insert a small JavaScript function into the <head> of the generated HTML.
Logic: The script checks if window.crypto.randomUUID exists. If it doesn't, it defines the function using a simple, universally compatible fallback based on Math.random() to generate unique IDs.

### Secondary change

Align parallel namespace and cucumber name calling the filename and run_set the same.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28841

- [ ] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/Rakefile
+++ b/testsuite/Rakefile
@@ -12,6 +12,49 @@ junit_results = '-f junit -o results_junit'
 build_number = ENV.fetch('BUILD_NUMBER', nil)
 result_folder = build_number.nil? ? 'results' : "results/#{build_number}"
 
+# Defines a portable JavaScript polyfill for crypto.randomUUID
+# See issue https://github.com/SUSE/spacewalk/issues/28841 for explanations
+def inject_polyfill_script(html_results)
+  # Define the polyfill script using a Ruby Heredoc (multiline string).
+  polyfill_script = <<-JS_POLYFILL
+<script>
+if (!window.crypto || !window.crypto.randomUUID) {
+  window.crypto = window.crypto || {};
+  window.crypto.randomUUID = function() {
+    var d = new Date().getTime();
+    var d2 = (performance && performance.now && (performance.now() * 1000)) || 0;
+    return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, function(c) {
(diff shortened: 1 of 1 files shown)


Etheryte:
For your consideration, if I'm not mistaken this stems from the fact that you're using HTTP, not HTTPS. I would recommend using HTTPS instead. [See MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID):

> Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility).

maximenoel8:
Solve by using https

PR #11135, opened 2025-11-05
Add AppStreams to SSM

wweellddeerr:
## What does this PR change?

This PR introduces AppStreams configuration feature to SSM through WebUI and API.

## Codespace

## GUI diff

After:

image

image

image

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24196

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


web/html/src/manager/systems/appstreams/ssm-appstreams-configure.tsx
```diff
@@ -0,0 +1,134 @@
...
+          onActionChange={handleActionChange}
+          getModuleAction={getModuleAction}
+        />
+      );
+    }
+  };
+
+  return (
+    <>
+      {scheduledMsg.length > 0 && <Messages items={scheduledMsg} />}
```
Etheryte: Is this necessary? Doesn't `Messages` already render nothing if there's no items?

web/html/src/manager/systems/appstreams/ssm-appstreams-channel-selection.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { Column } from "components/table/Column";
+import { HierarchicalRow, HierarchicalTable } from "components/table/HierarchicalTable";
+
+type ChannelSelectionProps = {
+  channels: HierarchicalRow[];
+};
+
+export const AppStreamsChannelSelection: React.FC<ChannelSelectionProps> = ({ channels }) => {
+  const getRowClass = (row: HierarchicalRow, index: number): string => {
```
Etheryte: I don't think this is necessary? Iirc we currently don't color separate even and odd rows, and if we do want to, we can do it in CSS. Please feel free to correct me if I'm wrong.

web/html/src/manager/systems/appstreams/ssm-appstreams-channel-selection.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from "react";
+
+import { Column } from "components/table/Column";
+import { HierarchicalRow, HierarchicalTable } from "components/table/HierarchicalTable";
```
Etheryte: Please use `Table` instead of this component, it's marked for removal https://github.com/SUSE/spacewalk/issues/28408. Now that I look at it, this should be marked as `@deprecated`, I will go do that now.

web/html/src/manager/systems/appstreams/ssm-appstreams-configure-list.tsx
```diff
@@ -0,0 +1,114 @@
...
+    <>
+      <p>{t("The following AppStream modules are available on the selected channel.")}</p>
+      <AppStreamActions numberOfChanges={numberOfChanges} onReset={onReset} onSubmit={onSubmitChanges} />
+      <Panel
+        headingLevel="h4"
+        icon="spacewalk-icon-software-channels"
+        title={channel.name}
+        buttons={[changeChannelButton]}
+      >
+        <table className="table table-striped">
```
Etheryte: Wouldn't it be more appropriate to use a `Table` here instead? If we ever want to change styling or other similar things, then we don't have to come and manually change this table.

web/html/src/manager/systems/appstreams/ssm-appstreams-configure.tsx
wweellddeerr: No, it is not necessary. Thanks for pointing it out, I just removed it.

web/html/src/manager/systems/appstreams/ssm-appstreams-channel-selection.tsx
wweellddeerr: You’re totally right. I was trying to mimic a similar channel selection table from the SSM JSP pages, but I ended up implementing something that doesn’t even exist there. I’ve removed it now, thanks for pointing it out.

web/html/src/manager/systems/appstreams/ssm-appstreams-channel-selection.tsx
wweellddeerr: Done, thank you.

web/html/src/manager/systems/appstreams/ssm-appstreams-configure-list.tsx
wweellddeerr: I’m actually currently evaluating it, as the `Table` component seems to be prepared for a flat structure, while this layout relies on `rowSpan` to group multiple AppStream rows under the same Module and Action.

Etheryte: `Table` supports nested rows, please see the example in storybook.

wweellddeerr: I refactored the component to use `Table`, thank you.

web/html/src/manager/systems/appstreams/ssm-appstreams-channel-selection.tsx
```diff
@@ -0,0 +1,50 @@
...
+          <Table data={channels} identifier={(row) => row.id}>
+            <Column
+              header="Channel Name"
+              columnKey="channelName"
+              columnClass="col"
+              cell={(row) => (
+                <>
+                  {row.parentId && (
+                    <img
+                      style={{ marginLeft: "4px", marginRight: "8px", verticalAlign: "middle" }}
```
Etheryte, line 32: Instead of inline styles, it would be preferable to use a css module, see e.g. `web/html/src/components/table/SearchPanel.module.scss` and `web/html/src/components/table/SearchPanel.tsx`.

Etheryte (approved):
Thank you very much for the updates, LGTM. 👍

java/code/src/com/redhat/rhn/manager/ssm/SsmManager.java
```diff
@@ -74,6 +76,15 @@ public class SsmManager {
 
     public static final String SSM_SYSTEM_FEATURE = "ftr_system_grouping";
 
+    private static final String LIST_SSM_SERVERS_IN_CHANNEL_SQL = """
+        SELECT DISTINCT ST.element as server_id
```
cbbayburt: Is DISTINCT necessary here?

java/code/src/com/suse/manager/webui/controllers/appstreams/AppStreamsController.java
```diff
@@ -192,4 +234,75 @@ public static String getPackagesInModule(Request req, Response res, User user) {
...
+    public static ModelAndView ssmAppstreamsChannelSelection(Request request, Response response, User user) {
+        List<ChannelTreeNode> channels = ChannelManager.getModularChannelsForSsm(user);
+        String channelsJson = GSON.toJson(channels);
+        Map<String, Object> data = new HashMap<>();
+        data.put("tabs", ViewHelper.getInstance().renderNavigationMenu(request, "/WEB-INF/nav/ssm.xml"));
+        data.put("channels", channelsJson);
+        return new ModelAndView(data, "templates/ssm/appstreams-channel-selection.jade");
+    }
+
+    private static ChannelAppStreamsJson getAppStreamChannelFromPath(User user, String channelId) {
```
cbbayburt: Why is this method called ..FromPath?

java/code/src/com/suse/manager/webui/controllers/appstreams/AppStreamsController.java
```diff
@@ -192,4 +234,75 @@ public static String getPackagesInModule(Request req, Response res, User user) {
...
+     *
+     * @param request the request
+     * @param response the response
+     * @param user the current user
+     * @return the ModelAndView object to render the page
+     */
+    public static ModelAndView ssmConfigureAppstreams(Request request, Response response, User user) {
+        var channel = getAppStreamChannelFromPath(user, request.params("channelId"));
+        if (channel == null) {
+            return ssmAppstreamsChannelSelection(request, response, user);
```
cbbayburt: I think a redirect would be safer here

schema/spacewalk/common/data/endpointNamespace.sql
```diff
@@ -5221,6 +5221,38 @@ INSERT INTO access.endpointNamespace (namespace_id, endpoint_id)
     WHERE ns.namespace = 'systems.migration' AND ns.access_mode = 'W'
     AND ep.endpoint = '/manager/api/systems/migration/schedule' AND http_method = 'POST'
     ON CONFLICT DO NOTHING;
+
+INSERT INTO access.endpointNamespace (namespace_id, endpoint_id)
+    SELECT ns.id, ep.id
+    FROM access.namespace ns, access.endpoint ep
+    WHERE ns.namespace = 'systems.appstreams' AND ns.access_mode = 'W'
```
cbbayburt: All the features in SSM tabs use the same namespace as the corresponding single-system feature. For example, SSM packages feature uses the same namespace as System > Software > Packages; SSM patches feature uses the same namespace as System > Software > Patches, etc.

Following the same pattern, you can use the existing `systems.software.appstreams` namespace instead. This way, modifying appstreams from the system details page, and from SSM is controlled by the same permission.

```suggestion
    WHERE ns.namespace = 'systems.software.appstreams' AND ns.access_mode = 'W'
```

schema/spacewalk/common/data/namespace.sql
```diff
@@ -687,6 +687,9 @@ INSERT INTO access.namespace (namespace, access_mode, description)
 INSERT INTO access.namespace (namespace, access_mode, description)
     VALUES ('systems.migration', 'W', 'Perform product migration on systems')
     ON CONFLICT (namespace, access_mode) DO NOTHING;
+INSERT INTO access.namespace (namespace, access_mode, description)
+    VALUES ('systems.appstreams', 'W', 'Manage AppStreams on systems (SSM)')
+    ON CONFLICT (namespace, access_mode) DO NOTHING;
```
cbbayburt: Since we're going to use the existing `systems.software.appstreams` namespace, you don't need to insert anything to the `namespace` table.
```suggestion
```

cbbayburt: BTW, if we were going to use a namespace, you would need to add a row into the `accessgroupnamespace` table for each access group that should have access to the new namespace by default.

If you don't add the namespace to any access group, effectively, only Uyuni admin (not even org admin) can access the namespace.

You don't need to do this for existing namespaces, unless you want to change the default access rules for them for some reason.

java/code/src/com/redhat/rhn/manager/ssm/SsmManager.java
wweellddeerr: It is actually not necessary, I removed it. Thanks.

java/code/src/com/suse/manager/webui/controllers/appstreams/AppStreamsController.java
wweellddeerr: Good catch. The `...FromPath` suffix is a leftover from a previous iteration where this value was strictly a path variable. Since I changed the logic later, the name is no longer accurate. I've renamed it to `resolveAppStreamChannel` and it looks much better now.

java/code/src/com/suse/manager/webui/controllers/appstreams/AppStreamsController.java
wweellddeerr: Agreed. A redirect keeps the URL consistent with the view. Changed to `response.redirect`.

schema/spacewalk/common/data/endpointNamespace.sql
wweellddeerr: This is a great point, I've updated the scripts to reuse the existing `systems.software.appstreams` namespace, it is a much better approach.

cbbayburt:
Thanks for addressing my comments, @wweellddeerr.

I have just one final suggestion:

For RBAC, we have this new ACL called `authorized_for(<namespace>)` that renders parts of the JSP conditionally, only if the user has access to the specific namespace. We don't have it everywhere, but I've been trying to use it as much as possible, especially when adding new stuff.

With this ACL, we can let the AppStreams tab render only if the user has access to `systems.software.appstreams`, so we can avoid 403 links as much as we can.

Here's some examples:

#### SSM tabs:
https://github.com/uyuni-project/uyuni/blob/8c6139b0a849a2e3864161e1b224719eb6659da7/java/code/webapp/WEB-INF/nav/ssm.xml#L27-L29

#### SSM index page shortcuts:
https://github.com/uyuni-project/uyuni/blob/8c6139b0a849a2e3864161e1b224719eb6659da7/java/code/webapp/WEB-INF/pages/ssm/ssmindex.jsp#L68-L80

cbbayburt (dismissed):

PR #11196, opened 2025-11-18
Move proxy initialization to run when proxy config is applied (bsc#1252948) 

rjpmestre:
## What does this PR change?

This PR moves the proxy initialization logic so that it gets executed only when the user applies a proxy configuration, instead of running immediately when the “Convert to Proxy” button is clicked.
As a result, a system will only be set/registered as a proxy after the configuration is confirmed, preventing partially initialized or cancelled conversions from appearing in the proxy system list.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28803

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/suse/proxy/ProxyConfigUtils.java
+++ b/java/core/src/main/java/com/suse/proxy/ProxyConfigUtils.java
@@ -7,10 +7,6 @@
  * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
  * along with this software; if not, see
  * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
- *
- * Red Hat trademarks are not licensed under GPLv2. No permission is
- * granted to use or replicate Red Hat trademarks that are incorporated
- * in this software or its documentation.
  */
 
 package com.suse.proxy;
@@ -26,7 +22,12 @@ import static com.suse.utils.Predicates.isAbsent;
 import static com.suse.utils.Predicates.isProvided;
 
 import com.redhat.rhn.common.util.FileUtils;
+import com.redhat.rhn.domain.channel.Channel;
 import com.redhat.rhn.domain.server.Pillar;
+import com.redhat.rhn.domain.server.Server;
+import com.redhat.rhn.domain.user.User;
+import com.redhat.rhn.manager.channel.ChannelManager;
+import com.redhat.rhn.manager.rhnpackage.PackageManager;
 
 import com.suse.manager.webui.utils.YamlHelper;
 import com.suse.proxy.model.ProxyConfig;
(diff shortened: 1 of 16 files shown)


java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
```diff
@@ -252,7 +252,7 @@ private void handleMissingProxyConfig(Pillar pillar, MinionServer proxy) {
                 throw new RhnRuntimeException(result.getMessage());
             }
         }
-        updater.update(request, systemManager, null);
+        updater.update(request, systemManager, GlobalInstanceHolder.SYSTEM_ENTITLEMENT_MANAGER, null);
```
cbosdo: Passing around the global instance is far for ideal for testability

java/code/src/com/redhat/rhn/frontend/xmlrpc/proxy/ProxyHandler.java
```diff
@@ -591,7 +593,9 @@ private int bootstrapProxy(User loggedInUser,
                 null, null, null
             );
 
-            proxyConfigUpdateFacade.update(request, systemManager, loggedInUser);
+            proxyConfigUpdateFacade.update(
+                    request, systemManager, GlobalInstanceHolder.SYSTEM_ENTITLEMENT_MANAGER, loggedInUser
```
cbosdo: Same here with the global instance.

java/code/src/com/suse/manager/webui/controllers/ProxyConfigurationController.java
```diff
@@ -144,7 +140,7 @@ public String updateProxyConfiguration(Request request, Response response, User
                 GSON.fromJson(request.body(), new TypeToken<ProxyConfigUpdateJson>() { }.getType());
 
         try {
-            proxyConfigUpdateFacade.update(data, systemManager, user);
+            proxyConfigUpdateFacade.update(data, systemManager, GlobalInstanceHolder.SYSTEM_ENTITLEMENT_MANAGER, user);
```
cbosdo: Here too

java/code/src/com/redhat/rhn/domain/action/HardwareRefreshAction.java
rjpmestre: Yeah, I agree its not ideal. The reason we’re passing `GlobalInstanceHolder.SYSTEM_ENTITLEMENT_MANAGER` instead of referencing it directly in the class is purely for testability. The way this `GlobalInstanceHolder` is setting everything just as static final fields makes hard for jmock to mock it.

rjpmestre: Update: ended up moving the "responsibility" of retrieving the entitlement manager into the context's getter. So, now we don’t need to pass it in the constructor. This way, when testing we can just mock this getter. Let me know what you think

cbosdo (approved):
LGTM

PR #11235, opened 2025-11-25
Fix exception error on spacecmd package refresh (bsc#1251995)

CDellaGiusta:
## What does this PR change?

While running the command

`spacecmd -u admin -p admin system_schedulepackagerefresh <minion>`

an error occourred:

`ERROR: com.redhat.rhn.common.translation.TranslationException: Could not find translator for class java.lang.String to class java.lang.Integer`

For some strange reason, the python script calling the rpc:

```
system_id = self.get_system_id(system)
self.client.system.schedulePackageRefresh(session, system_id, options.start_time )
```

is sending the system_id  as a string and not anymore as an integer.
Since no translator from string to integer is found, the error is raised.
         

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links

Issue(s): https://bugzilla.suse.com/show_bug.cgi?id=1251995
Port(s): 
5.1: https://github.com/SUSE/spacewalk/pull/29063
5.0: https://github.com/SUSE/spacewalk/pull/29064
- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/common/translation/Translator.java
+++ b/java/core/src/main/java/com/redhat/rhn/common/translation/Translator.java
@@ -168,6 +168,15 @@ public class Translator extends Translations {
     }
 
     /**
+     * Convert from String to Integer.
+     * @param str The string to convert
+     * @return the resulting Integer
+     */
+    public static Integer string2Int(String str) {
+        return Integer.parseInt(str);
+    }
+
+    /**
      * Convert from Double to String.
      * @param d The double to convert
      * @return The resulting string.
--- a/java/core/src/test/java/com/redhat/rhn/common/translation/test/TranslatorTest.java
+++ b/java/core/src/test/java/com/redhat/rhn/common/translation/test/TranslatorTest.java
@@ -17,6 +17,7 @@ package com.redhat.rhn.common.translation.test;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.redhat.rhn.common.translation.Translator;
@@ -106,6 +107,22 @@ public class TranslatorTest extends RhnBaseTestCase {
         assertFalse(Translator.string2boolean("false"));
         assertFalse(Translator.string2boolean("faLSe"));
         assertFalse(Translator.string2boolean("rock on"));
+    }
+
+    @Test
+    public void testString2Int() {
+        assertEquals(1000010006, Translator.string2Int("1000010006"));
+        assertEquals(123, Translator.string2Int("+123"));
+        assertEquals(0, Translator.string2Int("0"));
+        assertEquals(-1, Translator.string2Int("-1"));
+        assertEquals(-1000010006, Translator.string2Int("-1000010006"));
+
+        assertThrows(NumberFormatException.class, () -> Translator.string2Int(null));
+        assertThrows(NumberFormatException.class, () -> Translator.string2Int(""));
+        assertThrows(NumberFormatException.class, () -> Translator.string2Int("true"));
+        assertThrows(NumberFormatException.class, () -> Translator.string2Int("F"));
+        assertThrows(NumberFormatException.class, () -> Translator.string2Int("n"));
+        assertThrows(NumberFormatException.class, () -> Translator.string2Int("rock on"));
     }
 
     @Test
--- a/java/spacewalk-java.changes.carlo.uyuni-1251995-translation-exception
+++ b/java/spacewalk-java.changes.carlo.uyuni-1251995-translation-exception
@@ -1,0 +1,1 @@
+- Fix exception error on spacecmd package refresh (bsc#1251995)


mackdk (commented):
The change itself looks good, but I'm struggling to understand why it's needed. if the python code is sending a request with the system id wrapped in a `<string>` xml-rpc type instead of using `<i4>`, the request is wrong and should be fixed on the client. 

@lucidd What do you think?

CDellaGiusta:
>  I'm struggling to understand why it's needed.

Here is what I understood it happens:

- the customer is running this:
`spacecmd -u admin -p admin system_schedulepackagerefresh mlm-ci-51-suse-sshminion.mgr.suse.de`

- in spacecmd/src/spacecmd/system.py lines from 4517 on:
```
  for system in systems:
        system_id = self.get_system_id(system)
        if not system_id:
            continue

        self.client.system.schedulePackageRefresh(
            self.session, system_id, options.start_time
        )
```

- when intercepted in java, there is an exception: 
`com.redhat.rhn.common.translation.TranslationException: Could not find translator for class java.lang.String to class java.lang.Integer`
In the debugger you see that  the second parameter (system_id AKA id of the minion) is sent as a string and not as an integer

Could it be something related to the python version handling types in a different way?

mackdk:
> * when intercepted in java, there is an exception

What xml request is the server receiving? Is it something like:

```xml
        <param>
            <value>
                <string>1000010000</string>
            </value>
        </param>
```

or:

```xml
        <param>
            <value>
                <int>1000010000</int> <!-- <i4> </i4> is also valid -->
            </value>
        </param>
```

If it's the former, the client is sending a wrong request so the error is kind of expected. If it's the latter the server is processing an integer as a string for some reason and it should not happen.

>Could it be something related to the python version handling types in a different way?

Could be, my python knowledge is very limited on this front.

java/core/src/main/java/com/redhat/rhn/common/translation/Translator.java
```diff
@@ -167,6 +167,20 @@ public static boolean string2boolean(String str) {
...
+     * Convert from String to Integer.
+     * @param str The string to convert
+     * @return the resulting Integer
+     */
+    public static Integer string2Int(String str) {
+        try {
+            return Integer.parseInt(str);
+        }
+        catch (NumberFormatException ex) {
+            return 0;
```
rjmateus: This looks dangerous to me. It we cannot parse the number, returning 0 can hide errors and cause troubles.

CDellaGiusta: yes, this is a fair point. Amended, as well as the tests.

CDellaGiusta: Once approved here, I'll backport it to 5.1 and 5.0

lucidd:
> The change itself looks good, but I'm struggling to understand why it's needed. if the python code is sending a request with the system id wrapped in a `<string>` xml-rpc type instead of using `<i4>`, the request is wrong and should be fixed on the client.
> 
> @lucidd What do you think?

i agree it would be better to find out why this is happening and fix it in spacecmd.  
Those implicit conversation make the matching a lot more fragile and have been causing issues before so i'm not a big fan of adding more if not necessary.

CDellaGiusta:
> i agree it would be better to find out why this is happening and fix it in spacecmd. Those implicit conversation make the matching a lot more fragile and have been causing issues before so i'm not a big fan of adding more if not necessary.

ok, then I'll investigate the issue there and see what comes out

CDellaGiusta:
@mackdk @lucidd @rjmateus 

Here's the result of my preliminary investigation:

I tried to issue the following command:
>spacecmd -u admin -p admin system_schedulepackagerefresh client1.home.lab

This command sends the following request to the java XML backend:
```

<?xml version='1.0'?>
<methodCall>
<methodName>system.schedulePackageRefresh</methodName>
<params>
    <param>
    <value><string>781 ... 111</string></value>
    </param>
    <param>
    <value><string>1000010006</string></value>
    </param>
    <param>
    <value><dateTime.iso8601>20251126T08:20:53</dateTime.iso8601></value>
    </param>
</params>
</methodCall>
```

As you can see, the second parameter should be the minion ID, which is sent as a string instead of as an integer.
This way the backend does not find a "translator" from the supplied string argument to the expected integer one and issues an error.

It seems that the culprit is the python method "**get_system_id**" which is probably returning a string instead of an integer.

Do also consider the following:
1) if run with the minion ID instead of the minion name, the command do not issue an error

**this works!**
`spacecmd -u admin -p admin system_schedulepackagerefresh 1000010006`
**this issues an ERROR**
`spacecmd -u admin -p admin system_schedulepackagerefresh client1.home.lab`


2) all the spacecmd commands that use the method "get_system_id" have the very same error: here's a list of what I tried and failed:

- spacecmd -u admin -p admin system_schedulepackagerefresh client1.home.lab
- spacecmd -u admin -p admin system_show_packageversion client1.home.lab packagename
- spacecmd -u admin -p admin system_scheduleapplyconfigchannels client1.home.lab
- spacecmd -u admin -p admin system_listmigrationtargets client1.home.lab
- spacecmd -u admin -p admin system_reboot client1.home.lab

So, I'm closing the PR and all the ports. I forwarded bug  https://bugzilla.suse.com/show_bug.cgi?id=1251995 to ION since it's about python

PR #11266, opened 2025-12-01
SonarCloud error reduction fix step 5, miscellaneous 5 rules

CDellaGiusta:
## What does this PR change?

[SonarCloud fix: java:S6213 Restricted Identifiers should not be used as Identifiers](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS6213&rule_key=java%3AS6213)
[SonarCloud fix: java:S3358 Ternary operators should not be nested](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3358&rule_key=java%3AS3358)
[SonarCloud fix: java:S3599 Double Brace Initialization should not be used](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3599&rule_key=java%3AS3599)
[SonarCloud fix: java:S3012 Arrays and lists should not be copied using loops](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3012&rule_key=java%3AS3012)
[SonarCloud fix: java:S108 Nested blocks of code should not be left empty](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS108&rule_key=java%3AS108)

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Port(s):
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/core/src/test/java/com/redhat/rhn/frontend/xmlrpc/image/profile/test/ImageProfileHandlerTest.java
+++ b/java/core/src/test/java/com/redhat/rhn/frontend/xmlrpc/image/profile/test/ImageProfileHandlerTest.java
@@ -20,6 +20,7 @@ import static com.redhat.rhn.testing.ImageTestUtils.createImageStore;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
@@ -65,7 +66,7 @@ public class ImageProfileHandlerTest extends BaseHandlerTestCase {
     }
 
     @Test
-    public final void testListImageProfileTypes() {
+    void testListImageProfileTypes() {
         List<String> types = handler.listImageProfileTypes(admin);
         assertEquals(2, types.size(), "Wrong number of image profile types found.");
         assertTrue(types.stream().anyMatch(ImageProfile.TYPE_DOCKERFILE::equals));
@@ -73,7 +74,7 @@ public class ImageProfileHandlerTest extends BaseHandlerTestCase {
     }
 
     @Test
-    public final void testGetDetailsDockerfile() throws Exception {
+    void testGetDetailsDockerfile() throws Exception {
         ImageStore store = createImageStore("myregistry", admin);
         ActivationKey key = createActivationKey(admin);
         int result = handler.create(admin, "myprofile", ImageProfile.TYPE_DOCKERFILE,
@@ -98,7 +99,7 @@ public class ImageProfileHandlerTest extends BaseHandlerTestCase {
     }
 
     @Test
-    public final void testGetDetailsKiwi() throws Exception {
+    void testGetDetailsKiwi() throws Exception {
         ImageStore store = createImageStore("mystore", admin, ImageStoreFactory.TYPE_OS_IMAGE);
         ActivationKey key = createActivationKey(admin);
         int result = handler.create(admin, "myprofile", ImageProfile.TYPE_KIWI,
@@ -116,7 +117,7 @@ public class ImageProfileHandlerTest extends BaseHandlerTestCase {
     }
 
     @Test
-    public final void testListImageProfiles() throws Exception {
+    void testListImageProfiles() throws Exception {
         createImageStore("myregistry", admin, ImageStoreFactory.TYPE_REGISTRY);
         createImageStore("myosimagestore", admin, ImageStoreFactory.TYPE_OS_IMAGE);
 
@@ -148,7 +149,7 @@ public class ImageProfileHandlerTest extends BaseHandlerTestCase {
     }
 
     @Test
-    public final void testCreateImageProfile() throws Exception {
+    void testCreateImageProfile() throws Exception {
         ImageStore store = createImageStore("myregistry", admin);
         int result = handler.create(admin, "newprofile", ImageProfile.TYPE_DOCKERFILE,
                 "myregistry", "/path/to/dockerfile", "");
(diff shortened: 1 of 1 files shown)


java/core/src/main/java/com/redhat/rhn/frontend/action/systems/sdc/SystemOverviewAction.java
```diff
@@ -181,6 +180,15 @@ public ActionForward execute(ActionMapping mapping, ActionForm form,
...
     }
 
+    private static int getEntitlementComparator(Entitlement e1, Entitlement e2) {
+        if ((e1.isBase() && e2.isBase()) || (!e1.isBase() && !e2.isBase())) {
+            return e1.getHumanReadableLabel().compareTo(e2.getHumanReadableLabel());
+        }
+        else {
+            return (e1.isBase() ? -1 : 1);
+        }
+    }
```
mackdk, line 190: This can be simplified to:

```java
    private static int getEntitlementComparator(Entitlement e1, Entitlement e2) {
        if (e1.isBase() == e2.isBase()) {
            return e1.getHumanReadableLabel().compareTo(e2.getHumanReadableLabel());
        }

        return (e1.isBase() ? -1 : 1);
    }
```

Or you can even avoid defining an explicit compartor function:

```java
        List<Entitlement> entitlements = s.getEntitlements().stream()
            .sorted(Comparator.comparing(Entitlement::isBase).reversed()
                              .thenComparing(Entitlement::getHumanReadableLabel))
            .collect(Collectors.toList());
```

The `reversed()` part is needed because `false` sorts before `true`, but here we were returning -1 on `true` so it was sorted before.

java/core/src/main/java/com/redhat/rhn/manager/setup/ProductSyncManager.java
```diff
@@ -451,6 +450,15 @@ private SetupWizardProductDto convertProduct(final MgrSyncProductDto productIn)
...
     }
 
+    private static int sortBaseChannelOnTop(MgrSyncChannelDto a, MgrSyncChannelDto b) {
+        if (StringUtils.isBlank(a.getParentLabel())) {
+            return -1;
+        }
+        else {
+            return StringUtils.isBlank(b.getParentLabel()) ? 1 : 0;
+        }
+    }
```
mackdk: Not due to your change, but this comparator is kind of broken. If both `a` and `b` have blank parent label, `sortBaseChannelOnTop(a, b)`, `sortBaseChannelOnTop(b, a)` and `sortBaseChannelOnTop(a, a)`  all return `-1`, while the should always return `0`.  So, this code totally breaks the compare contract. I think this should simply be replaced by:

```java
        List<Channel> mandatoryChannelsOut = collect.get(true).stream()
            // Add base channel on top of everything else so it can be added first.
            // Base channels have empty parent, and java sorts false before true
            .sorted(Comparator.comparing(dto -> StringUtils.isNotBlank(dto.getParentLabel())))
            .map(mapping)
            .collect(Collectors.toList());
```

java/core/src/main/java/com/suse/manager/api/RouteFactory.java
```diff
@@ -334,6 +334,12 @@ private Gson initGsonWithSerializers() {
     private int compareSerializerHierarchy(ApiResponseSerializer<?> a, ApiResponseSerializer<?> b) {
         Class<?> aClass = a.getSupportedClass();
         Class<?> bClass = b.getSupportedClass();
-        return aClass.isAssignableFrom(bClass) ? -1 : (bClass.isAssignableFrom(aClass) ? 1 : 0);
+
+        if (aClass.isAssignableFrom(bClass)) {
+            return -1;
+        }
+        else {
+            return bClass.isAssignableFrom(aClass) ? 1 : 0;
```
mackdk, line 343: This comparator is broken as well, but honestly I'm not entirely sure about what logic this was trying to achieve, but this again return -1 when comparing the same class. Judging by the comment on line 319, I think this should be something like, but I'm not sure:

```java
private int compareSerializerHierarchy(ApiResponseSerializer<?> a, ApiResponseSerializer<?> b) {
    Class<?> aClass = a.getSupportedClass();
    Class<?> bClass = b.getSupportedClass();

    if (aClass.equals(bClass)) {
        return 0;
    }

    // Sort parent (more generic) classes first
    if (aClass.isAssignableFrom(bClass)) {
        return -1;
    }
    else if (bClass.isAssignableFrom(aClass)) {
        return 1;
    }

    // The two classes are unrelated, sort them alphabetically as siblings
    return aClass.getName().compareTo(bClass.getName());
}
``` 

Maybe the unit tests covers the logic, otherwise we can leave it as is with a comment :face_with_diagonal_mouth:

java/core/src/main/java/com/redhat/rhn/frontend/nav/NavTreeIndex.java
```diff
@@ -198,9 +198,7 @@ public String computeActiveNodes(String url, String lastActive) {
             String[] urls = new String[prefixes.length + 1];
 
             // Add the lastActive to the end
-            for (int i = 0; i < prefixes.length; i++) {
-                urls[i] = prefixes[i];
-            }
+            System.arraycopy(prefixes, 0, urls, 0, prefixes.length);
```
mackdk: This whole block inside the `if` could be simplified by using `Arrays.copyOf()` which copies and expands (or contract) the array at the same time:

```java
if (lastActive != null) {
    String[] urls = Arrays.copyOf(prefixes, prefixes.length + 1);
    urls[prefixes.length] = lastActive;
    prefixes = url;
}
```

Or even better we should throw away arrays and just use an `ArrayList` in this and the downstream methods :slightly_smiling_face:

java/core/src/test/java/com/redhat/rhn/taskomatic/core/test/TestDaemon.java
```diff
@@ -81,6 +81,7 @@ public void run() {
                 Thread.sleep(10000);
             }
             catch (InterruptedException e) {
+                //test should never be here
```
mackdk: Since this is unexpected, I guess resetting the interrupt flag in this case could be the best choice:

```suggestion
                Thread.currentThread().interrupt();
```

java/core/src/main/java/com/redhat/rhn/frontend/action/systems/sdc/SystemOverviewAction.java
CDellaGiusta: Fixed.

java/core/src/main/java/com/redhat/rhn/frontend/nav/NavTreeIndex.java
CDellaGiusta: I take the second suggestion and convert to ArrayList. Fixed

java/core/src/test/java/com/redhat/rhn/taskomatic/core/test/TestDaemon.java
CDellaGiusta: fair enough, fixed

java/core/src/main/java/com/redhat/rhn/manager/setup/ProductSyncManager.java
CDellaGiusta: fixed

java/core/src/main/java/com/suse/manager/api/RouteFactory.java
CDellaGiusta: added a TODO comment as agreed

mackdk (approved):

PR #11300, opened 2025-12-08
Revert "Use absolute in paths in salt scripts" (bsc#1253780)

m-czernek:
## What does this PR change?

In https://github.com/uyuni-project/uyuni/commit/c4f0d68b23ac166ce40f1c9bc9fae7c1a5c695d3, we modified the SLS to use absolute paths. This was based on sonarcloud issue - https://github.com/SUSE/spacewalk/issues/24387

The problem is that older Ubuntu and Debian systems (that were updated and not reinstalled, e.g. 16.04 -> 18.04 -> 20.04 -> 22.04 -> 24.04) do not necessarily have `usrmerge` installed. Usrmerge is installed by default only on new installations.

This means that we cannot assume `/usr/bin` path for a lot of binaries on some Debians and Ubuntus, such as `uname`, `cat`, `ls`, etc.

Given that this change is related to sonarcloud, I propose reverting the change. 

## Codespace

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): # 
- https://github.com/SUSE/spacewalk/issues/29075
- https://github.com/SUSE/spacewalk/issues/29295
Port(s): # TODO

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/bootloader/autoinstall.sls
+++ b/susemanager-utils/susemanager-sls/salt/bootloader/autoinstall.sls
@@ -11,10 +11,10 @@ mgr_copy_initrd:
 
 {% set loader_type = salt['cmd.run']('if [ -f /etc/sysconfig/bootloader ]; then source /etc/sysconfig/bootloader 2> /dev/null; fi;
 if [ -z "${LOADER_TYPE}" ]; then
-if [ $(/usr/bin/which grubonce 2> /dev/null) ] && [ !$(/usr/bin/which grub2-mkconfig 2> /dev/null) ]; then LOADER_TYPE="grub";
-elif [ $(/usr/bin/which elilo 2> /dev/null) ] && [ !$(/usr/bin/which grub2-mkconfig 2> /dev/null) ]; then LOADER_TYPE="elilo";
+if [ $(which grubonce 2> /dev/null) ] && [ !$(which grub2-mkconfig 2> /dev/null) ]; then LOADER_TYPE="grub";
+elif [ $(which elilo 2> /dev/null) ] && [ !$(which grub2-mkconfig 2> /dev/null) ]; then LOADER_TYPE="elilo";
 fi;
-fi; /usr/bin/echo "${LOADER_TYPE}"', python_shell=True) %}
+fi; echo "${LOADER_TYPE}"', python_shell=True) %}
 {% if loader_type == 'grub' %}
 mgr_create_grub_entry:
   file.append:
@@ -27,7 +27,7 @@ mgr_create_grub_entry:
 
 mgr_grub_boot_once:
   cmd.run:
-    - name: /usr/sbin/grubonce "{{ pillar.get('uyuni-reinstall-name') }}"
+    - name: grubonce "{{ pillar.get('uyuni-reinstall-name') }}"
     - onchanges:
       - file: mgr_create_grub_entry
 {% elif loader_type == 'elilo' %}
@@ -50,7 +50,7 @@ mgr_set_default_boot:
 
 mgr_elilo_copy_config:
   cmd.run:
-    - name: /sbin/elilo
+    - name: elilo
     - onchanges:
       - file: mgr_create_elilo_entry
       - file: mgr_set_default_boot
@@ -72,7 +72,7 @@ mgr_set_default_boot:
 
 mgr_generate_grubconf:
   cmd.run:
-    - name: /usr/sbin/grub2-mkconfig -o /boot/grub2/grub.cfg
+    - name: grub2-mkconfig -o /boot/grub2/grub.cfg
     - onchanges:
       - file: mgr_copy_kernel
       - file: mgr_copy_initrd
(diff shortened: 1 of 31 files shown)


rjmateus:
This also affects some older versions of SLES, like sles12: https://github.com/SUSE/spacewalk/issues/29295

rjmateus:
@m-czernek I don't think this revert is needed. Thomas already fixed part of the issues, and this commit should fix the remaining missing parts: https://github.com/uyuni-project/uyuni/pull/11355

What do you think?

m-czernek:
@rjmateus Sles 12 is unrelated to this issue. As in the description, certain set of Ubuntu/Debian systems, some binaries we assume to be in `/usr/bin` are in `/bin`. 

I see, for example, that we hardcode `/usr/bin/which` at [0], which would fail on an old Ubuntu system that does not explicitly install `usrmerge`. Note btw, that newly deployed Ubuntu systems would require `/usr/bin/which` even though they do not have `usrmerge` installed.

I see two options basically:
- We revert the original PR to use full binary paths (from a security perspective, I've discussed this with Orion & Wictor, see [1] for more info)
- For every binary, we try any possible path where it could be (non-user writable). When I was still in the bug squad, I created [2] as a POC, but this could suffer from problems for ssh-only management.

Would love to hear your opinion here; for now, I'm still leaning towards reverting the full paths (especially since we use `which`, which negates any security gained from using full binary paths anyways).

[0] https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/hardware/profileupdate.sls#L157
[1] https://github.com/SUSE/spacewalk/issues/29075#issuecomment-3646503605
[2] https://github.com/uyuni-project/uyuni/pull/11315

vzhestkov (approved):
I would vote for reverting it.

rjmateus:
@m-czernek Thank you for the detailed reply. For SLES is relatively simple to have the full path, since we can easily compute the path based on the OS.
For Debian is also possible, but it looks much more complex and unpredictable, we can look in a set of locations as you have on the pr: https://github.com/uyuni-project/uyuni/pull/11315

It doesn't look as a security issue, as discussed in the slack thread. So if the change in bringing more problems then the ones it fix, I think we should consider reverting it.
However, if I get it right, the autoinstall doesn't apply to Debian, and all issues should be fixed now. So we should be able to keep the changes to that file.
for the bootstrap is a bit more complex, and we can revert it and revisit in the future if a security issue arise. Another option could be have a conditional execution and you propose in the draft PR

agraul:
> @m-czernek Thank you for the detailed reply. For SLES is relatively simple to have the full path, since we can easily compute the path based on the OS. For Debian is also possible, but it looks much more complex and unpredictable, we can look in a set of locations as you have on the pr: #11315
> 

After looking into this a bit more, I tend to think that we don't need to revert the whole commit. There are not many places left that cause issues and it's good to be explicit where we can.

There are Salt states that should work on different OSes and where we don't know the location of all invoked commands. In this case, I propose to use `command -p $cmd`. It's a [built-in](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/command.html) command and uses a [PATH](https://gitweb.git.savannah.gnu.org/gitweb/?p=bash.git;a=blob;f=config-top.h#l68) similar to the `search_paths` list defined in #11315. From a security perspective it's better than running `which` (which uses a user-defined `PATH`) and we know that it's always available.

agraul:
To bring the discussion to a single place, this is what @m-czernek wrote in #11315 as a reply to my comment to keep using the full path to snapper.

> I'm not a fan of this approach. Yes, we know where snapper is. The point of the security PR was to always use full binary paths. The point of this PR [#11315] was to always use utils.find_binary. If we implement your command -p approach, I'm not a fan of having implicit reasons for where to use it and where not.

> This kind of "use a security principle only when it would have been insecure otherwise" leads to inconsistencies in the future, when someone sees that snapper uses a full path instead of e.g. command -p, and will use full path for other binaries when modifying/creating a SLS.

Marek raises a good point, using the same approach everywhere will make it easier to maintain the SLS files in the future. We only touch these files from time to time and I share the concern that someone will pick the wrong implementation (full path where `command -p` would have been appropriate) in the future. Once we have reached a conclusion I'll put it in a README file in `susemanager-sls` for future reference, but documenting alone won't be enough. We all know how easy it is too miss the documentation, especially when you already know how to do something.

Is everyone fine with **always** using `command -p <foo>`? That will protect against shell functions, shell aliases and PATH modifications (in Bash `command -p` sets `PATH` to `"/bin:/usr/bin:/sbin:/usr/sbin"`).

rjmateus:
I prefer to have the same approach across all calls. We need to keep consistency as much as possible to make maintenance a bit easier. 
From my point of view we can move forward with the approach @agraul sugested with the `command -p <foo>`.

agraul:
See https://github.com/uyuni-project/uyuni/pull/11380 for an implementation of `command -p` (and `command -v` where we currently use `which` for consistency)

m-czernek:
Closing in favor of https://github.com/uyuni-project/uyuni/pull/11380

PR #11304, opened 2025-12-11
Fix reposync mediaproduct fetch when URL contains auth token (bsc#1252388)

m-czernek:
## What does this PR change?

`yum_src.py` automatically adds a slash `/` at the end of every URL. But, when fetching media/products file from reposync, if we're authenticating against a MLM, `/` is added at the end of the auth token, which makes the token malformed and fails MLM auth process.

## Codespace

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): # https://github.com/SUSE/spacewalk/issues/28715
Port(s): # TODO

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
+++ b/python/spacewalk/satellite_tools/repo_plugins/yum_src.py
@@ -1437,7 +1437,14 @@ password={passwd}
         url = "media.1/products"
         media_products_path = os.path.join(self._get_repodata_path(), url)
         grabber = urlgrabber.grabber.URLGrabber()
-        mirror_group = MirrorGroup(grabber, self.repo.urls)
+        urls = []
+        for url in self.repo.urls:
+            # A URL might contain an auth token
+            if url.endswith("/"):
+                url = url[:-1]
+            urls.append(url)
+        mirror_group = MirrorGroup(grabber, urls)
+
         try:
             urlgrabber_opts = {}
             self.set_download_parameters(urlgrabber_opts, url, media_products_path)
--- a/python/spacewalk/spacewalk-backend.changes.mczernek.reposync-product-fix
+++ b/python/spacewalk/spacewalk-backend.changes.mczernek.reposync-product-fix
@@ -1,0 +1,2 @@
+- Fix reposync mediaproduct fetch when
+  URL contains auth token (bsc#1252388)
--- a/python/test/unit/spacewalk/satellite_tools/test_yum_src.py
+++ b/python/test/unit/spacewalk/satellite_tools/test_yum_src.py
@@ -534,6 +534,19 @@ class YumSrcTest(unittest.TestCase):
             assert not cs.get_mediaproducts()
             log_mock.assert_not_called()
 
+    def test_get_mediaproduct_adds_no_trailing_backslash(self):
+        cs = self._make_dummy_cs()
+        mirror_mock = Mock()
+
+        with patch(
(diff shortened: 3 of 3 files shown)


python/spacewalk/satellite_tools/repo_plugins/yum_src.py
```diff
@@ -1437,7 +1437,14 @@ def get_mediaproducts(self):
         url = "media.1/products"
         media_products_path = os.path.join(self._get_repodata_path(), url)
         grabber = urlgrabber.grabber.URLGrabber()
-        mirror_group = MirrorGroup(grabber, self.repo.urls)
+        urls = []
+        for url in self.repo.urls:
+            # A URL might contain an auth token
```
agraul, line 1442: Why do we change this here in `get_mediaproducts()`? Are other methods not affected the same? Can we move URL cleanup to the place where we first populate `self.repo.urls`?

m-czernek: We explicitly add a trailing backslash at [0]. The other methods delegate their requests to zypper, which actually removes the backslash from the requests to the other master (in the context of ISSv3 sync).

`get_mediaproducts()` is the only method that actually hand-crafts the requests and sends it to the master, which is why it was the only call within ISSv3 that failed. IIRC, there were 4 previous calls that passed the token verification, and the media product call was the 5th one, which failed because it requested `$URL/$TOKEN/` instead of `$URL/$TOKEN`.

Given that zypper actually removes the backslash from the URL, I considered it reasonable that `get_mediaproducts` conforms to the same behavior, and modified the behavior this low in the call stack.

Regarding [0], we do the same thing in all the repo plugins, which is why I was hesitant to remove the slash at the end:

```
repo_plugins/uln_src.py:        # Make sure baseurl ends with / and urljoin will work correctly
repo_plugins/yum_dnf_src.py:            # Make sure baseurl ends with / and urljoin will work correctly
repo_plugins/deb_src.py:        # Make sure baseurl ends with / and urljoin will work correctly
repo_plugins/yum_src.py:        # Make sure baseurl ends with / and urljoin will work correctly
repo_plugins/yum_src.py:        # Make sure baseurl ends with / and urljoin will work correctly
```

When I look at `urljoin` only, we use it only in [1] and based on [2], it seems to me like `urljoin` works correctly without the trailing slash. I'm not confident though that there's no other string concatenation that relies on the trailing slash.

[0] https://github.com/uyuni-project/uyuni/blob/master/python/spacewalk/satellite_tools/repo_plugins/yum_src.py#L321-L323
[1] https://github.com/uyuni-project/uyuni/blob/master/python/spacewalk/satellite_tools/download.py#L281-L292
[2] https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

agraul: We have the "ensure trailing slash" in reposync plugins since https://github.com/uyuni-project/uyuni/commit/978ae262a4188db12fcf2867c78b6f1b55d091cb

Looking at the example you linked from the docs and trying it out, there does not seem to be a reason to do it like that anymore. `urljoin` without a trailing slash is no problem. I understand that it's hard to verify that removing that code does not cause any other issue, our repo plugins grew into hard to understand beasts.

PR #11315, opened 2025-12-12
Do not hardcode binary paths in sls

m-czernek:
## What does this PR change?

This is an alternative approach to https://github.com/uyuni-project/uyuni/pull/11300

In https://github.com/uyuni-project/uyuni/pull/11300, we rely on PATH resolution. This has potential security implications. In this PR, we take a reduced approach, where we allow binaries to exist outside of `/usr/bin`, but only in predefined, root-only writable locations.

This has the advantage of relaxing the restrictions imposed by enforcing `/usr/bin` paths, but also being secure.

NOTE: This has a potential for SLS rendering performance issues in case of Salt SSH deployments. Testing is needed in such situations.

## Codespace

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
@@ -1,3 +1,4 @@
+{% import 'util/lib.sls' as utils %}
 ##
 ##  java bootstrapping calls certs.sls before this state
 ##
@@ -165,7 +166,7 @@ bootstrap_repo:
 {% set salt_minion_installed = (salt['pkg.info_installed']('venv-salt-minion', attr='version', failhard=False).get('venv-salt-minion', {}).get('version') != None) %}
 check_bootstrap_dbg:
   cmd.run:
-    - name: /usr/bin/echo "{{ salt_minion_installed }}"
+    - name: {{ utils.find_binary('echo') }} "{{ salt_minion_installed }}"
 {% set venv_available_request = salt_minion_installed or salt['http.query'](bootstrap_repo_url + 'venv-enabled-' + grains['osarch'] + '.txt', status=True, verify_ssl=False) %}
 {# Prefer venv-salt-minion if available and not disabled #}
 {%- set use_venv_salt = salt['pillar.get']('mgr_force_venv_salt_minion') or ((salt_minion_installed or (0 < venv_available_request.get('status', 404) < 300)) and not salt['pillar.get']('mgr_avoid_venv_salt_minion')) %}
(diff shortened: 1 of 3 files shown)


m-czernek:
Thank you @agraul for taking a look, much appreciated. 

The upside would be security related - the paths that we search instead of `PATH` are not user writable. This means that an attacker that gains non-sudo access to a minion cannot do path hacking to elevate their privileges to superuser. 

Of course, we're disregarding the fact that we already use `which` in the sls files, which negates the security impact, but that can be solved easily since this would make `which` unnecessary.

The downsides are twofold:

- One is practical, i.e. every we'd have to modify every binary to use the util class to find it. In addition, that brings the possibility of making subtle changes to the final SLS renders if one makes a typo, which could result in some L3s. This is annoying, but doable and verifiable. 
- One is technical, i.e. as Victor mentioned, when we manage minions via SSH, the compilation of the salt state is done on the master. I believe this should work regardless, but we'd have to verify this to be sure of course. Additionally, we'd have to verify if this approach doesn't produce performance impact (e.g. establishing SSH to the minion to find every binary would likely be significant, performance speaking).

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -165,7 +166,7 @@ bootstrap_repo:
 {% set salt_minion_installed = (salt['pkg.info_installed']('venv-salt-minion', attr='version', failhard=False).get('venv-salt-minion', {}).get('version') != None) %}
 check_bootstrap_dbg:
   cmd.run:
-    - name: /usr/bin/echo "{{ salt_minion_installed }}"
+    - name: {{ utils.find_binary('echo') }} "{{ salt_minion_installed }}"
```
agraul, line 169: `echo` is a shell built-in in `bash`, `zsh`, `fish`, `dash` and probably many others. I would just use it plainly. The shell will find `/usr/bin/echo` if needed anyway.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -189,10 +190,10 @@ salt-minion-package:
 {# hack until transactional_update.run is fixed to use venv-salt-call #}
 {# Writing  to the future - find latest etc overlay which was created for package installation and use that as etc root #}
 {# this only works here in bootstrap when we are not running in transaction #}
-{%- set pending_transaction_id = salt['cmd.run']('/usr/bin/snapper --no-dbus list --columns=number | /usr/bin/grep "+" | tr -d "+"', python_shell=True) %}
+{%- set pending_transaction_id = salt['cmd.run'](utils.find_binary('snapper') ~ ' --no-dbus list --columns=number | ' ~ utils.find_binary('grep') ~ ' "+" | tr -d "+"', python_shell=True) %}
```
agraul, line 193: We know where snapper is on transactional systems.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
```diff
@@ -335,7 +336,7 @@ copy_transactional_conf_file_to_etc:
     - name: /etc/transactional-update.conf
     - source: /usr/etc/transactional-update.conf
     - unless:
-      - /usr/bin/test -f /etc/transactional-update.conf
+      - {{ utils.find_binary('test') }} -f /etc/transactional-update.conf
```
agraul, line 339: `test` is defined as a shell built-in in POSIX

susemanager-utils/susemanager-sls/salt/util/lib.sls
```diff
@@ -0,0 +1,36 @@
+{%- set _find_binary_cache = {} -%}
+
+{%- macro find_binary(binary_name) -%}
```
agraul, line 3: This is very similar to running `command -p $cmd`. Shouldn't we use that instead? It's a [built-in](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/command.html) command and uses a [PATH](https://gitweb.git.savannah.gnu.org/gitweb/?p=bash.git;a=blob;f=config-top.h#l68) similar to the `search_paths` list you've defined below.

susemanager-utils/susemanager-sls/salt/bootstrap/init.sls
m-czernek: I'm not a fan of this approach. Yes, we know where snapper is. The point of the security PR was to always use full binary paths. The point of this PR was to always use `utils.find_binary`. If we implement your `command -p` approach, I'm not a fan of having implicit reasons for where to use it and where not.

This kind of "use a security principle only when it would have been insecure otherwise" leads to inconsistencies in the future, when someone sees that snapper uses a full path instead of e.g. `command -p`, and will use full path for other binaries when modifying/creating a SLS.

agraul:
Closing in favor of https://github.com/uyuni-project/uyuni/pull/11380

PR #11323, opened 2025-12-17
Maven migration next steps: Move all unit tests outside of “test” subpackage

CDellaGiusta:
## What does this PR change?
This PR aims to move all unit tests outside of “test” subpackages.

Please note: the very last commit is temporary while developing and rebasing, and will be fixed as permanent only when everything is ready to go, by running the script  `/java/scripts/migrate-tests/run_main.sh`

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/29235
Port(s): not backported
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/apitest/TestHandler.java
mackdk, line 1: This is actually part of our API, even if not documented. @mcalmer can we remove this `TestHandler` or does it have any usage?

java/core/src/main/java/org/cobbler/CobblerObject.java
```diff
@@ -571,9 +571,9 @@ public void setResolvedOwners(List<String> ownersIn) {
      * @return the created
      */
     public Date getCreated() {
-        Double time = (Double) dataMap.get(CTIME);
+        Long time = (Long) dataMap.get(CTIME);
```
mackdk: This change looks kind of dangerous: at runtime if the type of the object inside the map is actually `Double` this will throw `ClassCastException` and it seems to me that the data is coming from an external source, but I'm not familiar enough with this code.

 Why do we need this change?

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/apitest/TestHandler.java
mcalmer: It is from old spacewalk. The changes we did were more for syntax or new java libs.
So I think this can be removed if no test is using it.

java/core/src/main/java/org/cobbler/CobblerObject.java
mcalmer: Guessing: sometimes "time" (epoch) is send as a "double" but stored as "long". Maybe this was the try for an easy "double" to "long" conversion?

CDellaGiusta: Both DistroTest.testModified and DistroTest.testCreated have been fixed (only the tests) in https://github.com/uyuni-project/uyuni/pull/11342, while changes on CobblerObject have been deleted

PR #11326, opened 2025-12-17
Improve switching to salt bundle

vzhestkov:
## What does this PR change?

There was a list of logical traps in the original state causing impossibility to switch to the bundle or even try to start the salt bundle with wrong keys if there was another unsucessful attempt to switch to the bundle before, but then the salt keys for classic salt-minion were changed. In such case old keys stored in the bundle will be reused instead of copying the actual ones.

It also improves the speed of switching the services to disable classic salt-minion and start further processing with the salt bundle. It uses scheduled task inside the bundle to stop classic salt-minion. It's a bit more complex solution, but it allows to avoid getting stuck on attempt to stop `salt-minion` service when the state applied with `salt-minion` itself, in most cases it took ~90 seconds, with this change it should be ~5 seconds.

## Codespace

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: already covered

## Links

Tracks: https://github.com/SUSE/spacewalk/issues/28884

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/susemanager-sls.changes.vzhestkov.simplify-salt-bundle-switch
+++ b/susemanager-utils/susemanager-sls/susemanager-sls.changes.vzhestkov.simplify-salt-bundle-switch
@@ -1,0 +1,2 @@
+- Avoid possible issues in salt state switching to the salt bundle.
+  Improve the speed of applying the state.


susemanager-utils/susemanager-sls/susemanager-sls.changes.vzhestkov.simplify-salt-bundle-switch
```diff
@@ -0,0 +1,2 @@
+- Avoid possible issues in salt state switching to the salt bundle.
```
m-czernek, line 1: Maybe we could be more specific here?

```suggestion
- Avoid possible PKI and config issues when migrating to Salt Bundle.
```

m-czernek (approved):
+1, took me a while to orient myself in the changes, but it makes sense to me.

Leaving a small suggestion for the chlog.

You might want to squash the commits before merging - I'll leave that up to you though.

PR #11364, opened 2026-01-08
testsuite: enable to run the secondary https_connection.feature

alichiel:
## What does this PR change?
It enables to run basic https_connection tests.
The test feature was executed successfully several times on all MLM versions. The feature itself is already merged but has been never added to runs. This PR fixes that.

### DIFF
--- a/testsuite/run_sets/secondary.yml
+++ b/testsuite/run_sets/secondary.yml
@@ -7,6 +7,7 @@
 ## Secondary features BEGIN ##
 
 # IDEMPOTENT
+- features/secondary/https_connection.feature
 - features/secondary/srv_mgr_sync_list_products.feature
 - features/secondary/srv_monitoring.feature
 - features/secondary/srv_virtual_host_manager.feature


testsuite/run_sets/secondary.yml
```diff
@@ -7,6 +7,7 @@
 ## Secondary features BEGIN ##
 
 # IDEMPOTENT
+- features/secondary/https_connection.feature
```
srbarrios, line 10: I don't see any reason to don't run it on secondary_parallelizable.yml, instead of here, sequentially.

alichiel: It could, of course, but the reason I put it there is the https should be ano of the sanity tests (first running) tests and if this doesn't work, this is a big problem.

srbarrios: Well, then, let's have this feature much earlier, on core tests?

alichiel:
After the discussion with Oscar, let's keep it in Secondary.
If this needs to be moved to Core, we'll create another PR.

PR #11408, opened 2026-01-22
Reorganise the server helm values by components

cbosdo:
## What does this PR change?

Getting all the images in one big value is not really intuitive. Use a top level object for each of the components with image and tag plus potential other variables.

Also took that opportunity to add a value to disable the deployment of the database, but commented that this is a dangerous option to play with.

The DB host, port and names are also converted to values in the ConfigMap to change them more easily later.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: will document the kubernetes install once all the bits are more stable.

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28944

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/podman_runner/03_run_controller_and_registry_and_buildhost.sh
+++ b/testsuite/podman_runner/03_run_controller_and_registry_and_buildhost.sh
@@ -6,7 +6,7 @@ echo buildhostproductuuid > /tmp/buildhost_product_uuid
 
 AUTH_REGISTRY_USER=$(echo "$AUTH_REGISTRY_CREDENTIALS"| cut -d\| -f1)
 AUTH_REGISTRY_PASSWD=$(echo "$AUTH_REGISTRY_CREDENTIALS" | cut -d\| -f2)
-sudo -i podman run --rm -d --network network -v /tmp/testing:/tmp --name controller -h controller -v ${src_dir}/testsuite:/testsuite ghcr.io/$UYUNI_PROJECT/uyuni/ci-test-controller-dev:$UYUNI_VERSION
+sudo -i podman run --pull newer --rm -d --network network -v /tmp/testing:/tmp --name controller -h controller -v ${src_dir}/testsuite:/testsuite ghcr.io/$UYUNI_PROJECT/uyuni/ci-test-controller-dev:$UYUNI_VERSION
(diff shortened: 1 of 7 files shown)


microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/configuration/DefaultConfiguration.java
```diff
@@ -69,6 +82,21 @@ public String getDatabaseConnectionString() {
...
+
+    @Override
+    public int getDatabasePort() {
+        return configurationSource.requireInteger("database_port");
+    }
+
+    @Override
+    public String getDatabaseNameString() {
+        return configurationSource.requireString("database_name");
+    }
```
mackdk, line 101: `ConfigurationSource.requireString()` will throw an exception if the property is not present. I guess you now need to call `configurationSource.getString("...").orElse(...)` for all these properties and for `getDatabaseConnectionString()` as well.

microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/configuration/DefaultConfiguration.java
```diff
@@ -91,6 +119,16 @@ public int getBatchSize() {
...
+
+        String host = getDatabaseHostString();
+        int port = getDatabasePort();
+        String name = getDatabaseNameString();
+
+        if (!host.isEmpty() && port != 0 && !name.isEmpty()) {
+            String cnx = String.format("jdbc:postgresql://%s:%s/%s", host, port, name);
+            properties.setProperty("database_connection", cnx);
+        }
+        return properties;
```
mackdk, line 135: Would be great to add a unit for this method, since the logic now deviates from the base class and it's pretty important to ensure we get the connection property correct.

microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/configuration/DefaultConfiguration.java
mackdk: Or you can also add the defaults to `configuration-defaults.properties` so that those properties are never missing.

PR #11216, opened 2025-11-20
QE: Partially synchronize products in our test suites

srbarrios:
## What does this PR change?

This pull request adds new test scenarios for partial product synchronization in the Setup Wizard and introduces improvements for faster test execution. The main changes include new feature scenarios for partially synchronizing SLES 15 SP4, SLES 15 SP7, and SL Micro 6.1 products, updates to required package lists for these products, and enhancements to the test environment to support a faster test mode.

### New partial synchronization scenarios (testsuite/features/reposync/srv_sync_products.feature):

* Added scenarios to partially synchronize SLES 15 SP4, SLES 15 SP7, and SL Micro 6.1 products, focusing on recommended sub-products and client tools dependencies. These scenarios are tagged with `@fast` to allow for quicker execution.

### Package list updates (testsuite/features/support/constants.rb):

* Updated the required package lists for the new partial synchronization scenarios, specifying essential packages for SLES 15 SP4, SLES 15 SP7, and SL Micro 6.1 pool channels.

### Fast test execution support (testsuite/features/support/env.rb):

* Introduced the `$faster` environment variable to enable fast mode for tests.
* Added a `Before('@fast')` hook to skip scenarios tagged with `@fast` unless the `fast` environment variable is set, improving test suite flexibility and speed.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
ℹ️ If a major new functionality is added, it is **strongly recommended** that tests for the new functionality are added to the Cucumber test suite
- Cucumber tests were refactored

- [x] **DONE**

## Links

Ports:
- Manager-5.0
- Manager-5.1

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
[...]

PR #11414, opened 2026-01-23
Add --metadata-only support to mgr-sync

srbarrios:
## What does this PR change?

This PR introduces the ability to perform a metadata-only repository synchronization when adding products or channels via the `mgr-sync` CLI. This is particularly useful for environments where we want to bootstrap and onboard systems without waiting to fully download all RPMs.

This feature is intended to work in conjunction with partial synchronizations where we filter only the required packages to prepare a bootstrap repository. For that purpose, one would typically use a command like: `spacewalk-repo-sync -c <channel_name> --include dmidecode --include libunwind ...`.

### Difference between `--no-sync` and `--metadata-only`

**`--no-sync`**

* **Function**: Prevents any repository synchronization from being scheduled automatically after the channel is added.
* **Behavior**: The channel is added to your organization, but it remains empty. No background task is sent to Taskomatic. The channel status remains "Not Installed" (or shows "Sync Failed" in the WebUI if previously attempted), and it does not generate valid repository metadata usable for onboarding systems.

**`--metadata-only`**

* **Function**: Schedules a synchronization immediately but instructs the backend to download **only the metadata**, skipping the actual RPM package binaries.
* **Behavior**: The channel is added, and a sync task is scheduled with the `metadata_only=True` parameter. The server imports all package information into the database (allowing for dependency resolution, searching, and a "Green/Installed" status in the WebUI), but the package files are not stored on the disk. This results in valid repository metadata that can be used to bootstrap/onboard a system.

#### 1. Backend: Taskomatic Support (`RepoSyncTask.java`)

Modified `RepoSyncTask` to recognize a new ad-hoc parameter `metadata_only` from the Job Data Map.

* If `metadata_only` is true, Taskomatic appends the `--metadata-only` flag when invoking the underlying `spacewalk-repo-sync` Python script.
* Retained legacy support for `no_packages`.

#### 2. Backend Script: Argument Parsing (`spacewalk-repo-sync`)

Updated the executable script `/usr/bin/spacewalk-repo-sync` to parse the new `--metadata-only` CLI argument and pass it correctly to the `RepoSync` class constructor.

#### 3. CLI: New Argument (`cli.py`)

Added a new global argument to the `add` subparser:

* `--metadata-only`: A boolean flag that prevents the automatic download of RPM packages while still syncing metadata and populating the database.

#### 4. Logic: Taskomatic Integration (`mgr_sync.py`)

Since the standard XML-RPC API `channel.software.syncRepo` does not support custom flags like `metadata_only`, the synchronization logic was updated to:

* Bypass the standard API when `--metadata-only` is toggled.
* Communicate directly with the **Taskomatic XML-RPC API** (`tasko.scheduleSingleBunchRun`).
* **Robust Channel ID Lookup**: Tries to resolve channel labels to IDs using the local cache first. If the channel was just added and is missing from the cache, it performs a direct API lookup (`channel.software.getDetails`) to ensure the task can be scheduled immediately without errors.

#### 5. Testing (`test_channel_operations.py`)

Added a new test case `test_add_channel_metadata_only` to verify:

* Correct parsing of the new CLI flag.
* Proper resolution of channel IDs.
* Verification that the internal Taskomatic API is called with the expected payload (`metadata_only: True`) while the standard `syncRepo` call is suppressed.

### How to Test

1. Run `mgr-sync add channel <label> --metadata-only`.
2. Verify in `rhn_taskomatic_daemon.log` that `spacewalk-repo-sync` is executed with the `--metadata-only` argument.
3. Check the **WebUI Product Wizard**: The channel should show as **Installed** (Green) because the package data has been imported into the database (unlike `--no-packages` which leaves it empty/red).
4. Check the filesystem: Verify that RPMs were not downloaded to `/var/spacewalk/rhn/kickstart/...` or the corresponding repository cache.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage

- Unit tests were added

- [x] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/satellite_tools/spacewalk-repo-sync
+++ b/python/spacewalk/satellite_tools/spacewalk-repo-sync
@@ -211,6 +211,14 @@ def main():
     )
     parser.add_option(
         "",
+        "--metadata-only",
+        action="store_true",
+        dest="metadata_only",
+        default=False,
+        help="Do not download packages, only sync metadata",
+    )
+    parser.add_option(
+        "",
         "--sync-kickstart",
         action="store_true",
         dest="sync_kickstart",
@@ -277,7 +285,7 @@ def main():
     # pylint: disable-next=consider-using-f-string
     log2disk(0, "Command: %s" % str(sys.argv))
 
-    l_params = ["no_errata", "sync_kickstart", "fail", "no-strict"]
+    l_params = ["no_errata", "sync_kickstart", "fail", "no-strict", "metadata_only"]
     d_chan_repo = reposync.getChannelRepo()
     l_ch_custom = reposync.getCustomChannels()
     d_parent_child = reposync.getParentsChilds()
@@ -407,7 +415,7 @@ def main():
             # pylint: disable-next=consider-using-f-string
             "Please check 'reposync/%s.log' for sync log of this channel." % ch,
             notimeYN=True,
-        )
+            )
         sync = reposync.RepoSync(
             channel_label=ch,
             repo_type=options.repo_type,
@@ -424,6 +432,7 @@ def main():
             log_level=log_level,
             force_all_errata=options.force_all_errata,
             show_packages_only=options.show_packages,
+            metadata_only=options.metadata_only,
         )
         if options.batch_size:
             sync.set_import_batch_size(options.batch_size)
--- a/susemanager/src/mgr_sync/cli.py
+++ b/susemanager/src/mgr_sync/cli.py
@@ -130,6 +130,13 @@ def _create_add_subparser(subparsers):
         default=False,
         help="do not syncronize product channels automatically after adding",
     )
+    add_parser.add_argument(
+        "--metadata-only",
+        action="store_true",
+        dest="metadata_only",
+        default=False,
+        help="do not import any package",
+    )
 
 
 def _create_list_subparser(subparsers):
--- a/susemanager/src/mgr_sync/mgr_sync.py
+++ b/susemanager/src/mgr_sync/mgr_sync.py
@@ -173,6 +173,7 @@ class MgrSync(object):  # pylint: disable=too-few-public-methods
                     mirror=options.mirror,
                     no_optionals=options.no_optionals,
                     no_sync=options.no_sync,
+                    metadata_only=options.metadata_only,
                 )
             elif "credentials" in options.add_target:
                 self._add_credentials(options.primary, options.target)
@@ -310,7 +311,14 @@ class MgrSync(object):  # pylint: disable=too-few-public-methods
             self.log,
         )
 
-    def _add_channels(self, channels, mirror="", no_optionals=False, no_sync=False):
+    def _add_channels(
+        self,
+        channels,
+        mirror="",
+        no_optionals=False,
+        no_sync=False,
+        metadata_only=False,
+    ):
         """Add a list of channels.
 
         If the channel list is empty the interactive mode is started.
@@ -423,9 +431,15 @@ class MgrSync(object):  # pylint: disable=too-few-public-methods
                 channels_to_sync.append(channel)
 
         if not no_sync:
-            self._schedule_channel_reposync(channels_to_sync)
+            self._schedule_channel_reposync(
+                channels_to_sync,
+                current_channels=current_channels,
+                metadata_only=metadata_only,
+            )
 
-    def _schedule_channel_reposync(self, channels):
+    def _schedule_channel_reposync(
+        self, channels, current_channels=None, metadata_only=False
+    ):
         """Schedules a reposync for a set of channels.
 
         :param channels: the labels identifying the channels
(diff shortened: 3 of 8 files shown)


agraul:
Can you explain the difference between `--no-sync` and `--no-sync-packages`? Does one imply the other? When should a user use one or the other?

srbarrios:
> Can you explain the difference between `--no-sync` and `--no-sync-packages`? Does one imply the other? When should a user use one or the other?

@agraul Sure! I edited the PR description.

agraul:
> > Can you explain the difference between `--no-sync` and `--no-sync-packages`? Does one imply the other? When should a user use one or the other?
> 
> @agraul Sure! I edited the PR description.

I need to ask again, when do you think a user should use `--no-sync` and when `--no-sync-packages`? Maybe it would be better to turn `--no-sync` into the behavior your AI is adding

susemanager/src/mgr_sync/cli.py
```diff
@@ -130,6 +130,13 @@ def _create_add_subparser(subparsers):
         default=False,
         help="do not syncronize product channels automatically after adding",
     )
+    add_parser.add_argument(
+        "--no-sync-packages",
+        action="store_true",
+        dest="no_sync_packages",
+        default=False,
+        help="do not sync packages, only metadata",
```
rjmateus: This is not complitelly correct, and can be mistaken with the work planned for lazy reposync. The data would be synced, but not packages are imported, meaning the channel on MLM/uyuni will not have any packages assigned.

srbarrios:
> > > Can you explain the difference between `--no-sync` and `--no-sync-packages`? Does one imply the other? When should a user use one or the other?
> > 
> > 
> > @agraul Sure! I edited the PR description.
> 
> I need to ask again, when do you think a user should use `--no-sync` and when `--no-sync-packages`? Maybe it would be better to turn `--no-sync` into the behavior your AI is adding

I'm not convinced of modifying the current behavior of --no-sync because I'm not sure how it is actually used and on which scenarios we prefer to don't even trigger a reposync.

And I want to use `--no-sync-packages` in particular to have valid repositories without waiting to have all the packages of all the required channels synchronized in order to speed up our CI Tests.

To give you some idea, in BV test suite we take a day to have all the supported OS synchronized, before we can proceed testing the client tools bootstrapping, onboarding and running smoke tests on the minions.

susemanager/src/mgr_sync/cli.py
srbarrios: Exactly, that's the behavior I expect, and it's enough to speed up our bootstraps.
How you would rephrased? And how you would design it so we can achieve both goals?
Faster bootstraps not requiring all channels synchronized (meaning all packages downloaded) + Lazy Reposync (which I'm not sure what we expect).

rjmateus: I would keep it simple and just say something like "do not import any package", and would not mention anything related to metadata

m-czernek (commented):
I would suggest that `--no-sync-packages` is a confusing name, considering we have `--no-sync`.

We should focus on what it does, not what it does not do. I'd suggest something like `--metadata-only` or `--package-metadata-only`. 

If we do want to merge this, also don't forget to document this, especially why this feature should be used. Otherwise we'll get L3 questions from customers about this.

Finally, can you explain the the use case to me please? I understand that with full reposync, you had to wait until reposync completes, which takes a long time. But what can you do exactly with reposyncing only metadata? Is the intended use case of this feature metadata only sync, then manually upload only the packages you need, and proceed with bootstrapping?

srbarrios:
> Finally, can you explain the the use case to me please? I understand that with full reposync, you had to wait until reposync completes, which takes a long time. But what can you do exactly with reposyncing only metadata? Is the intended use case of this feature metadata only sync, then manually upload only the packages you need, and proceed with bootstrapping?

Yes, quite close.
To generate the bootstrap repository, I already solved without this feature, as we only need to use spacewalk-repo-sync with the option --include to pass the list of packages we require on the bootstrap repo, and in that case, the repositories are automatically valid. And I have the bootstrap repo generated.
Once we have that, to bootstrap a minion through a proxy, we will need a list of channels to synched or have a valid repository at least, this list is longer than just the channels required for the bootstrap repo generation, and in such case, some channels I just want to have valid empty repos, so I can onboard the minions. In theory, if I need, I can always make a full synchronization later on.

Here is the PoC https://github.com/uyuni-project/uyuni/pull/11216

Maybe the most important part is this part:
```
When(/^I use spacewalk-repo-sync to sync all channels for "([^"]*)" including only client tools dependencies$/) do |product_os_version|
  client_tools_dependencies = CLIENT_TOOLS_DEPENDENCIES_BY_PRODUCT[product_os_version]
  package_for_testing = PACKAGE_BY_CLIENT[CLIENT_BY_OS_PRODUCT_VERSION[product_os_version]]
  append_includes = (client_tools_dependencies + [package_for_testing]).compact.map { |pkg| "--include #{pkg}" }.join(' ')
  channels_to_synchronize = CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, product_os_version).clone ||
                            CHANNEL_TO_SYNC_BY_OS_PRODUCT_VERSION.dig(product, "#{product_os_version}-x86_64").clone
  channels_to_synchronize = filter_channels(channels_to_synchronize, ['beta']) unless $beta_enabled
  channels_to_synchronize.each do |channel|
    $command_output, _code = get_target('server').run("mgr-sync add channel #{channel} --no-sync", check_errors: false, verbose: true)
    $command_output, _code = get_target('server').run("spacewalk-repo-sync -c #{channel} #{append_includes}", check_errors: false, verbose: true)
  end
end
```

And I plan to change the line `$command_output, _code = get_target('server').run("mgr-sync add channel #{channel} --no-sync", check_errors: false, verbose: true)` with this new option. And yeah, I agree that better rename it.

susemanager/src/mgr_sync/mgr_sync.py
```diff
@@ -441,9 +455,37 @@ def _schedule_channel_reposync(self, channels):
...
+                if current_channels is None:
+                    current_channels = self._fetch_remote_channels()
+
+                for label in channels:
+                    if label in current_channels:
+                        channel_ids.append(str(current_channels[label].id))
+                    else:
+                        # Fallback if channel wasn't in the provided list
+                        # (unlikely if flow is correct, but safe to ignore or log)
+                        self.log.warn(f"Could not resolve ID for channel {label}")
```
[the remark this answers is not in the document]
srbarrios: When I change it to .format() I receive lint issues.

susemanager/src/test/test_mgr_sync/test_channel_operations.py
```diff
@@ -860,6 +860,83 @@ def test_sync_channels_interactive_with_children(self):
...
             stubbed_xmlrpm_call.assert_has_calls(expected_xmlrpc_calls)
 
+    def test_add_channel_no_sync_packages(self):
+        """Test adding a channel with --no-sync-packages option"""
+        channel_label = "rhel-i386-as-4"
+        channel_id = 101
+        org_id = 123
+
+        # pylint: disable-next=consider-using-f-string
+        options = get_options(f"add channel {channel_label} --no-sync-packages".split())
```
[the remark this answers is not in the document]
srbarrios: I will not, otherwise the lint fails.

python/spacewalk/satellite_tools/spacewalk-repo-sync
```diff
@@ -209,6 +209,14 @@ def main():
         default=False,
         help="Do not sync packages",
     )
+    parser.add_option(
+        "",
+        "--metadata-only",
+        action="store_true",
+        dest="metadata_only",
+        default=False,
+        help="Do not download packages, only sync metadata",
```
rjmateus, line 218: If I got it right, it will not synchronize any package, nor save the remote channel metadata in the MLM side. It will result in an empty channel in MLM.
This can be confused with lazy-repo sync proposal we had before, and I think is something that, in my opinion, we should avoid

srbarrios: I can show a working version in oscar-server.mgr.suse.de, we can take a look tomorrow if you have time.
It actually synchronize some packages, the release packages. And the channels shows as synchronized, even if not all packages are synchronized.

This unblocks faster onboardings in our test environments. It might be interesting for customers too.

Btw, the code just pass through different tools an option that has been always implemented inside RepoSync class.

agraul: If we add this flag to `spacewalk-repo-sync`, we better support customer setups using that. Until now, we didn't expose `metadata_only`, a feature used by Red Hat's CDN back when they created this part afaik.

I'm not sure we want to do that, given that we haven't fully decided how we'll design the lazy reposync CLI we've started working on in GSoC. It might lead to two implementations we need to support for the similar things. This decision shouldn't be done quickly.

PR #11449, opened 2026-02-06
(MU) Commit DB changes before refreshing pillar for SSH push minions (bsc#1253712)

rjpmestre:
## What does this PR change?

When updating a base channels for a SSH push minions, it receives the "non updated" pillars because saltutil.refresh_pillar ends up reading uncommitted DB data.
This PR ensures the transaction is committed before refreshing the pillar.

## Codespace

## GUI diff
No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/29019
Port(s): 
- M51 - https://github.com/SUSE/spacewalk/pull/29590
- M50 - https://github.com/SUSE/spacewalk/pull/29591

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/suse/manager/reactor/messaging/ChannelsChangedEventMessageAction.java
+++ b/java/core/src/main/java/com/suse/manager/reactor/messaging/ChannelsChangedEventMessageAction.java
@@ -14,6 +14,7 @@
  */
 package com.suse.manager.reactor.messaging;
 
+import com.redhat.rhn.common.hibernate.HibernateFactory;
 import com.redhat.rhn.common.messaging.EventMessage;
 import com.redhat.rhn.common.messaging.MessageAction;
 import com.redhat.rhn.domain.action.salt.ApplyStatesAction;
@@ -77,6 +78,12 @@ public class ChannelsChangedEventMessageAction implements MessageAction {
 
                     // Regenerate the pillar data
                     MinionPillarManager.INSTANCE.generatePillar(minion);
+
+                    // At this point all channel-related entities are updated in the Hibernate session,
+                    // but the changes have not yet been persisted to the database.
+                    // The following Salt modules do not share this transaction and _read directly from the DB_
+                    // using a separate connection, so we must commit here to make the changes visible.
+                    HibernateFactory.commitTransaction();
 
                     // push the changed pillar data to the minion
                     saltApi.refreshPillar(new MinionList(minion.getMinionId()));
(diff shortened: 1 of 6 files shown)


java/core/src/main/java/com/suse/manager/reactor/messaging/ChannelsChangedEventMessageAction.java
```diff
@@ -78,6 +79,9 @@ public void execute(EventMessage event) {
                     // Regenerate the pillar data
                     MinionPillarManager.INSTANCE.generatePillar(minion);
 
+                    // Commit the current transaction
```
mackdk: Please explain more in detail why the commit is needed here, it will help us in the future to understand why we should not remove this line of code 🙂  I would add the same description also in the body of the commit message.

java/core/src/test/java/com/redhat/rhn/frontend/xmlrpc/system/test/SystemHandlerTest.java
```diff
@@ -470,6 +470,9 @@ public void testSetChildChannelsDeprecated() throws Exception {
 
             handler.setChildChannels(admin, sid, ia32Children);
         }, "allowed invalid child channel to be set.");
+
+        // teardown
+        cleanupServers(server);
```
mackdk: This should not be needed. We have a method in the base test classes called `commitHappenend()`. Calling it will set the `committed` flag to `true`. This will automatically remove everything connected to the organization created by the test, including all systems. It should cover your use case without having to do any manual clean up.

Moreover, doing it like this won't cover the cases where the test fails. If an assert fails or an unexpected exception happens, the `cleanupServers()` won't be called and  the database will be polluted for the next tests.

CDellaGiusta (commented):
While wondering if it could be worth putting the `cleanupServers` call into the  `@AfterEach` `tearDown` method of the test classes involved I saw Thomas' comment about `commitHappened`

java/core/src/test/java/com/redhat/rhn/frontend/xmlrpc/system/test/SystemHandlerTest.java
rjpmestre: I did not realized this existed. Way cleaner

CDellaGiusta (dismissed):

rjpmestre:
> While wondering if it could be worth putting the `cleanupServers` call into the `@AfterEach` `tearDown` method of the test classes involved I saw Thomas' comment about `commitHappened`

Yeah, usually the choice is really about whether tests should be given a clean snapshot up front, or if they should be responsible for cleaning up after themselves. Then, theres also the question of granularity, like doing it per individual test vs once per test class.

If we would be able to use a db migration tool like flyway or liquibase, this would be a no brainer. 
But as it is, - ie, using mechanisms like `commitHappened` - i think introducing a new method that runs cleanup after each test might add complexity without much benefit.

Issue #11448, opened 2026-02-06
Ranking config channels leads to an Internal Server Error


denyskon:
### Problem description

When saving a changed ranking of configuration channels (no matter if you have actually changed anything), an Internal Server Error occurs and nothing is being saved.

### Steps to reproduce

1. Go to System -> Manage Configuration Channels -> View/Modify Rankings
2. Click on "Update Channel Rankings"

### Uyuni version

```bash
2026.01
```

### Uyuni proxy version (if used)

```bash

```

### uyuni-tools version used to deploy/manage the server or proxy (if relevant)

```bash

```

### Useful logs

```bash
2026-02-06 13:04:33,038 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-7] ERROR org.hibernate.engine.jdbc.spi.SqlExceptionHelper - ERROR: duplicate key value viola
tes unique constraint "rhn_servercc_sid_ccid_uq"                                                                                                       
  Detail: Key (server_id, config_channel_id)=(1000010030, 9) already exists.                                                                           
2026-02-06 13:04:33,038 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-7] WARN  org.apache.struts.action.RequestProcessor - Unhandled Exception thrown: class javax
.persistence.PersistenceException                                                                                                                      
2026-02-06 13:04:33,038 [ajp-nio-0:0:0:0:0:0:0:1-8009-exec-7] ERROR com.redhat.rhn.frontend.servlets.SessionFilter - Error during transaction. Rolling 
back                                                                                                                                                   
javax.servlet.ServletException: javax.persistence.PersistenceException: org.hibernate.exception.ConstraintViolationException: could not execute stateme
nt
```

kluchoslaw:
I can only confirm that it is not working in 2026.01 :(

mcalmer:
CC @CDellaGiusta @mackdk 
Maybe another hibernate glitch?

CDellaGiusta:
This PR will fix the issue in the next Uyuni release: https://github.com/uyuni-project/uyuni/pull/11495
Thanks @denyskon for your contribution

PR #11495, opened 2026-02-16
Fix exception while reordering config channels ranking

CDellaGiusta:
## What does this PR change?

This fixes a bug, reproducible this way

- create some config channels in >Configuration > Channels
- onboard a minion
- \> Systems > select the minion > Configuration
- \> Manage Configuration Channels > Subscribe to channels (subscribe the minion to some config channels)
- \> Manage Configuration Channels > View/Modify Rankings (modify the order of the channels)
- \> Update Channel Rankings

An exception is raised, like:
`ERROR org.hibernate.engine.jdbc.spi.SqlExceptionHelper - ERROR: duplicate key value violates unique constraint "rhn_servercc_sid_ccid_uq"  Detail: Key (server_id, config_channel_id)=(1000010000, 3) already exists.`

the annotation in Server class solves the problem:

```
@org.hibernate.annotations.CollectionType(
            type = "com.redhat.rhn.common.hibernate.ForceRecreationListType"
   )
    private List<ConfigChannel> configChannels = new ArrayList<>();
```

This bug has the same root cause as https://github.com/uyuni-project/uyuni/pull/10843

A unit test has been added to avoid this bug and another has been added to verify https://github.com/uyuni-project/uyuni/pull/10843

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Issue(s): https://github.com/uyuni-project/uyuni/issues/11448
Port(s): not backported to 5.1, 5.0 and 4.3 (bug from 5.2 forward)
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/domain/server/Server.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/server/Server.java
@@ -63,6 +63,7 @@ import org.apache.logging.log4j.Logger;
 import org.cobbler.CobblerConnection;
 import org.cobbler.SystemRecord;
 import org.hibernate.annotations.Cascade;
+import org.hibernate.annotations.CollectionType;
 import org.hibernate.annotations.ListIndexBase;
 import org.hibernate.annotations.Type;
 import org.hibernate.annotations.WhereJoinTable;
@@ -233,6 +234,9 @@ public class Server extends BaseDomainHelper implements Identifiable {
             inverseJoinColumns = @JoinColumn(name = "config_channel_id")
     )
     @ListIndexBase(1)
+    @CollectionType(
+            type = "com.redhat.rhn.common.hibernate.ForceRecreationListType"
+    )
     private List<ConfigChannel> configChannels = new ArrayList<>();
 
     @ManyToMany(fetch = FetchType.LAZY, cascade = CascadeType.ALL)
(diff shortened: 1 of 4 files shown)


java/core/src/main/java/com/redhat/rhn/domain/server/Server.java
```diff
@@ -233,6 +233,9 @@ public class Server extends BaseDomainHelper implements Identifiable {
             inverseJoinColumns = @JoinColumn(name = "config_channel_id")
     )
     @ListIndexBase(1)
+    @org.hibernate.annotations.CollectionType(
```
mackdk: Why the fully qualified class name? Can't we just `import org.hibernate.annotations.CollectionType;` and use `@CollectionType` ?

mackdk (dismissed):

java/core/src/main/java/com/redhat/rhn/domain/server/Server.java
CDellaGiusta: Yes. I wrongly thought there was a name mismatch. I corrected it also in the other place where I got this code from. I will squash everything in one commit.
BTW: the two unit tests I added should be useful when rebasing jakarta to master branch, to see if this annotation behaves as expected

mcalmer (approved):

PR #11514, opened 2026-02-19
Add DB disk usage monitoring with notifications, healthchecks, and UI alerts (bsc#1238890)

mbussolotto:
## What does this PR change?

- Periodically checks the database and server disk usage using a new Taskomatic task.
- Triggers a notification if the database and server disk usage is above a certain threshold.
- Displays the database and server disk check status on the login page to warn users.
- Integrates the database and server disk check into the server's health check.

The database and server disk check is scheduled to run every hour, at the top of the hour. This is defined by a cron expression in the rhnTaskoSchedule.sql file.

Warning Threshold:
- Variable: DISKCHECKALERT
- Default: 90 (for 90% usage)
- Action: If disk usage is above this percentage, a notification is shown

Critical Threshold:

- Variable: DISKTHRESHOLD
- Default: 95 (for 95% usage)
- Action: If disk usage is above this percentage, a notification is shown, and the container health check will fail.

## Codespace

## GUI diff
re-write on login page
Screenshot From 2026-03-09 09-33-33

new message on all pages to alert disk usage
Screenshot From 2026-03-19 17-20-54

- [x] **DONE**

## Documentation
- DISKCHECKALERT (default 90) and DISKTHRESHOLD (default 95) can be set  as an env var to change percentage before provide an alert or a critical issue. https://github.com/uyuni-project/uyuni-docs/issues/4760

- [x] **DONE**

## Test coverage

- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28795

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/get_pgsql_disk_severity.sh
+++ b/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/get_pgsql_disk_severity.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: 2025 SUSE LLC
+# SPDX-FileCopyrightText: 2026 SUSE LLC
 #
 # SPDX-License-Identifier: Apache-2.0
 


rjmateus:
@mbussolotto could you add to the description why we are removing this and what will be the alternative to this diskcheck and what will replace it?

java/core/src/main/java/com/suse/manager/utils/DiskCheckHelper.java
```diff
@@ -58,10 +54,9 @@ public DiskCheckSeverity executeDiskCheck() {
      * specific result.
      * @return the exit value as returned from the execution of the script.
      * @throws IOException          when an I/O error occurs during the execution of the script.
-     * @throws InterruptedException when the process is interrupted while waiting for a result from the script.
      */
-    protected int invokeExternalScript() throws IOException, InterruptedException {
-        final Process process = Runtime.getRuntime().exec(new String[]{DISKCHECK_SCRIPT, "-c"});
```
rjmateus, line 64: I think we need to have both checks, one doesn't replace the other.
The new one you are implementing if to check the disk usage for the database volume. The one you are deleting was to check the spacewalk and cache volumes. Both are needed.

mbussolotto: don't we want to: 
- shows on login page warning just for the db
- shutdown whatever container based on  health-cmd  (see https://github.com/SUSE/spacewalk/issues/28795#issuecomment-3642059321 ) ?

rjmateus: As far as I know, the goal was to enhance the UI alert to also include the DB data.
Then we should have the "shutdown whatever container based on health-cmd". It's 2 steps protection :) 
https://github.com/SUSE/spacewalk/issues/28795#issuecomment-3644115443

containers/server-postgresql-image/root/usr/bin/healthcheck.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -e
+
+DISK_USAGE=$(df --output=pcent /var/lib/pgsql/data | tail -1 | tr -d '%')
+
+if [ "$DISK_USAGE" -gt 95 ]; then
```
rjmateus: I'm not sure if we should have just the percentage or if we can have a second check that ensures they have more than a certain amount of free space, even if it is less then 5%, we don't return an error.
For example, if a customer has a really huge database (managing cent OS), they will have a huge disk, so having 5% of a 1Tb disk is 50Gb, which is enough space to operate the database.
I would say something above 2,5 Gb should be the minimum (5% of the minimum hardware requirements for the database).

java/core/src/main/java/com/suse/manager/utils/DBDiskCheckHelper.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.IOException;
+
+/**
+ * Class to execute a check of the available disk space using a sql function
+ */
+public class DBDiskCheckHelper extends DiskCheckHelper {
```
rjmateus, line 24: I think we should also add this to the `LoginController` or `LoginHelper` to check the database disk usage when someone enters the login page.

mbussolotto: yes, all that part is still missing, the PR is on draft :)

rjmateus: Ah, good point :) 
I didn't notice it was on draft, sorry for the noise :)

mbussolotto:
https://github.com/uyuni-project/uyuni/pull/11514/changes/763a0c7142d4248f1a36b425bdb92ec627027ce2 this looks really ugly , maybe @uyuni-project/naica can help with  it :)

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/get_pgsql_disk_severity.sh
```diff
@@ -0,0 +1,48 @@
...
+    COPY tmp_sys_df FROM PROGRAM 'df --output=pcent /var/lib/pgsql/data/ | tail -1';
+    SELECT content INTO raw_output FROM tmp_sys_df;
+    
+    usage_pct := trim(both ' %' from raw_output)::integer;
+
+    RETURN CASE
+        WHEN usage_pct >= $DB_CRIT THEN 3
+        WHEN usage_pct >= $DB_WARN THEN 2
+        ELSE 0
+    END;
```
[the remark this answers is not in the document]
mbussolotto: not an issue, the function is recreated during image upgrade

java/core/src/main/java/com/redhat/rhn/domain/notification/types/DiskCheck.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (c) 2022--2026 SUSE LLC
```
mackdk:
```suggestion
 * Copyright (c) 2026 SUSE LLC
```

java/core/src/main/java/com/redhat/rhn/domain/notification/types/DiskCheck.java
```diff
@@ -0,0 +1,65 @@
...
+    public NotificationSeverity getSeverity() {
+        if (this.severity.equals(DiskCheckSeverity.CRITICAL)) {
+            return NotificationSeverity.ERROR;
+        }
+
+        if (this.severity.equals(DiskCheckSeverity.ALERT)) {
+            return NotificationSeverity.WARNING;
+        }
+        // anything else is informational
+        return NotificationSeverity.INFO;
```
mackdk: You can use a switch expression here:
```java
        return switch (this.severity) {
            case CRITICAL -> NotificationSeverity.ERROR;
            case ALERT -> NotificationSeverity.WARNING;
            // anything else is informational
            default -> NotificationSeverity.INFO;
        };
```

java/core/src/main/java/com/redhat/rhn/taskomatic/task/DiskCheckTask.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void execute(JobExecutionContext context) {
+        final DiskCheckHelper diskCheckHelper = new DiskCheckHelper();
+        LocalizationService ls = LocalizationService.getInstance();
+        DiskCheckSeverity diskCheckSeverity = diskCheckHelper.executeDiskCheck();
+
+        if (diskCheckSeverity.compareTo(DiskCheckSeverity.ALERT) >= 0) {
```
mackdk, line 48: We could move this logic into the `DiskCheckSeverity` enum so we don't repeat it below:

```java
if (dbDiskCheckSeverity.needsAttention()) {
```

Where `needsAttention()` is:

```java
    public boolean needsAttention() {
        return this.compareTo(ALERT) >= 0;
    }
```

java/core/src/main/java/com/redhat/rhn/taskomatic/task/DiskCheckTask.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2009--2016 Red Hat, Inc.
```
mackdk: Should be SUSE here as well 🙂  And the "Red Hat trademarks" section can be removed.

java/core/src/main/java/com/suse/manager/utils/DBDiskCheckHelper.java
```diff
@@ -0,0 +1,55 @@
...
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
```
mackdk: Not needed for a new file 🙂

java/core/src/main/java/com/suse/manager/utils/DBDiskCheckHelper.java
```diff
@@ -0,0 +1,55 @@
...
+        try {
+            Object result = HibernateFactory.getSession()
+                    .createNativeQuery("SELECT get_pgsql_disk_severity()")
+                    .getSingleResult();
+
+            if (result instanceof Number) {
+                return ((Number) result).intValue();
+            }
+            throw new IOException("Database did not return a numeric value.");
+        }
```
mackdk: I _think_ this can be simplified to:
```java
            return HibernateFactory.getSession()
                    .createNativeQuery("SELECT get_pgsql_disk_severity()", Integer.class)
                    .getSingleResult();
```
Hibernate should throw a similar exception if the result is not of the correct type and you are catching `Exception`, so the explicit handle seems redundant to me. 

Anyway, the `createNativeQuery()` version without any class definition is deprecated so please add `Object.class` if want to keep the explicit return value check

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/get_pgsql_disk_severity.sh
```diff
@@ -0,0 +1,48 @@
...
+# Set thresholds from env vars or defaults
+DB_CRIT=${DISKTHRESHOLD:-95}
+DB_WARN=${DISKCHECKALERT:-90}
+
+run_sql() {
+    psql -v ON_ERROR_STOP=1 \
+        -h localhost \
+        -p "${PGPORT:-5432}" \
+        -U "$POSTGRES_USER" \
+        --no-password --no-psqlrc -d susemanager "$@"
```
mackdk: Is `"$@"` needed? Shellcheck is showing the warning `run_sql references arguments, but none are ever passed.` and it seems to me you're passing the sql through the standard input.

containers/server-image/root/usr/bin/healthcheck.sh
```diff
@@ -1,5 +1,43 @@
 #!/bin/bash
+set -o pipefail
+MANAGER_COMPLETE="/root/.MANAGER_SETUP_COMPLETE"
+
+check_manager_login() {
+    local attempts=2
+    local attempt=1
+
+    while [ "$attempt" -le "$attempts" ]; do
+        if curl --noproxy localhost --fail --silent --show-error --connect-timeout 1 --max-time 1 \
```
aaannz: Why `--connect-timeout 1 --max-time 1`? For system under heavy load this can be problematic. Why not some longer time like 10s or so?

containers/server-image/root/usr/bin/healthcheck.sh
```diff
@@ -1,5 +1,43 @@
...
-salt-call --local --no-color status.ping_master localhost |grep -q True
-curl --noproxy localhost --fail http://localhost/rhn/manager/login
+
+if [ $DISK_EXIT_CODE -eq 3 ]; then
+    echo "Healthcheck failed: spacewalk-diskcheck returned critical error (3)"
+    exit 1
+fi
+
+if [ -f "$MANAGER_COMPLETE" ]; then
+    set -e
```
aaannz: Nitpick `set -e` is already active from L25, isn't it?
```suggestion
```

containers/server-postgresql-image/root/docker-entrypoint-initdb.d/zz-source-upgrade-script.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: 2025 SUSE LLC
```
aaannz:
```suggestion
# SPDX-FileCopyrightText: 2026 SUSE LLC
```

containers/server-postgresql-image/root/docker-entrypoint-initdb.d/zz-create-db.sh
```diff
@@ -10,3 +10,4 @@ echo "CREATE ROLE $MANAGER_USER PASSWORD '$MANAGER_PASS' NOSUPERUSER NOCREATEDB
 echo "GRANT ALL PRIVILEGES ON DATABASE $MANAGER_DB_NAME TO $MANAGER_USER;" | run_sql -d $MANAGER_DB_NAME
 echo "ALTER DATABASE $MANAGER_DB_NAME OWNER TO $MANAGER_USER;" | run_sql -d $MANAGER_DB_NAME
 echo "GRANT ALL PRIVILEGES ON SCHEMA public TO $MANAGER_USER;" | run_sql -d $MANAGER_DB_NAME
+echo "ALTER DEFAULT PRIVILEGES FOR ROLE $POSTGRES_USER IN SCHEMA public GRANT EXECUTE ON FUNCTIONS TO $MANAGER_USER;" | run_sql -d $MANAGER_DB_NAME
```
aaannz:
```suggestion
echo "ALTER DEFAULT PRIVILEGES FOR ROLE $POSTGRES_USER IN SCHEMA public GRANT EXECUTE ON FUNCTION get_pgsql_disk_severity TO $MANAGER_USER;" | run_sql -d $MANAGER_DB_NAME
```

Wouldn't this work?

mbussolotto: yes, it's the only function so this would be the same

java/core/src/main/java/com/redhat/rhn/taskomatic/task/TaskHelper.java
```diff
@@ -86,6 +86,23 @@ public static void sendTaskoEmail(Integer orgId, String messageBody) {
...
+    public static void sendCustomEmail(Integer orgId, String subject, String messageBody) {
+        String[] recipients = MailHelper.getAdminRecipientsFromConfig();
+        if (orgId != null) {
+            List<String> emails = getActiveOrgAdminEmails(orgId);
+            recipients = !emails.isEmpty() ?
+                    emails.toArray(new String[0]) :
+                    MailHelper.getAdminRecipientsFromConfig();
+        }
+        MailHelper.withSmtp().sendEmail(recipients, subject, messageBody);
+    }
```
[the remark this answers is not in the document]
cbbayburt: You can also add an overload to the existing `sendTaskoEmail` with an additional `subject` parameter, then call it from the original one to avoid duplicated code.

mbussolotto:
thanks @cbbayburt , I'll have a look at `ActionMessages` implementation then

mbussolotto:
> Great work Michele! It's nice to have this feature back. Would you mind adding some automated E2E tests to cover these changes?
It could fit well on the "finishing" pipeline stage, along with other tests like backup/restore.

sure, I created an issue for it https://github.com/SUSE/spacewalk/issues/30120, not sure if we're going to tackle it in this PR or not

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/get_pgsql_disk_severity.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: 2025 SUSE LLC
```
aaannz:
```suggestion
# SPDX-FileCopyrightText: 2026 SUSE LLC
```

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/get_pgsql_disk_severity.sh
```diff
@@ -0,0 +1,47 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Set thresholds from env vars or defaults
+DB_CRIT=${DISKTHRESHOLD:-95}
+DB_WARN=${DISKCHECKALERT:-90}
+
+run_sql() {
+    PGHOST= PGHOSTADDR= psql -v ON_ERROR_STOP=1 \
```
aaannz: Nitpick:
I know we use it like this before, but this fails shellcheck. ( https://github.com/koalaman/shellcheck/wiki/SC1007 )


```suggestion
    PGHOST='' PGHOSTADDR='' psql -v ON_ERROR_STOP=1 \
```

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+#
+# Image identity is tracked via /etc/uyuni-image-ref (written at build time)
+# and a persisted copy at $PGDATA/.uyuni-image-ref.
+
+set -Eeo pipefail
+
+UPSTREAM_ENTRYPOINT="/usr/local/bin/docker-entrypoint.sh"
+IMAGE_REF_FILE="/etc/uyuni-image-ref"
+PGDATA="${PGDATA:-/var/lib/pgsql/data}"
+UPGDB_DIR="/docker-entrypoint-upgdb.d"
```
aaannz: Nitpick: I must say I am not a big fan of unpronounceable variable names 😄 

```suggestion
UPGRADE_HOOKS_DIR="/docker-entrypoint-upgdb.d"
```

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+        echo "No upgrade scripts found in $UPGDB_DIR, nothing to do."
+        return
+    fi
+
+    local pg_user pg_port
+    pg_user="${POSTGRES_USER:-postgres}"
+    pg_port="${PGPORT:-5432}"
+
+    echo "Starting temporary postgres server for upgrade scripts..."
+    PGUSER="${PGUSER:-${POSTGRES_USER:-postgres}}" \
```
aaannz: Upstream entry point also temporarily clears out notify socket. Is this something we should also do?


```suggestion
    PGUSER="${PGUSER:-${POSTGRES_USER:-postgres}}" NOTIFY_SOCKET='' \
```

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+        return 1
+    fi
+
+    local f
+    for f in "${files[@]}"; do
+        if ! pg_isready --host localhost --port "$pg_port" --username "$pg_user" > /dev/null 2>&1; then
+            echo "Temporary postgres is not ready before running: $f"
+            pg_ctl -D "$PGDATA" status || true
+            return 1
+        fi
```
aaannz: Why is there this check again when we already established database is running?

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+        if pg_isready --host localhost --port "$pg_port" --username "$pg_user" > /dev/null 2>&1; then
+            break
+        fi
+
+        sleep 1
+        attempt=$((attempt + 1))
+    done
+
+    if [ "$attempt" -gt "$attempts" ]; then
+        echo "Temporary postgres did not become ready for upgrades."
```
aaannz:
```suggestion
        log "Temporary postgres did not become ready for upgrades."
```

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+        return
+    fi
+
+    local files=()
+    while IFS= read -r -d '' f; do
+        files+=("$f")
+    done < <(find "$UPGDB_DIR" -maxdepth 1 -type f -print0 | sort -z)
+
+    if [ "${#files[@]}" -eq 0 ]; then
+        echo "No upgrade scripts found in $UPGDB_DIR, nothing to do."
```
aaannz:
```suggestion
        log "No upgrade scripts found in $UPGDB_DIR, nothing to do."
```

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+}
+
+# ---------------------------------------------------------------------------
+# Upgrade script runner
+# Mirrors the logic of docker_process_init_files from the upstream script.
+# ---------------------------------------------------------------------------
+
+run_upgrade_scripts() {
+    if [ ! -d "$UPGDB_DIR" ]; then
+        echo "No $UPGDB_DIR directory found, skipping upgrade scripts."
```
aaannz:
```suggestion
        log "No $UPGDB_DIR directory found, skipping upgrade scripts."
```

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+        echo ""
+    fi
+}
+
+# Persist the current image reference into the data directory so we can
+# detect a change on the next run.
+save_image_ref() {
+    local ref_path="$PGDATA/.uyuni-image-ref"
+    current_image_ref > "$ref_path"
+    echo "Stored image ref: $(cat "$ref_path")"
```
aaannz:
```suggestion
    log "Stored image ref: $(cat "$ref_path")"
```

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+    if [ "$(id -u)" = '0' ]; then
+        log "Running as root, dropping privileges to postgres..."
+        if command -v setpriv > /dev/null; then
+            exec setpriv --reuid=postgres --regid=postgres --clear-groups -- "$BASH_SOURCE" "$@"
+        elif command -v runuser > /dev/null; then
+            exec runuser -u postgres -- "$BASH_SOURCE" "$@"
+        else
+            exec su -p -s /bin/bash postgres -c "$BASH_SOURCE $*"
+        fi
+    fi
```
aaannz: This logic is probably not needed as we know what we have in the container. Just use the only one method.
And also - is this safe to do here? upstream entry point drops privileges only after creating directories.

containers/server-postgresql-image/root/uyuni-entrypoint.sh
```diff
@@ -0,0 +1,201 @@
...
+
+    if [ "$1" != 'postgres' ]; then
+        exec "$@"
+    fi
+
+    if ! db_already_exists; then
+        log "No existing database found – running initial setup."
+        mkdir -p "$PGDATA"
+        log "Delegating to upstream entrypoint: $UPSTREAM_ENTRYPOINT"
+        exec "$UPSTREAM_ENTRYPOINT" "$@"
```
aaannz: Let's pass pristine environment to the upstream entrypoint
```suggestion
        log "No existing database found – running upstream entrypoint: $UPSTREAM_ENTRYPOINT"
        exec "$UPSTREAM_ENTRYPOINT" "$@"
```

containers/server-postgresql-image/Dockerfile
```diff
@@ -41,5 +34,9 @@ LABEL org.uyuni.version="${PRODUCT_VERSION}"
 RUN /usr/sbin/groupmod -g 999 postgres && \
     /usr/sbin/usermod -u 999 -g 999 postgres
 
-ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
+RUN echo "${PRODUCT_VERSION}.%RELEASE%" > /etc/uyuni-image-ref
```
aaannz: I would use the reference label we have above for this. Just to include image name too.


```suggestion
RUN echo "${REFERENCE_PREFIX}/server-postgresql:${PRODUCT_VERSION}.%RELEASE%" > /etc/uyuni-image-ref
```

python/spacewalk/satellite_tools/spacewalk-diskcheck
```diff
@@ -137,44 +59,26 @@ function updateSeverity {
 # Main script
 
 CHECKONLY=false
```
aaannz:
```suggestion
```

mbussolotto:
> Aren't we now missing privileges grant in upgrade path now?

@aaannz yes I dropped it...I retested and they seems to be not required

containers/server-postgresql-image/root/uyuni-entrypoint.sh
mbussolotto: ~~seems `log` is not install...or should I create a function just for this?~~

containers/server-postgresql-image/root/uyuni-entrypoint.sh
mbussolotto: IIRC, without it the container wasn't able to run main psql entrypoint. But I can re-check

aaannz: I mean, upstream entry point has just this:
```bash
         if [ "$(id -u)" = '0' ]; then
                        # then restart script as postgres user
                        exec setpriv --reuid=postgres --regid=postgres --clear-groups -- "$BASH_SOURCE" "$@"
                fi
```

so I don't think we need additional runuser and su fallbacks.

And yes, this is needed for correct start, but only later. For example upstream entrypoint has:
```bash
               docker_setup_env
                # setup data directories and permissions (when run as root)
                docker_create_db_directories
                if [ "$(id -u)" = '0' ]; then
                        # then restart script as postgres user
                        exec setpriv --reuid=postgres --regid=postgres --clear-groups -- "$BASH_SOURCE" "$@"
                fi
```

so the docker_create_db_directories is run under calling user and only after that privs are dropped.

So IMO we should check if database is empty -> exec to upstream entry point.
If it is not, then change to postgres user and call the rest and exec.

java/core/src/main/java/com/redhat/rhn/domain/notification/types/DiskCheck.java
```diff
@@ -0,0 +1,62 @@
...
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+package com.redhat.rhn.domain.notification.types;
+
+import com.redhat.rhn.common.localization.LocalizationService;
+
+import com.suse.manager.utils.DiskCheckSeverity;
+
+public class DiskCheck implements NotificationData {
```
mcalmer, line 17: Please add new types also to `java/conf/rhn_java.conf`.
We use the comment in the config file also for documentation which checks exists.

mbussolotto:
@uyuni-project/release-engineering The PR is ready to be merged, but it would also requires to change on OBS/IBS the db base image tag from 18 to 18-contrib

PR #11532, opened 2026-02-24
Fix SLM 6.2 bootstrap script failure during first execution bsc#1258796

m-czernek:
## What does this PR change?

Bootstrap script on SLM 6.2 would fail the first time we execute it. This is because new `/etc` files are not available in an already opened transaction.

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/29834
Port(s): https://github.com/SUSE/spacewalk/pull/30033, https://github.com/SUSE/spacewalk/pull/30034

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/spacewalk/certs-tools/rhn_bootstrap_strings.py
+++ b/spacewalk/certs-tools/rhn_bootstrap_strings.py
@@ -415,7 +415,7 @@ function setup_bootstrap_repo() {{
 
     if [ -n "$CLIENT_REPO_URL" ]; then
         echo " adding client software repository at $repourl"
-        cat <<EOF >"$repopath"
+        repo_cmd="cat <<EOF >$repopath
 [$reponame]
 name=$reponame
 baseurl=$repourl
@@ -423,7 +423,15 @@ enabled=1
 autorefresh=1
 keeppackages=0
 gpgcheck=0
-EOF
+EOF"
+        if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
+              "$Z_CLIENT_CODE_VERSION" = "6" && \
+              "$Z_CLIENT_CODE_PATCHLEVEL" -ge "2" ]]
+        then
+            call_tukit "$repo_cmd"
+        else
+            bash <<< "$repo_cmd"
+        fi
     fi
 
     # Avoid modularity failsafe mechanism in dnf 4.2.7 or greater
@@ -819,7 +827,11 @@ if [ ! -z "$ORG_GPG_KEY" ]; then
            apt-get --yes install --no-install-recommends gnupg
            apt-key add $GPG_KEY
         else
-           rpm --import $GPG_KEY
+            if [ -z "$SNAPSHOT_ID" ]; then
+                rpm --import $GPG_KEY
+            else
+                call_tukit "rpm --import $GPG_KEY"
+            fi
         fi
         rm -f ${GPG_KEY}
     done
@@ -913,7 +925,7 @@ echo
         # we need to copy certificate to the trustroot outside of transaction for zypper
         cp "$ORG_CA_CERT" /etc/pki/trust/anchors/
         call_tukit "test -d '$CERT_DIR' || mkdir -p '$CERT_DIR'"
-        call_tukit "cp '/etc/pki/trust/anchors/$ORG_CA_CERT' '${CERT_DIR}/${CERT_FILE}'"
+        call_tukit "cp '$ORG_CA_CERT' '${CERT_DIR}/${CERT_FILE}'"
     else
         test -d "$CERT_DIR" || mkdir -p "$CERT_DIR"
         mv "$ORG_CA_CERT" "${CERT_DIR}/${CERT_FILE}"
@@ -943,11 +955,12 @@ if [[ $ACTIVATION_KEYS =~ , ]]; then
 fi
 
 SNAPSHOT_PREFIX=""
-if [ -n "$SNAPSHOT_ID" ]; then
-    if [ -d "/var/lib/overlay" ]; then
-        # SLM 6.1 and older requires writing to snapshot
-        SNAPSHOT_PREFIX="/var/lib/overlay/$SNAPSHOT_ID"
-    fi
+if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
+      "$Z_CLIENT_CODE_VERSION" = "6" && \
+      "$Z_CLIENT_CODE_PATCHLEVEL" -lt "2" ]]
+then
+    # SLM 6.1 and older requires writing to snapshot
+    SNAPSHOT_PREFIX="/var/lib/overlay/$SNAPSHOT_ID"
 fi
 
 MINION_ID_FILE="${{SNAPSHOT_PREFIX}}/etc/salt/minion_id"
@@ -964,7 +977,10 @@ if [ $VENV_ENABLED -eq 1 ]; then
     MINION_SERVICE="venv-salt-minion"
 fi
 
-if [ -z "$SNAPSHOT_PREFIX" ] && [ -n "$SNAPSHOT_ID" ]; then
+if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
+      "$Z_CLIENT_CODE_VERSION" = "6" && \
+      "$Z_CLIENT_CODE_PATCHLEVEL" -ge "2" ]]
+then
     # TU minion writing to /etc, create /etc directories
     mkdir -p $MINION_CONFIG_DIR
 fi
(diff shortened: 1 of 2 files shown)


spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -235,6 +235,14 @@
 if [ -x /usr/sbin/tukit ]; then
     new_transaction
     echo "Transactional system detected. Reboot will be required to finish bootstrapping"
+
+    SLM_6_2_NEWER=1
```
agraul: Is this really needed? We already set these variables:

``` sh
Z_CLIENT_CODE_BASE=slmicro
Z_CLIENT_CODE_VERSION=6
Z_CLIENT_CODE_PATCHLEVEL=2
```

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -415,15 +423,23 @@ def getRegistrationStackSh():
 
     if [ -n "$CLIENT_REPO_URL" ]; then
         echo " adding client software repository at $repourl"
-        cat <<EOF >"$repopath"
+        repo_cmd=$(cat <<EOF
```
agraul: I think we don't need `$(cat ...)` here. Would this also work?

``` sh
        repo_cmd="cat <<EOF >$repopath
[$reponame]
name=$reponame
baseurl=$repourl
enabled=1
autorefresh=1
keeppackages=0
gpgcheck=0
EOF"
```

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -415,15 +423,23 @@ def getRegistrationStackSh():
...
 baseurl=$repourl
 enabled=1
 autorefresh=1
 keeppackages=0
 gpgcheck=0
+INNER_EOF
 EOF
+)
+        if [ "$SLM_6_2_NEWER" -eq 1 ]; then
+            call_tukit "$repo_cmd"
```
agraul, line 431: Alternative:

``` sh
        # In SL Micro 6.2, the repo needs to be set up inside the transaction
        if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
              "$Z_CLIENT_CODE_VERSION" = "6" && \
              "$Z_CLIENT_CODE_PATCHLEVEL" -ge "2" ]]
        then
            call_tukit "$repo_cmd"
        else
            $repo_cmd
        fi
```

agraul:
Another approach could be to define a function that replaces the `cat` usage, e.g.
```sh
function cat_etc {
    local filename=$1; shift
    local text=$1; shift

    if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
        "$Z_CLIENT_CODE_VERSION" = "6" && \
        "$Z_CLIENT_CODE_PATCHLEVEL" -ge "2" ]]
    then
        call_tukit "cat >$filename <<<$text"
    else
        cat >"$filename" <<<"$text"
    fi
}
# call in setup_bootstrap_repo:
cat_etc "$repopath" "[$reponame]
name=$reponame
baseurl=$repourl
enabled=1
autorefresh=1
keeppackages=0
gpgcheck=0"
```

spacewalk/certs-tools/rhn_bootstrap_strings.py
m-czernek: I'm aware, but I originally wasn't a fan of copy-pasting a test that checks 3 variables multiple times. But, if you prefer this, I don't mind either way.

m-czernek:
@agraul this should be ready for re-review. You were right that `rpm --import` failed on all TU systems (I suppose nobody hit that condition just yet with TU minions), so I fixed that. Outside of this, unless I'm missing something else, the rest should be good.

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -943,11 +955,12 @@ def getRegistrationSaltSh(productName):
...
-        # SLM 6.1 and older requires writing to snapshot
-        SNAPSHOT_PREFIX="/var/lib/overlay/$SNAPSHOT_ID"
-    fi
+if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
+      "$Z_CLIENT_CODE_VERSION" = "6" && \
+      "$Z_CLIENT_CODE_PATCHLEVEL" -ge "2" ]]
+then
+    # SLM 6.1 and older requires writing to snapshot
+    SNAPSHOT_PREFIX="/var/lib/overlay/$SNAPSHOT_ID"
 fi
```
agraul, line 964: The conditional checks for `patchlevel >= 2`, but the body is for `patchlevel < 2`.

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -415,15 +415,23 @@ def getRegistrationStackSh():
...
 gpgcheck=0
-EOF
+EOF"
+        if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
+              "$Z_CLIENT_CODE_VERSION" = "6" && \
+              "$Z_CLIENT_CODE_PATCHLEVEL" -ge "2" ]]
+        then
+            call_tukit "$repo_cmd"
+        else
+            bash <<< "$repo_cmd"
```
agraul, line 433:
```suggestion
            $repo_cmd
```

spacewalk/certs-tools/rhn_bootstrap_strings.py
agraul, line 927: Does this work with SL Micro 6.2?

spacewalk/certs-tools/rhn_bootstrap_strings.py
m-czernek: I don't believe this works; when I test in a shell, I get:

```
[mczernek@t14s uyuni] $ echo $repo_cmd
cat <<EOF >/tmp/test [] name= baseurl= enabled=1 autorefresh=1 keeppackages=0 gpgcheck=0 EOF
[mczernek@t14s uyuni] $ $repo_cmd
cat: '<<EOF': No such file or directory
cat: '>/tmp/test': No such file or directory
cat: '[]': No such file or directory
cat: 'name=': No such file or directory
cat: 'baseurl=': No such file or directory
cat: 'enabled=1': No such file or directory
cat: 'autorefresh=1': No such file or directory
cat: 'keeppackages=0': No such file or directory
cat: 'gpgcheck=0': No such file or directory
cat: EOF: No such file or directory
```

spacewalk/certs-tools/rhn_bootstrap_strings.py
m-czernek: Yes; `$CERT_DIR` is `/usr/share/rhn`, and `/usr/share` is RO on 6.2 as well.

spacewalk/certs-tools/rhn_bootstrap_strings.py
m-czernek: Good catch, fixed!

spacewalk/certs-tools/rhn_bootstrap_strings.py
agraul: You're right, the `<<EOF` here document construct is not special when calling it like that and newlines separate commands.

So that leaves us with passing the variable as here string or passing it to `-c`. Let's keep it as you have it.

spacewalk/certs-tools/rhn_bootstrap_strings.py
agraul: I selected the wrong region, I mean `cp $ORG_CA_CERT /etc/pki/trust/anchors/` and then `call_tukit "cp '/etc/pki/trust/anchors/$ORG_CA_CERT' ...`

spacewalk/certs-tools/rhn_bootstrap_strings.py
```diff
@@ -943,11 +955,12 @@ def getRegistrationSaltSh(productName):
...
 
 SNAPSHOT_PREFIX=""
-if [ -n "$SNAPSHOT_ID" ]; then
-    if [ -d "/var/lib/overlay" ]; then
-        # SLM 6.1 and older requires writing to snapshot
-        SNAPSHOT_PREFIX="/var/lib/overlay/$SNAPSHOT_ID"
-    fi
+if [[ "$Z_CLIENT_CODE_BASE" = "slmicro" && \
+      "$Z_CLIENT_CODE_VERSION" = "6" && \
+      "$Z_CLIENT_CODE_PATCHLEVEL" -lt "2" ]]
```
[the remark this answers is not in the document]
m-czernek: ignoring, `[ -n "$SNAPSHOT_ID" ]` was just a check if we're on SLM < 6.2.

spacewalk/certs-tools/rhn_bootstrap_strings.py
m-czernek: Indeed, this was again passing on my system because `/etc/pki/trust/anchors/$ORG_CA_CERT` was there from a previous transaction. Good catch!

We must add this into a transaction (since zypper is called from inside of a transaction), so I'm fixing the source of the cert only. The `updateCertificates` should also execute in a transaction and should require no modifications.

PR #11550, opened 2026-02-28
Testsuite: Ansible basic tests added

alichiel:
## Ansible tests - basic test for playbook parameters editor
I added a couple of tests for ansible editor, mostly variables. The playbook uses a bash script to evaluate given parameters and the tests verify the correct values. Teh playbook also contains, as well as the script, quoted strings to verify the corectness of escaping.

### New scenarios 
The new scenarios were added to the existing feature, there is also a new playbook and bash script files 
Scenarios:
- Run playbook as it is
- Run playbook while changing parameters with values thru the web form fields
- Run playbook while changing empty parameters (defined as empty in the playbook) thru the web form fields
- Run playbook while changing null parameter (defined as null in the playbook) thru the web form fields
- Run playbook with crafted quoted strings while accessing the YAML editor, doing Edit and Save operations without changing the content
- Run playbook, editing the variables thru the web form fields and adding safe values, accessing the YAML editor, doing Edit and Save operations without changing the content

The 2 (CORRECTLY) failing scenarios with opened bugs:
- Scenario: Run the basic tests playbook while editing the NULL variable and changing its value 
  - [Bug 1258017 - Ansible WebUI feature - playbook variable without a value cannot be updated via text field](https://bugzilla.suse.com/show_bug.cgi?id=1258017)
``` 
    And I enter "256 * 256" as "p_null"                                                                                                               # features/step_definitions/navigation_steps.rb:298
      Unable to find field "p_null" that is not disabled (Capybara::ElementNotFound)
      ./features/step_definitions/navigation_steps.rb:299:in `/^I enter "([^"]*)" as "([^"]*)"$/'
      features/secondary/min_ansible_control_node.feature:178:in `I enter "256 * 256" as "p_null"'
``` 
- Scenario: Run the basic tests playbook on SAVE operation of YAML editor with pre-crafted strings
  - [Bug 1257831 - Ansible WebUI feature - updating variables or yaml file leads to errors](https://bugzilla.suse.com/show_bug.cgi?id=1257831)
``` 
Scenario: Run the basic tests playbook on SAVE operation of YAML editor with pre-crafted strings
    Then I should see a "42" text                                                                                                                     # features/step_definitions/navigation_steps.rb:700
      Text '42' not found (ScriptError)
      ./features/step_definitions/navigation_steps.rb:701:in `/^I should see a "([^"]*)" text$/'
      features/secondary/min_ansible_control_node.feature:209:in `I should see a "42" text'
``` 

### Results
- Uyuni, 2026.01 
``` 
19 scenarios (2 failed, 2 skipped, 15 passed)
223 steps (2 failed, 30 skipped, 191 passed)
6m35.853s
``` 

- Multi-Linux Manager 5.1, 5.1.2.20260228 
``` 
20 scenarios (2 failed, 2 skipped, 16 passed)
224 steps (2 failed, 17 skipped, 205 passed)
6m23.243s
``` 

## Other fixes
- flaky test step update 'Then I wait until I see "myself" text' should remove the false result evaluation

### DIFF
--- a/testsuite/features/secondary/min_ansible_control_node.feature
+++ b/testsuite/features/secondary/min_ansible_control_node.feature
@@ -98,7 +98,7 @@ Feature: Operate an Ansible control node in a normal minion
     And I follow "scheduled" in the content area
     Then I wait until I see "1 system successfully completed this action" text or "1 system failed to complete this action" text, refreshing the page
     When I follow "1 system"
-    And I wait until I see the system name "sle_minion" text
+    And I wait until I see the system name of "sle_minion"
     And I follow this "sle_minion" link
     Then I should see a "42" text
     And I should see a "101010" text
@@ -127,7 +127,7 @@ Feature: Operate an Ansible control node in a normal minion
     And I follow "scheduled" in the content area
     Then I wait until I see "1 system successfully completed this action" text or "1 system failed to complete this action" text, refreshing the page
     When I follow "1 system"
-    And I wait until I see the system name "sle_minion" text
+    And I wait until I see the system name of "sle_minion"
     And I follow this "sle_minion" link
     Then I should see a "4200" text
     And I should see a "(code 0)" text
@@ -157,7 +157,7 @@ Feature: Operate an Ansible control node in a normal minion
     And I follow "scheduled" in the content area
     Then I wait until I see "1 system successfully completed this action" text or "1 system failed to complete this action" text, refreshing the page
     When I follow "1 system"
-    And I wait until I see the system name "sle_minion" text
+    And I wait until I see the system name of "sle_minion"
     And I follow this "sle_minion" link
     Then I should see a "42000" text
     And I should see a "4242" text
@@ -189,7 +189,7 @@ Feature: Operate an Ansible control node in a normal minion
     And I follow "scheduled" in the content area
     Then I wait until I see "1 system successfully completed this action" text or "1 system failed to complete this action" text, refreshing the page
     When I follow "1 system"
-    And I wait until I see the system name "sle_minion" text
+    And I wait until I see the system name of "sle_minion"
     And I follow this "sle_minion" link
     Then I should see a "65536" text
     And I should see a "(code 0)" text
@@ -212,7 +212,7 @@ Feature: Operate an Ansible control node in a normal minion
     And I follow "scheduled" in the content area
     Then I wait until I see "1 system successfully completed this action" text or "1 system failed to complete this action" text, refreshing the page
     When I follow "1 system"
-    And I wait until I see the system name "sle_minion" text
+    And I wait until I see the system name of "sle_minion"
     And I follow this "sle_minion" link
     Then I should see a "42" text
     And I should see a "101010" text
@@ -242,7 +242,7 @@ Feature: Operate an Ansible control node in a normal minion
     And I follow "scheduled" in the content area
     Then I wait until I see "1 system successfully completed this action" text or "1 system failed to complete this action" text, refreshing the page
     When I follow "1 system"
-    And I wait until I see the system name "sle_minion" text
+    And I wait until I see the system name of "sle_minion"
     And I follow this "sle_minion" link
     Then I should see a "abc" text
     And I should see a "def" text
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -117,7 +117,7 @@ When(/^I wait at most (\d+) seconds until the event is completed, refreshing the
   end
 end
 
-When(/^I wait until I see the system name "([^"]*)" text$/) do |host|
+When(/^I wait until I see the system name of "([^"]*)"$/) do |host|
   system_name = get_system_name(host)
   step %(I wait until I see "#{system_name}" text)
 end


testsuite/features/secondary/min_ansible_control_node.feature
```diff
@@ -82,6 +83,164 @@ Feature: Operate an Ansible control node in a normal minion
...
+    And I enter "(21 + 21) * 100" as "p1"
+    And I enter "" as "p2"
+    And I enter "" as "p3"
+    And I enter "" as "p4"
+    And I enter "" as "p5"
+    And I click on "Save"
+    And I click on "Schedule"
+    And I wait until I see "Playbook execution has been scheduled" text
+    And I wait for "3" seconds
+    And I click on the link that contains href "/rhn/schedule/ActionDetails.do"
```
maximenoel8: We usually don't refer the page element directly in the step.
It should be encapsulated with high level keyword + switch case in the step.

testsuite/features/secondary/min_ansible_control_node.feature
```diff
@@ -82,6 +83,164 @@ Feature: Operate an Ansible control node in a normal minion
...
+    Given I am on the Systems overview page of this "sle_minion"
+    When I follow "Ansible" in the content area
+    And I follow "Playbooks" in the content area
+    And I wait until I see "/srv/playbooks" text
+    And I click on "/srv/playbooks"
+    And I wait until I see "/srv/playbooks/basic_tests.yml" text
+    And I click on "basic_tests.yml"
+    And I wait until I see "Playbook Content" text
+    And I click on "Edit variables"
+    And I enter "1" as "p1"
```
maximenoel8: On repeated actions, it's better to use this ruby/cucumber method:
https://github.com/SUSE/spacewalk/blob/26b678e13ae650733a3522c7d790d90e6aca3cac/testsuite/features/secondary/srv_password_restriction.feature#L49-L61

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -36,6 +36,29 @@
   raise ScriptError, "Path #{current_path} different than #{arg1}" unless current_path == arg1
 end
 
+When(/^I click on the link that contains href "([^"]+)"$/) do |href|
```
maximenoel8: As explain before, it's better to use high level words with switch case in the step. I would probably not even mention href but just the action.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -52,6 +75,14 @@
   raise ScriptError, "Text '#{text1}' or '#{text2}' not found" unless check_text_and_catch_request_timeout_popup?(text1, text2: text2, timeout: DEFAULT_TIMEOUT)
 end
 
+When(/^I wait until I see "([^"]*)" text or "([^"]*)" text, refreshing the page$/) do |text1, text2|
```
maximenoel8: I know it's not something we are doing for the other steps but I think we should merge both check text1 or text2 and add refresh as optional:

Something like that:

```ruby
When(/^I wait until I see "([^"]*)" text or "([^"]*)" text(?:, (refreshing the page))?$/) do |text1, text2, refresh_option|
  if refresh_option
    # Logic for the refreshing version
    repeat_until_timeout(message: "Couldn't find text '#{text1}' or text '#{text2}'") do
      break if has_content?(text1) || has_content?(text2)
      sleep(3)
      refresh_page
    end
  else
    # Logic for the standard version
    success = check_text_and_catch_request_timeout_popup?(text1, text2: text2, timeout: DEFAULT_TIMEOUT)
    raise ScriptError, "Text '#{text1}' or '#{text2}' not found" unless success
  end
end
```

maximenoel8 (changes requested):

maximenoel8:
Good work ! I added some comments and improvement needed.
The blocker FMPOV is the href step. The others are suggested QOL for us later and can be dismissed. 

And also check rubocop ;-)

testsuite/features/secondary/min_ansible_control_node.feature
alichiel: Yes, I know, but this was problematic so I went with href for purpose.

testsuite/features/secondary/min_ansible_control_node.feature
alichiel: OK, I had to write a new function but it's on the way.

testsuite/features/secondary/min_ansible_control_node.feature
alichiel: This is fixed now, will appear in the new update.

testsuite/features/step_definitions/navigation_steps.rb
alichiel: This function will be removed in the new update.

testsuite/features/step_definitions/navigation_steps.rb
alichiel: Great idea! I'll change it and update it.

alichiel:
## Update 
### Fixes 
- human readable steps 
- rubocop 
- steps update and merge: `^I wait until I see "([^"]*)" text or "([^"]*)" text(?:, (refreshing the page))?$`
- support functions 

### Results
``` 
Failing Scenarios:
cucumber features/secondary/min_ansible_control_node.feature:166 # Scenario: Run the basic tests playbook while editing the NULL variable and changing its value
cucumber features/secondary/min_ansible_control_node.feature:197 # Scenario: Run the basic tests playbook on SAVE operation of YAML editor with pre-crafted strings

19 scenarios (2 failed, 2 skipped, 15 passed)
208 steps (2 failed, 31 skipped, 175 passed)
6m59.683s
```

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -106,6 +117,11 @@
   end
 end
 
+When(/^I wait until I see the system name "([^"]*)" text$/) do |host|
```
Bischoff:
```
^I wait until I see the system name of "([^"]*)"$
```

like the one just below

alichiel: Updated.

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -106,6 +117,11 @@
   end
 end
 
+When(/^I wait until I see the system name of "([^"]*)"$/) do |host|
```
Bischoff, line 120: Sorry:

I wait until I see the name of

like next step.

alichiel: This is different function that doesn't do the job. That's why I created the new one.

Bischoff: Discussing it in slack

maximenoel8 (approved):

PR #11568, opened 2026-03-04
Fixes for hub registration and XMLRPC API

nadvornik:
## What does this PR change?

This PR
- Fixes migration from peripheral servers created by mgradm hub register command, which could create both salt and foreign systems.
- Adds sync.hub.listPeripheralServers that provides complete info for Hub XMLRPC API, including root CAs
- Remove Peripheral Server entitlement from migrated servers as the entitlement is no longer used.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- TBD

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28474

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/spacewalk-java.changes.nadvornik.hub-xmlrpc
+++ b/java/spacewalk-java.changes.nadvornik.hub-xmlrpc
@@ -1,0 +1,2 @@
+- Fix migration of peripheral servers to ISSv3
+- Add listPeripheralServers API method


java/core/src/main/java/com/suse/manager/hub/HubManager.java
```diff
@@ -1159,6 +1193,28 @@ public List<IssPeripheral> listRegisteredPeripherals(User user, PageControl pc)
...
+     * @param user the user
+     * @return a list of maps containing fqdn, id and root_ca
+     */
+    public List<Map<String, Object>> listPeripheralServers(User user) {
+        ensureSatAdmin(user);
+        List<IssPeripheral> peripherals = hubFactory.listPeripherals();
+        List<Map<String, Object>> result = new ArrayList<>();
+        for (IssPeripheral peripheral : peripherals) {
+            Optional<Server> server = ServerFactory.findByFqdn(peripheral.getFqdn());
+            server.ifPresent(s -> {
```
rjmateus, line 1207: In this change you get the list of hub servers, try to find curresponding system ID by the fqdn. If it is not found, the system would not be added to the return list.
Would make sense to have fallback in this case? For example, if it doesn't find the systemID it would add the peripheral to the list and use the periphetal ID as negative (example, if peripherral ID was 10, it would add -10 as the id).
Would this be possible? What do you think would break?

nadvornik: The Hub XMLRPC API expects systemIDs:
https://github.com/uyuni-project/hub-xmlrpc-api?tab=readme-ov-file#python-example

I think we can't break the assumption that the IDs returned by Hub XMLRPC API are the same as IDs on the hub server.

Also, this situation should not happen because on registration in getOrCreateManagerSystem, the system is always created.

PR #11592, opened 2026-03-10
spacecmd: Convert %-formatting and .format() to f-strings

Carbon14-48:
## What does this PR change?

The spacecmd codebase had pylint's `consider-using-f-string` warnings suppressed
on a per-line basis instead of actually fixing them. This PR does the conversion
for real — all `.format()` and `%`-style formatting that was flagged gets converted
to f-strings, and the suppression comments are removed.

Covers both source files in `spacecmd/src/spacecmd/` and test files in `spacecmd/tests/`.

No logic changes, just string formatting cleanup.

## GUI diff

N/A — spacecmd is a CLI tool, no GUI involved.

- [x] **DONE**

## Documentation

No documentation needed.

- [x] **DONE**

## Test coverage

All existing `spacecmd_unittests` pass. No new tests needed since this is a
formatting-only refactor.

- [x] **DONE**

## Links

No related issue — this is a code quality cleanup.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

### DIFF
--- a/spacecmd/src/spacecmd/activationkey.py
+++ b/spacecmd/src/spacecmd/activationkey.py
@@ -572,8 +572,7 @@ def do_activationkey_listpackages(self, args):
         details.get("packages"), key=lambda x: (x["name"]), reverse=True
     ):
         if "arch" in package:
-            # pylint: disable-next=consider-using-f-string
-            print("%s.%s" % (package["name"], package["arch"]))
+            print(f"{package['name']}.{package['arch']}")
         else:
             print(package["name"])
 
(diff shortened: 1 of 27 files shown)


m-czernek:
Thank you for the PR. The problem with `spacecmd` specifically is that we still support executing it on systems with Python 2.7, which is why we cannot use f-strings here, with no plans to deprecate/remove that support. I'm afraid we won't be able to accept this PR.

If you'd still be interested in contributing and improving the Python code quality, we still have a lot of resolvable pylint disable comments in  the `python/` directory. I wouldn't touch things like changing names of functions/classes as that can be quite complex, but there are easy (f-strings, docstrings, used-before-assignment, pointless-string-statement, try-except-raise, ...) as well as medium (arguments-renamed, unused-variable, ...) difficulty linting problems that you could take a look at.

I'd recommend keeping the PRs small, either one file at a time, or even parts of a file at a time, so that it's easier to review and test.

Carbon14-48:
Thanks for the detailed feedback @m-czernek  — I really appreciate it! I didn't realize spacecmd still needs to support Python 2.7, that makes total sense.

I'll close this PR and start fresh with the python/ directory instead, tackling the resolvable pylint disable comments there. I'll keep the PRs small — one file at a time, starting with the easier ones like consider-using-f-string, pointless-string-statement, and try-except-raise.

Thanks again for pointing me in the right direction!

PR #11600, opened 2026-03-10
GitHub actions security hardening

cbbayburt:
## What does this PR change?

This PR hardens PR labeling by replacing pull_request_target usage in the labeling/translation-labeling flow with a pull_request producer + workflow_run consumer model, while keeping labeling behavior aligned. It also simplifies workflow orchestration by consolidating label logic.

PR trigger workflow

- Added a generic PR producer workflow spacewalk/.github/workflows/pr-event-trigger.yml. It exists only to emit reusable workflow_run events for downstream consumers.
- Refactored spacewalk/.github/workflows/labeler.yml into multi-purpose jobs. Split responsibilities into metadata, path labels, external PR label, and translation label handling.
- Keeps privileged jobs metadata-only (no checkout/PR code execution).
- Migrated spacewalk/.github/workflows/check_translations.yml to pull_request. Computes needs_translation and passes it via artifact to the consumer workflow.
- Updated spacewalk/.github/workflows/update_translations.yml to use native concurrency and removed third-party cancel-workflow action.
- Move acceptance feedback comment job from pull_request_target to a safer 2-step (safe trigger, privileged commenter) flow

## Codespace

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/.github/workflows/acceptance_tests_feedback_comment.yml
+++ b/.github/workflows/acceptance_tests_feedback_comment.yml
@@ -1,35 +1,46 @@
-name: Ask for acceptance tests feedback
+name: Post acceptance tests feedback comment
+
 on:
-  pull_request_target:
-    paths:
-      - 'java/**'
-      - 'web/html/src/**'
-      - 'testsuite/**'
-      - '.github/workflows/acceptance_tests_sequential.yml'
-      - '.github/workflows/acceptance_tests_parallel.yml'
-      - '.github/workflows/acceptance_tests_feedback.yml'
-      - '.github/workflows/acceptance_tests_base.yml'
-      - '!java/*.changes*'
-      - '!testsuite/features/build_validation/**'
+  workflow_run:
+    workflows: ["Acceptance tests feedback trigger"]
+    types: [completed]
+
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   comment:
+    if: >-
+      ${{ github.event.workflow_run.conclusion == 'success' &&
+          github.event.workflow_run.event == 'pull_request' }}
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      contents: read
       pull-requests: write
     steps:
+      - name: Read PR number
+        id: pr
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd #v8.0.0
+        with:
+          script: |
+            const pullRequests = context.payload.workflow_run.pull_requests || [];
+            if (pullRequests.length !== 1 || !pullRequests[0].number) {
+              core.setFailed('Expected exactly one pull request in workflow_run payload.');
+              return;
+            }
+            core.setOutput('number', pullRequests[0].number.toString());
+
       - uses: actions-cool/maintain-one-comment@4b2dbf086015f892dcb5e8c1106f5fccd6c1476b # v3.2.0
         with:
+          number: ${{ steps.pr.outputs.number }}
           body: |
             :wave: Hello! Thanks for contributing to our project.
-            Acceptance tests will take some time (aprox. 1h), please be patient :coffee:
+            Acceptance tests will take some time (approx. 1h), please be patient :coffee:
 
-            You can see the progress at the end of this page and at https://github.com/uyuni-project/uyuni/pull/${{ github.event.pull_request.number  }}/checks
+            You can see the progress at the end of this page and at https://github.com/uyuni-project/uyuni/pull/${{ steps.pr.outputs.number }}/checks
             Once tests finish, if they fail, you can check :eyes: the cucumber report. See the link at the output of the action.
-            You can also check the artifacts section, which contains the logs at https://github.com/uyuni-project/uyuni/pull/${{ github.event.pull_request.number }}/checks.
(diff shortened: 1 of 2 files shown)


.github/workflows/check_translations.yml
```diff
@@ -12,85 +16,86 @@ on:
...
-        cd ..
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Count translation changes for base
+        id: base
+        run: |
+          cd master_repo
+          trs=$(git diff origin_check_translations check_translations | wc -l)
+          echo "line_count=${trs}" >> "$GITHUB_OUTPUT"
+          cd ..
```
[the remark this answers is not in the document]
cbbayburt: Keeping the existing behavior.

.github/workflows/check_translations.yml
```diff
@@ -12,85 +16,86 @@ on:
...
-        echo "Translations needed for PR: ${trs}"
-        echo "translations_needed_for_PR=${trs}" >> $GITHUB_ENV
-        cd ..
+      - name: Build translation result artifact
+        run: |
+          if [[ "${{ steps.base.outputs.line_count }}" != "${{ steps.pr.outputs.line_count }}" ]]; then
+            needs_translation=true
+          else
+            needs_translation=false
+          fi
```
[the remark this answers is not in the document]
cbbayburt: Keeping the existing behavior.

.github/workflows/labeler.yml
```diff
@@ -1,14 +1,178 @@
...
+              artifact_id: matchArtifact.id,
+              archive_format: 'zip',
+            });
+            let fs = require('fs');
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/translation-check-result.zip`, Buffer.from(download.data));
+
+      - name: Read translation result
+        id: translation
+        run: |
+          unzip translation-check-result.zip
```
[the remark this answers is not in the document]
cbbayburt: Done

.github/workflows/labeler.yml
```diff
@@ -1,14 +1,163 @@
...
+            core.setOutput('number', pullRequests[0].number.toString());
+
+      - name: Download translation result artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: translation-check-result
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ github.token }}
+          path: .
+
```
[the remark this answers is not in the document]
nodeg: Very valid point from Copilot. We should stick to pinning the version to a commit SHA instead of the version number as per https://docs.github.com/en/actions/reference/security/secure-use?learn=getting_started.

.github/workflows/pr-event-trigger.yml
```diff
@@ -0,0 +1,22 @@
+name: "PR event trigger"
```
szachovy, line 1: I don't get what this one is doing.

.github/workflows/labeler.yml
```diff
@@ -1,14 +1,163 @@
...
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: Number('${{ steps.pr.outputs.number }}'),
+                name: 'needs-translation',
+              });
+            } catch (error) {
+              if (error.status !== 404) {
+                throw error;
+              }
+            }
```
szachovy, line 164: Do we need to catch this error without modified handling?

.github/workflows/labeler.yml
```diff
@@ -1,14 +1,163 @@
...
+    if: needs.pr-metadata.result == 'success'
     permissions:
       contents: read
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
     - uses: actions/labeler@634933edcd8ababfe52f92936142cc22ac488b1b #v6.0.1
       with:
-        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        pr-number: "${{ needs.pr-metadata.outputs.pr-number }}"
```
szachovy, line 66: Can you give point here location of path based labels? They are not directly defined as in the other actions.

.github/workflows/pr-event-trigger.yml
cbbayburt: None of the PR triggers other than `pull_request_target` have elevated privileges, which we need for labelling/commenting PRs. The most common alternative to this is the `workflow_run` event, which also has elevated privileges.

However, `workflow_run` can't be connected to a PR event by itself, so we need a triggering mechanism to run these `workflow_run` actions.

This `pr-event-trigger.yml` is only for triggering these `workflow_run` actions on a PR event. Other actions watch for this via:

```
workflow_run:
    workflows: ["PR event trigger"]
    types: [completed]
```

So the flow is;

1. A PR is created/updated
2. "PR event trigger" dummy action runs and completes
3. Real actions pick up the completion and are triggered

It's a bit hacky, but GH doesn't offer a better alternative yet.

.github/workflows/labeler.yml
cbbayburt: This is just to ignore a 404, which means that the label doesn't exist (harmless). We probably don't want to obscure any other errors, so it's thrown as-is in any other case.

.github/workflows/labeler.yml
cbbayburt: It's at the default location, but sure, it wouldn't hurt to be explicit.

.github/workflows/pr-event-trigger.yml
szachovy: ok

.github/workflows/labeler.yml
szachovy: ok

PR #11616, opened 2026-03-13
QE: pick IPv6 interface for PXE boot from node attributes

NamelessOne91:
## What does this PR change?

It seems we can no longer rely on the assumption that we will have a network interface named **ens4** on the Uyuni proxy and **eth1** on all SUSE Manager proxies.
E.G. The newly added SL Micro 6.2 OS uses **ens4** as well.

However, we already pick up what the private network interface of the proxy is when initializing the corresponding node https://github.com/NamelessOne91/uyuni/blob/master/testsuite/features/support/remote_node.rb#L57
So it should just be enough to read it here.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Issue(s): #
Port(s):  

- 5.1 https://github.com/SUSE/spacewalk/pull/30085
- 5.0 https://github.com/SUSE/spacewalk/pull/30086
- 4.3 https://github.com/SUSE/spacewalk/pull/30087

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/support/retail.rb
+++ b/testsuite/features/support/retail.rb
@@ -1,4 +1,4 @@
-# Copyright (c) 2023-2024 SUSE LLC
+# Copyright (c) 2023-2026 SUSE LLC
 # Licensed under the terms of the MIT license.
 
 # This function returns the net prefix, caching it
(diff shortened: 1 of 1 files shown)


Bischoff (approved):

maximenoel8:
Good catch. Should it be backported to spacewalk ? I think it could avoid other issues.

NamelessOne91:
> Good catch. Should it be backported to spacewalk ? I think it could avoid other issues.

I guess it is not strictly needed but handy to have.
I`ll double check and prepare the backports

PR #11617, opened 2026-03-13
New test for mgradm backup / restore feature

ilmanzo:
## What does this PR change?

Introduce a new test feature for basic functionality of `mgradm backup` and `restore`

## Changelogs

- [x] No changelog needed

reference to https://github.com/SUSE/spacewalk/issues/26937

### DIFF
--- a/testsuite/features/secondary/srv_mgradm_backup.feature
+++ b/testsuite/features/secondary/srv_mgradm_backup.feature
@@ -1,0 +1,27 @@
+# Copyright (c) 2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+@containerized_server
+@mgradm_backup_restore
+Feature: Be able to backup and restore the server
+  As a system administrator
+  I want to back up and restore my containerized server
+  So that I can recover from system failures
+
+# exclude the repository volume due to space constraints
+  Scenario: Perform a test backup
+    When I backup the server excluding the spacewalk volume
+    Then the backup should exist
+
+# stop the service before the restore, otherwise podman network creation will fail
+  Scenario: Restore succeeds with force flag
+    Given the backup exist
+    When I stop the services on server
+    Then I restore the backup
+
+  Scenario: Remove backup and start services
+    Given the backup exist
+    When I remove the backup directory on server
+    And I start the services on server
+    Then service "uyuni-server" is running on server
+    And service "uyuni-db" is running on server
--- a/testsuite/features/step_definitions/backup_restore_steps.rb
+++ b/testsuite/features/step_definitions/backup_restore_steps.rb
@@ -1,0 +1,38 @@
+# Copyright (c) 2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+# @summary Step definitions for mgradm backup and restore operations
+
+When('I backup the server excluding the spacewalk volume') do
+  command = "mgradm backup create #{get_context('backup_dir')} --skipvolumes var-spacewalk"
+  out, code = get_target('server').run(command, runs_in_container: false, timeout: 300)
+  raise "Backup command failed with code #{code} and output: #{out}" unless code.zero?
+end
+
+When('I stop the services on server') do
+  get_target('server').run('mgradm stop', runs_in_container: false)
+end
+
+When('I start the services on server') do
+  get_target('server').run('mgradm start', runs_in_container: false)
+end
+
+When('I restore the backup') do
+  command = "mgradm backup restore #{get_context('backup_dir')} --force"
+  out, code = get_target('server').run(command, runs_in_container: false, timeout: 300)
+  raise "Restore command failed with code #{code} and output: #{out}" unless code.zero?
+end
+
+Then(/^the backup .* exist$/) do
+  raise "Backup directory #{get_context('backup_dir')} does not contain any tar files." unless backup_exists?
+end
+
+Then(/^service "(.*)" is running on server$/) do |service|
+  target = get_target('server')
+  output, _code = target.run("systemctl is-active #{service}", runs_in_container: false)
+  raise "Service #{service} is not running." unless output.strip == 'active'
+end
+
+When('I remove the backup directory on server') do
+  get_target('server').run("rm -rf #{get_context('backup_dir')}", runs_in_container: false)
+end
--- a/testsuite/features/support/backup_restore.rb
+++ b/testsuite/features/support/backup_restore.rb
@@ -1,0 +1,14 @@
+# Copyright (c) 2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+# Helper method to check if backup files exist in the specified directory
+#
+# @return [Boolean] true if backup files exist, false otherwise
+def backup_exists?
+  target = get_target('server')
+  output, _code = target.run("find #{get_context('backup_dir')} -maxdepth 1 -type f -name '*.tar*'", runs_in_container: false)
+  root_level = !output.lines.empty?
+  output, _code = target.run("find #{get_context('backup_dir')} -mindepth 2 -type f -name '*.tar*'", runs_in_container: false)
+  sub_level = !output.lines.empty?
+  root_level && sub_level
+end
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -270,6 +270,20 @@ After('@scope_cobbler') do |scenario|
   end
 end
 
+Before('@mgradm_backup_restore') do
+  add_context('backup_dir', '/var/lib/containers/storage/volumes/test_backup')
+end
+
+After('@mgradm_backup_restore') do |scenario|
+  if scenario.failed?
+    get_target('server').run("rm -rf #{get_context('backup_dir')}", runs_in_container: false)
+    # if server is stopped due to a previous restore failure, start it again
+    # to ensure the system is in a clean state for subsequent tests
+    output, _code = get_target('server').run('systemctl is-active uyuni-server', runs_in_container: false)
+    get_target('server').run('mgradm start', runs_in_container: false) unless output.strip == 'active'
+  end
+end
+
 AfterStep do
   next unless capybara_session_created?
 
--- a/testsuite/run_sets/secondary_finishing.yml
+++ b/testsuite/run_sets/secondary_finishing.yml
@@ -7,6 +7,7 @@
 ## Finishing Secondary features  BEGIN ##
 
 # these features are disruptive for other features, so they are better at the end
+- features/secondary/srv_mgradm_backup.feature
 - features/secondary/srv_rename_hostname.feature
 
 ## Finishing Secondary features END ##


alichiel:
Quick one: maybe the "command" is redundant in "I run the backup/restore command" since it should be in more human readable language like: "I want to backup the server", or something like this.

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,25 @@
...
+# we skip saving the spacewalk volume due to size constraints
+  Scenario: Perform a test backup
+    When I run the mgradm backup command
+    Then the backup must exist
+
+# we need to stop the service before the restore, otherwise podman network creation will fail
+  Scenario: Restore succeeds with force flag
+    Given the backup must exist
+    Then I stop the services on server
+    And I run the mgradm restore command
```
alichiel: For consideration: I miss the "Then" step like "Then I see the server was restored", to ensure the restore was successful. I know it's partially in the "restore" step.

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,25 @@
...
+# Licensed under the terms of the MIT license.
+
+Feature: mgradm backup and restore
+  As a system administrator
+  I want to back up and restore my containerized Uyuni/SUSE Manager instance
+  So that I can recover from system failures
+
+# we skip saving the spacewalk volume due to size constraints
+  Scenario: Perform a test backup
+    When I run the mgradm backup command
```
srbarrios: Comments are not printed on the test report, so I think it's better to include important details on the Gherkin sintax.

I would write the step definition like that:

```
When I backup the server except the spacewalk volume
```

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2017-2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+Feature: mgradm backup and restore
```
srbarrios:
```suggestion
Feature: Validate that we can backup and restore the server
```

testsuite/features/secondary/srv_mgradm_backup.feature
srbarrios: We try to hide technical details on the Gherkin syntax, at least when we find it doable. So in this step, I would write something like:

```
When I restore the server's backup
```

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2017-2026 SUSE LLC.
```
srbarrios: I would not use `mgradm_xxx.rb` but just something like `backup_steps.rb` or if you prefer `backup_restore_steps.rb`, so we highlight the functionality and not the implementation.

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
...
+
+After do |scenario|
+  if scenario.failed?
+    cleanup!
+    # if server is stopped due to a previous failure, start it again
+    # to ensure the system is in a clean state for subsequent tests
+    output, _code = get_target('server').run('systemctl is-active uyuni-server', runs_in_container: false)
+    get_target('server').run('mgradm start', runs_in_container: false) unless output.strip == 'active'
+  end
+end
```
srbarrios: I would move this After method into `env.rb`, even if we define it here it will be taken into account for any scenario during the whole cucumber execution.
So better to have the Cucumber Hooks defined at one place.

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,25 @@
...
+    Then the backup must exist
+
+# we need to stop the service before the restore, otherwise podman network creation will fail
+  Scenario: Restore succeeds with force flag
+    Given the backup must exist
+    Then I stop the services on server
+    And I run the mgradm restore command
+
+  Scenario: Remove backup and start services
+    Given the backup must exist
```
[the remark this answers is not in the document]
srbarrios: Don't trust Copilot here. Our idempotency is at Cucumber Feature layer, not by Scenarios.

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: I'm thinking if it might be more suited for a restricted context ? After all the backup files are created only when the backup test runs, and in other cases it will be an useless overhead. (see copilot's https://github.com/uyuni-project/uyuni/pull/11617#discussion_r2930815978 )

testsuite/features/secondary/srv_mgradm_backup.feature
ilmanzo: I got the point but I'm not sure how to implement a "successful restore" check (beside re-testing the whole product). Ideally the command itself should fail on a restore issue ? 

How about stating it as `I successfully restore the backup` ?

ilmanzo: see https://github.com/uyuni-project/uyuni/pull/11617#discussion_r2931379425

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2017-2026 SUSE LLC.
```
srbarrios: That's a new file, so we only need 2026

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,26 @@
...
+# Licensed under the terms of the MIT license.
+
+@containerized_server
+Feature: Validate that we can backup and restore the server
+  As a system administrator
+  I want to back up and restore my containerized Uyuni/SUSE Manager instance
```
srbarrios: SUSE Multi-Linux Manager.
(Or just write my containerized server)

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
...
+  root_level && sub_level
+end
+
+# Helper method to clean up backup files after tests
+def cleanup!
+  # remove any test backup files if the scenario fails to prevent interference with subsequent tests
+  get_target('server').run("rm -rf #{BACKUP_DIR}", runs_in_container: false)
+end
+
+When('When I backup the server excluding the spacewalk volume') do
```
srbarrios: Duplicated "When"

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
...
+
+Then('the command should succeed') do
+  raise "Command failed with output: #{@last_command_output}" if @last_command_output.nil?
+end
+
+Then('the backup must exist') do
+  raise "Backup directory #{BACKUP_DIR} does not contain any tar files." unless backup_exists?
+end
+
+Then(/^service "(.*)" is running on server$/) do |service|
```
srbarrios, line 30: I'm reviewing from my phone app, but I think we have already a step doing exactly this.

testsuite/features/step_definitions/backup_restore_steps.rb
srbarrios: Yes, make sense using a new tag to be more restrictive. But also this needs to be moved along with the rest of Cucumber hooks, on the env.rb file

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2017-2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+# Step definitions for mgradm backup and restore operations
+
+BACKUP_DIR = '/var/lib/containers/storage/volumes/test_backup'.freeze
+
+# Helper method to check if backup files exist in the specified directory
+def backup_exists?
```
srbarrios: Ideally we organize our test framework code into different folders. On this case, we would involve two kinds: the step definitions folder (those methods using Given/When/Then), and the support folder (for Ruby methods acting as helpers for the step definitons).

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
...
+  target = get_target('server')
+  output, _code = target.run("find #{BACKUP_DIR} -maxdepth 1 -type f -name '*.tar*'", runs_in_container: false)
```
srbarrios: Why we want to separate this check into 2 different levels instead of recursively look at all levels in one call?

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,26 @@
...
+
+# exclude the repository volume due to space constraints
+  Scenario: Perform a test backup
+    When I backup the server excluding the spacewalk volume
+    Then the backup must exist
+
+# stop the service before the restore, otherwise podman network creation will fail
+  Scenario: Restore succeeds with force flag
+    Given the backup must exist
+    Then I stop the services on server
```
srbarrios: We try to be a bit strict with Gherkin verbs. So, if we do an action that we expect some outcome later on, then is a _When_, not _Then_. Even if I could totally understand that it seems more natural language using a _Then_, when we already did some other action.

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,26 @@
...
+    And I successfully restore the backup
+
+  Scenario: Remove backup and start services
+    Given the backup must exist
+    And I remove the backup directory on server
```
srbarrios: When

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,26 @@
...
+    Then I start the services on server
```
srbarrios: And

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,26 @@
...
+    And service "uyuni-server" is running on server
```
srbarrios: Then

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
...
+
+# Helper method to clean up backup files after tests
+def cleanup!
+  # remove any test backup files if the scenario fails to prevent interference with subsequent tests
+  get_target('server').run("rm -rf #{BACKUP_DIR}", runs_in_container: false)
+end
+
+When('When I backup the server excluding the spacewalk volume') do
+  command = "mgradm backup create #{BACKUP_DIR} --skipvolumes srv-spacewalk"
+  @last_command_output = get_target('server').run(command, runs_in_container: false)
```
srbarrios: We try to avoid using global/class variables, instead we use set_context and get_context methods

testsuite/features/secondary/srv_mgradm_backup.feature
srbarrios: I'm not sure if we could made this check using _mgradm status_ or maybe checking the containers state with _podman ps_.
My reasoning is trying to make this test less coupled to systemd services, as we only have them on the podman flavor, but we will not have them on the Kubernetes cluster.
But we can always tackle this angle im the future.

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2017-2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+# Step definitions for mgradm backup and restore operations
+
+BACKUP_DIR = '/var/lib/containers/storage/volumes/test_backup'.freeze
+
+# Helper method to check if backup files exist in the specified directory
```
srbarrios: Not a big deal, but will be good to use YARD style for the header comments on Ruby methods.
You can see an example on: https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/support/keyvalue_store.rb

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: if you refer to https://github.com/uyuni-project/uyuni/blob/f3a04aa156a51a4fd4ba05ea535d514a6403213c/testsuite/features/step_definitions/command_steps.rb#L301 , it checks the service in the container and not in the host ; is there some other methods I can use ?

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: Because it's more robust than just checking for "any tar file must exist". A successful backup is done by archiving different stuff in the proper places.

testsuite/features/secondary/srv_mgradm_backup.feature
ilmanzo: yes, it's a bit tricky because container may be up and running but service is not responding for some reason (due to networking or something else). I'm towards a more E2E check, even leveraging the web interface perhaps ?

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: The constant and helper methods are being used only in this test. What's the final benefit of spreading them among many files ?

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: ok ; how about enforcing this good practice with a linter step upon push ?

testsuite/features/step_definitions/backup_restore_steps.rb
srbarrios: Right, that one.
Maybe you can extend it so we can choose to check inside or outside a container?
Otherwise, I would at least place both step definitions together on the same command_steps.rb file.

Pablogoliva (dismissed):

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017-2026 SUSE LLC.
```
Bischoff: 2026 **only**, this is new.

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
...
+# Licensed under the terms of the MIT license.
+
+@containerized_server
+@mgradm_backup_restore
+Feature: Validate that we can backup and restore the server
```
Bischoff: No. It's not a feature of the software to validate anything.
That's what your test wants to check.

Change to:
```
Feature: Be able to backup and restore the server
```

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
...
+  As a system administrator
+  I want to back up and restore my containerized server
+  So that I can recover from system failures
```
Bischoff, line 9: That's indeed the feature of the software that you want to test! Perfect.

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
...
+
+# exclude the repository volume due to space constraints
+  Scenario: Perform a test backup
+    When I backup the server excluding the spacewalk volume
+    Then the backup must exist
+
+# stop the service before the restore, otherwise podman network creation will fail
+  Scenario: Restore succeeds with force flag
+    Given the backup must exist
```
Bischoff: Are you sure you want to repeat it? We don't ensure idempotency between scenarios.
We only ensure idempotency at feature level.

(and it would be "Given the backup exists").

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
...
+    When I stop the services on server
+    Then I successfully restore the backup
+
+  Scenario: Remove backup and start services
+    Given the backup must exist
```
Bischoff: Same remark.

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
...
+    When I backup the server excluding the spacewalk volume
+    Then the backup must exist
+
+# stop the service before the restore, otherwise podman network creation will fail
+  Scenario: Restore succeeds with force flag
+    Given the backup must exist
+    When I stop the services on server
+    Then I successfully restore the backup
+
+  Scenario: Remove backup and start services
```
Bischoff, line 22: This is a cleanup scenario. Prefix it with "Cleanup: ".

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
...
+    Given the backup must exist
+    When I remove the backup directory on server
+    And I start the services on server
+    Then service "uyuni-server" is running on server
```
Bischoff, line 26: "should be"

testsuite/features/secondary/srv_mgradm_backup.feature
```diff
@@ -0,0 +1,27 @@
...
+@mgradm_backup_restore
+Feature: Validate that we can backup and restore the server
+  As a system administrator
+  I want to back up and restore my containerized server
+  So that I can recover from system failures
+
+# exclude the repository volume due to space constraints
+  Scenario: Perform a test backup
+    When I backup the server excluding the spacewalk volume
+    Then the backup must exist
```
Bischoff: "should exist"

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,38 @@
...
+
+# @summary Step definitions for mgradm backup and restore operations
+
+When('I backup the server excluding the spacewalk volume') do
+  command = "mgradm backup create #{get_context('backup_dir')} --skipvolumes var-spacewalk"
+  out, code = get_target('server').run(command, runs_in_container: false, timeout: 300)
+  raise "Backup command failed with code #{code} and output: #{out}" unless code.zero?
+end
+
+When('I stop the services on server') do
```
Bischoff, line 12: The title is wrong. What you do here is stopping the server, not the services that constitute it.
```
When(^I stop the server$`) do
```
Notice the regex anchors `^$`.

Also, this is a generic step. It should not be in this file dedicated to backup and restore.

Finally, are we sure this step does not exist already? Please grep for `mgradm`.

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,38 @@
...
+  get_target('server').run('mgradm stop', runs_in_container: false)
+end
+
+When('I start the services on server') do
```
Bischoff, line 16: Same remarks:
* wrong title
* missing anchors
* generic step to go to a generic steps file
* sure this is new?

testsuite/features/step_definitions/backup_restore_steps.rb
```diff
@@ -0,0 +1,38 @@
...
+  get_target('server').run('mgradm start', runs_in_container: false)
+end
+
+When('I successfully restore the backup') do
```
Bischoff: `successfully` is not needed here, it's implicit in all our `When` steps.

Missing anchors.

testsuite/features/step_definitions/backup_restore_steps.rb
Bischoff: In fact you have been mislead by `service` @srbarrios . He means `server` or `container`.

Here we see how important is to choose correct English wording.

Bischoff (changes requested):

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: thanks. See also https://github.com/uyuni-project/uyuni/pull/11617#discussion_r2931379425

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: Which can be a correct title ?

testsuite/features/step_definitions/backup_restore_steps.rb
ilmanzo: which anchors and where ?

testsuite/features/step_definitions/backup_restore_steps.rb
srbarrios: Hi @Bischoff I don't see where I could have a misleading.
What I'm proposing is to re-use:
 ``` When(/^I wait until "([^"]*)" service is active on "([^"]*)"$/) do |service, host| ```
 to have something like:
 ``` When(/^I wait until "([^"]*)" service is active on "([^"]*)"((?: inside the container)?)$/) do |service, host, inside_container| ```
Instead of adding a new step that does something similar inside `backup_restore_steps.rb`, when in fact, the step can be generalized for much more scenarios than just backup/restore.

ilmanzo:
closing because, looking at all the nitpicks and not answered questions, the test specs probably needs more refinement. Thanks for all the valuable contribution.

PR #11629, opened 2026-03-16
drop pgjdbc-ng

rjpmestre:
## What does this PR change?

This PR removes the pgjdbc-ng (and associated netty) dependency.

## Codespace

## GUI diff
No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/29095

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
```diff
@@ -32,68 +31,71 @@
...
+    private final Connection connection;
     private final List<ThreadPoolExecutor> executorServices = IntStream.range(0, THREAD_POOL_SIZE + 1).mapToObj(i ->
         new ThreadPoolExecutor(
                 1,
                 1,
                 0L,
                 TimeUnit.MILLISECONDS,
                 new LinkedBlockingQueue<>(),
-                new BasicThreadFactory.Builder()
+                BasicThreadFactory.builder()
```
mackdk, line 139: I think we have some more occurrences of `new BasicThreadFactory.Builder()` in other files, which I see is now deprecated. Please update also the other places so even CoPilot will be happy 😄

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
```diff
@@ -104,84 +106,98 @@ public PGEventStream() throws SaltException {
...
+            pollPendingNotifications();
+
+            // if we have any rows in suseSaltEvent that do not yet have a process task active
+            // then schedule tasks for them
+            // this can only happen in case we lost notifications somehow
+            List<Long> allJobs = SaltEventFactory.countSaltEvents(THREAD_POOL_SIZE + 1);
+
+            List<Long> missingJobs = IntStream.range(0, allJobs.size())
+                    .mapToObj(i -> executorServices.get(i).getActiveCount() > 0 ? 0L : allJobs.get(i))
+                    .filter(l -> l > 0)
```
mackdk: Why is this change needed? We weren't removing the zeros before at it seems to me that Copilot as a point when it says that this will screw the alignment when the code iterates from 0 to the number of threads in `handleNotifications()`. 

Could you explain better how this works?

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
```diff
@@ -104,84 +106,98 @@ public PGEventStream() throws SaltException {
...
+            }
+
+        }
+        catch (SQLException e) {
+            // DB connection is probably broken
+            // make sure that the callback does not use the old session
+            HibernateFactory.closeSession();
+            clearListeners(0, "Postgres notification connection was lost");
+            watchdogExecutor.shutdown();
+            executorServices.forEach(ThreadPoolExecutor::shutdown);
```
mackdk: Calling `shutdown()` on an executor is not enough as [it does not wait for previously submitted tasks to complete execution](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#shutdown--). You have to call `awaitTermination()`, we have an example in [CoCoAttestation `ProcessingThread`](https://github.com/uyuni-project/uyuni/blob/a30e7f36b4078f8882ffc13ac29162100e0fa294/microservices/coco-attestation/attestation-core/src/main/java/com/suse/coco/attestation/ProcessingThread.java#L124-L137).

Moreover, this whole `catch (SQLException e)`  block seems problematic, if it ever occurs:
1) we are attempting to shut down the `watchdogExecutor` from within the thread the executor is running. It should be ok, but maybe we should test if there might be concurrency problems.
2) We are closing  the hibernate session before the `executorServices` have stopped and they might still attempt to reopen and use a new session.
3) We are shutting down the watchdog, but the `connection` is not closed.

Maybe we should just call `close()` here? Do you think it's possible to cover these cases with unit tests?

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
```diff
@@ -258,6 +274,8 @@ public boolean isEventStreamClosed() {
     public void close() throws IOException {
         try {
             connection.close();
+            watchdogExecutor.shutdown();
+            executorServices.forEach(ThreadPoolExecutor::shutdown);
```
mackdk: Also here you need to wait for the threads to complete. And I think copilot has a point here: the connection should be closed only after all the executors and threads are shut down (at least the `watchdogExecutor` who's using it), and we also should check if it's not null before attempting to close it 😅 

Moreover, who is calling this `close()` method?

susemanager-utils/testing/docker/scripts/reset_pgsql_database.sh
```diff
@@ -51,3 +51,6 @@ echo "INSERT INTO  rhnChannelFamily (id, name, label, org_id)
       'private-channel-family-1', 1);" | spacewalk-sql --select-mode -
 echo "INSERT INTO  rhnPrivateChannelFamily (channel_family_id, org_id) VALUES  (1000, 1);" | spacewalk-sql --select-mode -
 
+echo "Increasing max_connections"
+su - postgres -c "psql -d postgres -c \"ALTER SYSTEM SET max_connections = 200;\""
+su - postgres -c "/usr/lib/postgresql/bin/pg_ctl restart"
```
mackdk: I don't think this change should be needed. If the tests are not passing without it, I think there is a problem somewhere in how we handle the resources clean-up  in the new code, and we need to investigate it.

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
```diff
@@ -104,84 +107,145 @@ public PGEventStream() throws SaltException {
...
+                stmt.execute("LISTEN " + SUSE_SALT_EVENT_NOTIFICATION_CHANNEL);
             }
 
-            startConnectionWatchdog();
+            watchdogExecutor.scheduleAtFixedRate(this::watchdogTask, 0, 5, TimeUnit.SECONDS);
 
             LOG.debug("Listening succeeded, making sure there is no event left in queue...");
-            notification(SaltEventFactory.countSaltEvents(THREAD_POOL_SIZE + 1));
+            handleNotification(SaltEventFactory.countSaltEvents(THREAD_POOL_SIZE + 1));
         }
```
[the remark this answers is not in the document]
mackdk: I think I agree on this one: if the connection is open when we catch the `SQLException` we need to close it before re-throwing the exception. And maybe we should also catch `RuntimeException`.

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
```diff
@@ -104,84 +106,98 @@ public PGEventStream() throws SaltException {
...
-                    clearListeners(0, "Postgres notification connection was lost");
-                }
-                catch (Exception e) {
-                    LOG.error("Unexpected exception:", e);
-                }
-                finally {
-                    // create new session for each run
-                    HibernateFactory.closeSession();
-                }
+    public void watchdogTask() {
```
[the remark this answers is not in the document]
mackdk: I think this makes sense as well, since you need it just for the unit test if I read the code correctly.

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
```diff
@@ -104,84 +106,98 @@ public PGEventStream() throws SaltException {
...
+
+            List<Long> missingJobs = IntStream.range(0, allJobs.size())
+                    .mapToObj(i -> executorServices.get(i).getActiveCount() > 0 ? 0L : allJobs.get(i))
+                    .filter(l -> l > 0)
+                    .toList();
+
+            if (missingJobs.stream().mapToLong(l -> l).sum() > 0) {
+                LOG.warn("Found {} events without a job. Scheduling...", missingJobs);
+                handleNotification(missingJobs);
+            }
```
mackdk: I'm wondering if this " we lost notifications somehow" is a real use case or not. We don't have anything like that in cocoattestation, we keep listening.

mackdk (changes requested):

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
rjpmestre: oh! you're right! I tried merging the if statement right below into this lambda, but it turned out not to be very readable. I ended up leaving this line by mistake.

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
rjpmestre: Thanks for the remarks, and i took it into consideration on my update. 
I created a `shutdown` method that centralizes all mechanisms we want to consider. This way, depending on the caller, we can also `awaitTermination` for `watchdogExecutor` thread (and/or newly created `pollForNotifications`)

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
rjpmestre: Good catch! `SaltReactor` would start the `PGEventStream`. But `SaltReactor.stop` was not close it.

susemanager-utils/testing/docker/scripts/reset_pgsql_database.sh
rjpmestre: I'll remove it and attempt running all tests again. Hopefully, the (re)-rework doesnt require this adjustment 🤞

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
rjpmestre: To answer your question, I started digging into the origin of that code block. I found that it was introduced in [this PR as part of a fix](https://github.com/SUSE/spacewalk/pull/7004), just a couple of months after the initial release of the functionality. I think that block wasn’t the “main” fix itself, but rather a safety mechanism to handle orphan jobs they probably kept finding on those tests.

For me, the PR description doesn’t fully explain why that code block in specific was added, so still I don’t have a definitive answer to your question. 
However, it made me reconsider my adaptation. From the beginning, I wasn’t that confident about running the pg notification poll/check before “whatever the watchdog was running” ..especially in terms of performance.

Seeing the stress tests in the PR triggered me into reworking my adaptation so it would replicates more closely the behavior we already have with `pgjdbc-ng`. I’d say this is the safest approach, at least until we can run similar stress tests again.

java/core/src/main/java/com/suse/manager/reactor/PGEventStream.java
rjpmestre: There was another occurrence, in `MessageQueueThreadPool` class

java/core/src/test/java/com/suse/manager/reactor/listener/PGEventStreamTest.java
```diff
@@ -0,0 +1,280 @@
...
+            allowing(mockConfigDefaults).getSaltEventConnectionWatchdogIntervalSeconds();
+            will(returnValue(1));
+            allowing(mockConfigDefaults).getSaltEventsPerCommit();
+            will(returnValue(configDefaults.getSaltEventsPerCommit()));
+            allowing(mockConfigDefaults).getSaltEventThreadPoolSize();
+            will(returnValue(configDefaults.getSaltEventThreadPoolSize()));
+            allowing(mockConfigDefaults).isPrometheusMonitoringEnabled();
+            will(returnValue(configDefaults.isPrometheusMonitoringEnabled()));
+            allowing(mockConfigDefaults).getJdbcConnectionString();
+            will(returnValue(configDefaults.getJdbcConnectionString()));
```
mackdk: Not sure to follow why you need to mock the `ConfigDefaults`. Are we missing those values in the test `rhn.conf`? Would it be possible to just call `Config.get().set...(property)` to set the value of the property you need and the clean them all in the tear down?

java/spacewalk-java.spec
```diff
@@ -112,6 +112,7 @@ BuildRequires:  istack-commons-runtime
 BuildRequires:  jade4j
 BuildRequires:  java-%{java_version}-openjdk-devel
 BuildRequires:  java-saml
+BuildRequires:  jctools
```
mackdk, line 114: Out of curiosity, do you know who requires jctools? I don't think it's needed to build though, it's marked as runtime dependency in maven.

mackdk (dismissed):
LGTM

java/core/src/test/java/com/suse/manager/reactor/listener/PGEventStreamTest.java
rjpmestre: There was no particular reason. I'll adapt it.

java/spacewalk-java.spec
rjpmestre: It appears to be an optional sub-dependency of log4j core (accordingly to our current) 2.17.2.
I couldn’t find any direct usage in our java source, nor any reference in our log4j xml files (ie, no `AsyncLogger` or `AsyncRoot` usages). 

atm, not sure this is enough to conclude we could just remove it. I'd suggest opening another card to run tests having this change isolated.

PR #11643, opened 2026-03-18
Avoid duplicates with reposync (bsc#1251821, bsc#1252793)

vzhestkov:
## What does this PR change?

Prefer the entries with `org_id` equal to `NULL` if exist in case if there are multiple instances of the same packages present, either vendored or from custom channels.

## Codespace

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: the logic in legacy code is overcomplicated and spread across different pars. Will be extremely hard to suggest any proper test.

## Links

Tracks: https://github.com/SUSE/spacewalk/issues/28612, https://github.com/SUSE/spacewalk/issues/28770

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/python/spacewalk/satellite_tools/reposync.py
+++ b/python/spacewalk/satellite_tools/reposync.py
@@ -3020,10 +3020,10 @@ class RepoSync(object):
         if self.org_id:
             param_dict["org_id"] = self.org_id
             # pylint: disable-next=invalid-name
-            orgidStatement = " = :org_id"
+            orgidStatement = "(p.org_id = :org_id or p.org_id is NULL)"
         else:
             # pylint: disable-next=invalid-name
-            orgidStatement = " is NULL"
+            orgidStatement = "p.org_id is NULL"
 
         h = rhnSQL.prepare(
             # pylint: disable-next=consider-using-f-string
@@ -3037,7 +3037,7 @@ class RepoSync(object):
               join rhnChecksumView c on p.checksum_id = c.id
               join rhnChannelPackage cp on p.id = cp.package_id
              where pn.name = :name
-               and p.org_id %s
+               and %s
                and pevr.version = :version
                and pevr.release = :release
                and pa.label = :arch
--- a/python/spacewalk/satellite_tools/reposync.py
+++ b/python/spacewalk/satellite_tools/reposync.py
@@ -3066,7 +3066,7 @@ class RepoSync(object):
         h = rhnSQL.prepare(
             # pylint: disable-next=consider-using-f-string
             """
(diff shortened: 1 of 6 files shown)


agraul (dismissed):
I'm not sure we have enough information to decide if we want to go this route. Until now, the "packages are not shared between orgs" was a deliberate decision. Why do we need to change that now? What alternatives have you considered?

If we decide to go in this direction, it would be great if we could clean up the code changes a bit, with explicit dependencies / assumptions.

rjmateus:
@agraul packages that are specific to a organization will not be shared to another.
What this introduces is that if the exact same package exists as a vendor one we would use that instead of make a org specific copy, that would lead to duplicated packages in dB and file system
That vendor package would already by available on both orgs, so we are not openings anything more.
The use case is in hub you use clm to create clone channels, and packages in those channels Will have org id null, to avoid duplicated. Issv2 uses the same Logic.
Now if we sync them to a peripheral, without this Change, the peripheral server would get them duplicated and use double the disk space

agraul:
> @agraul packages that are specific to a organization will not be shared to another. What this introduces is that if the exact same package exists as a vendor one we would use that instead of make a org specific copy, that would lead to duplicated packages in dB and file system That vendor package would already by available on both orgs, so we are not openings anything more. 

That makes sense.

> The use case is in hub you use clm to create clone channels, and packages in those channels Will have org id null, to avoid duplicated. 

That's new with this PR, right?

---

Thanks for the explanation. I'm okay with the approach.

We still need tests and to bring the ordering close to iterating over packages, as it's written the requirement will not be understood when we go back to the `for p in packs:` loop.

rjmateus:
> That's new with this PR, right?

yes, with this PR the packages from CLM project that are sync to the peripheral will get org id null, which was what was already happening on the HUB and was also the behavior in the peripheral when using ISSv2

rjmateus (approved):
LGTM, I thin kwe are good to go with the current approach. As an improvement, I think some comments can be added to the loop block to give more context about why we need this (and I think that would be enough to overcome Alex concerns)

vzhestkov:
In `for p in packs:` loop the packages hould be already ordered as we adjusted the query to return the entries with `org_id` with `NULL` values first. The idea here is to give the packages with `org_id` equal to `NULL` more priority as they are vendored, but they could be only picked to the custom channels from vendored sources and it's not the case here when one organisation could get the packages from the other, but different orgs can link to the vendored packages. And `disassociate_package_by_id` was implemented and colled to unlink the packages which are not vendored (have `org_id` other than `NULL`) in case if there is a better replacement with vendored package available.

PR #10813, opened 2025-09-02
Add GitHub job for automatic test run insertion into database

akayiz:
## What does this PR change?

Closes #10744 

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- In README.md

- [x] **DONE**

## Test coverage
ℹ️ If a major new functionality is added, it is **strongly recommended** that tests for the new functionality are added to the Cucumber test suite

- No tests

- [x] **DONE**

## Links

Issue(s): #10744 

- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

Issue #10744, opened 2025-08-19
Add GH Workflow for Inserting Test Runs into DB


akayiz:
Create a GitHub Action workflow that continuously adds new test runs to a database.

This implements the "Automated GitHub Actions Workflow" represented in the following diagram in the "Database Insertion" section at the bottom right.

<img width="1722" height="975" alt="Image" src="https://github.com/user-attachments/assets/6da55253-21a4-435e-9481-05fa674b5697" />

This diagram is from the [Predictive Test Selection README](https://github.com/uyuni-project/uyuni/blob/master/testsuite/ext-tools/predictive-test-selection/README.md)

pks2906:
Hi,

I reviewed the previous PR (#10813) and noticed the issue with accessing secrets in pull_request workflows, which prevents DB insertion from forked contributions.

I would like to work on an alternative approach using a scheduled (cron-based) GitHub Actions workflow to periodically collect test run data and insert it into the database.

This should avoid the secrets limitation and decouple DB insertion from PR workflows.

Please let me know if this approach sounds good.

Thanks!

akayiz:
Hi @pks2906

Sorry about the late reply.

Yes, this approach is exactly what we had in mind as a fix, I am wondering how you stumbled upon this issue/PR, but I am glad you did :)

Actually, @srbarrios at the time was kind enough that he took what I did and converted it into a cron-based workflow in #10937 exactly as you are suggesting, as this feature wasn't (and still isn't) a blocker to the project we were working on, we decided to postpone working on it and move on with more essential parts of the project, so till now, I didn't thoroughly review Oscar's PR.

Quickly thinking about it now, there are 2 ways to implement this in a cron-based workflow that come to my mind currently:

1. The way that Oscar did it in his PR, reusing the `insert_test_run_into_db.py` script I created that inserts a single PR run into the database and running it on all the past PRs, I believe the problem with this might be that we will miss multiple test runs on the same PR.

Taking a quick glance at Oscar's PR now, probably there is a way to tweak it to avoid this issue and insert all test runs, even multiple ones on a single PR.

2. The way I believe might be easier to implement, is to discard this `insert_test_run_into_db.py` script I created, and implement this using the backfilling process we already have and that I already use to backfill PRs into the database every 3 months, but rather than me doing it manually every 3 months, we'll have the workflow do it automatically every X hours/days.

[...]

PR #10937, opened 2025-10-03
Schedule the collection of PR tests results into Database using a new GH workflow

srbarrios:
## What does this PR change?

Schedule the collection of PR tests results into Database using a new GH workflow

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
ℹ️ If a major new functionality is added, it is **strongly recommended** that tests for the new functionality are added to the Cucumber test suite
- No tests

- [x] **DONE**

## Links

No ports

- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

PR #11648, opened 2026-03-19
ci: add scheduled workflow for test run insertion (cron-based)

pks2906:
## What does this PR change?

This PR introduces an **initial implementation of a scheduled GitHub Actions workflow** to handle insertion of test run data into the database.

While reviewing the previous attempt in PR #10813, I noticed that the workflow relied on the `pull_request` event, which prevents access to repository secrets for forked contributions. This makes database insertion fail in such cases.

Based on the discussion in that PR, where maintainers suggested a **periodic insertion workflow** as a better approach, this PR implements a cron-based workflow that:

- Runs independently of PR workflows
- Avoids the GitHub Actions secrets limitation for forks
- Lays the foundation for a decoupled and reliable data pipeline

### Current scope of this PR:
- Adds a scheduled GitHub Actions workflow (`cron`)
- Sets up the basic pipeline structure (checkout + Python setup)
- Prepares the groundwork for artifact fetching and processing

### Future improvements (follow-up PRs):
- Fetch test run artifacts from GitHub Actions
- Parse Cucumber/JSON test reports
- Insert processed data into the database

---

## Codespace

(No changes required)

---

## GUI diff

No difference.

Before:  
After:

- [x] **DONE**

---

## Documentation

This change only affects CI/CD workflows and does not impact user-facing functionality or APIs.

- [x] **DONE**

---

## Test coverage

- No tests: this change introduces CI/CD workflow infrastructure and does not affect application logic

Tests will be considered once the data processing and DB insertion logic are implemented.

- [x] **DONE**

---

## Links

Issue(s): #10744  
Related PR: #10813

- [x] **DONE**

---

## Changelogs

- [x] No changelog needed

This change only introduces internal CI/CD workflow improvements and does not affect packages or user-visible behavior.

---

## Re-run a test

(No re-run needed)

---

### DIFF
--- a/.github/workflows/insert-test-runs.yml
+++ b/.github/workflows/insert-test-runs.yml
@@ -1,0 +1,25 @@
+name: Insert Test Runs into DB (Scheduled)
+
+on:
+  schedule:
+    - cron: "0 */6 * * *"
+  workflow_dispatch:
+
+jobs:
+  insert-test-runs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: pip install requests
+
+      - name: Debug step
+        run: echo "Workflow is running successfully"
--- a/.github/workflows/insert-test-runs.yml
+++ b/.github/workflows/insert-test-runs.yml
@@ -1,4 +1,4 @@
-name: Insert Test Runs into DB (Scheduled)
+name: Insert Test Runs (cron)
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


akayiz:
Hi @pks2906!

As I mentioned in #10744, @srbarrios took a quick swing at this in #10937, I suggest you review his PR and let us know if you spot any bugs, and if it has any, based on their severity, you can choose either to continue with the approach he took or check the other approach I suggested in #10744.

Also, it is standard that with GH Actions workflows, you finish the whole workflow, test it on a fork as much as possible, open a PR, have people review it, and if it looks good we merge it, so I suggest you finish the whole workflow in a single PR, rather than splitting it into multiple PRs as you mentioned in the description.

Thank you again for taking the time to help with this :)

pks2906:
Hi @akayiz,

Thank you for the detailed explanation, this really helped clarify the issue.

You’re absolutely right about the `run-id` problem, I’m currently using `${{ github.run_id }}`, which refers to the scheduled workflow run itself, not the individual PR test runs. I see how this would result in incorrect data being inserted.

I’ll update the approach to fetch the actual test run IDs associated with each PR and pass those instead.

I’ll also take a closer look at #10937 and the backfilling approach you mentioned, as it seems like a more robust and scalable solution.

I’ll work on refining the implementation and update the PR shortly.

Thanks again for the guidance

pks2906:
Hi @akayiz,

I’ve updated the workflow implementation based on our discussion.

Currently, I am:
- Fetching recent workflow runs triggered by pull requests
- Extracting the corresponding PR number and run ID
- Passing those values into the `insert_test_run_into_db.py` script

I understand the concern regarding `${{ github.run_id }}` and have adjusted the approach to use the actual run IDs from the GitHub API instead.

At the moment, this implementation is focused on validating the overall workflow and data flow. I will next look into:

- Handling multiple test runs per PR correctly
- Improving robustness (pagination, filtering, etc.)
- Evaluating the backfilling approach mentioned in #10744 as a potentially better long-term solution

Additionally, I noticed that some workflow runs may not be associated with a PR (e.g., edge cases where `pull_requests` is empty). I can handle this case explicitly by filtering out such runs to avoid incorrect data insertion.

Please let me know if this direction aligns with what you had in mind, or if you’d prefer I pivot to the backfilling approach directly.

Thanks again for your guidance

akayiz:
Hi, sorry again about the late reply, I’m currently busy with my graduation project at university, we have an important milestone coming up, along with midterm exams next week. Also, I’m not currently working at Uyuni, so please bear with me 😅

I checked what you did, there is some stuff that's missing, and with all honesty, I believe even if we fix them, it will be error-prone, some of the concerns that come to mind:

- The `insert_test_run_into_db.py` needs more inputs: `commit-sha`, `modified-files`, and `cucumber-reports-path`
- We don't insert any workflow run that was caused by a PR, we focus only on the test workflow runs, currently named "TestFlow"
- I believe previously I have had some issues with the `per_page` GH API parameter where I would miss some PRs if there is a lot of them.

If I were to work on this now, I’d probably go with the backfilling approach I mentioned earlier. The scripts are already working and well-tested, so adapting them into a periodic GitHub Action would likely be a lot less error-prone. It would basically just run the following steps in a workflow:

backfilling

1. `pr_data_extraction.py <current_date - periodic_time>`
2. `runs_feature_result_extraction.py`
3. `backfill_test_runs_into_db.py`

You’d still need to test this properly. I suggest creating a blank database (e.g. using [Neon](https://neon.com/)), applying the [postgres_schema](https://github.com/uyuni-project/uyuni/blob/master/testsuite/ext-tools/predictive-test-selection/postgres_schema.sql), and testing the workflow on a fork before requesting it to be merged. The goal is to make sure it correctly inserts all test runs within the given time window and doesn’t miss anything.

If this feels like too much or gets confusing, that’s totally fine, feel free to drop it and I will take care of it later :)

PR #11407, opened 2026-01-21
Use 5.2 proxy

Bischoff:
## What does this PR change?

Use 5.2 proxy now that the product exists in SCC.
Base for 5.2 proxy is now SL micro 6.2.
It can also be SLES 15 SP7, but this will be done in a separate PR, after SCC fixed the issues with that combination.

To be merged at same time as https://github.com/SUSE/susemanager-ci/pull/1922 .

Piggyback: SLES 15 SP7 is not beta anymore.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

No documentation needed: only internal and user invisible changes.

- [x] **DONE**

## Test coverage

Cucumber tests were modified.

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28063
No ports, 5.2 (head) only.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "frontend_checks"
- [ ] Re-run test "spacecmd_unittests"

PR #11664, opened 2026-03-23
BV: Reposync in case of a SL micro 6.1 minion

Bischoff:
## What does this PR change?

Fixup of https://github.com/uyuni-project/uyuni/pull/11407

Head BV: Reposync in case of a SL micro 6.1 minion

Piggyback: also fixed the synchronization of the tools for each version of SLE / SL micro, both in MLM and Uyuni.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

No documentation needed: only internal and user invisible changes.

- [x] **DONE**

## Test coverage

Cucumber tests were fixed.

- [x] **DONE**

## Links

No ports, head only.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
+++ b/testsuite/features/build_validation/reposync/srv_sync_all_products.feature
@@ -196,8 +196,9 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I select "SUSE Linux Enterprise Micro 5.2 x86_64" as a product
     Then I should see the "SUSE Linux Enterprise Micro 5.2 x86_64" selected
     When I open the sub-list of the product "SUSE Linux Enterprise Micro 5.2 x86_64"
-    And I select "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64" as a product
-    Then I should see the "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64" selected
+    And I open the sub-list of the product "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64"
+    And I select "SUSE Multi-Linux Manager Beta Client Tools for SLE Micro 5 x86_64 (BETA)" as a product
+    Then I should see the "SUSE Multi-Linux Manager Beta Client Tools for SLE Micro 5 x86_64 (BETA)" selected
     When I click the Add Product button
     And I wait until I see "Selected channels/products were scheduled successfully for syncing." text
     And I wait until I see "SUSE Linux Enterprise Micro 5.2 x86_64" product has been added
@@ -235,8 +236,9 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I select "SUSE Linux Enterprise Micro 5.3 x86_64" as a product
     Then I should see the "SUSE Linux Enterprise Micro 5.3 x86_64" selected
     When I open the sub-list of the product "SUSE Linux Enterprise Micro 5.3 x86_64"
-    And I select "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64" as a product
-    Then I should see the "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64" selected
+    And I open the sub-list of the product "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64"
+    And I select "SUSE Multi-Linux Manager Beta Client Tools for SLE Micro 5 x86_64 (BETA)" as a product
+    Then I should see the "SUSE Multi-Linux Manager Beta Client Tools for SLE Micro 5 x86_64 (BETA)" selected
     When I click the Add Product button
     And I wait until I see "Selected channels/products were scheduled successfully for syncing." text
     And I wait until I see "SUSE Linux Enterprise Micro 5.3 x86_64" product has been added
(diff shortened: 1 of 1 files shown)


testsuite/features/build_validation/reposync/srv_sync_all_products.feature
```diff
@@ -380,6 +380,21 @@ Feature: Synchronize products in the products page of the Setup Wizard
     When I use spacewalk-common-channel to add channel "sl-micro-6.0-devel-uyuni-client" with arch "x86_64"
     And I wait until the channel "sl-micro-6.0-devel-uyuni-client-x86_64" has been synced
 
+@susemanager
+@slmicro61_minion
+  Scenario: Add SUSE Linux Micro 6.1
```
alichiel, line 390: I can see the very same scenario below with only difference of @uyuni tag. What is the other difference here please?

srbarrios: In other OSs, we also assure that "SUSE Multi-Linux Manager Client Tools" are selected. This particular step differs from MLM to Uyuni, as on Uyuni we synchronize it using `spacewalk-common-channel`.

MLM:
```
When I open the sub-list of the product "SUSE Linux Enterprise xxx x86_64"
And I open the sub-list of the product "SUSE Multi-Linux Manager Client Tools for xxx x86_64" if present
```

Uyuni:
```
When I use spacewalk-common-channel to add channel "xxx-devel-uyuni-client" with arch "x86_64"
And I wait until the channel "xxx-devel-uyuni-client-x86_64" has been synced
```

Why we don't have this step on SL Micro 6.1, I don't know.

Bischoff: @alichiel 

There are no other differences and we could even group them and remove the tags.

But they are all like this. Look for example at SL micro 6.0.

I can imagine there are reasons for this:
 * have an homogeneus structure for all minions ;
 * maybe the code for uyuni should be different. @maximenoel8 told me this morning the Uyuni cases may be incorrect.
 
 In the doubt I keep it like this.

Bischoff: @srbarrios 

> ```
> When I use spacewalk-common-channel to add channel "xxx-devel-uyuni-client" with arch "x86_64"
> And I wait until the channel "xxx-devel-uyuni-client-x86_64" has been synced
> ```
> 
> Why we don't have this step on SL Micro 6.1, I don't know.

??? we do have it, a few lines below:
```
@uyuni
@slmicro61_minion
  Scenario: Add SUSE Linux Micro 6.1 Uyuni Client tools
    When I use spacewalk-common-channel to add channel "sl-micro-6.1-devel-uyuni-client" with arch "x86_64"
    And I wait until the channel "sl-micro-6.1-devel-uyuni-client-x86_64" has been synced
```

alichiel: @Bischoff, in this case I would group them since there is no other difference. If the uyuni fails, we should fix the uyuni part (in PR related to uyuni change).

srbarrios: > There are no other differences and we could even group them and remove the tags.

Recording in our private slack about why we need to have differences between MLM and Uyuni: https://suse.slack.com/files/U02DQHJ4PSL/F0AN75T51KQ/screenshare_-_2026-03-23_2_56_12_pm.mp4

Bischoff: After verification, tools selection was wrong both for MLM and uyuni cases starting with SLE Micro 5.2.

This is fixed in a second commit.

alichiel:
I haven't tested but now it looks good. Approving.

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
```diff
@@ -196,8 +196,9 @@ Feature: Synchronize products in the products page of the Setup Wizard
     And I select "SUSE Linux Enterprise Micro 5.2 x86_64" as a product
     Then I should see the "SUSE Linux Enterprise Micro 5.2 x86_64" selected
     When I open the sub-list of the product "SUSE Linux Enterprise Micro 5.2 x86_64"
-    And I select "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64" as a product
-    Then I should see the "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64" selected
+    And I open the sub-list of the product "SUSE Multi-Linux Manager Client Tools for SLE Micro 5 x86_64"
```
srbarrios, line 199: Did it work well without (BETA) as part of this text?

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
srbarrios: Thank you

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
Bischoff: Yes.

"Beta" and "(BETA)" texts are NOT part of the sub-list name.
"Beta" and "(BETA)" texts ARE part of the checkbox name.

See screenshot

Image

testsuite/features/build_validation/reposync/srv_sync_all_products.feature
Bischoff: Welcome. I still need to understand why we had so much missing code for SL micro.

I am also not completely confident that those problems do not also affect SLES systems.

But I did not want to divert this PR too much from its original purpose.

srbarrios: Make sense 👍🏼

PR #11673, opened 2026-03-24
testsuite: hibernate tests added, Internal Server Error check added

alichiel:
## hibernate tests

Adding hibernate tests
- simple server **exception** log check 
- test case for BSC#1259499 

Additional change is for the function `check_text_and_catch_request_timeout_popup?` to catch the Internal Server Error

### DIFF
--- a/testsuite/features/secondary/srv_hibernate.feature
+++ b/testsuite/features/secondary/srv_hibernate.feature
@@ -2,6 +2,9 @@
 # Licensed under the terms of the MIT license.
 
 Feature: Hibernate
+  Hibernate is a bridge between WebUI and database,
+  so the CRUD operations should be triggered
+  in order to test hibernate function.
 
   Scenario: Log in as org admin user
     Given I am authorized


testsuite/features/secondary/srv_hibernate.feature
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Hibernate
+
```
srbarrios, line 8: Could you add the acceptance criteria please?

alichiel:
> Do you think the scenarios on this test could be covered by some other test running in core stage? Did you have in mind other kind of actions that could trigger a particular behavior to validate regarding the hibernate implementation?

From the hibernate perspective, the tool is kind of bridge and to test is is basically about "going thru the page and trigger db actions", like in this example.
I added additional log check as well.
All of this is initial, definitely not extensive and I believe there are other scenarios that should test hibernate.

testsuite/features/secondary/srv_hibernate.feature
alichiel: Will do.

alichiel:
> So from what I see this creates a test channel and repo with localhost url and then it deletes them. I think we do cover the case in other channel and repo management features. Can you please look if we have something duplicated? Maybe we can adjust the others?

This is for another reason, specific scenario that triggers the hibernate bugs. But of course, hibernate is being tested via various actions that triggers CRUD in DB. I have just chosen the feature that is specifically related to hibernate.

srbarrios:
To expand on potential scenarios: I think adding a scenario running the 'Dry Run' command would be very valuable, you can find it for instance while performing a product migration. Since dry runs trigger transactional processes that must eventually roll back, they could potentially catch Hibernate-specific issues that standard CRUD operations might miss. It would be great to get the developers' perspective on this.

testsuite/features/secondary/srv_hibernate.feature
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+Feature: Hibernate
```
Bischoff, line 4: Sorry to nitpick, but...

"To hibernate" is not a feature of our product.

Our users don't want to hibernate.

"Hibernate" is a technical, internal, component: a library.

Name this feature "crud_operations_on_channel_and_repository" or anything else that relates to what it *really* does.

You can put in a comment that these tests _might_ trigger hibernate problems.

alichiel: OK, from that perspective, it makes sense. I'll reconsider that.

alichiel:
After the discussion with the team, this one will land in different existing area (channels and repos), and the specifics like "hibernate" will be mentioned there. Also to consider scoping.
Thanks everyone for the participation!

PR #11681, opened 2026-03-26
Fix User x Address relationship (bsc#1259951)

rjpmestre:
## What does this PR change?

This PR fixes the User x Address relationship that resulted in issues associating addresses to users

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30099

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 SUSE LLC
+ * Copyright (c) 2025--2026 SUSE LLC
  * Copyright (c) 2009--2014 Red Hat, Inc.
  *
  * This software is licensed to you under the GNU General Public License,
@@ -16,12 +16,15 @@
 package com.redhat.rhn.domain.user;
 
 import com.redhat.rhn.domain.BaseDomainHelper;
+import com.redhat.rhn.domain.user.legacy.UserImpl;
 
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
 import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.GenerationType;
 import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
 import jakarta.persistence.SequenceGenerator;
 import jakarta.persistence.Table;
 
@@ -56,6 +59,10 @@ public class AddressImpl extends BaseDomainHelper implements Address {
     private String isPoBox;
     @Column(name = "type")
     private String privType;
+
+    @ManyToOne
+    @JoinColumn(name = "web_user_id")
+    private UserImpl user;
 
     /**
      * Protect the constructor
@@ -260,6 +267,21 @@ public class AddressImpl extends BaseDomainHelper implements Address {
             return Address.TYPE_MARKETING;
         }
         return privType;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public User getUser() {
+        return user;
+    }
+
+
+    /**
+     * {@inheritDoc}
+     */
+    public void setUser(User userIn) {
+        user = (UserImpl) userIn;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -36,232 +41,167 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
-    private String isPoBox;
+    private String isPoBox = "0";
+
     @Column(name = "type")
-    private String privType;
+    private String privType = Address.TYPE_MARKETING;
+
+    @ManyToOne
+    @JoinColumn(name = "web_user_id")
+    private UserImpl user;
```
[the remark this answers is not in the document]
rjpmestre: No problem. I don’t see a scenario where we retrieve an address and would want to avoid automatically fetching its user. As far as I know, there are no bulk operations on addresses or “find all” use cases.

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -36,232 +41,167 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
     public void setIsPoBox(String isPoBoxIn) {
-        this.isPoBox = isPoBoxIn;
+        isPoBox = isPoBoxIn;
     }
 
-    /**
-     * {@inheritDoc}
-     */
     @Override
     public String getType() {
```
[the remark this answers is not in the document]
rjpmestre: I set the initial value as `Address.TYPE_MARKETING`. If some process sets it as something else/null, it should handle it.

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -36,232 +41,167 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
     }
 
-    /**
-     * {@inheritDoc}
-     */
     @Override
     public String getAddress1() {
-        if (address1 == null) {
-            address1 = " ";
-        }
```
CDellaGiusta, line 90: What if someone or some sort of mistake or combination calls setAddress1(null)? This way you're discarding this kind of "protection" against the null value.  I have no idea why it was there in the first place, or if it is still needed now. The question is: is this needed? Or do we have to play "on the safe side"? Answering to this question lets you leave the code as it is or to restore the null check. No opinion about this. Ditto for the remaining members below

rjpmestre: You're right, I understand your concern as it does not help fixing the bug.
But will be an improvement. I'll add it (and explain it) on another PR.

CDellaGiusta: it's ok to leave it there. If it's an improvement and you see it as redundant code, there's no need for another PR, you can leave it here as it is

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
```diff
@@ -1351,7 +1356,9 @@ public void setTimeZone(RhnTimeZone timeZoneIn) {
         @Override
         public void setAddress(Address addressIn) {
             addresses.clear();
-            addresses.add((AddressImpl)addressIn);
+            AddressImpl address = (AddressImpl) addressIn;
+            address.setUser(UserImpl.this);
```
mackdk, line 1360: I think this should just be `this`, like in the code above
```suggestion
            address.setUser(this);
```

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
```diff
@@ -1386,7 +1393,9 @@ public Address getAddress() {
                     addr.setState(baddr.getState());
                     addr.setZip(baddr.getZip());
                 }
-                addresses.add((AddressImpl)addr);
+                AddressImpl address = (AddressImpl) addr;
+                address.setUser(UserImpl.this);
```
mackdk, line 1397: Same here:
```suggestion
                address.setUser(this);
```

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
rjpmestre: Using `this` here will refer to current `EnterpriseUserImpl` instance. We need the enclosing/outer `UserImpl`

PR #11691, opened 2026-03-27
Generate credentials for reporting service user

witekest:
## What does this PR change?

If reporting datasource is enabled in Grafana formula a dedicated service user is created and credentials saved in the pillar.

## Codespace

## GUI diff

No difference.

## Documentation
- Documentation issue was created: https://github.com/SUSE/spacewalk/issues/30174

- [x] **DONE**

## Test coverage
- No tests:

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28572

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/suse/manager/webui/services/pillar/MinionReportDBPillarGenerator.java
+++ b/java/core/src/main/java/com/suse/manager/webui/services/pillar/MinionReportDBPillarGenerator.java
@@ -51,8 +51,8 @@ public class MinionReportDBPillarGenerator extends MinionPillarGeneratorBase {
         String reportDBUser = "grafana_" + minion.getId();
         String reportDBPass = RandomStringUtils.secure().nextAlphanumeric(12);
         ConnectionManager localRcm = ConnectionManagerFactory.localReportingConnectionManager();
+        ReportDbHibernateFactory localRh = new ReportDbHibernateFactory(localRcm);
         try {
-            ReportDbHibernateFactory localRh = new ReportDbHibernateFactory(localRcm);
             ReportDBHelper dbHelper = ReportDBHelper.INSTANCE;
 
             pillar.add("reportdb_name", reportDBName);
@@ -64,13 +64,17 @@ public class MinionReportDBPillarGenerator extends MinionPillarGeneratorBase {
             else {
                 dbHelper.createDBUser(localRh.getSession(), reportDBName, reportDBUser, reportDBPass);
             }
-            localRcm.commitTransaction();
+            localRh.commitTransaction();
         }
         catch (Exception e) {
             LOG.error("Setting user/password failed", e);
             pillar.getPillar().clear();
             minion.getPillars().remove(pillar);
-            localRcm.rollbackTransaction();
+            localRh.rollbackTransaction();
+        }
+        finally {
+            localRh.closeSession();
+            localRh.closeSessionFactory();
         }
 
         return Optional.of(pillar);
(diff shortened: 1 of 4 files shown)


java/core/src/main/java/com/suse/manager/webui/services/pillar/MinionReportDBPillarGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public Optional<Pillar> generatePillarData(MinionServer minion) {
+        Pillar pillar = minion.getPillarByCategory(CATEGORY).orElseGet(() -> {
+            Pillar newPillar = new Pillar(CATEGORY, new HashMap<>(), minion);
+            minion.getPillars().add(newPillar);
+            return newPillar;
+        });
+        pillar.getPillar().clear();
+        String reportDBName = Config.get().getString(ConfigDefaults.REPORT_DB_NAME, "reportdb");
+        String reportDBUser = "repotdb" + minion.getId();
```
mcalmer: Better name the user by the client service which use it.
What about "grafana_ + id" ?


```suggestion
        String reportDBUser = "grafana_" + minion.getId();
```

witekest: Good point.

java/core/src/main/java/com/suse/manager/webui/services/pillar/MinionReportDBPillarGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+    public Optional<Pillar> generatePillarData(MinionServer minion) {
+        Pillar pillar = minion.getPillarByCategory(CATEGORY).orElseGet(() -> {
+            Pillar newPillar = new Pillar(CATEGORY, new HashMap<>(), minion);
+            minion.getPillars().add(newPillar);
+            return newPillar;
+        });
+        pillar.getPillar().clear();
+        String reportDBName = Config.get().getString(ConfigDefaults.REPORT_DB_NAME, "reportdb");
+        String reportDBUser = "grafana_" + minion.getId();
+        String reportDBPass = RandomStringUtils.secure().nextAlphanumeric(12);
```
mcalmer: It would not hurt to make the password a bit more secure :-)
Maybe use `20` or `24` ?

mcalmer (dismissed):
Maybe you want to change the password length but I do not want to block.

Another thing you can think about to improve this in the future:
This change the password every time when the pillar gets re-generated and the password is immediately changed in the DB. But only after the user run a state.apply for the service, the new password can be used by grafana. There is a time gap which can be very long when the pillar were re-generated globally but no highstate was executed.

Possible improvement for the future:
Keep the password until a change is explicitly requested.

java/core/src/main/java/com/suse/manager/webui/services/pillar/MinionReportDBPillarGenerator.java
witekest: Done

mcalmer (dismissed):

witekest:
> Another thing you can think about to improve this in the future: This change the password every time when the pillar gets re-generated and the password is immediately changed in the DB. But only after the user run a state.apply for the service, the new password can be used by grafana. There is a time gap which can be very long when the pillar were re-generated globally but no highstate was executed.

Thanks, indeed I was generating pillar data too often. I have removed it now from the second `generatePillar` method. Kudos for @aaannz as well for helping to spot it.

And yes, I think I can improve the implementation to generate the account only after the user runs `state.apply`.

mcalmer (approved):

PR #11731, opened 2026-04-04
Fix: correctly re-attach detached filters to CLM project (#11256)

### LINKED ISSUE #11256: Can't re-attach detached filter to CLM project
### Problem description

This was recently observed by an attendee completing lab exercises whilst I was teaching the SUSE Multi-Linux Manager Operations (SMGR201v5) course. The lab used SUSE Manager 5.0.2 but I've subsequently verified with Uyuni 2025.07 and 2025.10, the latter being the version I'm now reporting against.

The exercise involved detaching a filter from a CLM project and attaching a different filter but they mistakenly detached the wrong filter. They were then unable to re-attach the mistakenly detached filter. Whilst the detach was saved they had not (re)built the project. Attempts to re-attach the mistakenly detached filter resulted in no error, it just wasn't actioned.

The expectation was that the filter would be re-attached to the CLM project.

### Steps to reproduce

1. Create a new CLM project
2. Add a new environment
3. Attach any base channel as a source
4. Attach a new filter and Save
5. Build the project
6. Detach the attached filter and Save
7. Re-attach the detached filter and Save

It won't be re-attached ...

### Uyuni version

```bash
uyuni-container:~ # mgrctl exec "rpm -qa | grep -i uyuni" | grep release
release-notes-uyuni-2025.10-241000.2.1.uyuni5.noarch

Note Uyuni-Server-release package not installed either in host OS or Uyuni server container
```

### Uyuni proxy version (if used)

```bash
N/A
```

### uyuni-tools version used to deploy/manage the server or proxy (if relevant)

```bash
N/A
```

### Useful logs

```bash
N/A
```

### Additional information

Workaround is to recreate the detached filter but with a different name and then attach that.

Sahitya0805:
### Summary
This Pull Request resolves a bug in the Content Lifecycle Management (CLM) system where filters in a `DETACHED` state could not be re-attached to a project via the WebUI. 

### Changes
- **FilterApiController**: Updated the filter comparison logic to utilize `getActiveFilters()` instead of checking the entire filter collection. This ensuring that previously detached filters are correctly identified as needing re-attachment.
- **Java Compatibility**: Replaced `.toList()` with `.collect(Collectors.toList())` to ensure backward compatibility with Java 11.
- **ContentManager**: Added an explicit Hibernate session flush (`HibernateFactory.getSession().flush()`) during filter attachment and detachment to synchronize the relationship state with the database immediately.

### Verification
- Manually verified the logic for state transitions in `ContentProject.java`.
- Cross-referenced with existing domain-level tests in `ContentManagerTest.java`.

Fixes #11256.

### DIFF
--- a/java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/handlers/FilterApiController.java
+++ b/java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/handlers/FilterApiController.java
@@ -53,10 +53,13 @@ import org.apache.logging.log4j.Logger;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import spark.Request;
 import spark.Response;
@@ -112,11 +115,18 @@ public class FilterApiController {
                                 new NoSuchElementException("No content project found with label: " + projectLabel));
 
 
-        List<Long> filterIdsToDetach = dbContentProject.getProjectFilters()
+        // Use only active filters to avoid treating previously detached filters as still attached
+        Set<Long> activeFilterIdsSet = dbContentProject.getActiveFilters()
                 .stream()
-                .map(pf -> pf.getFilter().getId())
-                .filter(filterId -> !filtersIdToUpdate.contains(filterId))
-                .toList();
+                .map(ContentFilter::getId)
+                .collect(Collectors.toSet());
+
+        Set<Long> filtersIdToUpdateSet = new HashSet<>(filtersIdToUpdate);
+
+        List<Long> filterIdsToDetach = activeFilterIdsSet
+                .stream()
+                .filter(filterId -> !filtersIdToUpdateSet.contains(filterId))
+                .collect(Collectors.toList());
         filterIdsToDetach.forEach(filterId -> CONTENT_MGR.detachFilter(
                 projectLabel,
                 filterId,
@@ -125,12 +135,9 @@ public class FilterApiController {
 
         List<Long> filterIdsToAttach = filtersIdToUpdate
                 .stream()
-                .filter(filterId ->
-                        dbContentProject.getProjectFilters()
-                                .stream()
-                                .noneMatch(pf -> pf.getFilter().getId().equals(filterId))
-                )
-                .toList();
+                .distinct()
+                .filter(filterId -> !activeFilterIdsSet.contains(filterId))
+                .collect(Collectors.toList());
         filterIdsToAttach
                 .forEach(filterId ->
                         CONTENT_MGR.attachFilter(
--- a/java/spacewalk-java.changes.sahitya.clm-filter-reattach
+++ b/java/spacewalk-java.changes.sahitya.clm-filter-reattach
@@ -1,0 +1,2 @@
+- Fixed an issue where previously detached filters could
+  not be re-attached to a CLM project.


java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/handlers/FilterApiController.java
```diff
@@ -112,11 +113,16 @@ public static String updateFiltersOfProject(Request req, Response res, User user
...
+        // Use only active filters to avoid treating previously detached filters as still attached
+        List<Long> activeFilterIds = dbContentProject.getActiveFilters()
+                .stream()
+                .map(ContentFilter::getId)
+                .collect(Collectors.toList());
+
+        List<Long> filterIdsToDetach = activeFilterIds
                 .stream()
-                .map(pf -> pf.getFilter().getId())
                 .filter(filterId -> !filtersIdToUpdate.contains(filterId))
```
rjmateus: I think this filter can be added to the previous stream, and avoid going through it 2 time.

java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/handlers/FilterApiController.java
```diff
@@ -112,11 +113,16 @@ public static String updateFiltersOfProject(Request req, Response res, User user
...
-                .toList();
+                .collect(Collectors.toList());
```
rjmateus, line 129: The project is being built with java 17. Why change the working code to make it work with java 11?

java/core/src/main/java/com/redhat/rhn/manager/contentmgmt/ContentManager.java
```diff
@@ -582,6 +582,7 @@ public ContentFilter attachFilter(String projectLabel, Long filterId, User user)
                 .orElseThrow(() -> new EntityNotExistsException(ContentFilter.class, filterId));
         project.attachFilter(filter);
         ContentProjectFactory.save(project);
+        HibernateFactory.getSession().flush();
```
rjmateus: Could you elaborate on why this flush is needed?

java/core/src/main/java/com/redhat/rhn/manager/contentmgmt/ContentManager.java
```diff
@@ -600,6 +601,7 @@ public void detachFilter(String projectLabel, Long filterId, User user) {
                 .orElseThrow(() -> new EntityNotExistsException(ContentFilter.class, filterId));
         project.detachFilter(filter);
         ContentProjectFactory.save(project);
+        HibernateFactory.getSession().flush();
```
rjmateus: Could you elaborate on why this flush is needed?

rjmateus (changes requested):

changelog/11256.bugfix
```diff
@@ -0,0 +1 @@
+Fixed an issue where previously detached filters could not be re-attached to a CLM project.
```
cbbayburt: Changelog additions must be in a specific directory with a specific filename so our CI can pick them up. Please check out this section in the Wiki to see where to put it: https://github.com/uyuni-project/uyuni/wiki/Contributing#changelog-rules-for-packages-in-git

TL;DR: For changes in Java code, this entry must go to a new file with the following name:

`java/spacewalk-java.changes.<your_name>.<feature_name>`

java/core/src/main/java/com/redhat/rhn/manager/contentmgmt/ContentManager.java
```diff
@@ -583,6 +584,9 @@ public ContentFilter attachFilter(String projectLabel, Long filterId, User user)
                 .orElseThrow(() -> new EntityNotExistsException(ContentFilter.class, filterId));
         project.attachFilter(filter);
         ContentProjectFactory.save(project);
+        if (HibernateFactory.getSession().getFlushMode().equals(FlushModeType.COMMIT)) {
+            HibernateFactory.getSession().flush();
+        }
```
cbbayburt: The `getSession()` call creates a transaction, and the framework commits the transaction automatically when the request ends. So in most places, flush is not explicitly required. For your fix, you shouldn't need a `flush` here at all.

I guess Copilot couldn't pick this correctly.

The same applies to the other instance below.

cbbayburt (changes requested):

changelog/11256.bugfix
Sahitya0805: Thanks for pointing this out! I've moved the changelog entry to the required location using the java/spacewalk-java.changes.<name>.<feature> naming convention. Please let me know if there's anything else I should adjust.

java/core/src/main/java/com/redhat/rhn/manager/contentmgmt/ContentManager.java
Sahitya0805: Thanks for the clarification. I removed the explicit flush() calls from both attachFilter and detachFilter, since the transaction lifecycle already handles persistence. I verified that the original re-attach scenario still works without the manual flush, so the fix now relies only on the updated active-filter logic.

cbbayburt (changes requested):
Thanks for the changes, @Sahitya0805.

Meanwhile, I think changes to [.vscode/settings.json](https://github.com/uyuni-project/uyuni/pull/11731/changes#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357) was added by mistake.

Please remove it from the PR, and rebase your branch on top of current `master` so the checks run successfully.

Sahitya0805:
Thanks for the review! I removed the accidentally included .vscode/settings.json from the PR and force-pushed the updated branch. The branch is also up to date with the current master. Thanks again!

cbbayburt (dismissed):
Thanks, @Sahitya0805. It looks good to me now.

Currently, the `master` branch is locked for maintenance work, but should be open again in a week or so. Once it's open and @rjmateus also approves the PR, we can go ahead and merge it.

cbbayburt:
Oh, and meanwhile, please address the failing Changelogs and Checkstyle checks. Seems to be some simple formatting issues. The other failures seem to be CI problems; we can try re-running them later.

PR #11737, opened 2026-04-07
SonarCloud error reduction fix, miscellaneous 14 rules

CDellaGiusta:
## What does this PR change?

SonarCloud fix: java:S2159 Comparisons between unrelated types always return false
SonarCloud fix: java:S1171 Only static class initializers should be used
SonarCloud fix: java:S2386 Mutable fields should not be public static
SonarCloud fix: java:S3923 All branches should not have the same implementation
SonarCloud fix: java:S3252 Static base class members should not be accessed via derived types
SonarCloud fix: java:S3066 Enum fields should not be publicly mutable
SonarCloud fix: java:S3864 implementation is free to skip calls to peek() for optimization purposes
SonarCloud fix: java:S1161 Override should be used on overriding and implementing methods
SonarCloud fix: java:S2153 Unnecessary boxing and unboxing should be avoided
SonarCloud fix: java:S6213 Restricted Identifiers should not be used as identifiers
SonarCloud fix: java:S4165 Assignments should not be redundant
SonarCloud fix: java:S4838 Change Object to the type handled by the collection
SonarCloud fix: java:S2119 Random objects should be reused
SonarCloud fix: javabugs:S2259 Fix access on a value that can be null

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Port(s): 
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
@@ -49,21 +49,23 @@ public class ContentPropertiesValidator {
         if (StringUtils.isEmpty(label)) {
             result.addFieldError("label", "contentmanagement.label_required");
         }
-
-        if (!isLabelValid(label)) {
-            result.addFieldError("label", "contentmanagement.label_invalid");
+        else {
+            if (label.length() > 24) {
+                result.addFieldError("label", "contentmanagement.project_label_too_long");
+            }
         }
 
-        if (label.length() > 24) {
-            result.addFieldError("label", "contentmanagement.project_label_too_long");
+        if ((null != label) && (!isLabelValid(label))) {
+            result.addFieldError("label", "contentmanagement.label_invalid");
         }
 
         if (StringUtils.isEmpty(name)) {
             result.addFieldError("name", "contentmanagement.name_required");
         }
-
-        if (name.length() > 128) {
-            result.addFieldError("name", "contentmanagement.project_name_too_long");
+        else {
+            if (name.length() > 128) {
+                result.addFieldError("name", "contentmanagement.project_name_too_long");
+            }
         }
 
         ContentManager.lookupProjectByNameAndOrg(name, user).ifPresent(cp -> {
@@ -90,21 +92,23 @@ public class ContentPropertiesValidator {
         if (StringUtils.isEmpty(label)) {
             result.addFieldError("label", "contentmanagement.label_required");
         }
+        else {
+            if (label.length() > 16) {
+                result.addFieldError("label", "contentmanagement.environment_lbl_too_long");
+            }
+        }
 
         if (StringUtils.isEmpty(name)) {
             result.addFieldError("name", "contentmanagement.name_required");
         }
+        else {
+            if (name.length() > 128) {
+                result.addFieldError("name", "contentmanagement.environment_name_too_long");
+            }
+        }
 
-        if (!isLabelValid(label)) {
+        if ((null != label) && (!isLabelValid(label))) {
             result.addFieldError("label", "contentmanagement.label_invalid");
-        }
-
-        if (label.length() > 16) {
-            result.addFieldError("label", "contentmanagement.environment_lbl_too_long");
-        }
-
-        if (name.length() > 128) {
-            result.addFieldError("name", "contentmanagement.environment_name_too_long");
         }
 
         if (result.hasErrors()) {
@@ -125,7 +129,7 @@ public class ContentPropertiesValidator {
             result.addFieldError("filter_name", "contentmanagement.name_required");
         }
 
-        if (name.length() > 128) {
+        if ((null != name) && (name.length() > 128)) {
             result.addFieldError("filter_name", "contentmanagement.filter_name_too_long");
         }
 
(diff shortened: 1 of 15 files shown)


java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/admin/monitoring/AdminMonitoringHandler.java
```diff
@@ -36,13 +36,11 @@
...
-        messageMap = new HashMap<>();
-        messageMap.put("enable", "enable_again_to_sync_config");
-        messageMap.put("disable", "disable_again_to_sync_config");
-        messageMap.put("restart", "restart_needed");
-    }
+    private static Map<String, String> messageMap = Map.ofEntries(
+            new AbstractMap.SimpleEntry<>("enable", "enable_again_to_sync_config"),
+            new AbstractMap.SimpleEntry<>("disable", "disable_again_to_sync_config"),
+            new AbstractMap.SimpleEntry<>("restart", "restart_needed")
+    );
```
mackdk: In general, use `Map.entry()` instead of `new AbstractMap.SimpleEntry<>()` . In this case, you can use directly `Map.of()`, since you have only 6 parameters:

```java
        private static Map<String, String> messageMap = Map.of(
                "enable", "enable_again_to_sync_config",
                "disable", "disable_again_to_sync_config",
                "restart", "restart_needed"
        );

File: java/core/src/main/java/com/redhat/rhn/frontend/action/kickstart/KickstartTreeUpdateType.java [hunk 2]
```diff
@@ -39,14 +39,6 @@ public String getType() {
         return type;
     }
 
-    /**
-     * Set the type
-     * @param updateType the update type to set
-     */
-    public void setType(String updateType) {
-        type = updateType;
```
mackdk, line 47: I guess you can also make the `type` field `final` now.

java/core/src/test/java/com/redhat/rhn/frontend/xmlrpc/system/SystemHandlerProvisioningTest.java
mackdk, line 131: Please also add a call to `super.tearDown()`

java/core/src/main/java/com/redhat/rhn/frontend/action/errata/ErrataConfirmAction.java
```diff
@@ -124,9 +124,9 @@ public ActionForward confirmErrata(ActionMapping mapping,
 
         if (actionChain == null) {
             Action update = ActionManager.createErrataAction(user, user.getOrg(), currentErrata);
-            for (Object systemIn : systems) {
+            for (SystemOverview systemIn : systems) {
                 ActionFactory.addServerToAction(
-                        ((SystemOverview) systemIn).getId(),
+                        systemIn.getId(),
                         update);
             }
```
mackdk: Or even:
```java
            systems.forEach(systemIn -> ActionFactory.addServerToAction(systemIn.getId(), update));
```

java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
```diff
@@ -54,15 +54,15 @@ public static void validateProjectProperties(String label, String name, User use
             result.addFieldError("label", "contentmanagement.label_invalid");
         }
 
-        if (label.length() > 24) {
+        if ((null != label) && (label.length() > 24)) {
```
mackdk: Maybe instead of adding the null check everywhere, we can just add an `else` on the first `isEmpty()` check for both parameters to skip the subsequent checks if there is no value?

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/admin/monitoring/AdminMonitoringHandler.java
CDellaGiusta: I saw this possibility in the docs, but I went for the other option.
Fixed.

java/core/src/main/java/com/redhat/rhn/frontend/action/kickstart/KickstartTreeUpdateType.java
CDellaGiusta: good catch. Fixed

java/core/src/test/java/com/redhat/rhn/frontend/xmlrpc/system/SystemHandlerProvisioningTest.java
CDellaGiusta: fixed

java/core/src/main/java/com/redhat/rhn/frontend/action/errata/ErrataConfirmAction.java
CDellaGiusta: fixed

java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
CDellaGiusta: that was my first thought either. This way you're altering the sequence of the errors, and I wanted to be as conservative as possible. But let's assume the error sequence does not matter.
Fixed

mackdk: From what I see, the string empty case will trigger both `contentmanagement.label_required` and `contentmanagement.label_invalid`, but it is possible that we have some acceptance tests that are checking for all the messages.

CDellaGiusta: Since ContentPropertiesValidator.isLabelValid(null) -> NPE I added a null check before calling it.
BTW I also forgot to correct this in all the places. Now it should be fine

PR #11741, opened 2026-04-08
Fix  cleanup state timeouts (bsc#1258567)

ycedres:
When a user deletes a system from the UI, the operation hangs for exactly 240 seconds (4 minutes). This was caused by the removal of the minion keys in the cleanup_minion state before returning the result of the execution to master.

The cleanup_minion state was refactored to perform an asynchronous clean up by first removing the keys and secondly stopping the salt minion service. This allows the state execution to return.

## What does this PR change?

**add description**

## Codespace

## GUI diff

No difference.

Before:

After:

- [ ] **DONE**

## Documentation

- [ ] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/susemanager-utils/susemanager-sls/salt/cleanup_minion/init.sls
+++ b/susemanager-utils/susemanager-sls/salt/cleanup_minion/init.sls
@@ -32,45 +32,38 @@ mgrchannels_repo_clean_keyring:
     - name: /usr/share/keyrings/mgr-archive-keyring.gpg
 {%- endif %}
 
-mgr_mark_no_longer_managed:
-  file.absent:
-    - name: /etc/sysconfig/rhn/systemid
+{%- if not grains.get('transactional', False) %}
+mgr_async_identity_cleanup:
+  cmd.run:
+    - bg: True
+    - name: |
+        /usr/bin/sleep 1 && \
+        /usr/bin/rm -f /etc/sysconfig/rhn/systemid \
+                       {{ salt_config_dir }}/minion.d/susemanager.conf \
+                       {{ salt_config_dir }}/minion.d/master.conf \
+                       {{ salt_config_dir }}/pki/minion/minion.pem \
+                       {{ salt_config_dir }}/pki/minion/minion.pub \
+                       {{ salt_config_dir }}/pki/minion/minion_master.pub
+{%- endif %}
 
-mgr_remove_salt_config:
-  file.absent:
-    - name: {{ salt_config_dir }}/minion.d/susemanager.conf
-
-mgr_remove_salt_config_altname:
-  file.absent:
-     - name: {{ salt_config_dir }}/minion.d/master.conf
-
-mgr_remove_salt_priv_key:
-  file.absent:
-     - name: {{ salt_config_dir }}/pki/minion/minion.pem
-
-mgr_remove_salt_pub_key:
-  file.absent:
-     - name: {{ salt_config_dir }}/pki/minion/minion.pub
-
-mgr_remove_salt_master_key:
-
-  file.absent:
-     - name: {{ salt_config_dir }}/pki/minion/minion_master.pub
+{%- if not grains.get('transactional', False) %}
+mgr_schedule_salt_minion_stop:
+  schedule.present:
+    - function: cmd.run
+    - seconds: 3
+    - once: True
+    - persist: False
+    - maxrunning: 1
+    - job_args:
+      - |
+        /usr/bin/rm -rf /var/cache/{{ salt_minion_name }} && \
+        /usr/bin/rm -f {{ salt_config_dir }}/minion.d/_schedule.conf && \
+        /usr/bin/systemctl stop {{ salt_minion_name }}
+    - order: last
+{%- endif %}
 
 {%- if salt['pillar.get']('contact_method') not in ['ssh-push', 'ssh-push-tunnel'] %}
 mgr_disable_salt:
   cmd.run:
     - name: /usr/bin/systemctl disable {{ salt_minion_name }}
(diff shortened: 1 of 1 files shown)


susemanager-utils/susemanager-sls/salt/cleanup_minion/init.sls
```diff
@@ -32,45 +32,38 @@ mgrchannels_repo_clean_keyring:
...
+                       {{ salt_config_dir }}/minion.d/susemanager.conf \
+                       {{ salt_config_dir }}/minion.d/master.conf \
+                       {{ salt_config_dir }}/pki/minion/minion.pem \
+                       {{ salt_config_dir }}/pki/minion/minion.pub \
+                       {{ salt_config_dir }}/pki/minion/minion_master.pub
+{%- endif %}
 
-  file.absent:
-     - name: {{ salt_config_dir }}/pki/minion/minion_master.pub
+{%- if not grains.get('transactional', False) %}
```
m-czernek, line 49: Let's merge these two together. 
```suggestion
```

m-czernek (approved):
In my testing against SLM 6.1, I was unable to do anything better than what this PR does in terms of cleanup.

The only difference from the previous state is that we no longer clean files in `/etc/venv-salt-minion/minion.d/`. When I tried to clean them, Salt is unable to respond back to master even when executing it with `cmd.run bg=True name=sleep 10 && rm ...`.

As is, the cleanup disables the Salt service, and when users reboot the system, venv does not come back up. Users are able to re-bootstrap the system even with the files present.

So my thinking is, we can merge this PR as it improves the situation on non-TU minions, and when we have improvements to the way we handle TU minions, i.e. we have control over what is executed inside and outside of a transaction, we should be able to do better (currently, `schedule.present` does not seem to be working when executed within a transaction).

Note that the cleanest solution I found would be to create a one-off systemd timer, like:

```
mgr_async_cleanup_and_stop:
  cmd.run:
    - name: |
        /usr/bin/systemd-run --on-active=3 /bin/sh -c '\
          /usr/bin/rm -f /etc/sysconfig/rhn/systemid \
                         {{ salt_config_dir }}/minion.d/susemanager.conf \
                         {{ salt_config_dir }}/minion.d/master.conf \
                         {{ salt_config_dir }}/pki/minion/minion.pem \
                         {{ salt_config_dir }}/pki/minion/minion.pub \
                         {{ salt_config_dir }}/pki/minion/minion_master.pub \
                         {{ salt_config_dir }}/minion.d/_schedule.conf && \
          /usr/bin/rm -rf /var/cache/{{ salt_minion_name }} && \
          /usr/bin/systemctl stop {{ salt_minion_name }}'
```

But of course that's a no-go for TU minions right now, since we cannot execute that inside of a transaction.

susemanager-utils/susemanager-sls/salt/cleanup_minion/init.sls
vzhestkov: We can't really provide now any state which could work the same way for non-transactional and transactional clients, especially while keeping in mind that we are going to change the way how the state.apply will be triggered after implementing all the improvements we are going to implement. The cleanup is mostly required to remove the pki keys for the next attempts to bootstrap and maybe the config file to prevent registering back to the same server in case of accidental start of the minion service.

For the transactional systems for me looks more logical to have separated copy of the state to avoid mixing the logic inside one single state like: this part for transactional, this one for non-transactional and repeating similar conditions again and again, I would say it will be easier to have one condition which selects which state to apply and whole the content of the state is intended to be fully applied either on top of transactional or non-transactional client.

m-czernek: Sure, but what does the TU-cleanup look like at this point? Adding `schedule.present` doesn't seem to work within a transaction, and adding something like this:

```
mgr_async_identity_cleanup:
  cmd.run:
    - bg: True
    - name: |
        /usr/bin/sleep 10 && \
        /usr/bin/rm -f /etc/sysconfig/rhn/systemid \
                       {{ salt_config_dir }}/minion.d/susemanager.conf \
                       {{ salt_config_dir }}/minion.d/master.conf \
                       {{ salt_config_dir }}/pki/minion/minion.pem \
                       {{ salt_config_dir }}/pki/minion/minion.pub \
                       {{ salt_config_dir }}/pki/minion/minion_master.pub
```

Causes timeouts (so Salt client already does not return a response to master).

ycedres: So IUUC at this point for TU minions cleaning up the config dir causes a timeout, so we do not have a perfect solution for TU yet. And his improves the situation anyway for non-TU minions, right?

m-czernek: Yes, this PR improves the situation on all minions, with the caveat that the files created by MLM stay on TU minions now (at the cost of quick delete operations). So IMO we have three options here:

1. We merge the PR in its current form and accept that the cleanup on TU minions is essentially just disabling the service.
2. We still do the old cleanup on TU minions. This improves the situations on non-TU minions (non-TU minions delete without timeout), but it causes timeout errors in the UI for TU minions (this happens right now as well).
3. We do cleanup on that works on TU minions too (this I couldn't find how to achieve without causing timeouts) .

m-czernek:
Closing in favor of https://github.com/uyuni-project/uyuni/pull/12161

PR #11746, opened 2026-04-09
Fix the ReportDB user deletion script

maximenoel8:
## What does this PR change?

The Cleanup: delete read-only user scenario was passing its deletion step but the user remained in the database.

Root cause: The uyuni-setup-reportdb-user interactive script's delete path connects as the test user via psql to run DROP OWNED BY current_user. This connection fails silently inside  the container (no set -e in the script), leaving the role with owned privileges. The subsequent DROP ROLE then also fails silently, and the script exits 0 regardless.

Fix: Replace the expect + interactive script approach with a direct spacewalk-sql call using admin credentials — the same mechanism already used by the verification steps. Running DROP OWNED BY <user> as the reportdb admin (instead of as the user itself) always succeeds, and DROP ROLE then completes cleanly.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/29433
Port(s): 
 - 5.1: https://github.com/SUSE/spacewalk/pull/30233
 - 5.0: https://github.com/SUSE/spacewalk/pull/30234

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


mcalmer (changes requested):
uyuni-setup-reportdb-user is an official script documented to be used by users.
If the script is not working, this is a product bug and needs a bugreport. 
A workaround in the testuite does not help any user of our product.
The reason for this testvis, to check that this script is working.

maximenoel8:
Bug: https://bugzilla.suse.com/show_bug.cgi?id=1261841

PR #11761, opened 2026-04-13
Add server and proxy installation checks on rke2

ilmanzo:
## What does this PR change?

Add basic post-install health-check for the RKE2 kubernetes cluster resources.
Network reachability and product functions are done by subsequent existing tests

Added some helper methods to reach k8s objects and fixed `remote_node` class to make it work also with `kubectl` managed resources.

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30010

## Changelogs

- [x] No changelog needed


testsuite/features/support/remote_node.rb
```diff
@@ -34,13 +34,16 @@ def initialize(host, port: 22)
 
     $named_nodes[host] = @hostname
     if @host == 'server'
-      _out, _err, code = ssh('which mgrctl', host: @target)
-      @has_mgrctl = code.zero?
```
Pablogoliva, line 38: Why do you remove these lines?. We still have a non-containerized version.

testsuite/features/support/remote_node.rb
```diff
@@ -34,13 +34,16 @@ def initialize(host, port: 22)
...
+    end
+
+    if @host == 'server' && !@has_kubectl
       # Remove /etc/motd inside the container, or any output from run will contain the content of /etc/motd
       run('rm -f /etc/motd && touch /etc/motd')
-      out, _code = run('sed -n \'s/^java.hostname *= *\(.\+\)$/\1/p\' /etc/rhn/rhn.conf')
+      out, code = run('sed -n \'s/^java.hostname *= *\(.\+\)$/\1/p\' /etc/rhn/rhn.conf')
     else
-      out, _err, _code = ssh('hostname -f', host: @target)
+      out, _err, code = ssh('hostname -f', host: @target)
```
Pablogoliva, line 46: Why do you set these variables as public?. I mean, what's the matter of doing this?

testsuite/features/support/env.rb
```diff
@@ -691,6 +692,10 @@ def process_code_coverage
   skip_this_scenario unless $is_containerized_server
 end
 
+Before('@rke2') do
```
Pablogoliva, line 695: This feature is going to be execute in an isolated stage, I'm not sure if that's neccesary.

testsuite/features/support/commonlib.rb
```diff
@@ -33,16 +34,24 @@ def count_table_items
...
+  # If running RKE2, first check inside the pod
+  if running_rke2?
+    pod = get_pod_name('server', 'server')
+    patterns.each do |pattern, name|
+      _out, code = server.run_local("kubectl exec -n uyuni #{pod} -- rpm -q #{pattern}", check_errors: false)
+      return $product = name if code.zero?
+    end
   end
-  _product_raw, code = get_target('server').run('rpm -q patterns-suma_server', check_errors: false)
-  if code.zero?
```
Pablogoliva, line 42: This is necessary for non-containerized versions. Let's keep it until we stop supporting 4.3.

ilmanzo: the check has not been deleted, only moved some lines below

testsuite/features/support/remote_node.rb
ilmanzo: the check has simply changed to the lines below

testsuite/features/support/remote_node.rb
ilmanzo: so that we can use `code` in case of any error
see 
```ruby
 raise StandardError, "No FQDN for '#{@hostname}'. Response code: #{code}" if @full_hostname.empty?
```

btw in Ruby the underscore hasn't any special meaning, it's just a common convention to prefix with _ the variables you are going to declare but not use later.

testsuite/features/kubernetes/srv_rke2_install.feature
```diff
@@ -0,0 +1,16 @@
...
+
+@rke2
+
+Feature: RKE2 Server Deployment
+  In order to manage systems using a Kubernetes installation
+  As the system administrator
+  I want to ensure the server is correctly deployed on RKE2
+
+  Scenario: Ensure server cluster health and persistence
+    Given the first-time setup job is successful
```
Pablogoliva: The cucumber steps usually starts with capital letter.

testsuite/features/support/commonlib.rb
Pablogoliva: Okey. Calrified.

testsuite/features/support/remote_node.rb
Pablogoliva: Clarified.

testsuite/features/support/remote_node.rb
Pablogoliva: Clarified,

testsuite/features/support/env.rb
Pablogoliva: This is not very useful since we are going to run this feature in an isolated stage called kubernetes but are not useless either, let's keep it.

Pablogoliva (approved):

PR #11836, opened 2026-04-27
HardwareMapper rework

rjpmestre:
## What does this PR change?

This PR refactors `HardwareMapper` by extracting mapping logic into dedicated components.

The goal is to improve the overall code organization and readability, making easier to understand and maintain.
These changes aim to **preserve the existing behaviour** while separating the different concerns into more manageable pieces.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/31833
Port(s):

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


java/core/src/main/java/com/suse/manager/reactor/hardware/CpuInfoMapper.java
```diff
@@ -0,0 +1,196 @@
...
+     * Map PPC64 CPU information.
+     */
+    protected void mapPpc64CpuInfo(CPU cpu, ValueMap cpuInfo) {
+        cpu.setModel(CpuFieldTruncator.model(cpuInfo.getValueAsString("cpu")));
+        cpu.setVersion(CpuFieldTruncator.version(cpuInfo, "revision"));
+        cpu.setBogomips(CpuFieldTruncator.bogomips(cpuInfo, "bogomips"));
+        cpu.setVendor(CpuFieldTruncator.vendor(cpuInfo, "machine"));
+        cpu.setMHz(CpuFieldTruncator.mhz(StringUtils.substring(cpuInfo.get("clock")
+                .flatMap(ValueMap::toString)
+                .orElse("-1MHz"), 0, -3))); // remove MHz sufix
```
lucidd: instead of including the MHz here and then removing it again move the `orElse` outside

java/core/src/main/java/com/redhat/rhn/domain/action/HardwareRefreshAction.java
```diff
@@ -130,71 +132,96 @@ public void handleUpdateServerAction(ServerAction serverAction, JsonElement json
...
+                                    result.getDnsFqdns().stream()
+                            ),
+                            result.getCustomFqdns().stream()
+                    ).distinct().collect(Collectors.toList())
+            );
 
-        if (LOG.isDebugEnabled()) {
-            long duration = Duration.between(start, Instant.now()).getSeconds();
-            LOG.debug("Hardware profile updated for minion: {} ({} seconds)", server.getMinionId(), duration);
+            //
```
lucidd: missing comment

java/core/src/main/java/com/suse/manager/reactor/hardware/CpuInfoMapper.java
```diff
@@ -0,0 +1,196 @@
...
+            // Map common CPU information from grains
+            cpu.setNrsocket(grains.getValueAsLong("cpu_sockets").orElse(1L));
+            cpu.setNrCore(grains.getValueAsLong("cpu_cores").orElse(1L));
+            cpu.setNrThread(grains.getValueAsLong("cpu_threads").orElse(1L));
+            // Use our custom grain. Salt has a 'num_cpus' grain but it gives
+            // the number of active CPUs not the total num of CPUs in the system.
+            // On s390x this number of active and actual CPUs can be different.
+            cpu.setNrCPU(grains.getValueAsLong("total_num_cpus").orElse(0L));
+
+            var archSpecs = grains.get("cpu_arch_specs")
```
lucidd, line 108: i think this can use `grains.getMap` instead

rjpmestre: Do you have any suggestion how I could do it? 

`grains.getMap` returns an instance of the `ValueMap` and there is no method to return the underlying value as a `Map`. Am I missing something here?

java/core/src/main/java/com/redhat/rhn/domain/action/HardwareRefreshAction.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.redhat.rhn.domain.action;
```
CDellaGiusta, line 17: General comment: in this commit I see some different logical actions altogether:
- restructuring handleHardwareProfileUpdate
- introducing CpuFieldTruncator
- introducing CpuInfoMapper and modifying HardwareMapper.mapCpuInfo to use CpuInfoMapper
As a suggestion, if you take these steps separately, it can make the review much easier, then you can always squash them together before merging

java/core/src/main/java/com/suse/manager/reactor/hardware/network/NetworkMapper.java
```diff
@@ -0,0 +1,202 @@
...
+            Map<String, Optional<String>> netModules
+    ) {
+
+        NetworkInterface networkInterface = server.getNetworkInterface(name);
+        if (networkInterface == null) {
+            networkInterface = new NetworkInterface();
+
+            networkInterface.setName(name);
+            networkInterface.setServer(server);
+            server.addNetworkInterface(networkInterface);
```
CDellaGiusta, line 165: you're doing these 3 operations only if `networkInterface` is created: did you check you don't need them otherwise?

java/core/src/test/java/com/suse/manager/reactor/hardware/network/NetworkMapperTest.java
```diff
@@ -0,0 +1,450 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Tests for {@link NetworkMapper}
+ */
```
CDellaGiusta, line 45: not sure if these kind of comments are needed, since they should be quite straightforward (otherwise the class name should be changed). I guess it boils down to personal preference, though

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -0,0 +1,522 @@
...
+    private ValueMap props;
+
+    /**
+     * Maps hardware devices from udevdb.
+     *
+     * @param udevdb the udev database entries
+     * @param dbdev the device entry to process
+     * @return Optional error message if mapping failed
+     */
+    @SuppressWarnings("unchecked")
```
CDellaGiusta, line 153: what is this annotation about?

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -0,0 +1,522 @@
...
+    private static final Logger LOG = LogManager.getLogger(DeviceMapper.class);
+
+    public static final String BLOCK = "block";
+    public static final String PCI = "pci";
+    public static final String PCI_FORMAT = "%s|%s";
+    public static final String SCSI = "scsi";
+    public static final String USB = "usb";
+    public static final String USB_DEVICE_FORMAT = "%04x";
+
+    private static final Set<String> SCANNABLE_DEVICES_SUBSYSTEM = Set.of(PCI, USB, BLOCK, "ccw", SCSI);
```
CDellaGiusta: why not adding "ccw" as a constant string as well?

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -0,0 +1,522 @@
...
+
+        return null;
+    }
+
+    /**
+     * Classify device class from udev data.
+     * @param dbdev the udev device entry
+     * @return device class or null if it cannot be determined
+     */
+    @SuppressWarnings({"java:S131", "unchecked"})
```
CDellaGiusta: SonarCloud issues should be whether addressed, or if they are false positives or for any other reason, dismissed in SonarCloud directly (also adding some brief explanation). I've been  told we decided this way to avoid pollluting code with such annotations

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -0,0 +1,522 @@
...
+        subsystem = props.getValueAsString(UDEV_KEY_SUBSYSTEM);
+
+        if (!SCANNABLE_DEVICES_SUBSYSTEM.contains(subsystem)) {
+            return null;
+        }
+
+        device = new Device();
+
+        device.setBus(subsystem);
+        device.setDriver(StringUtils.defaultIfBlank(props.getValueAsString(UDEV_KEY_DRIVER), "unknown"));
```
CDellaGiusta: worth putting "unknown" as a constant string (something like `Device.DRIVER_UNKNOWN `?)

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -0,0 +1,522 @@
...
+            String usbDeviceDesc = props.getValueAsString(UDEV_KEY_ID_MODEL_ID);
+            return String.format(PCI_FORMAT, vendorId, usbDeviceDesc);
+        }
+
+        String devType = props.getValueAsString(UDEV_KEY_DEVTYPE);
+        if (DEVTYPE_USB_INTERFACE.equals(devType)) {
+            String driver = props.getValueAsString(UDEV_KEY_DRIVER);
+            if (driver.equals("usbhid")) {
+                return "USB HID Interface";
+            }
```
CDellaGiusta, line 258: worth putting "usbhid" and "hub" as constants?

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -0,0 +1,522 @@
...
+                }
+                yield switch (pciSubCode) {
+                    case PCI_CLASS_INPUT_KEYBOARD -> Device.CLASS_KEYBOARD;
+                    case PCI_CLASS_INPUT_MOUSE -> Device.CLASS_MOUSE;
+                    case PCI_CLASS_INPUT_SCANNER -> Device.CLASS_SCANNER;
+                    default -> null;
+                };
+            }
+            case PCI_BASE_CLASS_SERIAL -> {
+                if (pciSubCode == null) {
```
CDellaGiusta, line 557: isn't this case already included in the following default?

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -0,0 +1,522 @@
...
+        PciClassCodes pciBaseCode = PciClassCodes.fromBaseCode(baseClass);
+        PciClassCodes pciSubCode = PciClassCodes.fromSubCode(pciBaseCode, subClass);
+
+        if (null == pciBaseCode) {
+            return null;
+        }
+
+        return switch (pciBaseCode) {
+            case PCI_BASE_CLASS_STORAGE -> {
+                if (pciSubCode == null) {
```
CDellaGiusta, line 517: ditto: is this already included on the default case below?

java/core/src/main/java/com/suse/manager/reactor/hardware/cpu/CpuInfoMapper.java
```diff
@@ -10,15 +10,22 @@
...
 package com.suse.manager.reactor.hardware.cpu;
 
+import static com.suse.manager.reactor.hardware.CpuArchUtil.ARCH_I386;
+import static com.suse.manager.reactor.hardware.CpuArchUtil.ARCH_X86_64;
+import static com.suse.manager.reactor.hardware.HardwareConstants.CPU_MHZ_NOT_APPLICABLE;
+import static com.suse.manager.reactor.hardware.HardwareConstants.GRAIN_CPU_ARCH;
+import static com.suse.manager.reactor.hardware.HardwareConstants.GRAIN_TOTAL_NUM_CPUS;
+import static com.suse.manager.reactor.hardware.cpu.CpuFieldTruncator.CPU_BOGOMIPS_LENGTH;
+import static com.suse.manager.reactor.hardware.cpu.CpuFieldTruncator.CPU_STEPPING_LENGTH;
+import static com.suse.manager.reactor.hardware.cpu.CpuFieldTruncator.CPU_VENDOR_LENGTH;
```
CDellaGiusta, line 20: I would prefer to use, for example `CpuArchUtil.ARCH_I386` as a constant instead of using the import, but I guess this is also a matter of personal preference

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
```diff
@@ -376,10 +432,10 @@ protected boolean mapScsiDevice(List<Map<String, Object>> udevdb) {
             String dpath = props.getValueAsString(UDEV_KEY_DEVPATH);
             Matcher m;
             if (StringUtils.isNotBlank(idpath)) {
-                m = Pattern.compile(".*scsi-(\\d+):(\\d+):(\\d+):(\\d+)").matcher(idpath);
+                m = SCSI_ID_PATH_REGEX.matcher(idpath);
             }
             else {
-                m = Pattern.compile(".*/(\\d+):(\\d+):(\\d+):(\\d+)/block/").matcher(dpath);
+                m = SCSI_DEVPATH_REGEX.matcher(dpath);
```
CDellaGiusta, line 438: here I'm a bit unsure: on one hand seeing directly what is the regex could be handy... on the other hand a compact constant looks better

CDellaGiusta (dismissed):
code looks good, just have a look at my suggestions, and dismiss them if not relevant.
A great job, anyway: now code looks much more modular!

java/core/src/main/java/com/suse/manager/reactor/hardware/network/NetworkMapper.java
rjpmestre: Yeah, i'd say it's fine. 
We're retrieving this network interface instance via `server.getNetworkInterface(name);`, just like the original code.

So this means the name is being used as an identifier here, and IF we're getting a non null value, that also implies that the  the server-network interface relationship was already set up.

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: it has to do with unchecked casts we keep doing to ValueMap objects. So, in this instance, specifically about 
`props = new ValueMap((Map<String, Object>) dbdev.get(UDEV_ENTRIES));`

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: Yeah, sorry, thats on me. Initially I was trying to avoid doing such a rework, and was only extracting literals that were used more than once.
But while reviewing the whole thing, I thought it would just be more consistent to have everything tidy and grouped together in same type of structure. Thats what my last commit was all about.

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: I forgot to remove the warning entry. I "fixed it" by adding a blank default case:
`default -> { // checkstyle 
}`

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: [for sure, same as here](https://github.com/uyuni-project/uyuni/pull/11836#discussion_r3895225660)

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: [for sure, same as here](https://github.com/uyuni-project/uyuni/pull/11836#discussion_r3895225660)

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: Nah, if `pciSubCode` was null, it wouldn't get matched.

Edit: it seems it has to do with the java version. Our 17 version does not support the `case null`. But it is supported at 21

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: replied at https://github.com/uyuni-project/uyuni/pull/11836#discussion_r3895337362

java/core/src/test/java/com/suse/manager/reactor/hardware/network/NetworkMapperTest.java
rjpmestre: Awh, for some reason I assumed checkstyle would complain about not having any javadoc.
But yeah, you're right, it's quite pointless.

java/core/src/main/java/com/suse/manager/reactor/hardware/cpu/CpuInfoMapper.java
rjpmestre: I agree, this one belongs here.  I ended up reorganizing these constants on this [commit](https://github.com/uyuni-project/uyuni/pull/11836#discussion_r3895225660).

java/core/src/main/java/com/suse/manager/reactor/hardware/device/DeviceMapper.java
rjpmestre: I think it works, especially because we were already able to gather so many literals and magic numbers over these classes.

From my experience, having them organized like this makes it easier to maintain, especially when we need to find and update one of these values. 
They're usually all at the same level, so you know exactly where to look, then, we also know which usages have the same purpose because they're using the same constant. So if we had duplicated literals instead, one might think the other occurrence is just unrelated or a different case.

Thats why for me, it doesn't make much difference whether the literal is used as a regex or something else.

PR #11845, opened 2026-04-28
Implement readiness script (bsc#1263157)

mbussolotto:
## What does this PR change?

as title

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30492

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/server-image/root/docker-entrypoint-init
+++ b/containers/server-image/root/docker-entrypoint-init
@@ -6,6 +6,9 @@
 # Fail if any init script fails
 set -e
 
+# Mark that initialization is in progress
+touch /var/run/uyuni-init-pending
+
 for f in $(find /docker-entrypoint-init.d/ -type f -name  '*.sh' | sort); do
     echo "Executing $f..."
     bash "$f"
@@ -37,6 +40,9 @@ unset SCC_USER
 unset SCC_PASS
 unset ISS_PARENT
 unset MANAGER_ENABLE_TFTP
+
+# Remove the in-progress marker - init completed successfully
+rm -f /var/run/uyuni-init-pending
 
 # This needs to be last for systemd to become PID1
 exec /usr/lib/systemd/systemd
--- a/containers/server-image/root/docker-entrypoint-init.d/90-debug-long-run.sh
+++ b/containers/server-image/root/docker-entrypoint-init.d/90-debug-long-run.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 SUSE LLC
+#
+# SPDX-License-Identifier: GPL-2.0-Only
+
+if test "$DEBUG_LONG_START" = "y"; then
+    sleep 10m
+fi
--- a/containers/server-image/root/usr/bin/healthcheck.sh
+++ b/containers/server-image/root/usr/bin/healthcheck.sh
@@ -1,48 +1,16 @@
-#!/bin/bash
-set -o pipefail
-MANAGER_COMPLETE="/var/spacewalk/.MANAGER_SETUP_COMPLETE"
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 SUSE LLC
+#
+# SPDX-License-Identifier: GPL-2.0-Only
 
-check_manager_login() {
-    local attempts=2
-    local attempt=1
-
-    while [ "$attempt" -le "$attempts" ]; do
-        if curl --noproxy localhost --fail --silent --show-error --connect-timeout 10 --max-time 20 \
-            http://localhost/rhn/manager/login > /dev/null; then
-            return 0
-        fi
-
-        sleep 0.5
-        attempt=$((attempt + 1))
-    done
-
-    return 1
-}
-
-set +e
-/usr/bin/spacewalk-diskcheck
-DISK_EXIT_CODE=$?
 set -e
 
-RESTART_MARKER="/var/run/uyuni-restart-pending"
-if [ -f "${RESTART_MARKER}" ]; then
-    echo "Restart pending – reporting unhealthy"
-    exit 1
-fi
+# This script is used on podman as is doesn't make a difference between liveness and readiness checks.
 
-if [ $DISK_EXIT_CODE -eq 3 ]; then
-    echo "Healthcheck failed: spacewalk-diskcheck returned critical error (3)"
-    exit 1
-fi
+/usr/bin/liveness-check.sh
 
-if [ -f "$MANAGER_COMPLETE" ]; then
-    /usr/bin/systemctl is-active multi-user.target
-    salt-call --local --no-color status.ping_master localhost |grep -q True
-    if ! check_manager_login; then
-        echo "Healthcheck failed: manager login endpoint is not reachable after retries."
-        exit 1
-    fi
-else
-    echo "Healthcheck failed: setup file not found. Skipping app checks, disk is healthy."
-    exit 1
-fi
+# Check that the login page shows up
+curl --noproxy localhost --fail --silent --show-error --connect-timeout 10 --max-time 20 \
+        --retry 2 --retry-all-errors --retry-delay 1 \
+        http://localhost/rhn/manager/login > /dev/null
+exit $?
--- a/containers/server-image/root/usr/bin/startup-check.sh
+++ b/containers/server-image/root/usr/bin/startup-check.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 SUSE LLC
+#
+# SPDX-License-Identifier: GPL-2.0-Only
+
+# Exit codes:
+# 0 - Ready
+# Any thing else: Pending or Failed
+
+# Check if initialization is still in progress
+if [ -f "/var/run/uyuni-init-pending" ]; then
+    echo "Initialization in progress"
+    exit 1
+fi
+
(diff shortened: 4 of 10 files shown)


aaannz:
I don't think checking for the reindex is a sufficient approach as there might be some other long term things happening.
I think we should modify init scripts a little.


Add extra mark file next to setup finished, something like `/var/lib/.startInProgress`, which would be setup by entry point and removed once scripts are finished.

Intentionally not in persisted folder so that if script fails, we do not have to have trap to remove that file.

And when this file is present we use that as a readiness probe.

aaannz:
Also the problem with healthcheck returning 0 is that when this happens, mgradm script returns sooner then everything is setup.

Original healthcheck was returning code 0 when setup complete mark file was missing and so mgradm exited almost immediately when nothing was done at all.

rjmateus:
@aaannz I like your proposal. The only thing that we may be missing is what to do when an init script fails.
In the behaviour you propose, it would have the `/var/lib/.startInProgress` but the init would fail. In this case, should we have an error file to sign the health check script? If we don't have something in place, the health check script will continue thinking that the init is still in progress.

mbussolotto:
@aaannz @rjmateus , this should be similar to your proposal, I used `/var/run` folder since we're already using it for `/var/run/uyuni-restart-pending`

containers/server-image/root/usr/bin/healthcheck.sh
```diff
@@ -20,10 +24,20 @@ check_manager_login() {
...
 set +e
+/usr/bin/readiness.sh
+READINESS_CODE=$?
 /usr/bin/spacewalk-diskcheck
 DISK_EXIT_CODE=$?
 set -e
 
+if [ $READINESS_CODE -eq 2 ]; then
+    echo "Healthcheck passed: Initialization in progress"
+    exit 0
```
mcalmer: Wouldn't this just cause go directly to healthy state? Is this really what we want?

aaannz: Indeed we cannot have it like this.
But I think we don't have to run it in healthcheck at all.

After reading the podman docs there is `healt-startup-cmd` we can use for this and we can add infinite time there which will wait until container is started up.

And have regular healthcheck started after startup which then can actually stop the container is something goes wrong.

We need to adapt uyuni-tools too.

containers/server-image/root/docker-entrypoint-init
```diff
@@ -6,6 +6,12 @@
 # Fail if any init script fails
 set -e
 
+# Mark that initialization is in progress
+touch /var/run/uyuni-init-pending
+
+# Cleanup on failure - mark init as failed
+trap 'touch /var/run/uyuni-init-failed; rm -f /var/run/uyuni-init-pending' ERR
```
aaannz: I am not sure this will work as expected. `/var/run` is not persisted.

Init script exits with an error -> trap handler creates this file -> container immediately exits and non-persistent data are lost.

This would IMO work only for brief milliseconds readiness accidentally run during container shutdown.

cbosdo:
> @aaannz I like your proposal. The only thing that we may be missing is what to do when an init script fails. In the behaviour you propose, it would have the `/var/lib/.startInProgress` but the init would fail. In this case, should we have an error file to sign the health check script? If we don't have something in place, the health check script will continue thinking that the init is still in progress.

The startup health checks on both podman and Kubernetes are binary things: either it succeeds and we are good to continue or it fails and the start is not finished. The start is considered failed after a given number of failed attempts that we define. On podman, this number is controlled by `--health-startup-retries` and a `0` means the container is never restarted. On the Kubernetes side however, this is defined by `failureThreshold` and needs to have a minimum of `1`… which means there is no in(de)finite start time.

This means that it's fine to have the `/var/lib/.startInProgress` file in a non-persisted folder as real startup failures are not really reported by the script using the exit code.

containers/server-image/root/usr/bin/readiness.sh
```diff
@@ -0,0 +1,24 @@
...
+# Check if initialization failed
+if [ -f "/var/run/uyuni-init-failed" ]; then
+    echo "Readiness failed: Initialization scripts failed"
+    exit 1
+fi
+
+# Check if initialization is still in progress
+if [ -f "/var/run/uyuni-init-pending" ]; then
+    echo "Readiness not ready: Initialization in progress"
+    exit 2
```
cbosdo: That is not needed with the podman and Kubernetes startup health checks… they only consider `0` as a success or anything else as a failure. In-progress is not a thing.

containers/server-image/root/usr/bin/readiness.sh
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: GPL-2.0-Only
+
+# Exit codes:
+# 0 - Ready
+# 1 - Failed
+# 2 - In progress
+
+# Check if initialization failed
+if [ -f "/var/run/uyuni-init-failed" ]; then
+    echo "Readiness failed: Initialization scripts failed"
```
cbosdo: I think that we need to give proper names to those scripts:
* `startup-check.sh`: checking if we reached the services start
* `readiness-check.sh`: if the pod is ready to handle HTTP queries
* `liveness-check.sh`: would be to check if the service is in a badly broken state that requires a restart. I have no real idea if we need this one as it's quite hard to come with a really good detection here.

On Kubernetes, the liveness check calls the `api.getVersion` unauthenticated HTTP query to check that we are good to get traffic. This could be used for podman too…

containers/server-image/root/usr/bin/startup-check.sh
```diff
@@ -0,0 +1,24 @@
...
+# 2 - In progress
+
+# Check if initialization failed
+if [ -f "/var/run/uyuni-init-failed" ]; then
+    echo "Readiness failed: Initialization scripts failed"
+    exit 1
+fi
+
+# Check if initialization is still in progress
+if [ -f "/var/run/uyuni-init-pending" ]; then
```
cbosdo, line 11: Do we want to also check that we reached the `spacewalk.target` service? Note that it doesn't mean that the queries would be ready to be handled, but it's a good time to start the check for this.

cbosdo (changes requested):

containers/server-image/root/usr/bin/readiness.sh
rjmateus: > liveness-check.sh: would be to check if the service is in a badly broken state that requires a restart. I have no real idea if we need this one as it's quite hard to come with a really good detection here.

for this one maybe we can consider calling `spacewalk-service status` and check the exit code. If I remember correctly it return 0 if all services are health and a different number in case something is wrong.

containers/server-image/root/usr/bin/healthcheck.sh
rjmateus: The option Ondrej proposed looks the safest one. Even if in kubernetes it cannot be infinite, we can play around with the parameters to give plenty of time to finish the initialization/migration.

containers/server-image/root/docker-entrypoint-init
rjmateus: @aaannz then the question in here is also how to behave in an error on start-up. After the restart, should we re-run the start-up procedure? If we need some special behaviour in case of failure, we need to save that file in a persistent location.
I think the scripts are protected to be able to run multiple times (since they run at every container start) but if it fails, can it land in an inconsistent state?
At least the database schema migration can cope with it.

containers/server-image/root/usr/bin/readiness.sh
cbosdo: > > liveness-check.sh: would be to check if the service is in a badly broken state that requires a restart. I have no real idea if we need this one as it's quite hard to come with a really good detection here.
> 
> for this one maybe we can consider calling `spacewalk-service status` and check the exit code. If I remember correctly it return 0 if all services are health and a different number in case something is wrong.

Our current `healthcheck.sh` script is a mix of liveness and readiness: checking for disk space is a liveness check, but checking the login page is a readiness check.

containers/server-image/root/docker-entrypoint-init
cbosdo: The scripts need to behave at each start… otherwise it will be hard to figure out what to run when.

cbosdo:
podman implementation needs https://github.com/uyuni-project/uyuni-tools/pull/781

cbosdo (dismissed):

containers/server-image/root/docker-entrypoint-init.d/90-debug-long-run.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 SUSE LLC
+#
+# SPDX-License-Identifier: GPL-2.0-Only
+
+if test "$DEBUG_LONG_START" = "y"; then
```
mbussolotto, line 6: now that we startupProbe, livenessProbe and readinessProbe, can't we remove it?

cbosdo: That's what I wonder… would this be useful to toggle this on for tests in the future?

cbosdo (approved):

PR #11848, opened 2026-04-29
Postponing action execution if not ready to run

CDellaGiusta:
## What does this PR change?

This implementation makes possible for an action execution to be postponed if the Action is not yet ready to run, i.e if the implementation of the Action method `isReadyToRun() `returns false.

This sums up to the condition where a Taskomatic task was triggered before the corresponding Action was actually committed in the database.

In both cases, the quartz event scheduling the Action execution is postponed in time by a number of seconds equal to the retry count (that is 1 second the first time, 2 seconds the second time and so on) up to a maximum of 25 retries (totaling 5 min 25 sec maximum)

This implementation is made with the CoCo attestation action in mind (it could not have all the input data to be ready to execute at scheduling time), but it can be used in future more in general cases, when for any reason, the Action could not be immediately ready to be executed.

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: manually tested
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/30374
Port(s): 
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
+++ b/java/core/src/main/java/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
@@ -108,27 +108,54 @@ public class MinionActionExecutor extends RhnJavaJob {
         }
 
         Action action = ActionFactory.lookupById(actionId);
+        long queuedServerActionsNumber = countQueuedServerActions(action);
 
         if (log.isDebugEnabled()) {
-            log.debug("Number of Queued Server Actions for {}: {}", actionId, countQueuedServerActions(action));
+            log.debug("Number of Queued Server Actions for {}: {}", actionId, queuedServerActionsNumber);
         }
 
         // HACK: it is possible that this Taskomatic task triggered before the corresponding Action was really
         // COMMITted in the database. If it is the case, reschedule the Job.
+        // Also: the Action could not be ready for execution yet. If this is the case, reschedule the Job as well
         int retryCount = (Integer) context.getTrigger().getJobDataMap().getOrDefault("retryCount", 1);
+        boolean failedNoServerActionsYet = (0 == queuedServerActionsNumber);
+        boolean failedNotReadyToRun = ((null != action) && (!action.isReadyToRun()));
 
-        if (retryCount <= MAX_RETRIES && countQueuedServerActions(action) == 0) {
-            log.debug("Couldn't find queued actions for action {}, scheduling retry {} of {}",
-                actionId, retryCount, MAX_RETRIES
-            );
+        if ((retryCount <= MAX_RETRIES) && (failedNoServerActionsYet || failedNotReadyToRun)) {
+            if (failedNoServerActionsYet) {
+                log.debug("Couldn't find queued actions for action {}, scheduling retry {} of {}",
+                        actionId, retryCount, MAX_RETRIES);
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
```diff
@@ -131,6 +141,19 @@ public void execute(JobExecutionContext context) {
             }
         }
 
+        if (failedNotReadyToRun) {
+            log.debug("Action {} is not yet ready to run after {} retries, marking it as failed",
+                    actionId, MAX_RETRIES);
+
+            action.getServerActions().forEach(
```
rjmateus: In location, action cannot be null since the condition is checked before. However, I think we can check again that the action is not null, since the check is done in an indirect mechanism. If someone changes the condition above, we can risk having a null point in here.

rjmateus (approved):
The overall code look good to me. I think we can make just one small improvement in the checks

java/core/src/main/java/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
CDellaGiusta: This was also my first approach, as I prefer to have a safety check even if redundant. I removed it since the IDE suggested me the condition was already checked, and I suspect SonarCloud will complain about this.
So I re-introduced the null check, leaving an explanatory comment so we can safely dismiss any SonarCloud issue if it pops up.

java/core/src/main/java/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
```diff
@@ -108,21 +108,31 @@ public void execute(JobExecutionContext context) {
...
-            );
+        if ((retryCount <= MAX_RETRIES) && (failedNoServerActionsYet || failedNotReadyToRun)) {
+            if (failedNotReadyToRun) {
+                log.debug("Action {} is not yet ready to run, scheduling retry {} of {}",
+                        actionId, retryCount, MAX_RETRIES);
+            }
+            else {
+                log.debug("Couldn't find queued actions for action {}, scheduling retry {} of {}",
+                        actionId, retryCount, MAX_RETRIES);
+            }
```
wweellddeerr: Just a minor suggestion: I would invert the `if/else` sequence here to make the logging more precise.

If `failedNoServerActionsYet` is true, it means the database transaction hasn't even finished committing the actions yet. If both conditions happen to be true and we check `failedNotReadyToRun` first, we'll log that the action is "not ready to run", which can be misleading and mask the root cause. 

Checking `failedNoServerActionsYet` first ensures the log accurately reflects a database concurrency delay rather than a business logic wait.

wweellddeerr (approved):

java/core/src/main/java/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java
CDellaGiusta: Giving `failedNoServerActionsYet` priority makes sense. I took your suggestion.

PR #11849, opened 2026-04-29
Test external CA on RKE2 server and proxy

ilmanzo:
## What does this PR change?

reference card: https://github.com/SUSE/spacewalk/issues/29464

Add end-to-end Cucumber tests that verify replacing the cert-manager
  self-signed CA with an externally provided one on both the server and
  proxy RKE2 clusters, and that deployments recover afterwards.

The test scenario is hard to recreate because we'd need to have infrastructure with a CA already in place, and deploy MLM-RKE2 on that.

To simulate a realistic scenario, we use the regular deployment and then switch to an external CA after the install.

This also create an issue when the controller needs to communicate with the machines, as previous SSL trust is being broken. So we need to reconfigure the controller trust as well.

  New files:
  - srv_rke2_external_ca.feature: 4 scenarios covering CA replacement on server, leaf cert re-issuance verification, proxy CA injection via configmap, and deployment recovery on both clusters.
  - rke2_steps.rb: step definitions for generating an EC CA, replacing the uyuni-ca secret (deleting the Certificate CR first to prevent cert-manager from overwriting it), deleting and waiting for leaf secrets, verifying issuer chains, re-generating proxy certificates, and transferring secrets between clusters.
  - certificate.yaml.erb: shared ERB template for cert-manager Certificate resources, replacing the inline YAML string in kubernetes.rb.

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/kubernetes/srv_rke2_external_ca.feature
+++ b/testsuite/features/kubernetes/srv_rke2_external_ca.feature
@@ -1,0 +1,39 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+# Verifies that MLM supports replacing the self-signed CA with an externally
+# provided one on both the server and proxy RKE2 clusters.
+
+@rke2
+Feature: RKE2 External CA Replacement
+
+  Scenario: Pre-requisite: Back up CA certificates before external CA replacement
+    Given The Kubernetes cluster is ready on "server"
+    And I back up the CA certificates on the server and proxy
+
+  Scenario: Replace self-signed CA with an external CA
+    When I generate an external CA on "server"
+    And I replace the uyuni-ca secret with the external CA on "server"
+    And I delete the leaf certificate secrets on "server"
+    Then the "uyuni-cert" secret on "server" should be re-issued within 10 minutes
+    And the "db-cert" secret on "server" should be re-issued within 5 minutes
+    And the "uyuni-cert" certificate on "server" should be signed by the external CA
+    And the "db-cert" certificate on "server" should be signed by the external CA
+
+  Scenario: Verify server deployments recover after CA replacement
+    Then the "db" deployment on "server" should become ready within 5 minutes
+    And the "uyuni" deployment on "server" should become ready within 15 minutes
+
+  Scenario: Inject external CA into proxy cluster
+    Given The Kubernetes cluster is ready on "proxy"
+    When I re-generate the proxy certificate on the server using the external CA
+    And I transfer the proxy certificate from the server to "proxy"
+    And I update the uyuni-ca configmap on "proxy" with the external CA
+    Then the "proxy-cert" certificate on "proxy" should be signed by the external CA
+
+  Scenario: Verify proxy deployments recover after CA replacement
+    Then the "uyuni-proxy" deployment on "proxy" should become ready within 10 minutes
+    And the "uyuni-proxy-tftp" deployment on "proxy" should become ready within 10 minutes
+
+  Scenario: Cleanup: Restore original CA certificates after external CA tests
+    When I restore the original CA certificates on the server and proxy
--- a/testsuite/features/step_definitions/rke2_steps.rb
+++ b/testsuite/features/step_definitions/rke2_steps.rb
@@ -1,4 +1,4 @@
-# Copyright 2026 SUSE LLC
+# Copyright (c) 2026 SUSE LLC.
 # Licensed under the terms of the MIT license.
 
 Given('The first-time setup job is successful') do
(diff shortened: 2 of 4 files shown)


testsuite/features/kubernetes/srv_rke2_external_ca.feature
```diff
@@ -0,0 +1,33 @@
...
+
+  Scenario: Inject external CA into proxy cluster
+    Given The Kubernetes cluster is ready on "proxy"
+    When I re-generate the proxy certificate on the server using the external CA
+    And I transfer the proxy certificate from the server to "proxy"
+    And I update the uyuni-ca configmap on "proxy" with the external CA
+    Then the "proxy-cert" certificate on "proxy" should be signed by the external CA
+
+  Scenario: Verify proxy deployments recover after CA replacement
+    Then the "uyuni-proxy" deployment on "proxy" should become ready within 10 minutes
```
Pablogoliva, line 35: The cucumber steps start with capital letters. Change it pls :)

testsuite/features/kubernetes/srv_rke2_external_ca.feature
```diff
@@ -0,0 +1,33 @@
...
+
+  Scenario: Replace self-signed CA with an external CA
+    Given The Kubernetes cluster is ready on "server"
+    When I generate an external CA on "server"
+    And I replace the uyuni-ca secret with the external CA on "server"
+    And I delete the leaf certificate secrets on "server"
+    Then the "uyuni-cert" secret on "server" should be re-issued within 10 minutes
+    And the "db-cert" secret on "server" should be re-issued within 5 minutes
+    And the "uyuni-cert" certificate on "server" should be signed by the external CA
+    And the "db-cert" certificate on "server" should be signed by the external CA
```
Pablogoliva, line 21: Same as below. The steps start with capital letter.

ilmanzo: they do. "And", "Then" start with a capital letter. Other features file on the same project follow the same convention.

testsuite/features/kubernetes/srv_rke2_external_ca.feature
ilmanzo: they do. "And", "Then" start with a capital letter. Other features file on the same project follow the same convention.
see https://cucumber.io/docs/gherkin/reference

testsuite/features/kubernetes/srv_rke2_external_ca.feature
ilmanzo: see 
https://cucumber.io/docs/gherkin/reference

Pablogoliva (dismissed):

testsuite/features/support/env.rb
```diff
@@ -734,6 +735,139 @@ def process_code_coverage
...
+  $context[:external_ca] = {
+    server: get_target('server'),
+    proxy: proxy_target,
+    backup_dir: backup_dir,
+    external_ca_dir: ca_dir
+  }
+  $external_ca_feature = scenario.location.file
+end
+
+Before do |scenario|
```
maximenoel8: We already have a Before do |scenario|
We should not use this method for only one feature. They are used for all scenarios

testsuite/features/support/env.rb
```diff
@@ -734,6 +735,139 @@ def process_code_coverage
...
+  $stdout.puts 'Restoring original CA certificates after @external_ca feature...'
+  begin
+    restore_external_ca($context[:external_ca])
+  rescue StandardError => e
+    warn "Failed to restore CA certificates: #{e.message}"
+  end
+  $external_ca_feature = nil
+end
+
+AfterAll do
```
maximenoel8: We already have a AfterAll. Same, it will be apply to all features and not only the feature you are creating

maximenoel8 (changes requested):

maximenoel8:
Don't use Before scenario and afterall only for one feature.
The better approach is to use scenario with Cleanup: and Pre-requisite: prefix within the feature
All scenarios are run in a features.
EG: https://github.com/uyuni-project/uyuni/blob/4a834ad43915223c748621547b7811c65676a12d/testsuite/features/secondary/allcli_action_chain.feature

ilmanzo:
> Don't use Before scenario and afterall only for one feature. The better approach is to use scenario with Cleanup: and Pre-requisite: prefix within the feature All scenarios are run in a features. EG: https://github.com/uyuni-project/uyuni/blob/4a834ad43915223c748621547b7811c65676a12d/testsuite/features/secondary/allcli_action_chain.feature

I see. So you suggest to add extra feature steps for the setup and cleanup, even if they are strictly part of the testing . Got it right ?

maximenoel8:
> > Don't use Before scenario and afterall only for one feature. The better approach is to use scenario with Cleanup: and Pre-requisite: prefix within the feature All scenarios are run in a features. EG: https://github.com/uyuni-project/uyuni/blob/4a834ad43915223c748621547b7811c65676a12d/testsuite/features/secondary/allcli_action_chain.feature
> 
> I see. So you suggest to add extra feature steps for the setup and cleanup, even if they are strictly part of the testing . Got it right ?

Yes that's correct. We just prefix with Pre-requisite or cleanup. When we review, we know it's not important but should work.
The default behavior is that all scenarios are going to run even if one is failing. So you don't really need before after. We just use it for general action (screenshot, try to reconnect if web session crash etc.)

maximenoel8 (approved):
LGTM, thank you !

PR #11867, opened 2026-05-04
Implementation of IBM z series CoCo attestation

CDellaGiusta:
## What does this PR change?

This PR implements the confidential computing attestation for the IBM z series machines, on the MLM server side.

**IMPORTANT:** 
Please consider **only the last 3 commits** (they are built on top of [Restructuring CoCo attestation PR](https://github.com/uyuni-project/uyuni/pull/11788) that has to be considered as its ancestor)

This PR is into a chain of PRs:

CoCo attestation, MLM server base part: https://github.com/uyuni-project/uyuni/pull/11788
On top of that: MLM server IBM part: https://github.com/uyuni-project/uyuni/pull/11867

CoCo attestation, microservices base part: https://github.com/uyuni-project/uyuni/pull/11868
On top of that: microservices IBM part: https://github.com/uyuni-project/uyuni/pull/11856

For all the solution details, please see [RFC: Confidential Compute Attestation for IBM Z series](https://github.com/uyuni-project/uyuni-rfc/pull/113)

## Codespace

Check if you already have a running container clicking on [Running CodeSpace](https://github.com/codespaces)

[Create CodeSpace](https://codespaces.new/uyuni-project/uyuni)  [About billing for Github Codespaces](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces) [CodeSpace Billing Summary](https://github.com/settings/billing/summary) [CodeSpace Limit](https://github.com/settings/billing/spending_limit)

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- Documentation issue was created: [Link for SUSE Multi-Linux Manager contributors](https://github.com/SUSE/spacewalk/issues/new?template=ISSUE_TEMPLATE_DOCUMENTATION.md&labels=documentation&projects=SUSE/spacewalk/31), [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/new).
- (OPTIONAL) [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pulls)

- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/30376
Port(s): 
- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

[...]

PR #11856, opened 2026-04-30
CoCo attestation microservices : Implementation of IBM z series CoCo attestation microservices

CDellaGiusta:
## What does this PR change?

This PR changes the Confidential computing attestation mechanism on the **_microservices_** side to support IBM z series CoCo attestation.

**IMPORTANT:** 
Please consider **only the last 4 commits** (they are built on top of [CoCo attestation microservices : changes to support attestation on more platforms](https://github.com/uyuni-project/uyuni/pull/11868) that has to be considered as its ancestor)

This PR is into a chain of PRs:

CoCo attestation, MLM server base part: https://github.com/uyuni-project/uyuni/pull/11788
On top of that: MLM server IBM part: https://github.com/uyuni-project/uyuni/pull/11867

CoCo attestation, microservices base part: https://github.com/uyuni-project/uyuni/pull/11868
On top of that: microservices IBM part: https://github.com/uyuni-project/uyuni/pull/11856

For all the solution details, please see [RFC: Confidential Compute Attestation for IBM Z series](https://github.com/uyuni-project/uyuni-rfc/pull/113)

## Codespace

Check if you already have a running container clicking on [Running CodeSpace](https://github.com/codespaces)

[Create CodeSpace](https://codespaces.new/uyuni-project/uyuni)  [About billing for Github Codespaces](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces) [CodeSpace Billing Summary](https://github.com/settings/billing/summary) [CodeSpace Limit](https://github.com/settings/billing/spending_limit)

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- Documentation issue was created: [Link for SUSE Multi-Linux Manager contributors](https://github.com/SUSE/spacewalk/issues/new?template=ISSUE_TEMPLATE_DOCUMENTATION.md&labels=documentation&projects=SUSE/spacewalk/31), [Link for community contributors](https://github.com/uyuni-project/uyuni-docs/issues/new).
- (OPTIONAL) [Documentation PR](https://github.com/uyuni-project/uyuni-docs/pulls)
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/30376
Port(s): 
- [x] **DONE**

## Changelogs

[...]

PR #11788, opened 2026-04-17
Restructuring CoCo attestation to support attestation on more platforms

CDellaGiusta:
## What does this PR change?

This PR changes the Confidential computing attestation mechanism on the MLM server side, to prepare the support for more attestation types than the only one currently supported (AMD). In particular, this is a preliminary work that forms the basis to support IBM z series CoCo attestation.

This PR is into a chain of PRs:

CoCo attestation, MLM server base part: https://github.com/uyuni-project/uyuni/pull/11788
On top of that: MLM server IBM part: https://github.com/uyuni-project/uyuni/pull/11867

CoCo attestation, microservices base part: https://github.com/uyuni-project/uyuni/pull/11868
On top of that: microservices IBM part: https://github.com/uyuni-project/uyuni/pull/11856

For all the solution details, please see [RFC: Confidential Compute Attestation for IBM Z series](https://github.com/uyuni-project/uyuni-rfc/pull/113)

## Codespace

Check if you already have a running container clicking on [Running CodeSpace](https://github.com/codespaces)

[Create CodeSpace](https://codespaces.new/uyuni-project/uyuni)  [About billing for Github Codespaces](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces) [CodeSpace Billing Summary](https://github.com/settings/billing/summary) [CodeSpace Limit](https://github.com/settings/billing/spending_limit)

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30374
Port(s): # 
- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
[...]

PR #11868, opened 2026-05-04
CoCo attestation microservices : changes to support attestation on more platforms

CDellaGiusta:
## What does this PR change?

This PR changes the Confidential computing attestation mechanism on the **_microservices_** side, to prepare the support for more attestation types than the only one currently supported (AMD). In particular, this is a preliminary work that forms the basis to support IBM z series CoCo attestation.

This PR is into a chain of PRs:

CoCo attestation, MLM server base part: https://github.com/uyuni-project/uyuni/pull/11788
On top of that: MLM server IBM part: https://github.com/uyuni-project/uyuni/pull/11867

CoCo attestation, microservices base part: https://github.com/uyuni-project/uyuni/pull/11868
On top of that: microservices IBM part: https://github.com/uyuni-project/uyuni/pull/11856

For all the solution details, please see [RFC: Confidential Compute Attestation for IBM Z series](https://github.com/uyuni-project/uyuni-rfc/pull/113)

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/30375
Port(s): 
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


microservices/coco-attestation/attestation-module/src/main/java/com/suse/coco/model/AttestationStatus.java
```diff
@@ -11,8 +11,58 @@
...
+    }
+
+    /**
+     * Checks if status is PENDING
+     *
+     * @return true if status is PENDING
+     */
+    public boolean isProcessingAttestationVerification() {
+        return (this == PENDING);
+    }
```
mackdk: These two methods seems redundant to me, at least for now. They are used only once outside of this enum and I think currently it's just easier and more readable to use directly a `==` check. Maybe in the future if we have more states we can revisit this implementation idea.

microservices/coco-attestation/attestation-module/src/main/java/com/suse/coco/model/AttestationStatus.java
```diff
@@ -11,8 +11,58 @@
...
+    public AttestationStatus getProcessingResultStatus(boolean success) {
+        if (success && isProcessingAttestationRequest()) {
+            return AttestationStatus.QUEUED;
+        }
+        if (success && isProcessingAttestationVerification()) {
+            return AttestationStatus.SUCCEEDED;
+        }
+
+        return AttestationStatus.FAILED;
+    }
```
mackdk: I think `resolveNextStatus()` or even simply `nextStatus()` are better names for this method :-) Moreover you can simplify the implementation:

```java
public AttestationStatus resolveNextStatus(boolean success) {
    if (!success) {
        return FAILED;
    }

    return switch (this) {
        case REQUESTED -> QUEUED;
        case PENDING   -> SUCCEEDED;
        default        -> FAILED;
    };
}
```

Open question: should we return `FAILED` if `success` is `true` but the current status is not either `REQUESTED` or `PENDING`? Maybe returning `this` would be a better result in this (unsupported) case?

microservices/coco-attestation/attestation-module/src/main/java/com/suse/coco/module/AttestationWorker.java
```diff
@@ -18,14 +18,21 @@
...
-    boolean process(SqlSession session, AttestationResult attestationResult);
+    boolean processAttestationRequest(SqlSession session, AttestationResult attestationResult);
+
+    /**
+     * Process the attestation verification of a given attestation result.
+     * @param session the active mybatis database session
+     * @param attestationResult the attestation result to process.
+     * @return <code>true</code> if the processing succeeded, <code>false</code> otherwise.
+     */
+    boolean processAttestationVerification(SqlSession session, AttestationResult attestationResult);
```
mackdk: Maybe we can drop the `Attestation` part from the name? Seems kind of redundant. Maybe just `processAttestationRequest` and `processVerification`?

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/CertificateHelper.java
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+
+package com.suse.common.utilities;
```
mackdk: I would avoid the `utilities` package, as this project is basically all composed of utilities. I think this class should live in `com.suse.common.security`

microservices/uyuni-java-common/src/main/java/com/suse/common/security/CertificateHelper.java
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * @param urlIn url of the file to download
+     * @return downloaded file as string content
+     * @throws IOException if something goes wrong
+     */
+    private static String downloadStringContent(String urlIn) throws IOException {
+        try (BufferedInputStream in = new BufferedInputStream(new URL(urlIn).openStream())) {
+            return new String(in.readAllBytes(), StandardCharsets.UTF_8);
+        }
```
mackdk: I don't think this code is safe enough for a generic "download a file from the internet" utility. It reads everything in a string, it doesn't handle timeouts  and the URL constructor is going to be deprecated from java 20. I would recommend using `java.net.http.HttpClient` here, where you can handle timeouts, follow redirects and check the return code explicitly.

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/JsonUtilities.java
mackdk, line 1: While creating json is easy enough to be done without dependencies (even if this code doesn't handle escaping), if you also need parsing I think it would be better to just use `jackson`, which is the java standard for that. Not sure what others think about that @rjmateus @mcalmer 

And if we end keeping this class, please move the package like `CertificateHelper` :-)

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/ShellCommandExecutor.java
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+
+package com.suse.common.utilities;
```
mackdk: I'd suggest `com.suse.common.io.process`

microservices/uyuni-java-common/src/test/java/com/suse/common/utilities/JsonUtilitiesTest.java
```diff
@@ -0,0 +1,227 @@
...
+                    "YUTu5Z9vzE+OYUBNroD/my/T4TIz9JHxL0T/Sw3vxGDcehSDpr8wJAJQzBr0cgfNg1Kiwe3Ivy9P" +
+                    "oegY+2UwJL5AXy6Pq3+4G7m4FeJeL7XjL6bX0SmvQ9fKJzOFLZUaVsleltolgSQVZ0nHnSKEg587" +
+                    "i3Cp8jcN4b9xcJuOaPosu/uvrfmic2Dp7l3FcZJ8SGEou1o9W3W3f/w6foCWTVrP3oc97HkPPJgH" +
+                    "/uMq6WPwBIh6YU00Yhi5AH/yzJUJ37hgFHu25Y4/8eubCbauIO9Jo9MQ7PQWjhtO24//WJsh5rnT" +
+                    "On3AJr+OkEcPFN0ZSvWLJbLW7vjZJSAAmPuKomftGUgeWp8KG9X1HWx3FlGnaAFWFRKbFzE1p7mv" +
+                    "EPlEFVAtPJ/9g3y3BFg4zuWZZnWCvSJ9r6ZOu9jZQw3sORmqvJbEg5bqFlXYioLtTKqfn4ppOqTM" +
+                    "5SqsZ1jJHWorzcmOHgqVsXM09nfOFC06kX/OwqE4vb3w/3Lxms8n4XRHrsO7ODG5DoBdPldxmNRm" +
+                    "iG+LAPdlUybbc6LjfNj0cy7Af8KT+izmhVp0bHKQg+tcHqngrAD56zRwXzxfqQL7P14DdFhRmuI6" +
+                    "9PKkto+bnNMjfyY28q4=";
+
```
mackdk: These declaration are a bit too long for the java code. Move these to resource files

microservices/uyuni-java-common/src/test/java/com/suse/common/utilities/CertificateHelperTest.java
```diff
@@ -0,0 +1,173 @@
...
+            KOtfJqGVWEjVGv7XJz/9kNF2ht0csGBc8w2o7uCJob054ThO2m67Np375SFTWsPK
+            6Wrxoj7bQ7gzyE84FJKZ9d3OVG3ZXQIUH0AzfAPilbLCIXVzUstG2MQ0HKKlS43N
+            b3Y3LIU/Gs4m6Ri+kAewQ3+ViCCCcPDMyu/9KTVcH4k4Vfc3iosJocsL6TEa/y4Z
+            XDlx4b6cpwoG1iZnt5LmTl/eeqxJzy6kdJKt2zyknIYf48FWGysj/4+16oh7cGvm
+            oLr9Oj9FpsToFpFSi0HASIRLlk2rREDjjfAVKM7t8RhWByovEMQMCGQ8M4+uKIw8
+            y4+ICw2/O/TOHnuO77Xry7fwdxPm5yg/rBKupS8ibEH5glwVZsxsDsrFhsP2JjMM
+            B0ug0wcCampAMEhLNKhRILutG4UI4lkNbcoFUCvqShyepf2gpx8GdOfy1lKQ/a+F
+            SCH5Vzu0nAPthkX0tGFuv2jiJmCG6sivqf6UHedjGzqGVnhO
+            -----END CERTIFICATE-----
+            """;
```
mackdk: This can also be resource file instead of being in the java test code.

mackdk (changes requested):

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/ShellCommandExecutor.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        /**
+         * Checks if there is a meaningful standard error
+         * @return true if there is a meaningful standard error
+         */
+        public boolean hasStandardError() {
+            return standardError != null && !standardError.isBlank();
+        }
+    }
```
mackdk: I would just extract this `record` to its own file

microservices/coco-attestation/attestation-module/src/main/java/com/suse/coco/model/AttestationStatus.java
CDellaGiusta: done

microservices/coco-attestation/attestation-module/src/main/java/com/suse/coco/model/AttestationStatus.java
CDellaGiusta: renamed and simplified.
As for the default value, I would highlight that "you should not be here", and hence put it to FAILED

microservices/coco-attestation/attestation-module/src/main/java/com/suse/coco/module/AttestationWorker.java
CDellaGiusta: when in doubt, I usually prefer to be verbose.
Changed

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/CertificateHelper.java
CDellaGiusta: moved package

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/ShellCommandExecutor.java
CDellaGiusta: moved

microservices/uyuni-java-common/src/test/java/com/suse/common/utilities/CertificateHelperTest.java
CDellaGiusta: moved to resources

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/ShellCommandExecutor.java
CDellaGiusta: done

microservices/uyuni-java-common/src/main/java/com/suse/common/utilities/JsonUtilities.java
CDellaGiusta: This class was not effectively used, and I lost track of it during the development.
We only need to synthesise json and never need to parse (using postgresql parsing) so I dropped the class and its usages

microservices/uyuni-java-common/src/test/java/com/suse/common/utilities/JsonUtilitiesTest.java
CDellaGiusta: dropped with the class

microservices/uyuni-java-common/src/main/java/com/suse/common/security/CertificateHelper.java
CDellaGiusta: I rewrote it according to your suggestions. Please review it.

mackdk (approved):

rjmateus (approved):

PR #11940, opened 2026-05-13
Update pg_hba.conf

nadvornik:
## What does this PR change?

This PR updates  pg_hba.conf:

- Disable external connection to the main db
- Enforce SSL on external connection to reportdb
- Disable SSL on internal connections
- Add pg_hba.conf section for customization, that is preserved on update
- Add variables to 00-mgrSetup.sh for enabling ssl (currently disabled)

The internal network is identified by hostname `uyuni-server` - name of the container on podman
and by `."$NAMESPACE".svc.cluster.local` domain on kubernetes.

00-mgrSetup.sh previously tried to enable ssl on reportdb, but the option was later overwritten by spacewalk-setup.
Now it is disabled explicitly, can be enabled by a variable (but this is needed only for external db).

The localhost entries are needed for the healthcheck https://github.com/uyuni-project/uyuni/blob/14c5425d9f536efffd3fd2658e463c87287efc0c/containers/server-postgresql-image/root/usr/bin/healthcheck.sh#L17

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- TBD
- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30434

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
```diff
@@ -7,24 +7,58 @@
...
-    PG_HBA_CHANGED="y"
+NEW_CONFIG="local all all peer
+local replication all peer
+
+$START_MARKER
+$CUSTOM_CONFIG
+$END_MARKER
+
+host all all 127.0.0.1/8 scram-sha-256
+host all all ::1/128 scram-sha-256"
```
aaannz, line 45: Did you find out why are these needed?

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
```diff
@@ -7,24 +7,58 @@
...
+    # Capture existing content between markers
+    CUSTOM_CONFIG=$(sed -n "/$START_MARKER/,/$END_MARKER/{ /$START_MARKER/d; /$END_MARKER/d; p; }" "$HBA_FILE")
+else
+    # Capture existing content, excluding our legacy configuration
+    # This preserves user-added rules and comments.
+    S="[[:space:]]\{1,\}"
+    CUSTOM_CONFIG=$(sed -e "/^local${S}all${S}all${S}peer/d" \
+                        -e "/^local${S}replication${S}all${S}peer/d" \
+                        -e "/^host${S}all${S}all${S}.*scram-sha-256/d" \
+                        "$HBA_FILE")
```
aaannz: I cannot say I like this at all.

Checking the documentation once again, [pg_hba.conf allows includes](https://www.postgresql.org/docs/18/auth-pg-hba-conf.html#AUTH-PG-HBA-CONF):

> include             file
> include_if_exists   file
> include_dir         directory

How about instead of this, we have our own pg_hba_mlm.conf which we include at the top of the pg_hba.conf and drop this custom lookup.

So this script instead creates that `pg_hba_mlm.conf` with our rules and checks/updates 01-pg_hba.conf to have this include as a first rule.

rjmateus:
@nadvornik, this also means that connections from the server to the `db` and `report-db`, which are internal to podman communication, will not be on SLL? It would be correct to assume that this removes the need to have `db` and `report-db` as CName for the database certificate for podman deployments?

I think we also need to change the java code to not enforce ssl when using the local names (`db` and `report-db`): https://github.com/uyuni-project/uyuni/blob/master/java/core/src/main/java/com/redhat/rhn/common/conf/ConfigDefaults.java#L1004-L1005

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
nadvornik: yes, see above

nadvornik:
> @nadvornik, this also means that connections from the server to the `db` and `report-db`, which are internal to podman communication, will not be on SLL? It would be correct to assume that this removes the need to have `db` and `report-db` as CName for the database certificate for podman deployments?
> 
yes, removing this requirement is the next step


> I think we also need to change the java code to not enforce ssl when using the local names (`db` and `report-db`): https://github.com/uyuni-project/uyuni/blob/master/java/core/src/main/java/com/redhat/rhn/common/conf/ConfigDefaults.java#L1004-L1005

Isn't this controlled by `db_ssl_enabled` from rhn.conf? I did not notice any problem with java.

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
aaannz: Right, sorry I missed that. So I am for changing the healthcheck to use socket and not have localhost there.

But it is still strange. On my system I have:

> $ cat /var/lib/pgsql/data/pg_hba.conf
> local all all peer
> local replication all peer
> host all all uyuni-server scram-sha-256
> hostssl all all all scram-sha-256
> host all all all reject

And

> bash-4.4# pg_isready -U postgres -h localhost -p 5432
> localhost:5432 - accepting connections
> bash-4.4# pg_isready -U susemanager -h localhost -p 5432
> localhost:5432 - accepting connections
> bash-4.4#

rjmateus:
> Isn't this controlled by db_ssl_enabled from rhn.conf? I did not notice any problem with java.

That is a good point, it disabling the SSL for both databases would not add the sll to the connection with the `verify-full` option.

> yes, removing this requirement is the next step

After those changes, and disabling the `db_ssl_enabled` and `report_db_ssl_enabled` what would be missing to remove the requirement for DNames on the DB certificate?
I'm just curious, because I don't know what would be missing, besides adapting the mgradm tool to not add them to the self-sign certificates.

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
nadvornik: Our config can't be included on the top, the custom rules must go before our rules to have some effect.

I am not sure if we want to give the users that much freedom - add any rule at any place -that can easily break things.

What about other way around - including custom config from the man config managed by us?

In any case we have to handle migration and split our rules and custom rules from existing configs.

nadvornik:
> After those changes, and disabling the `db_ssl_enabled` and `report_db_ssl_enabled` what would be missing to remove the requirement for DNames on the DB certificate? I'm just curious, because I don't know what would be missing, besides adapting the mgradm tool to not add them to the self-sign certificates.

Do not add them to generated certificates and do not check for them in 3rd party certificates.

rjmateus:
> Do not add them to generated certificates and do not check for them in 3rd party certificates.

Ok, so the missing part would be to adapt uyuni-tool to cope with these changes, and not add or check by names. Thank you Vladimir.

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
aaannz: Why can't we import our rules on the top? We are not adding any reject here.
So we pass our rules, if it matches fine. If not, then customers rules and if nothing matched, it will be rejected.

I don't know hot to properly handle migrations here. How do you match what is ours and what not?

nadvornik: > Why can't we import our rules on the top? We are not adding any reject here. So we pass our rules, if it matches fine. If not, then customers rules and if nothing matched, it will be rejected.

Use cases:
- disable reportdb completely if it is not needed - that must go to the top
- adding access to the main db (to specified host or globally) - any order

Anything else?

> 
> I don't know hot to properly handle migrations here. How do you match what is ours and what not?

The sed command above removes the rules added by us. What remains is the current custom config.

aaannz: > disable reportdb completely if it is not needed - that must go to the top

That should be done on the firewall level. Just don't allow anything to the database

> The sed command above removes the rules added by us. What remains is the current custom config.

You mean that matches all those strings we have from 5.0?

aaannz: We should IMO have strictest rules for our features - than includes access to the reportdb as it is our feature.

pg_hba should be adapted only to make those rules less strict if they need it.

aaannz (approved):

Issue #11705, opened 2026-04-01
Remove defaults for the sso configuration


Ladeburger:
The SSO Configuration defaults in /usr/share/rhn/config-defaults/rhn_java_sso.conf  including an optional Patameter called onelogin.saml2.security.requested_authncontext which is set to Password. While this Parameter is ignored by for example Chrome, edge obey them. And, as the comparison is set to exact, one needs to login to his System with a password, other mechanism (yubikey, windows hello, ..) will NOT work. Setting this Parameter Empty in /etc/rhn/rhn.conf did not work also, it seems as the configurations get merged. My current "workaround" for this is to edit the defaults File and comment them, but this is by far no real solution. As these Parameters are optional and 
not required for a working SSO Configuration i would like to have them removed from the defaults file, letting the user configure them at whish in the rhn.conf file

These are the Parameters in Question
```
# Authentication context.
# Set Empty and no AuthContext will be sent in the AuthNRequest
# You can set multiple values (comma separated them)

onelogin.saml2.security.requested_authncontext = urn:oasis:names:tc:SAML:2.0:ac:classes:Password

# Allows the authn comparison parameter to be set, defaults to 'exact'
onelogin.saml2.security.requested_authncontextcomparison = exact
```

Sahitya0805:
Hi, I’d like to work on this issue. Could you please assign it to me?

PR #11950, opened 2026-05-14
Stop forcing password-based SAML auth by default

officialasishkumar:
## What does this PR change?

The shipped SSO defaults at `/usr/share/rhn/config-defaults/rhn_java_sso.conf` force the SAML SP to request the `Password` AuthnContext with comparison `exact`. Strict SAML clients such as Microsoft Edge honor the request literally, so the IdP is required to authenticate the user with a password, breaking modern flows the IdP would otherwise allow (Yubikey, FIDO2, Windows Hello). Trying to disable the behavior by setting the option to an empty string in `/etc/rhn/rhn.conf` is reported in #11705 not to work because the empty value is dropped during config merging and the default keeps applying.

The OneLogin SAML toolkit only emits an `AuthnContextClassRef` in the `AuthnRequest` when `requested_authncontext` is configured. This PR comments the two `requested_authncontext` / `requested_authncontextcomparison` entries out so the default behaves as "let the IdP choose" — which is the documented intent of the surrounding comment ("Set Empty and no AuthContext will be sent in the AuthNRequest").

Administrators who need the previous strict behavior can still uncomment the lines or set the properties in `/etc/rhn/rhn.conf` under the `java.sso.*` namespace, which is the supported override mechanism for `SSOConfig`.

The accompanying inline comments in the config file are rewritten to spell out the new default behavior and the override mechanism.

## End-User Impact & Release Notes

Change in config defaults, see changelog entry.

- [x] **DONE**

## Codespace

## GUI diff

No difference.

The change is in the default SAML SP configuration only; the Uyuni Web UI is unchanged. Users will simply find that passwordless SSO logins through stricter SAML clients (notably Microsoft Edge) now succeed without manually editing the defaults.

- [x] **DONE**

## Documentation

The defaults are documented inline in the same config file with the rewritten comments.

- [x] **DONE**

## Test coverage
- No tests: already covered

`SSOControllerTest` exercises `Saml2Settings` assembly through `SettingsBuilder`; that path is unchanged by this PR (it only commentes out two optional defaults). There is no behavior to add a unit test against — the test would be checking that a property the toolkit no longer sees is no longer set.

- [x] **DONE**

## Links

Issue(s): #11705

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

A delta entry under `java/spacewalk-java.changes.officialasishkumar.remove-sso-authncontext-defaults` describes the user-visible change.

### DIFF
--- a/java/conf/rhn_java_sso.conf
+++ b/java/conf/rhn_java_sso.conf
@@ -144,12 +144,18 @@ onelogin.saml2.security.want_assertions_encrypted = false
 onelogin.saml2.security.want_nameid_encrypted = false
 
 # Authentication context.
-# Set Empty and no AuthContext will be sent in the AuthNRequest
-# You can set multiple values (comma separated them)
-onelogin.saml2.security.requested_authncontext = urn:oasis:names:tc:SAML:2.0:ac:classes:Password
+# Optional. If set, the IdP is asked to authenticate the user with one of
+# the listed contexts (e.g. urn:oasis:names:tc:SAML:2.0:ac:classes:Password).
+# When unset, no AuthnContext is sent and the IdP is free to pick the most
+# appropriate authentication method, so passwordless options such as
+# Yubikey, FIDO2 or Windows Hello keep working with strict SAML clients
+# (e.g. Microsoft Edge). Multiple contexts can be comma separated.
+# onelogin.saml2.security.requested_authncontext = urn:oasis:names:tc:SAML:2.0:ac:classes:Password
 
-# Allows the authn comparison parameter to be set, defaults to 'exact'
-onelogin.saml2.security.requested_authncontextcomparison = exact
+# Comparison parameter for the requested AuthnContext. Defaults to 'exact'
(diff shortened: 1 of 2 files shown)


admd (approved):
There is risk that user who are already using it with the current defaults and relies on the Password constraint for security policy or compliance audit reasons will lose that enforcement after upgrading. But they can revert  to previous behaviour by uncommenting those lines  so should be ok.

I will add a note in release notes.

cbbayburt:
@officialasishkumar, thank you for your contribution. As a rule, all commits to the repo must be signed with a valid GPG key. Please see the contribution guide for details: https://github.com/uyuni-project/uyuni/wiki/Contributing#always-sign-your-git-commits

Once you rebase your branch with signed commits, we can proceed to merge the PR.

officialasishkumar:
Rebased the branch and pushed a GPG-signed commit in 1bd94be.

cbbayburt (approved):

PR #8658, opened 2024-04-24
Chained container proxies configuration refactor

rjpmestre:
## What does this PR change?

Introduces the Chain of Responsibility pattern to streamline the creation of a proxy container configuration file. This  enhances code readability as we extracted a portion of code from the SystemManager class (which was dedicated to this flow) and relocated it into a dedicated structures with clear and defined steps. Additionally, it also improves testability as each step has a narrow scope.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: we commented each object/step with their purpose

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Related:
- https://github.com/uyuni-project/uyuni/pull/8634
- https://github.com/uyuni-project/uyuni/issues/8157

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "javascript_lint"
- [ ] Re-run test "spacecmd_unittests"

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

PR #11953, opened 2026-05-14
Add unified services to merge/clone erratas and/or packages

rjpmestre:
## What does this PR change?

This PR was created to address card https://github.com/SUSE/spacewalk/issues/28058 and introduces three new API endpoints for channel/errata synchronization:

1. `channel.software.syncErrataWithPackagesFromSource`
2. `channel.software.syncErrataWithPackagesFromOriginal`  
3. `channel.software.syncErrataWithPackagesFromVendor`

These replace the fragmented `channel.software.mergeErrata`, `channel.software.mergePackages`, `errata.clone`, `errata.cloneAsync`, `errata.cloneAsOriginal` and `errata.cloneAsOriginalSync` endpoints by providing:

- **Granular control**: Choose to sync erratas only, packages only, or both
- **Flexible filtering**: Filter by advisory names, date ranges, or both
- **Async support**: Execute large syncs in background without timing out
- **Consistent interface**: Same parameters and response structure across all three

---

### 1. syncErrataWithPackagesFromSource

**Description**: Sync erratas and/or packages from an explicit source channel to a target channel

**Use Case**: Merging content between any two channels (eg, dev -> staging, vendor -> custom, etc.)

**Signature**:
```
syncErrataWithPackagesFromSource(
    sessionKey,
    sourceChannelLabel,      // Source channel (required)
    targetChannelLabel,      // Target channel (required)
    operation,               // "ERRATA_ONLY" | "PACKAGES_ONLY" | "ERRATA_AND_PACKAGES"
    async,                   // Execute asynchronously
    forceRefresh,            // Force cache refresh
    filters,                 // {advisoryNames: [...], startDate: Date, endDate: Date}
    options                  // {alignPackageModules: boolean}
)
```

**Brief summary of how it works**:
- Direct lookup from source channel
- Syncs erratas and/or packages based on operation. If any filters are used, then considers the matching erratas and/or associated packages. Otherwise, considers ALL packages in the source channel.

---

### 2. syncErrataWithPackagesFromOriginal

**Description**: Sync erratas and/or packages to a cloned channel from its original channel

**Use Case**: Updating cloned channels with content from their originals

**Signature**:
```
syncErrataWithPackagesFromOriginal(
    sessionKey,
    targetChannelLabel,      // Target cloned channel (required, must be cloned)
    operation,               // "ERRATA_ONLY" | "PACKAGES_ONLY" | "ERRATA_AND_PACKAGES"
    async,                   // Execute asynchronously
    forceRefresh,            // Force cache refresh
    filters                  // {advisoryNames: [...], startDate: Date, endDate: Date}
)
```

**Brief summary of how it works**:
- Resolves original channel from clone relationship
- Walks up clone tree to find which channel contains the errata (using cascading org resolution: original org -> user org -> vendor)
- Syncs all packages from each errata in the original channel

---

### 3. syncErrataWithPackagesFromVendor

**Description**: Sync vendor erratas and/or packages to user channels. It only considers erratas and packages that are newer than the ones already present in the destination channel

**Use Case**: Applying vendor security updates to custom channels

**Signature**:
```
syncErrataWithPackagesFromVendor(
    sessionKey,
    targetChannelLabel,      // Target channel (required)
    operation,               // "ERRATA_ONLY" | "PACKAGES_ONLY" | "ERRATA_AND_PACKAGES"
    async,                   // Execute asynchronously
    forceRefresh,            // Force cache refresh
    filters                  // {advisoryNames: [...], startDate: Date, endDate: Date}
)
```

**How it works**:
- Walks up clone tree to determine suitable erratas (using cascading org resolution: original org -> user org -> vendor)
- Syncs packages where target already has matching name+arch
- Prevents adding unwanted vendor packages

---

## Common Flow

All three services share the same execution pattern:

### 1. Validation Phase
- Validate request fields (required parameters, date range consistency)
- Validate channels access and user permissions
- Service specific validations:
  - **FromSource**: Validate source channel exists
  - **FromOriginal**: Validate target is a cloned channel, original exists
  - **FromVendor**: No source validation needed
- Check for concurrent async operations on same channel

### 2. Async Dispatch (ie async=true)
- Publish event to MessageQueue (`SyncFrom[Source|Vendor|Original]ErrataEvent`)
- Return empty response immediately
- Event handler executes sync later with async=false

### 3. Sync Execution (ie if async=false OR the event action getting executing)

**Errata sync** (if operation includes erratas):
- Resolve erratas to sync (depending on the strategy:`Source`, `Vendor`, or `Original`)
- Filter out erratas already in target
- Clone missing erratas to target channel via `ErrataManager.cloneErrataForOrg()` or `ErrataManager.cloneErrata()`

**Package sync** (if operation includes packages):
- Calculate packages to sync (depending on the strategy:`Source`, `Vendor`, or `Original`)
- Filter out packages already in target
- Add packages to target channel
- Refresh newest packages

**Post-sync**:
- Align modular metadata if requested (FromSource only)
- Force cache refresh if requested

### 4. Return Response
- Return `SyncResponse` with sets of synced erratas and packages
- If async=true, both sets are empty (sync happens in background)

---

## Differences

**FromSource**:
- Considers all erratas directly associated with a source channel and filters them
- If filters applied: considers only packages associated with matching erratas
- If no filters: considers all packages from source channel (even if they are not associated to an errata)

**FromOriginal**:
- Target channel must be cloned
- Resolves erratas cascading up the clone tree, up to vendor (channel's org -> User's org -> Vendor (later overrides earlier))
- For each errata, finds which channel in tree contains it
- Gets all packages from that errata in that channel

**FromVendor**:
- Target channel doen't need to be cloned
- Resolves erratas cascading up the clone tree, up to vendor (original channel's org -> User's org -> Vendor (later overrides earlier))
- For each errata, finds packages in that errata
- Only considers packages where target channel has matching name+arch

---

## Summary Table

| Aspect | FromSource | FromOriginal | FromVendor |
|--------|------------|--------------|------------|
| **Source channel** | Required (explicit) | Inferred from clone | Vendor channel (automatic) |
| **Target requirements** | Any channel | Must be cloned | Any channel |
| **Errata resolution** | Direct lookup from source | Cascading org: original org → user org → vendor | Cascading org: original org → user org → vendor |
| **Package strategy** | With filters: errata-associated<br>No filters: ALL source packages | All packages from errata in original channel | Only packages with matching name+arch in target |
| **Deprecating EndPonts** | `channel.software.mergeErrata` | `errata.cloneAsOriginal*` | `errata.clone*` |
|  | `channel.software.mergePackages` | `errata.cloneAsOriginalAsync` | `errata.cloneAsync` |

---

## Deprecated Services

1. All deprecated APIs now delegate to the new service implementations.
2. `channel.software.mergeErrata` was async by default. **Is now sync**.

---

## Architecture Flow

The typical workflow for these services is as follows:

```
XML-RPC Handler (ChannelSoftwareHandler)
    ↓
Parse parameters → Build ErrataCriteria + SyncRequest
    ↓
Service Interface (SyncFrom[Source|Vendor|Original]Service)
    ↓
ServiceImpl (SyncFrom[Source|Vendor|Original]ServiceImpl)
    ↓
┌─────────────────────────────────────────────────────────┐
│ Validation (ChannelSoftwareValidationHelper)            │
│  - Request fields (dates, operation, etc.)              │
│  - Channel existence & user access                      │
│  - Service-specific rules (cloned, source exists, etc.) │
│  - No concurrent async jobs                             │
└─────────────────────────────────────────────────────────┘
    │
    │             ┌───────────────────────────────────────────────┐
 async -> true -> │ Publish Event to MessageQueue                 │
    │             │ (SyncFrom[Source|Vendor|Original]ErrataEvent) │
    ↓             │ Return empty SyncResponse                     │
  false           └───────────────────────────────────────────────┘
    │                 │
    │                 ↓
    │             ┌────────────────────────────────────────────────┐
    │             │ Later: EventAction executes                    │
    │             │ (SyncFrom[Source|Vendor|Original]ErrataAction) │
    │             │ Return empty SyncResponse                      │
    │             └────────────────────────────────────────────────┘
    │                 │
    ↓                 ↓
┌─────────────────────────┐
│  1. Errata Resolution   │
│  2. Clone erratas       │
│  3. Package Resolution  │
│  4. Clone packages      │
└─────────────────────────┘
    │
    ↓
Return SyncResponse(erratas, packages)
```

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28058

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
```diff
@@ -1323,5 +1325,120 @@ public static ErrataOverview cloneErratum(Long originalEid, String advisory, Str
...
+     *   - If org is null: returns ONLY vendor erratas (org_id IS NULL)
+     *
+     * @param org the organization to filter by (ignored if channel is provided, mutually exclusive with channel)
+     * @param channel the channel to filter by (mutually exclusive with org)
+     * @param advisories list of advisory names to filter by (optional)
+     * @param startDate filter erratas issued on or after this date (optional)
+     * @param endDate filter erratas issued on or before this date (optional)
+     * @return Set of matching erratas matching the criteria
+     */
+    public static Set<Errata> lookupErrataByChannelOrOrg(
```
CDellaGiusta: please put this as `private` to avoid potential misuse, especially with `org` and `channel` arguments

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
```diff
@@ -1323,5 +1325,120 @@ public static ErrataOverview cloneErratum(Long originalEid, String advisory, Str
...
+            List<String> advisories,
+            Date startDate,
+            Date endDate
+    ) {
+        StringBuilder hql = new StringBuilder("SELECT DISTINCT e FROM Errata e");
+        Map<String, Object> params = new HashMap<>();
+
+        // Mutually exclusive channel / org filter
+        if (channel != null) {
+            hql.append(" WHERE :channel IN elements(e.channels) ");
```
CDellaGiusta, line 1412: Just wondering if it's worth putting `assert(null==org) `just to avoid misuse?

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
```diff
@@ -1323,5 +1325,120 @@ public static ErrataOverview cloneErratum(Long originalEid, String advisory, Str
...
+
+        if (endDate != null) {
+            hql.append(" AND e.lastModified < :endDate ");
+            params.put("endDate", endDate);
+        }
+
+        Query<Errata> query = getSession().createQuery(hql.toString(), Errata.class);
+        params.forEach(query::setParameter);
+
+        return new HashSet<>(query.list());
```
CDellaGiusta, line 1445: why not returning a `List<Errata>` so the last line could be `return query.list()` ?

CDellaGiusta (dismissed):
LGTM, please just have a look at the couple of minor nitpicks I suggested

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
rjpmestre: Sure, but lets throw an `IllegalArgumentException` instead. `assert` is usually discouraged

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
rjpmestre: The callers (of the callers) are using these results to perform multiple `.remove()` operations (like `ErrataManager.filterErrataRequiringMerge(...)` or `ErrataResolver.resolveFromCascadingOrg(...) `).
One could then argue then why not letting the callers convert that. And my argument is that this method is already preparing it to be a set anyway because query already has `SELECT DISTINCT e FROM Errata e`. In the end its all about performance.

java/core/src/main/java/com/suse/spec/channel/software/dto/ErrataCriteria.java
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (c) 2026 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+package com.suse.spec.channel.software.dto;
```
cbbayburt: I'm not sure about these new packages `impl/spec`. Even in the `com.suse.` path, we organize packages per feature, unless I've missed something.

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/serializer/SyncErrataWithPackagesResponseSerializer.java
```diff
@@ -0,0 +1,78 @@
...
+ *                          the gpg key it was signed with.")
+ *              #struct_end()
+ *          #array_end()
+ *     #struct_end()
+ */
+public class SyncErrataWithPackagesResponseSerializer extends ApiResponseSerializer<SyncErrataWithPackagesResponse> {
+
+    @Override
+    public Class<SyncErrataWithPackagesResponseSerializer> getSupportedClass() {
+        return SyncErrataWithPackagesResponseSerializer.class;
```
cbbayburt:
```suggestion
        return SyncErrataWithPackagesResponse.class;
```

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/serializer/SyncErrataWithPackagesResponseSerializer.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public SerializedApiResponse serialize(SyncErrataWithPackagesResponse src) {
+        SerializationBuilder builder = new SerializationBuilder();
+
+        Set<Errata> erratas = new HashSet<>(src.erratas());
+        builder.add("errata", erratas);
+
+        Set<Package> packages = new HashSet<>(src.packages());
+        builder.add("package", packages);
```
cbbayburt: Array property names are defined as `erratas` and `packages` in the DTO and apidoc.
```suggestion
        Set<Errata> erratas = new HashSet<>(src.erratas());
        builder.add("erratas", erratas);

        Set<Package> packages = new HashSet<>(src.packages());
        builder.add("packages", packages);
```

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/errata/ErrataHandler.java
```diff
@@ -827,9 +837,14 @@ private List<Errata> lookupVendorAndUserErrataByAdvisoryAndOrg(String advisoryNa
...
+     * @deprecated Use syncErrataWithPackagesFromVendor instead
      */
+    @Deprecated(since = "2026.06")
     public Object[] clone(User loggedInUser, String channelLabel, List<String> advisoryNames) {
-        return clone(loggedInUser, channelLabel, advisoryNames, false, false);
+        ErrataCriteria criteria = new ErrataCriteria(advisoryNames, null, null);
+        SyncRequest request = new SyncRequest(criteria, SyncOperation.ERRATA_AND_PACKAGES, false, false, false);
+        SyncResponse response = syncFromVendorService.sync(loggedInUser, channelLabel, request);
+        return response.erratas().toArray();
     }
```
cbbayburt, line 848: These deprecated endpoints used `lookupAccessibleErratum()` for each advisory and returned a `FaultException` if any were missing. IIUC, now as long as there's one match, the missing ones are silently dropped. We should ideally not change the existing contract for the deprecated methods.

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/channel/software/ChannelSoftwareHandler.java
```diff
@@ -1751,24 +1771,33 @@ public Object[] mergeErrata(User loggedInUser, String mergeFromLabel,
...
-
-        if (!UserManager.verifyChannelAdmin(loggedInUser, mergeTo)) {
-            throw new PermissionCheckFailureException();
-        }
+            String mergeToLabel, String startDateStr, String endDateStr) {
+        Date startDate = Optional.ofNullable(startDateStr)
+                .map(d -> Date.from(
+                        LocalDate.parse(d).atStartOfDay(ZoneId.systemDefault()).toInstant()
+                ))
+                .orElse(null);
```
cbbayburt: The changed date/time handling on these deprecated methods can break existing customer scripts. I think we should try to use the old parsing (with timestamps), or at least support all formats that were supported before.

cbbayburt (changes requested):
Hi @rjpmestre, here's my review, mostly from the API perspective. There are a couple of small changes needed, listed below.

Apart from that, the sheer amount of code changes in the PR worries me a bit. I wasn't able to review it line-by-line, and since it includes both some deep refactoring and API contract changes, we should be really careful reviewing this thoroughly before merging. I'd say let's have at least a couple more reviews before proceeding.

So take my review as API-focused, not for the whole Java code.

java/core/src/main/java/com/suse/spec/channel/software/dto/ErrataCriteria.java
rjpmestre: No, you're right.

Ever since I worked on the proxy service-related PRs (such as [#8658](https://github.com/uyuni-project/uyuni/pull/8658) and [#9703](https://github.com/uyuni-project/uyuni/pull/9703)), I've been thinking about different ways to improve service organization, data flow, and the consistency of boundaries between classes in general.

I've been writing an RFC to bring these ideas up for discussion, and establish some official guidelines for developing services like these. Basically have a place where the "unwritten" rules are.. well, written :P
As I was tackling this card, I kind of found myself applying more of those ideas. So, I understand how this can come across as unnecessarily disruptive.

I think you're right, though. I'll adapt this PR to follow the current organization and keep these kind of  changes for a future PR. Maybe once we have chance to reach some conclusions from the RFC.

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/errata/ErrataHandler.java
rjpmestre: Alright. I took the chance to make it fault tolerant. In the end not only that was not the point as the caller also ends up not having a clue something has failed. I'll address it.

rjpmestre: While I was following @cbbayburt comment and adjusting these legacy calls behave as before, I noticed that  `channel.software` `mergeErrata` allows providing a list of `errataNames` but simply ignores the ones that cannot be matched ([link](https://github.com/uyuni-project/uyuni/blob/df5312ee7a856b9c62fddbe68c2d17d0b5279709/java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/channel/software/ChannelSoftwareHandler.java#L1797)).

As I mentioned, we can preserve the legacy behavior in the deprecated endpoints, no questions about it. 
But now i'm questioning myself what do we want for the new unified endpoints, when request has errata names? 
I see 3 obvious options right now:
- Be always strict and fail if any errata name is not matched 
- Ignore non matched errata names
- Add a field to the output with the list of the `missingErrataNames` 

I don't have a strong preference, on one hand i'm not a fan of aborting the operation just because a single errata name doesn't match. On the other hand, silently ignoring them means the caller to infer what happened by comparing the requested names with the ones on the output. The 3rd option feels like a compromise, but i'm open to other ideas.

What do you think? pinging @rjmateus @cbbayburt

cbbayburt: Considering the way these API endpoints are usually consumed by customer scripts, I think it's better to be strict, so deviant behavior doesn't slip unnoticed.

Most likely, the API client scripts would get the errata lists from other API endpoints and feed them into these methods. So if any are missing, that might indicate a problem with their system.

No strong opinion here, though. I'd say option 3 is not necessary, but option 2, together with a proper log message with missing names, is also a valid option IMO.

rjmateus: I think we should move forward with option 1), and the default behaviour. 
We can also have an overloaded method, with a flag named something like "ignoreMismatch", that is it's passed as true, it would behave as you put on option 2.
The method with the more restrict mode, can be the default, and would delegate the execution on the method with the boolean option set to false.
What do you think?

mcalmer: I have the feeling we should spend an option for this. Maybe `ignoreMissing` or `failMissing`
But when we ignore, we should report missing things in the result.

rjpmestre: Today we had a meeting and we reached an agreement. We'll add a boolean flag for these endpoint to control strict vs lenient processing behavior.
By default, strict mode will used. If any of the provided errata names is not matched, it fails the entire request.
Lenient mode will process all valid errata names.

Strict mode returns a list of failures, capped at a fixed constant `MAX_MISMATCH_OUTPUT` (~10-20 items). This way we avoid polluting the output.
If the limit is exceeded, we'll keep counting how many more failed to be informative. The output will append "N failures and more"
The goal was to keep feedback useful but bounded.

Lenient mode should also return what failed, returning a list of unprocessed errata alongside the success response. Same limit and "N failures and more" pattern applies as in strict mode.


Note: from this meeting we also agreed creating [this card](https://github.com/SUSE/spacewalk/issues/31927) for investigating current endpoints behaviour

PR #11956, opened 2026-05-15
Add RKE2 TFTP container sanity tests

ilmanzo:
## What does this PR change?

The TFTP service now runs as a dedicated Kubernetes pod (proxy-tftpd image) instead of the removed tftpsync packages. Add two scenarios to verify the new container:

- Smoke: tftp deployment readiness + service endpoint presence
- File transfer: end-to-end TFTP get via NodePort using curl tftp://, exercising the full path through the tftp pod's HTTP proxy to the uyuni pod's /srv/tftpboot

Also adds `get_tftp_node_port()` helper to `kubernetes.rb`

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/29930

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/kubernetes/srv_rke2_tftpd_sanity.feature
+++ b/testsuite/features/kubernetes/srv_rke2_tftpd_sanity.feature
@@ -1,0 +1,19 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@rke2
+Feature: RKE2 TFTP container sanity checks on the server
+  In order to provision systems via PXE boot
+  As the system administrator
+  I want to ensure the TFTP container is correctly deployed and serving files
+
+  Scenario: TFTP deployment is ready and service has active endpoints
+    Given The Kubernetes cluster is ready on "server"
+    And the "tftp" deployment on "server" should become ready within 5 minutes
+    Then the "tftp" service on "server" should have at least one active endpoint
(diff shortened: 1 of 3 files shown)


testsuite/features/kubernetes/srv_rke2_tftpd_sanity.feature
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@rke2
+Feature: RKE2 TFTP container sanity checks
```
srbarrios: As tftp is running in both server and proxy, I would either make the sanity check in both, either describe on the title that we aim to test the server's one.

srbarrios (dismissed):

Pablogoliva (dismissed):

testsuite/features/kubernetes/srv_rke2_tftpd_sanity.feature
ilmanzo: I'd stay with just the server to start simple. When needed the tests can be easily extended to the proxy in a second time (just add new line in the feature with a different target)

NamelessOne91 (approved):

Pablogoliva (approved):

maximenoel8 (approved):

PR #11958, opened 2026-05-15
Entity Address default constructor - 2 (bsc#1259951)

rjpmestre:
## What does this PR change?

This PR addresses the missing default values on AddressImpl which should have been part of https://github.com/uyuni-project/uyuni/pull/11681

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30099

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
@@ -68,6 +68,16 @@ public class AddressImpl extends BaseDomainHelper implements Address {
      * Protect the constructor
      */
     protected AddressImpl() {
+        this.address1 = " ";
+        this.address2 = " ";
+        this.city = " ";
+        this.state = " ";
+        this.zip = " ";
+        this.country = " ";
+        this.phone = " ";
+        this.fax = " ";
+        this.isPoBox = "0";
+        this.privType = TYPE_MARKETING;
     }
 
     /**


mackdk (commented):
This change feels wrong to me, I'd rather avoid initializing the object with empty fields. 

If the UI workflow allows it, a better fix would be to remove the factory method `UserFactory.createAddress()` and replace it with a public constructor that requires all mandatory fields (specifically those with `NOT NULL` constraints in the DB).

Of course, if we’re intentionally creating an empty object to populate it later, your approach is the only immediate fix, and it's totally fine to have it in now. In this case, I'd suggest to create a card in our board to refactor this logic, so we can tackle the problem in a separate PR.

rjpmestre:
> This change feels wrong to me, I'd rather avoid initializing the object with empty fields.
> 
> If the UI workflow allows it, a better fix would be to remove the factory method `UserFactory.createAddress()` and replace it with a public constructor that requires all mandatory fields (specifically those with `NOT NULL` constraints in the DB).
> 
> Of course, if we’re intentionally creating an empty object to populate it later, your approach is the only immediate fix, and it's totally fine to have it in now. In this case, I'd suggest to create a card in our board to refactor this logic, so we can tackle the problem in a separate PR.

That was my exact reading as well. IIRC, it looks like initially, it was designed to force users to have an address, but later on they must have realized that maybe it wasn’t a good idea to push new users (especially during installation) filling these sort of data.

So, yeah, this would be just a temporary fix.

When I initially got into it and notice this stuff, I started reworking this whole thing, but at a point I realized I was starting to look a lot more as an update and..probably would require more testing than this type of fix would need.

But I’m planning to get back to it!

PR #12006, opened 2026-05-22
Check object before using the in operator (bsc#1264967)

bisht-richa:
## What does this PR change?

**Fix Page Salt > Formula Catalog page bug**

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: 

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30673
Port(s): [# **5.1**](https://github.com/SUSE/spacewalk/pull/31543)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

- [x] Re-run test "frontend_checks"

# Before you merge

### DIFF
--- a/web/html/src/components/table/Table.tsx
+++ b/web/html/src/components/table/Table.tsx
@@ -223,7 +223,7 @@ export const Table = forwardRef<TableRef, TableProps>((props, ref) => {
                 onClick={() => expanded.toggle(props.identifier(item))}
                 cell={() => {
                   const hasChildren =
-                    item && typeof item === "object" && "children" in item && item.children.length > 0;
+                    !!item && typeof item === "object" && "children" in item && item.children.length > 0;
                   const isExpanded = expanded.has(props.identifier(item));
                   return (
                     <i
--- a/web/html/src/components/table/Table.tsx
+++ b/web/html/src/components/table/Table.tsx
@@ -21,7 +21,9 @@ import { useExpanded } from "./useExpanded";
 
 type ArrayElement<A> = A extends readonly (infer T)[] ? T : never;
 function hasChildRows(item: any): item is { children: any[] } {
-  return "children" in item && Array.isArray(item.children) && item.children.length > 0;
+  return (
+    item && typeof item === "object" && "children" in item && Array.isArray(item.children) && item.children.length > 0
+  );
 }
 
 export type TableLoadInfo = {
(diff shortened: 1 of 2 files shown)


web/html/src/components/table/Table.tsx
```diff
@@ -21,7 +21,9 @@ import { useExpanded } from "./useExpanded";
 
 type ArrayElement<A> = A extends readonly (infer T)[] ? T : never;
 function hasChildRows(item: any): item is { children: any[] } {
-  return "children" in item && Array.isArray(item.children) && item.children.length > 0;
+  return (
+    item && typeof item === "object" && "children" in item && Array.isArray(item.children) && item.children.length > 0
+  );
 }
 
 export type TableLoadInfo = {
```
EMaksy, line 29: Just as an idea:
`item && ...` is functionally fine, but `&&` does not always return a boolean in JavaScript. It returns the first falsy operand, or the last operand if all are truthy. So if `item` is `null` or `undefined`, `item && ...` returns that value instead of `false`.

Using `!!item && ...` normalizes the first check to a boolean, which matches this type guard’s boolean return.

```ts
function hasChildRows(item: any): item is { children: any[] } {
  return !!item && typeof item === "object" && Array.isArray(item.children) && item.children.length > 0;
}
```

Reference:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND

[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!)


What do you think  @bisht-richa  ?

web/html/src/components/table/Table.tsx
```diff
@@ -21,7 +21,9 @@ import { useExpanded } from "./useExpanded";
 
 type ArrayElement<A> = A extends readonly (infer T)[] ? T : never;
 function hasChildRows(item: any): item is { children: any[] } {
```
EMaksy, line 23: Just curious  why we use "any" as type ?

bisht-richa: Good question. This was already in the existing code, so I kept the same typing while working on the changes. I didn't look into the type here, but we could probably improve it later if needed.

EMaksy: Yeah for sure, i thought maybe there is a historical reason or something :)

EMaksy (approved):
The code change lgtm 👍 
Thank you!

mcalmer (approved):

PR #12032, opened 2026-05-28
Fix PostgreSQL peer auth failure when POSTGRES_USER != postgres

admd:
## What does this PR change?

When `POSTGRES_USER` is not '`postgres`', initdb creates a superuser with a different name (e.g. dbadmin), leaving no 'postgres' DB role. After` 01-pg_hba.sh` activates peer auth via` pg_ctl reload`, all Unix socket connections are restricted to the DB role matching the OS user, which is always '`postgres`'. Scripts using` -U $POSTGRES_USER` therefore fail
authentication.

Fix by creating the 'postgres' role in 01-pg_hba.sh before the reload. The check connects as 'postgres' first: success means the role already exists (upgrade/restart path); failure means trust auth is still in memory (fresh install), so the role is created via POSTGRES_USER.

Update 02-sys-database-collation.sh, get_pgsql_disk_severity.sh and zz-evr_t.sh to use -U postgres on socket connections, since peer auth is active when these scripts run.

--
When `POSTGRES_USER` is set to any value other than `postgres` (e.g. `dbadmin` in the test suite, or a user-defined value in production), the `uyuni-db` container crashes on every start after the initial database setup.

This was the flow

1. `initdb` is called with `--username=$POSTGRES_USER`, creating that user as the cluster superuser. The `postgres` DB role does **not** exist.
2. `01-pg_hba.sh` (introduced in c2fef597, reloading added in 9d22ee08) rewrites `pg_hba.conf` and calls `pg_ctl reload`,
   activating peer auth on the Unix socket:
                    `local all all peer`

3. Peer auth maps the connecting OS user to the DB role of the same name. The postgres process always runs as OS user `postgres`, so Unix socket connections can only authenticate as DB role `postgres`.

4. All subsequent upgdb scripts (`02-sys-database-collation.sh`, `get_pgsql_disk_severity.sh`, `zz-evr_t.sh`) use
`-U $POSTGRES_USER` on the socket. When `POSTGRES_USER != postgres` this fails peer auth and the container exits.

This PR create the `postgres` DB role before calling `pg_ctl reload`. The approach avoids any dependency on TCP or `PGPASSWORD`:

Try connecting as `postgres` via socket.
- **Success** → role already exists (upgrade/restart path, peer auth active) → skip creation.
- **Failure** → role does not exist; trust auth is still in memory (fresh install path) → create the role via `POSTGRES_USER`, which works under trust.

`02-sys-database-collation.sh`, `get_pgsql_disk_severity.sh`, `zz-evr_t.sh` Change socket connections from `-U $POSTGRES_USER` to `-U postgres`. These scripts run after `01-pg_hba.sh` has already activated peer auth, so `$POSTGRES_USER` can never work on the socket unless it equals `postgres`. Using `-U postgres` is both correct and honest about the peer auth constraint.

I Verified locally using the podman test suite runner (`testsuite/podman_runner/07_server_setup.sh`) with `POSTGRES_USER=dbadmin`, which previously caused the container to crash immediately after the
initdb phase.

## Codespace

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered
- Unit tests were added
- Cucumber tests were added

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
+++ b/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
@@ -68,6 +68,16 @@ if [[ "$(< "$HBA_FILE")" != "$NEW_CONFIG" ]]; then
     PG_HBA_CHANGED=1
 fi
 
+# Ensure 'postgres' role exists for local socket connections under peer auth.
+# initdb creates the cluster superuser as POSTGRES_USER (not 'postgres'), so peer
+# auth would fail without a matching 'postgres' role for the OS user.
+# If we can connect as 'postgres' the role already exists; otherwise trust auth
+# is still in memory (fresh install) so we can create it via POSTGRES_USER.
+if ! PGHOST= PGHOSTADDR= psql -p "${PGPORT:-5432}" -U postgres --no-password --no-psqlrc -c "" 2>/dev/null; then
+    PGHOST= PGHOSTADDR= psql -p "${PGPORT:-5432}" -U "${POSTGRES_USER}" --no-psqlrc \
+        -c "CREATE ROLE postgres SUPERUSER LOGIN;"
+fi
+
 # restart database to pickup new pg_hba changes
 if [ -n "$PG_HBA_CHANGED" ]; then
     pg_ctl -D "$PGDATA" reload
--- a/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/02-sys-database-collation.sh
+++ b/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/02-sys-database-collation.sh
@@ -6,7 +6,7 @@
 function run_sql() {
     PGHOST='' PGHOSTADDR='' psql -v ON_ERROR_STOP=1 \
         -p "${PGPORT:-5432}" \
-        -U "${POSTGRES_USER:-postgres}" \
+        -U postgres \
         --no-password --no-psqlrc --tuples-only --no-align "$@"
 }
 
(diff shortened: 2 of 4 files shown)


admd:
@Pablogoliva can you test these changes to verify the change at your end too.

admd:
@aaannz @nadvornik I only fixed it more keeping in the testsuit part, please review it carefully that I didn't break any other scenerio. I needed quite some help from the gemini on this one.

aaannz:
For 5.1 I set `local all all trust` instead of peer, so anyone with access to the socket has access to the database.
Which is IMO ok, as anyone gaining access to the database container can easily do a `su postgres` so having it set to peer is only making things harder on us.

I am not sure how hardcoded postgres user will cope with upgrades where people used different superuser.

mcalmer (approved):

admd:
> For 5.1 I set `local all all trust` instead of peer, so anyone with access to the socket has access to the database. Which is IMO ok, as anyone gaining access to the database container can easily do a `su postgres` so having it set to peer is only making things harder on us.
> 
> I am not sure how hardcoded postgres user will cope with upgrades where people used different superuser.

Ondrej, wouldn't that case be covered with the [else block]( https://github.com/uyuni-project/uyuni/pull/12032/changes#diff-173889f5e3b404a227a636954ac0d0c63b430cb6ee9d35c7deded573832416a4R77) that 

```
PGHOST= PGHOSTADDR= psql -p "${PGPORT:-5432}" -U "${POSTGRES_USER}" --no-psqlrc \
        -c "CREATE ROLE postgres SUPERUSER LOGIN;"
```

admd:
@aaannz what also works in my testing env is creating a temporary hba file to allow trust auth during upgrades and after upgrade, remove that. That looks more cleaner and avoid all this hardcoding and role part.

But as I said, this one is tricky so feel free to come with a better solution please. At least this PR should help with identifiying the issue. There must be better, more complete solution.

mbussolotto (approved):
Seems ok to me

aaannz:
I would prefer https://github.com/uyuni-project/uyuni/pull/12035

admd:
thank you Ondrej, that looks simple and clean. If you don't have any concerns to not use peer authentica and just use trust, we are good. 

I will close this PR. :+1:

PR #12036, opened 2026-05-29
SonarCloud error reduction fix step 16, miscellaneous 8 rules

CDellaGiusta:
## What does this PR change?

SonarCloud error reduction fix step 16, miscellaneous 8 rules

SonarCloud fix: java:S1149 Replace StringBuffer with unsynchronized class StringBuilder
SonarCloud fix: java:S1488 Local variables not immediately returned
SonarCloud fix: java:S2583 Conditionally executed code should be reachable
SonarCloud fix: java:S2692 indexOf checks should not be for positive numbers
SonarCloud fix: java:S5164 ThreadLocal variables should be cleaned up when no longer used
SonarCloud fix: java:S2226 (AdminListAction) Servlets should not have mutable instance fields
SonarCloud fix: java:S2226 (ProxyClientsAction) Servlets should not have mutable instance fields
SonarCloud fix: java:S3655 Optional value should only be accessed after calling isPresent()

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links

Port(s):
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/frontend/action/kickstart/ScheduleKickstartWizardAction.java
+++ b/java/core/src/main/java/com/redhat/rhn/frontend/action/kickstart/ScheduleKickstartWizardAction.java
@@ -72,6 +72,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import jakarta.servlet.http.HttpServletRequest;
@@ -550,22 +551,22 @@ public class ScheduleKickstartWizardAction extends RhnWizardAction {
                 if (rec != null && rec.getProfile() != null && profile.getName().equals(rec.getProfile().getName())) {
                     if (StringUtils.isBlank(form.getString(KERNEL_PARAMS_TYPE))) {
                         form.set(KERNEL_PARAMS_TYPE, KERNEL_PARAMS_CUSTOM);
-                        if (rec.getKernelOptions().isEmpty()) {
-                            form.set(KERNEL_PARAMS, CobblerObject.INHERIT_KEY);
-                        }
-                        else {
-                            form.set(KERNEL_PARAMS, profile.convertOptionsMap(rec.getKernelOptions().get()));
-                        }
+
+                        String kernelOptions = rec.getKernelOptions()
+                                .map(profile::convertOptionsMap)
+                                .orElse(CobblerObject.INHERIT_KEY);
+
+                        form.set(KERNEL_PARAMS, kernelOptions);
                     }
 
                     if (StringUtils.isBlank(form.getString(POST_KERNEL_PARAMS_TYPE))) {
                         form.set(POST_KERNEL_PARAMS_TYPE, KERNEL_PARAMS_CUSTOM);
-                        if (rec.getKernelOptionsPost().isEmpty()) {
-                            form.set(POST_KERNEL_PARAMS, CobblerObject.INHERIT_KEY);
-                        }
-                        else {
-                            form.set(POST_KERNEL_PARAMS, profile.convertOptionsMap(rec.getKernelOptionsPost().get()));
-                        }
+
+                        String kernelOptionsPost = rec.getKernelOptionsPost()
+                                .map(profile::convertOptionsMap)
+                                .orElse(CobblerObject.INHERIT_KEY);
+
+                        form.set(POST_KERNEL_PARAMS, kernelOptionsPost);
                     }
                 }
             }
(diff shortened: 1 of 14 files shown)


java/core/src/main/java/com/suse/manager/webui/websocket/RemoteMinionCommands.java
```diff
@@ -405,7 +406,7 @@ private WebSession getWebSession() {
      * will result in IllegalStateException.
      */
     private void sendMessage(Session session, AbstractSaltEventDto dto) {
-        synchronized (session) {
+        synchronized (lockObj) {
```
mackdk: I get why you're doing this, but here there's a catch: the `session` object is passed as parameter, this means multiple concurrent calls of `sendMessage` are fine, as long as they are executed on different `session` object. So, a single instance of `RemoteMinionCommands` can handle multiple `session` concurrently. 

If we switch to the `lockObj` field, multiple calls of `sendMessage` with different `session`s will be handled one at a time, because the synchronization will now be at `RemoteMinionCommands` instance level.

I don't like synchronizing on the `session` either (mainly because it's an object out of our scope and we don't know what else might be locking on it), but I'm not sure if this change is totally ok without a proper concurrent multi-threaded test.

java/core/src/main/java/com/redhat/rhn/frontend/action/systems/sdc/ProxyClientsAction.java
```diff
@@ -95,23 +93,26 @@ public ActionForward execute(ActionMapping mapping, ActionForm formIn,
...
                     clients.add(client);
                 }
             }
         }
         return clients;
     }
+
+    @Override
+    protected DataResult<?> getDataResult(User user, PageControl pc, ActionForm formIn) {
+        //dummy implementation, never used since execute() is overridden and calls getDataResultImpl
```
mackdk, line 115: If we are overriding directly `execute` does it still make sense that this class extends `BaseSystemsAction` or would it work the same if it extended directly `RhnListAction`?

java/core/src/main/java/com/redhat/rhn/frontend/action/kickstart/ScheduleKickstartWizardAction.java
```diff
@@ -549,22 +550,24 @@ protected void addRequestAttributes(RequestContext ctx, KickstartScheduleCommand
...
+                        Optional<Map<String, Object>> kernelOptions = rec.getKernelOptions();
                         form.set(KERNEL_PARAMS_TYPE, KERNEL_PARAMS_CUSTOM);
-                        if (rec.getKernelOptions().isEmpty()) {
+                        if (kernelOptions.isEmpty()) {
                             form.set(KERNEL_PARAMS, CobblerObject.INHERIT_KEY);
                         }
                         else {
-                            form.set(KERNEL_PARAMS, profile.convertOptionsMap(rec.getKernelOptions().get()));
+                            form.set(KERNEL_PARAMS, profile.convertOptionsMap(kernelOptions.get()));
                         }
```
mackdk: Here and in all places in this file, I'd recommend to use the `map()` and `orElse()` instead of doing an `if` on`isEmpty()`, since it kinds of defeats the purpose of using `Optional`:
```java
String kernelOptions = rec.getKernelOptions()
    .map(options -> profile.convertOptionsMap(options))
    .orElse(CobblerObject.INHERIT_KEY);

form.set(KERNEL_PARAMS, kernelOptions);
```

mackdk (commented):
Looks mostly good. I would leave out of this comment the change of synchronization in `RemoteMinioncommands`. While it's indeed better and safer than the original implementation, I think it needs a specific unit test coverage to ensure its correctness.

java/core/src/main/java/com/suse/manager/webui/websocket/RemoteMinionCommands.java
CDellaGiusta: ok, I dropped the commit addressing this. I will remove it from sonar cloud warnings as well

java/core/src/main/java/com/redhat/rhn/frontend/action/systems/sdc/ProxyClientsAction.java
CDellaGiusta: unfortunately not.
`ProxyClientsAction.execute()` uses `getSetDecl` and `SHOW_NO_SYSTEMS` which are defined in `BaseSystemsAction`

java/core/src/main/java/com/redhat/rhn/frontend/action/kickstart/ScheduleKickstartWizardAction.java
CDellaGiusta: as per our chat, this solution does not work for  `addKernelParamsAttribute` and `addPostKernelParamsAttribute` methods. I created a card to fix the issue: https://github.com/SUSE/spacewalk/issues/30916

mackdk (dismissed):

PR #12061, opened 2026-06-01
Fix Channel/Errata/Package bidirectional relationship mappings

rjpmestre:
## What does this PR change?

While working on https://github.com/SUSE/spacewalk/issues/28058 i noticed inconsistent behavior when adding/removing erratas and packages to channels. 
Further investigation revealed these NxN relationships  between Channel/Errata/Package had incorrect hibernate mappings that violated best practices.
These are the sort of issues that may get more relevant after upgrading hibernate 5 to  7 upgrade (because it got stricter about enforcing these specifications).

Two main problems currently are:
- Both sides declaring @JoinTable. Hibernate expects one of the side to own the relationship. 
  I set the ownership hierarchy to be like: channel owns errata and package, then errata owns package
- Inappropriate cascade on peer associations. This has more to do with our usage pattern. We usually create these objects separately and link then, instead of passing the new objects for automatic persistence. As example, we dont' usually:
```
 Channel channel = new Channel();
 Errata errata = new Errata(); 
 channel.addErrata(errata);
```

instead , we usually do 
```
Channel channel = ChannelFactory.createChannel();  // Created separately
Errata errata = ErrataFactory.createErrata();     // Created separately
channel.addErrata(errata);  
```

As a follow up, this PR also updates best practices in managing/updating these relationships.
Adding now an element to these relationships like `channel.getErrata().add(errata);` will mean we'll also have to update the other side of the relationship, in this case with `errata.getChannels().add(channel);`

While doable, this it's an error prone process.  Best practice would be NOT use the direct `.getX().add(...)` (and the following reverse update), but delegate this responsibility to an `add`  method to the owner side. In this example, then, `Channel` entity we should have
``` 
public void addErrata(Errata errataIn) {
        if (erratas.add(errataIn)) { // Only sync if actually added
            errataIn.getChannels().add(this); // Only sync if actually added
        }
    }
```
Same concept for removing elements or clear.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/28058

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
@@ -189,13 +189,13 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
             inverseJoinColumns = @JoinColumn(name = "org_trust_id"))
     private Set<Org> trustedOrgs;
 
-    @ManyToMany(fetch = FetchType.LAZY)
+    @ManyToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST}, fetch = FetchType.LAZY)
     @JoinTable(name = "rhnChannelErrata",
             joinColumns = @JoinColumn(name = "channel_id"),
             inverseJoinColumns = @JoinColumn(name = "errata_id"))
-    private Set<Errata> erratas;
+    private Set<Errata> erratas = new HashSet<>();
 
-    @ManyToMany(fetch = FetchType.LAZY)
+    @ManyToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST}, fetch = FetchType.LAZY)
     @JoinTable(name = "rhnChannelPackage",
             joinColumns = @JoinColumn(name = "channel_id"),
             inverseJoinColumns = @JoinColumn(name = "package_id"))
@@ -604,41 +604,12 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
     }
 
     /**
-     * Sets the erratas set for this channel
-     * @param erratasIn The set of erratas
-     */
-    public void setErratas(Set<Errata> erratasIn) {
-        this.erratas = erratasIn;
-    }
-
-    /**
      * Adds a single errata to the channel
      * @param errataIn The errata to add
      */
     public void addErrata(Errata errataIn) {
-        if (erratas.add(errataIn)) {
-            errataIn.getChannels().add(this);
-        }
-    }
-
-    /**
-     * Removes a single errata from the channel
-     * @param errataIn The errata to remove
-     */
-    public void removeErrata(Errata errataIn) {
-        if (erratas.remove(errataIn)) {
-            errataIn.getChannels().remove(this);
-        }
-    }
-
-    /**
-     * Adds a single package to the channel
-     * @param packageIn The package to add
-     */
-    public void addPackage(Package packageIn) {
-        if (packages.add(packageIn)) {
-            packageIn.getChannels().add(this);
-        }
+        erratas.add(errataIn);
+        errataIn.getChannels().add(this);
     }
 
     /**
@@ -652,6 +623,16 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
     }
 
     /**
+     * Removes a single errata from the channel
+     * @param errataIn The errata to remove
+     */
+    public void removeErrata(Errata errataIn) {
+        erratas.remove(errataIn);
+        errataIn.getChannels().remove(this);
+    }
+
+
+    /**
      * Removes multiple erratas from the channel
      * @param erratasIn The collection of erratas to remove
      */
@@ -659,6 +640,22 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
         for (Errata errata : erratasIn) {
             removeErrata(errata);
         }
+    }
+
+    /**
+     * Clears out the errata associated with this channel.
+     */
+    public void clearErratas() {
+        removeErratas(new ArrayList<>(getErratas()));
+    }
+
+    /**
+     * Adds a single package to the channel
+     * @param packageIn The package to add
+     */
+    public void addPackage(Package packageIn) {
+        packages.add(packageIn);
+        packageIn.getChannels().add(this);
     }
 
     /**
@@ -672,14 +669,22 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
     }
 
     /**
+     * Removes a package from the packages set.
+     * @param packageIn The package to remove.
+     */
+    public void removePackage(Package packageIn) {
+        packages.remove(packageIn);
+        packageIn.getChannels().remove(this);
+    }
+
+
+    /**
      * Removes multiple packages from the channel
      * @param packagesIn The collection of packages to remove
      */
     public void removePackages(Collection<Package> packagesIn) {
         for (Package pkg : packagesIn) {
-            if (packages.remove(pkg)) {
-                pkg.getChannels().remove(this);
-            }
+            removePackage(pkg);
         }
     }
 
@@ -710,15 +715,6 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
         return ChannelFactory.getErrataCount(this);
     }
 
-
-    /**
-     * Sets the packages set for this channel
-     * @param packagesIn The set of erratas
-     */
-    public void setPackages(Set<Package> packagesIn) {
-        this.packages = packagesIn;
-    }
-
     /**
      *
      * @param sourcesIn The set of yum repo sources
@@ -733,15 +729,6 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      */
     public Set<ContentSource> getSources() {
         return sources;
-    }
-
-    /**
-     * Removes a package from the packages set.
-     * @param packageIn The package to remove.
-     */
-    public void removePackage(Package packageIn) {
-        packages.remove(packageIn);
-        packageIn.getChannels().remove(this);
     }
 
     /**
--- a/java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
@@ -27,6 +27,7 @@ import com.redhat.rhn.domain.user.User;
 import com.redhat.rhn.manager.channel.ChannelManager;
 import com.redhat.rhn.manager.system.SystemManager;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
@@ -37,6 +38,7 @@ import org.hibernate.type.YesNoConverter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
@@ -193,13 +195,13 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
     @JoinTable(name = "rhnChannelErrata",
             joinColumns = @JoinColumn(name = "channel_id"),
             inverseJoinColumns = @JoinColumn(name = "errata_id"))
-    private Set<Errata> erratas = new HashSet<>();
+    private Set<Errata> erratas;
 
     @ManyToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST}, fetch = FetchType.LAZY)
     @JoinTable(name = "rhnChannelPackage",
             joinColumns = @JoinColumn(name = "channel_id"),
             inverseJoinColumns = @JoinColumn(name = "package_id"))
-    private Set<Package> packages = new HashSet<>();
+    private Set<Package> packages;
 
     @ManyToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST}, fetch = FetchType.LAZY)
     @JoinTable(name = "rhnChannelContentSource",
@@ -600,7 +602,7 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @return Returns the set of erratas for this channel.
      */
     public Set<Errata> getErratas() {
-        return erratas;
+        return Collections.unmodifiableSet(erratas);
     }
 
     /**
@@ -608,8 +610,11 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param errataIn The errata to add
      */
     public void addErrata(Errata errataIn) {
+        if (errataIn == null) {
+            return;
+        }
         erratas.add(errataIn);
-        errataIn.getChannels().add(this);
+        errataIn.addChannelInternal(this);
     }
 
     /**
@@ -617,9 +622,10 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param erratasIn The collection of erratas to add
      */
     public void addErratas(Collection<Errata> erratasIn) {
-        for (Errata errata : erratasIn) {
-            addErrata(errata);
+        if (CollectionUtils.isEmpty(erratasIn)) {
+            return;
         }
+        erratasIn.forEach(this::addErrata);
     }
 
     /**
@@ -627,8 +633,11 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param errataIn The errata to remove
      */
     public void removeErrata(Errata errataIn) {
+        if (errataIn == null) {
+            return;
+        }
         erratas.remove(errataIn);
-        errataIn.getChannels().remove(this);
+        errataIn.removeChannelInternal(this);
     }
 
 
@@ -637,9 +646,10 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param erratasIn The collection of erratas to remove
      */
     public void removeErratas(Collection<Errata> erratasIn) {
-        for (Errata errata : erratasIn) {
-            removeErrata(errata);
+        if (CollectionUtils.isEmpty(erratasIn)) {
+            return;
         }
+        erratasIn.forEach(this::removeErrata);
     }
 
     /**
@@ -654,8 +664,11 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param packageIn The package to add
      */
     public void addPackage(Package packageIn) {
+        if (packageIn == null) {
+            return;
+        }
         packages.add(packageIn);
-        packageIn.getChannels().add(this);
+        packageIn.addChannelInternal(this);
     }
 
     /**
@@ -663,9 +676,10 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param packagesIn The collection of packages to add
      */
     public void addPackages(Collection<Package> packagesIn) {
-        for (Package pkg : packagesIn) {
-            addPackage(pkg);
+        if (CollectionUtils.isEmpty(packagesIn)) {
+            return;
         }
+        packagesIn.forEach(this::addPackage);
     }
 
     /**
@@ -673,8 +687,11 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param packageIn The package to remove.
      */
     public void removePackage(Package packageIn) {
+        if (packageIn == null) {
+            return;
+        }
         packages.remove(packageIn);
-        packageIn.getChannels().remove(this);
+        packageIn.removeChannelInternal(this);
     }
 
 
@@ -683,9 +700,17 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
      * @param packagesIn The collection of packages to remove
      */
     public void removePackages(Collection<Package> packagesIn) {
-        for (Package pkg : packagesIn) {
-            removePackage(pkg);
+        if (CollectionUtils.isEmpty(packagesIn)) {
+            return;
         }
+        packagesIn.forEach(this::removePackage);
+    }
+
+    /**
+     * Clears out the packaged associated with this channel.
+     */
+    public void clearPackages() {
+        removePackages(new ArrayList<>(getPackages()));
     }
 
     /**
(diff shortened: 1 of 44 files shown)


java/core/src/main/java/com/redhat/rhn/domain/errata/Errata.java
CDellaGiusta: I would put this as = = new HashSet<>(); as well for uniformity

CDellaGiusta: ditto: I would put this as = = new HashSet<>(); as well for uniformity

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -188,17 +189,17 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
             inverseJoinColumns = @JoinColumn(name = "org_trust_id"))
     private Set<Org> trustedOrgs;
 
-    @ManyToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST}, fetch = FetchType.LAZY)
+    @ManyToMany(fetch = FetchType.LAZY)
     @JoinTable(name = "rhnChannelErrata",
             joinColumns = @JoinColumn(name = "channel_id"),
             inverseJoinColumns = @JoinColumn(name = "errata_id"))
     private Set<Errata> erratas;
```
CDellaGiusta, line 198: Could it be possible to set it as an empty set, so to avoid "null" checks all around the code? My guess it that you should do this for all member sets/collections

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -615,7 +616,71 @@ public void setErratas(Set<Errata> erratasIn) {
...
+        }
+    }
+
+    /**
+     * Adds a single package to the channel
+     * @param packageIn The package to add
+     */
+    public void addPackage(Package packageIn) {
+        if (packages.add(packageIn)) {
+            packageIn.getChannels().add(this);
```
CDellaGiusta: in all your getX().add(Y) you should ensure that getX() is not going to return null (or check for it). Please do choose a pattern and use it everywhere (for uniformity)

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -615,7 +616,71 @@ public void setErratas(Set<Errata> erratasIn) {
...
+
+    /**
+     * Adds multiple erratas to the channel
+     * @param erratasIn The collection of erratas to add
+     */
+    public void addErratas(Collection<Errata> erratasIn) {
+        for (Errata errata : erratasIn) {
+            addErrata(errata);
+        }
+    }
```
CDellaGiusta, line 660: I would move the "multiple" versions close to the "single" ones (e.g. `addErratas` moved close to `addErrata`)

java/core/src/main/java/com/redhat/rhn/domain/errata/Errata.java
```diff
@@ -804,6 +808,30 @@ public void setPackages(Set<Package> p) {
...
+        for (Package pkg : packagesIn) {
+            addPackage(pkg);
+        }
+    }
+
+    /**
+     * Clears out the Packages associated with this errata.
+     */
+    public void clearPackages() {
+        if (this.getPackages() != null && !this.getPackages().isEmpty()) {
```
CDellaGiusta: can getPackages be null? Also, if this.getPackages() is empty, the mechanism works anyway

java/core/src/main/java/com/redhat/rhn/domain/errata/Errata.java
```diff
@@ -834,13 +862,19 @@ public String toString() {
...
-        if (this.getChannels() != null) {
+        if (this.getChannels() != null && !this.getChannels().isEmpty()) {
+            List<Channel> channelsCopy = new ArrayList<>(this.getChannels());
+            for (Channel channel : channelsCopy) {
+                channel.removeErrata(this);
+            }
             this.getChannels().clear();
         }
         Iterator<ErrataFile> i = IteratorUtils.getIterator(this.getFiles());
         while (i.hasNext()) {
```
CDellaGiusta: since you're here ... 
`for (ErrataFile pf : this.getFiles()) {`

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
```diff
@@ -232,7 +230,7 @@ public static List<Errata> addToChannel(List<Errata> errataList, Channel chan,
                                             User user, boolean inheritPackages, boolean performPostActions) {
         List<com.redhat.rhn.domain.errata.Errata> toReturn = new ArrayList<>();
         for (Errata errata : errataList) {
-            errata.addChannel(chan);
+            chan.addErrata(errata);
```
CDellaGiusta, line 233: question: are these two operations equivalent? Same happens in the tests

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
```diff
@@ -284,7 +282,7 @@ public static List<Errata> addToChannel(List<Errata> errataList, Channel chan,
      */
     public static void addToChannel(Errata errata, Channel chan, User user,
                                       Set<Package> packages) {
-        errata.addChannel(chan);
+        chan.addErrata(errata);
```
CDellaGiusta, line 285: ditto

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
```diff
@@ -1109,7 +1107,7 @@ public static void syncErrataDetails(ClonedErrata cloned) {
         cloned.setAdvisoryStatus(original.getAdvisoryStatus());
 
         // Copy the packages
-        cloned.setPackages(new HashSet<>(original.getPackages()));
```
CDellaGiusta, line 1112: my understanding at this point is that nowhere in the code `setPackages` should be called, and it should be replaced by `replacePackages`. If this is right, probably it's worth put `setPackages` as private/protected, so that it can be used only by hibernate? This way you can also spot leftovers

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/errata/ErrataHandler.java
```diff
@@ -1171,7 +1171,7 @@ public Errata create(User loggedInUser, Map<String, Object> errataInfo,
             newErrata.addKeyword(keyword);
         }
 
-        newErrata.setPackages(new HashSet<>());
```
CDellaGiusta, line 1174: if this method is private/protected you can also spot leftovers like this one (see previous comment)

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -188,17 +189,17 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
             inverseJoinColumns = @JoinColumn(name = "org_trust_id"))
     private Set<Org> trustedOrgs;
 
-    @ManyToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST}, fetch = FetchType.LAZY)
+    @ManyToMany(fetch = FetchType.LAZY)
```
CDellaGiusta: question: is this modification persisting the relationship? (when you change the erratas, are they now persisted?). This question is valid for all db changes you put in place in the relationships

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -188,17 +189,17 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
...
     @JoinTable(name = "rhnChannelErrata",
             joinColumns = @JoinColumn(name = "channel_id"),
             inverseJoinColumns = @JoinColumn(name = "errata_id"))
     private Set<Errata> erratas;
 
-    @ManyToMany(cascade = {CascadeType.MERGE, CascadeType.PERSIST}, fetch = FetchType.LAZY)
+    @ManyToMany(fetch = FetchType.LAZY)
```
CDellaGiusta: ditto (persistence)

java/core/src/main/java/com/redhat/rhn/manager/errata/cache/ErrataCacheManager.java
```diff
@@ -266,16 +264,16 @@ public static void insertCacheForChannelErrataAsync(List<Long> channelIdsToUpdat
...
-     * @param errataId ID of the errata to update the cache for. Assumes the errata is published
+     * @param errata the errata to update the cache for. Assumes the errata is published
      */
-    public static void addErrataRefreshing(Collection<Long> channelIdsToUpdate, Long errataId) {
-        for (Long cid : channelIdsToUpdate) {
-            ChannelFactory.addErrataToChannel(Set.of(errataId), cid);
-            List<Long> pids = ErrataFactory.listErrataChannelPackages(cid, errataId);
-            ErrataCacheManager.insertCacheForChannelPackages(cid, errataId, pids);
-            ErrataManager.replaceChannelNotifications(errataId, cid, new Date());
+    public static void addErrataRefreshing(Set<Channel> channelToUpdate, Errata errata) {
```
CDellaGiusta: better 'channelsToUpdate' (with ending 'S')?

java/core/src/test/java/com/redhat/rhn/domain/channel/ChannelTest.java
```diff
@@ -177,6 +178,62 @@ public void testIsBaseChannel() {
...
+        // Assert the errata also contains the channel
+        assertEquals(1, errata.getChannels().size());
+        assertEquals(channel, errata.getChannels().iterator().next());
+
+        // Remove the errata
+        channel.removeErrata(errata);
+
+        assertTrue(channel.getErratas().isEmpty());
+        assertTrue(errata.getChannels().isEmpty());
+    }
```
CDellaGiusta, line 268: I would add to these tests the checks if the channel/errata are really saved/persisted, to be sure that we're not losing associations in the database

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
mackdk: Indeed, I was wondering the same. Do we have some unit test checking the cascading behavoiur of these entities. If not, I think we should add some before doing this refactoring because to me it feels we are changing how it works in this PR.

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -188,17 +189,17 @@ public class Channel extends BaseDomainHelper implements Comparable<Channel> {
...
     @JoinTable(name = "rhnChannelPackage",
             joinColumns = @JoinColumn(name = "channel_id"),
             inverseJoinColumns = @JoinColumn(name = "package_id"))
-    private Set<Package> packages;
+    private Set<Package> packages = new HashSet<>();
```
mackdk: Seems to me this was already initialized in the default constructor

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -615,7 +616,71 @@ public void setErratas(Set<Errata> erratasIn) {
      * @param errataIn The errata to add
      */
     public void addErrata(Errata errataIn) {
-        erratas.add(errataIn);
+        if (erratas.add(errataIn)) {
```
mackdk: Here and in the other `add` methods I think we should also check the parameter for `null`. `HashSet` is going to accept `null` values but I don't think hibernate will be happy with them.

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -615,7 +616,71 @@ public void setErratas(Set<Errata> erratasIn) {
...
+    }
+
+    /**
+     * Adds multiple erratas to the channel
+     * @param erratasIn The collection of erratas to add
+     */
+    public void addErratas(Collection<Errata> erratasIn) {
+        for (Errata errata : erratasIn) {
+            addErrata(errata);
+        }
```
mackdk: Or simply `erratasIn.forEach(this::addErrata);` 

Both versions are not null safe, though, so we could check it to ensure it's fully safe:

```java
        if (CollectionUtils.isEmpty(erratasIn)) {
            return;
        }

        erratasIn.forEach(this::addErrata);
```

Same applies to the other methods working on collections.

java/core/src/main/java/com/redhat/rhn/domain/errata/Errata.java
mackdk: I agree, but I recommend creating a default constructor and initialize all the fields there, since we have more than one to initialize

java/core/src/main/java/com/redhat/rhn/domain/errata/Errata.java
mackdk: I agree, the null check seems redundant now. If it is, we can remove it, otherwise please invert this if and use `CollectionsUtils.isEmpty()` to avoid nesting the method main logic

java/core/src/main/java/com/redhat/rhn/domain/errata/Errata.java
```diff
@@ -834,13 +862,19 @@ public String toString() {
      * Clears out the Channels associated with this errata.
      */
     public void clearChannels() {
-        if (this.getChannels() != null) {
+        if (this.getChannels() != null && !this.getChannels().isEmpty()) {
```
mackdk: Same applies here

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
mackdk: I agree. With the pattern implemented in this PR the `package`/`errata` collections should be mutated only using the new add/remove methods. I think the `setPackages`/`setErratas` are not even needed by hibernate since we are using field-based access (annotations are on the fields). In general with this approach, the `getPackages()`/`getErratas()` should also return immutable collections, to ensure only the owner modifies it, but I don't think we can afford to be that strict with our codebase 😅

java/core/src/test/java/com/redhat/rhn/domain/errata/ErrataTestBuilder.java
```diff
@@ -0,0 +1,343 @@
...
+    private List<Package> packages = new ArrayList<>();
+    private List<Channel> channels = new ArrayList<>();
+    private List<Cve> cves = new ArrayList<>();
+    private List<Bug> bugs = new ArrayList<>();
+
+    public ErrataTestBuilder() {
+        this.errata = new Errata();
+    }
+
+    public ErrataTestBuilder orgId(Long orgIdIn) {
```
mackdk, line 71: Not sure if we have a convention for naming here, but I think most builder in our codebase use the `with...` pattern for simple fields. Not that I have any preference, though.

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/errata/ErrataHandler.java
rjpmestre: This would be a clear.
I ended up removing it  the direct sets

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
rjpmestre: We usually update entities that are already managed by hibernate. And, for managed entities, hibernate tracks and handles them.
So the cascade only applies for transient or detached entities, if we were doing something like `channel.getPackages().add(new Package(...))`

Well, In any case, I ended up rolling this back.

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
rjpmestre: I ended up removing the setter for these collections and replace their usages for proper operations that reflect the intention like `.addPackage(...)`, '.clearChannels()", etc

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
rjpmestre: Sorry, didn't get your mean here. Can you clarify?

java/core/src/test/java/com/redhat/rhn/domain/errata/ErrataTestBuilder.java
rjpmestre: I went with the shorter builder style here since I've did use it it a few times elsewhere in the codebase already :\
I just find it a bit cleaner due to the context of a builder pattern.

java/core/src/main/java/com/redhat/rhn/domain/errata/ErrataFactory.java
CDellaGiusta: I mean: has adding a channel to an errata the very same effects of adding an errata to a channel? if not, you cannot invert the logic

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
CDellaGiusta: can `errataIn` be null?

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -615,7 +616,71 @@ public void setErratas(Set<Errata> erratasIn) {
...
+            errataIn.getChannels().add(this);
+        }
+    }
+
+    /**
+     * Removes a single errata from the channel
+     * @param errataIn The errata to remove
+     */
+    public void removeErrata(Errata errataIn) {
+        if (erratas.remove(errataIn)) {
```
CDellaGiusta: ditto can `errataIn` be null?

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -671,14 +736,12 @@ public Set<ContentSource> getSources() {
...
+     * Removes a package from the packages set.
+     * @param packageIn The package to remove.
      */
-    public void removePackage(Package packageIn, User user) {
-            List<Long> list = new ArrayList<>();
-            list.add(packageIn.getId());
-            ChannelManager.removePackages(this, list, user);
+    public void removePackage(Package packageIn) {
+        packages.remove(packageIn);
+        packageIn.getChannels().remove(this);
```
CDellaGiusta: can `packa`geIn be nuill?

java/core/src/main/java/com/redhat/rhn/domain/errata/Errata.java
CDellaGiusta: I would remove the iterator in favour of a for loop

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
```diff
@@ -671,14 +736,12 @@ public Set<ContentSource> getSources() {
...
-     * Removes a single package from the channel
-     * @param user the user doing the remove
-     * @param packageIn The package to remove
+     * Removes a package from the packages set.
+     * @param packageIn The package to remove.
      */
-    public void removePackage(Package packageIn, User user) {
-            List<Long> list = new ArrayList<>();
-            list.add(packageIn.getId());
-            ChannelManager.removePackages(this, list, user);
```
CDellaGiusta, line 681: the old call to `ChannelManager.removePackages(this, list, user);` was actually testing `UserManager.verifyChannelAdmin` . Did you avoid this check on purpose?
Also, did you check that the `remove_packages`  query did the same effect? If so, can you remove the query from Channel_queries.xml?

java/core/src/test/java/com/redhat/rhn/domain/errata/ErrataFactoryTest.java
```diff
@@ -246,66 +238,11 @@ public static Errata createTestErrata(Long orgId) throws Exception {
...
-        ef = ErrataFactory.createErrataFile(ErrataFactory.
-                lookupErrataFileType("RPM"),
-                    "SOME FAKE CHECKSUM: 123456789012",
-                    "test errata file" + TestUtils.randomString(), errataFilePackages);
-
-        e.addFile(ef);
-        e.setSeverity(Severity.getById(1));
+    public static Errata createTestErrata(Long orgId, Optional<String> advisory) {
+        ErrataTestBuilder builder = new ErrataTestBuilder().orgId(orgId).addDummyPackage(true);
+        advisory.ifPresent(builder::advisory);
```
CDellaGiusta, line 244: well, you are missing to fill the errata with all the stuff in fillOutErrata, so you are creating an "empty" errata here? For example this test errata has no errata file.... not sure you're testing the same thing

CDellaGiusta (dismissed):
LGTM but please check you addressed all comments

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
rjpmestre: Well, it can. I'd usually rather have that responsibility on the caller's side. Of course there could be exceptions but I think failing silently would just be masking a bug in caller's logic.
I'm not sure we're we stand on this in terms of best practices so I'll add the null check on these

java/core/src/main/java/com/redhat/rhn/domain/channel/Channel.java
rjpmestre: I was ok dropping `UserManager.verifyChannelAdmin` here because the only usage of this `removePackage` here was in this class test.
Other services that perform package removal would do it directly via `channel.getPackages().remove(pkg)` without invoking through this method. Now, this brings a bit the point i mentioned earlier on other reply about the entity roles/scope. 
I see this as an operation that should be handled at the service layer, where the use case is defined, before invoking `removePackage`.

java/core/src/test/java/com/redhat/rhn/domain/errata/ErrataFactoryTest.java
rjpmestre: Oh! No, I added those fields as the default values in `ErrataTestBuilder`.
I just wanted a bit of flexibility creating erratas for testing purposes but I was ok with the default values.

mackdk (dismissed):

cbbayburt (dismissed):
LGTM 👍

PR #12075, opened 2026-06-05
Get rid of strutstestcase dependency

CDellaGiusta:
## What does this PR change?

this PR aims at getting rid of the dependency with test project strutstestcase

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30950
Port(s): # none (5.2)
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


rjpmestre (dismissed):
LGTM. Happy to let it go :)

java/core/src/test/java/com/redhat/rhn/frontend/action/configuration/ChannelOverviewActionTest.java
```diff
@@ -88,7 +86,7 @@ public void testCreateChannelWithSpacesInLabel() {
 
     private ConfigChannel doCreateChannelAction(String channelLabel) {
         // create a config channel via action and return it
-        request.setMethod(HttpServletRequestSimulator.POST);
+        request.setMethod("POST");
```
rjmateus, line 89: Should we use a constant instead of a hardcoded string? Since we use spark framework, maybe use `spark.route.HttpMethod`?
Or maybe create a constant data for it?

rjmateus (dismissed):
Looks good to me. Just one small idea that I would like to hear your opinion.

java/core/src/test/java/com/redhat/rhn/frontend/action/configuration/ChannelOverviewActionTest.java
CDellaGiusta: Initially I had in mind to create a `setPostMethod` and `setGetMethod` methods. However a couple of SparkTestUtils methods use to pass it as a string, so they should be refactored as well.
One possible tradeoff could be to define  a `setPostMethod` and `setGetMethod` and substitute them when the call is explicit, and then define a string constant in RhnMockHttpServletRequest.
Not sure whether the effort is worth, though, since we're talking of testing files that in line of principle should go away with time.

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+public class BaseStrutsTestCase {
+
+    private static final String LOCATION_KEY = "Location";
+
+    public class StrutsTestHttpServletResponse extends RhnMockHttpServletResponse {
```
mackdk: For all these inner classes I'd rather have them `static` and pass the dependencies through the constructor rather referencing the outer class fields directly and keep the reference to the test. In this way we don't have to worry about memory leaks.

Moreover, these should not be `public`. You can either have them `private` and keep them here or move them outside and have them package visible.

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+            return requestDispatcher;
+        }
+
+        @Override
+        public InputStream getResourceAsStream(String string) {
+            try {
+                File file = new File(string);
+
+                if (!file.exists()) {
+                    file = new File(new File(".").getAbsolutePath() + StrutsTestCaseUtility.prependSlash(string));
```
mackdk: Please use `Strings.CS.prependIfMissing(string, "/")` from `commons-lang3`, instead of custom implementation.

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+                }
+
+                if (file.exists()) {
+                    return new FileInputStream(file);
+                }
+                else {
+                    return getClass().getResourceAsStream(StrutsTestCaseUtility.prependSlash(string));
+                }
```
mackdk: Please remove the redundant `else` and use `Strings.CS.prependIfMissing(string, "/")`

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+        setRequestPathInfo("", pathInfo);
+    }
+
+    public void setRequestPathInfo(String moduleName, String pathInfo) {
+        actionPath = Optional.ofNullable(stripActionPath(pathInfo));
+
+        if (moduleName != null) {
+            if (!moduleName.isEmpty()) {
+                moduleName = StrutsTestCaseUtility.prependSlash(moduleName);
+                moduleName = StrutsTestCaseUtility.appendSlash(moduleName);
```
mackdk: Same here:
```suggestion
                moduleName = Strings.CS.prependIfMissing(moduleName, "/");
                moduleName = Strings.CS.appendIfMissing(moduleName, "/");
```

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+        StrutsTestCaseUtility.verifyNoActionMessages(request);
+    }
+
+    public ActionForm getActionForm() {
+        ActionConfig actionConfig = StrutsTestCaseUtility.getActionConfig(actionPath.orElse(null),
+                request, servletContext);
+        if (isRequest(actionConfig)) {
+            return (ActionForm) request.getAttribute(actionConfig.getAttribute());
+        }
+        else {
```
mackdk: Please remove this redundant `else`

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+
+                if (file.exists()) {
+                    return new FileInputStream(file);
+                }
+                else {
+                    return getClass().getResourceAsStream(StrutsTestCaseUtility.prependSlash(string));
+                }
+            }
+            catch (Exception e) {
+                System.out.println(e.getMessage());
```
mackdk: Maybe we can use a logger here instead of `System.out`?

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
```diff
@@ -0,0 +1,334 @@
...
+            return StrutsTestCaseUtility.removeSessionId(response.getHeader(LOCATION_KEY));
+        }
+        else {
+            try {
+                return request.getContextPath() + StrutsTestCaseUtility.removeSessionId(getForward());
+            }
+            catch (NullPointerException e) {
+                return null;
+            }
+        }
```
mackdk, line 336: I think we should avoid catching `NullPointerException` and do a proper null check:
```java
if (response.containsHeader(LOCATION_KEY)) {
    return StrutsTestCaseUtility.removeSessionId(response.getHeader(LOCATION_KEY));
}

if (request == null) {
    return null;
}

return StringUtils.defaultString(request.getContextPath()) + StringUtils.defaultString(StrutsTestCaseUtility.removeSessionId(getForward()));
```
This code is not fully equivalent to the previous, but I don't think we want to have `null/Index.do` if `request.getContextPath()` is `null` and and `getForward()` is `/Index.do`

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,276 @@
...
+
+    public static void verifyActionMessages(HttpServletRequest request, String[] expectedMessages) {
+        verifyActionMessages(request, expectedMessages, Globals.MESSAGE_KEY, ACTION_MESSAGE_LABEL);
+    }
+
+    private static void verifyNoActionMessages(HttpServletRequest request, String key, String messageLabel) {
+        ActionMessages actionMessages = (ActionMessages) request.getAttribute(key);
+        if (null == actionMessages) {
+            return;
+        }
```
mackdk: You can leverage on modern java here to enforce the type and also avoid a `ClassCastException`:

```suggestion
        if ( !(request.getAttribute(key) instanceof ActionMessages actionMessages) {
            return;
        }
```

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,276 @@
...
+        Iterator iterator = actionMessages.get();
+        if (iterator.hasNext()) {
+            StringBuilder messageText = new StringBuilder();
+            while (iterator.hasNext()) {
+                messageText.append(" ").append(((ActionMessage) iterator.next()).getKey());
+            }
+
+            fail("Got message(s) [%s] while no action messages expected of type [%s]"
+                    .formatted(messageText.toString(), messageLabel));
+        }
```
mackdk: I think this can use modern java as well:

```java
        Iterator<?> iterator = actionMessages.get();
        List<String> messageKeys = new ArrayList<>();
        iterator.forEachRemaining(m -> {
            if (m instanceof ActionMessage actionMessage) {
                messageKeys.add(actionMessage.getKey());
            }
        });

        if (!messageKeys.isEmpty()) {
            fail("Got message(s) [%s] while no action messages expected of type [%s]"
                    .formatted(String.join(" ", messageKeys), messageLabel));
        }
```

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    private static void verifyActionMessages(HttpServletRequest request, String[] expectedMessages,
+                                             String key, String messageLabel) {
+        ActionMessages actionMessages = (ActionMessages) request.getAttribute(key);
+
+        if (null == actionMessages) {
+            fail("Expected messages of type [%s], but received none.".formatted(messageLabel));
+        }
```
mackdk: Same as before:
```suggestion
        if ( !(request.getAttribute(key) instanceof ActionMessages actionMessages) {
            fail("Expected messages of type [%s], but received none.".formatted(messageLabel));
            return;
        }
```

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,276 @@
...
+        Iterator iterator = new TransformIterator(actionMessages.get(), m -> ((ActionMessage) m).getKey());
+        String[] actionMessageKeys = (String[]) IteratorUtils.toArray(iterator, String.class);
+        Arrays.sort(actionMessageKeys);
+
+        for (int i = 0; i < expectedMessages.length; i++) {
+            if (!expectedMessages[i].equals(actionMessageKeys[i])) {
+                fail("Messages not matching: expected: [%s], actual: [%s]"
+                        .formatted(expectedMessages[i], actionMessageKeys[i]));
+            }
+        }
```
mackdk: Or using modern java:
```java
Iterator<?> iterator = actionMessages.get();
List<String> actualList = StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator, 0), false)
    .flatMap(m -> {
        if (m instanceof ActionMessage actionMessage) {
            return Stream.of(actionMessage.getKey());
        }

        return Stream.empty();
    })
    .sorted()
    .toList();

List<String> expectedList = Arrays.stream(expectedMessages).sorted().toList();
assertEquals(expectedList, actualList,
    () -> "Messages not matching: expected: %s, actual: %s".formatted(expectedList, actualList));
```

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,276 @@
...
+        return stringIn;
+    }
+
+    public static String appendSlash(String stringIn) {
+        if (!stringIn.endsWith("/")) {
+            return stringIn + "/";
+        }
+        return stringIn;
+    }
```
mackdk: As already said, we should prefer `commons-lang3` methods.

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
CDellaGiusta: done

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/BaseStrutsTestCase.java
CDellaGiusta: ok, done

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: done

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: done

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: done

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: done

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: replaced and deleted

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,271 @@
...
+        assertEquals(forwardPath, actualForwardPath, "Expected and actual forward path not matching");
+    }
+
+    private static String getModuleName(HttpServletRequest request) {
+        String moduleName = "";
+        if (null != request.getServletPath()) {
+            moduleName = request.getServletPath();
+        }
+
+        if (Strings.isBlank(moduleName)) {
```
mackdk: Better use the same method from `org.apache.commons.lang3.StringUtils` rather than `org.apache.logging.log4j.util.Strings`

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,271 @@
...
+            if (null != request.getAttribute(RequestProcessor.INCLUDE_SERVLET_PATH)) {
+                moduleName = (String) request.getAttribute(RequestProcessor.INCLUDE_SERVLET_PATH);
+            }
+        }
+
+        if (Strings.isNotEmpty(moduleName)) {
```
mackdk: Same, better use `StringUtils`

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,271 @@
...
+    private static Optional<String> getForwardNameWhenTiles(ForwardConfig expectedForward, HttpServletRequest request,
+                                                            ServletContext context) {
+        String tilesForward = null;
+        ComponentDefinition definition = getTilesForward(expectedForward.getPath(), request, context);
+        if (definition != null) {
+            tilesForward = definition.getPath();
+        }
+        if (tilesForward != null) {
+            return Optional.of(request.getContextPath() +
+                    org.apache.commons.lang3.Strings.CS.prependIfMissing(tilesForward, "/"));
```
mackdk: If you use `StringUtils` above you won't need to qualify Strings here and below 🙂

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
```diff
@@ -0,0 +1,271 @@
...
+
+            int queryIndex = pathCopy.indexOf("?");
+            if (queryIndex > 0) {
+                urlString = buffer.delete(removeIndex, queryIndex).toString();
+            }
+            else {
+                urlString = buffer.delete(removeIndex, buffer.length()).toString();
+            }
+        }
+        return urlString;
```
mackdk: How about this version?
```java
        int removeIndex = Strings.CI.indexOf(urlString, ";jsessionid=");
        if (removeIndex == -1) {
            // No jsessionid to remove
            return urlString;
        }

        // Search for parameter start after jsessionid, if any, to avoid removing part of the query string
        int queryIndex = Strings.CI.indexOf(urlString, "?", removeIndex);
        if (queryIndex == -1) {
            return urlString.substring(0, removeIndex);
        }

        StringBuilder buffer = new StringBuilder(urlString);
        return buffer.delete(removeIndex, queryIndex).toString();
```
Also the checks for `removeIndex > 0` and `queryIndex > 0` feels more like bugs since the are skipping the first item, like the sonarcloud warning of the other PR. In fact, if `urlString` is exacly `";jsessionid=<whatever>"` the current method will return `";jsessionid=<whatever>"`. I doubt it will ever happen, but still I think the expected behaviour should be to remove it.

mackdk (commented):
Sorry, I've added a couple of more comments that I missed in the first review 😅

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: yes, fixed through all the occurrences in the file

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/StrutsTestCaseUtility.java
CDellaGiusta: it behaves in equivalent way + more precise. Suggestion taken and implemented

mackdk (approved):

mcalmer (approved):

PR #12077, opened 2026-06-06
Add RKE2 reposync storage benchmark

AkashKumar7902:
## What does this PR change?

Adds an optional RKE2 storage benchmark for the Uyuni testsuite using the real repository synchronization path:

```bash
spacewalk-repo-sync -c <channel> --url=file:///<mounted-source-repo>/
```

The benchmark verifies that the source repository is mounted in the Uyuni server pod, creates a unique software channel through the Uyuni API, runs `spacewalk-repo-sync` inside the server pod, and stores a compact JSON summary and raw artifacts under:

```text
/var/spacewalk/uyuni-bench/results/reposync/<backend>-<timestamp>/
```

The stable summary contains workload metadata, source and API-verified imported package counts, exit code, storage backend, monotonic wall duration, and paths to raw stdout, stderr, and optional `/usr/bin/time -v` output. It deliberately does not parse human-readable `spacewalk-repo-sync` logs for phase metrics.

Cluster-side `kubectl` commands run from the testsuite controller through `get_target('localhost').run_local`. This uses the kubeconfig installed on the controller by Sumaform and does not require SSH access to the Uyuni ingress hostname or an RKE2 node.

The benchmark is opt-in only and is not part of the normal testsuite flow. The scenario uses the standard `@long_running` tag so the normal 40-minute scenario watchdog does not terminate the configured four-hour workload.

Files added or updated:

- `testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature`
- `testsuite/features/step_definitions/storage_benchmark_steps.rb`
- `testsuite/run_sets/kubernetes_storage_benchmark.yml`

The input repository must be mounted into the server pod separately, for example:

```text
server host: /srv/uyuni-bench-source
pod path:    /uyuni-bench-source
source URL:  file:///uyuni-bench-source/leap-15.6-backports-update/
```

RPM datasets are not committed to this repository.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- No user-facing documentation needed: this is an opt-in testsuite benchmark.

- [x] **DONE**

## Test coverage

The rebased branch was checked with:

- Ruby 3.3 syntax validation;
- targeted RuboCop;
- YAML parsing and run-set/feature linkage;
- generated shell-command syntax;
- focused assertions for integer parsing, controller-local Kubernetes routing, Running/Ready pod selection;
- searches confirming there are no parser-derived metric references or server-host SSH calls in the benchmark;
- `git diff --check`.

A previous live run of the same mounted-repository workload imported 3,027 of 3,027 packages with exit code 0. The current revision also validates Sumaform external-cluster routing. A new full run still requires a healthy deployed RKE2/Uyuni environment and the source mount supplied through the Sumaform Helm values.

- [x] **DONE**

## Links

Issue(s): openSUSE/mentoring#252  
Port(s): none

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
+++ b/testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
@@ -1,0 +1,18 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+
+@rke2
+@scope_storage_benchmark
+@benchmark
+Feature: RKE2 storage benchmark
+  In order to compare storage backends for Uyuni on Kubernetes
+  As the system administrator
+  I want to run a real Uyuni repository synchronization workload
+
+  Scenario: Run a local file repository synchronization benchmark
+    Given The Kubernetes cluster is ready on "server"
+    And the reposync benchmark source repository is mounted in the server pod
+    When I create a unique reposync benchmark channel
+    And I run the reposync benchmark for the mounted source repository
+    Then the reposync benchmark should finish successfully
--- a/testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
+++ b/testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
@@ -1,0 +1,16 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+
+@rke2
+@long_running
+Feature: RKE2 storage benchmark
+  In order to compare storage backends for Uyuni on Kubernetes
+  As the system administrator
+  I want to run a real Uyuni repository synchronization workload
+
+  Scenario: Run a local file repository synchronization benchmark
+    Given the reposync benchmark source repository is mounted in the server pod
+    When I create a unique reposync benchmark channel
+    And I run the reposync benchmark for the mounted source repository
+    Then the reposync benchmark should finish successfully
--- a/testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
+++ b/testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2026 SUSE LLC
+# Copyright (c) 2026 Akash Kumar <meakash7902@gmail.com>
 # Licensed under the terms of the MIT license.
 #
 
--- a/testsuite/features/step_definitions/storage_benchmark_steps.rb
+++ b/testsuite/features/step_definitions/storage_benchmark_steps.rb
@@ -1,0 +1,223 @@
+# Copyright (c) 2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+### Step definitions for optional real-world storage benchmark workloads.
+
+require 'base64'
+require 'json'
+require 'shellwords'
+require 'time'
+
+require_relative '../support/storage_benchmark_parser'
+
+UYUNI_BENCH_DEFAULT_SOURCE_REPO = '/uyuni-bench-source/leap-15.6-backports-update'.freeze
+UYUNI_BENCH_DEFAULT_RESULTS_PARENT = '/var/spacewalk/uyuni-bench/results/reposync'.freeze
+
+# Return the Kubernetes server pod running Uyuni.
+def reposync_benchmark_server_pod
+  return @reposync_benchmark_server_pod if @reposync_benchmark_server_pod
+
+  pod = get_pod_name('server', 'server')
+  raise ScriptError, 'Unable to determine the Uyuni server pod name' if pod.nil? || pod.empty?
+
+  @reposync_benchmark_server_pod = pod
+end
+
+# Run a shell command on the RKE2 host.
+def reposync_benchmark_run_on_host(command, timeout: DEFAULT_TIMEOUT, verbose: true, check_errors: true)
+  get_target('server').run_local(command, timeout: timeout, verbose: verbose, check_errors: check_errors)
+end
+
+# Run a shell command inside the Uyuni server pod.
+def reposync_benchmark_run_in_server_pod(command, timeout: DEFAULT_TIMEOUT, verbose: true, check_errors: true)
+  pod = Shellwords.escape(reposync_benchmark_server_pod)
+  escaped_command = Shellwords.escape(command)
+  reposync_benchmark_run_on_host(
+    "kubectl -n uyuni exec #{pod} -- sh -lc #{escaped_command}",
+    timeout: timeout,
+    verbose: verbose,
+    check_errors: check_errors
+  )
+end
+
+# Path to the mounted source repository as seen by the Uyuni server container.
+def reposync_benchmark_source_repo
+  ENV.fetch('UYUNI_BENCH_SOURCE_REPO', UYUNI_BENCH_DEFAULT_SOURCE_REPO).delete_suffix('/')
+end
+
+# URL passed to spacewalk-repo-sync.
+def reposync_benchmark_source_url
+  "file://#{reposync_benchmark_source_repo}/"
+end
+
+# Storage backend label stored in the benchmark summary.
+def reposync_benchmark_storage_backend
+  ENV.fetch('UYUNI_BENCH_STORAGE_BACKEND', ENV.fetch('STORAGE_BACKEND', 'unknown'))
+end
+
+# Parse an integer environment variable and raise a clear error for invalid values.
+def reposync_benchmark_integer_env(name, default, minimum:)
+  value = ENV.fetch(name, default).to_s
+  valid_value = value.match?(/\A\d+\z/) && value.to_i >= minimum
+  raise ScriptError, "#{name} must be an integer >= #{minimum}, got #{value.inspect}" unless valid_value
+
+  value.to_i
+end
+
+# Directory for this benchmark run as seen by the Uyuni server container.
+def reposync_benchmark_results_dir
+  return @reposync_benchmark_results_dir if @reposync_benchmark_results_dir
+
+  @reposync_benchmark_results_dir =
+    if ENV['UYUNI_BENCH_RESULTS_DIR']
+      ENV.fetch('UYUNI_BENCH_RESULTS_DIR').delete_suffix('/')
+    else
+      timestamp = Time.now.utc.strftime('%Y%m%d%H%M%S')
+      File.join(UYUNI_BENCH_DEFAULT_RESULTS_PARENT, "#{reposync_benchmark_storage_backend}-#{timestamp}")
+    end
+end
+
+# Unique channel label for this benchmark run.
+def reposync_benchmark_channel_label
+  return @reposync_benchmark_channel_label if @reposync_benchmark_channel_label
+
+  @reposync_benchmark_channel_label =
+    ENV.fetch('UYUNI_BENCH_CHANNEL_LABEL') do
+      "uyuni-bench-reposync-#{Time.now.utc.strftime('%Y%m%d%H%M%S')}"
+    end
+end
+
+# Write JSON content into the server pod without depending on host path sharing.
+def reposync_benchmark_write_json_in_server_pod(path, payload)
+  encoded = Base64.strict_encode64(JSON.pretty_generate(payload))
+  reposync_benchmark_run_in_server_pod(
+    "mkdir -p #{Shellwords.escape(File.dirname(path))} && " \
+    "printf '%s' #{Shellwords.escape(encoded)} | base64 -d > #{Shellwords.escape(path)}"
+  )
+end
+
+Given('the reposync benchmark source repository is mounted in the server pod') do
+  repo = Shellwords.escape(reposync_benchmark_source_repo)
+  command = "test -r #{repo}/repodata/repomd.xml && find #{repo} -name '*.rpm' | wc -l"
+  output, = reposync_benchmark_run_in_server_pod(command, timeout: 300)
+  @reposync_benchmark_source_package_count = reposync_benchmark_last_integer(output)
+
+  raise ScriptError, "No RPM packages found in #{reposync_benchmark_source_repo}" if @reposync_benchmark_source_package_count.zero?
+
+  log "Reposync benchmark source: #{reposync_benchmark_source_url}"
+  log "Reposync benchmark source RPMs: #{@reposync_benchmark_source_package_count}"
+end
+
+When('I create a unique reposync benchmark channel') do
+  label = reposync_benchmark_channel_label
+  name = label
+  summary = 'Uyuni reposync storage benchmark channel'
+  arch = ENV.fetch('UYUNI_BENCH_CHANNEL_ARCH', 'channel-x86_64')
+  parent = ENV.fetch('UYUNI_BENCH_PARENT_CHANNEL', '')
+
+  assert_equal(1, $api_test.channel.software.create(label, name, summary, arch, parent))
+  log "Created reposync benchmark channel: #{label}"
+end
+
+When('I run the reposync benchmark for the mounted source repository') do
+  @reposync_benchmark_summary_path = File.join(reposync_benchmark_results_dir, 'summary.json')
+  @reposync_benchmark_stdout_path = File.join(reposync_benchmark_results_dir, 'spacewalk-repo-sync.stdout.log')
+  @reposync_benchmark_stderr_path = File.join(reposync_benchmark_results_dir, 'spacewalk-repo-sync.stderr.log')
+  @reposync_benchmark_time_path = File.join(reposync_benchmark_results_dir, 'time.verbose.log')
+
+  channel = Shellwords.escape(reposync_benchmark_channel_label)
+  source_url = Shellwords.escape(reposync_benchmark_source_url)
+  results_dir = Shellwords.escape(reposync_benchmark_results_dir)
+  stdout_path = Shellwords.escape(@reposync_benchmark_stdout_path)
+  stderr_path = Shellwords.escape(@reposync_benchmark_stderr_path)
+  time_path = Shellwords.escape(@reposync_benchmark_time_path)
+  timeout = reposync_benchmark_integer_env('UYUNI_BENCH_REPOSYNC_TIMEOUT', '14400', minimum: 1)
+
+  sync_command = "spacewalk-repo-sync -c #{channel} --url=#{source_url}"
+  command = "mkdir -p #{results_dir} && " \
+            'if [ -x /usr/bin/time ]; then ' \
+            "/usr/bin/time -v -o #{time_path} #{sync_command}; " \
+            "else #{sync_command}; fi " \
+            "> #{stdout_path} 2> #{stderr_path}"
+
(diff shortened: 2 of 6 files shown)


testsuite/test/storage_benchmark_parser_test.rb
```diff
@@ -0,0 +1,201 @@
...
+      3027
+    OUTPUT
+
+    assert_equal(3027, reposync_benchmark_last_integer(output))
+    assert_equal(0, reposync_benchmark_last_integer('no integer here'))
+  end
+
+  private
+
+  def full_reposync_log
```
cbosdo: You could add patches sync to your test: IIRC I sent you a log with some a few weeks ago.

cbosdo (commented):
I don't see the `testsuite/ext-tools` folder in this PR code. Is this due to an outdated PR description or a missing `git add`?

Since the commits history of this PR have no special value, you should squash them together and add a git commit description that could be useful for the future developers (may be you too) doing archeology. The PR description is usually a good thing to have as commit message.

I am not really fluent in ruby and the testsuite, I'll let @Pablogoliva or @uyuni-project/qe  check the code more thoroughly

testsuite/ext-tools/uyuni-bench/uyuni_bench/workloads/repodata.py
```diff
@@ -0,0 +1,449 @@
...
+    repo_dir: Path,
+    mode: str,
+    iteration_dir: Path,
+    extra_args: Optional[Sequence[str]] = None,
+) -> Dict[str, Any]:
+    if shutil.which("createrepo_c") is None:
+        raise RuntimeError("createrepo_c is not installed or not in PATH")
+
+    if mode == "full":
+        shutil.rmtree(repo_dir / "repodata", ignore_errors=True)
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. This comment refers to the earlier Python repodata benchmark helper, 
 which is no longer part of this PR.                                            
                                                                                
   The current implementation uses the Uyuni Cucumber testsuite directly and    
 runs `spacewalk-repo-sync` against a mounted local `file://` repository. Since 
 this file/path no longer exists in the PR, resolving as outdated.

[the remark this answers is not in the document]
AkashKumar7902: Thanks for catching this. This was part of the initial Python repodata helper, which has since been removed from the PR. The current benchmark no longer deletes or regenerates repodata directly; it runs the real `spacewalk-repo-sync` path against a mounted local `file://` repository, so this cleanup concern is no longer applicable.

testsuite/ext-tools/uyuni-bench/uyuni_bench/workloads/repodata.py
```diff
@@ -0,0 +1,449 @@
...
+        command,
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    duration = time.monotonic() - start
+
+    stdout_path.write_text(completed.stdout, encoding="utf-8", errors="replace")
+    stderr_path.write_text(completed.stderr, encoding="utf-8", errors="replace")
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. This applied to the earlier Python helper that captured `createrepo_c` output. That code is no longer in the PR. The current implementation runs `spacewalk-repo-sync` in the server pod and redirects stdout/stderr directly to log files under the benchmark result directory.

testsuite/ext-tools/uyuni-bench/uyuni_bench/workloads/repodata.py
```diff
@@ -0,0 +1,449 @@
...
+    with _urlopen(entry.url, timeout=300) as response, temporary.open("wb") as handle:
+        shutil.copyfileobj(response, handle, length=1024 * 1024)
+
+    actual_size = temporary.stat().st_size
+    if entry.size and actual_size != entry.size:
+        temporary.unlink(missing_ok=True)
+        raise RuntimeError(
+            f"downloaded size mismatch for {entry.href}: "
+            f"expected {entry.size}, got {actual_size}"
+        )
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. This was valid for the old dataset-preparation code, but that helper and manifest download path have been removed. The current benchmark uses a pre-mounted source repository as input and records package counts from the source repo and the Uyuni API instead.

testsuite/ext-tools/uyuni-bench/uyuni_bench/workloads/repodata.py
```diff
@@ -0,0 +1,449 @@
...
+
+    if verify_checksum and entry.checksum and entry.checksum_type:
+        actual = file_checksum(temporary, entry.checksum_type)
+        if actual != entry.checksum:
+            temporary.unlink(missing_ok=True)
+            raise RuntimeError(f"checksum mismatch for {entry.href}")
+
+    temporary.replace(destination)
+    return {"href": entry.href, "status": "downloaded", "bytes": entry.size}
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. This referred to the removed Python dataset preparation flow. The current PR no longer computes per-file manifest byte totals, so there is no `entry.size` vs actual-size accounting path left to fix here.

testsuite/ext-tools/uyuni-bench/uyuni_bench/workloads/repodata.py
```diff
@@ -0,0 +1,449 @@
...
+def prepare_dataset(
+    repo_url: str,
+    repo_dir: Path,
+    arches: Optional[Sequence[str]],
+    limit: Optional[int],
+    workers: int,
+    force: bool = False,
+    verify_checksum: bool = False,
+    dataset_name: Optional[str] = None,
+) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. This was for the old `ThreadPoolExecutor` dataset-preparation logic, which is no longer part of the PR. The benchmark now expects the source repository to be mounted before the run and does not use worker threads for package preparation.

testsuite/ext-tools/uyuni-bench/uyuni_bench/workloads/repodata.py
```diff
@@ -0,0 +1,449 @@
...
+    started_at = utc_now()
+    print(
+        f"Preparing {len(entries)} packages in {repo_dir} "
+        f"from {metadata['repo_url']} with {workers} workers"
+    )
+    counters: Dict[str, int] = {"downloaded": 0, "skipped": 0, "failed": 0}
+    bytes_seen = 0
+    failures: List[Dict[str, Any]] = []
+
+    with ThreadPoolExecutor(max_workers=workers) as executor:
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. This also applied to the removed worker-based preparation code. The current reposync benchmark does not expose a worker-count option, so this failure mode is no longer present.

testsuite/features/step_definitions/storage_benchmark_steps.rb
```diff
@@ -0,0 +1,169 @@
...
+
+# Run a shell command inside the Uyuni server pod.
+def uyuni_bench_run_in_server_pod(command, timeout: DEFAULT_TIMEOUT, verbose: true, check_errors: true)
+  pod = Shellwords.escape(uyuni_bench_server_pod)
+  escaped_command = Shellwords.escape(command)
+  uyuni_bench_run_on_host(
+    "kubectl -n uyuni exec #{pod} -- bash -lc #{escaped_command}",
+    timeout: timeout,
+    verbose: verbose,
+  )
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. I changed the pod command wrapper from `bash -lc` to `sh -lc`. The benchmark commands are POSIX-shell compatible, so this keeps the opt-in benchmark a bit less dependent on the container having Bash available.

testsuite/features/step_definitions/storage_benchmark_steps.rb
```diff
@@ -0,0 +1,169 @@
...
+  uyuni_bench_upload_to_host
+  uyuni_bench_copy_to_server_pod
+  uyuni_bench_run_in_server_pod("cd #{UYUNI_BENCH_POD_DIR} && ./uyuni-bench --version")
+end
+
+When('I prepare the repodata generation benchmark dataset') do
+  dataset = Shellwords.escape(uyuni_bench_dataset_path)
+  repo_dir = Shellwords.escape(uyuni_bench_repo_dir)
+  workers = ENV.fetch('UYUNI_BENCH_PREPARE_WORKERS', '4').to_i
+  timeout = ENV.fetch('UYUNI_BENCH_PREPARE_TIMEOUT', '7200').to_i
```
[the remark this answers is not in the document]
AkashKumar7902: Thanks. The old `UYUNI_BENCH_PREPARE_WORKERS` path has been removed with the Python preparation helper. For the remaining integer inputs, I added explicit validation so invalid values such as a non-numeric timeout fail early with a clear `ScriptError` instead of being silently converted with `to_i`.

testsuite/Rakefile
```diff
@@ -7,12 +7,21 @@ require 'yaml'
...
 require 'cucumber/rake/task'
 require 'rake/task'
+require 'rake/testtask'
 require 'parallel'
 
 junit_results = '-f junit -o results_junit'
 build_number = ENV.fetch('BUILD_NUMBER', nil)
 result_folder = build_number.nil? ? 'results' : "results/#{build_number}"
 
+namespace :test do
```
Pablogoliva: I don't understand what's the matter of it

testsuite/features/support/storage_benchmark_parser.rb
```diff
@@ -0,0 +1,298 @@
+# Copyright (c) 2026 SUSE LLC.
+# Licensed under the terms of the MIT license.
+
+require 'time'
+
+# Parse useful fields from /usr/bin/time -v output.
+def reposync_benchmark_parse_time_metrics(output)
```
Pablogoliva, line 7: These units tests are difficult to maintian and it's not our common approach. I would say leave just the e2e tests.

testsuite/test/storage_benchmark_parser_test.rb
```diff
@@ -0,0 +1,247 @@
...
+      {
+        'command_being_timed' => '"spacewalk-repo-sync -c channel"',
+        'user_time_seconds' => '12.34',
+        'maximum_resident_set_size_kbytes' => '123456'
+      },
+      metrics
+    )
+  end
+
+  def test_last_integer_ignores_shell_noise
```
Pablogoliva: Also check linux logs it's a nightmare as a tests. So please use another approach.

Pablogoliva (changes requested):

testsuite/Rakefile
AkashKumar7902: Removed the `test:unit` rake task (and the `rake/testtask` require) in 2a772ddb — it was only there to run the parser unit tests, which are gone now.

testsuite/features/support/storage_benchmark_parser.rb
AkashKumar7902: Removed the unit tests and kept only the e2e cucumber scenario (2a772ddb).

One question on the parser itself (`storage_benchmark_parser.rb`): the e2e still uses it to turn the `spacewalk-repo-sync` log into the phase timings/throughput in `summary.json`. Do you want me to keep that as the e2e's result parsing, or would you rather the e2e just record the wall-clock duration + raw logs and drop the log-derived metrics entirely? Happy to simplify it further if that's the other approach you had in mind.

testsuite/test/storage_benchmark_parser_test.rb
AkashKumar7902: Agreed — removed the log-parsing unit tests entirely (2a772ddb); the benchmark now relies only on the e2e run.

testsuite/features/step_definitions/storage_benchmark_steps.rb
```diff
@@ -0,0 +1,223 @@
...
+    verbose: false,
+    check_errors: false
+  )
+  sync_stdout, = reposync_benchmark_run_in_server_pod(
+    "test -f #{stdout_path} && cat #{stdout_path} || true",
+    verbose: false,
+    check_errors: false
+  )
+  sync_stderr, = reposync_benchmark_run_in_server_pod(
+    "test -f #{stderr_path} && cat #{stderr_path} || true",
```
Pablogoliva: The system commands should be executed through the run method. Not like this.

testsuite/Rakefile
AkashKumar7902: Confirmed after rebasing onto current `master`: the `test:unit` task and `rake/testtask` require remain removed. This benchmark no longer changes the Rakefile.

testsuite/features/support/storage_benchmark_parser.rb
AkashKumar7902: Simplified in `9f684a3eab` as suggested. The parser and all log-derived phase/throughput metrics are removed. The e2e scenario now records monotonic wall time, exit status, source and API-verified package counts, plus paths to raw stdout, stderr, and `/usr/bin/time` output.

testsuite/test/storage_benchmark_parser_test.rb
AkashKumar7902: Addressed in `9f684a3eab`. The human-log parser and its unit-test fixtures are both gone; the scenario validates stable end-to-end outcomes and preserves raw artifacts instead of parsing application/Linux log wording.

testsuite/features/step_definitions/storage_benchmark_steps.rb
AkashKumar7902: Addressed in `9f684a3eab`. The host-command helper now calls `get_target('server').run(..., runs_in_container: false)`, so `kubectl` still executes on the RKE2 host while using the standard `RemoteNode` command path.

AkashKumar7902:
@cbosdo @Pablogoliva The branch is now rebased onto current `master`, and the `cucumber.yml` conflict was resolved by retaining the current upstream profile list and adding only `storage_benchmark`.

The PR remains a single signed-off commit (`9f684a3eab`) with an updated durable message. The description now matches the four-file diff and no longer claims a Rake task, parser unit tests, or parser-derived metrics.

Following the review, the human-log parser and derived phase metrics are gone. The e2e scenario keeps stable wall time, exit status, source/API package counts, and raw artifacts. Host-side `kubectl` execution now uses `RemoteNode#run(..., runs_in_container: false)`. Ruby syntax, RuboCop, YAML/profile linkage, shell syntax, routing assertions, stale-reference searches, and diff checks pass; the live workload still requires a mounted repository and RKE2/Uyuni deployment.

AkashKumar7902:
Updated in `2bb9807598` for the Sumaform external-cluster path. Cluster readiness, Running/Ready server-pod discovery, and every `kubectl exec` now run from the testsuite controller through the registered `localhost` target and its kubeconfig; `SERVER` remains the Uyuni API/ingress endpoint. The pod command also selects the `uyuni` container explicitly.

Ruby syntax, targeted RuboCop 1.82.1, YAML/run-set linkage, focused controller-routing and readiness assertions, and `git diff --check` pass. I have not claimed a new full workload run: the current lab has only about 18 GiB free and the Uyuni server stack is not healthy enough to benchmark safely. The next live run will be through Sumaform after storage headroom and pod health are restored, with the source mount supplied in the Sumaform Helm values.

testsuite/features/step_definitions/storage_benchmark_steps.rb
```diff
@@ -0,0 +1,258 @@
...
+        '--container',
+        'uyuni',
+        reposync_benchmark_server_pod,
+        '--',
+        'sh',
+        '-lc',
+        command
+      ]
+    )
+  get_target('localhost').run_local(
```
Pablogoliva, line 76:
```suggestion
  get_target('controller').run_local(
```

AkashKumar7902: The current testsuite target registry does not define a `controller` key. It maps `localhost` to `HOSTNAME`, and `RemoteNode` raises `NotImplementedError` for an unregistered target. In a Sumaform run, `HOSTNAME` identifies the testsuite controller, where the kubeconfig and `kubectl` are installed, so I kept `get_target("localhost").run_local` in f5eb57ad64. Changing this call to `controller` alone would fail before `kubectl` executes; introducing that name would require a separate framework-wide target and environment export.

cbosdo (approved):

testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2026 SUSE LLC
```
cbosdo: The copyright is not correct: change the SUSE LLC with your full name and email.

testsuite/features/step_definitions/storage_benchmark_steps.rb
```diff
@@ -0,0 +1,257 @@
+# Copyright (c) 2026 SUSE LLC.
```
cbosdo: Copyright issue here too

cbosdo (changes requested):
You need to fix the copyright headers before being able to merge this PR

testsuite/features/kubernetes/srv_rke2_storage_benchmark.feature
AkashKumar7902: done

testsuite/features/step_definitions/storage_benchmark_steps.rb
AkashKumar7902: done

cbosdo (dismissed):
The PR looks good but rubocop config needs to be changed to not only match SUSE LLC copyrights

testsuite/features/support/network_utils.rb
```diff
@@ -21,7 +21,7 @@ def ssh_command(command, host, port: 22, timeout: DEFAULT_TIMEOUT, buffer_size:
   begin
     Timeout.timeout(timeout) do # Enforce timeout on the entire SSH operation
       Net::SSH.start(host, nil, port: port, verify_host_key: :never, timeout: timeout, keepalive: true, max_pkt_size: buffer_size, config: true) do |ssh|
-        stdout, stderr, exit_code = ssh_exec!(ssh, command, timeout: DEFAULT_TIMEOUT)
+        stdout, stderr, exit_code = ssh_exec!(ssh, command, timeout: timeout)
```
Pablogoliva: Leave the previous timeout pls. That is intended.

AkashKumar7902: ok, removed, 
the below is going to be needed beforing starting reposync benchmark testsuite run
```
  # Shared testsuite command limit
  export DEFAULT_TIMEOUT=14400

```

Pablogoliva: Perfect

Pablogoliva (approved):

PR #12093, opened 2026-06-10
taskotop needs to properly handle timezones (bsc#1267871)

cbosdo:
## What does this PR change?

The data taskotop contain the timezone. Converting to local timezone will show correct times while the current way drops it.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: taskotop is an ncurses tool

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30904
Port(s): 5.1: https://github.com/SUSE/spacewalk/pull/30927 5.2: https://github.com/SUSE/spacewalk/pull/31391

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/utils/taskotop
+++ b/utils/taskotop
@@ -27,7 +27,6 @@ from spacewalk.server import rhnSQL
 
 from io import BytesIO
 
-
 DEFAULT_LOGFILE = "./taskotop.log"
 
 parser = argparse.ArgumentParser(
@@ -207,7 +206,8 @@ class CursesDisplayBuilder:
             value = value[0:-1]
         log_debug(
             # pylint: disable-next=consider-using-f-string
-            "y=%d x=%d value '%s'  value length %d" % (ypos, addxpos, value, len(value))
+            "y=%d x=%d value '%s'  value length %d"
+            % (ypos, addxpos, value, len(value))
         )
         if len(value) > 0:
             screen.addstr(ypos, addxpos, value)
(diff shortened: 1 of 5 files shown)


utils/taskotop
```diff
@@ -663,7 +655,7 @@ def format_elapsed_time(rowdata, data_key, width):
     if rowdata["end_time"]:
         end = rowdata["end_time"]
 
-    td = end.replace(tzinfo=None) - rowdata["start_time"].replace(tzinfo=None)
+    td = end.astimezone() - rowdata["start_time"].astimezone()
```
m-czernek, line 658: Just wondering here, is there any possibility of the data in the DB using a different TZ than the currently configured locale one? For example the user explicitly changing the postgres TZ without changing the system's TZ, or something similar. 

Although I suppose that the data entered into the DB are generated by the Java Taskomatic side and would also take the system TZ configuration, so this shouldn't be a problem.

m-czernek (dismissed):
Leaving a non-blocking comment to consider before merging, though I suspect that both the Java and Python parts simply use the system's TZ configuration, so +1.

utils/taskotop
cbosdo: Yes it is possible: currently the DB is in UTC and the main container (where taskotop runs) could be set to any time zone. The timestamps in the DB are automatically saved with the timezone info: those just need to be properly handled on the taskotop side.

aaannz (approved):

PR #12101, opened 2026-06-11
Rework Address entity

rjpmestre:
## What does this PR change?

This PR reworks the Address entity.

During https://bugzilla.suse.com/show_bug.cgi?id=1259951 I couldn't help to notice how outdated some Address relationships and data handling were.
To keep the original issue focused and avoid increasing its scope, we decided to implement the immediate fixes first (PR [#11681](https://github.com/uyuni-project/uyuni/pull/11681) and [#11958](https://github.com/uyuni-project/uyuni/pull/11958)) and follow up with this dedicated rework.

It main focus is to: 
-  Remove legacy address type: MARKETING/BILLING/SHIPPING/SERVICE  as only MARKETING was used;
-  Remove completely unused fields from `web_user_site_info`:  `email_uc`, `alt_first_names`, `alt_last_name`, `address3`, `address4`, `url`, `oracle_site_id`, and 8 `_ol` locale fields
- Remove the implicit creation of empty Address records when a user is created. User creation and address management are now treated as separate concerns, allowing a user to exist without an address until one is explicitly provided. **NOTE: ** Code seems to suggest that user should be able to have multiple addresses but at the same time it seems to support just one. We can postpone this decision.

This is not intended to be a comprehensive redesign of the address model. Additional improvements remain possible, but they would require broader changes across related entities / dto. Those will be addressed separately in future work.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30716

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


java/core/src/main/resources/com/redhat/rhn/frontend/strings/java/StringResource_en_US.xml
```diff
@@ -266,24 +266,12 @@
           <context context-type="sourcefile">Navigation Menu</context>
         </context-group>
       </trans-unit>
-      <trans-unit id="address type B" xml:space="preserve">
```
CDellaGiusta, line 269: I have been told that StringResource_en_US.xml is the ONLY file you should edit. Am I wrong?

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
CDellaGiusta: you are missing the change in this method, aren't you?

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
```diff
@@ -1059,62 +1076,23 @@ public void setHiddenPanes(Set<Pane> p) {
...
-            for (Address addressIn : addrA) {
-                if (addressIn.getType().equals(Address.TYPE_MARKETING)) {
-                    addr = addressIn;
-                }
-                if (addressIn.getType().equals("B")) {
-                    baddr = addressIn;
-                }
-            }
+            AddressImpl current = addresses.iterator().next();
+            current.setUser(null);
```
CDellaGiusta, line 888: what is the reason to nullify the user of the first of the addresses?

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -39,259 +43,226 @@ public class AddressImpl extends BaseDomainHelper implements Address {
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "WEB_USER_SITE_INFO_SEQ")
     @SequenceGenerator(name = "WEB_USER_SITE_INFO_SEQ", sequenceName = "WEB_USER_SITE_INFO_ID_SEQ", allocationSize = 1)
     private Long id;
-    @Column
-    private String address1;
-    @Column
+
+    @Column(length = 128)
```
CDellaGiusta: Here and in other parts: is it worth specifying the field length ? What if the database field is extended/changed? There are still a bunch of such specifications around, but I feel it's legacy code

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -39,259 +43,226 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
+    private String email;
+
+    @Column(name = "address1", length = 128, nullable = false)
```
CDellaGiusta: here and in other parts: name attribute should be specified only if the column name is different from the member name, otherwise it's redundant. See the above "email" field, it has no name specification. I would remove the name annotation spefication if redundant

CDellaGiusta (changes requested):
In general, you should prefer smaller commits to help reviewers follow a line of though, and then, if the case, squash them into one/some related and more consistent commits.
This commit was quite difficult to follow, since there are mixed logical operations you did, whose side-effects are quite difficult to track:
- removed fields (in code and database)
- cosmetics (e.g removing this.something)
- removing null checks
- removal of methods (getPrivType, getIsPoBox)
- removal of address types (TARGET_ADDRESS_BILLING and the like)
- changes in code, in pages, in db

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
rjpmestre: You're right, I mixed this with the @Size annotation. But we're not using hibernate-validator anyway. I'm dropping these.

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
rjpmestre: Yes, this has to do with a note I left of the PR.

This legacy functionality was clearly designed to support multiple addresses, including the different address types. For some reason, we seem to have been reducing these capabilities. We stopped using the address types, reducing all to `Marketing`,  and now, our address manager actually only allows users to create a single address. After that, they can only edit it (and not even delete it!).

This is where I drew a line. I was ok removing address types, but I still think allowing users to have multiple addresses and properly manage them is a valuable thing we might want to get back.

My plan was to apply these changes for now, and then discuss where we want to take the address manager going forward.

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -39,286 +39,225 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
-
-    /**
-     * {@inheritDoc}
-     */
-    public User getUser() {
-        return user;
+        if (!(o instanceof AddressImpl other)) {
+            return false;
+        }
+        return id != null && id.equals(other.id);
```
mackdk: Since the `id` is autogenerated, with this implementation you won't be able to compare new entities, until they are persisted into the database. I would just compare all the fields excluding the `id` in both the `equals` and `hashcode`

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -39,286 +39,225 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
 
-
     /**
      * {@inheritDoc}
      */
-    public void setUser(User userIn) {
-        user = (UserImpl) userIn;
+    @Override
+    public int hashCode() {
+        return getClass().hashCode();
```
mackdk: This implementation looks wrong. It should be consistent with the `equals`

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -39,286 +39,225 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
+                ", email='" + email + '\'' +
+                ", address1='" + address1 + '\'' +
+                ", address2='" + address2 + '\'' +
+                ", city='" + city + '\'' +
+                ", state='" + state + '\'' +
+                ", zip='" + zip + '\'' +
+                ", country='" + country + '\'' +
+                ", phone='" + phone + '\'' +
+                ", fax='" + fax + '\'' +
+                '}';
```
mackdk:
```java
        return new StringJoiner(", ", AddressImpl.class.getSimpleName() + "[", "]")
                .add("id=" + id)
                .add("email='" + email + "'")
                .add("address1='" + address1 + "'")
                .add("address2='" + address2 + "'")
                .add("city='" + city + "'")
                .add("state='" + state + "'")
                .add("zip='" + zip + "'")
                .toString();
```

schema/spacewalk/susemanager-schema.spec
```diff
@@ -20,7 +20,7 @@
 %{!?productprettyname: %global productprettyname Uyuni}
 
 Name:           susemanager-schema
-Version:        5.2.11
+Version:        5.2.12
```
mackdk: This is not needed, it will be changed automatically when we release the version

java/core/src/main/java/com/redhat/rhn/domain/user/AddressImpl.java
```diff
@@ -39,286 +39,225 @@ public class AddressImpl extends BaseDomainHelper implements Address {
...
+    @Column(nullable = false)
     private String country;
+
     @Column
     private String phone;
+
     @Column
     private String fax;
-    @Column(name = "is_po_box")
-    private String isPoBox;
```
mackdk, line 59: Personally, I would keep the `isPoBox` field, but we should make it a proper `boolean`.

schema/spacewalk/susemanager-schema.spec
rjpmestre: ahh I needed that for running tests locally. wasn't supposed to go in. mb

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
```diff
@@ -818,7 +821,11 @@ public String getPhone() {
      */
     @Override
     public void setPhone(String phoneIn) {
-        getAddress().setPhone(phoneIn);
+        Address address = getAddress();
+        if (address == null) {
+            throw new IllegalStateException(USER_HAS_NO_ADDRESS);
```
lucidd: instead of failing it would be better if we did two things:

- have getters reflect optionality
- remove any getters and setters that are forwarded

this provides more clarity of what can be null and gives a way to directly handle it in a standard way

java/core/src/main/java/com/redhat/rhn/domain/user/Address.java
```diff
@@ -184,8 +171,9 @@ public interface Address extends Serializable {
 
     /**
      * Sets the user associated with this address
+     * @param <T> Type of User
      * @param user New value for user
      */
-    void setUser(User user);
+    <T extends User> void setUser(T user);
```
lucidd: why is this needed?

java/core/src/main/java/com/redhat/rhn/frontend/taglibs/AddressTag.java
lucidd, line 146: these getters previously returned " " when they were null. Now that they are returning null it needs to be handled

java/core/src/main/java/com/redhat/rhn/domain/user/Address.java
rjpmestre: My bad, I forgot to remove this while I was exploring something

java/core/src/main/java/com/redhat/rhn/domain/user/legacy/UserImpl.java
rjpmestre: You're right! Your comment made me realize I had forgot we could also get rid of the `EnterpriseUser`

java/core/src/main/java/com/redhat/rhn/frontend/taglibs/AddressTag.java
rjpmestre: @lucidd and I ended up having a chat about this and this is just for record.
So, we noticed that there is a discrepancy. What is considered the minimum required data for an address in the db, the webui and the backend all have different expectations. 
It's not that this is something that's prone to failure, but it is inefficient.
At this point we discussed a possible rework, but in the end of the day even if we decide to go down that route, we (as a team) first review and align on the goals of this **address manager**. And since we're already planning to redesign some of this screens in the future anyway, maybe would be the ideal time to address this.

CDellaGiusta (dismissed):
Code looks good, there's only one catch: database schema has changed meanwhile.
My suggestion is to rebase on master, then change the upgrade schema directories to the appropriate value.

PR #12105, opened 2026-06-12
Fix test leftovers causing side effects

CDellaGiusta:
## What does this PR change?

When run in random order, tests sometimes fail due to leftovers left by other previously executed tests causing unexpected side effects. This PR aims at fixing them.
The order in which `ant` executes tests didn't show those problems at the moment, but when running tests with `maven`, this order changes and side effects are popping up.

To run a test in random order:
`mvn test -DskipTests=false -Dsurefire.runOrder=random | tee testLog.log`

at the beginning of the test, the random seed used is shown, so that the test can be repeated in the same order:

e.g.
`mvn test -DskipTests=false -Dsurefire.runOrder=random   -Dsurefire.runOrder.random.seed=68149531682814 `

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/30954
Port(s): none (5.2)
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/core/src/test/java/org/cobbler/MockConnection.java
```diff
@@ -277,7 +277,7 @@ public Object invokeMethod(String name, Object... args) {
                     // Log a warning because in production this should not happen, however it is possible that the
                     // Mock doesn't implement everything as desired.
                     this.log.warn(
-                            String.format("Requested attribute \"%s\" for uuid \"%s\" not found!", args[0], args[1])
+                            String.format("Requested attribute \"%s\" for uuid \"%s\" not found!", args[1], args[0])
```
mackdk: No need to use `String.format()` here, we can just use log4j formatting capability:
```java
his.log.warn("Requested attribute \"{}\" for uuid \"{}\" not found!", args[1], args[0]);
```

mackdk (dismissed):
LGTM. As a minor suggestion maybe we can move the new methods `deleteAll()` and `testDeleteLastMgrSyncRefresh()` to `TestUtils` or to a new test-only factory so we don't have to put methods meant only for test inside our production code.

java/core/src/test/java/org/cobbler/MockConnection.java
CDellaGiusta: suggestion taken. fixed

CDellaGiusta:
> LGTM. As a minor suggestion maybe we can move the new methods `deleteAll()` and `testDeleteLastMgrSyncRefresh()` to `TestUtils` or to a new test-only factory so we don't have to put methods meant only for test inside our production code.

I didn't consider that kind of potential "pollution". I agree. I moved the two methods in TestUtils

CDellaGiusta:
I discovered that running tests in random order was sending a thousand of mails to galaxy-noise.
Hence I added this commit: [fix avoid sending emails while testing](https://github.com/uyuni-project/uyuni/pull/12105/changes/e8b302f667ba90b4a21d7587dab1851e51839c8e)

java/core/src/test/java/com/redhat/rhn/testing/TestNoOpMail.java
```diff
@@ -0,0 +1,70 @@
...
+import com.redhat.rhn.common.messaging.Mail;
+
+// This mailer is used in test configuration ("buildconf/rhn.conf.postgresql-example", copied to "buildconf/rhn.conf")
+// in the entry:
+//          web.mailer_class = com.redhat.rhn.testing.TestNoOpMail
+// in order to avoid flooding "galaxy-noise@suse.de" mailbox with error messages when unit tests are run
+// The method MailHelper.createMailFromConfig() looks for the configuration entry "web.mailer_class"
+// to create a child object of class Mail.
+// In production, the entry "web.mailer_class" is not found in rhn.conf and so the mailer is the default one (SmtpMail)
+// When testing, the entry "web.mailer_class" is found and points to create an instance of this class
```
mackdk: This can just be a javadoc comment, I guess

java/core/src/test/java/com/redhat/rhn/testing/TestNoOpMail.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class TestNoOpMail implements Mail  {
```
mackdk: I feel the `Test` prefix is redundant here since this implementation is already not accessible from the main code. I would just call it `NoOpMail`.

I kind of don't like that this is both a mail and a mailer, but I guess that's for another refactoring 😅

mackdk (dismissed):
LGTM, just minor low priority comments as usual 😄

java/core/src/test/java/com/redhat/rhn/testing/TestNoOpMail.java
CDellaGiusta: it may well be, changed

java/core/src/test/java/com/redhat/rhn/testing/TestNoOpMail.java
CDellaGiusta: changed class name and reference in buildconf/rhn.conf.postgresql-example

mackdk (dismissed):

java/core/src/main/java/com/redhat/rhn/domain/image/ImageInfoFactory.java
rjpmestre, line 483: I know your change has nothing to do this. I did noticed a couple of things we should check in the future:
- the method name is by itself is underselling what the query really does. In my opinion this should' be named something like `findByNameAndVersionAndStore` and the comment should be updated explaining it can be null.
- I think the logic of the method could be simplified to something like 
```
public static Optional<ImageInfo> lookupByName(String name, String version, ImageStore imageStore) {
        return getSession()
                .createQuery("""
                                FROM ImageInfo WHERE name = :name
                                AND (version = :version OR (version IS NULL AND :version IS NULL))
                                AND store = :imageStore ORDER BY revisionNumber, id""", ImageInfo.class)
                .setParameter("name", name)
                .setParameter("version", version)
                .setParameter("imageStore", imageStore)
                .setMaxResults(1)
                .uniqueResultOptional();
    }
```

java/core/src/test/java/com/redhat/rhn/domain/channel/ChannelFamilyTest.java
```diff
@@ -76,9 +76,13 @@ public static ChannelFamily ensureChannelFamilyExists(User user, String label, S
             cf = ChannelFamilyFactoryTest.createTestChannelFamily(user, true, TestUtils.randomString());
             cf.setName(name);
             cf.setLabel(label);
-            ChannelFamilyFactory.save(cf);
         }
-        return cf;
+        else {
+            cf.setName(name);
```
rjpmestre: `cf.setName(name);` its applied on both cases, can be moved outside the if statement.

java/core/src/test/java/com/redhat/rhn/domain/image/ImageInfoFactoryTest.java
```diff
@@ -354,8 +354,7 @@ public final void testScheduleBuild() throws Exception {
         prd.setBaseproduct(true);
         prd = TestUtils.saveAndReload(prd);
         info.setInstalledProducts(Collections.singleton(prd));
-        ImageInfoFactory.save(info);
-        info = TestUtils.saveAndFlush(info);
+        info = ImageInfoFactory.save(info);
```
rjpmestre, line 357: You'd only need to assign the `save` method to a variable IF you were persisting a new entity. 
As this is already managed (from line 324), when you save, the managed entity will be updated.

You can confirm this behaviour. Just assign the save to another variable
```
ImageInfo infoCopy = ImageInfoFactory.save(info);
```
you'll notice both `info` and `info` will refer to the same object instance

```suggestion
        ImageInfoFactory.save(info);
```

java/core/src/test/java/com/redhat/rhn/taskomatic/task/sshservice/SSHServiceWorkerTest.java
```diff
@@ -93,6 +94,12 @@ public void setUp() throws Exception {
         sampleSystemInfo = Json.GSON.fromJson(jsonResult, SystemInfo.class);
     }
 
+    @AfterEach
+    public void tearDownSSHServiceWorkerTest() {
+        // The worker commits, so mark the session for cleanup on every test
+        commitHappened();
```
rjpmestre, line 100: I get the point of `commitHappened`. I think i have also used it in the past. There must be a way for us to automate this by counting commits or something on hibernate session. But nothing to do for now. Just throwing the idea out

rjpmestre (dismissed):
Looks good! I just left a few nitpick comments. Nothing that would prevent from merging it.

java/core/src/main/java/com/redhat/rhn/domain/image/ImageInfoFactory.java
CDellaGiusta: fully agree! Here I added an `id` in the order, since this query in one test in particular was giving 2 values, and the choice of the one to be returned should be deterministic. Since in this PR  I'm focusing on test side effects, my guess your valid proposal should be for another PR.

java/core/src/test/java/com/redhat/rhn/domain/channel/ChannelFamilyTest.java
CDellaGiusta: Yes. Changed, removed `else` part and changed the comment

java/core/src/test/java/com/redhat/rhn/taskomatic/task/sshservice/SSHServiceWorkerTest.java
CDellaGiusta: Fully agree, we're all on the same page. Probably one of the next PRs

java/core/src/test/java/com/redhat/rhn/domain/image/ImageInfoFactoryTest.java
mackdk: I beg to differ on this :slightly_smiling_face: 

>You'd only need to assign the save method to a variable IF you were persisting a new entity.

Actually, no, if you are saving a new entity, `save()` calls `persist()` which modifies the object in-place (in fact by contract that method returns `void`). If the object already exists like in this case (aka it has an id), `save` would either call `merge()` or return the same instance depending on the state of the entity. In this specific case, since the session is never flushed or cleared, it still contains the object we pass and `save()` indeed returns the same instance immediately.

However, if this test or the underlying code changes in the future (e.g. a `flush()`, `clear()` is introduced or the transaction boundaries change), the entity will become detached. Our `save()` implementation will call `merge()` to attach the instance to the session. `merge()` does not update the object in-place but returns a completely new one. If we don't reassign the variable, we will lose the attached instance leading to possible bugs.

Long story short, in my opinion it is simply a safer practice to just always reassign the instance after calling `save()`, without having to worry about what Hibernate will be doing behind the scene.

rjpmestre: oh, right, I wasn't considering a scenario where you'd detach the entity. fair enough

CDellaGiusta: besides this, the test fails if you do not reassign the variable ...

lucidd (approved):

PR #12108, opened 2026-06-12
Add extra volumes to server Helm chart

AkashKumar7902:
## What does this PR change?

Adds support for configuring additional Kubernetes volumes and volume mounts
in the server Helm chart through:

- `server.extraVolumes`
- `server.extraVolumeMounts`

This allows users to mount optional content volumes, custom certificates, or
operator-managed volumes without modifying the chart templates.

## Codespace

## GUI diff

No difference.

Before: N/A

After: N/A

- [x] **DONE**

## Documentation

Documentation updated in:

- `containers/server-helm/README.md`

- [x] **DONE**

## Test coverage

- Unit tests were added

Note: `helm unittest` was not run locally because the Helm unittest plugin is not installed in this environment.

- [x] **DONE**

## Links

Issue(s): none
Port(s): none

- [x] **DONE**

## Changelogs

Changelog added:

- `containers/server-helm/server-helm.changes.akash.extra-volumes`

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/server-helm/README.md
+++ b/containers/server-helm/README.md
@@ -71,6 +71,24 @@ This would be overridden by each claim's `volumes.<claim>.storageClass` value.
 The storage class supports a special `"-"` value to force the storage class to the empty string.
 This is different from the `""` value representing the default storage class by not setting any storage class at all.
 
+Additional volumes can be attached to the server pod with `server.extraVolumes` and mounted into the server container with `server.extraVolumeMounts`.
+This is useful for optional data sources such as benchmark repositories or externally managed content volumes without changing the chart templates.
+
+For example:
+
+```yaml
+server:
+  extraVolumes:
+    - name: uyuni-bench-source
+      persistentVolumeClaim:
+        claimName: uyuni-bench-source
+
+  extraVolumeMounts:
+    - name: uyuni-bench-source
+      mountPath: /uyuni-bench-source
+      readOnly: true
+```
+
 ### Node Tuning
 
 For each of the components it is possible to tune the node where the pod will be scheduled.
--- a/containers/server-helm/README.md
+++ b/containers/server-helm/README.md
@@ -72,20 +72,20 @@ The storage class supports a special `"-"` value to force the storage class to t
 This is different from the `""` value representing the default storage class by not setting any storage class at all.
 
 Additional volumes can be attached to the server pod with `server.extraVolumes` and mounted into the server container with `server.extraVolumeMounts`.
-This is useful for optional data sources such as benchmark repositories or externally managed content volumes without changing the chart templates.
+This is useful for optional content volumes like `inter-server-sync` data without changing the chart templates.
 
 For example:
 
 ```yaml
 server:
   extraVolumes:
-    - name: uyuni-bench-source
+    - name: inter-server-sync-data
       persistentVolumeClaim:
-        claimName: uyuni-bench-source
+        claimName: inter-server-sync-data
 
   extraVolumeMounts:
-    - name: uyuni-bench-source
-      mountPath: /uyuni-bench-source
+    - name: inter-server-sync-data
+      mountPath: /inter-server-sync-data
       readOnly: true
 ```
 
--- a/containers/server-helm/templates/server.yaml
+++ b/containers/server-helm/templates/server.yaml
@@ -269,8 +269,8 @@ spec:
         - name: mirror
           mountPath: /mirror
 {{- end }}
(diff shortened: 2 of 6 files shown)


mbussolotto:
@AkashKumar7902 changelog is missing https://github.com/uyuni-project/uyuni/wiki/Contributing#changelog-rules-for-packages-in-git . But the rest LGTM cc @cbosdo 

morever, please use the default template for the PR

containers/server-helm/templates/server.yaml
```diff
@@ -269,6 +269,9 @@ spec:
         - name: mirror
           mountPath: /mirror
 {{- end }}
+{{- with .Values.server.extraVolumeMounts }}
```
cbosdo: Don't we need to wrap this in a test if the value is set?

containers/server-helm/templates/server.yaml
```diff
@@ -540,3 +543,6 @@ spec:
           path: {{ .Values.server.mirror.hostPath }}
   {{- end }}
 {{- end }}
+{{- with .Values.server.extraVolumes }}
```
cbosdo: Same here, we may need to test if the value is set before using it

containers/server-helm/README.md
```diff
@@ -71,6 +71,24 @@ This would be overridden by each claim's `volumes.<claim>.storageClass` value.
 The storage class supports a special `"-"` value to force the storage class to the empty string.
 This is different from the `""` value representing the default storage class by not setting any storage class at all.
 
+Additional volumes can be attached to the server pod with `server.extraVolumes` and mounted into the server container with `server.extraVolumeMounts`.
+This is useful for optional data sources such as benchmark repositories or externally managed content volumes without changing the chart templates.
```
cbosdo:
```suggestion
This is useful for optional content volumes like `inter-server-sync` data without changing the chart templates.
```
I know the original use case is benchmark data, but users won't have that case. A mounted volume to export the data in order to reload them on another server will make more sense to users.

containers/server-helm/README.md
```diff
@@ -71,6 +71,24 @@ This would be overridden by each claim's `volumes.<claim>.storageClass` value.
...
+server:
+  extraVolumes:
+    - name: uyuni-bench-source
+      persistentVolumeClaim:
+        claimName: uyuni-bench-source
+
+  extraVolumeMounts:
+    - name: uyuni-bench-source
+      mountPath: /uyuni-bench-source
+      readOnly: true
```
cbosdo, line 89:
```suggestion
server:
  extraVolumes:
    - name: inter-server-sync-data
      persistentVolumeClaim:
        claimName: inter-server-sync-data

  extraVolumeMounts:
    - name: inter-server-sync-data
      mountPath: /inter-server-sync-data
      readOnly: true
```

This example could almost be used as is by some users.

containers/server-helm/values.yaml
```diff
@@ -43,6 +43,16 @@ server:
     # Using this is strongly discouraged as it is tied to where the server runs.
     hostPath: ""
 
+  # extraVolumes defines additional Kubernetes volumes for the server pod.
+  # This can be used to mount benchmark datasets, custom certificates or
+  # operator-managed volumes without modifying the chart templates.
```
cbosdo:
```suggestion
  # extraVolumes defines additional Kubernetes volumes for the server pod.
  # This can be used to mount operator-managed volumes without modifying the chart templates.
```

cbosdo (changes requested):

AkashKumar7902:
Thanks, addressed the general PR comment: added the server-helm changelog fragment and updated the PR description using the default template.

containers/server-helm/templates/server.yaml
AkashKumar7902: The `with` block already acts as the value check here. It only renders the block when `.Values.server.extraVolumeMounts` is non-empty, so empty/default values do not produce output. I can switch this to an explicit `if` if you prefer.

cbosdo (approved):

containers/server-helm/templates/server.yaml
cbosdo: > The `with` block already acts as the value check here. It only renders the block when `.Values.server.extraVolumeMounts` is non-empty, so empty/default values do not produce output. I can switch this to an explicit `if` if you prefer.

I didn't know that. Thanks for the info. You could have left it, but the new way is good too: no need to revert the change.

PR #12158, opened 2026-06-24
Automatic cleanup of Config and commits in test environment ()

CDellaGiusta:
## What does this PR change?

This PR aims to do an automatic cleanup of the tests. There are actually two cleanup issues:

- `Config` object: it can be set by calls like `Config.get().setString(configKey, newValue)`: since `Config` is a global object, they can interfere with other following tests
- Commits in the database: some of the methods in code do an explicit database commit, leaving some tables with unwanted persistent rows.

For both these issues, an automatic cleanup mechanism has been created, so to avoid forgetting any tracking and subsequent cleanup of the objects involved.

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/31017
Port(s): 
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/core/src/main/java/com/redhat/rhn/common/hibernate/HibernateCommitListener.java
```diff
@@ -0,0 +1,21 @@
...
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+
+package com.redhat.rhn.common.hibernate;
+
+@FunctionalInterface
+public interface HibernateCommitListener {
```
mackdk: There is a tag interface `java.util.EventListener` you should extend.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/EventListener.html

java/core/src/test/java/com/redhat/rhn/testing/TestUtils.java
```diff
@@ -486,6 +489,18 @@ public static void deleteLastMgrSyncRefresh() {
...
+    /**
+     * Deletes the org of a user, looking up the org id to check if it's committed in the database
+     * @param userIn the given user
+     */
+    public static void deleteOrgOfUser(User userIn) {
+        Stream.of(userIn)
+                .map(u -> u.getOrg().getId())
+                .map(OrgFactory::lookupById)
+                .filter(Objects::nonNull)
+                .forEach(org -> OrgFactory.deleteOrg(org.getId(), userIn));
```
mackdk: I think you should use an `Optional` here:
```java
        Optional.ofNullable(userIn)
            .map(u -> u.getOrg().getId())
            .map(OrgFactory::lookupById)
            .ifPresent(org -> OrgFactory.deleteOrg(org.getId(), userIn));
```

Moreover, I think it would be cleaner to ensure `deleteOrg` does not throw exception if the organization id is not present rather than look up twice the organization object we already have.

```java
    public static void deleteOrg(Long oid, User user) {
        Org org = OrgFactory.lookupById(oid);
        if (org == null) {
            return null;
        }

        ...
```
It's funny though, if you look at the implementation of `deleteOrg` the `id` is enough for all the operations it does.

mackdk (approved):

java/core/src/main/java/com/redhat/rhn/common/hibernate/HibernateCommitListener.java
CDellaGiusta: fixed

java/core/src/test/java/com/redhat/rhn/testing/TestUtils.java
CDellaGiusta: fixed with Optional

rjpmestre:
What is the scope of this cleanup? How is it intended to be used?

CDellaGiusta:
> What is the scope of this cleanup? How is it intended to be used?

When you run tests in random order, you find out that test pass only because they are executed always in the same order, since many of them leave leftovers behind. This PR aims at automatically cleaning up the stuff tests leave, in particular in the two points cited in the PR header: the Config object (if you force a Config setting it is automatically reset) and the commits in the database (if your test somehow have some hjidden commit, then it's automatically detected and cleaned, without your test having to remember/state the old `commitHappened() `method)

rjpmestre (approved):

PR #12170, opened 2026-06-25
Implement use of JUnit5 extensions for test salt utils

CDellaGiusta:
## What does this PR change?

Implement use of JUnit5 extensions for test salt utils, replacing SaltTestCaseUtils

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): https://github.com/SUSE/spacewalk/issues/29207
Port(s): 
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


mackdk (changes requested):
This is not the correct approach for implementing the extension in all the tests that require the salt root path, because with this implementation you will still need to extend a base class, which we don't want in the long run.

The correct approach would be to create a `@SaltRootPath` annotation and, in the `beforeEach` of the `SaltTestCaseExtension`, use reflection to set all the annotated fields. 

Something like:
```java
@Target(ElementType.FIELD)
@Retention(RetentionPolicy.RUNTIME)
public @interface SaltRootPath {
}

public class SaltTestCaseExtension implements BeforeEachCallback, AfterEachCallback {
    protected Path tmpSaltRoot;

    @Override
    public void beforeEach(ExtensionContext extensionContextIn) throws Exception {
        tmpSaltRoot = setupSaltConfigurationForTests();

        extensionContextIn.getRequiredTestInstances().getAllInstances().forEach(instance -> {
            Field[] instanceFields = instance.getClass().getDeclaredFields();
            for (Field field : instanceFields) {
                if (field.isAnnotationPresent(SaltRootPath.class)) {
                    if (!Path.class.equals(field.getType())) {
                        throw new IllegalStateException("Only Path fields can be annotated with @SaltRootPath");
                    }

                    try {
                        field.setAccessible(true);
                        field.set(instance, tmpSaltRoot);
                    }
                    catch (Exception ex) {
                        throw new IllegalStateException("Unable to set salt root path for field " + field, ex);
                    }
                }
            }
        });
    }
```

In this way in your test you can just have a field:

```java
    @SaltRootPath
    private Path saltRootPath;
```

and it will be initialised with the correct value in your test.

mackdk (approved):

java/core/src/test/java/com/redhat/rhn/testing/TestCaseExtension.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+public class TestCaseExtension {
```
rjpmestre, line 24: It would be nice to get some javadocs on the class and/or methods

java/core/src/test/java/com/redhat/rhn/testing/TestCaseExtension.java
```diff
@@ -0,0 +1,77 @@
...
+
+    protected List<Field> listAllFields(Object obj) {
+        List<Field> fieldList = new ArrayList<>();
+        Class<? extends Object> tmpClass = obj.getClass();
```
rjpmestre, line 28: By default `?` is implicitly bounded to Object ( like you have below in `Class<?> extensionObjectClass`)
thi one can be simplified to `Class<?>`

rjpmestre (approved):
Other than a couple of nitpicks, looks good :)

PR #12176, opened 2026-06-26
Grant db access to uyuni-server-attestation containers (bsc#1269267)

nadvornik:
## What does this PR change?

Grant db access to uyuni-server-attestation containers

Mgradm allows creating multiple replicas of the container, this PR allows at most 4 replicas.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- Cucumber tests: wip

- [x] **DONE**

## Links

https://bugzilla.suse.com/show_bug.cgi?id=1269267

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
+++ b/containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
@@ -26,7 +26,7 @@ if [ -f "$NAMESPACE_FILE" ] ; then
 host all all .kube-system.svc.cluster.local reject"
 else
     # podman
-    INTERNAL_ACCESS=uyuni-server
+    INTERNAL_ACCESS="uyuni-server uyuni-server-attestation-0 uyuni-server-attestation-1 uyuni-server-attestation-2 uyuni-server-attestation-3"
     EXTERNAL_RULES=""
 fi
 
--- a/containers/server-postgresql-image/server-postgresql-image.changes.nadvorni.allow_coco
+++ b/containers/server-postgresql-image/server-postgresql-image.changes.nadvorni.allow_coco
@@ -1,0 +1,1 @@
+- Grant access to uyuni-server-attestation container (bsc#1269267)


containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
```diff
@@ -26,7 +26,7 @@ if [ -f "$NAMESPACE_FILE" ] ; then
 host all all .kube-system.svc.cluster.local reject"
 else
     # podman
-    INTERNAL_ACCESS=uyuni-server
+    INTERNAL_ACCESS="uyuni-server uyuni-server-attestation-0 uyuni-server-attestation-1 uyuni-server-attestation-2 uyuni-server-attestation-3"
```
rjmateus, line 29: Can't we put it as a wildcard? This way we are assuming we will not have more than 4 parallel instances.
It should be more than enough, but it looks strange

nadvornik: unfortunately, it does not support wildcards

rjmateus: For podman, why don't we allow all connections from inside the podman network we create for uyuni deployment?
If a user gets access to the podman network he would already got access to much more, and it would not be a security problem.

mcalmer: I think it is not easy to find out what this network is.

nadvornik: mgradm would have to push the network details to the container and keep it up to date.

This  is self-contained.

mcalmer (approved):
For me this is ok. podman run all on the same host. Having more than 1 container for attestation is already special. 4 should be enough. And I think if there is a customer with special needs, he can now make use of custom rules and add more.

containers/server-postgresql-image/root/docker-entrypoint-upgdb.d/01-pg_hba.sh
rjmateus: agree.

aaannz: But technically this limits us to 4 replicas, right?

We could use `samenet` and explicitly forbid access from the podman host in a rule before that.
We can get this by parsing `/proc/net/route` and get gateway ip address.

But this change is definitely not for this time and release cycle.

rjmateus: `samenet` should only allow connection in the podman network, right? Since it's a virtual network that is linked as bridge to the host network, it should not allow connection from the host itself.
Or we can just get the ip of the host and set it "/24". 
According to Geminy, we should be able to get the sub-net from `cat /proc/net/fib_trie` (I didn't try it)

aaannz: Host and external connections will come from gateway and for them we mandate hostssl.

Rest are ruled by `host` entries, so those without ssl, thus must disallow access from the gateway.

Getting the IP of it is the only slightly complicated thing as there is no sugercandy like `ip` and we must decode ip ourselves.
And we cannot forget about IPv6 as soon ™️ there will be mandate on running IPv6 only stack.

PR #12203, opened 2026-07-03
Implement badge component

bisht-richa:
## What does this PR change?

**Implemented new status and informational badges**

## Codespace

[Create CodeSpace](https://codespaces.new/uyuni-project/uyuni)  [About billing for Github Codespaces](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces) [CodeSpace Billing Summary](https:
Screenshot 2026-07-01 at 15 12 03

Screenshot 2026-07-03 at 13 49 30
Screenshot 2026-07-02 at 10 16 05

Screenshot 2026-07-02 at 11 26 28

Screenshot 2026-07-02 at 11 38 00

Screenshot 2026-07-02 at 15 49 31

## GUI diff

- [x] **DONE**

## Documentation
- No documentation needed: **Implemented new status and informational badges**

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Issue(s): [#31085](https://github.com/SUSE/spacewalk/issues/31085)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/web/html/src/components/ComplianceBadge.tsx
+++ b/web/html/src/components/ComplianceBadge.tsx
@@ -1,3 +1,4 @@
+import { Badge } from "./badge/Badge";
 type Props = {
   percentage: number;
   compliant: number;
@@ -10,18 +11,15 @@ type Props = {
  */
 export const ComplianceBadge = ({ percentage, compliant, total }: Props) => {
   if (total === 0) {
-    return <span className="badge badge-default">{t("No Scans")}</span>;
+    return <Badge text={t("No Scans")} />;
   }
 
   const getColorClass = () => {
-    if (percentage >= 80) return "badge-success";
-    if (percentage >= 50) return "badge-warning";
-    return "badge-danger";
+    if (percentage >= 80) return "success";
+    if (percentage >= 50) return "warning";
+    return "error";
   };
 
-  return (
-    <span className={`badge ${getColorClass()}`}>
-      {compliant}/{total} ({percentage.toFixed(0)}%)
-    </span>
-  );
+  const badgeText = `${compliant}/${total} (${percentage.toFixed(0)}%)`;
+  return <Badge text={badgeText} color={getColorClass()} />;
 };
--- a/web/html/src/components/badge/Badge.test.tsx
+++ b/web/html/src/components/badge/Badge.test.tsx
@@ -1,4 +1,5 @@
 import { render, screen } from "@testing-library/react";
+
 import { Badge } from "./Badge";
 
 describe("Badge Component", () => {
@@ -12,6 +13,13 @@ describe("Badge Component", () => {
 
     expect(badge.className).toContain("badge--status");
     expect(badge.className).toContain("badge--default");
+  });
+
+  test("renders the small badge with the correct CSS class", () => {
+    render(<Badge text="Active" small />);
+    const badge = screen.getByText("Active");
+
+    expect(badge.className).toContain("badge--sm");
   });
 
   test("uses the default special color", () => {
--- a/web/html/src/components/badge/Badge.tsx
+++ b/web/html/src/components/badge/Badge.tsx
@@ -8,6 +8,7 @@ type BadgeProps =
       icon?: string;
       variant?: "status";
       color?: StatusColor;
+      title?: string;
       small?: boolean;
     }
   | {
@@ -29,7 +30,7 @@ export const Badge = (props: BadgeProps) => {
     .filter(Boolean)
     .join(" ");
   return (
-    <span className={badgeClassName} data-bs-toggle="tooltip" title={variant === "special" ? props.title : undefined}>
+    <span className={badgeClassName} title={props.title} {...(props.title ? { "data-bs-toggle": "tooltip" } : {})}>
       {icon && <i className={`fa ${icon}`} aria-hidden="true" />}
       {text}
     </span>
--- a/web/html/src/components/badge/Badge.tsx
+++ b/web/html/src/components/badge/Badge.tsx
@@ -8,22 +8,28 @@ type BadgeProps =
       icon?: string;
       variant?: "status";
       color?: StatusColor;
+      small?: boolean;
     }
   | {
       text: string;
       icon?: string;
       variant: "special";
       color?: SpecialColor;
+      title?: string;
+      small?: boolean;
     };
 
 export const Badge = (props: BadgeProps) => {
-  const { text, icon, variant = "status", color = variant === "status" ? "default" : "gray" } = props;
+  const { text, icon, small = false, variant = "status", color } = props;
+
+  const resolvedColor = color ?? (variant === "status" ? "default" : "gray");
   // Determine the class name based on the status prop
 
-  const badgeClassName = ["badge", `badge--${variant}`, `badge--${color}`].join(" ");
-
+  const badgeClassName = ["badge", `badge--${variant}`, `badge--${resolvedColor}`, small && "badge--sm"]
+    .filter(Boolean)
+    .join(" ");
   return (
-    <span className={badgeClassName}>
+    <span className={badgeClassName} data-bs-toggle="tooltip" title={variant === "special" ? props.title : undefined}>
       {icon && <i className={`fa ${icon}`} aria-hidden="true" />}
       {text}
     </span>
--- a/web/html/src/components/badge/Badge.tsx
+++ b/web/html/src/components/badge/Badge.tsx
@@ -30,7 +30,7 @@ export const Badge = (props: BadgeProps) => {
     .join(" ");
   const formatBadgeText = (text: string) => text.charAt(0).toUpperCase() + text.slice(1);
   return (
-    <span className={badgeClassName} data-bs-toggle="tooltip" title={variant === "special" ? props.title : undefined}>
+    <span className={badgeClassName} title={props.title} {...(props.title ? { "data-bs-toggle": "tooltip" } : {})}>
       {icon && <i className={`fa ${icon}`} aria-hidden="true" />}
       {formatBadgeText(text)}
     </span>
--- a/web/html/src/components/badge/Badge.tsx
+++ b/web/html/src/components/badge/Badge.tsx
@@ -30,7 +30,7 @@ export const Badge = (props: BadgeProps) => {
     .join(" ");
   const formatBadgeText = (text: string) => text.charAt(0).toUpperCase() + text.slice(1);
   return (
-    <span className={badgeClassName}>
+    <span className={badgeClassName} data-bs-toggle="tooltip" title={variant === "special" ? props.title : undefined}>
       {icon && <i className={`fa ${icon}`} aria-hidden="true" />}
       {formatBadgeText(text)}
     </span>
--- a/web/html/src/components/badge/Badge.tsx
+++ b/web/html/src/components/badge/Badge.tsx
@@ -30,7 +30,7 @@ export const Badge = (props: BadgeProps) => {
     .join(" ");
   const formatBadgeText = (text: string) => text.charAt(0).toUpperCase() + text.slice(1);
   return (
-    <span className={badgeClassName} title={props.title} {...(props.title ? { "data-bs-toggle": "tooltip" } : {})}>
+    <span className={badgeClassName}>
       {icon && <i className={`fa ${icon}`} aria-hidden="true" />}
       {formatBadgeText(text)}
     </span>
--- a/web/html/src/components/badge/CounterBadge.test.tsx
+++ b/web/html/src/components/badge/CounterBadge.test.tsx
@@ -4,12 +4,17 @@ import { CounterBadge } from "./CounterBadge";
 
 describe("CounterBadge Component", () => {
   test("renders the counter badge with the correct count", () => {
-    render(<CounterBadge count="1" />);
+    render(<CounterBadge count={1} />);
     expect(screen.getByText("1")).toBeTruthy();
   });
 
+  test("renders the counter badge with a large number value", () => {
+    render(<CounterBadge count={999} />);
+    expect(screen.getByText("999+")).toBeTruthy();
+  });
+
   test("uses the default status class", () => {
-    render(<CounterBadge count="1" />);
+    render(<CounterBadge count={1} />);
     const badge = screen.getByText("1");
 
     expect(badge.className).toContain("badge");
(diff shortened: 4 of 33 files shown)


web/html/src/components/badge/Badge.tsx
EMaksy, line 5: What do you think about extracting the shared props to reduce duplication?

```code
type StatusColor = "default" | "success" | "warning" | "error" | "info" | "running";

type SpecialColor = "gray" | "green" | "yellow" | "blue" | "red";

type DefaultBadgeProps = {
  text: string;
  icon?: string;
  title?: string;
  small?: boolean;
};

type StatusProps = DefaultBadgeProps & {
  variant?: "status";
  color?: StatusColor;
};

type SpecialProps = DefaultBadgeProps & {
  variant: "special";
  color?: SpecialColor;
};

type BadgeProps = StatusProps | SpecialProps;
```

web/html/src/components/badge/Badge.tsx
```diff
@@ -8,22 +8,29 @@ type BadgeProps =
...
       variant: "special";
       color?: SpecialColor;
+      title?: string;
+      small?: boolean;
     };
 
 export const Badge = (props: BadgeProps) => {
-  const { text, icon, variant = "status", color = variant === "status" ? "default" : "gray" } = props;
-  // Determine the class name based on the status prop
+  const { text, icon, small = false, variant = "status", color } = props;
```
EMaksy, line 24: We could also destruct in the function call like this what do you think ? 
```
export const Badge = ({
  text,
  icon,
  title,
  small = false,
  variant = "status",
  color,
}: BadgeProps) => {
```

With this we don't need also to use again props in line 33 `props.title --> title`

web/html/src/components/badge/Badge.tsx
```diff
@@ -8,22 +8,29 @@ type BadgeProps =
...
-  const badgeClassName = ["badge", `badge--${variant}`, `badge--${color}`].join(" ");
+  const resolvedColor = color ?? (variant === "status" ? "default" : "gray");
+  // Determine the class name based on the status prop
 
+  const badgeClassName = ["badge", `badge--${variant}`, `badge--${resolvedColor}`, small && "badge--sm"]
+    .filter(Boolean)
+    .join(" ");
   return (
-    <span className={badgeClassName}>
+    <span className={badgeClassName} title={props.title} {...(props.title ? { "data-bs-toggle": "tooltip" } : {})}>
```
EMaksy, line 33: We should try to avoid the usage of props.<property>, better destruct it in line 24 or in the function itself 
```code 
export const Badge = ({
  text,
  icon,
  title,
  small = false,
  variant = "status",
  color,
}: BadgeProps) => {
  const resolvedColor =
    color ?? (variant === "status" ? "default" : "gray");

  const badgeClassName = [
    "badge",
    `badge--${variant}`,
    `badge--${resolvedColor}`,
    small && "badge--sm",
  ]
    .filter(Boolean)
    .join(" ");

  return (
    <span
      className={badgeClassName}
      title={title}
      data-bs-toggle={title ? "tooltip" : undefined}
    >
      {icon && <i className={`fa ${icon}`} aria-hidden="true" />}
      {text}
    </span>
  );
};
``` 

what do you think ?

EMaksy: Oh i think we could also simplify data-bs-toggle here to  from `{...(props.title ? { "data-bs-toggle": "tooltip" } : {}` to -->       data-bs-toggle={title ? "tooltip" : undefined}

```
 return (
    <span
      className={badgeClassName}
      title={title}
      data-bs-toggle={title ? "tooltip" : undefined}
    >
      {icon && <i className={`fa ${icon}`} aria-hidden="true" />}
      {text}
    </span>
```

web/html/src/components/badge/Badge.test.tsx
```diff
@@ -14,6 +15,13 @@ describe("Badge Component", () => {
     expect(badge.className).toContain("badge--default");
   });
 
+  test("renders the small badge with the correct CSS class", () => {
```
EMaksy, line 18: It would be nice to add a test if the tooltip is rendered or not

```
  test("renders tooltip attribute when title is provided", () => {
    render(<Badge text="Active" title="This is active" />);
    const badge = screen.getByText("Active");

    expect(badge.getAttribute("title")).toBe("This is active");
    expect(badge.getAttribute("data-bs-toggle")).toBe("tooltip");
  });

  test("does not render tooltip attribute when title is not provided", () => {
    render(<Badge text="Active" />);
    const badge = screen.getByText("Active");

    expect(badge.getAttribute("data-bs-toggle")).toBeNull();
  });
```

web/html/src/manager/notifications/notifications.tsx
```diff
@@ -91,7 +93,7 @@ class Notifications extends Component<Props, State> {
         <i className={DEPRECATED_unsafeEquals(this.state.websocket, null) ? "fa fa-bell-slash" : "fa fa-bell"}></i>
         {!DEPRECATED_unsafeEquals(this.state.websocket, null) && this.state.unreadMessagesLength > 0 ? (
           <div id="notification-counter" className={this.state.classStyle}>
-            {this.state.unreadMessagesLength}
+            <CounterBadge count={this.state.unreadMessagesLength} status="highlight" />
```
EMaksy, line 96: In CounterBadge we changed the `count` prop type from `string` to `number`. 
I noticed that `this.state.unreadMessagesLength` has the type `any`, which we  should avoid.

After checking the Java backend, it returns a `long`:

```
private static Object prepareUserNotifications(Session session, User user) {
    return UserNotificationFactory.unreadUserNotificationsSize(user); 
}
``` 

``` 
public static long unreadUserNotificationsSize(User userIn) {
    return unreadUserNotificationsSize(userIn.getId());
}

public static long unreadUserNotificationsSize(long userIdIn) {
    // ... SQL Query ...
    CriteriaQuery<Long> count = criteria.select(builder.count(root));
    // ...
}
```

So with this i think we can set the type of unreadMessagesLength to number but, maybe this is even out of scope of this pr, i just wanted to share that chaning existing types could lead to problems

web/html/src/components/badge/CounterBadge.test.tsx
```diff
@@ -4,12 +4,17 @@ import { CounterBadge } from "./CounterBadge";
 
 describe("CounterBadge Component", () => {
   test("renders the counter badge with the correct count", () => {
-    render(<CounterBadge count="1" />);
+    render(<CounterBadge count={1} />);
     expect(screen.getByText("1")).toBeTruthy();
   });
 
+  test("renders the counter badge with a large number value", () => {
```
EMaksy, line 11: In addition to my comment in CounterBadge.tsx 
It would be nice to add more cases here, to catch diffrent values for example:
```code
test.each([
  [-1, "0"],
  [0, "0"],
  [999, "999"],
  [1000, "999+"],
])("renders count %s as %s", (count, expected) => {
  render(<CounterBadge count={count} />);
  expect(screen.getByText(expected)).toBeTruthy();
});
```

web/html/src/components/badge/CounterBadge.tsx
```diff
@@ -1,11 +1,12 @@
 type CounterBadgeProps = {
-  count?: string;
+  count: number;
   status?: "default" | "highlight";
 };
 
 export const CounterBadge = ({ count, status = "default" }: CounterBadgeProps) => {
   // Determine the class name based on the status prop
   const badgeClassName = `badge counter-badge badge--${status}`;
+  const displayCount = count >= 999 ? "999+" : count;
```
EMaksy, line 9: Do we allow negative numbers ?
What about the count 999 ? 999+ would let me expect then there are more then 999 :) 

We could add a guard here for negative numbers and 999, 1000 and so on:

```
  const safeCount = Math.max(0, count);
  const displayCount = safeCount >= 1000 ? "999+" : safeCount;
``` 
This would render it like this 
```
  -1 → 0
   0 → 0
 999 → 999
1000 → 999+
1500 → 999+
```
What do you think ? Maybe worth also to add tests here :)

web/html/src/components/ComplianceBadge.tsx
```diff
@@ -10,18 +11,15 @@ type Props = {
  */
 export const ComplianceBadge = ({ percentage, compliant, total }: Props) => {
```
EMaksy, line 12: It would be nice to add tests for ComplianceBadge :)

EMaksy: Something like this maybe ? 
``` 
import { render, screen } from "@testing-library/react";

import { ComplianceBadge } from "./ComplianceBadge";

describe("ComplianceBadge", () => {
  test("shows the default badge when there are no scans", () => {
    render(<ComplianceBadge percentage={0} compliant={0} total={0} />);

    expect(screen.getByText("No Scans")).toHaveClass(
      "badge--status",
      "badge--default"
    );
  });

  test.each([
    [80, "badge--success"],
    [50, "badge--warning"],
    [49, "badge--error"],
  ])("uses the correct color for %s percent", (percentage, expectedClass) => {
    render(
      <ComplianceBadge
        percentage={percentage}
        compliant={5}
        total={10}
      />
    );

    expect(screen.getByText(`5/10 (${percentage}%)`)).toHaveClass(expectedClass);
  });

  test("rounds the displayed percentage", () => {
    render(
      <ComplianceBadge percentage={66.6} compliant={2} total={3} />
    );

    expect(screen.getByText("2/3 (67%)")).toBeTruthy();
  });
});
```

web/html/src/manager/admin/setup/products/products.tsx
```diff
@@ -881,11 +882,7 @@ class CheckListItem extends Component<CheckListItemProps, CheckListItemState> {
         {...DEPRECATED_onClick(handleDescriptionClick)}
       >
         {currentItem.label}&nbsp;
-        {currentItem.recommended ? (
-          <span className="recommended-tag" title={"This extension is recommended"}>
-            {t("recommended")}
-          </span>
-        ) : null}
+        {currentItem.recommended ? <Badge text={t("Recommended")} small color="yellow" variant="special" /> : null}
```
EMaksy, line 885: Ehmm careful here, changing the badge text from lowercase `"recommended"` to capitalized `"Recommended"` will break existing Cucumber XPath tests. 

Note that test suites differ between Uyuni and MLM.... so if this works on uyuni it will break the ci on the backport! 

Checkout testsuite/features/step_definitions/setup_steps.rb line 136
```
Then(/^I should see that the "(.*?)" product is "(.*?)"$/) do |product, recommended|
  xpath = "//span[text()[normalize-space(.) = '#{product}'] and ./span/text() = '#{recommended}']"
  raise ScriptError, "xpath: #{xpath} not found" unless find(:xpath, xpath)
end
 
``` 



Check these tests:  
- testsuite/features/reposync/srv_sync_products.feature:55-57
- testsuite/features/build_validation/reposync/srv_sync_all_products.feature:863-865


For this we will need to adjust the tests or keep the lowercase "recommended", better to reach out  to QE before backporting

web/html/src/manager/images/image-view-overview.tsx
```diff
@@ -72,9 +72,10 @@ function StatusIcon(props: StatusIconProps) {
   } else if (action?.status === 2) {
     return (
       <span>
-        <i className="fa fa-check-circle fa-1-5x text-success" title={t("Successful")} />
+        <i className="fa fa-check-circle fa-1-5x text-success" data-bs-toggle="tooltip" title={t("Successful")} />
         <a
           title={t("Go to event")}
+          data-bs-toggle="tooltip"
```
EMaksy, line 78: is the repeated class in <i ... data-bs-toggle="tooltip" ... /> needed ? 

Maybe it is enough only on <a .... /> ?

web/html/src/manager/images/image-view-overview.tsx
```diff
@@ -84,9 +85,10 @@ function StatusIcon(props: StatusIconProps) {
   } else if (action?.status === 3) {
     return (
       <span>
-        <i className="fa fa-times-circle-o fa-1-5x text-danger" title={t("Failed")} />
+        <i className="fa fa-times-circle-o fa-1-5x text-danger" data-bs-toggle="tooltip" title={t("Failed")} />
```
EMaksy, line 88: is the repeated class in <i ... data-bs-toggle="tooltip" ... /> needed ?

Maybe it is enough only on <a .... /> ?

web/html/src/manager/salt/keys/key-management.tsx
```diff
@@ -67,26 +68,26 @@ function actionsFor(id, state, update, enabled) {
 
 const stateMapping = {
   accepted: {
-    uiName: t("accepted"),
+    uiName: t("Accepted"),
```
EMaksy, line 71: Here we have again issues with the cucumber test suite, we must be carefull, these change will break the cucumber ci on mlm again, please adjust the ci or keep them lowercase :) 
Maybe worth contacting QE on this how to adjust  before backporting to MLM

EMaksy (changes requested):
Hey @bisht-richa, great work! I left a few minor comments regarding tests and type usage, but the rest looks good to me.

There is one important issue that needs to be addressed: changing the status labels from lowercase to capitalized text breaks the affected Cucumber tests because their text assertions are case-sensitive.

For example:

```tsx
t("accepted") → t("Accepted")
```

The current PR CI already shows failures such as:

```text
Text 'accepted' not found
```

Uyuni and MLM use different test run sets, including scenarios tagged with `@uyuni` and `@susemanager`. We should therefore either preserve the existing lowercase labels or update all affected Cucumber expectations for the relevant environments. Please also consider the existing translation keys, which currently use the lowercase strings.

Some additional failures may only become visible when the changes are backported to MLM, so please have a look at the related comments.

Once the Cucumber tests and translations are addressed, this should be good to go. 🙂

web/html/src/manager/notifications/notifications.tsx
bisht-richa: Thanks for the update. However, I think the Special Badge is intended for displaying a counter, so it would be better for the count prop to be a number rather than a string. If this causes any issues, we can address them separately in the future.

web/html/src/manager/salt/keys/key-management.tsx
bisht-richa: I think I’ll update the Cucumber tests to match the new Badge changes. This way, we’ll keep the tests consistent with how Badge text is displayed and avoid this issue in the future.

web/html/src/components/badge/Badge.test.tsx
bisht-richa: Ideally, the Badge component should not include a tooltip. I’ve removed it since we currently don’t have any use cases for it.

PR #11286, opened 2025-12-04
SonarCloud error reduction fix step 11, refactoring to reduce cognitive complexity

CDellaGiusta:
## What does this PR change?

SonarCloud error reduction fix, rules:
[SonarCloud fix: java:S3776 Cognitive Complexity of methods should not be too high](https://sonarcloud.io/organizations/uyuni-project/rules?open=java%3AS3776&rule_key=java%3AS3776)

- ActionManager.java
- ErrataFactory.java
- ServerActionService.java
- CVEAuditManager.java
- KubernetesManager.java
- SystemHardwareAction.java

## Codespace

Check if you already have a running container clicking on [Running CodeSpace](https://github.com/codespaces)

[Create CodeSpace](https://codespaces.new/uyuni-project/uyuni)  [About billing for Github Codespaces](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces) [CodeSpace Billing Summary](https://github.com/settings/billing/summary) [CodeSpace Limit](https://github.com/settings/billing/spending_limit)

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- No documentation needed: only internal and user invisible changes
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): https://github.com/uyuni-project/uyuni/issues/9878
Port(s):
- [x] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

## Re-run a test

If you need to re-run a test, please mark the related checkbox, it will be unchecked automatically once it has re-run:

- [ ] Re-run test "changelog_test"
- [ ] Re-run test "backend_unittests_pgsql"
- [ ] Re-run test "java_pgsql_tests"
- [ ] Re-run test "schema_migration_test_pgsql"
- [ ] Re-run test "susemanager_unittests"
- [ ] Re-run test "frontend_checks"
- [ ] Re-run test "spacecmd_unittests"

# Before you merge

PR #12212, opened 2026-07-06
Restructuring of ActionType using enums

CDellaGiusta:
## What does this PR change?

As per [this discussion](https://github.com/uyuni-project/uyuni/pull/11286#discussion_r2601635962), an enum called ActionTypeEnum has been introduced to handle what now is handled by ActionType, plus getting  rid of complexity of the method Action.createAction (every ActionTypeEnum has an Action class to be created).

Te use of ActionTypeEnum then is spread to almost substitute ActionType when referring a specific action type, instead of comparing object looked up from the database.

ActionType could be completely substituted by ActionTypeEnum if it was not for the following 3 members

- private Character triggersnapshot;
- private Character unlockedonly;
 these two members are NOT used in the java code anywhere

- private String name;
- private boolean maintenancemodeOnly;
 these two members are currently actively used

**In summary:**
there may be a chance to get rid of the ActionType object moving name and maintenancemodeOnly in ActionTypeEnum, but this has to be evaluated in another PR 

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links
Issue(s): #
Port(s): 
5.2: https://github.com/SUSE/spacewalk/pull/31247
5.1: not backported
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
```diff
@@ -0,0 +1,175 @@
...
+        return label;
+    }
+
+    /**
+     * Create a new Action from scratch related to the enum.
+     * @return the created Action
+     * @throws ReflectiveOperationException if something goes wrong
+     */
+    public Action createAction() throws ReflectiveOperationException  {
+        return actionClass.asSubclass(Action.class).getDeclaredConstructor().newInstance();
```
mackdk: Why not simply?

```java
return actionClass.getDeclaredConstructor().newInstance();
```

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+    private ActionTypeEnum actionTypeEnum;
+    private User schedulerUser;
+    private Org org;
+    private String actionName;
+    private Date earliest;
+
+    /**
+     * default constructor
+     */
+    public ActionBuilder() {
```
rjpmestre: This constructor seems to be unnecessary

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public String getPackageActionName() {
+        return switch (this) {
+            case TYPE_PACKAGES_REMOVE -> "Package Removal";
+            case TYPE_PACKAGES_UPDATE -> "Package Install/Upgrade";
+            case TYPE_PACKAGES_VERIFY -> "Package Verify";
+            case TYPE_PACKAGES_REFRESH_LIST -> "Package List Refresh";
+            case TYPE_PACKAGES_DELTA -> "Package Synchronization";
+            case TYPE_PACKAGES_LOCK -> "Lock packages";
+            default -> "";
```
mackdk, line 149: I understand why you did it in this way, but this feels wrong on multiple levels. We shouldn't have a method returning an hardcoded string only for some specific actions, also considering that these strings are not localized.

Are these strings visible in the UI on the actions we generate? If so, they should be localized using the proper label (and there is no need to restrict this behaviour only to the package actions). If not, maybe we can change this logic and find a better alternative.

Anyway, I would keep it like this for now and have a deeper look in a separate PR.

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Retrieves an action type enum by its label.
+     *
+     * @param labelIn the label to search
+     * @return the ActionTypeEnum matching the given label, if present
+     */
+    public static Optional<ActionTypeEnum> of(String labelIn) {
+        return Arrays.stream(ActionTypeEnum.values())
```
mackdk, line 160: You can shortcut returning `Optional.empty()` if `labelIn` is `null`.

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+
+        Optional.ofNullable(org).
+                ifPresentOrElse(action::setOrg,
+                        () -> Optional.ofNullable(schedulerUser)
+                                .ifPresent(u -> action.setOrg(u.getOrg())));
+
+        Optional.ofNullable(actionName)
+                .ifPresentOrElse(action::setName,
+                        () -> Optional.ofNullable(actionType)
+                                .ifPresent(t -> action.setName(t.getName())));
```
mackdk: The `ifPresentOrElse`/`ifPresent` nesting here is quite hard to read for me. I would suggest this alternative version:
```java
        Optional.ofNullable(org)
            .or(() -> Optional.ofNullable(schedulerUser).map(User::getOrg))
            .ifPresent(action::setOrg);

        Optional.ofNullable(actionName)
            .or(() -> Optional.ofNullable(actionType).map(ActionType::getName))
            .ifPresent(action::setName);
```

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+
+        //in perl (modules/rhn/RHN/DB/Scheduler.pm) version is given a 2. So that's what I did.
+        action.setVersion(2L);
+        action.setArchived(0L); //not archived
```
mackdk, line 139: This smells like a boolean to me... another idea for a separate refactoring PR 🙂

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
```diff
@@ -0,0 +1,175 @@
...
+                .filter(type -> Strings.CS.equals(labelIn, type.getLabel()))
```
mackdk: Here I would simply use `Objects.equals()`

java/core/src/main/java/com/redhat/rhn/domain/action/ActionFactory.java
```diff
@@ -715,6 +465,31 @@ public static Action lookupById(Long id) {
...
+        }
+        return lookupActionTypeByLabel(actionTypeEnumIn.getLabel());
+    }
+
+    /**
+     * Checks if the action type is in maintenance mode only
+     * @param actionTypeEnumIn
+     * @return true if the action type exists and is in maintenance mode only
+     */
+    public static boolean isMaintenancemodeOnly(ActionTypeEnum actionTypeEnumIn) {
```
rjpmestre: I understand `isMaintenancemodeOnly` existing method already has the camelcase wrong. I'd avoid propagating it and rename this method to `isMaintenanceModeOnly`

java/core/src/main/java/com/redhat/rhn/frontend/action/schedule/ActionChainEditAction.java
```diff
@@ -192,4 +193,9 @@ private void setAttributes(HttpServletRequest request, DynaActionForm form,
     public void populateMaintenanceWindows(HttpServletRequest request, Set<Long> systemIds) {
         MaintenanceWindowHelper.prepopulateMaintenanceWindows(request, systemIds);
     }
+
+    @Override
+    public ActionTypeEnum referenceMaintenanceWindowsType() {
+        return null;
+    }
```
mackdk, line 200: I don't think this class needs to implement `MaintenanceWindowsAware`. You can just call directly `MaintenanceWindowHelper.prepopulateMaintenanceWindows(request, systemsWithMaintAwareAction)` on line 186, since this execution is not associated with an action type.

java/core/src/main/java/com/redhat/rhn/frontend/action/configuration/ssm/ConfigConfirmAction.java
```diff
@@ -81,23 +77,21 @@ protected void processForm(RequestContext ctxt, ActionForm formIn) {
...
         ctxt.getRequest().setAttribute("date", picker);
 
-        Set<Long> systems = getSystemIds(ctxt, ActionFactory.TYPE_CONFIGFILES_DEPLOY);
+        Set<Long> systems = getSystemIds(ctxt, ActionTypeEnum.TYPE_CONFIGFILES_DEPLOY);
         populateMaintenanceWindows(ctxt.getRequest(), systems);
         ActionChainHelper.prepopulateActionChains(ctxt.getRequest());
     }
 
-    private Set<Long> getSystemIds(RequestContext ctxt, ActionType actionType) {
+    private Set<Long> getSystemIds(RequestContext ctxt, ActionTypeEnum actionTypeEnum) {
```
rjpmestre: No need to parameterize `actionTypeEnum`. Its always '`ActionTypeEnum.TYPE_CONFIGFILES_DEPLOY`

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
```diff
@@ -4277,7 +4278,7 @@ private Long[] schedulePackagesAction(User loggedInUser, List<Integer> sids,
      * @return package action id
      */
     private Long schedulePackagesUpdateAction(User loggedInUser, List<Integer> sids,
-            Date earliestOccurrence, ActionType acT) {
+            Date earliestOccurrence, ActionTypeEnum actionTypeEnum) {
```
rjpmestre: This is also a private method invoked only by `SystemHandler.schedulePackageUpdate(...)`, always setting `actionTypeEnum to  'ActionTypeEnum.TYPE_PACKAGES_UPDATE'. No need to parameterize it.

rjpmestre (dismissed):

java/core/src/main/java/com/redhat/rhn/frontend/action/MaintenanceWindowsAware.java
```diff
@@ -34,7 +38,16 @@ public interface MaintenanceWindowsAware {
...
+        if (ActionFactory.isMaintenancemodeOnly(referenceMaintenanceWindowsType())) {
+            MaintenanceWindowHelper.prepopulateMaintenanceWindows(request, systemIds);
+        }
+    }
 
+    /**
+     * Retrieves the reference action type to check if the maintenance mode only is on
+     * @return the reference action type
+     */
+    ActionTypeEnum referenceMaintenanceWindowsType();
```
mackdk, line 51: This name is kind of confusing: I would suggest `getAssociatedActionType()` or `getMaintenenceActionType()`.

java/core/src/main/java/com/redhat/rhn/domain/action/ActionFactory.java
```diff
@@ -715,6 +465,31 @@ public static Action lookupById(Long id) {
...
+        ActionType actionType = lookupActionTypeByEnum(actionTypeEnumIn);
```
mackdk, line 490: I would short circuit this method to `false` if `actionTypeEnumIn` is `null` and document it in the javadoc. In this way you could just return `null` in `ImportFileConfirmAction` without having to override both methods.

java/core/src/main/java/com/redhat/rhn/frontend/action/MaintenanceWindowsAware.java
mackdk, line 1: In the end, this `MaintenanceWindowsAware` interface pattern was not really a good fit to begin with. Almost all the classes that implement it just call the method directly. In those cases, it would be not very different from including the check `ActionFactory.isMaintenancemodeOnly()` into an overloaded version of `MaintenanceWindowHelper.prepopulateMaintenanceWindows()` which also takes the `actionTypeEnum` as an additional parameter.

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Sets the action type
+     * @param actionTypeIn the action type
+     * @return the builder
+     */
+    public ActionBuilder ofType(ActionType actionTypeIn) {
+        actionTypeEnum = ActionTypeEnum.of(actionTypeIn).orElse(null);
+        return this;
+    }
```
mackdk, line 46: Maybe we should also store the `actionTypeIn` instance. In this way, we could call `ActionFactory.lookupActionTypeByLabel()` on line 117 only if it's still `null`.

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Retrieves an action type enum given an action type
+     *
+     * @param actionTypeIn the label to search
+     * @return the ActionTypeEnum matching the given label, if present
+     */
+    public static Optional<ActionTypeEnum> of(ActionType actionTypeIn) {
+        return actionTypeIn == null ? Optional.empty() : of(actionTypeIn.getLabel());
```
mackdk: Since you are returning an optional you could prefer the optional features over the ternary operator:
```java
        return Optional.ofNullable(actionTypeIn)
                .map(ActionType::getLabel)
                .flatMap(ActionTypeEnum::of);
```

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+                action.setActionType(actionType);
+            }
+            catch (ReflectiveOperationException eIn) {
+                LOG.error("Error while creating action of type {}", actionTypeEnum.getLabel(), eIn);
+                throw new RuntimeException(eIn);
+            }
+        }
+        else {
+            action = new Action();
+            actionType = null;
```
mackdk: Do you know which places in the code falls back to this `else`, if any? I tried to search for them, but I could not find any occurrence. If we have no explicit place that needs this, we could throw an illegal state exception and maybe also have the type as a constructor parameter, instead of being optional.

mackdk (approved):
Great change. I have left some minor nitpicking comments and some discussion points than can also be addressed in separate PRs.

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
CDellaGiusta: right. removed

java/core/src/main/java/com/redhat/rhn/domain/action/ActionFactory.java
CDellaGiusta: if not now, when? Challenge taken and method renamed

java/core/src/main/java/com/redhat/rhn/frontend/action/configuration/ssm/ConfigConfirmAction.java
CDellaGiusta: the method was short and called once only, so I removed it and substituted the call with the actual code

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/system/SystemHandler.java
CDellaGiusta: fixed

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
CDellaGiusta: fixed

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
CDellaGiusta: They all sink down to an argument to Action.setName(). But no idea if there are other implications, I just moved them around. Good idea for a separate PR

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
CDellaGiusta: much more expressive and clear, suggestion taken

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
CDellaGiusta: I have no idea of the meaning of these two lines, I just found them and moved them. Separate PR!

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
CDellaGiusta: fixed

java/core/src/main/java/com/redhat/rhn/frontend/action/schedule/ActionChainEditAction.java
CDellaGiusta: I would like to keep the removal of interface MaintenanceWindowsAware in a separate PR

java/core/src/main/java/com/redhat/rhn/frontend/action/MaintenanceWindowsAware.java
CDellaGiusta: since we may get rid of all them, let's keep this name for now

java/core/src/main/java/com/redhat/rhn/domain/action/ActionFactory.java
CDellaGiusta: fixed

java/core/src/main/java/com/redhat/rhn/frontend/action/MaintenanceWindowsAware.java
CDellaGiusta: As said, we'll try to remove the whole interface in another PR

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
CDellaGiusta: suggestion taken

java/core/src/main/java/com/redhat/rhn/domain/action/ActionTypeEnum.java
CDellaGiusta: suggestion taken

java/core/src/main/java/com/redhat/rhn/domain/action/ActionBuilder.java
CDellaGiusta: yes, there are 12 action types present in the database that have no entry as an ActionTypeEnum value, because the old createAction had a fallback mechanism in which after that old huge if chain, it has an else clause with new Action()
So I reported this fallback mechanism here as well.
4 action types are actually used in python code:
- activation.schedule_pkg_install
- activation.schedule_deploy
- kickstart_host.schedule_virt_host_pkg_install
- kickstart_guest.schedule_virt_guest_pkg_install

the remaining are possible leftovers:

- solarispkgs.install
- solarispkgs.remove
- solarispkgs.patchInstall
- solarispkgs.patchRemove
- solarispkgs.patchClusterInstall
- solarispkgs.patchClusterRemove
- solarispkgs.refresh_list
- proxy.deactivate

mackdk (dismissed):

PR #12223, opened 2026-07-08
QE: Move the installation of RKE2 on slmicro6.2 to the testsuite

Pablogoliva:
## What does this PR change?

Move the installation of RKE2 on slmicro6.2 to the testsuite

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed

- [x] **DONE**

## Test coverage
- Cucumber tests were edited

- [x] **DONE**

## Links

Issue(s): https://github.com/orgs/SUSE/projects/32/views/25?filterQuery=assignee%3A%40me&pane=issue&itemId=216718511&issue=SUSE%7Cspacewalk%7C31335
Port(s): 
- Manager 5.2: 

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge


maximenoel8:
See copilot comments. 
Instead of using curl, could you add it to upload_files.
I found in the fast that dynamically downloading files with curl is causing a lot of testsuite instability, or we should have the package on mirror, something clause to our server.

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,41 @@
...
+# Replaces the sumaform salt/kubernetes_common/install_rke2.sls state for SL-Micro 6.2.
+# Before running this feature, the SCC_SLMICRO_REGCODE environment variable must be
+# available on the server node so the SCC registration step can resolve it.
+
+Feature: Install RKE2 on SL-Micro 6.2
+  In order to deploy Uyuni as a Kubernetes-based application on SL-Micro 6.2
+  As the system administrator
+  I want to install and configure RKE2 on the server
+
+  Scenario: Register the server to SCC to access RKE2 packages
```
maximenoel8: If we want to test devel, we need to continue using devel repository IMPOV

Pablogoliva: Yes, it is done with sumaform. So this comment is outdated.

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,41 @@
...
+    Then service "rke2-server" is enabled on "server"
+    And service "rke2-server" is active on "server"
+
+  Scenario: Create symlinks for RKE2 tools
+    When I run "mkdir -p /usr/local/bin && ln -sf /var/lib/rancher/rke2/bin/kubectl /usr/local/bin/kubectl" on "server"
+    And I run "ln -sf /var/lib/rancher/rke2/bin/crictl /usr/local/bin/crictl" on "server"
+    And I run "ln -sf /var/lib/rancher/rke2/bin/ctr /usr/local/bin/ctr" on "server"
+
+  Scenario: Set up RKE2 environment variables
+    When I run "printf 'export PATH=$PATH:/opt/rke2/bin\nexport KUBECONFIG=/etc/rancher/rke2/rke2.yaml\n' > /etc/profile.d/rke2_vars.sh" on "server"
```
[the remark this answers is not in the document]
Pablogoliva: Outdated.

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,41 @@
...
+    When I run "transactional-update register -r $SCC_SLMICRO_REGCODE" on "server"
+    And I run "transactional-update apply" on "server"
+
+  Scenario: Create the RKE2 configuration file
+    When I run "mkdir -p /etc/rancher/rke2" on "server"
+    And I run "printf 'tls-san:\n  - %s\ningress-controller: traefik\nselinux: true\nkubelet-arg:\n  - seccomp-default=true\n' $(hostname -f) > /etc/rancher/rke2/config.yaml" on "server"
+    Then file "/etc/rancher/rke2/config.yaml" should exist on "server"
+
+  Scenario: Install RKE2 via RPM method
+    When I run "INSTALL_RKE2_VERSION='v1.35.4+rke2r1' INSTALL_RKE2_METHOD=rpm INSTALL_RKE2_SELINUX=true curl -sfL https://get.rke2.io | sh -" on "server"
```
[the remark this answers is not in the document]
Pablogoliva: Outdated

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+#
+# Replaces the sumaform salt/kubernetes_common/install_rke2.sls state for SL-Micro 6.2.
+# Before running this feature, the SCC_SLMICRO_REGCODE environment variable must be
+# available on the server node so the SCC registration step can resolve it.
+
+Feature: Install RKE2 on SL-Micro 6.2
```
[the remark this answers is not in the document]
Pablogoliva: Done

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@transactional_server
+@rke2
+Feature: Install RKE2 on transactional systems
+
+  Scenario: Check that RKE2 configuration
```
ktsamis: Small one. "Check that RKE2 configuration" is missing a word. Maybe "Check the RKE2 configuration".

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,66 @@
...
+    Then the environment variable "RKE2_VERSION" is set on "server"
+    And the environment variable "CERT_MANAGER_VERSION" is set on "server"
+    And the environment variable "CERT_MANAGER_NAMESPACE" is set on "server"
+    And the environment variable "TRAEFIK_FILE" is set on "server"
+    And the environment variable "LOCAL_PATH_STORAGE_FILE" is set on "server"
+    And the environment variable "LOCAL_PATH_PROVISIONER_STORAGE_FILE" is set on "server"
+    And the environment variable "LOCAL_PATH" is set on "server"
+    And the environment variable "LOCAL_PATH_NAMESPACE" is set on "server"
+    And the environment variable "SELF_SIGNED_PATH" is set on "server"
```
ktsamis: Three of these names do not match what sumaform #2243 exports:

| here | sumaform #2243 |
| --- | --- |
| `TRAEFIK_FILE` | `TRAEFIK_PATH` |
| `LOCAL_PATH_STORAGE_FILE` | `LOCAL_PATH_PROVISIONER_PATH` |
| `SELF_SIGNED_PATH` | not exported at all, it is only a Jinja variable |

So this scenario fails right away. Could we pick one set of names and use it in both PRs?

There is also a fourth one on the sumaform side. `variables_local-path-provisioner` sits inside a guard that now skips SL Micro, so `LOCAL_PATH` and `LOCAL_PATH_NAMESPACE` never get written either. I left a comment there too.

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,66 @@
...
+    And the environment variable "HELM_CHART_DIRECTORY" is set on "server"
+    And the environment variable "KUBECONFIG" is set on "server"
+    And file "/etc/rancher/rke2/config.yaml" should exist on "server"
+
+  Scenario: Install RKE2 via RPM method
+    When I run "curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION='$RKE2_VERSION' INSTALL_RKE2_METHOD=rpm sh -" on "server" outside the container
```
ktsamis: The single quotes stop the shell from expanding this. I tried it on mlm-ci-head-rke2-server:

```
single-quoted (as here): [$KUBECONFIG]
double-quoted:           [/etc/rancher/rke2/rke2.yaml]
```

So `INSTALL_RKE2_VERSION` gets the literal string `$RKE2_VERSION`. The install script then builds a repo URL out of it and runs `zypper install rke2-server-$RKE2_VERSION`, which fails.

Double quotes fix it. The good news is that `INSTALL_RKE2_METHOD=rpm` does pass through sudo, I checked that on the same box, so the inline approach works.

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,66 @@
...
+    And service "rke2-server" is active on "server"
+
+  Scenario: Create symlinks for RKE2 tools
+    When I run "ln -sf /var/lib/rancher/rke2/bin/kubectl /usr/local/bin/kubectl" on "server"
+    And I run "ln -sf /var/lib/rancher/rke2/bin/crictl /usr/local/bin/crictl" on "server"
+    And I run "ln -sf /var/lib/rancher/rke2/bin/ctr /usr/local/bin/ctr" on "server"
+
+## Install helm
+  Scenario: Install Helm
+    When I run "curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash" on "server" outside the container
```
ktsamis: Copilot raised this on line 23 and it got marked outdated, but I think the point still stands here.

`curl ... | bash` takes its exit code from bash, so if curl fails the step still passes. This one has no `-f` either, so an HTML error page would get piped into bash.

`set -o pipefail`, or downloading first and then running it, would make it fail properly.

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,66 @@
...
+    When I apply the RKE2 YAML file "$LOCAL_PATH_STORAGE_FILE" on "server"
+    And I set "$LOCAL_PATH_PROVISIONER_STORAGE_FILE" storage class as default on "server"
+    And I run "mkdir -p $LOCAL_PATH" on "server" outside the container
+    And I run "restorecon -R -v $LOCAL_PATH" on "server" outside the container
+    And I run "kubectl delete pods --all -n $LOCAL_PATH_NAMESPACE" on "server" outside the container
+
+## Install MLM/Uyuni
+  Scenario: Install Uyuni
+    When I run "cd $SELF_SIGNED_PATH && helm dependencies build" on "server" outside the container
+    And I run "cd $SELF_SIGNED_PATH && helm upgrade --install uyuni ./selfsigned -f ./selfsigned/values.yaml -n uyuni" on "server" outside the container
```
ktsamis: `-n uyuni` does not create the namespace. sumaform #2243 puts `create_uyuni_namespace` behind a guard that skips SL Micro, so nothing creates it any more.

Adding `--create-namespace` here would cover it, same as the cert-manager scenario above.

Same question for the persistent volumes. `apply_var_spacewalk_file` and `apply_var_pgsql_file` are also skipped for SL Micro now, and nothing here replaces them. Are they still needed, or does local-path handle it?

testsuite/features/support/commonlib.rb
```diff
@@ -49,6 +49,7 @@ def count_table_items
 # @return [String] The product name.
 def product
   return $product unless $product.nil?
+  return if get_target('server').run_local('[[ "${UYUNI_NOT_INSTALLED:-}" == "true" ]]', check_errors: false).last.zero?
```
ktsamis: This returns nil without setting `$product`, so the cache on the line above never kicks in.

`Before('@uyuni')` and `Before('@susemanager')` in env.rb both call `product`, so this ends up running an SSH command before every tagged scenario.

A nil product also breaks `BASE_CHANNEL_BY_CLIENT[product][client]` further down in this file and in api_common.rb, if anything reaches it.

testsuite/features/support/env.rb
```diff
@@ -148,8 +148,7 @@ def capybara_register_driver
 World(MiniTest::Assertions)
 
 # Initialize the API client
-$api_test = new_api_client
-
+$api_test = new_api_client unless get_target('server').run_local('[[ "${UYUNI_NOT_INSTALLED:-}" == "true" ]]', check_errors: false).last.zero?
```
ktsamis: This same check is now in four places, and two of them run at file load time. That is an extra SSH round trip for every cucumber process in every environment, not only RKE2.

Could we work it out once here into a global, and let the other three read that?

testsuite/run_sets/install_mlm_on_rke2.yml
```diff
@@ -0,0 +1,11 @@
+# This file describes the order of features in a normal testsuite run.
+#
+# If you create new features, please see conventions about naming of the
+# feature files in testsuite/documentation/guidelines.md in "Rules for features" chapter,
+# as well as guidelines about idempotency in "Idempotency" chapter.
+
+## Secondary features BEGIN ##
```
ktsamis: Copy paste leftover. This says "Secondary features BEGIN" but it is an install run set.

ktsamis (changes requested):
The approach makes sense to me. Moving the install into the testsuite is the right call, because cucumber can reboot and wait where a salt state cannot.

I tested a couple of things on mlm-ci-head-rke2-server and found some issues. Two of them will stop the run on the first try, the rest are smaller. Comments inline.

One general point. Every step here is its own Scenario. If one fails, the rest still run against a half installed system. Merging install, reboot and enable into one scenario would fail earlier and read better.

ktsamis:
Just a note to rebase this before merging, master has moved on a fair bit since it was opened.

testsuite/features/kubernetes/srv_install_rke2.feature
Pablogoliva: The variables are already addressed and added.

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@transactional_server
+@rke2
+@no_user_creation
+Feature: Install RKE2 on transactional systems
+
+  Scenario: Reboot the server to activate to active everything before starting
```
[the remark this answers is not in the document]
Pablogoliva: Done

testsuite/features/support/env.rb
```diff
@@ -293,6 +292,14 @@ def process_code_coverage
   process_code_coverage
 end
 
+After('@install_mlm_on_rke2') do |scenario|
+  return unless scenario.passed?
+
+  bashrc_file = '/root/.bashrc'
+  get_target('localhost').run("sed -i 's/export UYUNI_NOT_INSTALLED=true/export UYUNI_NOT_INSTALLED=false/g' #{bashrc_file}")
+  get_target('localhost').run("source #{bashrc_file}")
+end
```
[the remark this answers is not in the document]
Pablogoliva: I have tried and this thing just changes the value of the env variable during runtime. In the next stage it completely changes and if we start the api_test instance here we are making it more complicated.

testsuite/features/support/remote_node.rb
```diff
@@ -52,8 +55,13 @@ def initialize(host, port: 22)
...
     # Determine OS version and OS family both inside the container and on the local host
     # in the case of non-containerized systems, both fields will be identical:
-    @os_version, @os_family = get_os_version
     @local_os_version, @local_os_family = get_os_version(runs_in_container: false)
+    if !uyuni_not_installed:
+      @os_version, @os_family = get_os_version()
+    else
+      @os_version = @local_os_version
+      @os_family = @local_os_family
+    end
```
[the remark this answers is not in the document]
Pablogoliva: Outdated, Rubocop complained about it.

testsuite/features/support/commonlib.rb
```diff
@@ -49,6 +49,7 @@ def count_table_items
 # @return [String] The product name.
 def product
   return $product unless $product.nil?
+  return if ENV.key?('UYUNI_NOT_INSTALLED') && ENV['UYUNI_NOT_INSTALLED'] == 'true'
```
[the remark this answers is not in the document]
Pablogoliva: It's intented. Whem that variable is in that point it returns as expexcted.

Pablogoliva:
> See copilot comments. Instead of using curl, could you add it to upload_files. I found in the fast that dynamically downloading files with curl is causing a lot of testsuite instability, or we should have the package on mirror, something clause to our server.

We can download the script and leave it in the machine. I prefer the curl way because eveything is updated, but no strong opinion. Also we can see in the next week how it goes. https://docs.rke2.io/install/methods#tarball

testsuite/features/kubernetes/srv_install_rke2.feature
```diff
@@ -0,0 +1,33 @@
...
+
+  Scenario: Reboot the server to activate everything before starting
+    When I reboot the "server" host through SSH, waiting until it comes back
+
+  Scenario: Check the RKE2 configuration
+    Then the environment variable "RKE2_VERSION" is set on "server"
+    And file "/etc/rancher/rke2/config.yaml" should exist on "server"
+
+  Scenario: Install RKE2 via RPM method
+    When I run "curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION=$RKE2_VERSION INSTALL_RKE2_METHOD=rpm sh -" on "server"
```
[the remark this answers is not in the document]
Pablogoliva: That the same, as Maxime said we can download the script instead of the curl way. But this way is the same but saving the script, so it not relevant.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
```diff
@@ -0,0 +1,39 @@
+
+
+@transactional_server
+@rke2
+@no_user_creation
+Feature: Install MLM dependencies on RKE2
```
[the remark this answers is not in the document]
Pablogoliva: True, the copyright is missing.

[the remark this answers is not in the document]
Pablogoliva: Added.

testsuite/features/kubernetes/srv_install_mlm_on_rke2.feature
```diff
@@ -0,0 +1,21 @@
+## Install MLM/Uyuni
+
+@transactional_server
+@rke2
+@no_user_creation
+Feature: Install MLM on RKE2
```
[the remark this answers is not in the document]
Pablogoliva: Added.

testsuite/features/support/remote_node.rb
```diff
@@ -32,13 +32,16 @@ def initialize(host, port: 22)
     @hostname = out.strip
     raise LoadError, "We can't connect to #{@host} through SSH." if @hostname.empty?
 
+    # A Bare Host has no container to exec into, even though the mgrctl binary ships in the base image.
+    uyuni_not_installed = !ssh('kubectl get deployment uyuni -n uyuni', host: @target).last.zero? && !ssh('podman container exists uyuni-server', host: @target).last.zero?
+
     $named_nodes[host] = @hostname
     if @host == 'server'
-      @has_mgrctl = ssh('which mgrctl', host: @target).last.zero?
+      @has_mgrctl = ssh('which mgrctl', host: @target).last.zero? && !uyuni_not_installed
```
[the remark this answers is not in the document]
Pablogoliva: Yes that's intended. I want it to use mgrctl if MLM is installed and avoid it if not because it's not going to work.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
```diff
@@ -0,0 +1,39 @@
...
+    And the environment variable "LOCAL_PATH_PROVISIONER_PATH" is set on "server"
+    And the environment variable "LOCAL_PATH_PROVISIONER_STORAGE_CLASS" is set on "server"
+    And the environment variable "LOCAL_PATH" is set on "server"
+    And the environment variable "LOCAL_PATH_NAMESPACE" is set on "server"
+    And file "/etc/rancher/rke2/config.yaml" should exist on "server"
+
+  ## Install helm
+  Scenario: Install Helm
+    When I run "curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash" on "server"
+
```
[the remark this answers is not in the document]
Pablogoliva: That's the official way to install helm, I see no reason to modify it.

testsuite/features/support/env.rb
```diff
@@ -293,6 +292,14 @@ def process_code_coverage
   process_code_coverage
 end
 
+After('@install_mlm_on_rke2') do |scenario|
+  return unless scenario.passed?
```
ktsamis: `return` in a cucumber hook raises `LocalJumpError: unexpected return`. The block is defined in a file cucumber `require`s, so by the time the hook runs the enclosing top level is gone. I reproduced it outside the testsuite:

```
ERROR: LocalJumpError: unexpected return
```

It only fires when the scenario fails, which is exactly when you want the rest of the run to carry on. Every other hook in this file uses `next`. Please use `next unless scenario.passed?`.

testsuite/features/support/env.rb
```diff
@@ -293,6 +292,14 @@ def process_code_coverage
...
+
+  bashrc_file = '/root/.bashrc'
+  get_target('localhost').run("sed -i 's/export UYUNI_NOT_INSTALLED=true/export UYUNI_NOT_INSTALLED=false/g' #{bashrc_file}")
+  get_target('localhost').run("source #{bashrc_file}")
```
ktsamis: This line does nothing. It opens its own SSH session, sources the file there, and throws the session away. Neither the cucumber process nor any later command sees it.

The `sed` above is what actually carries the state, because each pipeline stage starts a fresh login shell. Please drop the `source` and put that reason in a comment, otherwise the next reader takes this line to mean the variable is live from here on.

testsuite/features/kubernetes/srv_install_mlm_on_rke2.feature
```diff
@@ -0,0 +1,22 @@
...
+    And the environment variable "HELM_CHART_DIRECTORY" is set on "server"
+    And the environment variable "SELF_SIGNED_PATH" is set on "server"
+    And file "/etc/rancher/rke2/config.yaml" should exist on "server"
+
+  Scenario: Update OCI app version
+    And I run "python3 $PYTHON_HELM_CHART_PATH -o $HELM_CHART_URL/$HELM_CHART_NAME --chart-file $SELF_SIGNED_PATH/Chart.yaml $DEVEL_FLAG" on "server"
+
+@install_mlm_on_rke2
+  Scenario: Install Uyuni
+    And I run "kubectl create namespace uyuni" on "server"
```
ktsamis: `kubectl create namespace uyuni` fails with `AlreadyExists` on a second run, and `testsuite/run_sets/install_mlm_on_rke2.yml` is headed `# IDEMPOTENT`. Re-running the stage against the same environment will not work.

```
kubectl create namespace uyuni --dry-run=client -o yaml | kubectl apply -f -
```

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
```diff
@@ -0,0 +1,40 @@
...
+  Scenario: Install cert-manager and trust-manager
+    When I run "helm install cert-manager oci://quay.io/jetstack/charts/cert-manager --version $CERT_MANAGER_VERSION --namespace $CERT_MANAGER_NAMESPACE --create-namespace --set crds.enabled=true --timeout 10m0s --wait" on "server"
```
ktsamis: Same idempotency problem: `helm install` fails if the release already exists. trust-manager two lines down already uses `helm upgrade --install`. Please make cert-manager match it.

ktsamis (changes requested):
Rebase is in, thanks, and the env var names line up with sumaform now. I checked all seven the dependencies feature asserts and they are all exported outside the SL Micro guards.

What is left is mostly runtime behaviour rather than style.

Two things that do not need a code comment:

- Copilot's "missing copyright header" comments on the two new feature files are false positives. Both have it on lines 1-2.
- `testsuite/features/support/env.rb` changed mode from 100644 to 100755 in this PR. That looks accidental, please revert it.

Order: https://github.com/uyuni-project/sumaform/pull/2270 should merge before this, and https://github.com/SUSE/susemanager-ci/pull/2191 last.

testsuite/features/kubernetes/srv_install_rke2.feature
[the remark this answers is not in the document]
ktsamis: This one is real, please do not mark it outdated again. With `-sfL` a failing curl leaves `sh` with empty stdin and `sh` exits 0, so the step passes with RKE2 not installed. Prefixing the command with `set -o pipefail; ` fixes it.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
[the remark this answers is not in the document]
ktsamis: Agreed, and there is a second problem on the same line: the script comes from helm's `main` branch, unpinned. `rke2_version` and `cert_manager_version` are both pinned in sumaform, so helm is the odd one out here and a new helm release breaks this without anything changing on our side.

Please pin the ref and add `set -o pipefail; ` and `-fsSL`. That also covers what @maximenoel8 asked for further up.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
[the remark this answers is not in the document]
ktsamis: False positive, the header is on lines 1-2 of this file.

testsuite/features/kubernetes/srv_install_mlm_on_rke2.feature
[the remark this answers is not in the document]
ktsamis: False positive, the header is on lines 1-2 of this file.

testsuite/features/support/env.rb
Pablogoliva: Done

testsuite/features/kubernetes/srv_install_mlm_on_rke2.feature
Pablogoliva: But, we never execute sumaform twice nor the isolated salt states. We never re-run salt states a second time trough the same environment.

testsuite/features/support/env.rb
Pablogoliva: That's true, the source command is useless.
Why do you want me to leave a comment? I don't understand this statement: `otherwise the next reader takes this line to mean the variable is live from here on.` If someone sees that the variable is changed in the bashrc file why is going to think that the variable is live on?

Pablogoliva: Source line deleted.

testsuite/features/kubernetes/srv_install_mlm_on_rke2.feature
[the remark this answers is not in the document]
Pablogoliva: Because I added them, but in the first revision the weren't.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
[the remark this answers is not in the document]
Pablogoliva: Yes, but because I added them after the first revision.

testsuite/features/kubernetes/srv_install_mlm_on_rke2.feature
Pablogoliva: But it's ok I guess. We can do it for sumaform idempotency.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
Pablogoliva: As I said above, we never execute salt states twice. But it's ok I guess. Adding that thing.

testsuite/features/kubernetes/srv_install_mlm_on_rke2.feature
Pablogoliva: Nervermind, that's the testusite not sumaform. My bad, it's true it can be used more than once.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
Pablogoliva: Nervermind, that's the testusite not sumaform. My bad, it's true it can be used more than once. But added those states.

testsuite/features/kubernetes/srv_install_rke2.feature
ktsamis: Fixed.

testsuite/features/kubernetes/srv_install_rke2.feature
ktsamis: No longer an issue, but only incidentally: `@has_mgrctl` is false before MLM is installed, so `run` stops wrapping the command in `mgrctl exec '...'` and the remote shell expands it. Nothing was changed to make that true, so it comes back if that logic moves.

testsuite/features/kubernetes/srv_install_rke2.feature
ktsamis: Still open on the current head, line 17 is unchanged. A failed download leaves `sh` with empty stdin and the step passes with RKE2 not installed. `set -o pipefail; ` in front is all it needs.

testsuite/features/kubernetes/srv_install_rke2.feature
ktsamis: Namespace part is fixed, thanks.

The second half is still open. `apply_var_spacewalk_file` and `apply_var_pgsql_file` sit behind an `is_supported_os` that excludes SL Micro, and nothing here replaces them. Every other local-path environment gets them by default, so either they are needed here too or that default is wrong elsewhere. Which one is it?

testsuite/features/support/commonlib.rb
ktsamis: Two corrections to my own comment: the early return sits before the SSH call, so there is no extra round trip, and the four checks are ENV lookups.

What is left is the `@return [String]` doc and nil reaching `BASE_CHANNEL_BY_CLIENT[product][client]`. The install run sets never get there, so updating the doc comment is enough for me.

testsuite/features/support/env.rb
ktsamis: Same correction as on commonlib.rb: these are ENV lookups, they cost nothing. Dropping the round trip claim.

A single `uyuni_not_installed?` helper would still beat four copies of the condition.

testsuite/run_sets/install_mlm_on_rke2.yml
ktsamis: Still there on the current head, and `install_rke2.yml:7` has the same copy paste.

ktsamis (commented):
Re-reviewed against the current head. Most of the earlier points are fixed, I replied in the threads that are still open. Four new ones below.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
```diff
@@ -0,0 +1,40 @@
...
+    And the environment variable "TRAEFIK_FILE" is set on "server"
+    And the environment variable "LOCAL_PATH_PROVISIONER_PATH" is set on "server"
+    And the environment variable "LOCAL_PATH_PROVISIONER_STORAGE_CLASS" is set on "server"
+    And the environment variable "LOCAL_PATH" is set on "server"
+    And the environment variable "LOCAL_PATH_NAMESPACE" is set on "server"
+    And file "/etc/rancher/rke2/config.yaml" should exist on "server"
+
+  ## Install helm
+  Scenario: Install Helm
+    When I run "curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash" on "server"
```
ktsamis, line 21: Re-raising this, the thread was resolved but the line is unchanged.

- The script comes from helm's `main`, unpinned. `rke2_version` and `cert_manager_version` are both pinned in sumaform, so helm is the odd one out and a new helm release breaks this with nothing changing on our side.
- `curl ... | bash` with no `-fsSL` and no `set -o pipefail; ` still passes when the download fails, and an HTML error page gets piped into bash.

Please pin the ref and add both.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
```diff
@@ -0,0 +1,40 @@
...
+
+  Scenario: Install cert-manager and trust-manager
+    When I run "helm upgrade --install cert-manager oci://quay.io/jetstack/charts/cert-manager --version $CERT_MANAGER_VERSION --namespace $CERT_MANAGER_NAMESPACE --create-namespace --set crds.enabled=true --timeout 10m0s --wait" on "server"
+    Then I wait until "cert-manager" helm chart is deployed in namespace "cert-manager" on "server"
+    When I run "helm upgrade --install trust-manager oci://quay.io/jetstack/charts/trust-manager --install --namespace $CERT_MANAGER_NAMESPACE --wait" on "server"
```
ktsamis: `--install` is passed twice now, once as part of `helm upgrade --install` and once further along the line. Harmless, but drop one.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@transactional_server
+@rke2
+@no_user_creation
+Feature: Install MLM dependencies on RKE2
```
ktsamis: Trailing whitespace here and on line 8.

Separately, the scenario below opens with `And`; `Then` reads better as a first step. Same in `srv_install_mlm_on_rke2.feature`.

testsuite/features/support/remote_node.rb
```diff
@@ -32,13 +32,16 @@ def initialize(host, port: 22)
     @hostname = out.strip
     raise LoadError, "We can't connect to #{@host} through SSH." if @hostname.empty?
 
+    # A Bare Host has no container to exec into, even though the mgrctl binary ships in the base image.
+    uyuni_not_installed = !ssh('kubectl get deployment uyuni -n uyuni', host: @target).last.zero? && !ssh('podman container exists uyuni-server', host: @target).last.zero?
```
ktsamis, line 36: This runs for every node, not only the server, so every environment now pays two extra SSH commands per node at init, RKE2 or not.

For a non-server node the result only picks the `@local_os_version` branch below, which produces the same values `get_os_version` would, since `@has_mgrctl` is nil there. Could the probe move inside `if @host == 'server'` and default to false elsewhere?

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
ktsamis: Done in 4ddbf555b98. Also indented the `## Install Traefik` and `## Set local-path-provisioner` comments to match `## Install helm`.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
ktsamis: Done in 4ddbf555b98.

testsuite/features/support/commonlib.rb
ktsamis: Done in 4ddbf555b98, `@return [String, nil]`.

testsuite/run_sets/install_mlm_on_rke2.yml
ktsamis: Done in 4ddbf555b98, and I took the two on master with it.

Every other run set labels the section after itself and closes it with a matching END. `kubernetes_sanity_checks.yml` and `kubernetes_tests.yml` carried `secondary.yml`'s label with no END at all, from 65e3c3cf, and the two new files inherited it. All four now read `## <run set> BEGIN ##` / `## <run set> END ##`. Comments only, the Rakefile just `YAML.safe_load`s these.

testsuite/features/support/remote_node.rb
Pablogoliva: Ok, I accept this one.

testsuite/features/kubernetes/srv_install_mlm_dependencies_on_rke2.feature
Pablogoliva: It will fail later. There're some checks of it later. But it's ok, we can add it.

Pablogoliva: Also to the installation of RKE2 by itself.

ktsamis:
superseded by https://github.com/uyuni-project/uyuni/pull/12529

PR #12226, opened 2026-07-09
REVERTED - use temp table for hidden packages in update_needed_cache (bsc#1267912)

mbussolotto:
## What does this PR change?

improve performance using temp table.

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed
- [x] **DONE**

## Test coverage
- No tests
- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30908
5.1: https://github.com/SUSE/spacewalk/pull/31087

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/schema/spacewalk/common/views/suseServerAppStreamHiddenPackagesView.sql
+++ b/schema/spacewalk/common/views/suseServerAppStreamHiddenPackagesView.sql
@@ -12,6 +12,8 @@
 -- granted to use or replicate Red Hat trademarks that are incorporated
 -- in this software or its documentation.
 --
+-- Consider converting it to a MATERIALIZED VIEW with an index on
+-- (sid, pid). see bsc#1267912
 CREATE OR REPLACE VIEW suseServerAppStreamHiddenPackagesView AS
 
 -- If a package is part of any appstream,
--- a/schema/spacewalk/upgrade/susemanager-schema-5.2.13-to-susemanager-schema-5.2.14/300-update-needed-cache-performance.sql
+++ b/schema/spacewalk/upgrade/susemanager-schema-5.2.13-to-susemanager-schema-5.2.14/300-update-needed-cache-performance.sql
@@ -1,0 +1,58 @@
+--
+-- Copyright (c) 2026 SUSE LLC
+--
+-- This software is licensed to you under the GNU General Public License,
+-- version 2 (GPLv2). There is NO WARRANTY for this software, express or
+-- implied, including the implied warranties of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+-- along with this software; if not, see
+-- http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+--
+
+-- Replace the MATERIALIZED CTE in update_needed_cache with a temp table
+-- to give the planner accurate row estimates and enable a fast indexed anti-join.
+-- See bsc#1267912 for details.
+create or replace function rhn_server.update_needed_cache(
+    server_id_in in numeric
+) returns void as $$
+    begin
+      delete from rhnServerNeededCache
+        where server_id = server_id_in;
+      create temp table if not exists tmp_hidden_packages on commit drop as
+        select pid from suseServerAppStreamHiddenPackagesView where sid = server_id_in
+        with no data;
+      insert into tmp_hidden_packages
+        select pid from suseServerAppStreamHiddenPackagesView where sid = server_id_in;
+      analyze tmp_hidden_packages;
+      create index on tmp_hidden_packages(pid);
+
+      insert into rhnServerNeededCache
+             (server_id, errata_id, package_id, channel_id)
+      select distinct sp.server_id, x.errata_id, p.id, x.channel_id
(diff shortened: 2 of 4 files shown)


schema/spacewalk/upgrade/susemanager-schema-5.2.13-to-susemanager-schema-5.2.14/300-update-needed-cache-performance.sql
```diff
@@ -0,0 +1,58 @@
+--
+-- Copyright (c) 2025 SUSE LLC
```
aaannz:
```suggestion
-- Copyright (c) 2026 SUSE LLC
```

rjmateus:
@mbussolotto do you have data about how much this query gets improved?

mbussolotto:
@rjmateus , not exactly...but we had an example where the previous code took ~25 min, and after this change a reasonable time

rjmateus:
What is a reasonable time? Would be great to have a comparison before and after this change.
Maybe run an explain query on the BV server could already give some information. What do you think?

mbussolotto:
@rjmateus same query that tooks 25 minutes, now it's taking 25 milliseconds. The materialized view has a bad estimation, it chose a nested loop anti-join. Instead the temp table provides correct stats, so the planner first filters the row than runs the anti-join.

cbosdo (changes requested):
I have a hard time understanding what really changes and its implications. May be my SQL / postgresql knowledge is not good enough. 

I any way, for such a kind of change, I would expect a commit message with details about the reasoning behind it and the performance metrics you get. This will save future devs when they'll stumble on that commit to debug it.

rjmateus:
@mbussolotto can you provide a bit more information in the commit message and past the explain of the before and after queries? 
This change is really cool, and the performance improvement is HUGE. If we have a bit more information in the PR description, it may be important to others in the future in case someone needs to work on query improvements :)

cbosdo:
> @mbussolotto can you provide a bit more information in the commit message and past the explain of the before and after queries? This change is really cool, and the performance improvement is HUGE. If we have a bit more information in the PR description, it may be important to others in the future in case someone needs to work on query improvements :)

Remember that PR descriptions could go away at any time (GH is just a provider)… but commit messages are to stay for ever

aaannz (commented):
This looks neat. First I was not aware you could do `with materialized` and had to read it up.

So seems to me even original author assumed she hiddenpackagesview is not very optimized as materialized force this to be computed only once separately instead of inlining the query what postgres would do in this instance.

I wonder if this can be tackled by optimizing that view instead. Maybe we can switch that view to materiazalized view, those can also be indexed. Drawback we will need to hunt down when that view is refreshed.

mbussolotto:
> This looks neat. First I was not aware you could do `with materialized` and had to read it up.
> 
> So seems to me even original author assumed she hiddenpackagesview is not very optimized as materialized force this to be computed only once separately instead of inlining the query what postgres would do in this instance.
> 
> I wonder if this can be tackled by optimizing that view instead. Maybe we can switch that view to materiazalized view, those can also be indexed. Drawback we will need to hunt down when that view is refreshed.

TBH I don't see the advantage of a materialized view. The temp table approach already cuts execution from 25 min to 24ms (99.9% reduction), without touching other codebase.

rjmateus:
> TBH I don't see the advantage of a materialized view. The temp table approach already cuts execution from 25 min to 24ms (99.9% reduction), without touching other codebase.

I agree. With an optimization of this scale, I don't think we need to make more changes for now to fix this performance issue.
The question can be if any other queries are using this view in an inefficient way, and in that case, the change suggested by Ondrej would make a considerable difference and improve the overall situation.

aaannz:
That was my point. This view is used in 5 another queries. Converting it to materialized view + index would help all of them.
Drawback is identifying where and when to refresh it.

Overall I am not against this change, or course.

mbussolotto:
I tried to replicate the issue on my env and test the other queries, but I didn't have any performance issue so far. Since this would require some code changes (add all the `refresh` might be complicated), I'd say to leave it as it is. But I'll add a comment to `suseServerAppStreamHiddenPackagesView`, so next time will see a performance issue we know where to start

schema/spacewalk/common/views/suseServerAppStreamHiddenPackagesView.sql
```diff
@@ -12,6 +12,8 @@
 -- granted to use or replicate Red Hat trademarks that are incorporated
 -- in this software or its documentation.
 --
+-- TODO: Consider converting it to a MATERIALIZED VIEW with an index on
```
cbosdo: Adding TODOs is bad: we will loose track of it and it will never be implemented. Better create a GH issue for this.

cbosdo (changes requested):
Please do not add TODOs in the code… this is not a sustainable task management with a project of that scale.

Also add `(Common Table Expression)` next to the first `CTE` in the commit description. This would be helpful for the pg-noobs like me… I had to search it up on the web!

cbosdo (approved):

mbussolotto:
@antogaro @uyuni-project/release-engineering can we please merge it?

PR #12229, opened 2026-07-10
Removal of interface MaintenanceWindowsAware

CDellaGiusta:
## What does this PR change?

This PR removes interface `MaintenanceWindowsAware`. 
That interface was aimed at overriding the method `populateMaintenanceWindows`, which in turn was checking if a particular action type (different for each class) was in "MaintenanceModeOnly" state.

It turns out that this method was directly called within the classes. So the strategy was:

1. Moving the method `populateMaintenanceWindows` to the utility class `MaintenanceWindowHelper`, and adding the action type directly as parameter, instead of calling an overriding method `referenceMaintenanceWindowsType` on each class
2. Calling `MaintenanceWindowHelper.populateMaintenanceWindows` directly on every class, where the previous `populateMaintenanceWindows` from the interface MaintenanceWindowsAware was called, specifying directly the reference action type  as a parameter
3. Tweaking two cases of RhnAction hierarchies where the reference action type was different in the child classes
4. Getting rid of the `MaintenanceWindowsAware` straight away

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- No tests: already covered
- [x] **DONE**

## Links

Issue(s): #
Port(s): # 
5.2: https://github.com/SUSE/spacewalk/pull/31248
5.1: not backported
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


mackdk (approved):
LGTM. I would personally drop the `reference` prefix in the naming, seems redundant to me, but I don't have a strong preference.

PR #12240, opened 2026-07-13
Initialize the Cobbler API client lazily in cobbler_steps.rb

AkashKumar7902:
## What does this PR change?

`testsuite/features/step_definitions/cobbler_steps.rb` used to create its Cobbler API client at file load time:

```ruby
$cobbler_test = CobblerTest.new
```

Cucumber loads every step-definition file for every run set. Since the `CobblerTest` constructor connects to the Cobbler XML-RPC API immediately, an unrelated run set could fail before its first scenario even when it never uses Cobbler.

This replaces the eager global object with a lazy proxy. The real `CobblerTest` is instantiated on the first valid Cobbler method call, and existing step definitions keep using `$cobbler_test`.

The proxy advertises and forwards only public `CobblerTest` methods. Unknown, private, typo, and implicit-conversion calls follow normal Ruby `NoMethodError` behavior without constructing the network client.

Behavior notes:

- Cobbler scenarios still create the same client on first use.
- Run sets that never use Cobbler no longer depend on Cobbler during step-definition loading.
- A real Cobbler connectivity problem still fails the first Cobbler step; it is not hidden.
- The current chart already exposes `/cobbler_api` through `uyuni-ingress-nossl`. The original benchmark run did not preserve enough pod, endpoint, ingress, and request diagnostics to prove whether its bootstrap failure was a readiness race or another transient condition, so this PR does not claim to fix that deployment-side cause.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- No documentation needed: this changes internal testsuite initialization only.

- [x] **DONE**

## Test coverage

The updated branch was checked with:

- Ruby 3.3 syntax validation;
- targeted RuboCop;
- a focused behavior harness confirming that introspection and unknown/implicit-conversion calls do not initialize the client, valid calls initialize once and forward arguments/blocks, and unknown calls raise `NoMethodError`;
- `git diff --check`.

Cobbler scenarios continue to exercise the real client through the proxy.

- [x] **DONE**

## Links

Issue(s): none  
Port(s): none

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/step_definitions/cobbler_steps.rb
+++ b/testsuite/features/step_definitions/cobbler_steps.rb
@@ -3,7 +3,27 @@
 
 ### This file contains the definitions for all steps concerning Cobbler.
 
-$cobbler_test = CobblerTest.new
+# Lazily initialize the Cobbler client: not every run set exercises Cobbler,
+# and the Cobbler API may not be reachable at step-definition load time.
+class LazyCobblerTest
+  # Instantiates the real CobblerTest on first use and forwards the call to it.
+  #
+  # @param name [Symbol] The name of the method being called.
+  # @return [Object] The result of the forwarded method call.
+  def method_missing(name, *, &)
+    @cobbler_test ||= CobblerTest.new
+    @cobbler_test.public_send(name, *, &)
+  end
+
+  # Declares that every method is handled through method_missing.
+  #
+  # @param _name [Symbol] The name of the method being looked up.
+  # @return [Boolean] Always true.
+  def respond_to_missing?(_name, _include_private = false)
+    true
+  end
+end
+$cobbler_test = LazyCobblerTest.new
 
 # cobbler daemon
 Given(/^cobblerd is running$/) do
--- a/java/core/src/main/java/com/suse/manager/webui/controllers/DownloadController.java
+++ b/java/core/src/main/java/com/suse/manager/webui/controllers/DownloadController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015--2024 SUSE LLC
+ * Copyright (c) 2015--2026 SUSE LLC
  *
  * This software is licensed to you under the GNU General Public License,
  * version 2 (GPLv2). There is NO WARRANTY for this software, express or
@@ -28,11 +28,12 @@ import com.redhat.rhn.domain.channel.Comps;
 import com.redhat.rhn.domain.channel.MediaProducts;
 import com.redhat.rhn.domain.channel.Modules;
 import com.redhat.rhn.domain.rhnpackage.Package;
-import com.redhat.rhn.domain.rhnpackage.PackageEvr;
 import com.redhat.rhn.domain.rhnpackage.PackageFactory;
 import com.redhat.rhn.domain.server.MinionServer;
 
 import com.suse.cloud.CloudPaygManager;
+import com.suse.manager.webui.controllers.download.PackageInfo;
+import com.suse.manager.webui.controllers.download.PackagePathParser;
 import com.suse.manager.webui.utils.token.Token;
 import com.suse.manager.webui.utils.token.TokenParser;
 import com.suse.manager.webui.utils.token.TokenParsingException;
@@ -51,7 +52,6 @@ import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Arrays;
 import java.util.Base64;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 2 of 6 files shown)


Bischoff (dismissed):

Pablogoliva (dismissed):

cbosdo:
> Found while running the (opt-in) RKE2 storage benchmark run set against a server deployed on an external Kubernetes cluster, where the Cobbler API is not reachable at the address the suite derives during bootstrap: the run aborted during step-definition loading, before any scenario started. With this change the same run set loads, runs, and passes end to end (verified on a live deployment, `spacewalk-repo-sync` of 3027 packages).

While this seems a nice change, I don't understand why it was failing in your case… Was the server pod not ready? That would be the only reason for `/cobbler_api` to not be available: there is a route for it in `uyuni-ingress-nossl`.

testsuite/features/step_definitions/cobbler_steps.rb
```diff
@@ -3,7 +3,27 @@
...
+    @cobbler_test.public_send(name, *, &)
+  end
+
+  # Declares that every method is handled through method_missing.
+  #
+  # @param _name [Symbol] The name of the method being looked up.
+  # @return [Boolean] Always true.
+  def respond_to_missing?(_name, _include_private = false)
+    true
+  end
```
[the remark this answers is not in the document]
AkashKumar7902: Addressed in `a9a1d9d82f`. `method_missing` now checks `CobblerTest.public_method_defined?` before constructing the client and delegates unknown calls to `super`; `respond_to_missing?` mirrors the same check. I verified that introspection and implicit-conversion calls do not initialize the client, unknown calls raise `NoMethodError`, and a valid call initializes once and forwards normally.

AkashKumar7902:
@cbosdo @Pablogoliva @Bischoff You are right that the chart already exposes `/cobbler_api`: `uyuni-ingress-nossl` routes it to the `web` service. I did not retain pod, endpoint, ingress, or direct-request diagnostics from the original run, so I cannot prove whether it was a server-readiness race or another transient bootstrap condition.

The server readiness probe currently checks `/rhn/manager/api/api/getVersion`, not the Cobbler XML-RPC call itself. This change does not hide a Cobbler failure from Cobbler scenarios: their first valid call still constructs `CobblerTest` and performs the API operation, failing there if Cobbler is unavailable. It only prevents a run set with no Cobbler scenarios from making that network call while Cucumber loads every step-definition file.

I also rebased onto current `master` and tightened the proxy in `a9a1d9d82f` so unknown and implicit-conversion calls cannot initialize the client. If the bootstrap failure reproduces, I will capture pod readiness, `web` endpoints, ingress state, DNS, and a direct `/cobbler_api` request to diagnose the deployment-side cause separately.

cbosdo:
Let's close it for now. Feel free to reopen if the problem happens again. But first we need to get data to analyze to see what really happened

ktsamis:
Reopening: we now have the data @cbosdo asked for.

`manager-Head-dev-acceptance-tests-podman` build 9442 failed with `SSL_read: unexpected eof` and `ECONNRESET` across several workers. `/var/log/uyuni-tools.log` on the server shows why:

```
10:59:59  mgrctl cp /tmp/zz-cobblerapi.conf server:/etc/apache2/conf.d/zz-cobblerapi.conf   x3
10:59:59  mgrctl exec -i systemctl restart apache2
11:00:00  mgrctl cp ...                                                                     x2
11:00:00  mgrctl exec -i systemctl restart apache2                                          x3
11:00:01  mgrctl exec -i systemctl restart apache2
```

Five copies and five restarts, one per parallel worker, because `$cobbler_test = CobblerTest.new` at line 6 runs on every worker and `CobblerTest#initialize` unconditionally injects the config and restarts Apache. `apache2.service` is `PartOf` `spacewalk.target`, so systemd turned them into three full stop/start cycles of the whole stack and killed the requests the other workers had in flight. The restart count per stage matches the worker count exactly: 1 for the single-process stages, 5 for the 5-worker one.

One caveat on sequencing. Lazy init alone does not fix this, and in one respect makes it worse: it moves the remaining restart out of load time and into a running scenario, where the other workers are mid-request. So:

1. https://github.com/uyuni-project/uyuni/pull/12508 makes the injection and the restart conditional on the config not already being there, which is what removes the damage.
2. https://github.com/uyuni-project/uyuni/pull/12507 retries API calls that fail with a transport-level error.
3. This PR on top, as the cleanup it always was: no reason to build the client on workers that never run a cobbler scenario.

ktsamis (commented):
### Still open

- [ ] **The PR does two unrelated things.** The description only covers the lazy Cobbler client. But uyuni-project/uyuni@e38636f1eba7a2e043f4d747e7f1fa079b85a7ce ("Support external Kubernetes test controllers") adds a whole new `KUBERNETES_CLUSTER_MODE=external` mode across [`commonlib.rb`](https://github.com/uyuni-project/uyuni/blob/8f4cc6c17fa41ae8570c9b635b1b8298118c0a81/testsuite/features/support/commonlib.rb#L13-L16) and [`env.rb`](https://github.com/uyuni-project/uyuni/blob/8f4cc6c17fa41ae8570c9b635b1b8298118c0a81/testsuite/features/support/env.rb#L90). That is 15 of the 39 added lines and it is not mentioned anywhere. Please either split it into its own PR, or update the description. The line "No documentation needed: this changes internal testsuite initialization only" is no longer true either.

- [ ] **Nothing sets `KUBERNETES_CLUSTER_MODE`.** I grepped uyuni, https://github.com/SUSE/susemanager-ci and https://github.com/uyuni-project/sumaform. The only two hits are the ones this PR adds. No pipeline sets the variable and there is no doc under `testsuite/documentation/`. So the new branches are unreachable today. Say in the description which pipeline will set it, and document the contract it needs: `KUBERNETES_CLUSTER_MODE`, `PRODUCT` and `SERVER`.

### Done

- [x] **`method_missing` dropped keyword arguments.** It forwarded positional arguments and the block, but not keywords, so the next `CobblerTest` method to take a keyword would have failed with `ArgumentError: wrong number of arguments` in a confusing place. Fixed in uyuni-project/uyuni@8f4cc6c17fa41ae8570c9b635b1b8298118c0a81: [`cobbler_steps.rb:13`](https://github.com/uyuni-project/uyuni/blob/8f4cc6c17fa41ae8570c9b635b1b8298118c0a81/testsuite/features/step_definitions/cobbler_steps.rb#L13) is now `def method_missing(name, ...)` forwarding through `public_send(name, ...)`. I ran the proxy standalone on Ruby 3.4.3: positional, keyword and block calls all forward; unknown methods still raise `NoMethodError` without constructing the client; `respond_to?` still answers from `CobblerTest.public_method_defined?`. `...` with a leading parameter needs Ruby 3.0+ and `testsuite/.rubocop.yml` sets `TargetRubyVersion: 3.3`, so that is fine.

- [x] **Dismissed approvals re-requested.** The force pushes dismissed the approvals from @Bischoff and @Pablogoliva, both of whom reviewed the Cobbler change only. I re-requested them.

- [x] **The rest of the proxy checks out.** I checked all 21 public `CobblerTest` methods. None collides with an `Object` instance method, so every call the step definitions make reaches `method_missing`. And `$cobbler_test` in [`cobbler_steps.rb:29`](https://github.com/uyuni-project/uyuni/blob/8f4cc6c17fa41ae8570c9b635b1b8298118c0a81/testsuite/features/step_definitions/cobbler_steps.rb#L29) plus `$is_transactional_server` in [`env.rb:90`](https://github.com/uyuni-project/uyuni/blob/8f4cc6c17fa41ae8570c9b635b1b8298118c0a81/testsuite/features/support/env.rb#L90) are the only two load-time touches of the server in the whole testsuite, so the pair of changes is complete for what it sets out to do. RuboCop is clean on all three files.

testsuite/features/support/env.rb
```diff
@@ -87,7 +87,7 @@
 $is_gh_validation = ENV['PROVIDER'].include? 'podman'
 $is_containerized_server = %w[k3s podman].include? ENV.fetch('CONTAINER_RUNTIME', '')
 $is_rke2 = ENV.fetch('CONTAINER_RUNTIME', '').include? 'rke2'
-$is_transactional_server = transactional_system?('server', runs_in_container: false)
+$is_transactional_server = !external_kubernetes_cluster? && transactional_system?('server', runs_in_container: false)
```
Pablogoliva: Why are you modifying the is transactional_server?. What's the matter of it?

testsuite/features/step_definitions/cobbler_steps.rb
```diff
@@ -3,7 +3,30 @@
...
+  # Declares which CobblerTest methods are handled through method_missing.
+  #
+  # @param name [Symbol] The name of the method being looked up.
+  # @param include_private [Boolean] Whether private methods should be considered.
+  # @return [Boolean] Whether the public method is available on CobblerTest.
+  def respond_to_missing?(name, include_private = false)
+    CobblerTest.public_method_defined?(name) || super
+  end
+end
+$cobbler_test = LazyCobblerTest.new
```
Pablogoliva: Pls leave it simple. We have been discussing about it, and we prefer that previous way at least that's 100% necessary.

testsuite/features/support/env.rb
AkashKumar7902: this was added because an external cluster controller might not have SSH access to
  a sumaform managed server host, so calling transactional_system? could fail during
  testsuite startup.

i have removed it for now

testsuite/features/step_definitions/cobbler_steps.rb
AkashKumar7902: ok, i saw you committed changes on aug 31, closing this pr for now

PR #12245, opened 2026-07-13
moving ServerAction related methods from ActionFactory to ServerActionFactory

CDellaGiusta:
## What does this PR change?

This PR moves ServerAction related methods from ActionFactory to ServerActionFactory, with tests.
During this process, several methods are simplified into one base method call, namely:

- 6 flavours of method listServerActionsForServer 
- 3 flavours of method rescheduleAllServerActions
- 2 flavours of method updateServerActions
- 3 flavours of method addServerToAction

## Codespace

## GUI diff
No difference.
- [x] **DONE**

## Documentation
- [x] **DONE**

## Test coverage
- Unit tests were added
- [x] **DONE**

## Links
Issue(s): #
Port(s): # 
5.2: https://github.com/SUSE/spacewalk/pull/31251
5.1: not backported
- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test


java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
```diff
@@ -307,4 +319,192 @@ private static void addServerToActionPrivate(Server serverIn, Action parentActio
         parentActionIn.addServerAction(sa);
     }
 
+
+    /**
+     * Check to see if a server has a pending kickstart scheduled
+     * @param serverId server
+     * @return true if found, otherwise false
+     */
+    @SuppressWarnings("unchecked")
```
mackdk, line 328: I don't think this `SuppressWarnings` is needed

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
```diff
@@ -307,4 +319,192 @@ private static void addServerToActionPrivate(Server serverIn, Action parentActio
...
+        return findFirstPendingActionForServerAndActionType(serverId, "kickstart.initiate").isPresent();
+    }
+
+    /**
+     * Check if there is a pending product migration in the schedule. Return the
+     * action ID if available or null otherwise.
+     * @param serverId server
+     * @return ID of a possibly scheduled migration or null.
+     */
+    @SuppressWarnings("unchecked")
```
mackdk, line 339: Same here, you can remove this

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
```diff
@@ -307,4 +319,192 @@ private static void addServerToActionPrivate(Server serverIn, Action parentActio
...
+        return findFirstPendingActionForServerAndActionType(serverId, "distupgrade.upgrade").orElse(null);
+    }
+
+    /**
+     * Check if the server has a scheudled reboot action. Return the action
+     * if available or null otherwise
+     * @param serverId server
+     * @return reboot Action or null otherwise
+     */
+    @SuppressWarnings("unchecked")
```
mackdk, line 350: Same here

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
```diff
@@ -307,4 +319,192 @@ private static void addServerToActionPrivate(Server serverIn, Action parentActio
...
+        return session.createQuery("""
+                        FROM ServerAction WHERE
+                            server.id = :serverId
+                            AND parentAction.actionType.label = :label
+                            AND status.id in ( 0, 1 )""", ServerAction.class)
+                .setParameter("serverId", serverId)
+                .setParameter("label", actionTypeLabel)
+                .list().stream()
+                .findFirst()
+                .map(ServerAction::getParentAction);
```
mackdk, line 367: You can remove the `.list()` as Hibernate offers directly a `stream()` method (avoiding to traverse all the results) and you can invert `findFirst()` and `map()` for simplier code:
```java
                .stream()
                .map(ServerAction::getParentAction)
                .findFirst();
```
Or we could even change the query to `SELECT parentAction FROM ...` and avoid the mapping entirely.

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
```diff
@@ -307,4 +319,192 @@ private static void addServerToActionPrivate(Server serverIn, Action parentActio
...
+        return getSession()
+                .createQuery("""
+                        SELECT COUNT(sa.server.id)
+                        FROM ServerAction sa
+                        WHERE sa.parentAction.id = :aid AND sa.status.id = :stid""", Long.class)
+                .setParameter("aid", action.getId())
+                .setParameter("stid", status.getId())
+                .uniqueResult()
+                .intValue();
+    }
```
mackdk, line 416: This method is kind of weird. From what I see it's used only in:
```
    // Get the number of ServerAction objects that match
    // the passed in ActionStatus
    private long getActionStatusCount(ActionStatus status) {
        return ActionFactory.getServerActionCountByStatus(this, status);
    }
```
where the return value is used as `long` rendering the `intValue()` conversion kind of useless. I would it and just return a primitive `long` (as `SELECT COUNT(...)` should never return `null`).

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
```diff
@@ -0,0 +1,510 @@
...
+     * @param status    {@link ActionStatus} object that needs to be set
+     */
+    public static void updateServerActions(Action actionIn, List<Long> serverIds, ActionStatus status) {
+        LOG.debug("Action status {} is going to be set for these servers: {}", status.getName(), serverIds);
+
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("action_id", actionIn.getId());
+        parameters.put("status", status.getId());
+        parameters.put("completed", ActionFactory.STATUS_COMPLETED.getId());
+        parameters.put("failed", ActionFactory.STATUS_FAILED.getId());
```
mackdk, line 267: Since you don't need to modify the `parameters` map, you can reduce the boilerplate:
```java
        Map<String, Object> parameters = Map.of(
            "action_id", actionIn.getId(),
            "status", status.getId(),
            "completed", ActionFactory.STATUS_COMPLETED.getId(),
            "failed", ActionFactory.STATUS_FAILED.getId()
        );
```

mackdk (approved):

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+    // history of the following two constants:
+    // REMAINING_TRIES = 10L was extracted from the ActionManager.scheduleAction method, then moved to
+    // ActionFactory.createAddServerAction and then to ServerActionFactory
+    // There was a comment "hmm 10?". Not sure what the hesitation is, but that comment was retained.
+    // REMAINING_TRIES_FROM_PERL = 5L is commonly seen around with a comment like "arbitrary number from perl"
+    // There is a chance both values could be merged into one, but no-one knows
+    public static final Long REMAINING_TRIES = 10L;
+    public static final Long REMAINING_TRIES_FROM_PERL = 5L;
```
mcalmer, line 64: We removed all perl code a long time ago. I would suggest to remove the _FROM_PERL and use only the generic REMAINING_TRIES

CDellaGiusta: As you can see in my comment above the constants, I simply grouped the two values used around the code in two constants. I was assuming the two values should be set differently where they appear. If from the historical point of view you are sure I can safely substitute 5 tries (value of REMAINING_TRIES_FROM_PERL) with 10 tries (value of REMAINING_TRIES), I will go for it and remove the "_FROM_PERL" variable, as well as updating the comment

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
CDellaGiusta: right, this is a leftover of the method that has be moved. Removed

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
CDellaGiusta: ditto, removed

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
CDellaGiusta: ditto, removed

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
CDellaGiusta: suggestion taken

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
CDellaGiusta: also this is something I just moved. But your suggestion makes sense, so I removed the intValue() and returned a `long`

java/core/src/main/java/com/redhat/rhn/domain/action/server/ServerActionFactory.java
CDellaGiusta: ditto, suggestion implemented

PR #12247, opened 2026-07-13
Testsuite: added a stopper text for check_text? and according minionssh test adjustment

alichiel:
## What does this PR change?

Adding a stopper text to the check_test? function that ends the function immediatelly. This will cut the unnecessary sleep time in tests. 

minssh_tunnel.feature has been updated accordingly - the tests fail immediatelly when the specific situation accurs - for example bootstrapping already bootstrapped system.

## Testing 

- tested with Multi-Linux Manager 5.1.4 with SLES SSH minion

## Additional info 

Please note that the existing test minssh_tunnel.feature is not fixed by this PR, only the execution time is being cut significantly.

### DIFF
--- a/testsuite/features/step_definitions/navigation_steps.rb
+++ b/testsuite/features/step_definitions/navigation_steps.rb
@@ -48,6 +48,10 @@ When(/^I wait at most (\d+) seconds until I see "([^"]*)" text$/) do |seconds, t
   raise ScriptError, "Text '#{text}' not found" unless check_text?(text, timeout: seconds.to_i)
 end
 
+When(/^I wait at most (\d+) seconds until I see "([^"]*)" but "([^"]*)" text$/) do |seconds, text, stopper|
+  raise ScriptError, "Text '#{text}' not found or '#{stopper}' appeared" unless check_text?(text, stopper: stopper, timeout: seconds.to_i)
+end
+
 When(/^I wait until I see "([^"]*)" text or "([^"]*)" text(?:, (refreshing the page))?$/) do |text1, text2, refresh_option|
   if refresh_option
     # refreshing the page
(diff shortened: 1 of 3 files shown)


Bischoff (approved):

testsuite/features/step_definitions/navigation_steps.rb
```diff
@@ -48,6 +48,10 @@
   raise ScriptError, "Text '#{text}' not found" unless check_text?(text, timeout: seconds.to_i)
 end
 
+When(/^I wait at most (\d+) seconds until I see "([^"]*)" but "([^"]*)" text$/) do |seconds, text, stopper|
```
ktsamis, line 51: No argument for implementation or the need for it, just a nitpick for grammar. I would use "unless" instead of "but" here. Or something like that. I don't know why but "but" seems wrong here for an exception.

alichiel: It should be correct. This sentence exactly mean: "I want to see A but not B". "but not" is simply shortened to "but".
https://dictionary.cambridge.org/grammar/british-grammar/but

ktsamis (approved):

NamelessOne91 (approved):

antogaro (approved):
lgtm

testsuite/features/step_definitions/navigation_steps.rb
maximenoel8: The wording is very confusing and your link doesn't work in this case.
For example, "I want tea, but coffee" doesn't mean "I want tea, but not coffee". It only  works after a negative or universal/restrictive word like 
"Nobody but you would say that"
"Everywhere but here."

Please change for `I see "A" but not "B" text` or even better `I see "A" text but do not see "B" text`

PR #12281, opened 2026-07-20
Testsuite: proxy port listening check update that changed with podman 6.0, also compatible backwards

alichiel:
With change to podman 6.0, the ports check of proxy doesn't work anymore. The entire routing withing the Linux kernel is handled by netavark via nftables. This change will the test issue and it's compatible backwards.


ktsamis (approved):

antogaro (approved):
lgtm

PR #12293, opened 2026-07-21
feat(openapi): migrate admin.configuration and admin.payg handlers to OpenAPI contract

masterhj:
Migrates the `admin.configuration` handler (`configure`) and the `admin.payg` handler
(`create`, `setDetails`, `list`, `getDetails`, `delete`) to the OpenAPI contract approach.
Two handlers in one PR as discussed.

## Note :-

These two handlers need parameter-side documentation support that the parsers did not have
yet, so this PR adds it. Both additions are opt-in behind a new annotation ,a request
parameter that does not carry it goes down exactly the same code path as before.

The first addition is `@LegacyDocParam(type = ...)`, the parameter-side counterpart to the
existing `@LegacyDocResponse`. `admin.configuration.configure` takes a `map` in the legacy
docs, but OpenAPI has no `map` primitive and swagger-core normalises it to `object`, so the
type has to be carried separately. The annotation is read in `UyuniSwaggerReader` and emitted
as `x-uyuni-doc-param-type`, alongside the `x-uyuni-doc-response-*` extensions already in use,
and both parsers pick it up when rendering the parameter list.

The reason this is applied in the reader rather than declared directly on the getter is worth
spelling out, because it looks like avoidable indirection otherwise. Swagger's
`@Schema(extensions = ...)` does propagate to the property schema and works fine for
`admin.configuration`, whose `content` property is an inline object. It does not work for
`admin.payg.setDetails`, whose `details` property is a `$ref` to `PaygDetails`: OpenAPI 3.0
collapses a `$ref` property to a bare `{"$ref": ...}` and drops the sibling annotations, so the
extension never reaches the spec. Re-applying it in the reader after `resolveAndRegisterSchema`
is what makes it survive for both shapes.

The second addition is the expansion of a nested struct parameter into its properties, which
mirrors the array-of-struct expansion already implemented for return values, and reuses
`printArrayStructProperties`. `admin.payg.setDetails` takes a `PaygDetails` struct with twelve
scalar fields, and the legacy doclet lists all of them under the parameter. This is not a
change to existing rendering: a `$ref` parameter was not handled at all before, because
`displayType()` returns `schema.getType()`, which is null for a `$ref`, so the parameter came
out as a literal `[.null]#null#` line in the AsciiDoc. The expansion fills that gap rather than
altering behaviour that previously worked. Since it is gated on the annotation, an unannotated
`$ref` parameter still renders the old way, if you would rather have that fixed
unconditionally for every handler, I am happy to change it, it just seemed safer not to touch
the default path in this PR.

One smaller parser change comes with it. `@LegacyDocResponse(name = ...)` already labelled the
array-item struct in AsciiDoc but not in DocBook, so `admin.payg.list` and `getDetails` picked
up the schema-derived `payg ssh data` instead of the serializer's `SSH data`. The label is now
passed down through `renderReturnSchema` to `renderStructList` in the same way `typeOverride`
already was. It has to be the full label rather than just `name()`, otherwise `api`'s
`method_info - Available...` loses its description, which the before/after comparison caught.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Unit tests were added: `AdminConfigurationHandlerContractTest` and
  `AdminPaygHandlerContractTest` validate the generated OpenAPI schema against the handler
  responses. The `admin.payg` test covers all five methods, including the 13-field `create`
  body and the nested `details` struct on `setDetails`.

- [x] **DONE**

## Changelogs

- [x] No changelog needed

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/admin/configuration/AdminConfigurationHandlerApi.java
+++ b/java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/admin/configuration/AdminConfigurationHandlerApi.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2026 SUSE LLC
+ *
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ */
+package com.redhat.rhn.frontend.xmlrpc.admin.configuration;
+
+import com.redhat.rhn.domain.user.User;
+
+import com.suse.manager.api.docs.ApiEndpointDoc;
+import com.suse.manager.api.docs.LegacyDocParam;
+
+import java.util.Map;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.tags.Tag;
+
+/**
+ * API contract for {@link AdminConfigurationHandler}.
+ */
+@Tag(name = "admin.configuration", description = "Provides methods to configure the Uyuni server.")
+public interface AdminConfigurationHandlerApi {
+
+    /**
+     * Configure server.
+     *
+     * @param loggedInUser the current user
+     * @param content the Uyuni configuration formula data
+     * @return 1 on success
+     */
+    @ApiEndpointDoc(
+        summary = "Configure server.",
+        requestClass = ConfigureRequest.class,
+        isIntegerResponse = true
+    )
+    int configure(User loggedInUser, Map<String, Object> content);
+
+    @Schema(name = "ConfigureServerRequest")
+    interface ConfigureRequest {
+
+        /**
+         * @return the Uyuni configuration formula data
+         */
+        @LegacyDocParam(type = "map")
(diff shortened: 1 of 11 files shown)


java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/admin/configuration/AdminConfigurationHandlerApi.java
```diff
@@ -0,0 +1,54 @@
...
+    @Schema(name = "ConfigureServerRequest")
+    interface ConfigureRequest {
+
+        /**
+         * @return the Uyuni configuration formula data
+         */
+        @LegacyDocParam(type = "map")
+        @Schema(type = "object", description = "the Uyuni configuration formula data",
+                requiredMode = Schema.RequiredMode.REQUIRED)
+        Object getContent();
```
wweellddeerr, line 52: Please avoid using `Object` as return type. In this case you can keep the same type as in the implementation method: `Map<String, Object>`, swagger core will convert it automatically and you can even remote the `type = "object"` from the `@Schema` annotation.

java/core/src/main/java/com/suse/manager/api/docs/LegacyDocParam.java
wweellddeerr, line 1: There's already a `LegacyDocResponse` annotation that serves the same purpose and it has no `@Target` restriction, so it could be applied directly to a request getter as-is. Introducing a separate annotation duplicates the annotation class, the extension constant, and the reader/parser lookup logic in both pasers. Please consider genereralizing `LegacyDocResponse` instead of introducing a parallel annotation for the request side.

PR #12299, opened 2026-07-22
Remove salt dependency from spacewalk-proxy-salt

vzhestkov:
## What does this PR change?

Even if `salt-broker` is intended to serve Salt traffic it has no idea about Salt in genral as it operates on lower level with ZeroMQ sockets and just proxying two pairs of sockets, so technically `salt` module is not used at all and all of the dependencies it brings with `salt` are redundant.

To simplify and reduce the size of `salt-broker` container we need to drop these dependencies and use just direct dependencies to the python modules used by `salt-broker`: PyYAML and pyzmq besides of python base modules.

## Codespace

## GUI diff

No difference.

## Documentation

## Test coverage
- No tests: just reduces the size of `salt-broker` container.

## Links

Issue(s): #

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/proxy/proxy/spacewalk-proxy.spec
+++ b/proxy/proxy/spacewalk-proxy.spec
@@ -98,7 +98,9 @@ Spacewalk Proxy components.
 
 %package salt
 Summary:        A ZeroMQ Proxy for Salt Minions
-Requires(pre):  salt
+Requires:       python3-base
+Requires:       python3-PyYAML
+Requires:       python3-pyzmq
 
 %description salt
 A ZeroMQ Proxy for Salt Minions
@@ -174,6 +176,7 @@ fi
 
 %files salt
 %{_bindir}/salt-broker
+%dir %{_sysconfdir}/salt
 %config(noreplace) %{_sysconfdir}/salt/broker
 %dir %{_sysconfdir}/salt/broker.d
 
--- a/proxy/proxy/spacewalk-proxy.spec
+++ b/proxy/proxy/spacewalk-proxy.spec
@@ -34,6 +34,7 @@ License:        GPL-2.0-only
 URL:            https://github.com/uyuni-project/uyuni
 #!CreateArchive: %{name}
 Source0:        https://github.com/spacewalkproject/spacewalk/archive/%{name}-%{version}.tar.gz
+BuildRequires:  ca-certificates
 BuildRequires:  make
 BuildRequires:  python3
 BuildRequires:  spacewalk-backend >= 1.7.24
--- a/proxy/proxy/spacewalk-proxy.changes.vzhestkov.remove-salt-broker-dep
+++ b/proxy/proxy/spacewalk-proxy.changes.vzhestkov.remove-salt-broker-dep
@@ -1,0 +1,1 @@
+- Remove salt dependency from spacewalk-proxy-salt package


proxy/proxy/spacewalk-proxy.spec
```diff
@@ -98,7 +98,9 @@ Spacewalk Proxy components.
 
 %package salt
 Summary:        A ZeroMQ Proxy for Salt Minions
-Requires(pre):  salt
+Requires:       python3-base
+Requires:       python3-PyYAML
+Requires:       python3-pyzmq
```
cbosdo, line 104: Do you want to add a `Conflicts: salt` as both now own `/etc/salt`?

vzhestkov: I don't think it's really needed, in reality even if anyone will try to install them on the same system it shouldn't be a problem. They can co-exist, but just salt-broker doesn't need anything from `salt`.

vzhestkov: And if I remember right the entries from spec marked as `%dir`, could be owned by multiple packages.

cbosdo: Then I'm fine

meaksh (dismissed):
LGTM, thanks

proxy/proxy/spacewalk-proxy.spec
```diff
@@ -175,6 +177,7 @@ fi
 %files salt
 %defattr(-,root,root)
 %{_bindir}/salt-broker
+%dir %{_sysconfdir}/salt
```
mcalmer, line 180: Providing same directory by multiple packages is possible, but you need to take care that this directory has in all packages which defines it, the same permission and owner.
Please check

meaksh: With the changes in this PR, there won't be salt anymore inside the `proxy-salt-broker` container, IIUC so we should be fine here.

vzhestkov: We are using this package in the container only, and the `salt-broker` is running from https://github.com/uyuni-project/uyuni/blob/master/containers/proxy-salt-broker-image/Dockerfile#L46, @cbosdo correct me if I'm wrong but seems like it's running as `root`?
Ideally the spec should contain something like:
```
%dir %attr(0750, root, salt) %{_sysconfdir}/salt
%config(noreplace) %attr(0640, root, salt) %{_sysconfdir}/salt/broker
%dir %attr(0750, root, salt) %{_sysconfdir}/salt/broker.d
```
but in this case it will require `salt` package what we are going to avoid.

mcalmer: Can the salt group be created by a separate package like we have it for prometheus and grafana user?

cbosdo: Yes, for now the salt-broker runs as `root`.

It seems though that you miss another folder. In my test build I have
```
[   25s] spacewalk-proxy-common-5.2.4-260400.2.1.uyuni6.noarch.rpm: directories not owned by a package:
[   25s]  - /etc/pki
```

cbosdo: Adding a `BuildRequires: ca-certificates` is enough to pass the build error. I'm waiting for images builds of the salt broker and httpd to see if it's all fine there too

vzhestkov: Just checked inside the container of 5.1 and it looks strange for me:
```
# rpm -qf /etc/pki
p11-kit-0.23.22-150500.8.3.1.x86_64
ca-certificates-2+git20240416.98ae794-150300.4.3.3.noarch
spacewalk-config-5.1.9-150700.3.6.5.noarch
```
why `spacewalk-config`?

vzhestkov: Oh I see why adding this build requirement solve the issue: spacewalk-proxy-common contains:
```
%dir %{_sysconfdir}/pki/tls
%dir %{_sysconfdir}/pki/tls/certs
%dir %{_sysconfdir}/pki/tls/private
```

But I don't understand why it was made this way and why this package is creating these directories.

cbosdo: This is where the Apache certs and key are copied in the proxy-httpd container

cbosdo:
```
%dir %{_sysconfdir}/pki/tls/certs
%dir %{_sysconfdir}/pki/tls/private
```
Those can be removed: we now mount the secrets directly in apache conf folder.

cbosdo (dismissed):
Tested to work: both the salt-broker and the httpd images.

cbosdo (approved):

m-czernek (approved):

PR #12305, opened 2026-07-22
Improve healthcheck (bsc#1271124)

nadvornik:
## What does this PR change?

This PR:
- increases healthcheck start-period to 600s
- calls spacewalk-diskcheck also on service startup

This means the container has enough time to start.
If the container has been taken down because of low disk space, it won't try to start the services again.

UPDATE: I moved the startup check to the entrypoint and added similar changes for postgres container
UPDATE2: Added fix for https://bugzilla.suse.com/show_bug.cgi?id=1273144 -  Specify correct healthcheck command in Dockerfile, so it does not need to be changed on podman commandline https://github.com/uyuni-project/uyuni-tools/pull/837

This does not affect kubernetes - the healthcheck from the Dockerfile is ignored there.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: fix

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/31163 https://github.com/SUSE/spacewalk/issues/31380
https://bugzilla.suse.com/show_bug.cgi?id=1271124
Port(s): https://github.com/SUSE/spacewalk/pull/31621 https://github.com/SUSE/spacewalk/pull/31397

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/server-image/Dockerfile.uyuni
+++ b/containers/server-image/Dockerfile.uyuni
@@ -143,4 +143,4 @@ LABEL io.artifacthub.package.readme-url="%SOURCEURL_WITH(README.md)%"
 LABEL org.uyuni.version="${PRODUCT_VERSION}"
 
 CMD ["/docker-entrypoint-init"]
-HEALTHCHECK --interval=10s --start-period=180s --timeout=120s --retries=1 CMD ["/usr/bin/liveness-check.sh"]
+HEALTHCHECK --interval=10s --start-period=600s --timeout=120s --retries=1 CMD ["/usr/bin/liveness-check.sh"]
--- a/containers/server-image/Dockerfile.uyuni
+++ b/containers/server-image/Dockerfile.uyuni
@@ -143,4 +143,4 @@ LABEL io.artifacthub.package.readme-url="%SOURCEURL_WITH(README.md)%"
 LABEL org.uyuni.version="${PRODUCT_VERSION}"
 
 CMD ["/docker-entrypoint-init"]
-HEALTHCHECK --interval=10s --start-period=600s --timeout=120s --retries=1 CMD ["/usr/bin/liveness-check.sh"]
+HEALTHCHECK --interval=10s --start-period=600s --timeout=120s --retries=1 CMD ["/usr/bin/healthcheck.sh"]
--- a/python/spacewalk/satellite_tools/spacewalk-diskcheck
+++ b/python/spacewalk/satellite_tools/spacewalk-diskcheck
@@ -8,15 +8,19 @@ usage: $0 options
 
 OPTIONS:
     -h              Show this message
+    -f              Skip checking if spacewalk services are running
 EOF
 }
 
 function parseArguments {
-    while getopts "h?" opt; do
+    while getopts "h?f" opt; do
         case "$opt" in
             h|\?)
                 usage
                 exit 0
+                ;;
+            f)
+                SKIP_SERVICES_CHECK=1
                 ;;
         esac
     done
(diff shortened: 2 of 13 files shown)


aaannz:
I thought we agreed on using entrypoint init script.

nadvornik:
@aaannz I think it would restart it too fast. Or is there any way to handle it?

aaannz:
We should restart only on clean exit code as we have `Restart=on-success` from the systemd side on a host and I think when entrypoint exits with nonzero, podman will not restart automatically. I think podman will auto restart only on unhealthy state.

Also longterm plan is to drop systemd in the container completely.

containers/server-image/Dockerfile.uyuni
```diff
@@ -142,4 +142,4 @@ LABEL org.opensuse.reference="${REFERENCE_PREFIX}/server:${PRODUCT_VERSION}.%REL
 LABEL org.uyuni.version="${PRODUCT_VERSION}"
 
 CMD ["/docker-entrypoint-init"]
-HEALTHCHECK --interval=10s --start-period=180s --timeout=120s --retries=1 CMD ["/usr/bin/liveness-check.sh"]
+HEALTHCHECK --interval=10s --start-period=600s --timeout=120s --retries=1 CMD ["/usr/bin/liveness-check.sh"]
```
mbussolotto, line 145: I think in this way the container would be marked as `Starting` for 600s ...
```suggestion
HEALTHCHECK --interval=30s --start-period=180s --timeout=180s --retries=10 CMD ["/usr/bin/liveness-check.sh"]
```
in this way:
- 180 for starting
- 30*10+180

So in this way it should be marked as unhealthy after 660. The problem in this case it would not be at start time (because it would pass from `Starting` -> `Unhealthy`), but at running time, since the container would be stop after 480s instead of 120

mbussolotto:
>  I think podman will auto restart only on unhealthy state.

currently we have 
> --health-on-failure=stop

so systemd should stop it (and not restart), if unhealthy

aaannz:
> > I think podman will auto restart only on unhealthy state.
> 
> currently we have
> 
> > --health-on-failure=stop
> 
> so systemd should stop it (and not restart), if unhealthy

This is podman config, so podman is supposed to stop the container on unhealthy. Not sure what exit code is then presented to the systemd which then can do it's thing.

On that particular cloud instance this was not working as expected as it was constantly being restarted.

mbussolotto:
> On that particular cloud instance this was not working as expected as it was constantly being restarted.

Ohh I didn't notice this was a cloud instance indeed.

containers/server-image/Dockerfile.uyuni
nadvornik: As I understand it, the healthcheck is executed during start period and if it succeeds then the container is marked healthy immediately. Just failures are ignored - the container stays as starting.

python/spacewalk/satellite_tools/spacewalk-diskcheck
aaannz, line 58: I wonder why we even have this check in the diskcheck. We check this also in the liveness-check. IMO instead of adding new option and conditions, just remove this part completely.

aaannz (approved):

python/spacewalk/spacewalk-backend.changes.nadvornik.startup_healthcheck
```diff
@@ -0,0 +1 @@
+- Allow using spacewalk-diskcheck without running service
```
aaannz, line 1:
```suggestion
- Allow using spacewalk-diskcheck without running service
  (bsc#1271124)
```

aaannz (changes requested):

aaannz (dismissed):
After discussion I missed that there are other changefiles with the bug number included. So it if fine with me.

mbussolotto (dismissed):

aaannz (dismissed):

aaannz (approved):

PR #12323, opened 2026-07-23
Wait for taskomatic in SaltReactor (bsc#1265472)

nadvornik:
## What does this PR change?

Taskomatic is started as the last service, after tomcat and salt.
The SaltReactor however calls taskomatic API for scheduling actions.
There is ~15s window when SaltReactor creates pending actions in the database, but it can't notify taskomatic, so the actions stay pending forever.

This PR  delays processing of salt events until the taskomatic API is available.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation
- No documentation needed: fix

- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30757
https://bugzilla.suse.com/show_bug.cgi?id=1265472

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
+++ b/java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
@@ -24,6 +24,7 @@ import com.redhat.rhn.domain.server.MinionServer;
 import com.redhat.rhn.domain.server.MinionServerFactory;
 import com.redhat.rhn.manager.action.ActionManager;
 import com.redhat.rhn.manager.system.SystemManager;
+import com.redhat.rhn.taskomatic.TaskomaticApi;
 import com.redhat.rhn.taskomatic.TaskomaticApiException;
 
 import com.suse.cloud.CloudPaygManager;
@@ -100,6 +101,11 @@ public class SaltReactor {
     // Indicate that the reactor has been stopped
     private volatile boolean isStopped = false;
 
+    // The thread that initializes the event stream connection asynchronously
+    private Thread initializerThread;
+
+    private TaskomaticApi taskomaticApi = new TaskomaticApi();
+
     /**
      * Processing salt events
      * @param saltApiIn instance to talk to salt
@@ -148,7 +154,49 @@ public class SaltReactor {
 
         MessageQueue.publish(new RefreshGeneratedSaltFilesEventMessage());
 
-        connectToEventStream();
+        startEventStreamAsynchronously();
+    }
+
+    /**
+     * Set the taskomatic api instance.
+     * @param taskomaticApiIn the taskomatic api
+     */
+    public void setTaskomaticApi(TaskomaticApi taskomaticApiIn) {
+        this.taskomaticApi = taskomaticApiIn;
+    }
+
+    /**
+     * Start the salt reactor connection asynchronously, waiting for Taskomatic to be responsive first.
+     */
+    private void startEventStreamAsynchronously() {
+        if (initializerThread != null && initializerThread.isAlive()) {
+            return;
+        }
+        initializerThread = new Thread(() -> {
+            int retries = 0;
+            while (!isStopped && !taskomaticApi.isRunning()) {
+                if (retries == 24) {
+                    LOG.error("Taskomatic API is not online after 2 minutes. Still waiting...");
+                }
+                else {
+                    LOG.info("Waiting for Taskomatic API to become online before connecting to the Salt event bus...");
+                }
+                try {
+                    Thread.sleep(5000L);
+                }
+                catch (InterruptedException e) {
+                    LOG.error("Interrupted while waiting for Taskomatic API to start", e);
+                    Thread.currentThread().interrupt();
+                    return;
+                }
+                retries++;
+            }
+            if (!isStopped) {
+                LOG.info("Taskomatic API is online. Connecting to the Salt event bus...");
+                connectToEventStream();
+            }
+        }, "salt-event-stream-initializer");
+        initializerThread.start();
     }
 
     /**
@@ -156,6 +204,9 @@ public class SaltReactor {
      */
     public void stop() {
         isStopped = true;
+        if (initializerThread != null) {
+            initializerThread.interrupt();
+        }
         if (eventStream != null) {
             eventStream.removeEventListener(listener);
             try {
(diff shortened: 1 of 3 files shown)


java/core/src/test/java/com/suse/manager/reactor/SaltReactorTest.java
```diff
@@ -0,0 +1,142 @@
...
+import com.suse.manager.webui.services.iface.SaltApi;
+import com.suse.manager.webui.services.iface.SystemQuery;
+import com.suse.salt.netapi.event.EventStream;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit test for {@link SaltReactor}.
+ */
+public class SaltReactorTest extends BaseTestCase {
```
mackdk, line 37: Does this test needs to access the database? It doesn't seem the case to me since you are basically passing mocks to all services. If that's true, I don't think you don't need to extend `BaseTestCase`.

For new tests, I would also suggest to follow junit 5 best practices and have the class and the test methods with package visibility instead of public.

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
```diff
@@ -148,14 +154,49 @@ public void start() {
...
+        this.taskomaticApi = taskomaticApiIn;
+    }
+
+    /**
+     * Start the salt reactor connection asynchronously, waiting for Taskomatic to be responsive first.
+     */
+    private void startEventStreamAsynchronously() {
+        initializerThread = new Thread(() -> {
+            while (!isStopped && !taskomaticApi.isRunning()) {
+                LOG.warn("Waiting for Taskomatic API to become online before connecting to the Salt event bus...");
```
mackdk, line 174: I don't this should be at level warning, info should be enough

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
```diff
@@ -148,14 +154,49 @@ public void start() {
...
+                try {
+                    Thread.sleep(5000L);
```
mackdk, line 176: This means we are going to wait forever if for some reason taskomatic is not starting, unless someone calls `stop()` and I did not find any usage of it a part from the unit test. Maybe we should put a limit on the number of retries? We could also adapt the delays as we retry instead of having a fixed 5 second delay.

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
```diff
@@ -148,14 +154,49 @@ public void start() {
...
+                }
+                catch (InterruptedException e) {
+                    LOG.error("Interrupted while waiting for Taskomatic API to start", e);
+                    Thread.currentThread().interrupt();
+                    return;
+                }
+            }
+            if (!isStopped) {
+                LOG.warn("Taskomatic API is online. Connecting to the Salt event bus...");
```
mackdk, line 185: This should rather be logged at info or debug instead of being a warning

java/core/src/test/java/com/suse/manager/reactor/SaltReactorTest.java
```diff
@@ -0,0 +1,142 @@
...
+ * This software is licensed to you under the GNU General Public License,
+ * version 2 (GPLv2). There is NO WARRANTY for this software, express or
+ * implied, including the implied warranties of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+ * along with this software; if not, see
+ * http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+ *
+ * Red Hat trademarks are not licensed under GPLv2. No permission is
+ * granted to use or replicate Red Hat trademarks that are incorporated
+ * in this software or its documentation.
```
mackdk, line 13: We don't need the `Red Hat trademarks` for new files

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
```diff
@@ -148,14 +154,49 @@ public void start() {
...
+     */
+    public void setTaskomaticApi(TaskomaticApi taskomaticApiIn) {
+        this.taskomaticApi = taskomaticApiIn;
+    }
+
+    /**
+     * Start the salt reactor connection asynchronously, waiting for Taskomatic to be responsive first.
+     */
+    private void startEventStreamAsynchronously() {
+        initializerThread = new Thread(() -> {
```
mackdk, line 172: I know we are probably calling `start()` exactly only once so this is a non-issue, but maybe we should still check if `initializerThread` is `null` or not running before creating and starting a new thread.

java/core/src/test/java/com/suse/manager/reactor/SaltReactorTest.java
nadvornik: fixed

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
nadvornik: fixed

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
nadvornik: Waiting forever is better than running without taskomatic - this would create actions pending forever in the database.
Now I log an error after 2 minutes but keep waiting.

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
nadvornik: fixed

java/core/src/main/java/com/suse/manager/reactor/SaltReactor.java
nadvornik: fixed

java/core/src/test/java/com/suse/manager/reactor/SaltReactorTest.java
nadvornik: fixed

mackdk (approved):
LGTM

PR #12341, opened 2026-07-27
Add Salt package download benchmark

AkashKumar7902:
## What does this PR change?

Adds an opt-in Cucumber benchmark that downloads every binary RPM from one Uyuni software channel on a configured set of Salt minions.

### Benchmark flow

1. Read the channel label and Salt minion IDs from environment variables.
2. Capture the channel's binary-RPM snapshot through the Uyuni API and confirm that every minion is registered and subscribed.
3. Find the ready Uyuni server pod.
4. Prepare the minions with:
   - `state.apply channels`
   - `grains.item os_family osarch`
   - `pkg.get_repo`
   - `pkg.refresh_db`
   - `pkg.list_repo_pkgs`
5. Clear the RPM payload cache outside the measured interval.
6. Run one concurrent Salt `cmd.run_all` workload on all selected minions:

   ```console
   zypper --quiet --non-interactive --no-refresh download \
     --all-matches \
     --repo susemanager:<channel-label> \
     '*'
   ```

7. Inventory the downloaded RPMs, match them to the channel checksums, and confirm that the channel snapshot remained unchanged.
8. Write a structured JSON result.

Commands follow this path:

```text
testsuite controller
  -> get_target('localhost').run_local
  -> kubectl exec into the Uyuni server pod
  -> salt --list <minions>
  -> Salt minions
```

Only the package download workload is timed. Setup, cache cleanup, and verification run outside the measured interval. The RPMs are downloaded into each minion's zypper package cache and are not installed.

The benchmark expects an existing channel and existing subscriptions. It does not create channels, subscribe systems, install a Helm release, or configure a StorageClass.

### Prerequisites

- The testsuite controller can reach the Kubernetes cluster with `kubectl`.
- The Uyuni server pod is running and ready in the `uyuni` namespace.
- The selected Salt minions are registered, reachable, and subscribed to the configured channel.
- The selected minions are SUSE systems with the same architecture.
- The channel contains binary RPM packages and is available on every selected minion.
- #12240 is present so a run set without Cobbler scenarios does not initialize the Cobbler client during Cucumber startup.

### Required inputs

| Variable | Description | Example |
| --- | --- | --- |
| `UYUNI_BENCH_MINIONS` | JSON array containing the exact Salt minion IDs | `["minion-1.example.test","minion-2.example.test"]` |
| `UYUNI_BENCH_CHANNEL` | Uyuni software channel label | `uyuni-bench-channel` |

The normal testsuite environment must also provide `SERVER` and its API configuration.

### Optional inputs

| Variable | Default | Description |
| --- | --- | --- |
| `UYUNI_BENCH_TIMEOUT_SECONDS` | `14400` | Timeout for the measured all-minion download workload |
| `UYUNI_BENCH_STORAGE_CLASS` | unset | StorageClass name recorded as result metadata |
| `SCENARIO_HARD_LIMIT` | workload timeout + 1800 seconds | Overall Cucumber scenario watchdog override |

Control operations use a maximum timeout of 600 seconds. `UYUNI_BENCH_STORAGE_CLASS` does not select or configure storage.

## How to run

Run the dedicated Rake task from a configured testsuite controller:

```bash
source /root/.bashrc
cd /root/spacewalk/testsuite

export UYUNI_BENCH_MINIONS='["minion-1.example.test","minion-2.example.test"]'
export UYUNI_BENCH_CHANNEL='uyuni-bench-channel'
export UYUNI_BENCH_TIMEOUT_SECONDS=14400
export UYUNI_BENCH_STORAGE_CLASS='nfs-csi'

bundle exec rake cucumber:kubernetes_package_download_benchmark
```

If StorageClass metadata is not needed, omit `UYUNI_BENCH_STORAGE_CLASS`.

Results are written to:

```text
testsuite/results/package-download/<timestamp>-<pid>/result.json
```

The result contains the measured duration, per-minion command status, package and payload counts, downloaded bytes, channel snapshot digest, and verification errors.

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- The configuration and run instructions are included in this PR description.

- [x] **DONE**

## Test coverage

Validated with:

- Ruby syntax checks for the benchmark steps and Rakefile;
- RuboCop for the benchmark step definitions;
- focused checks for API snapshot mapping and the Salt preflight sequence;
- focused checks for controller target selection, `kubectl exec` serialization, Salt arguments, zypper arguments, and timeout values;
- focused checks for RPM cache verification and result schema;
- run-set linkage validation;
- `git diff --check`.

- [x] **DONE**

## Links

Issue(s): openSUSE/mentoring#252  
Related: #12077  
Depends on: #12240  
Related Sumaform storage work: uyuni-project/sumaform#2246  
Port(s): none

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
+++ b/testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
@@ -1,7 +1,7 @@
 # Copyright (c) 2026 SUSE LLC
 # Licensed under the terms of the MIT license.
 
-@benchmark @long_running
+@benchmark
 Feature: Salt channel package download on configured minions
   In order to measure package storage performance
   As an operator of an existing Uyuni environment
--- a/testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
+++ b/testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
@@ -1,0 +1,18 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@benchmark @long_running
+Feature: Salt channel package download on configured minions
+  In order to measure package storage performance
+  As an operator of an existing Uyuni environment
+  I want configured Salt minions to download every package in a configured channel
+
+  Scenario: Download every configured channel package on every configured minion
+    Given the Salt package download benchmark inputs are valid
+    And the initial configured channel package snapshot is valid
+    And a ready server pod is reachable from the benchmark controller
+    And the benchmark minions are ready for the configured channel
+    When I clear RPM payload caches on the benchmark minions outside the measurement
+    And I execute and record the channel package downloads
+    Then the package download result report should exist
+    And every configured minion should have downloaded every channel package
--- a/testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
+++ b/testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
@@ -1,4 +1,4 @@
-# Copyright (c) 2026 SUSE LLC
+# Copyright (c) 2026 Akash Kumar <meakash7902@gmail.com>
 # Licensed under the terms of the MIT license.
 
 @benchmark
--- a/testsuite/features/step_definitions/package_download_benchmark_steps.rb
+++ b/testsuite/features/step_definitions/package_download_benchmark_steps.rb
@@ -7,152 +7,51 @@ require 'json'
 require 'shellwords'
 require 'time'
 
-PACKAGE_DOWNLOAD_BENCHMARK_DEFAULT_WORKLOAD_TIMEOUT = 14_400
-PACKAGE_DOWNLOAD_BENCHMARK_CONTROL_TIMEOUT = 600
-PACKAGE_DOWNLOAD_BENCHMARK_CACHE_ROOT = '/var/cache/zypp/packages'.freeze
-PACKAGE_DOWNLOAD_BENCHMARK_IDLE_POLL_SECONDS = 2
-PACKAGE_DOWNLOAD_BENCHMARK_SOURCE_ARCHES = %w[src nosrc source srcpackage].freeze
+PACKAGE_DOWNLOAD_DEFAULT_TIMEOUT = 14_400
+PACKAGE_DOWNLOAD_CONTROL_TIMEOUT = 600
+PACKAGE_DOWNLOAD_CACHE_ROOT = '/var/cache/zypp/packages'.freeze
+PACKAGE_DOWNLOAD_IDLE_POLL_SECONDS = 2
+PACKAGE_DOWNLOAD_SOURCE_ARCHES = %w[src nosrc source srcpackage].freeze
 
-# Parse one required JSON environment variable.
-def package_download_benchmark_json_env(name)
-  value = ENV.fetch(name, nil)
-  raise "#{name} must be set" if value.nil?
-
-  JSON.parse(value)
-rescue JSON::ParserError => e
-  raise "#{name} must contain valid JSON: #{e.message}"
-end
-
-# Return the validated Salt minion inventory.
-def package_download_benchmark_minions
-  minions = package_download_benchmark_json_env('UYUNI_BENCH_MINIONS')
+# Load the configured benchmark inputs.
+def package_download_inputs
+  minions = JSON.parse(ENV.fetch('UYUNI_BENCH_MINIONS', ''))
   raise 'UYUNI_BENCH_MINIONS must be a non-empty JSON array' unless minions.is_a?(Array) && !minions.empty?
 
-  minions.each do |minion|
-    valid = minion.is_a?(String) && !minion.empty? && !minion.start_with?('-') && !minion.match?(/[,\s[:cntrl:]]/)
-    raise "Invalid Salt minion ID: #{minion.inspect}" unless valid
-  end
-  raise 'UYUNI_BENCH_MINIONS must not contain duplicate IDs' unless minions.uniq.length == minions.length
+  valid_minions =
+    minions.all? do |minion|
+      minion.is_a?(String) &&
+        !minion.empty? &&
+        !minion.start_with?('-') &&
+        !minion.match?(/[,\s[:cntrl:]]/)
+    end
+  raise 'UYUNI_BENCH_MINIONS contains an invalid Salt ID' unless valid_minions
+  raise 'UYUNI_BENCH_MINIONS contains duplicate Salt IDs' unless minions.uniq.length == minions.length
 
-  minions
-end
+  channel = ENV.fetch('UYUNI_BENCH_CHANNEL', '')
+  raise 'UYUNI_BENCH_CHANNEL is invalid' unless channel.match?(/\A[A-Za-z0-9][A-Za-z0-9_.-]*\z/)
 
-# Return the validated Uyuni software channel label.
-def package_download_benchmark_channel
-  channel = ENV.fetch('UYUNI_BENCH_CHANNEL', nil)
-  raise 'UYUNI_BENCH_CHANNEL must be set' if channel.nil?
-
-  valid = channel.match?(/\A[A-Za-z0-9][A-Za-z0-9_.-]*\z/) && channel.length <= 128
-  raise 'UYUNI_BENCH_CHANNEL must be a valid software channel label' unless valid
-
-  channel
-end
-
-# Return the validated package-download workload timeout.
-def package_download_benchmark_workload_timeout
-  value = ENV.fetch('UYUNI_BENCH_TIMEOUT_SECONDS', PACKAGE_DOWNLOAD_BENCHMARK_DEFAULT_WORKLOAD_TIMEOUT.to_s)
-  timeout = Integer(value, 10)
+  timeout = Integer(ENV.fetch('UYUNI_BENCH_TIMEOUT_SECONDS', PACKAGE_DOWNLOAD_DEFAULT_TIMEOUT.to_s), 10)
   raise 'UYUNI_BENCH_TIMEOUT_SECONDS must be between 1 and 86400' unless timeout.between?(1, 86_400)
 
-  timeout
+  {
+    minions: minions,
+    channel: channel,
+    repo_alias: "susemanager:#{channel}",
+    storage_class: ENV.fetch('UYUNI_BENCH_STORAGE_CLASS', nil),
+    timeout_seconds: timeout
+  }
+rescue JSON::ParserError
+  raise 'UYUNI_BENCH_MINIONS must contain valid JSON'
 rescue ArgumentError
   raise 'UYUNI_BENCH_TIMEOUT_SECONDS must be an integer'
 end
 
-# Return the optional StorageClass result metadata.
-def package_download_benchmark_storage_class
-  value = ENV.fetch('UYUNI_BENCH_STORAGE_CLASS', nil)
-  return if value.nil?
-
-  raise 'UYUNI_BENCH_STORAGE_CLASS must not be empty' if value.empty?
-  raise 'UYUNI_BENCH_STORAGE_CLASS must not exceed 253 characters' if value.length > 253
-
-  value
-end
-
-# Load the complete benchmark input contract.
-def package_download_benchmark_inputs
-  {
-    minions: package_download_benchmark_minions,
-    channel: package_download_benchmark_channel,
-    storage_class: package_download_benchmark_storage_class,
-    timeout_seconds: package_download_benchmark_workload_timeout
-  }
-end
-
-# Return one required string field from a channel package API record.
-def package_download_benchmark_package_string(package, field, allow_blank: false)
-  value = package[field]
-  raise "Package #{package['id'].inspect} has invalid #{field}" unless value.is_a?(String)
-
-  value = value.strip
-  raise "Package #{package['id'].inspect} has blank #{field}" if !allow_blank && value.empty?
-
-  value
-end
-
-# Convert a channel package API record to the production package tuple.
-def package_download_benchmark_package(package)
-  raise "Invalid channel package record: #{package.inspect}" unless package.is_a?(Hash)
-
-  id = package['id']
-  raise "Channel package has invalid id: #{id.inspect}" unless id.is_a?(Integer) && id.positive?
-
-  name = package_download_benchmark_package_string(package, 'name')
-  version = package_download_benchmark_package_string(package, 'version')
-  release = package_download_benchmark_package_string(package, 'release', allow_blank: true)
-  epoch = package_download_benchmark_package_string(package, 'epoch', allow_blank: true)
-  api_arch = package_download_benchmark_package_string(package, 'arch_label')
-  checksum = package_download_benchmark_package_string(package, 'checksum')
-  checksum_type = package_download_benchmark_package_string(package, 'checksum_type')
-  arch = api_arch.sub(/-deb\z/, '')
-
-  raise "Package #{id} has an invalid retracted flag" unless [true, false, nil].include?(package['retracted'])
-  raise "Package #{id} has an invalid name" unless name.match?(/\A[A-Za-z0-9][A-Za-z0-9+_.-]*\z/)
-  raise "Package #{id} has an invalid architecture" unless arch.match?(/\A[A-Za-z0-9][A-Za-z0-9_.-]*\z/)
-  raise "Package #{id} is not a binary package (#{api_arch})" if PACKAGE_DOWNLOAD_BENCHMARK_SOURCE_ARCHES.include?(arch)
-  raise "Package #{id} has a non-numeric epoch" unless epoch.empty? || epoch.match?(/\A[0-9]+\z/)
-  raise "Package #{id} has an invalid checksum" unless checksum.match?(/\A[0-9A-Fa-f]+\z/)
-  raise "Package #{id} has an invalid checksum type" unless checksum_type.match?(/\A[A-Za-z0-9_-]+\z/)
-  raise "Package #{id} is not an RPM package (#{api_arch})" if api_arch.end_with?('-deb')
-
-  evr = +''
-  evr << "#{epoch}:" unless epoch.empty?
-  evr << version
-  evr << "-#{release}" unless release.empty? || release == 'X'
-  raise "Package #{id} has an invalid universal EVR" unless evr.match?(/\A[A-Za-z0-9][A-Za-z0-9+_.~:^%-]*\z/)
-
-  {
-    id: id,
-    name: name,
-    api_arch: api_arch,
-    arch: arch,
-    epoch: epoch,
-    version: version,
-    release: release,
-    evr: evr,
-    cache_evr: evr.sub(/\A0:/, ''),
-    checksum: checksum,
-    checksum_type: checksum_type,
-    retracted: package['retracted'],
-    tuple: [name, arch, evr]
-  }
-end
-
-# Raise when one value maps to more than one channel package.
-def package_download_benchmark_reject_duplicates(packages, description, &identity)
-  duplicates = packages.group_by(&identity).select { |_value, records| records.length > 1 }
-  return if duplicates.empty?
-
-  details = duplicates.map { |value, records| "#{value.inspect} (ids #{records.map { |record| record[:id] }.join(', ')})" }
-  raise "Channel contains duplicate #{description}: #{details.join('; ')}"
-end
-
-# Return a benchmark API client pointed at the configured Uyuni endpoint.
-def package_download_benchmark_api
-  @package_download_benchmark_api ||=
+# Return an API client for the configured Uyuni server.
+def package_download_api
+  @package_download_api ||=
     begin
-      host = ENV.fetch('SERVER', nil).strip
+      host = ENV.fetch('SERVER', '').strip
       raise 'SERVER must not be empty' if host.empty?
 
       if $api_test.is_a?(ApiTestXmlrpc)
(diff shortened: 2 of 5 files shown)


testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2026 SUSE LLC
+# Licensed under the terms of the MIT license.
+
+@benchmark @long_running
```
Pablogoliva, line 4: Remove long_running tag.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,1065 @@
...
+# Build a kubectl exec command for the Uyuni container.
+def package_download_benchmark_kubectl(pod, argv)
+  ['kubectl', '--namespace', 'uyuni', 'exec', '--container', 'uyuni', pod, '--', *argv]
+end
+
+# Find one ready Uyuni server pod.
+def package_download_benchmark_server_pod
+  stdout, stderr, code = package_download_benchmark_run(
+    [
+      'kubectl',
```
Pablogoliva, line 259: There's an standalone function for this called run, pls use it instead putting the command in a array of strings.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,1065 @@
...
+
+  pod_name
+rescue JSON::ParserError, KeyError => e
+  raise "Unable to parse the Uyuni server pod response: #{e.message}"
+end
+
+# Build one synchronous Salt list-target command.
+def package_download_benchmark_salt(inputs, function, arguments, timeout_seconds: inputs[:timeout_seconds])
+  [
+    'salt',
```
Pablogoliva, line 295: Maybe here we have to use venv-salt, I have to check.

Pablogoliva, line 295: Also use run instead of it.

Pablogoliva (changes requested):

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,1065 @@
...
+end
+
+# Build a kubectl exec command for the Uyuni container.
+def package_download_benchmark_kubectl(pod, argv)
+  ['kubectl', '--namespace', 'uyuni', 'exec', '--container', 'uyuni', pod, '--', *argv]
+end
+
+# Find one ready Uyuni server pod.
+def package_download_benchmark_server_pod
+  stdout, stderr, code = package_download_benchmark_run(
```
Pablogoliva, line 257:
```suggestion
  get_target('controller').run(
```

testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
AkashKumar7902: Addressed in `c8594d3a52`. Removed `@long_running`. The dedicated run set keeps the normal watchdog active and derives its limit from `UYUNI_BENCH_TIMEOUT_SECONDS` plus 30 minutes for setup and verification; an explicit `SCENARIO_HARD_LIMIT` still wins.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Addressed in `c8594d3a52`. Removed the custom `Open3` runner. Pod discovery, shared Salt calls, cache reset, and the measured workload now execute through `RemoteNode#run`; argv values are serialized with `Shellwords`, stdout/stderr remain separate, nonzero statuses remain reportable, and TERM/KILL process timeouts are preserved.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: I checked the current testsuite conventions and kept the server-side `salt` CLI. Existing server-side calls use `salt`, `salt-run`, and `salt-key`; `venv-salt-minion` and `venv-salt-call` are selected for minion-side operations. I found no server-side `venv-salt` master CLI convention to use here.

AkashKumar7902: Addressed in `c8594d3a52`. This Salt execution now goes through `get_target('localhost').run(...)` with separate output, explicit status handling, safe quoting, and the existing process timeout semantics.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Addressed in `c8594d3a52` using the registered controller target name. The current target map defines `localhost -> HOSTNAME` for the test runner and has no `controller` key, so the implementation uses `get_target('localhost').run(...)` rather than introducing a second alias.

AkashKumar7902:
@Pablogoliva Review updates are pushed in `c8594d3a52` on top of current `master`:

- removed `@long_running` while keeping an active run-set-specific watchdog;
- replaced every custom `Open3` execution path with `RemoteNode#run` on the registered `localhost` controller target;
- preserved shell-safe argv handling, separate stdout/stderr, explicit exit status, and TERM/KILL timeouts;
- verified that plain `salt` is the established server-side CLI while `venv-salt-*` selection is minion-side; and
- corrected the PR description so it no longer claims closed Sumaform #2241 supplies `UYUNI_BENCH_MINIONS`.

Ruby syntax, targeted RuboCop, runner/quoting assertions, and diff checks pass. A new full benchmark run still requires deployed Uyuni infrastructure and minions.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,1087 @@
...
+def package_download_benchmark_kubectl(pod, argv)
+  ['kubectl', '--namespace', 'uyuni', 'exec', '--container', 'uyuni', pod, '--', *argv]
+end
+
+# Find one ready Uyuni server pod.
+def package_download_benchmark_server_pod
+  timeout = 30
+  command =
+    package_download_benchmark_timeout_command(
+      [
```
Pablogoliva, line 268: For system commands pls use the method run or run_local. Instead of the one you have created there.

Pablogoliva (changes requested):
Change the way to use system commands.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Updated in 2d693f5c9f. The custom package_download_benchmark_timeout_command and GNU timeout/TERM/KILL wrapper are removed. All four controller-side command sites now call get_target("localhost").run_local directly; package_download_benchmark_kubectl_command only performs shell-safe serialization. Pod discovery uses the normal framework default, while only the genuinely long Salt/cache/workload operations pass an explicit RemoteNode timeout. I also fixed ssh_command so the requested timeout reaches ssh_exec! instead of silently reverting to DEFAULT_TIMEOUT (250 seconds). Ruby syntax, targeted RuboCop, command-routing/quoting assertions, timeout propagation, and diff checks pass.

AkashKumar7902:
@Pablogoliva Follow-up on the system-command review is complete on current head.

- The benchmark now calls the standard `RemoteNode#run_local` method directly at every controller execution site.
- The shared SSH helper now actually forwards the caller's timeout to the SSH channel; a real transport timeout reports 124, while unrelated transport failures remain -1.

I also ran the full mounted-volume pipeline against the external RKE2 environment. The isolated checkout included the declared #12240 dependency; its `testsuite` directory was bind-mounted into the controller, with `kubectl` and the RKE2 kubeconfig mounted read-only.

```text
1 scenario passed; 8 / 8 steps passed
measured workload:        264.839248 seconds
successful Salt targets:  2 / 2
verified packages:        6,054 / 6,054
verified payload bytes:   37,105,623,752
missing/mismatched/extra: 0 / 0 / 0
timed_out:                false
uncertain_completion:     false
channel snapshot stable:  yes
```

The PR description now contains the updated timeout model and validation result. Please re-review the current head when convenient.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: The live end-to-end run confirms the server-side CLI choice: plain `salt` successfully addressed both configured clients, including their channel state, repository refresh, concurrent `cmd.run_all`, and verification calls. Both targets returned complete structured results and all 6,054 payloads verified. I am keeping the established server-side `salt` convention.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,1075 @@
...
+  command = package_download_benchmark_kubectl_command(pod, salt)
+  started_at = Time.now.utc
+  started_monotonic = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
+  stdout = ''
+  stderr = ''
+  code = nil
+  exception = nil
+  begin
+    stdout, stderr, code = get_target('localhost').run_local(
```
Pablogoliva:
```suggestion
    stdout, stderr, code = get_target('controller').run_local(
```
Also in the other occurrences.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,1075 @@
...
+
+# Query and verify the structured downloaded-package inventory after timing.
+def package_download_benchmark_verify(inputs, pod)
+  control_timeout = [inputs[:timeout_seconds], PACKAGE_DOWNLOAD_BENCHMARK_CONTROL_TIMEOUT].min
+  command_timeout = [control_timeout - 15, 1].max
+  returns =
+    package_download_benchmark_salt_call(
+      inputs,
+      pod,
+      'cmd.run_all',
```
Pablogoliva: What's the matter of this?

AkashKumar7902: It runs a Python inventory command on each minion to collect the paths and sizes of files under /var/cache/zypp/packages. The returned inventory is used for post-download verification.

  Order:

  1. Before benchmark
     cmd.run_all → clear old RPM cache

  2. During benchmark (timed)
     cmd.run_all → run zypper download

  3. After benchmark
     cmd.run_all → inventory downloaded RPM files

  4. Verification
     Compare inventory with expected channel packages

 So line 848 is post-benchmark verification and is not included in the measured duration.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: the existing code already runs on the controller. 

plus, the controller is not defined in:
https://github.com/uyuni-project/uyuni/blob/master/testsuite/features/support/constants.rb

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,710 @@
...
+      }
+    end
+  workload_errors = []
+  workload_errors << workload_error unless workload_error.nil?
+  workload_errors << 'one or more minions failed the package download' if workload_per_minion.any? { |minion| minion[:status] == 'failed' }
+  workload = {
+    status: workload_errors.empty? ? 'passed' : 'failed',
+    command: zypper,
+    timeout_seconds: timeout,
+    timed_out: workload_error.to_s.include?('exited with 124'),
```
cbosdo: This exit code is not in zypper's man page with the others… Where does it come from?

testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2026 SUSE LLC
```
cbosdo: Copyright should be yours: change the SUSE LLC with your full name

testsuite/run_sets/kubernetes_package_download_benchmark.yml
cbosdo, line 1: How comes that you add a new run set? Shouldn't this run after the reposync benchmark to reuse the imported repository? In which case it should be in the same run_set.

testsuite/features/support/network_utils.rb
```diff
@@ -21,11 +21,13 @@ def ssh_command(command, host, port: 22, timeout: DEFAULT_TIMEOUT, buffer_size:
...
     Timeout.timeout(timeout) do # Enforce timeout on the entire SSH operation
       Net::SSH.start(host, nil, port: port, verify_host_key: :never, timeout: timeout, keepalive: true, max_pkt_size: buffer_size, config: true) do |ssh|
-        stdout, stderr, exit_code = ssh_exec!(ssh, command, timeout: DEFAULT_TIMEOUT)
+        stdout, stderr, exit_code = ssh_exec!(ssh, command, timeout: timeout)
       end
     end
   rescue Timeout::Error
-    puts "SSH operation timed out after #{timeout} seconds."
+    stderr = "SSH operation timed out after #{timeout} seconds."
+    exit_code = 124
```
cbosdo: Where is this exit_code coming from?

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,710 @@
...
+# Prepare a fixed package snapshot before the benchmark:
+# 1. Read the channel packages, subscribed systems, and Salt-to-Uyuni system ID mapping.
+# 2. Keep the binary RPMs and verify that every selected minion is registered and subscribed.
+# 3. Save the package records and their SHA-256 digest for comparison after the download.
+# If the digest changes, the result is invalid because the tested package set was not stable.
+Given('the initial configured channel package snapshot is valid') do
+  inputs = @package_download_inputs
+  api = $api_test
+
+  # 1. Read the current channel state from the Uyuni API.
```
cbosdo: AI tends to add numbered comments, but this is a pain to maintain in the long run. Remove the numbered list items from them please.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,710 @@
...
+  @package_download_inputs = inputs.merge(
+    packages: packages,
+    system_ids: system_ids,
+    snapshot_captured_at: Time.now.utc,
+    snapshot_digest: Digest::SHA256.hexdigest(JSON.generate(snapshot_records.sort_by(&:first))),
+    subscribed_system_count: subscribed_ids.length
+  )
+end
+
+Given('a ready server pod is reachable from the benchmark controller') do
```
cbosdo, line 265: This function needs to be shared with the other PR as it's a duplicate. Better base this PR on the other one to share code with it

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,710 @@
...
+# 1. Apply the Salt channels state to update the minion repository configuration.
+# 2. Read the OS family and architecture, then require matching SUSE clients.
+# 3. Verify that the configured Uyuni repository exists and is enabled.
+# 4. Refresh the repository metadata.
+# 5. Verify that every RPM from the initial snapshot is available to every minion.
+Given('the benchmark minions are ready for the configured channel') do
+  inputs = @package_download_inputs
+  pod = @package_download_pod
+
+  # 1. Run `state.apply channels` so Uyuni updates the assigned software channel
```
cbosdo: Again numbered comments…

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: ok, should i just remove the numbers, or the enitre comment ?

testsuite/features/step_definitions/package_download_benchmark_steps.rb
```diff
@@ -0,0 +1,710 @@
...
+  zypper = [
+    'zypper',
+    '--quiet',
+    '--non-interactive',
+    '--no-refresh',
+    'download',
+    '--all-matches',
+    '--repo',
+    inputs[:repo_alias],
+    '*'
```
cbosdo, line 453: I wonder if we could have issue with packages conflicting each other in the repo

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: it comes from network_utils.rb where we throw this exit code when timeout happens

testsuite/features/support/network_utils.rb
AkashKumar7902: it is a timeout error code which we are synthesizing

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: updated

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: updated, removed the numbers

testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
AkashKumar7902: done

testsuite/run_sets/kubernetes_package_download_benchmark.yml
AkashKumar7902: I though of keeping it separate , will be making changes to accomodate both

testsuite/features/support/network_utils.rb
```diff
@@ -21,11 +21,13 @@ def ssh_command(command, host, port: 22, timeout: DEFAULT_TIMEOUT, buffer_size:
   begin
     Timeout.timeout(timeout) do # Enforce timeout on the entire SSH operation
       Net::SSH.start(host, nil, port: port, verify_host_key: :never, timeout: timeout, keepalive: true, max_pkt_size: buffer_size, config: true) do |ssh|
-        stdout, stderr, exit_code = ssh_exec!(ssh, command, timeout: DEFAULT_TIMEOUT)
+        stdout, stderr, exit_code = ssh_exec!(ssh, command, timeout: timeout)
```
Pablogoliva, line 24: Leave the previous timeout pls

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Kept `get_target('localhost')`: there is no `controller` key in `ENV_VAR_BY_HOST`, and the merged reposync benchmark (#12077) uses `get_target('localhost').run_local` for the same kubectl calls. The pod lookup step is gone in 2c1ac3cd0; the benchmark now calls `reposync_benchmark_server_pod` from that file, so both benchmarks share one implementation. OK to resolve?

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: This runs after the timed workload and lists what zypper actually downloaded, so the result can be checked against the channel package list. In 2c1ac3cd0 the inline Python script is replaced by `find /var/cache/zypp/packages -type f -printf '%s\t%p\n'`, which prints one "size, path" line per file and is easier to read.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Removed in 70fb1fba: the synthesized exit code and the `timed_out` field are gone. A timeout now surfaces as the exception raised by the framework helper and is recorded as `uncertain_completion` in the result.

testsuite/features/build_validation/benchmarks/srv_salt_package_download_benchmark.feature
AkashKumar7902: Done in 47c5389a (author name and email).

testsuite/run_sets/kubernetes_package_download_benchmark.yml
AkashKumar7902: Agreed and done in 2c1ac3cd0: the separate run set is removed. The feature now lives next to the reposync one as `features/kubernetes/srv_rke2_package_download_benchmark.feature` and is the second entry of `run_sets/kubernetes_storage_benchmark.yml`. It picks up the channel label the reposync scenario created (shared through `reposync_benchmark_channel_label`), waits until the server has generated the channel's repodata, and subscribes the configured minions to it through `system.scheduleChangeChannels` before measuring. `UYUNI_BENCH_CHANNEL_LABEL` still lets it run on its own against an existing channel.

testsuite/features/support/network_utils.rb
AkashKumar7902: Removed in 70fb1fba, and 2c1ac3cd0 drops the whole network_utils.rb change from this PR (see the thread below).

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Removed in 7e90cfc5 and 2229bd82; no numbered list items remain in the step file.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Done in 2c1ac3cd0 now that #12077 is merged and the branch is rebased on master: the duplicated pod lookup step is removed and the benchmark calls `reposync_benchmark_server_pod`; it also reuses `reposync_benchmark_run_in_server_pod` (repodata readiness check), `reposync_benchmark_integer_env` (timeout parsing) and `reposync_benchmark_storage_backend` (result label).

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: Removed in 7e90cfc5 and 2229bd82; no numbered list items remain in the step file.

testsuite/features/step_definitions/package_download_benchmark_steps.rb
AkashKumar7902: No: `zypper download` only fetches RPMs into the package cache. It selects the packages with a pool query (name match, `--repo` limits it to the benchmark repository) and never runs the dependency solver, so packages that conflict with each other, obsolete each other, or come in several versions are all downloaded side by side; `--all-matches` is what keeps every version instead of only the best one. Nothing is skipped because it is installed; the only skip is an RPM already present in `/var/cache/zypp/packages`, and that cache is wiped in the step before the measurement. The post-run inventory check is the real completeness proof, because `zypper download` can still exit 0 when a single transfer fails. The full run against a 6,054-package channel completed with 0 missing, mismatched or extra payloads. I added a two-line comment above the zypper command saying this in 2c1ac3cd0.

testsuite/features/support/network_utils.rb
AkashKumar7902: Reverted in 2c1ac3cd0; network_utils.rb is no longer touched by this PR. As agreed on #12077, the run instructions export `DEFAULT_TIMEOUT=14400` before starting the benchmark run set.

PR #12346, opened 2026-07-28
Use helm-docs and helm-schema to minimize doc duplicates in charts

cbosdo:
## What does this PR change?

The helm chart's README and values schema are now generated using
helm-docs and helm schema. The source of truth were it all starts is the
values.yaml.
    
The README.md and the values.schema.json are generated using a pre-commit hook.
A github action verifies that those have been updated or not modified directly.

## GUI diff

No difference.

Before:

After:

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- GH action tests were added

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge


containers/server-helm/values.yaml
```diff
@@ -1,957 +1,1019 @@
...
+    # @default `nil`
+    # @schema
+    # type: ["object", "null"]
+    # @schema
+    annotations: null
 
-# exposeJavaDebug will expose the 8001, 8002 and 8003 ports to connect a Java debugger
-# to taskomatic, search server and tomcat respectively
+# exposeJavaDebug -- Expose ports 8001, 8002, and 8003 for Java debugging if true
 exposeJavaDebug: false
```
aaannz, line 314: I don't know enough about the schema generator, but even though here it is as a boolean, when I check values.schema.json it has value `"default": "false",` which is string value without any other type info. Previously it was explicitly marked as type boolean and also with boolean value.
When I consulted this with the AI, this apparently means any string value pass validation and helm then assumes any value as a true

containers/proxy-helm/values.yaml
```diff
@@ -58,174 +165,285 @@ cert:
...
+placement:
+
+# @schema
+# $ref: "#/definitions/nodeSelectorConfig"
+# @schema
+# proxy -- Proxy pod placement configuration
+proxy:
+
+# @schema
+# ref: "#/definitions/componentConfig"
```
aaannz:
```suggestion
# $ref: "#/definitions/componentConfig"
```

containers/proxy-helm/values.yaml
```diff
@@ -58,174 +165,285 @@ cert:
...
+  # @schema
+  # httpd.tag -- Tag for the Apache server container, overwriting the global one
   tag:
 
-  # Apache configuration tuning to use
+  # httpd.tuning -- Custom configuration for the apache server
   tuning: ""
 
+# @schema
+# ref: "#/definitions/componentConfig"
```
aaannz:
```suggestion
# $ref: "#/definitions/componentConfig"
```

containers/proxy-helm/values.yaml
```diff
@@ -58,174 +165,285 @@ cert:
...
+  # @schema
+  # squid.tag -- Tag for the Squid server container, overwriting the global one
   tag:
 
-  # Squid configuration tuning to use
+  # squid.tuning -- Custom configuration for the squid server
   tuning: ""
 
+# @schema
+# ref: "#/definitions/componentConfig"
```
aaannz:
```suggestion
# $ref: "#/definitions/componentConfig"
```

containers/proxy-helm/values.yaml
```diff
@@ -58,174 +165,285 @@ cert:
...
   tag:
 
-  # Use the host network to bypass the Kubernetes network layers.
+  # tftp.hostNetwork -- Use the host network to bypass the Kubernetes network layers.
   # This may be needed if not using a LoadBalancer for the TFTP server.
   # Note that this may be more complex to manage on multi-node clusters.
   hostNetwork: false
 
+# @schema
+# ref: "#/definitions/componentConfig"
```
aaannz:
```suggestion
# $ref: "#/definitions/componentConfig"
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,56 @@
...
+        language: system
+        entry: /bin/sh
+        args:
+          - -c
+          - |
+            output=$(helm-docs -c containers/server-helm -s file -z '^services\.[^.]*\..*$' -z '^volumes\.[^.]*\..*$' -x 2>&1 | grep "Error parsing information")
+            if [ -n "$output" ]; then
+              printf "%b\n" "$output"
+              exit 1
+            fi
```
aaannz: Is it intentional that when `helm-docs` binary is missing this test passes?

.github/workflows/helm-charts.yml
```diff
@@ -0,0 +1,34 @@
...
+      - 'containers/proxy-helm/**'
+      - 'containers/server-helm/**'
+
+permissions:
+  contents: read
+
+jobs:
+  helm-charts-tests:
+    runs-on: ubuntu-latest
+    container: registry.opensuse.org/home/cbosdonnat/branches/systemsmanagement/uyuni/master/docker/containers/uyuni-helm:latest
```
aaannz: I assume this is temporary, right?

containers/proxy-helm/test.sh
```diff
@@ -1,11 +1,52 @@
 #!/bin/sh
 
-# Requires to install helm unittest plugin:
-#   helm plugin install --verify=false https://github.com/helm-unittest/helm-unittest
-helm unittest .
+HELM_UNITTEST=helm-unittest
+which $HELM_UNITTEST 2>/dev/null 1>&2
+if test $? -ne 0; then
```
aaannz: Personally I abandoned using `which` as it is a separate binary which may not be present. Where `command` is in every POSIX shell.


```suggestion
if command -v "${HELM_UNITTEST}" >/dev/null then
```

.pre-commit-config.yaml
cbosdo: Nope, it's not intentional…

.github/workflows/helm-charts.yml
cbosdo: yes, until the images are moved to their final destination

containers/proxy-helm/test.sh
cbosdo: I wasn't aware of that one. I'll try it out.

containers/proxy-helm/values.yaml
cbosdo: oops!

containers/server-helm/values.yaml
cbosdo: Nice catch! I'll probably need to force the type in the schema declaration for this.

cbosdo: While investigating this I found out another serious issue with the server properties: none of the custom one where taken into account in the schema: it only used the components. Fixing this too

.pre-commit-config.yaml
cbosdo: I just harmonized the pre-commit hooks to now run the `tests.sh` like other scripts.

containers/proxy-helm/test.sh
cbosdo: Working like a charm, adopting it!

containers/proxy-helm/values.yaml
cbosdo: I removed the whole componentConfig as it hides additional properties

containers/proxy-helm/values.yaml
cbosdo: I removed the whole componentConfig as it hides additional properties

containers/proxy-helm/values.yaml
cbosdo: I removed the whole componentConfig as it hides additional properties

containers/proxy-helm/values.yaml
cbosdo: I removed the whole componentConfig as it hides additional properties

.github/workflows/helm-charts.yml
cbosdo: Fixed now

PR #12360, opened 2026-07-31
Fixed invalid date format (bsc#1271467)

bisht-richa:
## What does this PR change?

**Improve handling of invalid issue_date values when creating filters via the API.**

## Codespace

## GUI diff
For filters containing an invalid issue_date stored in the database, the UI now displays a warning message and allows the user to select a new date and save the filter.
Screenshot 2026-07-31 at 16 33 33

- [x] **DONE**

## Documentation
- No documentation needed: **handling of invalid issue_date values *

- [x] **DONE**

## Test coverage

- No tests: already covered

- [x] **DONE**

## Links

Issue(s): [#](https://github.com/SUSE/spacewalk/issues/31240)
Port(s): [# **5.1**](https://github.com/SUSE/spacewalk/pull/31535) [# **5.2**](https://github.com/SUSE/spacewalk/pull/31557)

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
@@ -23,6 +23,9 @@ import com.redhat.rhn.manager.contentmgmt.ContentManager;
 
 import org.apache.commons.lang3.StringUtils;
 
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.regex.Pattern;
 
 /**
@@ -146,5 +149,27 @@ public class ContentPropertiesValidator {
     public static boolean isLabelValid(String label) {
         return Pattern.compile(CreateChannelCommand.CHANNEL_LABEL_REGEX).matcher(label).find() &&
                 Pattern.compile(CreateChannelCommand.CHANNEL_NAME_REGEX).matcher(label).find();
+    }
+
+    /**
+     * Validate an issue_date and build_date filter value.
+     *
+     * @param field the filter field
+     * @param value the filter value
+     */
+    public static void validateIssueDate(String field, String value) throws ValidatorException {
+
+        if ((!"issue_date".equals(field) && !"build_date".equals(field)) || StringUtils.isEmpty(value)) {
+            return;
+        }
+
+        try {
+            OffsetDateTime.parse(value, DateTimeFormatter.ISO_DATE_TIME);
+        }
+        catch (DateTimeParseException e) {
+            ValidatorResult result = new ValidatorResult();
+            result.addFieldError(field, "contentmanagement.invalid_issue_date");
+            throw new ValidatorException(result);
+        }
     }
 }
--- a/java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
@@ -19,12 +19,11 @@ import com.redhat.rhn.common.validator.ValidatorException;
 import com.redhat.rhn.common.validator.ValidatorResult;
 import com.redhat.rhn.domain.user.User;
 import com.redhat.rhn.manager.channel.CreateChannelCommand;
+import com.redhat.rhn.manager.contentmgmt.ContentManagementUtils;
 import com.redhat.rhn.manager.contentmgmt.ContentManager;
 
 import org.apache.commons.lang3.StringUtils;
 
-import java.time.OffsetDateTime;
-import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.regex.Pattern;
 
@@ -157,14 +156,14 @@ public class ContentPropertiesValidator {
      * @param field the filter field
      * @param value the filter value
      */
-    public static void validateIssueDate(String field, String value) throws ValidatorException {
+    public static void validateDateCriteria(String field, String value) throws ValidatorException {
 
-        if ((!"issue_date".equals(field) && !"build_date".equals(field)) || StringUtils.isEmpty(value)) {
+        if (!ContentManagementUtils.isDateCriteria(field, value)) {
             return;
         }
 
         try {
-            OffsetDateTime.parse(value, DateTimeFormatter.ISO_DATE_TIME);
+            ContentManagementUtils.parseDateCriteria(value);
         }
         catch (DateTimeParseException e) {
             ValidatorResult result = new ValidatorResult();
--- a/java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
+++ b/java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
@@ -53,6 +53,9 @@ import com.suse.manager.api.ReadOnly;
 
 import org.apache.commons.lang3.StringUtils;
 
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -745,10 +748,28 @@ public class ContentManagementHandler extends BaseHandler {
         if (!criteria.containsKey("matcher") || !criteria.containsKey("field")) {
             throw new InvalidArgsException("Incomplete filter criteria");
         }
+
+        String field = (String) criteria.get("field");
+        String value = StringUtils.trimToNull((String) criteria.get("value"));
+        validateCriteriaValue(field, value);
+
         return of(new FilterCriteria(
                 FilterCriteria.Matcher.lookupByLabel((String) criteria.get("matcher")),
-                (String) criteria.get("field"),
-                StringUtils.trimToNull((String) criteria.get("value"))));
+                field,
+                value));
+    }
+
+    private void validateCriteriaValue(String field, String value) {
+        if (!"issue_date".equals(field) || value == null || value.isEmpty()) {
+            return;
+        }
+
+        try {
+            OffsetDateTime.parse(value, DateTimeFormatter.ISO_DATE_TIME);
+        }
+        catch (DateTimeParseException e) {
+            throw new InvalidArgsException("Invalid issue_date format.");
+        }
     }
 
     /**
--- a/java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
+++ b/java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
@@ -53,9 +53,6 @@ import com.suse.manager.api.ReadOnly;
 
 import org.apache.commons.lang3.StringUtils;
 
-import java.time.OffsetDateTime;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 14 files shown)


java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/response/ProjectFilterResponse.java
```diff
@@ -58,7 +58,7 @@ public void setCriteriaKey(String criteriaKeyIn) {
      */
     public void setCriteriaValue(String criteriaValueIn) {
         // If we have a date as a criteria value we need to format it with the current user timezone
-        if (this.criteriaKey.equals("issue_date")) {
+        if (this.criteriaKey.equals("issue_date ")) {
```
EMaksy: Oh i guess a typo iwith an empty space here inn "issue_date " ? 
It therefore never matches real issue_date filters or ?

java/core/src/main/java/com/redhat/rhn/frontend/xmlrpc/contentmgmt/ContentManagementHandler.java
```diff
@@ -745,10 +748,28 @@ private Optional<FilterCriteria> createCriteria(Map<String, Object> criteria) {
...
+
         return of(new FilterCriteria(
                 FilterCriteria.Matcher.lookupByLabel((String) criteria.get("matcher")),
-                (String) criteria.get("field"),
-                StringUtils.trimToNull((String) criteria.get("value"))));
+                field,
+                value));
+    }
+
+    private void validateCriteriaValue(String field, String value) {
```
parlt91: It looks like you're basically doing the validation separately in two different places. Here and in `FilterApiController`. In addition the validation logic would be bypassed during tests that call `ContentManager` directly since they don't go through either endpoint.

I'd suggest to put a single validation method in a place e.g. `contentmgmt/validation/ContentPropertiesValidator.java`
and then use it in `ContentManger.createFilter()` function.

There is already a validation here that you could use for orientation
```
ContentPropertiesValidator.validateFilterProperties(name);
```

java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/handlers/FilterApiController.java
```diff
@@ -275,10 +286,18 @@ public static String createContentFilter(Request req, Response res, User user) {
     public static String updateContentFilter(Request req, Response res, User user) {
         FilterRequest updateFilterRequest = FilterHandler.getFilterRequest(req);
 
+        String updateCriteriaValue = StringUtils.trimToNull(updateFilterRequest.getCriteriaValue());
+        if (!validateCriteriaValue(updateFilterRequest.getCriteriaKey(), updateCriteriaValue)) {
+            return json(GSON, res, HttpStatus.SC_BAD_REQUEST, ResultJson.error(Collections.emptyList(),
+                    Collections.singletonMap("criteriaValue",
+                            Collections.singletonList("Invalid issue_date format."))),
```
parlt91: This string should be localized instead of being in English only.
you can create a key e.g. `"contentmanagement.invalid_date_format"` use it here instead and add it to the 
`frontend/strings/java/StringResource_en_US.xml`
translation file.
```
<trans-unit id="contentmanagement.invalid_date_format" xml:space="preserve">
  <source>Invalid date format. It must follow ISO 8601 offset format (e.g. 2026-06-04T22:00:00Z).</source>
</trans-unit>
```

parlt91 (changes requested):
The UI part look great, however for the java part I'd like to have the validation in one place instead of spreading different validations over multiple instead. See comments below.

java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
```diff
@@ -147,4 +150,26 @@ public static boolean isLabelValid(String label) {
...
+
+    /**
+     * Validate an issue_date filter value.
+     *
+     * @param field the filter field
+     * @param value the filter value
+     */
+    public static void validateIssueDate(String field, String value) throws ValidatorException {
+
+        if (!"issue_date".equals(field) || StringUtils.isEmpty(value)) {
```
parlt91: It looks like there can also be `build_date` when using the package filter. It would also suffer from missing date validation if used incorrectly. I think we should add it here as well:
```
if ((!"issue_date".equals(field) && !"build_date".equals(field)) ||
      StringUtils.isEmpty(value)) {
```
Mind the parenthesis!
Just add it here and in other places where the check is used.

java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/response/FilterResponse.java
```diff
@@ -62,12 +63,16 @@ public void setCriteriaKey(String criteriaKeyIn) {
      */
     public void setCriteriaValue(String criteriaValueIn) {
         // If we have a date as a criteria value we need to format it with the current user timezone
-        if (this.criteriaKey.equals("issue_date")) {
-            DateTimeFormatter timeFormatter = DateTimeFormatter.ISO_DATE_TIME;
-            OffsetDateTime offsetDateTime = OffsetDateTime.parse(
-                    criteriaValueIn, timeFormatter);
-            Date criteriaValueDate = Date.from(Instant.from(offsetDateTime));
-            criteriaValueIn = ViewHelper.getInstance().renderDate(criteriaValueDate);
+        if ("issue_date".equals(this.criteriaKey) && criteriaValueIn != null && !criteriaValueIn.isEmpty()) {
```
parlt91: There is also the `ProjectFilterResponse` class that is used when editing the filters from the project page. It has a similar check that also needs this change to make sure the ISE can't happen from there

parlt91 (commented):
Thanks
It looks much better now. I found two more things you should look into.

java/core/src/main/java/com/suse/manager/webui/controllers/contentmanagement/response/ProjectFilterResponse.java
```diff
@@ -58,12 +58,14 @@ public void setCriteriaKey(String criteriaKeyIn) {
...
-        if (this.criteriaKey.equals("issue_date")) {
-            DateTimeFormatter timeFormatter = DateTimeFormatter.ISO_DATE_TIME;
-            OffsetDateTime offsetDateTime = OffsetDateTime.parse(
-                    criteriaValueIn, timeFormatter);
-            Date criteriaValueDate = Date.from(Instant.from(offsetDateTime));
-            criteriaValueIn = ViewHelper.getInstance().renderDate(criteriaValueDate);
+        if (("issue_date".equals(this.criteriaKey) || "build_date".equals(this.criteriaKey)) &&
+            criteriaValueIn != null &&
+            !criteriaValueIn.isEmpty()) {
+                DateTimeFormatter timeFormatter = DateTimeFormatter.ISO_DATE_TIME;
```
parlt91, line 66: Almost there! Just wrap this into a try-catch block like you did for `FilterResponse` ;)

parlt91 (dismissed):
Just one more small thing

mackdk (dismissed):
I was pinged on the 5.2 version of this PR, but I'm commenting here to keep the discussion in a single places. Java code looks good as is, so feel free to merge it, but there is still a bit of duplication we can remove if you can spend more time on this. 

What about creating a class (maybe in `com.redhat.rhn.manager.contentmgmt`):
```java
public class ContentManagementUtils {

    private static final Set<String> SUPPORTED_DATE_FIELDS = Set.of("issue_date", "build_date");

    public static boolean isDateCriteria(String field, String value) {
        return field != null && SUPPORTED_DATE_FIELDS.contains(field) && StringUtils.isNotEmpty(value);
    }

    public static Date parseDateCriteria(String value) throws DateTimeParseException {
        OffsetDateTime offsetDateTime = OffsetDateTime.parse(value, DateTimeFormatter.ISO_DATE_TIME);
        return Date.from(offsetDateTime.toInstant());
    }
}
```
Then the `ContentPropertiesValidator` becomes:
```java
    public static void validateIssueDate(String field, String value) throws ValidatorException {

        if (!ContentManagementUtils.isDateCriteria(field, value)) {
            return;
        }

        try {
            ContentManagementUtils.parseDateCriteria(value);
        }
        catch (DateTimeParseException e) {
            ValidatorResult result = new ValidatorResult();
            result.addFieldError(field, "contentmanagement.invalid_issue_date");
            throw new ValidatorException(result);
        }
    }
```
while the `setCriteriaValue` of `FilterResponse` and `ProjectFilterResponse` becomes:
```java
     public void setCriteriaValue(String criteriaValueIn) {
        // If we have a date as a criteria value we need to format it with the current user timezone
        if (ContentManagementUtils.isDateCriteria(criteriaKey, criteriaValueIn)) {
            try {
                Date criteriaValueDate = ContentManagementUtils.parseDateCriteria(criteriaValueIn);
                criteriaValueIn = ViewHelper.getInstance().renderDate(criteriaValueDate);
            }
            catch (DateTimeParseException ignored) {
                // Keep the raw value if it cannot be parsed.
            }
        }

        this.criteriaValue = criteriaValueIn;
    }
```

web/html/src/manager/content-management/list-filters/filter-form.tsx
```diff
@@ -2,6 +2,7 @@ import { Fragment, useEffect, useState } from "react";
 
 import { Button } from "components/buttons";
 import { DateTime, DEPRECATED_Select, Form, Radio, Text } from "components/input";
+import { Messages, Utils as MessagesUtils } from "components/messages/messages";
```
EMaksy, line 5: Just an idea, maybe it would make sense to rename Utils overall to MessageUtils or even use dedicated functions to be more explicit, but this may be out of scope, just a general idea :) 

Example:
```
export function createInfoMessages(...): MessageType[] {}
export function createSuccessMessages(...): MessageType[] {}
export function createWarningMessages(...): MessageType[] {}
export function createErrorMessages(...): MessageType[] {}
```

web/html/src/manager/content-management/shared/type/filter.type.ts
```diff
@@ -42,4 +42,5 @@ export type FilterFormType = {
   kernelId?: number;
   kernelName?: string;
   channelId?: number;
+  issueDateFormatError?: boolean;
```
EMaksy, line 45: Maybe we could make the name a bit more explicit ? 
What about hasIssueDateParseError or isIssueDateInvalid ? 
With this we directly know if it is a boolean :)

web/html/src/manager/content-management/list-filters/filter.utils.ts
```diff
@@ -86,7 +98,13 @@ export function mapResponseToFilterForm(filtersResponse: FilterServerType[] = []
     // Custom filters mappers for complex filter forms
     // If this starts growing we could define mapper functions in the enum itself, for now it's enough. (ex: mapCriteriaValueToRequest())
     if (filterResponse.criteriaKey === clmFilterOptions.ISSUE_DATE.key) {
-      filterForm[clmFilterOptions.ISSUE_DATE.key] = localizedMoment(filterResponse.criteriaValue);
+      const parsedDate = parseIssueDateValue(filterResponse.criteriaValue);
+
+      // This allows the UI to warn the user that the stored value is invalid or non-empty issue_date.
+      filterForm.issueDateFormatError = !parsedDate && !_isEmpty(filterResponse.criteriaValue);
+
+      // If the stored issue_date is invalid, use the current date so the user
```
EMaksy, line 106: It seems like a part of the comment was removed ? 

Something like this ? 
// If the stored issue_date is invalid, use the current date so the user can correct and save the filter.

java/core/src/main/java/com/redhat/rhn/domain/contentmgmt/validation/ContentPropertiesValidator.java
```diff
@@ -147,4 +150,26 @@ public static boolean isLabelValid(String label) {
...
+     * @param value the filter value
+     */
+    public static void validateIssueDate(String field, String value) throws ValidatorException {
+
+        if ((!"issue_date".equals(field) && !"build_date".equals(field)) || StringUtils.isEmpty(value)) {
+            return;
+        }
+
+        try {
+            OffsetDateTime.parse(value, DateTimeFormatter.ISO_DATE_TIME);
```
EMaksy, line 167: I see that in other places ISO_OFFSET_DATE_TIME is used instead of ISO_DATE_TIME?
Which one is correct one and which should be used when ? :D

EMaksy (commented):
Hey @bisht-richa, thanks for working on this —> this is a tricky issue. I have a couple of questions about the PR.

Should the recovery handling also cover build_date? The backend validates both issue_date and build_date, and PackageFilter parses build_date using the same date-time format. Therefore, an existing malformed build_date value seems able to cause the same problem, while the UI currently only warns about issue_date.

Would it make sense to cover both in this PR, or should build_date be handled in a follow-up?

It would also be helpful to add tests for malformed stored date values and the UI recovery path, ideally covering both date filter types.

web/html/src/manager/content-management/list-filters/filter-form.tsx
bisht-richa: Yes, out of scope for this PR. I will create a separate issue or address this when we start UX work for the Alert component.

mackdk (approved):

PR #12373, opened 2026-08-04
Remove tftpd check from server sanity test

maximenoel8:
## What does this PR change?

Remove tftpd check from server.
It is deployed on proxy now.
We don't have check on proxy => https://github.com/SUSE/spacewalk/issues/31523
Discussion: https://suse.slack.com/archives/C02D134507M/p178575447862100

## Codespace

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): 
Port(s):  No need

- [x] **DONE**

## Changelogs

- [x] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/testsuite/features/core/allcli_sanity.feature
+++ b/testsuite/features/core/allcli_sanity.feature
@@ -25,7 +25,6 @@ Feature: Sanity checks
     And podman container "uyuni-server" should be running on "server"
     And podman container "uyuni-server" should be healthy on "server"
     And podman container "uyuni-saline-0" should be running on "server"
-    And podman container "uyuni-tftpd" should be running on "server"
     And podman container "uyuni-server-attestation-0" should be running on "server"
     And podman container "uyuni-hub-xmlrpc-0" should be running on "server"
 


ktsamis (approved):

testsuite/features/core/allcli_sanity.feature
```diff
@@ -25,7 +25,6 @@ Feature: Sanity checks
     And podman container "uyuni-server" should be running on "server"
     And podman container "uyuni-server" should be healthy on "server"
     And podman container "uyuni-saline-0" should be running on "server"
-    And podman container "uyuni-tftpd" should be running on "server"
```
nodeg, line 28: It is not necessarily wrong to remove this check, but keep in mind we have the `tftpd` container both on the proxy and the server. This was a change that is not properly documented and the container is also not started by default and needs to be enabled with systemd for example IIRC.

nodeg:
```bash
uyuni-long-master-server:~ # mgradm install podman --help | grep tftpd
      --tftpd-enable         Start the TFTP server container
      --tftpd-image string   Image for TFTPD container (default "uyuni/proxy-tftpd")
      --tftpd-tag string     Tag for TFTPD container, overrides the global value if set
```

maximenoel8: OK, should the fix be on sumaform side ? Is the tftp always deployed with proxy retail  ?

ktsamis: deploy_tftp = true in modules/server_containerized/variables.tf is always there, in all our envs, so yes indeed we should not remove this from the server because this is deployed, we need to check it. I will add the check in proxy as well without removing this

ktsamis (changes requested):
yes in the end we should not remove this, we need only to add it to the proxy, this container is deployed by default in all our envs so we need to keep testing for its existence

`deploy_tftp = true` in `modules/server_containerized/variables.tf`

PR #12379, opened 2026-08-04
Proxy fqdn

nadvornik:
## What does this PR change?

Improved handling of proxy FQDNs

- Allow specifying additional fqdns when creating a proxy config - this is a alternative for the import from certificate (which has been dropped)
- Validate and update proxy connection path
  - Fixed updating on kubernetes
  - Specified proxy parent must exist in the database before - this ensures that proxy connection can't diverge from salt connection
- In UI the free text for parent has be changed to select
- On kubernetes all specified FQDNs are configured for Ingress and Traefik

Better control over used proxy FQDN

In environments where an intermediate Proxy server has both **internal** and **external** FQDNs:
- Internal minions must address the Proxy using its internal FQDN (e.g., `proxy-internal.local`).
- The central Uyuni Server (located externally) must address the Proxy using its external FQDN (e.g., `proxy-external.com`).

In this PR:
rhnServerPath.hostname contains the internal FQDN used by minion to connect the proxy or used by proxy to connect the parent

Primary FQDN is the external FQDN used for Salt SSH and for monitoring.

This cleanly decouples the network perspective of the internal minions from the network perspective of the external server, allowing both incoming minion connections and outgoing Salt-SSH tunnels to operate concurrently and correctly.

For compatibility, it sets the primary fqdn to minion ID by default - that is the FQDN used for bootstrapping of existing minions.

Bootstapping

This PR allows to choose particular proxy FQDN, that is used for minion->proxy connection.

## Codespace

## GUI diff

Before:
Proxy selector shows one random FQDN per proxy.

After:
Proxy selector shows primary FQDN followed by indented additional FQDNs.
It is possible to enter additional FQDNs when creating proxy config.

- [x] **DONE**

## Documentation
- TBD
- [x] **DONE**

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/30754

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

## Re-run a test

# Before you merge

### DIFF
--- a/containers/proxy-httpd-image/uyuni-configure.py
+++ b/containers/proxy-httpd-image/uyuni-configure.py
@@ -198,6 +198,10 @@ with open(config_path + "httpd.yaml", encoding="utf-8") as httpdSource:
         smlm_conf.write(file_content)
         smlm_conf.truncate()
 
+    aliases = ""
+    if "additional_fqdns" in config and config["additional_fqdns"]:
+        aliases = "\n\tServerAlias " + " ".join(config["additional_fqdns"])
+
     with open("/etc/apache2/vhosts.d/ssl.conf", "w", encoding="utf-8") as file:
         file.write(f"""
 <IfDefine SSL>
@@ -205,7 +209,7 @@ with open(config_path + "httpd.yaml", encoding="utf-8") as httpdSource:
 <VirtualHost _default_:443>
 
 	DocumentRoot "/srv/www/htdocs"
-	ServerName {config["proxy_fqdn"]}
+	ServerName {config["proxy_fqdn"]}{aliases}
 
 	ErrorLog /proc/self/fd/2
 	TransferLog /proc/self/fd/1
--- a/java/core/src/main/java/com/redhat/rhn/domain/server/MinionServer.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/server/MinionServer.java
@@ -315,8 +315,9 @@ public class MinionServer extends Server implements SaltConfigurable {
                 // the system is connected to a proxy
                 // check if serverPath already exists
                 Optional<ServerPath> path = ServerFactory.findServerPath(this, proxy.get());
-                if (path.isEmpty() || path.get().getPosition() != 0) {
-                    // proxy path does not exist -> create it
+                if (path.isEmpty() || path.get().getPosition() != 0 ||
+                        !path.get().getHostname().equals(hostname.orElse(proxy.get().getHostname()))) {
+                    // proxy path does not exist or hostname changed -> create it
                     Set<ServerPath> proxyPaths = ServerFactory.createServerPaths(this, proxy.get(),
                                                  hostname.orElse(proxy.get().getHostname()));
                     getServerPaths().clear();
--- a/java/core/src/main/java/com/redhat/rhn/domain/server/MinionServer.java
+++ b/java/core/src/main/java/com/redhat/rhn/domain/server/MinionServer.java
@@ -110,6 +110,20 @@ public class MinionServer extends Server implements SaltConfigurable {
     }
 
     /**
+     * Return the name of the primary FQDN, falling back to the minion id if none is known.
+     *
+     * This is used for connection from proxy to minion for Salt SSH
+     *
+     * @return the primary FQDN name or the minion id
+     */
+    @Override
+    public String getPrimaryFqdnName() {
+        return Optional.ofNullable(findPrimaryFqdn())
+                .map(ServerFQDN::getName)
+                .orElseGet(this::getMinionId);
+    }
+
+    /**
      * Gets kernel live version.
      *
      * @return the kernel live version
--- a/java/core/src/main/java/com/suse/manager/webui/controllers/ProxyController.java
+++ b/java/core/src/main/java/com/suse/manager/webui/controllers/ProxyController.java
@@ -47,11 +47,12 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.stream.Collectors;
+import java.util.Optional;
 
 import spark.ModelAndView;
 import spark.Request;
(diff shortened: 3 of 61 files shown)


java/core/src/main/java/com/suse/manager/webui/controllers/ProxyController.java
```diff
@@ -56,6 +62,46 @@
...
+            }
+            ParentRepresentation other = (ParentRepresentation) obj;
+            return java.util.Objects.equals(this.primaryFqdn, other.primaryFqdn);
+        }
+
+        @Override
+        public int hashCode() {
+            return java.util.Objects.hash(primaryFqdn);
+        }
+    }
```
mackdk, line 103: This class can be `record` to reduce a bit the boilerplate:
```java
    private record ParentRepresentation(String primaryFqdn, List<String> additionalFqdns)
            implements Comparable<ParentRepresentation> {

        @Override
            public int compareTo(ParentRepresentation o) {
                return this.primaryFqdn.compareTo(o.primaryFqdn);
            }
    
            @Override
            public boolean equals(Object obj) {
                if (this == obj) {
                    return true;
                }
                if (!(obj instanceof ParentRepresentation other)) {
                    return false;
                }
                return Objects.equals(this.primaryFqdn, other.primaryFqdn);
            }
    
            @Override
            public int hashCode() {
                return Objects.hash(primaryFqdn);
            }
        }
```

mackdk (dismissed):
LGTM. 

I think the code duplication among `bootstrapByProxyFqdn`, `bootstrapWithPrivateSshKeyByProxyFqdn`, `bootstrapByProxyFqdn` and `bootstrapWithPrivateSshKeyByProxyFqdn` could be further reduced by introducing a private method that takes all the parameters and delegates to the `maybe...` methods the handling of the `null` ones.

containers/proxy-httpd-image/uyuni-configure.py
```diff
@@ -198,14 +198,18 @@ def get_server_version(fqdn):
...
+
     with open("/etc/apache2/vhosts.d/ssl.conf", "w", encoding="utf-8") as file:
         file.write(f"""
 <IfDefine SSL>
 <IfDefine !NOSSL>
 <VirtualHost _default_:443>
 
 	DocumentRoot "/srv/www/htdocs"
-	ServerName {config["proxy_fqdn"]}
+	ServerName {config["proxy_fqdn"]}{aliases}
```
aaannz, line 212: A nitpick, but I first missed there it as `\n\tServerAlias` in the variable.
Why not have this on the separate line so it is clearly visible {aliases} is not an additional parameter to servername?
We then can drop `\n\t` from the alias start.

java/core/src/main/java/com/redhat/rhn/domain/server/MinionServer.java
```diff
@@ -109,6 +109,20 @@ public void setMinionId(String minionIdIn) {
...
+     * This is used for connection from proxy to minion for Salt SSH
+     *
+     * @return the primary FQDN name or the minion id
+     */
+    @Override
+    public String getPrimaryFqdnName() {
+        return Optional.ofNullable(findPrimaryFqdn())
+                .map(ServerFQDN::getName)
+                .orElseGet(this::getMinionId);
+    }
```
aaannz, line 124: I think this is worth a note in the release notes.

Also shouldn't we check if we can use this as is?
Minion ID can be some pretty arbitrary string so IMO we should check if it conforms to the DNS naming.

java/core/src/main/java/com/redhat/rhn/domain/server/MinionServer.java
```diff
@@ -109,6 +109,20 @@ public void setMinionId(String minionIdIn) {
...
 
+    /**
+     * Return the name of the primary FQDN, falling back to the minion id if none is known.
+     *
+     * This is used for connection from proxy to minion for Salt SSH
+     *
+     * @return the primary FQDN name or the minion id
+     */
+    @Override
+    public String getPrimaryFqdnName() {
```
aaannz, line 120: Very nitpicky 😄 but this function name is now `getPrimaryFullyQualifiedDomainNameName()`

And similarly below.

java/core/src/main/java/com/suse/manager/webui/controllers/ProxyController.java
```diff
@@ -99,6 +145,41 @@ public void initRoutes(ProxyController proxyController, JadeTemplateEngine jade)
...
+        for (ParentRepresentation rep : parentReps) {
+            Map<String, String> primaryOpt = new HashMap<>();
+            primaryOpt.put("value", rep.getPrimaryFqdn());
+            primaryOpt.put("label", rep.getPrimaryFqdn());
+            parentOptions.add(primaryOpt);
+
+            for (String addFqdn : rep.getAdditionalFqdns()) {
+                Map<String, String> addOpt = new HashMap<>();
+                addOpt.put("value", addFqdn);
+                addOpt.put("label", "\u00A0\u00A0\u00A0\u00A0" + addFqdn);
```
aaannz, line 176: This looks strange. Why putting nbsp here to the data?

schema/spacewalk/upgrade/susemanager-schema-5.3.1-to-susemanager-schema-5.3.2/104-add-bootstrap-by-proxy-fqdn-endpoints.sql
```diff
@@ -0,0 +1,27 @@
+INSERT INTO access.namespace (namespace, access_mode, description)
+    SELECT 'api.system.bootstrap_by_proxy_fqdn', 'W', 'Bootstrap a system for management via either Salt or Salt SSH using a proxy FQDN.'
+    WHERE NOT EXISTS (SELECT 1 FROM access.namespace WHERE namespace = 'api.system.bootstrap_by_proxy_fqdn' AND access_mode = 'W');
+
+INSERT INTO access.endpoint (class_method, endpoint, http_method, scope, auth_required)
+    SELECT 'com.redhat.rhn.frontend.xmlrpc.system.SystemHandler.bootstrapByProxyFqdn', '/manager/api/system/bootstrapByProxyFqdn', 'POST', 'A', True
+    WHERE NOT EXISTS (SELECT 1 FROM access.endpoint WHERE endpoint = '/manager/api/system/bootstrapByProxyFqdn' AND http_method = 'POST');
```
aaannz, line 7: Why these internal selects when the same could be done like in original fixtures:
`ON CONFLICT (namespace, access_mode) DO NOTHING;`

web/html/src/manager/proxy/container-config.tsx
```diff
@@ -202,16 +208,16 @@ export function ProxyConfig({ noSSL }: { noSSL: boolean }) {
           validators={[Validation.matches(/^[A-Za-z0-9-]+(\.[A-Za-z0-9-]+)*$/)]}
           invalidHint={t("Has to be a valid FQDN address")}
         />
-        <Text
+        <DEPRECATED_Select
```
aaannz, line 211: What is the difference between `DEPRECATED_Select` and `Select`?

PR #12435, opened 2026-08-12
Fix playwright-ruby-client Transport to prevent indefinite test run hangs

maximenoel8:
## What does this PR change?

Adds a monkey-patch for playwright-ruby-client 1.60.0 (`features/support/playwright_transport_patch.rb`) that fixes three bugs in `Playwright::Transport` which together caused the testsuite to hang indefinitely — until the Jenkins 10-hour job timeout — whenever `Timeout.timeout` fired inside `repeat_until_timeout` during a Playwright call:

**Bug 1 — Torn protocol frames (root cause of the `SyntaxError` crash):**
`send_message` wrote the 4-byte LE length header and JSON payload as two separate `IO#write` calls. A `Timeout::Error` delivered via `Thread#raise` (as used by `Timeout.timeout`) could land between them, leaving the Node.js driver with a partial frame on stdin. It then crashed its message loop with `SyntaxError: Unexpected token ... is not valid JSON` (Node >= 16 banner: `<anonymous_script>:1`). Fixed by merging both writes into one buffer and shielding the write with `Thread.handle_interrupt(Object => :never)`.

**Bug 2 — Clean EOF not signalled (direct cause of the 10-hour hang):**
`handle_stdout`'s while loop exited silently when `read(4)` returned nil on EOF (normal driver exit after the watchdog's SIGKILL). `on_driver_closed` was never called in the normal-exit path, so all pending Ruby callbacks waited forever. Fixed by calling `on_driver_closed` after the loop exits.

**Bug 3 — Modern Node.js crash not detected:**
`handle_stderr` only matched the Node <= 14 crash banner `'undefined:1'`. Node >= 16 prints `'<anonymous_script>:1'`, so `on_driver_crashed` never fired and pending callbacks were again left hanging. Fixed by matching both banners.

Also:
- Adds a 10-second `Timeout.timeout` backstop in the After hook's `web_session_is_active?` call as a defence-in-depth layer.
- Switches `repeat_until_timeout`'s inner deadline loop from `Time.new` (wall clock) to `Process.clock_gettime(Process::CLOCK_MONOTONIC)` so NTP adjustments on long-running test hosts cannot skew the timeout.

A `TODO:` comment in the patch file links upstream for removal once fixed in playwright-ruby-client itself.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

No documentation needed.

- [x] **DONE**

## Test coverage

- [x] **DONE**

## Links

Issue(s): 
Port(s): 
  - 5.2: https://github.com/SUSE/spacewalk/pull/31638
  - 5.1: https://github.com/SUSE/spacewalk/pull/31639
- Upstream playwright-ruby-client issue: https://github.com/YusukeIwaki/playwright-ruby-client/issues/392

- [x] **DONE**

## Changelogs

- [x] No changelog needed

### DIFF
--- a/testsuite/features/support/commonlib.rb
+++ b/testsuite/features/support/commonlib.rb
@@ -110,9 +110,9 @@ def repeat_until_timeout(timeout: DEFAULT_TIMEOUT, retries: nil, message: nil, r
       # At the time of writing some of the problems described have been addressed.
       # However, at least https://bugs.ruby-lang.org/issues/15886 remains reproducible and code below
       # works around it by adding an additional check between loops
-      start = Time.new
+      deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
       attempts = 0
-      while (Time.new - start <= timeout) && (retries.nil? || attempts < retries)
+      while Process.clock_gettime(Process::CLOCK_MONOTONIC) <= deadline && (retries.nil? || attempts < retries)
         last_result = yield
         attempts += 1
       end
--- a/testsuite/features/support/env.rb
+++ b/testsuite/features/support/env.rb
@@ -172,10 +172,17 @@ After do |scenario|
   log "This scenario took: #{current_epoch - @scenario_start_time} seconds"
   if scenario.failed?
     begin
-      # web_session_is_active? can raise if the session went stale after a long step.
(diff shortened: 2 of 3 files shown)


Bischoff (approved):
Cool, thanks.

I suggest to hammer them upstream until they fix the upstream bug and we can remove the monkeypatch.

PR #12447, opened 2026-08-13
feat(openapi): migrate six handlers

masterhj:
Migrates the six namespaces that #12433 unblocks


wweellddeerr:
I noticed that several response fields are missing `requiredMode` in `DeltaImageDoc`, `FileListDoc` / `FilePreservationDoc` and `CustomInfoKeyDoc`, please check it.

You should also improve the contract tests for `ScheduleHandlerContractTest`, only 8 of the 15 methods are covered.

wweellddeerr (approved):

PR #12460, opened 2026-08-18
Accept str DISTTAG in RPM_Header.modularity_label()

ilcalimero:
## What does this PR change?

`RPM_Header.modularity_label()` falls back to the RPM `DISTTAG` tag to detect
modular packages, comparing it against the bytes literal `b"module("`. The rpm
Python bindings shipped with Python 3.13 return that tag as `str`, so the
comparison raises:

```
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

The exception is not caught and it kills `spacewalk-repo-sync`: the channel log
stops right after `Regenerating bootstrap repositories` without ever reaching
`Sync completed.`, and the patches of the channel are then reported as
`not found in database ... skipped`. Nothing in the log mentions RPM headers or
modularity, which makes the failure look like a network or database problem.

Only channels carrying packages tagged with `DISTTAG` are affected, in practice
the modular AppStream repositories of the EL8 clones. Seen on Uyuni 5.3.2
(server image, Python 3.13) syncing Oracle Linux 8 AppStream.

The fix normalizes the tag to `str` before the comparison, so that both the old
(bytes) and the new (str) bindings are handled. The returned value keeps the
type of the original tag, so the behaviour of the callers does not change.

## GUI diff

No difference.

- [x] **DONE**

## Documentation

- [x] **DONE**

## Test coverage

- Unit tests were added

`python/test/unit/uyuni/test_rhn_rpm.py` covers `modularity_label()` with
`DISTTAG` as both `str` and `bytes`, with a non-modular `DISTTAG`, with
`MODULARITYLABEL` taking precedence over `DISTTAG`, and with no modularity tag
at all. The parametrized `str` case fails with the `TypeError` above on the
unpatched code.

- [x] **DONE**

## Links

Issue(s): #
Port(s): #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

`python/uyuni/uyuni-common-libs.changes.<user>.disttag-str` added.

### DIFF
--- a/python/test/unit/uyuni/test_rhn_rpm.py
+++ b/python/test/unit/uyuni/test_rhn_rpm.py
@@ -1,0 +1,77 @@
+#  pylint: disable=missing-module-docstring,invalid-name
+#
+# Copyright (c) 2026 SUSE LLC
+#
+# This software is licensed to you under the GNU General Public License,
+# version 2 (GPLv2). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+# along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+
+try:
+    import rpm
+except ImportError:
+    # The rpm bindings are a native module that is not always available where
+    # the unit tests run. Only the two tags read by modularity_label() are
+    # needed here, so stub the module with their real values.
+    rpm = MagicMock()
+    rpm.RPMTAG_MODULARITYLABEL = 5096
+    rpm.RPMTAG_DISTTAG = 1155
+    sys.modules["rpm"] = rpm
+
+# pylint: disable-next=wrong-import-position
+from uyuni.common.rhn_rpm import RPM_Header
+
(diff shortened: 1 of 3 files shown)


ilcalimero:
Hi, a gentle ping on this one. The workflows are still waiting for maintainer
approval (first-time contributor gate), so the CI has not run yet on the
current commit.

For context: the only check that failed on the first run was `Changelogs`,
because the changelog lines exceeded 67 characters. That is fixed, and I
verified it locally with `.github/workflows/changelogs/changelogs.py`. The unit
test was also updated to stub the `rpm` bindings, which are not installed on
the runners.

The change itself is a two-line type normalization in `modularity_label()`:
with the Python 3.13 rpm bindings the DISTTAG tag comes back as `str`, the
comparison against `b"module("` raises a `TypeError`, and
`spacewalk-repo-sync` dies without ever logging `Sync completed.` on channels
with modular packages (EL8 AppStream clones). Happy to adjust anything if you
would like it done differently.

ilcalimero:
Hi! Just a friendly nudge on this one — no rush at all, I know everyone has
plenty on their plate.

The CI has been green since 28 August. I re-checked today: `master` has moved
on by around 350 commits in the meantime, but none of them touched
`rhn_rpm.py`, so the branch still merges cleanly and the change is unchanged.

If it helps, I am happy to rebase it, reshape it, or adjust anything you would
prefer done differently — just say the word and I will follow up. And thanks
for all the work you put into Uyuni!

